PR #10813, opened 2019-11-18
Fix crash on access to non-initialized lateinit property

planarvoid:
This fixes part of this issue #10019
It's a crash in the HistoryViewModel where an non-initialized lateinit property is accessed. The crash is linked from the original issue: crash: https://sentry.io/share/issue/7d4c3139eb874b45a4b27b0dd3ef97c1/

To test:
* Test that the Post/History works as expected
* Try to rotate the device
* The data is not refreshed
* Try to switch offline/online
* The data is refreshed on going back online

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/HistoryListFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/HistoryListFragment.kt
@@ -45,7 +45,7 @@ class HistoryListFragment : Fragment() {
     }
 
     interface HistoryItemClickInterface {
-        fun onHistoryItemClicked(revision: Revision, revisions: ArrayList<Revision>)
+        fun onHistoryItemClicked(revision: Revision, revisions: List<Revision>)
     }
 
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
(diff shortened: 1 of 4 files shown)


WordPress/src/main/java/org/wordpress/android/ui/posts/HistoryListFragment.kt
```diff
@@ -147,9 +147,12 @@ class HistoryListFragment : Fragment() {
             }
         })
 
-        viewModel.showDialog.observe(this, Observer {
-            if (it is Revision && activity is HistoryItemClickInterface) {
-                (activity as HistoryItemClickInterface).onHistoryItemClicked(it, viewModel.revisionsList)
+        viewModel.showDialog.observe(this, Observer { showDialogItem ->
+            if (showDialogItem.historyListItem is Revision && activity is HistoryItemClickInterface) {
```
malinajirka: `showDialogItem` is nullable, we might want to consider checking that it's not null or adding requireNoNull. Wdyt?

SideNote:
Since we are touching this code, we might want to consider moving the `activity is HistoryItemClickInterface` check into onAttach and throwing an exception when the activity doesn't implement mentioned interface. Wdyt?

WordPress/src/main/java/org/wordpress/android/viewmodel/history/HistoryViewModel.kt
```diff
@@ -81,30 +81,33 @@ class HistoryViewModel @Inject constructor(
...
         })
-    }
-
-    fun create(localPostId: Int, site: SiteModel) = launch {
-        if (isStarted) {
-            return@launch
-        }
-
-        val post: PostModel? = withContext(bgDispatcher) { postStore.getPostByLocalPostId(localPostId) }
+        launch {
```
malinajirka: It seems the default dispatche is `Dispatchers.Default` - do you really want to perform all these operations in a background thread?

malinajirka: Ohh I noticed that it was an existing behavior before these changes. I'm wondering how come it doesn't crash on `[liveadata].value = `. I thought it crashes when performed on a background thread 😕 .

malinajirka: Ok ignore me... `ScopedViewModel` contains a variable called `defaultDispatcher` but we actually pass main dispatcher there... We might want to rename that variable as it's quite confusing:).

planarvoid:
thanks for the review @malinajirka! I've implemented your suggestions + added an extra non-null check to the other observer so this should be ready for another check

malinajirka:
I've reviewed+tested the changes and LGTM ;) Thanks

PR #10819, opened 2019-11-19
Update FluxC version

malinajirka:
To test:
- CI should be enough

PR submission checklist:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -106,5 +106,5 @@ buildScan {
 
 ext {
     daggerVersion = '2.22.1'
-    fluxCVersion = 'f57b2b2f8b884efbdc6df7ba8f7de8aec3fac6ea'
+    fluxCVersion = '1.5.3-beta-3'
 }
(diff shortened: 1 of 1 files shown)


mchowning:
I think this `1.5.3-beta-3` tag is confusing because it is a commit _after_ the `1.5.3` release. Seems like we should give it a `1.5.4-beta-1` tag or just use the commit hash (and probably delete the `1.5.3-beta-2` and `1.5.3-beta-3` fluxc tags.

malinajirka:
Thanks for noticing this @mattchowning! 
I've removed the 1.5.3-beta-3 and created a new 1.5.4-beta-1 tag. I'm not able to remove the `1.5.3-beta-2` tag, it keeps failing with 
```
Failed to delete tag 1.5.3-beta-2 on remote: failed to push some refs to 'git@github.com:wordpress-mobile/WordPress-FluxC-Android.git'
```.
I asked on Slack if someone could remove it for me, so it should be gone soon.

PR #10820, opened 2019-11-19
Create a Sequencing snackbar feature

### LINKED ISSUE #10771: Failed upload Snackbar shown after successful upload
### Expected behavior
When I publish a post and the upload succeeds a positive Snackbar is shown.

### Actual behavior
When I successfully publish a post but there is another upload which fails, the failed upload snackbar is shown which is very confusing.

1. Switch to a site where you are in Contributor role.
2. Create a new post and add an image
3. This post will always fail to upload since you don't have permissions to upload an image as a contributor (this reproduction steps will work until we fix #10666 )
4. Now create a new post without an image and publish it
5. Notice the UploadService publishes the post, but also retries to upload the failing post -> the failed upload message is shown which is very confusing since the upload from step 2 actually succeeded

Emulator API 29

develric:
Partially Fixes #10771 

This is a draft, uploaded to make easier to elaborate/evaluate on the approach to solve the issue. Some notes/doubts:

- coroutine improvements in SnackbarSequencer: for example job cancellation?
- modifications/improvements on estimateNextAvailableTimeSlot algorithm
- Snackbar.LENGTH_INDEFINITE: for now those snackbar not added to the sequencer; decide if trigger an exception in case they are accidentally added or silently overwrite with some default
- ...

To test:

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/util/SnackbarSequencer.kt
+++ b/WordPress/src/main/java/org/wordpress/android/util/SnackbarSequencer.kt
@@ -1,0 +1,91 @@
+package org.wordpress.android.util
+
+import com.google.android.material.snackbar.Snackbar
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import org.wordpress.android.util.AppLog.T
+import org.wordpress.android.widgets.WPSnackbar
+import javax.inject.Inject
+import javax.inject.Singleton
+import kotlin.coroutines.CoroutineContext
+
+private const val DELAY_LOWER_MARGIN_MS: Long = 1000
+private const val DELAY_HIGHER_MARGIN_MS: Long = 20000
+private const val CLOCK_BEHIND_MARGIN_MS: Long = 2 * DELAY_HIGHER_MARGIN_MS
+
+// Taken from com.google.android.material.snackbar.SnackbarManager.java
+// Did not find a way to get them directly from the android framework for now
+private const val SHORT_DURATION_MS = 1500
+private const val LONG_DURATION_MS = 2750
+
+@Singleton
+class SnackbarSequencer @Inject constructor() : Snackbar.Callback(), CoroutineScope {
+    private var job: Job = Job()
+
+    private var nextAvailableTimeSlot: Long = 0
+    private var nextDelay: Long = 0
+
+    override val coroutineContext: CoroutineContext
+        get() = Dispatchers.Default + job
+
+    private suspend fun schedule(snackbar: WPSnackbar, delay: Long) {
+        delay(delay)
+        withContext(Dispatchers.Main) {
+            AppLog.d(T.UTILS, "SnackbarSequencer > show snackbar [${snackbar.message}]")
+            snackbar.show()
+        }
+    }
+
+    @Synchronized
+    private fun estimateNextAvailableTimeSlot(durationInMs: Int, snackbarCreationTimestamp: Long): Long {
+        val currTime = System.currentTimeMillis()
+        val lastAvailableTimeSlot = nextAvailableTimeSlot
+        val lastDelay = nextDelay
+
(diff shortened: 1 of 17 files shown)


WordPress/src/main/java/org/wordpress/android/util/SnackbarSequencer.kt
```diff
@@ -0,0 +1,91 @@
...
+private const val DELAY_HIGHER_MARGIN_MS: Long = 20000
+private const val CLOCK_BEHIND_MARGIN_MS: Long = 2 * DELAY_HIGHER_MARGIN_MS
+
+// Taken from com.google.android.material.snackbar.SnackbarManager.java
+// Did not find a way to get them directly from the android framework for now
+private const val SHORT_DURATION_MS = 1500
+private const val LONG_DURATION_MS = 2750
+
+@Singleton
+class SnackbarSequencer @Inject constructor() : Snackbar.Callback(), CoroutineScope {
```
malinajirka: Thanks for looking into this issue @develric! ;) I realize it's still just a draft PR so I haven't reviewed the code line by line, I just looked at the overall approach. 
My first thought when we discussed this issue was having a queue of messages waiting to be shown (with a limited capacity) and [adding onDismiss callback](https://bit.ly/2D26Q1P) in which we'd check if the queue is empty or if we should show a SnackBar. I'm not sure if it would work or which approach is better, I just wanted to share my thoughts so we can both think about it. Wdyt?
One thing I realized while reviewing this code is that we need to be extra careful not to leak any activities/views. `WPSnackbar` holds a reference to the view == activity. If we decide to go ahead with either of the approaches, we should make sure we keep just a WeakReference to the context (making sure the optional action callbacks don't leak the context might be even more tricky). Wdyt?

develric: Hi @malinajirka and thanks for keeping an eye on this (much appreciated 🙇‍♂️). I try to give some more detail below hoping it helps on further discussion/elaboration. Let me know wdyt 😊.

>My first thought when we discussed this issue was having a queue of messages...

This was the first approach I tried actually. What I got tough is that at least on the emulator with my mac a bit loaded I suspect android missed some callbacks event and I ended up having messages in queue that would stay there until some (if any) next snackbar. So in case we decide to go with this approach I'm not sure it's enough to rely on the callback as unique mechanism to poll from the queue 🤷‍♂️.
The intention of what I tried to implement was to introduce some scheduled delay in between the snackbars show and to saturate it just to not make the snackbar show up too far from the timestamp when it was generated (basic attempt in the estimateNextAvailableTimeSlot that for sure we can improve on). In this way we should not be worried on callbacks to be skipped even if of course it can be less precise on snackbar duration and can basically fall in something similar to the snackbar overlapping we are having now in cases where we have many snachbars (due to delay saturation in estimateNextAvailableTimeSlot). Maybe one possibility in those cases can be to "emulate" the queue with loss you were suggesting skipping the snackbar show (but I wonder if it can make more sense to keep a small timing sequencing even if not lasting for all the desired snackbar duration?).

> we need to be extra careful not to leak any activities/views.
Absolute good point! Thanks!

malinajirka: Thanks for the explanation! ❤️ 

>This was the first approach I tried actually. What I got tough is that at least on the emulator with my mac a bit loaded I suspect android missed some callbacks event and I ended up having messages in queue that would stay there until some (if any) next snackbar. So in case we decide to go with this approach I'm not sure it's enough to rely on the callback as unique mechanism to poll from the queue 🤷‍♂️.

ohh, I wouldn't expect it to miss the callbacks 🤔. Perhaps it misses them when the activity isn't resumed or something like that 😞 🤷‍♂ . In that case your approach seems reasonable:).


You mentioned you have concerns about changing the behavior in the entire app. I can't say I'm feeling differently. One thought which just crossed my mind would be handle the specific scenario we are having troubles with manually - show a snackbar when an upload fails and somehow replace it with a snackbar saying something like "1 Upload succeeded and 1 upload failed`. We already do a similar thing in Notifications, they have unique ids though 🤷‍♂.  => I'm not saying we should do it, again just something we might want to think about.

develric:
Closing in favor of #10856

PR #10826, opened 2019-11-21
Fix null site model in PostListActivity

### LINKED ISSUE #8363: Rare NPE crash in ActivityLogListFragment
```
Caused by kotlin.TypeCastException: null cannot be cast to non-null type org.wordpress.android.fluxc.model.SiteModel at 
org.wordpress.android.ui.activitylog.list.ActivityLogListFragment.onViewCreated(ActivityLogListFragment.kt:57)
```

5ba662b6f8b88c2963e156c0-fabric

planarvoid:
Fixes #8363

We're still passing nullable `SiteModel` to the intent when creating activities. This PR should fix that when creating the `PostListActivity`. All the places should now be checked for non-null `SiteModel`. I don't know which solution is the best. 

`viewCurrentBlogPages` had already `NonNull` parameter but we were still passing a nullable SiteModel. That's why is my solution different than the one with `viewCurrentBlogPosts`.

To test:

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.java
```diff
@@ -619,7 +619,11 @@ private void viewPosts() {
 
     private void viewPages() {
         requestNextStepOfActiveQuickStartTask();
-        ActivityLauncher.viewCurrentBlogPages(requireActivity(), getSelectedSite());
+        SiteModel selectedSite = getSelectedSite();
+        if (selectedSite != null) {
+            ToastUtils.showToast(getActivity(), R.string.site_cannot_be_loaded);
```
malinajirka: Shouldn't this line be in an `else` clause? Wdyt?

planarvoid: Good catch, I've fixed it in https://github.com/wordpress-mobile/WordPress-Android/pull/10826/commits/60bcddc5f23f2dbf7d73d95142fbf13964814ea6

planarvoid:
@malinajirka thanks for the check! I've fixed what you found and it should be ready for another check.

PR #10834, opened 2019-11-22
Reader PhotoViewer no longer crashes with null imgUrl

### LINKED ISSUE #10831: IllegalArgumentException: Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkParameterIsNo...
Sentry Issue: [WORDPRESS-ANDROID-2MK](https://sentry.io/organizations/a8c/issues/1098054051/?referrer=github_integration)

```
IllegalArgumentException: Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkParameterIsNotNull, parameter imgUrl
    at org.wordpress.android.util.image.ImageManager.loadWithResultListener
    at org.wordpress.android.ui.reader.views.ReaderPhotoView.loadImage(ReaderPhotoView.java:108)
    at org.wordpress.android.ui.reader.views.ReaderPhotoView.onLayout(ReaderPhotoView.java:193)
...
(44 additional frame(s) were not displayed)

Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkParameterIsNotNull, parameter imgUrl
```

jd-alexander:
Fixes #10831

To test:
1. Simulate the `imgUrl` as being `null`.
2. Go to Reader. 
3. Click on an article.
4. Click on an image.
5. The app shouldn't crash. 

For further discussion, it just shows a loading spinner. 

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/views/ReaderPhotoView.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/views/ReaderPhotoView.java
@@ -190,7 +190,9 @@ public class ReaderPhotoView extends RelativeLayout {
             if (mIsInitialLayout) {
                 mIsInitialLayout = false;
                 AppLog.d(AppLog.T.READER, "reader photo > initial layout");
-                loadImage();
+                if (mLoResImageUrl != null && mHiResImageUrl != null) {
+                    loadImage();
+                }
             }
         }
     }
--- a/WordPress/src/main/java/org/wordpress/android/util/image/ImageManager.kt
+++ b/WordPress/src/main/java/org/wordpress/android/util/image/ImageManager.kt
@@ -142,24 +142,22 @@ class ImageManager @Inject constructor(private val placeholderManager: ImagePlac
     fun loadWithResultListener(
         imageView: ImageView,
         imageType: ImageType,
-        imgUrl: String?,
+        imgUrl: String,
         scaleType: ScaleType = CENTER,
         thumbnailUrl: String? = null,
         requestListener: RequestListener<Drawable>
     ) {
         val context = imageView.context
         if (!context.isAvailable()) return
-        imgUrl?.let {
-            GlideApp.with(context)
-                    .load(it)
-                    .addFallback(imageType)
-                    .addPlaceholder(imageType)
-                    .addThumbnail(context, thumbnailUrl, requestListener)
-                    .applyScaleType(scaleType)
-                    .attachRequestListener(requestListener)
-                    .into(imageView)
-                    .clearOnDetach()
-        }
+        GlideApp.with(context)
+                .load(imgUrl)
+                .addFallback(imageType)
+                .addPlaceholder(imageType)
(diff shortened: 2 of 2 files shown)


malinajirka:
Thanks @jd-alexander! 

The solution works as expected. However, I like to follow the Fail-Fast principle. Changing the non-null type to nullable prevents the app from crashing, but we'll never know if the users can see their images (and that we have bugs in the app). In an ideal world, everything would be written in Kotlin and we wouldn't have this problem. We are not there yet, but I think it'd be nice to try to keep the interfaces as clean as possible. Wdyt?

ReaderPhotoView.loadImage() is invoked from two places 
1. setImageUrl(..) -> which invokes `mHiResImageUrl = ReaderUtils.getResizedImageUrl(` which always returns non-null string -> this scenario is safe.
2. `onLayout(..)` -> this is apparently the scenario in which the mHiResImageUrl is null (stack trace in Sentry confirms it).

How would you feel about modifying `ReaderPhotoViewerFragment` so it's never opened with a null imgUrl -> we could show a toast message and log an exception into Sentry in case someone tries to open it with a null imgUrl. This solution would prevent the app from crashing. Moreover it'd help us investigate how come we are opening the fragment with a null url. 
Another solution which is simpler and is ok imo is to check we don't call `loadImage` from `onLayout` unless the url is set. We'd still just hide the bug we have, but we wouldn't change the interface of ImageManager which is used all over the app and could potential hide multiple bugs. Wdyt?

jd-alexander:
> Thanks @jd-alexander!

> The solution works as expected. However, I like to follow the Fail-Fast principle. Changing the non-null type to nullable prevents the app from crashing, but we'll never know if the users can see their images (and that we have bugs in the app). In an ideal world, everything would be written in Kotlin and we wouldn't have this problem. We are not there yet, but I think it'd be nice to try to keep the interfaces as clean as possible. Wdyt?

Thanks for the review! Let's see what can be done. 

> ReaderPhotoView.loadImage() is invoked from two places

> setImageUrl(..) -> which invokes mHiResImageUrl = ReaderUtils.getResizedImageUrl( which always returns non-null string -> this scenario is safe.
onLayout(..) -> this is apparently the scenario in which the mHiResImageUrl is null (stack trace in Sentry confirms it).

> How would you feel about modifying ReaderPhotoViewerFragment so it's never opened with a null imgUrl -> we could show a toast message and log an exception into Sentry in case someone tries to open it with a null imgUrl. This solution would prevent the app from crashing. Moreover it'd help us investigate how come we are opening the fragment with a null url.

This sounds like a good solution as it preserving the ability to be notified of the bug and the user is being warned.

> Another solution which is simpler and is ok imo is to check we don't call loadImage from onLayout unless the url is set. We'd still just hide the bug we have, but we wouldn't change the interface of ImageManager which is used all over the app and could potential hide multiple bugs. Wdyt?

 I would feel more comfortable with the first solution as the user isn't left wondering as to what's happening when the bug occurs and yes, the `ImageManager`'s API would be restored since having no else on conditions or error logic would hide other bugs.

PR #10848, opened 2019-11-25
Stop hiding data when there is no connection

### LINKED ISSUE #10836: Stats widgets are showing "No network available" after some time
### Expected behavior
Stats widgets will refresh when the network is available.

### Actual behavior
Stats widgets, even with a constant WiFi and/or cellular connection, will randomly show "No network available".

image000000

I can't reliable reproduce it - just add a widget and let the phone sit locked for a while?

planarvoid:
Fixes #10836

The idea behind this change is to not allow the `error` state of no connection to override the actual data shown. The widget refreshes the data periodically and I think there might actually be a fake no-connection response. I couldn't actually reproduce this error. 

Once you have a widget data, we save a flag into the shared prefs for the given widget and never replace the data with the error (if the user still has the access token). 

To test:
* Try to add/remove/change Today widget for various sites
* Lock the phone
* Check that it does not hide the data when you come back

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/prefs/AppPrefs.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/prefs/AppPrefs.java
@@ -128,6 +128,7 @@ public class AppPrefs {
         STATS_WIDGET_SELECTED_SITE_ID,
         STATS_WIDGET_COLOR_MODE,
         STATS_WIDGET_DATA_TYPE,
+        STATS_WIDGET_HAS_DATA,
 
         // Keep the local_blog_id + local_post_id values that have HW Acc. turned off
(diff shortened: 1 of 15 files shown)


develric (approved):
Hi @planarvoid , the approach seems reasonable and the code looks good! 👍

I was not able either to fully reproduce it, but I verified the desired difference in behavior with the following steps:

- Activate the Today widget and check that it shows the correct today stats
- turn on air mode
- click on the widget to open the stats in the app
- come back with back button to the home screen

Doing the above you get the no network message before your modification whereas the stats are preserved as it should be using your modification. AFAIU this is the correct expected behavior, so good job! 😎

PR #10850, opened 2019-11-25
Fixing NoSuchElement list empty in PostDayViewsUseCase

### LINKED ISSUE #10830: NoSuchElementException: List is empty.
Sentry Issue: [WORDPRESS-ANDROID-5RK](https://sentry.io/organizations/a8c/issues/1265129016/?referrer=github_integration)

```
NoSuchElementException: List is empty.
    at org.wordpress.android.ui.stats.refresh.lists.detail.PostDayViewsUseCase.buildUiModel(PostDayViewsUseCase.kt:80)
    at org.wordpress.android.ui.stats.refresh.lists.detail.PostDayViewsUseCase.buildUiModel(PostDayViewsUseCase.kt:25)
    at org.wordpress.android.ui.stats.refresh.lists.sections.BaseStatsUseCase$updateState$2.invokeSuspend(BaseStatsUseCase.kt:197)
...
(5 additional frame(s) were not displayed)

List is empty.
```

Introduced in 13.3 - seems like an easy win.

develric:
Fixes #10830

This PR does two things:

1. Attempt to fix issue #10830: was not actually able to reproduce but applied same approach as used in `OverviewUseCase.buildUiModel` (that is introduced a check for the list not being empty)
2. In `PostDayViewsUseCase.fetchRemoteData` in case of error I used `selectedDateProvider.onDateLoadingFailed` insetad of `selectedDateProvider.onDateLoadingSucceeded`

## To test:

Since it was not possible to reproduce, I think we can only make sure we did not introduce issues in the relevant stats card, so smoke test the post day/views card with different sites/posts and compare with the not modified WordPress app to check you get the same visualizations.

PR submission checklist:


develric:
Hey @planarvoid 👋

> Do you think you can add a test for the PostDayViewsUseCase for the case when the dayViews list is empty?

Sure, makes sense! I added a test case as requested; since I'm not much familiar with the stats part of the app (as you are instead 😊), I'm not fully sure the Empty state is what we actually expect (that is if I created the correct preconditions for the test) so appreciate your usual valuable feedback! 😎

planarvoid:
hey @develric, thanks for the changes. If I understand it correctly, we're fixing here an edge case where the `fetch` actually returns some data (so the result is not empty) and the `get` does return an empty list. I have no idea how this happens but I think we can emulate it in the test. The way you wrote it it will never go to the `buildUiModel` method because `fetch` is empty and it will just emit `State.EMPTY` and that's it. I think to get the correct behaviour we need something like this. In this we actually get a `SUCCESS` result and an empty list. I don't think this is a state we want but it's what we were getting.
```
val forced = false
        whenever(emptyModel.dayViews).thenReturn(listOf())
        whenever(model.dayViews).thenReturn(listOf(Day("2019-10-10", 50)))
        whenever(store.getPostDetail(site, postId)).thenReturn(emptyModel)
        whenever(store.fetchPostDetail(site, postId, forced)).thenReturn(
                OnStatsFetched(
                        model
                )
        )

        val result = loadData(true, forced)

        Assertions.assertThat(result.state).isEqualTo(SUCCESS)
        Assertions.assertThat(result.data).isEmpty()
```

develric:
Thanks for the clarifying explanation @planarvoid 🙇‍♂️! I have reported it [here](f9433e9) with an extra comment to get the context. Let me know 😊.

planarvoid:
thanks for the change @develric ! The comment makes a lot of sense 👍

PR #10856, opened 2019-11-26
Create a Sequencing snackbar feature 

### LINKED ISSUE #10771: Failed upload Snackbar shown after successful upload

develric:
Fixes #10771 and supersedes #10820 

The code in this PR was mostly done in Pair Programming/Brainstorming with @malinajirka , so credits to him 🙇‍♂️. 

The basic idea is to introduce a snackbar sequencer to avoid that multiple snackbars, triggered nearly together, overlap not allowing to the user to check/understand the ongoing messages. We introduced a SnackbarSequencer class and a SnackbarSequencerInfo class. The client code can enqueue SnackbarSequencerInfo element into the SnackbarSequencer instead of directly use the WPSnackbar class. In this way the sequencer consumes the queue of SnackbarSequencerInfo visualuazing in sequence the snackbars.

To avoid memory leaks due to the fact that we keep into the sequencer queue  references to view/context and listeners that can eventually be no more active at the time we try to show the snackbar, we keep those references as Weak/Soft references.

For now we have used the sequencer only in upload posts related snackbars to limit the impact as much as possible to the #10771 issue. It will be eventually extended in a future time to all snackbars. NOTE: only snackbars that are not `Snackbar.LENGTH_INDEFINITE` are allowed, otherwise an `IllegalArgumentException` is triggered.

For the unit testing we introduced `kotlinx-coroutines-test` version 1.2.1 in the testImplementation tree. In this way we can use the `advanceTimeBy` function to test the coroutine/delay.

Pinging also @planarvoid to have a third couple of eyes on it.

## To test:
- Use the instructions in the #10771 (or use Charles Proxy and its Block List feature) to trigger an upload fail and check that now the multiple snackbars are shown in sequence and you can open the snackbar actions in between
- (this is tricky to execute and describe tbh) Use the AS to start debugging WP app and attach a memory profiler session:
    - place a breakpoint in `SnackbarSequencer.start` (in `val item = snackBarQueue.peek()`)
    - Open PostsList > Drafts and trash a post
    - the above breakpoint is hit
    - in the app exit from PostsListActivity using the back button and going to the main activity
    - in the profiler check that the PostsListActivity is still in memory
    - resume debugger and check in the logcat that you get the message `SnackbarSequencer > start context was null` or `SnackbarSequencer > start context was not alive` (this means the snackbar got the Activity was no more valid)
    - in the memory profiler force some cycles of garbage collection (using the bin icon); and check the PostsListActivity gets garbaged (can need 10/12 cycles or maybe more depending on memory pressure)
- Unit testing must pass

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/util/SnackbarItem.kt
+++ b/WordPress/src/main/java/org/wordpress/android/util/SnackbarItem.kt
@@ -27,8 +27,10 @@ fun getSnackbarDurationMs(snackbarItem: SnackbarItem): Long {
 class SnackbarItem(
     val info: Info,
     val action: Action? = null,
-    val callback: Callback? = null
+    dismissCallback: ((transientBottomBar: Snackbar?, event: Int) -> Unit)?
 ) {
+    val dismissCallback = SoftReference(dismissCallback)
+
     class Info(
         view: View,
         val textRes: UiString,
@@ -44,9 +46,10 @@ class SnackbarItem(
         val clickListener = SoftReference(clickListener)
     }
 
-    class Callback(
-        snackbarCallback: Snackbar.Callback
-    ) {
-        val snackbarCallback = SoftReference(snackbarCallback)
+    val snackbarCallback = object: Snackbar.Callback() {
+        override fun onDismissed(transientBottomBar: Snackbar?, event: Int) {
+            this@SnackbarItem.dismissCallback.get()?.invoke(transientBottomBar, event)
+            super.onDismissed(transientBottomBar, event)
+        }
     }
 }
--- a/WordPress/src/main/java/org/wordpress/android/util/SnackbarItem.kt
+++ b/WordPress/src/main/java/org/wordpress/android/util/SnackbarItem.kt
@@ -27,7 +27,7 @@ fun getSnackbarDurationMs(snackbarItem: SnackbarItem): Long {
 class SnackbarItem(
     val info: Info,
     val action: Action? = null,
-    dismissCallback: ((transientBottomBar: Snackbar?, event: Int) -> Unit)?
+    dismissCallback: ((transientBottomBar: Snackbar?, event: Int) -> Unit)? = null
 ) {
     val dismissCallback = SoftReference(dismissCallback)
 
@@ -46,7 +46,7 @@ class SnackbarItem(
         val clickListener = SoftReference(clickListener)
     }
 
-    val snackbarCallback = object: Snackbar.Callback() {
+    val snackbarCallback = object : Snackbar.Callback() {
         override fun onDismissed(transientBottomBar: Snackbar?, event: Int) {
             this@SnackbarItem.dismissCallback.get()?.invoke(transientBottomBar, event)
             super.onDismissed(transientBottomBar, event)
--- a/WordPress/src/main/java/org/wordpress/android/util/SnackbarItem.kt
+++ b/WordPress/src/main/java/org/wordpress/android/util/SnackbarItem.kt
@@ -14,22 +14,22 @@ private const val LONG_DURATION_MS = 2750L
 
 const val INDEFINITE_SNACKBAR_NOT_ALLOWED = "Snackbar.LENGTH_INDEFINITE not allowed in getSnackbarDurationMs."
 
-fun getSnackbarDurationMs(snackbarItem: SnackbarItem): Long {
-    return when (snackbarItem.info.duration) {
-        Snackbar.LENGTH_INDEFINITE ->
-            throw IllegalArgumentException(INDEFINITE_SNACKBAR_NOT_ALLOWED)
-        Snackbar.LENGTH_LONG -> LONG_DURATION_MS
-        Snackbar.LENGTH_SHORT -> SHORT_DURATION_MS
-        else -> snackbarItem.info.duration.toLong()
-    }
-}
-
 class SnackbarItem(
     val info: Info,
     val action: Action? = null,
     dismissCallback: ((transientBottomBar: Snackbar?, event: Int) -> Unit)? = null
 ) {
     val dismissCallback = SoftReference(dismissCallback)
+
+    fun getSnackbarDurationMs(): Long {
+        return when (info.duration) {
+            Snackbar.LENGTH_INDEFINITE ->
+                throw IllegalArgumentException(INDEFINITE_SNACKBAR_NOT_ALLOWED)
+            Snackbar.LENGTH_LONG -> LONG_DURATION_MS
+            Snackbar.LENGTH_SHORT -> SHORT_DURATION_MS
+            else -> info.duration.toLong()
+        }
+    }
 
     class Info(
         view: View,
--- a/WordPress/src/main/java/org/wordpress/android/util/SnackbarItem.kt
+++ b/WordPress/src/main/java/org/wordpress/android/util/SnackbarItem.kt
@@ -14,39 +14,39 @@ private const val LONG_DURATION_MS = 2750L
 
 const val INDEFINITE_SNACKBAR_NOT_ALLOWED = "Snackbar.LENGTH_INDEFINITE not allowed in getSnackbarDurationMs."
 
-fun getSnackbarDurationMs(snackbarSequencerInfo: SnackbarSequencerInfo): Long {
-    return when (snackbarSequencerInfo.snackbarInfo.duration) {
+fun getSnackbarDurationMs(snackbarItem: SnackbarItem): Long {
+    return when (snackbarItem.info.duration) {
         Snackbar.LENGTH_INDEFINITE ->
             throw IllegalArgumentException(INDEFINITE_SNACKBAR_NOT_ALLOWED)
         Snackbar.LENGTH_LONG -> LONG_DURATION_MS
         Snackbar.LENGTH_SHORT -> SHORT_DURATION_MS
-        else -> snackbarSequencerInfo.snackbarInfo.duration.toLong()
+        else -> snackbarItem.info.duration.toLong()
     }
 }
 
-class SnackbarActionInfo(
-    val textRes: UiString,
-    clickListener: OnClickListener
+class SnackbarItem(
+    val info: Info,
+    val action: Action? = null,
+    val callback: Callback? = null
 ) {
-    val clickListener = SoftReference(clickListener)
-}
+    class Info(
+        view: View,
+        val textRes: UiString,
+        val duration: Int
+    ) {
+        val view = WeakReference(view)
+    }
 
-class SnackbarCallbackInfo(
-    snackbarCallback: Snackbar.Callback
-) {
-    val snackbarCallback = SoftReference(snackbarCallback)
-}
+    class Action(
+        val textRes: UiString,
+        clickListener: OnClickListener
+    ) {
+        val clickListener = SoftReference(clickListener)
+    }
 
-class SnackbarInfo(
-    view: View,
-    val textRes: UiString,
-    val duration: Int
-) {
-    val view = WeakReference(view)
+    class Callback(
+        snackbarCallback: Snackbar.Callback
+    ) {
+        val snackbarCallback = SoftReference(snackbarCallback)
+    }
 }
-
-class SnackbarSequencerInfo(
-    val snackbarInfo: SnackbarInfo,
-    val snackbarActionInfo: SnackbarActionInfo? = null,
-    val snackbarCallbackInfo: SnackbarCallbackInfo? = null
-)
--- a/WordPress/src/main/java/org/wordpress/android/util/SnackbarSequencer.kt
+++ b/WordPress/src/main/java/org/wordpress/android/util/SnackbarSequencer.kt
@@ -7,7 +7,9 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
+import org.wordpress.android.ui.utils.UiHelpers
 import org.wordpress.android.util.AppLog.T
+import org.wordpress.android.viewmodel.ContextProvider
 import org.wordpress.android.widgets.WPSnackbar
 import javax.inject.Inject
 import javax.inject.Singleton
@@ -23,7 +25,10 @@ private const val SHORT_DURATION_MS = 1500
 private const val LONG_DURATION_MS = 2750
 
 @Singleton
-class SnackbarSequencer @Inject constructor() : Snackbar.Callback(), CoroutineScope {
+class SnackbarSequencer @Inject constructor(
+    private val uiHelper: UiHelpers,
+    private val contextProvider: ContextProvider
+) : Snackbar.Callback(), CoroutineScope {
     private var job: Job = Job()
 
     private var nextAvailableTimeSlot: Long = 0
@@ -32,11 +37,54 @@ class SnackbarSequencer @Inject constructor() : Snackbar.Callback(), CoroutineSc
     override val coroutineContext: CoroutineContext
         get() = Dispatchers.Default + job
 
-    private suspend fun schedule(snackbar: WPSnackbar, delay: Long) {
+    private suspend fun schedule(snackbarSequencerInfo: SnackbarSequencerInfo, delay: Long) {
         delay(delay)
+        System.gc()
         withContext(Dispatchers.Main) {
-            AppLog.d(T.UTILS, "SnackbarSequencer > show snackbar [${snackbar.message}]")
-            snackbar.show()
+            //AppLog.d(T.UTILS, "SnackbarSequencer > show snackbar [${snackbar.message}]")
+
+            snackbarSequencerInfo.context?.let {
+                if (it.get() == null) {
+                    AppLog.d(T.UTILS, "SnackbarSequencer > schedule got null context (garbaged?)")
+                } else {
+                    AppLog.d(T.UTILS, "SnackbarSequencer > schedule context [${it.get()}]")
+                }
+            }
+
+            snackbarSequencerInfo.snackbarInfo.view.get()?.let {view ->
+                snackbarSequencerInfo.snackbarInfo.let { snackbarInfo ->
+                    val message = uiHelper.getTextOfUiString(contextProvider.getContext(), snackbarInfo.textRes)
+
+                    val snackbar = WPSnackbar.make(
+                           view,
+                           message,
+                           snackbarInfo.duration
+                    )
+
+                    snackbarSequencerInfo.snackbarActionInfo?.let {
+                        snackbar.setAction(
+                                uiHelper.getTextOfUiString(contextProvider.getContext(), it.textRes),
+                                it.clickListener.get()
+                        )
+                    }
+
+                    snackbarSequencerInfo.snackbarCallbackInfo?.let {
+                        snackbar.addCallback(
+                                it.snackbarCallback.get()
+                        )
+                    }
+
+                    AppLog.d(T.UTILS, "SnackbarSequencer > schedule Showing snackbar [$message]")
+
+                    snackbar.show()
+                }
+            } ?: AppLog.d(
+                        T.UTILS,
+                    "SnackbarSequencer > schedule " +
+                            "skipping snackbar [${uiHelper.getTextOfUiString(
+                                    contextProvider.getContext(),
+                                    snackbarSequencerInfo.snackbarInfo.textRes)}]"
+            )
         }
     }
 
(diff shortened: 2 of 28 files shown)


WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
```diff
@@ -0,0 +1,119 @@
...
+
+        // When
+        for (item in items) {
+            sequencer.enqueue(item)
+        }
+
+        // Then
+        verify(wpSnackbar, times(1)).show()
+        advanceTimeBy(checkPoints[0])
+        verify(wpSnackbar, times(1)).show()
```
malinajirka: This test confused me for a bit -> if I understand it correctly, we advance time and we check that nothing else was shown even though there was an item in the queue. Is this intentional?

If it wasn't intentional we might want to change it to something like this:
```
  // Then
        verify(wpSnackbar, times(1)).show()
        advanceTimeBy(checkPoints[0])

        verify(wpSnackbar, times(2)).show()
        advanceTimeBy(checkPoints[1])

        verify(wpSnackbar, times(3)).show()
        advanceTimeBy(checkPoints[2])

        // Nothing else shown
        verify(wpSnackbar, times(3)).show()
        verifyNoMoreInteractions(wpSnackbar)
    }

    private fun getCheckPoints(items: List<SnackbarSequencerInfo>): List<Long> {
        return listOf(
                DURATION_EXTRA_MARGIN + getSnackbarDurationMs(items[0]),
            DURATION_EXTRA_MARGIN + getSnackbarDurationMs(items[1]),
            DURATION_EXTRA_MARGIN + getSnackbarDurationMs(items[2])
        )
    }
```
We check that an item is shown and we advance time by it's duration + duration_extra_margin -> we check that the next item is shown etc. 

If you wanted to test that the next item in the queue is not shown when you advance the time by less than duration+extra_martin, I'd consider something like this and I'd probably create a new test for this scenario.

```
 // Then
        verify(wpSnackbar, times(1)).show()
        advanceTimeBy(checkPoints[0] - 100)
      // verify the next item not shown yet
        verify(wpSnackbar, times(1)).show()
        advanceTimeBy(100)


        verify(wpSnackbar, times(2)).show()
        advanceTimeBy(checkPoints[1])

        verify(wpSnackbar, times(3)).show()
        advanceTimeBy(checkPoints[2])

        // Nothing else shown
        verify(wpSnackbar, times(3)).show()
        verifyNoMoreInteractions(wpSnackbar)
```
Wdyt?

WordPress/src/main/java/org/wordpress/android/util/SnackbarSequencer.kt
```diff
@@ -0,0 +1,114 @@
...
+import org.wordpress.android.widgets.WPSnackbar
+import org.wordpress.android.widgets.WPSnackbarWrapper
+import java.util.LinkedList
+import java.util.Queue
+import javax.inject.Inject
+import javax.inject.Named
+import javax.inject.Singleton
+import kotlin.coroutines.CoroutineContext
+
+const val DURATION_EXTRA_MARGIN = 500L
```
malinajirka: I think we should increase the MARGIN_DURATION, since it's currently causing this little animation twitch.
snackbar-short-margin
Wdyt?

malinajirka: @develric I just found out that the system handles the animation automatically when you set the MARGIN_DURATION to 0L. In that case I think we should completely remove it. Wdyt?

malinajirka:
I've tested the app and it works as expected except of the animation issue I mentioned above.

@planarvoid Since we wrote big parts of the code with Riccardo, I'll leave the review + approval + merge to you. Thanks! ;)

WordPress/src/main/java/org/wordpress/android/util/SnackbarSequencerInfo.kt
```diff
@@ -0,0 +1,52 @@
...
+
+class SnackbarActionInfo(
+    val textRes: UiString,
+    clickListener: OnClickListener
+) {
+    val clickListener = SoftReference(clickListener)
+}
+
+class SnackbarCallbackInfo(
+    snackbarCallback: Snackbar.Callback
```
planarvoid: I think I'd rather pass a lambda here instead of the `Snackbar.Callback`. It makes calling this function easier. Something like this:
```
class SnackbarCallbackInfo(
    callback: (transientBottomBar: Snackbar?, event: Int) -> Unit
) {
    val snackbarCallback: SoftReference<Snackbar.Callback> = SoftReference(object : Snackbar.Callback() {
        override fun onDismissed(transientBottomBar: Snackbar?, event: Int) {
            callback(transientBottomBar, event)
            super.onDismissed(transientBottomBar, event)
        }
    })
}
```

WordPress/src/main/java/org/wordpress/android/util/SnackbarSequencerInfo.kt
```diff
@@ -0,0 +1,52 @@
...
+}
+
+class SnackbarActionInfo(
+    val textRes: UiString,
+    clickListener: OnClickListener
+) {
+    val clickListener = SoftReference(clickListener)
+}
+
+class SnackbarCallbackInfo(
```
planarvoid: I think this object is not necessary. Inlining it makes everything more readable. What do you think about something like this: 
```
class SnackbarSequencerInfo(
    val snackbarInfo: SnackbarInfo,
    val snackbarActionInfo: SnackbarActionInfo? = null,
    val dismissCallback: ((transientBottomBar: Snackbar?, event: Int) -> Unit)? = null,
    val shownCallback: ((snackbar: Snackbar?) -> Unit)? = null
) {
    val snackbarCallback: SoftReference<Snackbar.Callback> = SoftReference(object : Snackbar.Callback() {
        override fun onDismissed(transientBottomBar: Snackbar?, event: Int) {
            dismissCallback?.invoke(transientBottomBar, event)
            super.onDismissed(transientBottomBar, event)
        }

        override fun onShown(sb: Snackbar?) {
            super.onShown(sb)
            shownCallback?.invoke(sb)
        }
    })
}
```

WordPress/src/main/java/org/wordpress/android/util/SnackbarSequencer.kt
```diff
@@ -0,0 +1,114 @@
...
+) : CoroutineScope {
+    private var job: Job = Job()
+
+    private val snackBarQueue: Queue<SnackbarSequencerInfo> = LinkedList()
+
+    override val coroutineContext: CoroutineContext
+        get() = bgDispatcher + job
+
+    fun enqueue(item: SnackbarSequencerInfo) {
+        synchronized(this@SnackbarSequencer) {
```
planarvoid: I'd consider using `launch(mainThread)` here instead since using the main thread would ensure synchronous access and I think it makes it easier to understand. There is also no heavy work necessary on the thread. It would mean replacing the next launch with `withContext(bgThread)`

WordPress/src/main/java/org/wordpress/android/util/SnackbarSequencerInfo.kt
```diff
@@ -0,0 +1,52 @@
...
+
+class SnackbarInfo(
+    view: View,
+    val textRes: UiString,
+    val duration: Int
+) {
+    val view = WeakReference(view)
+}
+
+class SnackbarSequencerInfo(
```
planarvoid: I'd simplify the names of these objects to something like this so we don't have to use `SnackbarInfo` in each name:
```
data class SnackbarItem(
   val info: SnackbarItem.Info,
   val actino: SnackbarItem.Action,
   val callback: SnackbarItem.Callback
) {
   class Info()
   class Action()
   class Callback()
}
```

WordPress/src/main/java/org/wordpress/android/util/SnackbarSequencer.kt
malinajirka: Ohh, great idea! I was curious so I tested it and it works as expected.

malinajirka: Here is a patch in case you want to use it @develric - https://cloudup.com/cdAXfGQft6N

WordPress/src/main/java/org/wordpress/android/util/SnackbarSequencerInfo.kt
develric: Much more readable, agree and changed 👍!

WordPress/src/main/java/org/wordpress/android/util/SnackbarSequencer.kt
develric: Makes sense, thanks for the patch 👍!

WordPress/src/main/java/org/wordpress/android/util/SnackbarSequencer.kt
develric: I'm not sure to fully understand the why of this behavior, but that's it 🤷‍♂️! Removed.

WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
develric: I was trying to condense both checks in the same test but makes sense to separate in a dedicated one. Changed 👍!

WordPress/src/main/java/org/wordpress/android/util/SnackbarSequencerInfo.kt
develric: I chatted a bit also with @malinajirka about this one, let me know if I targeted your intention 😊. I removed the `val` in `val dismissCallback` to try to avoid keeping strong reference (let me know if it makes sense). Changes here 3f60566.

develric:
Hi @malinajirka and @planarvoid , i should have followed up to all your comments. Waiting for your second pass review, thanks! 😊

WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
```diff
@@ -76,42 +81,91 @@ class SnackbarSequencerConcurrentTest {
...
+        verifyNoMoreInteractions(wpSnackbar)
     }
 
-    private fun getCheckPoints(items: List<SnackbarSequencerInfo>): List<Long> {
+    private fun getCheckPoints(items: List<SnackbarItem>): List<Long> {
         return listOf(
-            getSnackbarDurationMs(items[0]),
-            DURATION_EXTRA_MARGIN + getSnackbarDurationMs(items[1]),
-            DURATION_EXTRA_MARGIN + getSnackbarDurationMs(items[2])
+            SNACKBAR_DURATION_MARGIN + getSnackbarDurationMs(items[0]),
```
malinajirka: Could you please elaborate what is the purpose of `SNACKBAR_DURATION_MARGIN` here?

WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
```diff
@@ -76,42 +81,91 @@ class SnackbarSequencerConcurrentTest {
...
+        val checkPoints = getCheckPoints(items)
+
+        // When
+        for (item in items) {
+            sequencer.enqueue(item)
+        }
+
+        // Then
+        verify(wpSnackbar, times(1)).show()
+        advanceTimeBy(checkPoints[0] - 2 * SNACKBAR_DURATION_MARGIN)
```
malinajirka: We might want to consider removing the `2 *` here or setting `SNACKBAR_DURATION_MARGIN` to `200L`. Wdyt?

WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
```diff
@@ -76,42 +81,91 @@ class SnackbarSequencerConcurrentTest {
...
+        verify(wpSnackbar, times(1)).show()
+        advanceTimeBy(2 * SNACKBAR_DURATION_MARGIN)
         verify(wpSnackbar, times(2)).show()
```
malinajirka: We might want to consider removing all the lines below this line (exclusive). I think we are technically testing the same thing as "snackbars are shown in sequence with the correct duration".It would make the purpose of the test clearer imo. Wdyt?

WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
malinajirka: Btw I'm also not sure the name makes sense - it's basically just a random number right? We might want to consider hardcoding `100`. I think it's clear enough when used only within 3 lines of code. Wdyt?

WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
develric: The intent was to have the check points not fall exactly on the snachbar durations but just a bit later so to not making the checks exactly on the edge cases. Let me know if it makes any sense 😊.

WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
develric: This pairs a bit with the previous comment. Basically in respect to the check points I want to sample the state 1 SNACKBAR_DURATION_MARGIN before the snackbar duration end (so the 2* accounts for the initial offset in the checkpoints). I agree the naming is not great at all! I wait for your suggestion on keeping the extra margin in the checkpoints above (in that case maybe better to keep also the margin as parametric, otherwise we can get rid of it)

WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
develric: Well, I think you are right 😊! Let me know what you think of my answers to the previous points and I will change also this one as you suggested, thanks 🙇‍♂️!

WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
malinajirka: I see. Imo it makes the tests more complicated and they'll work even when you remove it. But I'm find with keeping it. Thanks for the explanation.

WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
malinajirka: So the unnecessary extra margins adds one more level of complexity. This make me even more unsure whether it's worth it. Wdyt @planarvoid ?

WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
malinajirka: After the discussion we had on slack. I think we might want to consider do the following
- remove SNACKBAR_DURATION_MARGIN
- rename "getCheckpoints" to "getDurations"
- replace advanceTimeBy(checkPoints[0]) with advanceTimeBy(durations[0] + 1)

WordPress/src/main/java/org/wordpress/android/util/SnackbarSequencerInfo.kt
planarvoid: looks great! thanks @develric (and good idea with removing of the `val`, I haven't thought of that!)

WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
planarvoid: I think this test is quite complicated and it doesn't have to be. For example why do we need different `TEST_DURATIONS` and not just have the same duration for all the items? And maybe we can just have 2 and not 3. These make it feel not like a unit test but more like something testing more than 1 item.

Saying that, I think we can merge this test and the previous one, have just one duration (500 ms), do something like:
- verify item 1 was emitted
- wait 250 ms
- verify nothing was emitted
- wait 250 ms
- verify item 2 was emitted

malinajirka: >I think this test is quite complicated and it doesn't have to be. For example why do we need different TEST_DURATIONS and not just have the same duration for all the items? And maybe we can just have 2 and not 3 items? These make it feel not like a unit test but more like something testing more than 1 item.

Good point! ;)

>Saying that, I think we can merge this test and the previous one, have just one duration (500 ms), do something like:

Actually I suggested splitting the test in two tests. I think it's nice for the test to have only a single reason to fail = test only one behavior. The fact the bodies of the two tests are almost identical isn't that important imo. I kind of like that the name of each of the tests explicitly says what it does. Wdyt?

WordPress/src/main/java/org/wordpress/android/util/SnackbarItem.kt
```diff
@@ -0,0 +1,55 @@
...
+import java.lang.ref.WeakReference
+
+// Taken from com.google.android.material.snackbar.SnackbarManager.java
+// Did not find a way to get them directly from the android framework for now
+private const val SHORT_DURATION_MS = 1500L
+private const val LONG_DURATION_MS = 2750L
+
+const val INDEFINITE_SNACKBAR_NOT_ALLOWED = "Snackbar.LENGTH_INDEFINITE not allowed in getSnackbarDurationMs."
+
+fun getSnackbarDurationMs(snackbarItem: SnackbarItem): Long {
```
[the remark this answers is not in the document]
develric: Good point, agree! Changed here f0dba9a

WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
develric: Good points from both of you here 👍! I applied a bit from both 😊.
Simplified the tests complexity reducing to 2 elements and using same duration from all snackbars. Also kept the 2 tests separated since I also feel this makes the test intent more clear. Let me know wdyt 😊. Changes here 353e66d85e25c4cb24cb050ae9228a85ce46655e.

WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
develric: see [here](https://github.com/wordpress-mobile/WordPress-Android/pull/10856#discussion_r351933046)

WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
develric: Maybe we can consider this addressed [here](https://github.com/wordpress-mobile/WordPress-Android/pull/10856#discussion_r351933046)?

develric:
Hey @malinajirka, @planarvoid ! Thanks both of you for the thorough review and comments! I hope I followed up to all of them accordingly. Let me know wdyt! 😊

WordPress/src/test/java/org/wordpress/android/util/SnackbarSequencerConcurrentTest.kt
planarvoid: Thanks for the changes, it looks much simpler and more readable. I think you're both right and it's correct to keep both tests 👍

mzorz:
While working on some other feature I need to user Snackbars for, found out about the Sequencer and so came here looking for info about why / how the Sequencer was needed, given the following:

> The basic idea is to introduce a snackbar sequencer to avoid that multiple snackbars, triggered nearly together, overlap not allowing to the user to check/understand the ongoing messages.

collided with my assumptions about how Snackbars work, as can be read in the [training documentation](https://developer.android.com/training/snackbar/showing):

> The system does not show multiple Snackbar objects at the same time, so if the view is currently displaying another Snackbar, the system queues your Snackbar and displays it after the current Snackbar expires or is dismissed.

I ran a quick test and called Snackbar several times only to confirm what you guys noted - any snackbar currently being shown on the screen will be closed immediately as a new call to `.show()` is issued 💥 

Been reading the code implementation of Snackbar and SnackbarManager in the support library. Looks like other people have discovered the [same issue ](https://stackoverflow.com/questions/33264764/when-is-snackbar-isqueued-used-how-to-queue-the-snackbar) as well, but their workarounds are much simpler given they still could count on `setCallback()` not being deprecated 😕.

I guess this is yet another inconsistency in the Android OS 🙃 - I wonder why they don't fix it 
[...] 
Went to Android's issue tracker to search for it and / or open a new issue and saw this one is closed as `won't fix` https://issuetracker.google.com/issues/37069975 😞 

Thanks for the work done here @develric @malinajirka @planarvoid , hope this can help others as well. 👏

PR #10703, opened 2019-10-25
Add Gallery Block

mkevins:
## Description

This PR is a first iteration for a semi-cross-platform Gallery block. This adds the gallery block, which should allow viewing, editing, and creating gallery blocks within the editor. In this 1st iteration, upload options are disabled, and new images can only be added via the WordPress Media Library.

### Related PRs
gutenberg-mobile: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1498
gutenberg: https://github.com/WordPress/gutenberg/pull/18265

PR submission checklist:

- [ ] I have considered adding unit tests where possible.

- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #10857, opened 2019-11-27
Use plural method for multiple media selection handling

mkevins:
This PR aims to consolidate various media selection handling by refactoring methods that utilize both singular and plural paradigms to only use plural methods where possible.

We should preserve the behavior that occurs when multiple images are selected via the media library intent (multiple selection is possible for the image block even though the `allowMultipleSelection` flag is false).

Related gutenberg-mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1621

**To test:**

---

**Image block:**

When multiple items are selected via the media library in the image block, the additional items are appended to the editor as new image blocks.

---

**Gallery block:**

Related PR: https://github.com/wordpress-mobile/WordPress-Android/pull/10703

Checkout related Gallery PR and build with source.

* In a post, add a gallery block
* Tap "ADD MEDIA"
* Tap "WordPress Media Library"
* Select multiple images

**Expected behavior:**

The selected images should be added to the gallery, and _not_ added as new image blocks.

---

Update release notes:


libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragmentAbstract.java
```diff
@@ -42,6 +42,8 @@
     public abstract void removeMedia(String mediaId);
     public abstract boolean showSavingProgressDialogIfNeeded();
     public abstract boolean hideSavingProgressDialog();
+    // Called from block editor when a media selection is cancelled
```
malinajirka: I think this comment might be a bit misleading -> it's actually called from the EditPostActivity when the activity wants to inform the block editor fragment that media selection was cancelled. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -2468,7 +2468,23 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+                case RequestCodes.STOCK_MEDIA_PICKER_SINGLE_SELECT:
+                case RequestCodes.MEDIA_LIBRARY:
+                case RequestCodes.PICTURE_LIBRARY:
+                case RequestCodes.TAKE_PHOTO:
+                case RequestCodes.VIDEO_LIBRARY:
+                case RequestCodes.TAKE_VIDEO:
+                case RequestCodes.STOCK_MEDIA_PICKER_MULTI_SELECT:
+                    mEditorFragment.mediaSelectionCancelled();
+                    return;
+                default:
```
malinajirka: NitPick: We might want to consider removing `default` and listing all the "cases" so when someone adds a new item into the ENUM, lint will inform them that they need to handle the new case in this switch. Wdyt?

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragmentAbstract.java
mkevins: You're right! I've updated the comment.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
mkevins: I tried this, but it seems the lint is not working in this way. Perhaps this is because the `RequestCode` is an enum-like-class with static int fields rather than an actual enum? I've left a comment above the switch, which will hopefully help.

malinajirka: My bad - I didn't realize `RequestCode` is technical a regular int. Sorry for the misleading suggestion. Thank you for the comment ❤️

PR #10860, opened 2019-11-27
Add non-null check for activity in Gutenberg fragment

### LINKED ISSUE #10829: NullPointerException: Attempt to invoke virtual method 'android.app.Application android.app.Activity.getApplication()' ...
Sentry Issue: [WORDPRESS-ANDROID-6QW](https://sentry.io/organizations/a8c/issues/1328088290/?referrer=github_integration)

```
NullPointerException: Attempt to invoke virtual method 'android.app.Application android.app.Activity.getApplication()' on a null object reference
    at org.wordpress.android.editor.GutenbergEditorFragment.initOtherMediaImageOptions(GutenbergEditorFragment.java:369)
    at org.wordpress.android.editor.GutenbergEditorFragment.access$400(GutenbergEditorFragment.java:60)
    at org.wordpress.android.editor.GutenbergEditorFragment$1.onGetOtherMediaImageOptions(GutenbergEditorFragment.java:289)
...
(10 additional frame(s) were not displayed)

Attempt to invoke virtual method 'android.app.Application android.app.Activity.getApplication()' on a null object reference
```

planarvoid:
Fixes #10829

This PR adds a non-null check for activity to the `onGetOtherMediaImageOptions` callback from Gutenberg. 

To test:
* I don't know how to reproduce this issue but the fix should be straightforward

PR submission checklist:

### DIFF
--- a/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/GutenbergEditorFragment.java
+++ b/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/GutenbergEditorFragment.java
@@ -26,6 +26,7 @@ import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.core.util.Consumer;
+import androidx.fragment.app.FragmentActivity;
 import androidx.fragment.app.FragmentManager;
 import androidx.fragment.app.FragmentTransaction;
 import androidx.lifecycle.LiveData;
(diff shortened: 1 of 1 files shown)


libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/GutenbergEditorFragment.java
```diff
@@ -378,10 +379,12 @@ public void run() {
 
     private ArrayList<MediaOption> initOtherMediaImageOptions() {
         ArrayList<MediaOption> otherMediaOptions = new ArrayList<>();
+        FragmentActivity activity = getActivity();
```
malinajirka: AFAIU the surrounded code, the editor will be opened, but the "stock photos" option will be missing. I think we might want to log this into a log (maybe even sentry). Wdyt?

planarvoid: I don't think it will be opened if the activity is null though but I can add tracking in case something else crashes after it passes this check 👍

malinajirka:
LGTM, thanks!

PR #10862, opened 2019-11-27
Issue/10859 bottom sheet not dismissed

### LINKED ISSUE #10859: Cannot dismiss reader subfilter bottom sheet after config change
This is a bug that came out while reviewing the #10792 PR (thanks @planarvoid for catching it 🙇‍♂️). This is actually valid also for the WPMainActivity create fab bottom sheet.

### Expected behavior
When having the bottom sheet opened in the reader and you rotate the device, you should get the bottom sheet still visualized. If now you select a bottom sheet item you should have the bottom sheet dismissed and the reader loading the relevant posts.

### Actual behavior
After the device rotation the bottom sheet is not dismissed when you tap on one of its items.

- Have a tag selected
- Click on the filter
- Rotate your device
- Try clicking on any site/tag and nothing happens

Pixel 3 emu, API 28.

develric:
Fixes #10859 

We fixed the issue both for the bottom sheet in reader and in main activity (create fab bottom sheet).

While debugging it I also actually noticed that in the onSubfilterClicked, the equality to compute the it.filtered was not super correct in cases when the filter is not the same object you have in the collection so added an equality function to the SubfilterListItem (since the PR is small I included it here, I know it's not super clean hope its fine enough 😅) 

## To test:
- follow steps in #10859 with the reader bottom sheet
- also check that the selected item in the bottom sheet is always the desired one
- repeat the #10859 steps for the create fab

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
@@ -19,6 +19,7 @@ import androidx.annotation.Nullable;
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.core.app.RemoteInput;
 import androidx.fragment.app.Fragment;
+import androidx.fragment.app.FragmentManager;
 import androidx.lifecycle.Observer;
 import androidx.lifecycle.ViewModelProvider;
 import androidx.lifecycle.ViewModelProviders;
@@ -169,7 +170,6 @@ public class WPMainActivity extends AppCompatActivity implements
     private SiteModel mSelectedSite;
     private WPMainActivityViewModel mViewModel;
     private FloatingActionButton mFloatingActionButton;
-    private MainBottomSheetFragment mMainBottomSheetFragment;
     private static final String MAIN_BOTTOM_SHEET_TAG = "MAIN_BOTTOM_SHEET_TAG";
 
     @Inject AccountStore mAccountStore;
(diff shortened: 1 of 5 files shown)


WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
```diff
@@ -404,16 +404,26 @@ private void initViewModel() {
...
+            });
+
+            mViewModel.isBottomSheetShowing().observe(this, isShowing -> {
+                if (!isShowing) {
+                    FragmentManager fm = getSupportFragmentManager();
+                    if (fm != null) {
+                        MainBottomSheetFragment bottomSheet =
+                                (MainBottomSheetFragment) fm.findFragmentByTag(MAIN_BOTTOM_SHEET_TAG);
+
+                        if (bottomSheet != null) {
```
planarvoid: What do you think about moving the construction of the fragment here? The logic should be something like:
```
if (isShowing && bottomSheet == null) {
          MainBottomSheetFragment bottomSheet = new MainBottomSheetFragment();
          bottomSheet.show(getSupportFragmentManager(), MAIN_BOTTOM_SHEET_TAG);
} else if (!isShowing && bottomSheet != null) {
          bottomSheet.dismiss();
}
```
this would make the `mFloatingActionButton.setOnClickListener` very simple (just `mViewModel.setIsBottomSheetShowing(true);`

develric: I was thinking about this initially since I also agree it cleans the click listener a lot, if I'm not wrong though in this way we need an instance variable (like mBottomSheet) to keep track of the null condition and I preferred to go the other way. Anyway I'm 50/50 between them so if you prefer the other way round or if I'm misunderstanding what you were suggesting let me know and I will just change it 👍.

develric:
Hi @planarvoid thanks for reviewing this 😊. I followed up to your comment, let me know wdyt and I will act accordingly 🙇‍♂️. About the refresh in orientation change it was not introduced with this PR so if you agree I would investigate it in a dedicated PR. Let me know 😊.

WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
develric: Don't know why but I totally misunderstood your suggestion 🤦‍♂️! Forget my previous comment! Changed here 80284ee

PR #10868, opened 2019-11-28
Gutenberg/activate preformat block

marecar3:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1264

GB mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1615

To test:

Follow test instructions on gb-mobile PR: wordpress-mobile/gutenberg-mobile#1615

PR submission checklist:


libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/AztecEditorFragment.java
```diff
@@ -2061,7 +2061,7 @@ private static String toHtml(SpannableStringBuilder builder,
         clearMetaSpans(builder);
         parser.syncVisualNewlinesOfBlockElements(builder);
         Format.postProcessSpannedText(builder, true);
-        return EndOfBufferMarkerAdder.removeEndOfTextMarker(parser.toHtml(builder, false));
+        return EndOfBufferMarkerAdder.removeEndOfTextMarker(parser.toHtml(builder, false, false));
```
hypest: I think we can use the `@JvmOverloads` annotation in the Aztec method in Kotlin to allow the Java side to call the method without the new extra parameter). Check out [this answer on SO](https://stackoverflow.com/a/34518024).

marecar3: Thanks for the hint. 
I have added annotation and now we don't need to pass extra param to the Kotlin method.

marecar3:
Hey @hypest, I think that I have addressed all issues so this one is ready for another iteration of review.

hypest:
I've LGTM'd the PR but, let's update the Aztec hash to a tag before we merge.

hypest:
It seems that the failing Firebase tests are not related to the work on this PR and happening on other commits merged to develop already so, we'll not fix them as part of this PR and merge it.

PR #10877, opened 2019-11-29
Fix crash in UploadStarter on Samsung devices with Android 5

### LINKED ISSUE #10827: ExceptionInInitializerError
Sentry Issue: [WORDPRESS-ANDROID-38R](https://sentry.io/organizations/a8c/issues/1115836993/?referrer=github_integration)

```
NoSuchFieldException: _state
    at java.lang.Class.getDeclaredField(Class.java:929)
    at java.util.concurrent.atomic.AtomicReferenceFieldUpdater$AtomicReferenceFieldUpdaterImpl.<init>(AtomicReferenceFieldUpdater.java:183)
    at java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater(AtomicReferenceFieldUpdater.java:67)
    at kotlinx.coroutines.JobSupport.<clinit>(JobSupport.kt)
    at kotlinx.coroutines.JobKt__JobKt.Job(Job.kt:361)
...
(21 additional frame(s) were not displayed)

RuntimeException: java.lang.NoSuchFieldException: _state
    at java.util.concurrent.atomic.AtomicReferenceFieldUpdater$AtomicReferenceFieldUpdaterImpl.<init>(AtomicReferenceFieldUpdater.java:202)
    at java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater(AtomicReferenceFieldUpdater.java:67)
    at kotlinx.coroutines.JobSupport.<clinit>(JobSupport.kt)
    at kotlinx.coroutines.JobKt__JobKt.Job(Job.kt:361)
    at kotlinx.coroutines.JobKt.Job
...
(20 additional frame(s) were not displayed)

ExceptionInInitializerError: None
    at org.wordpress.android.ui.uploads.UploadStarter.<init>(UploadStarter.kt:64)
    at org.wordpress.android.ui.uploads.UploadStarter_Factory.get(UploadStarter_Factory.java:81)
    at org.wordpress.android.ui.uploads.UploadStarter_Factory.get(UploadStarter_Factory.java:16)
    at org.wordpress.android.modules.DaggerAppComponent.injectWordPress(DaggerAppComponent.java:4849)
    at org.wordpress.android.modules.DaggerAppComponent.inject(DaggerAppComponent.java:4124)
...
(17 additional frame(s) were not displayed)
```

### LINKED ISSUE #10872: IndexOutOfBoundsException: 129, -122
Sentry Issue: [WORDPRESS-ANDROID-8W](https://sentry.io/organizations/a8c/issues/1021101226/?referrer=github_integration)

```
IndexOutOfBoundsException: 129, -122
    at android.text.PackedIntVector.deleteAt(PackedIntVector.java:218)
    at android.text.DynamicLayout.reflow(DynamicLayout.java:311)
    at android.text.DynamicLayout.-wrap0(DynamicLayout.java)
    at android.text.DynamicLayout$ChangeWatcher.reflow(DynamicLayout.java:662)
    at android.text.DynamicLayout$ChangeWatcher.onSpanChanged(DynamicLayout.java:691)
...
(15 additional frame(s) were not displayed)

129, -122
```

malinajirka:
Fixes #10827 

Fixes a crash on Samsung devices running Android 5. The app is crashing because it can't find `_state` variable in the runtime. However, we don't directly use this variable in our code. As it turned out the variable is used in `Mutex.kt`. This PR simply removes the usage of `Mutex` from our codebase and replaces it with `@Synchronized`.

More info about why the app is crashing can be find in @jd-alexander [comment](https://github.com/wordpress-mobile/WordPress-Android/issues/10827#issuecomment-557365282).

To test:
I don't have a Samsung phone running Android 5. But we should at least verify it still works as expected on other phones.
1. Turn off the internet connection
2. Create a post
3. Publish it
4. Turn on internet connection + pull to refresh
5. Make sure the post gets uploaded 

PR submission checklist:

Thank you @jd-alexander for the initial investigation which really helped me fix this issue!

@jkmassel We might want to consider merging this change as hotfix - wdyt? It crashed 3.6k times to 1.2k users in the last 90 days. I'm a bit worried that it might not work on Samsung devices running Android 5 at all - I haven't checked the stats though. I believe the change is very safe.


WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadStarter.kt
```diff
@@ -141,40 +133,41 @@ class UploadStarter @Inject constructor(
 
     /**
      * This is meant to be used by [checkConnectionAndUpload] only.
+     *
+     * The method needs to be synchronized from the following reasons. When the app comes to foreground both
```
shiki: > When the app comes to foreground both `queueUploadFromAllSites` and `queueUploadFromSite` are invoked

Not sure if this is true. The `queueUploadFromSite` is only called when the `PostListMainViewModel` is started or when swiping to refresh. 🤔 They can still run in parallel. But that's harder to explain. 😄

WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadStarter.kt
```diff
@@ -141,40 +133,41 @@ class UploadStarter @Inject constructor(
...
+     * `queueUploadFromAllSites` and `queueUploadFromSite` are invoked. The problem is that they can run in parallel
+     * and `uploadServiceFacade.isPostUploadingOrQueued(it)` might return out-of-date result and a same post is added
+     * twice.
      */
+    @Synchronized
```
shiki: Just noting an observation that I didn't notice before. 

The intent of `launch`-ing in here: 

https://github.com/wordpress-mobile/WordPress-Android/blob/aed0c79108ba08424af38b3eb28c2772a98bdc40/WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadStarter.kt#L135-L140

... was so we could run `SELECT` queries in parallel, and then queue all the posts in parallel. But because of the concurrency issues in `isPostUploadingOrQueued()`, we wrapped this in a Mutex which essentially negated the `launch(ioDispatcher)`. 🤷‍♂ 

I think I overengineered `UploadStarter` in the first place anyway so it doesn't bother me that much. 😄 We _could_ just simplify `UploadStarter` in the future or make `UploadService` able to handle concurrency. It's a shame that we can't be concurrent if phones have multiple processors now.

shiki (approved):
LGTM! 🎉 

- [x] Verified that all sites' posts are still uploaded on startup
- [x] Tested given testing steps

jd-alexander:
This can be closed right @shiki ?

shiki:
Yes! Thank you, @jd-alexander. 

Closing this in favor of #10878 which will be released with `release/13.7`.

WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadStarter.kt
malinajirka: Good catch:D! I agree we could use just one launch and not use a launch in a foreach loop. However, the UploadService uploads the items serially + the data bandwidth is limited, so I'm not sure it has any real impact on the UX.

WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadStarter.kt
malinajirka: Great eye ;)! I think what we originally wanted to write was "When the device connects to the internet both queueUploadFromAllSites and queueUploadFromSite are invoked` -> `queueUploadFromAllSites` is invoked from `connectionStatus` observer and `queueUploadFromSite` is invoked from the `UploadWorker`. I copy pasted the comment without thinking about it 😞 .

PR #10878, opened 2019-11-29
Fix crash in UploadStarter on Samsung devices with Android 5

shiki:
This has the same contents as #10877. The difference is this one is targeting `release/13.7`. 

## Testing 

Please do the same testing steps in #10877.

 
## Submitter Checklist

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -10,7 +10,6 @@
 * Block editor: Images clickable for fullscreen preview
 * Fixed time displayed on Post Conflict Detected and Unpublished Revision dialogs
 * Block editor: Fix issue when removing image/page break block crashes the app
-* Removed support for Giphy
 * Fixed a crash on Samsung devices running Android 5 (Lollipop)
  
 13.6
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -10,6 +10,8 @@
 * Block editor: Images clickable for fullscreen preview
 * Fixed time displayed on Post Conflict Detected and Unpublished Revision dialogs
 * Block editor: Fix issue when removing image/page break block crashes the app
+* Removed support for Giphy
+* Fixed a crash on Samsung devices running Android 5 (Lollipop)
  
 13.6
 -----
--- a/WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadStarter.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadStarter.kt
@@ -12,7 +12,6 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.sync.Mutex
 import org.wordpress.android.analytics.AnalyticsTracker.Stat
 import org.wordpress.android.fluxc.Dispatcher
 import org.wordpress.android.fluxc.generated.UploadActionBuilder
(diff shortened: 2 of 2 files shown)


RELEASE-NOTES.txt
```diff
@@ -10,6 +10,8 @@
 * Block editor: Images clickable for fullscreen preview
 * Fixed time displayed on Post Conflict Detected and Unpublished Revision dialogs
 * Block editor: Fix issue when removing image/page break block crashes the app
+* Removed support for Giphy
```
jd-alexander: Should this update for Giphy be included here? Let me know 🙏

shiki: Gah! That's from `develop`. Thanks for noticing that. 😞

shiki: Fixed in db5472a758dc32ad8af3546fa035fbb9cb4c31a1

jd-alexander:
LGTM 🚀

jd-alexander (approved):
LGTM 🚀

ParaskP7:
👋 @malinajirka !

It's been almost 3 years already, I know... 😅 But, let me ask you this coroutines related question below, just in case you still remember the answer. 🤞

As part of this [Compile Warnings PR](https://github.com/wordpress-mobile/WordPress-Android/pull/17210), that I am currently working on, I was trying to resolve the `@Synchronized annotation` related warning. To that end, I ended-up suppressing this warning instead, at least until I discuss it with you first (see [here](https://github.com/wordpress-mobile/WordPress-Android/commit/175f0e364cf4b923a64c7cf036ae75a244924c9e)). In this commit you will also notice that I added a couple of links as resources to help us figure out how to properly solve this warning:

- Kotlin Issue: [Forbid @synchronized annotation on suspend functions](https://youtrack.jetbrains.com/issue/KT-27333/Forbid-Synchronized-annotation-on-suspend-functions)
- Blog Post: [Coroutines and Java Synchronization Don't Mix](https://blog.danlew.net/2020/01/28/coroutines-and-java-synchronization-dont-mix/)

To that end, and as far as you aware and remember, do you think we could/should resolve this `@Synchronized annotation` related warning, maybe by using the something like this recommended `Mutex` solution, or, do you think that we should just ignore that altogether, suppressing it as I did and only touch that piece of logic again, when and if need be in the distant future. Wdyt? 🤔

malinajirka:
@ParaskP7 I do remember this code and convo relatively well considering how long it has been - time flies :). 
1. I'm wondering how come you pinged me even though I'm not mentioned on this PR at all :D? Just out of curiosity.
2. We were originally using the solution with a Mutex, but refactored it in this PR because it was crashing on Samsung devices with Android 5. Since we do not support Android 5 anymore, I believe it is safe to revert this change and start using mutex again.
3. I'd personally vote for fixing this issue, since if the code doesn't work with coroutines it might lead to race conditions which might result in double-uploads of posts and such - these kind of issues might be there without us knowing about them. However, that doesn't mean we shouldn't fix them. Moreover, I believe the fix is relatively simple and the code is tested for most parts.

Let me know if I can provide some further context.

ParaskP7:
@malinajirka time flies indeed! 🛸 

1. I just annotated this line on AS and it showed your name. Then, I copied the revision number (d5b994850d80de50edbcf3c8237aa9daf3ffedd0) which pointed to this PR. From the [3 commits](https://github.com/wordpress-mobile/WordPress-Android/pull/10878/commits) included in this PR, I am seeing you being involved in the main first one as well. 🧙 
2. Thank you so much for the extra clarification on that. Yes, I agree, we can revert this change and start using `Mutex` again, especially since we do not need to support Android 5 anymore (`API Level 21`). PS: Out current minimum SDK version is the `24` (Android 7).
3. I am voting on fixing this issue as well. It indeed seems a bit dangerous to have this kind of race conditions sitting there, waiting to bite us when we would least expect it, and even worse, as you said, having them there without us even knowing about them.

> Moreover, I believe the fix is relatively simple and the code is tested for most parts.

👍 🌟 🤞

> Let me know if I can provide some further context.

Not only you provided all the context I needed, you also did it at the speed of light, thank you for that, so much appreciated! 🙇 

FYI: I'll proceed with updating my PR and revert this change to `Mutex` again. 💯

PR #10892, opened 2019-12-03
Allow user to change display name and password at the same time during signup

### LINKED ISSUE #10835: Can't change display name during signup
On the signup screen after email confirmation there are 3 fields: Display name, Username and Password

It's impossible to update the display name and the password at the same time (I'm not sure if it's a local check or an API limitation).

Screenshot_1574411160

- Signup by email address in wpandroid
- Confirm email address, click in "Sign up to WordPress.com" in the received email.
- Update the display name, fill the password field and tap "Continue"
- Get the "The password setting cannot be changed at the same time as any other setting dialog. 💥

renanferrari:
Fixes #10835 

**The problem:** it's impossible for the user to change their display name and password at the same time during email signup. When trying to do so, the user gets a dialog stating that "The password setting cannot be changed at the same time as any other setting".

**The cause:** this is an API limitation (thank you @hypest for confirming this) that the current implementation doesn't take into consideration when it tries to update the display name and password [with a single network call](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/accounts/signup/SignupEpilogueFragment.java#L710-L712).

**The solution:** updating each account field (display name, username and password) separately – each with their own network call.

The main target of this PR was the `SignupEpilogueFragment` class, where I added some extra logic to manage the chain of updates. I suggest looking into the [`updateAccountOrContinue()`](https://github.com/renanferrari/WordPress-Android/blob/51485bbbc08b02d658bb0d5aeabb8823730937e4/WordPress/src/main/java/org/wordpress/android/ui/accounts/signup/SignupEpilogueFragment.java#L739-L760) method, which is where most of this logic happens. 

I also tried to keep in mind scenarios where it would be necessary to recover from failures that occurred in the middle of the chain. Even though I wasn't able to test this as much as I wanted, the order in which the updates occur (username -> display name -> password) were thought of with this in mind and I've also added some extra functionality that would allow for a more consistent behavior in those cases, which were the main reasons for the changes you can see [here](https://github.com/renanferrari/WordPress-Android/blob/51485bbbc08b02d658bb0d5aeabb8823730937e4/WordPress/src/main/java/org/wordpress/android/ui/accounts/signup/SignupEpilogueFragment.java#L515-L530) and [here](https://github.com/renanferrari/WordPress-Android/blob/51485bbbc08b02d658bb0d5aeabb8823730937e4/WordPress/src/main/java/org/wordpress/android/ui/accounts/signup/SignupEpilogueFragment.java#L728-L737). Let me know if I need to go into more detail about this.

To test
---

1. Make sure you have OAuth credentials with user creation capabilities.
2. If you have logged in to the app, sign out.
3. On the initial screen, tap **Sign up for WordPress.com**.
4. Tap **Sign up with email**.
5. Enter an email address that is not associated with a WordPress.com account.
6. On the next screen, tap **Open mail**.
7. On the signup email sent to the address used in step 5, tap **Sign up to WordPress.com**.
8. On the epilogue screen, fill in the _Display Name_ and _Password_ fields and tap **Continue**.
9. Confirm that **no error dialog appeared** and that the values were **correctly updated**.

Notes
---

1. For this PR, I did the minimum necessary to ensure the expected behavior and because of that, the proposed implementation here is extremely simple and tries to make good use of existing structures in the current code. However, this part of the signup flow seems to be a good candidate for further refactoring. As an example, I believe that the readability and testability of the `SignupEpilogueFragment` class could greatly benefit from the separation of responsibilities that a ViewModel could bring.

2. While working on this PR, I found another existing bug where the username, display name, and password fields don't seem to retain their values after configuration changes (screen rotation, for example). In an attempt to maintain the focus of this PR, I haven't looked into this problem any further. Let me know if I should open a specific issue for it.

---

**PR submission checklist:**

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/accounts/signup/SignupEpilogueFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/accounts/signup/SignupEpilogueFragment.java
@@ -521,9 +521,7 @@ public class SignupEpilogueFragment extends LoginBaseFormFragment<SignupEpilogue
     }
 
     protected boolean changedUsername() {
-        final String savedUsername = mAccount.getAccount().getUserName();
-        final String initialUsername = getArguments().getString(ARG_USERNAME);
-        return !TextUtils.equals(!TextUtils.isEmpty(savedUsername) ? savedUsername : initialUsername, mUsername);
+        return !TextUtils.equals(mAccount.getAccount().getUserName(), mUsername);
     }
 
     /**
--- a/WordPress/src/main/java/org/wordpress/android/ui/accounts/signup/SignupEpilogueFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/accounts/signup/SignupEpilogueFragment.java
@@ -475,7 +475,7 @@ public class SignupEpilogueFragment extends LoginBaseFormFragment<SignupEpilogue
                    && !TextUtils.isEmpty(mAccountStore.getAccount().getEmail())) {
             endProgress();
             populateViews();
-        } else if (event.causeOfChange == AccountAction.PUSH_SETTINGS && mSignupEpilogueListener != null) {
+        } else if (event.causeOfChange == AccountAction.PUSH_SETTINGS) {
             mHasMadeUpdates = true;
 
             if (mIsUpdatingDisplayName) {
@@ -503,7 +503,7 @@ public class SignupEpilogueFragment extends LoginBaseFormFragment<SignupEpilogue
                             + event.error.type + " - " + event.error.message);
             endProgress();
             showErrorDialog(getString(R.string.signup_epilogue_error_generic));
-        } else if (mSignupEpilogueListener != null) {
+        } else {
             mHasMadeUpdates = true;
             AnalyticsTracker.track(mIsEmailSignup
                     ? AnalyticsTracker.Stat.SIGNUP_EMAIL_EPILOGUE_UPDATE_USERNAME_SUCCEEDED
--- a/WordPress/src/main/java/org/wordpress/android/ui/accounts/signup/SignupEpilogueFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/accounts/signup/SignupEpilogueFragment.java
@@ -103,6 +103,12 @@ public class SignupEpilogueFragment extends LoginBaseFormFragment<SignupEpilogue
     protected boolean mIsAvatarAdded;
     protected boolean mIsEmailSignup;
 
+    private boolean mIsUpdatingDisplayName = false;
+    private boolean mIsUpdatingPassword = false;
+    private boolean mHasUpdatedDisplayName = false;
+    private boolean mHasUpdatedUsername = false;
+    private boolean mHasUpdatedPassword = false;
+
     private static final String ARG_DISPLAY_NAME = "ARG_DISPLAY_NAME";
     private static final String ARG_EMAIL_ADDRESS = "ARG_EMAIL_ADDRESS";
     private static final String ARG_IS_EMAIL_SIGNUP = "ARG_IS_EMAIL_SIGNUP";
@@ -275,7 +281,9 @@ public class SignupEpilogueFragment extends LoginBaseFormFragment<SignupEpilogue
                 // Start progress and wait for account to be fetched before populating views when
                 // email does not exist in account store.
                 if (TextUtils.isEmpty(mAccountStore.getAccount().getEmail())) {
-                    startProgress(false);
+                    if (!isInProgress()) {
+                        startProgress(false);
+                    }
                 } else {
                     // Skip progress and populate views when email does exist in account store.
                     populateViews();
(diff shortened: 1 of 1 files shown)


WordPress/src/main/java/org/wordpress/android/ui/accounts/signup/SignupEpilogueFragment.java
```diff
@@ -476,23 +504,29 @@ public void onUsernameChanged(OnUsernameChanged event) {
...
+            updateAccountOrContinue();
         }
     }
 
     protected boolean changedDisplayName() {
-        return !TextUtils.equals(getArguments().getString(ARG_DISPLAY_NAME), mDisplayName);
+        final String savedDisplayName = mAccount.getAccount().getDisplayName();
+        final String initialDisplayName = getArguments().getString(ARG_DISPLAY_NAME);
+        return !TextUtils.equals(!TextUtils.isEmpty(savedDisplayName)
+                ? savedDisplayName : initialDisplayName, mDisplayName);
```
maxme: I'm not sure to understand the logic here (and in the `changedUsername` function):
- You check that the stored value in the AccountModel is empty.
- If it's empty you use the initial string generated when the fragment starts and compare with the textfield.

This function will return false if the stored value is empty and the initial generated text is "username" and the display name textfield hasn't been changed. So If I'm not mistaken, in this case, the display name won't be changed and will stay empty. 

Note: I haven't checked how the display name in AccountModel can be empty.

renanferrari: > I'm not sure to understand the logic here (and in the `changedUsername` function):
> 
> * You check that the stored value in the AccountModel is empty.
> * If it's empty you use the initial string generated when the fragment starts and compare with the textfield.

That's correct. Since separate updates are now being made, the most assertive way I found to verify that the username/display name actually changed was to check the last stored value rather than relying on the values inside the arguments bundle.

Initially, this check only existed in `changedUsername()` and the reason I assumed it could be empty is because there was a similar check inside [`populateViews()`](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/accounts/signup/SignupEpilogueFragment.java#L593-L594). The reason why I later replicated this check to `changedDisplayName()` (even though I wasn't sure it could ever be empty) has to do with the case you just mentioned:

> This function will return false if the stored value is empty and the initial generated text is "username" and the display name textfield hasn't been changed. So If I'm not mistaken, in this case, the display name won't be changed and will stay empty.

First of all, thank you for pointing this out. Yes, I think you're correct. If the stored display name is empty and the user decides to keep the generated one, the generated one won't be saved and the stored one will remain empty.

After investigating this a bit further, I now believe this behavior is rooted in a bug already present in the current implementation, where the generated display name won't be saved if the user doesn't change any value in the epilogue screen. You can verify this by doing the following:
1. Follow the steps 1 to 7 described in the "To test" section of this PR.
2. On the epilogue screen, don't change any of the values but take note of the generated display name.
3. Tap **Continue**.
4. On the next screen, go to the _Me_ tab.
5. Verify that the display name is not the same as the one generated in the epilogue screen (in my case it was the same as the username).

_Note: I realize it may be a little tricky to reproduce this, as I think it may depend on the email you're using to sign up, but I was able to consistently do so, even with the latest version of the app in production (13.7)._

In the proposed implementation, I believe this will only happen in the case you mentioned, where the stored display name is empty, which I'm also not sure if it'll ever be.

I actually stumbled upon this problem while testing some scenarios before submitting this PR but (mistakenly) assumed it was a problem with my own implementation. It only started working as expected after I replicated the check from `changedUsername()` to `changedDisplayName()`, which finally explains why it is there as well.

If we can confirm that the stored display name will never be empty, I think it would be possible to simplify the logic of `changedDisplayName()` so that it simply compares the value of the text field with the stored one. What do you think? Maybe @theck13 can help clarify this as well?

maxme: Thanks for the investigation!

> If we can confirm that the stored display name will never be empty, I think it would be possible to simplify the logic of changedDisplayName() so that it simply compares the value of the text field with the stored one. What do you think? 

Sounds good to me, I haven't checked yet how the AccountModel is initialized, but I guess it's done during user creation POST call (after the email confirmation).

theck13: I think the `mAccount.getAccount().getDisplayName()` is always not empty after the account is created before showing the signup epilogue screen.  The `createDisplayNameFromEmail()` method exists to create a slightly more user-friendly display name based on the email address rather than the default display name that matches the username.  In that case, we should _always_ save the display name.  We _could_ optimize it a little and save the display name only if it doesn't match the username, but that might be confusing to future maintainers.

WordPress/src/main/java/org/wordpress/android/ui/accounts/signup/SignupEpilogueFragment.java
```diff
@@ -451,16 +475,20 @@ public void onAccountChanged(OnAccountChanged event) {
                    && !TextUtils.isEmpty(mAccountStore.getAccount().getEmail())) {
             endProgress();
             populateViews();
-        } else if (changedUsername()) {
-            startProgress(false);
-            PushUsernamePayload payload = new PushUsernamePayload(
-                    mUsername, AccountUsernameActionType.KEEP_OLD_SITE_AND_ADDRESS);
-            mDispatcher.dispatch(AccountActionBuilder.newPushUsernameAction(payload));
         } else if (event.causeOfChange == AccountAction.PUSH_SETTINGS && mSignupEpilogueListener != null) {
```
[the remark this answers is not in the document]
renanferrari: Good catch! Just removed them 👍

WordPress/src/main/java/org/wordpress/android/ui/accounts/signup/SignupEpilogueFragment.java
renanferrari: Thanks for the input, @theck13!

> I think the `mAccount.getAccount().getDisplayName()` is always not empty after the account is created before showing the signup epilogue screen.

Great, that's my understanding too. I'll proceed with the approach I mentioned earlier then.

renanferrari: For future reference, I have also applied the same simplification from `changedDisplayName()` to `changedUsername()` as it didn't seem to ever be empty at that point as well. 

A possible future improvement – beyond the scope of this PR – would be to investigate whether it would be possible to remove similar checks being made inside `populateViews()` and `undoChanges()`.

PR #10898, opened 2019-12-04
Issue/10634 material pages list

### LINKED ISSUE #10634: Implement Material theme on Pages List screen
We should switch widgets used at Pages LIst screen to their material counterparts, remove color references where possible, and switch to using material `textAppearance` attributes.

khaykov:
Fixes #10634 

This PR moves Pages screen to Material Theme and adds Dark Theme support. Changes are cosmetic.

At the time, not all screens have Dark Theme support, so the layout will look broken at most of the places.

comparison 

To test:
- Navigate to Pages screen and make sure it looks as expected in both light and dark themes. The theme can be changed from App Settings.

This PR requires one designer and one developer to review it.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/pages/PageItemViewHolder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/pages/PageItemViewHolder.kt
@@ -11,7 +11,6 @@ import android.widget.PopupMenu
 import android.widget.RadioButton
 import android.widget.TextView
 import androidx.annotation.LayoutRes
-import androidx.core.widget.CompoundButtonCompat
 import androidx.recyclerview.widget.RecyclerView
 import org.wordpress.android.R
 import org.wordpress.android.ui.ActionableEmptyView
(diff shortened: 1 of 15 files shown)


mattmiklic:
This is looking good to me. 👍

khaykov:
Updated the style of the primary button (used in empty view) after discussion with @mattmiklic .

Screenshot_1575928168

WordPress/src/main/java/org/wordpress/android/ui/pages/PageItemViewHolder.kt
```diff
@@ -163,10 +162,6 @@ sealed class PageItemViewHolder(internal val parent: ViewGroup, @LayoutRes layou
                 radioButton.setOnClickListener {
                     onParentSelected(pageItem)
                 }
-
-                @Suppress("DEPRECATION")
-                CompoundButtonCompat.setButtonTintList(radioButton,
```
planarvoid: when we remove this, the radio button has the light blue color, right? At least that's what see in the app now. Is that what we want? 😄 
Screenshot 2019-12-11 at 14 44 00

khaykov: Yes, there is no reason for this radio buttons to be special ones :)

khaykov:
Thank you for taking a look, @planarvoid ! Yes, removal of a grey background is part of the plan :)

khaykov:
@planarvoid Since you only commented on visual changes (and all of them are intentional) I'm going to merge this PR since couple of other ones depend on it :)

Again, thanks for taking the time to check it!

PR #10899, opened 2019-12-04
Update post object from a background thread using coroutines

planarvoid:
There is a deadlock happening in the `EditPostActivity` which is causes by the post settings fragment calling `updateInTransaction` at the same time as the `EditPostActivity`. The main reason is that the aztec fragment method `getContent` jumps on the main thread and blocks it. There is a race condition, however, it happens quite often. I've added a new method called `updateAndForget` which calls the `updateInTransaction` asynchronously with a callback. This ensures that we call the `updateInTransaction` always from a background thread and should be called whenever we don't need the result of the `updateInTransaction` method (which is the case of the post being updated from post settings). 

The `updateAndForget` method doesn't block the current thread.

To test:
* Go to media
* Add a photo with +
* Choose "Take a photo"
* Wait until the photo gets uploaded
* Click on the success snackbar button to create a new post
* The post gets created with the image and the app doesn't freeze

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
@@ -810,6 +810,7 @@ public class EditPostActivity extends AppCompatActivity implements
         if (mReactNativeRequestHandler != null) {
             mReactNativeRequestHandler.destroy();
         }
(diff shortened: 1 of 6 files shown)


malinajirka:
Closing in favor of https://github.com/wordpress-mobile/WordPress-Android/pull/10900

PR #10915, opened 2019-12-06
Issue/10631 material post list

### LINKED ISSUE #10631: Implement Material theme on Post List screen
We should switch widgets used at the Post List screen to their material counterparts, remove color references where possible, and switch to using material `textAppearance` attributes.

The gradient element at the tabs will require special treatment for elevation in dark mode.

A potential issue with separators (trailing invisible items in the list).

khaykov:
Fixes #10631 

This PR moves Posts list screen to Material Theme and adds Dark Theme support. There are some code changes for cosmetic reasons.

At the time, not all screens have Dark Theme support, so the layout will look broken at most of the places.

comparison_posts

To test:
- Navigate to Posts screen and make sure it looks as expected in both light and dark themes. The theme can be changed from App Settings.
 

PR submission checklist:


mattmiklic:
This is looking good. 👍 For the loading state, it looks like we're using a gradient of gray and white. Could we maybe change this to gray/semi-transparent gray instead, so the same skeleton could work in both light and dark modes?

khaykov:
As discussed with @mattmiklic, I updated the skeleton color.

shiki:
Taking a look. 👀

WordPress/src/main/java/org/wordpress/android/ui/posts/PostsListActivity.kt
```diff
@@ -157,7 +165,27 @@ class PostsListActivity : AppCompatActivity(),
...
         authorSelection = findViewById(R.id.post_list_author_selection)
+
+        val elevationOverlayProvider = ElevationOverlayProvider(this)
+        val appbarElevation = resources.getDimension(R.dimen.appbar_elevation)
+        val appBarColor = elevationOverlayProvider.compositeOverlayIfNeeded(
+                this.getColorFromAttribute(R.attr.wpColorAppBar),
+                appbarElevation
+        )
+
+        appbar.setBackgroundColor(appBarColor)
```
shiki: I'm not sure I understand what this does. It looks like even without this line, the app bar still looks the same. 🤔 

Without | With this line
--------|-------
without background color |       with background color set

Am I looking at the wrong thing? 😅

WordPress/src/main/java/org/wordpress/android/ui/posts/adapters/AuthorSelectionAdapter.kt
```diff
@@ -122,11 +128,33 @@ class AuthorSelectionAdapter(context: Context) : BaseAdapter() {
...
+        ) {
             super.bind(state, imageManager, uiHelpers)
             val context = itemView.context
+
+            // Because it's a custom popup list we need to manage colors of list items manually
+            val elevationOverlayProvider = ElevationOverlayProvider(context)
+            val appbarElevation = context.resources.getDimension(R.dimen.appbar_elevation)
+            val elevatedSurfaceColor = elevationOverlayProvider.compositeOverlayWithThemeSurfaceColorIfNeeded(
+                    appbarElevation
+            )
```
shiki: This is quite a long dance. It was a bit difficult for me to understand at first. But that's just me coming from no experience with Dark Mode. 😅 

It feels like overkill to do this now in this PR, but do you have plans on making utility methods for this in the future? Or all usage of `ElevationOverlayProvider` is probably going to look different?

WordPress/src/main/java/org/wordpress/android/ui/posts/PostsListActivity.kt
khaykov: Yeah, you are right! It was supposed to fix a shadow issue in light mode, but I guess some subsequent changes in layout fixed it instead :)

WordPress/src/main/java/org/wordpress/android/ui/posts/adapters/AuthorSelectionAdapter.kt
khaykov: I was thinking about using an extension for this, but I got to it a bit late -  at the moment I have a bunch of outstanding PRs with similar functionality, so I'll wait till the number reduces a bit so I can make a better decision on implementation and avoid conflicts.

shiki: Yup! That sounds like the best approach to me. 🙂

khaykov:
@shiki Again, thank you so much for taking a look at this PR :) All the visual changes you mentioned are intentional. We took this opportunity to clean up and revamp the app style a bit. Sorry, I should have mentioned it in the PR description 🙇 

I also just realized that Search hint text color in Toolbar is black in light mode, and fixed it :)

shiki:
> Sorry, I should have mentioned it in the PR description

No worries. 🙂 

LGTM!

PR #10916, opened 2019-12-06
Accessibility : Aztec Media Options TalkBack Improvements

### LINKED ISSUE #10894: Accessibility : Aztec Media Options TalkBack Improvements
For this audit the governing principles are as follows ([inspired by a previous audit](https://github.com/wordpress-mobile/WordPress-Android/issues/7288))

- All the clickable elements should be marked as so.
- All nonactionable and decorative elements should be non-focusable.
- Relevant content should be grouped together.
- All important elements should be described through content description.

The following is a list of opportunities to improve the accessibility of the Editor's Media Options.

### TalkBack Issues

* Media Options Dialog
	* When you press the switch that opens the media options the main area that get's focused says "Out of Pager" I guess it's to signify that there's a view that's overlaying the main pager so the user knows to hover to discover the new current view. Possibly renaming the pager i.e. changing what it's announced as might make it clearer as to what the user is interacting with eg. renaming the pager's announcement to "Editor"

* Gallery
	* Disable "Person reading device with notifications" announcement on the permissions view within the media options dialog.
	
* Free Photo Picker
	* "List" is read by the announcer.
	* Descriptions provided for images here are really good. Good to replicate for other pickers. 

* WordPress Media Picker
	* Search - Double check to see what actually powers the media search since the filenames aren't being found during the process when using TalkBack.

jd-alexander:
Fixes #10894

To test:

For all below scenarios follow the steps of : 

1. Creating a new post with Aztec as the selected editor. 
2. Expand the media toolbar. 

#### Gallery

1. Select an image in the grid. Instead of hearing "Unlabeled" button, you will hear "Image Thumbnail".
2. After selecting or unselecting an image you will hear the selected state being announced. 

#### Free Photo Picker

1. Once you focus on the list you will hear "Photo Library, List" so there's more context as to what the list contains. 
2. The selected state of images is also being announced for this picker. 

#### WordPress Media Picker

1. Item's selected status is being announced. 
2. The Confirm action now has a content description that states that the action actually confirms the selected images so that will assist in making it less vague. 
3. When the "Done" `ActionMode` is triggered it now announces that the selected images are cleared. I attempted to add a content description to this button but finding an actual reference to it has been difficult. 

	Attempts made were as follows : 

	1. When the menu is inflated I tried to use several versions of the `action_mode_close_button` identifier to fetch the view. When `findViewById` was run on the menu and the activity the view wasn't found even though that's the id. 
	
	2. Utilized the layout inspector to verify the view id but it's still a bit tricky figuring out where the best place to grab it actually is. So I settled on this approach for the time being to not commit any more time to this issue. 

PR submission checklist:


jd-alexander:
Hi @khaykov Thanks a lot for the review. I really appreciate it 🙇 

> "Close" button that appears after you select an image from WP medial library or initial photo picker is announced as "Done" which could be misleading. Maybe it should say something like "Cancel" ?

Thanks for pointing this out. This is something I have been investigation to do but I haven't been able to find a way to apply a title to the action since it's not defined in the xml menu resource. The same comments in the WordPress Media Picker section of the PR description applies here as well. Let me know if you have any thoughts on this. 

> While Image Selected action is announced in the initial image picker, it's not announced in the WP media library picker (the only announcement is when all selected images are unselected).

Thanks. Will add the behavior to this picker as well. 

>  Also, selected status is not announced when you hover over the thumbnail, so it's impossible to tell if the image is selected or not until you double-tap on it. 

A good observation! 
 
> Also, when an image is selected, the border around it announced with the number in blue circle - it's a bit confusing, as there is no context to that announcement.

👍 

> "Confirm selected images" button is announced in the WP media library, but not in the initial image selector (whatever it is called).

Indeed. I will standardize the behavior for all pickers.

jd-alexander:
Hi @khaykov I implemented all the changes you requested. All behavior is now standardized across all pickers. 

* Done button is now "Cancel" 6bb18da and Confirm Button just says "Confirm" due to `setContentDescription` only being available on API 26 and up 225a111. This seems more consistent with other actions, where just one word is being utilized. 
* All media pickers now support announcement of the image's selected status when focused. 85b64d1 71c6f2b
* Image Thumbnail layouts have been made more accessible, so views that aren't important for accessible are no longer being announced. ea18b82 e6712e8

jd-alexander:
> Thank you so much for the changes, and figuring how to deal with Action Mode - always tricky :)
Indeed 😄 

> I have only one small comment - when you select a selected image, it's selected status is announced as Image Selected, so it sounds like Image Thumbnail Image Selected, Button, Row 1...etc. I think just saying "Selected" might be good enough, and maybe adding some commas to separate it from the other words (I believe commas make the announcer pause for a bit). What do you think?

Thanks! Great observation. This was resolved in 0e21306


> Also, I noticed Gutenberg submodule commit in this PR, and was wondering is it supposed to be here :)
Yes, I was having some Gutenberg issues and that went in. It's removed now!

PR #10917, opened 2019-12-06
Accessibility : TalkBack EmptyView announcement fix

### LINKED ISSUE #10907: [TalkBack] Empty screens are spoken with “Tap to save posts...“
## Expected behavior

Only the text should be spoken.

## Actual behavior

The extra unrelated text is spoken.

1. Navigate to Reader → Followed Sites
2. Tap on Search
3. Enter a very long text that should have no results
4. Turn on TackBack
5. Hover on the empty list

Tested on Pixel 3, Android 9.0, WPAndroid `develop` ()

---

FYI @jd-alexander 🙂

jd-alexander:
Fixes issue when the content description of the TextView remains set so the actual TextView text isn't being read.

Fixes #10907

What caused this issue was that when announcing the empty view the default behavior is to utilize the text of the `TextViews` but there's is a case where an icon is being injected in the string that's being set to a `TextView`.
https://github.com/wordpress-mobile/WordPress-Android/blob/caeaf58e0137b56b7c27fbc5913107dfcbbf422c/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java#L1517
 so TalkBack is reading the symbols that represent the icon in the string so the content description was leveraged to set a text that TalkBack truly understands. But this wasn't being cleared so it was being used as the default content description being supplied for announcement for every empty state. The solution is a bit hacky but it's the easiest way to address the issue since adding icons to the Text of the `TextView` made it inaccessible in the first place.

To test:

1. Navigate to Reader → Followed Sites
2. Tap on Search
3. Enter a very long text that should have no results
4. Turn on TackBack
5. Hover on the empty list

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/ActionableEmptyView.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/ActionableEmptyView.kt
@@ -137,6 +137,13 @@ class ActionableEmptyView : LinearLayout {
         layout.layoutParams = params
     }
 
+    /**
+     * Announces the empty view when the empty state occur. Due to the formatting of subtitle text in certain
+     * circumstances TalkBack isn't able to properly make it's announcement so in cases like these the content
+     * description is dynamically added before doing so. Since the appropriate measure is to rely on the TextView's
+     * text as the content for announcement the content description of the subtitle is cleared so that the subtitle
+     * text can be default.
+     */
     fun announceEmptyStateForAccessibility() {
         val subTitle = if (!TextUtils.isEmpty(subtitle.contentDescription)) {
             subtitle.contentDescription
@@ -145,5 +152,7 @@ class ActionableEmptyView : LinearLayout {
         }
 
         announceForAccessibility("${title.text}.$subTitle")
+
+        subtitle.contentDescription = null
     }
 }
--- a/WordPress/src/main/java/org/wordpress/android/ui/ActionableEmptyView.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/ActionableEmptyView.kt
@@ -138,11 +138,9 @@ class ActionableEmptyView : LinearLayout {
     }
 
     /**
-     * Announces the empty view when the empty state occur. Due to the formatting of subtitle text in certain
+     * Announces the empty view when the empty state occurs. Due to the formatting of subtitle text in certain
      * circumstances TalkBack isn't able to properly make it's announcement so in cases like these the content
-     * description is dynamically added before doing so. Since the appropriate measure is to rely on the TextView's
-     * text as the content for announcement the content description of the subtitle is cleared so that the subtitle
-     * text can be default.
+     * description is dynamically added before doing so.
      */
     fun announceEmptyStateForAccessibility() {
         val subTitle = if (!TextUtils.isEmpty(subtitle.contentDescription)) {
--- a/WordPress/src/main/java/org/wordpress/android/ui/ActionableEmptyView.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/ActionableEmptyView.kt
@@ -152,7 +152,9 @@ class ActionableEmptyView : LinearLayout {
         }
 
         announceForAccessibility("${title.text}.$subTitle")
+    }
 
+    fun resetSubTitleContentDescription() {
         subtitle.contentDescription = null
     }
 }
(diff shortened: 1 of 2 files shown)


WordPress/src/main/java/org/wordpress/android/ui/ActionableEmptyView.kt
```diff
@@ -145,5 +152,7 @@ class ActionableEmptyView : LinearLayout {
         }
 
         announceForAccessibility("${title.text}.$subTitle")
+
+        subtitle.contentDescription = null
```
shiki: This feels like an odd place to have a `contentDescription` being set to `null`. I wouldn't expect a method like `announceEmptyStateForAccessibility` to do that. 

I investigated a bit and it looks like the cause of the issue is `ReaderPostListFragment` is being reused whenever we switched from _Saved Posts_ to _Search_ and to other filters. This means that the `mActionableEmptyView` is behaving as it should but it's **state** is just not being reset properly. There's a `contentDescription` not being reset somewhere. 

The fix should most probably be in how `ReaderPostListFragment` uses `ActionableEmptyView`. What do you think?

jd-alexander: That's a good point especially since that's the location where this issue is arising. Will address. Thanks for looking into this 🙏

jd-alexander: I made some changes 465485e Let me know if this is along the lines that you are thinking.

shiki: Looks okay but probably a tad unnecessary considering that `ReaderPostListFragment` uses `mActionableEmptyView.subtitle` directly. It's probably simpler to just do this instead of creating a function:

```
mActionableEmtpyView.subtitle.setContentDescription(null);
```

If we do want functions, a better, more elaborate, but promotes encapsulation is to have:

```
class ActionableEmptyView {
    private lateinit var subtitle: WPTextView // <-- private

   func setSubtitleText(text: String) {
      subtitle.text = ...
      subtitle.contentDescription = ...
   } 
}
```

Hiding `subtitle` and having `setSubtitleText` means we'll never forget to set `contentDescription` again. This is just a rough translation of what I'm thinking. The work would definitely be much bigger. 😄 

Anyway, what do you think of simplifying it?

jd-alexander: Good points. 

> mActionableEmtpyView.subtitle.setContentDescription(null);

Yes, I was thinking to do this but didn't want to access it directly but it's already public so might as well access it directly. 

> Anyway, what do you think of simplifying it for now?

I will just use the first option for now since it's highly unlikely this will popup again because most `TextView` objects don't need a content description since the text acts like that.

jd-alexander: Simplified  🙇 5834fd1

WordPress/src/main/java/org/wordpress/android/ui/ActionableEmptyView.kt
```diff
@@ -137,6 +137,13 @@ class ActionableEmptyView : LinearLayout {
         layout.layoutParams = params
     }
 
+    /**
```
shiki: @jd-alexander I guess we'll need to remove this? 😉

jd-alexander: Good idea. Modified in f1d6f75

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -1434,6 +1434,7 @@ private void setEmptyTitleDescriptionAndButton(boolean requestFailed) {
         int heightToolbar = getActivity().getResources().getDimensionPixelSize(R.dimen.toolbar_height);
         int heightTabs = getActivity().getResources().getDimensionPixelSize(R.dimen.tab_height);
         mActionableEmptyView.updateLayoutForSearch(false, getEmptyViewTopMargin());
+        mActionableEmptyView.subtitle.setContentDescription(null);
```
shiki: I think the most appropriate location is here where most of the `subtitle` manipulation are happening:

https://github.com/wordpress-mobile/WordPress-Android/blob/ff4c3b3876720a69aafb5ee837d4e38981dcd3d8/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java#L1541-L1560

This line will be more discoverable there.

jd-alexander: I didn't put it there because the line that's actually setting the `contentDescription` is here

PR #10921, opened 2019-12-08
Accessibility : Post Settings TalkBack Improvements

### LINKED ISSUE #10895: Accessibility : Post Settings TalkBack Improvements
For this audit the governing principles are as follows ([inspired by a previous audit](https://github.com/wordpress-mobile/WordPress-Android/issues/7288))

- All the clickable elements should be marked as so.
- All nonactionable and decorative elements should be non-focusable.
- Relevant content should be grouped together.
- All important elements should be described through content description.

### Talkback Audit 

* Post Settings
	* First option says "Status Draft in pager". Correct the pager reference. 
	* Second options reads "Publish Immediately, Imediately". One of the "immediately"s need to be removed.
	* Password,Categories, Tags,Slug, Location, Excerpt reads as"not set,not set" 
	* Excerpt even when set still says "not set". This issue correlates to the avoid so it should be addressed in one go.

jd-alexander:
Fixes #10895

This PR addresses several Settings that are represented as `TextViews` that had their `hints` being announced along with their `text` properties, which are identical. So now, the hint announcements are disabled, so the announcement is more concise. 

Another enhancement that was made was the addition of accessibility headings to several headings such as Categories & Tags, Featured Image, to name a few. Defining headings as "accessibility headings" allows the user to perform heading navigation and also to be informed of the headings during traversal. 

To test:

**Disabled hints**

1. Go to Post Settings. 
2. Select either Password, Categories or Location and only one property i.e text will be announced instead of the duplicate behavior before this change.
3. Navigate to Post Settings → Publish
4. Select Date & Time and Notifications and the hints aren't announced. 

**Accessibility headings**

1. Go to Post Settings. 
2. Select either Categories & Tags, Featured Image or More Options and you hear heading being announced. 
3. To verify the heading navigation simply swipe up until you reach the setting of "Headings". Then you can swipe left or right to cycle through the headings. For more details visit the [support documentation on navigation.
](https://support.google.com/accessibility/android/answer/6006598?hl=en)

PR submission checklist:

#### Reviewing 

### DIFF
--- a/libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/AccessibilityUtils.java
+++ b/libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/AccessibilityUtils.java
@@ -45,6 +45,11 @@ public class AccessibilityUtils {
         });
     }
 
+    /**
+     * When the minsdk is 28 this can be replaced by adding android:accessibilityHeading="true" as a property to the
+     * view's xml declaration.
+     * @param view that will become a heading.
+     */
     public static void enableAccessibilityHeading(@NonNull View view) {
         setAccessibilityDelegateSafely(view, new AccessibilityDelegateCompat() {
             @Override public void onInitializeAccessibilityNodeInfo(View host, AccessibilityNodeInfoCompat info) {
--- a/libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/AccessibilityUtils.java
+++ b/libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/AccessibilityUtils.java
@@ -12,6 +12,8 @@ import androidx.core.view.accessibility.AccessibilityNodeInfoCompat;
 
 import com.google.android.material.snackbar.Snackbar;
 
+import org.wordpress.android.util.AppLog.T;
+
 import static android.content.Context.ACCESSIBILITY_SERVICE;
 
 public class AccessibilityUtils {
@@ -35,7 +37,7 @@ public class AccessibilityUtils {
     }
 
     public static void disableHintAnnouncement(@NonNull TextView textView) {
-        ViewCompat.setAccessibilityDelegate(textView, new AccessibilityDelegateCompat() {
+        setAccessibilityDelegateSafely(textView, new AccessibilityDelegateCompat() {
             @Override public void onInitializeAccessibilityNodeInfo(View host, AccessibilityNodeInfoCompat info) {
                 super.onInitializeAccessibilityNodeInfo(host, info);
                 info.setHintText(null);
@@ -44,11 +46,24 @@ public class AccessibilityUtils {
     }
 
     public static void enableAccessibilityHeading(@NonNull View view) {
-        ViewCompat.setAccessibilityDelegate(view, new AccessibilityDelegateCompat() {
+        setAccessibilityDelegateSafely(view, new AccessibilityDelegateCompat() {
             @Override public void onInitializeAccessibilityNodeInfo(View host, AccessibilityNodeInfoCompat info) {
                 super.onInitializeAccessibilityNodeInfo(host, info);
                 info.setHeading(true);
             }
         });
+    }
+
+    public static void setAccessibilityDelegateSafely(View view,
+                                                      AccessibilityDelegateCompat accessibilityDelegateCompat) {
+        if (ViewCompat.hasAccessibilityDelegate(view)) {
+            final String errorMessage = "View already has an AccessibilityDelegate.";
+            if (PackageUtils.isDebugBuild()) {
+                throw new RuntimeException(errorMessage);
+            }
+            AppLog.e(T.UTILS, errorMessage);
+        } else {
+            ViewCompat.setAccessibilityDelegate(view, accessibilityDelegateCompat);
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/AccessibilityUtils.java
```diff
@@ -26,4 +33,22 @@ public static int getSnackbarDuration(Context ctx, int defaultDuration) {
         return defaultDuration == Snackbar.LENGTH_INDEFINITE ? Snackbar.LENGTH_INDEFINITE
                 : isAccessibilityEnabled(ctx) ? SNACKBAR_WITH_ACTION_DURATION_IN_MILLIS : defaultDuration;
     }
+
+    public static void disableHintAnnouncement(@NonNull TextView textView) {
+        ViewCompat.setAccessibilityDelegate(textView, new AccessibilityDelegateCompat() {
```
malinajirka: It might be a nice improvement to check if the view already has a delagate and log an error(possible crash in Debug perhaps?) if it does as we'll override i. Wdyt?

libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/AccessibilityUtils.java
```diff
@@ -26,4 +33,22 @@ public static int getSnackbarDuration(Context ctx, int defaultDuration) {
...
+            @Override public void onInitializeAccessibilityNodeInfo(View host, AccessibilityNodeInfoCompat info) {
+                super.onInitializeAccessibilityNodeInfo(host, info);
+                info.setHintText(null);
+            }
+        });
+    }
+
+    public static void enableAccessibilityHeading(@NonNull View view) {
+        ViewCompat.setAccessibilityDelegate(view, new AccessibilityDelegateCompat() {
```
malinajirka: Same as above

libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/AccessibilityUtils.java
```diff
@@ -26,4 +33,22 @@ public static int getSnackbarDuration(Context ctx, int defaultDuration) {
...
+    public static void disableHintAnnouncement(@NonNull TextView textView) {
+        ViewCompat.setAccessibilityDelegate(textView, new AccessibilityDelegateCompat() {
+            @Override public void onInitializeAccessibilityNodeInfo(View host, AccessibilityNodeInfoCompat info) {
+                super.onInitializeAccessibilityNodeInfo(host, info);
+                info.setHintText(null);
+            }
+        });
+    }
+
+    public static void enableAccessibilityHeading(@NonNull View view) {
```
malinajirka: NitPick: We might want to add a comment to this method that it can be removed and replaced with `android:accessibilityHeading="true"` when minSdk == 28. Wdyt?

libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/AccessibilityUtils.java
jd-alexander: This sounds like a good idea so if anyone comes in the future trying to add a delegate there are no weird side effects.

libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/AccessibilityUtils.java
jd-alexander: Awesome suggestion!

jd-alexander:
> Thank you @jd-alexander! I've checked the code and tested the app and it look great overall.
I have left some comments with possible minor improvements, but none of them is necessary. I marking this PR as ready for merge - feel free to merge it without any changes.

Thanks for the review @malinajirka 😄 

> Possible improvement: I think it'd be nice to add a heading to "Status, Publish, Password". It's a bit weird that you can now scan through headings but you never reach the first section. Wdyt?

This is a really good idea. I am going to think of a heading that would work here.

Options in my mind are : 

1. Publishing
2. General 

Is this something editorial could help with?

malinajirka:
>Options in my mind are :
> - Publishing
> - General
>
>Is this something editorial could help with?

I like both suggestions. Yes, I think editorial could definitely help. We might want to create a gif/image with notes so they can quickly understand what we are trying to achieve.

jd-alexander:
> I like both suggestions. Yes, I think editorial could definitely help. We might want to create a gif/image with notes so they can quickly understand what we are trying to achieve.

Sounds good.

jd-alexander:
@malinajirka I just checked iOS and the heading is called "Publish" so I am thinking to just add that for consistency.

libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/AccessibilityUtils.java
jd-alexander: Resolved in 76bcce4

libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/AccessibilityUtils.java
jd-alexander: Resolved in 30ff588

libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/AccessibilityUtils.java
jd-alexander: Resolved in 30ff588

jd-alexander:
@malinajirka Publish header has been added in dd8f9e5 & 23479a6

Let me know your thoughts on the manner in which I am setting the accessibility delegate. Thank you 🙇

PR #10925, opened 2019-12-10
Saving and restoring subfilter to/from shared prefs

### LINKED ISSUE #10884: User selected subfilter is not restored closing and opening the app again
With [Reader navigation revised](https://href.li/?https://github.com/wordpress-mobile/WordPress-Android/pull/10792) as part of the Information Architecture project, we introduced a subfilter capability.

### Expected behavior
When selecting a subfilter we want the user selction to be saved so that it can be restored when closing/opening the app.

### Actual behavior
When selecting a subfilter (different from the All Sites that is default one) closing and opening again the app, defaults the subfilter to the default one (All Sites)

- In the Reader select a subfilter other than All SItes
- Close tha app
- Open it again and notice the subfilter is defaulted again to All Sites

### Note
A solution can be to save the selected subfilter to the shared prefs as it is currently done for the main filter.

develric:
Fixes #10884 

This is related to the Information Architecture modifications currently behind feature flag in zalpha flavor. Test the feature in zalpha then smoke test the reader in vanilla flavor.

### zalpha flavor
- In the Reader select a subfilter other than All SItes
- Close the app
- Open it again and notice the subfilter is not All Sites but is set again to the subfilter that was selected before to close the app
- Test the above with All Sites/Different Tags/Different Sites

### vanilla flavor
- Smoke test the reader

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterListItem.kt
```diff
@@ -1,19 +1,107 @@
...
 
 sealed class SubfilterListItem(val type: ItemType) {
     open var isSelected: Boolean = false
-    open val onClickAction: ((filter: SubfilterListItem) -> Unit)? = null
+    open var onClickAction: ((filter: SubfilterListItem) -> Unit)? = null
     open val label: UiString? = null
 
+    companion object {
+        @JvmStatic
+        fun fromJson(json: String): SubfilterListItem {
```
planarvoid: What do you think about moving this logic into a separate injected class where it could be tested? Maybe create something like `SubfilterListItemMapper`? Actually can't you use a simpified data class (without inheritance, with nullable fields) and Gson? Using JSON object feels a little error prone.

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterListItem.kt
```diff
@@ -30,35 +118,46 @@ sealed class SubfilterListItem(val type: ItemType) {
...
+        override var onClickAction: ((filter: SubfilterListItem) -> Unit)?
+    ) : SubfilterListItem(SITE_ALL) {
+        override val label: UiString = UiStringRes(R.string.reader_filter_all_sites)
+    }
 
     data class Site(
-        override val label: UiString,
         override var isSelected: Boolean = false,
-        override val onClickAction: (filter: SubfilterListItem) -> Unit,
+        override var onClickAction: ((filter: SubfilterListItem) -> Unit)?,
```
planarvoid: I think it might really be a better solution to have a different object that is stored to AppPrefs which you'd use to build the `SubfilterListItem`. The mapper could be called directly in the `ViewModel` and this way you wouldn't need to make this a `var`. What do you think?

develric: Makes sense 👍! Introduced a `SubfilterListItemMapper`

develric:
Hi @planarvoid 👋! Agree on the Gson usage. I made some modifications; not fully sure of it, but hope I got your comments intention. Let me know, thanks 😊!

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterListItem.kt
```diff
@@ -30,35 +33,46 @@ sealed class SubfilterListItem(val type: ItemType) {
...
     data class SectionTitle(override val label: UiString) : SubfilterListItem(SECTION_TITLE)
 
     object Divider : SubfilterListItem(DIVIDER)
 
     data class SiteAll(
-        override val label: UiString,
         override var isSelected: Boolean = false,
-        override val onClickAction: (filter: SubfilterListItem) -> Unit
-    ) : SubfilterListItem(SITE_ALL)
+        override val onClickAction: ((filter: SubfilterListItem) -> Unit)
```
planarvoid: I think this additional parenthesis is added by accident :-), the same goes for the Site and Tag classes.

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterListItemMapper.kt
```diff
@@ -0,0 +1,92 @@
...
+        json: String,
+        onClickAction: ((filter: SubfilterListItem) -> Unit),
+        isSelected: Boolean
+    ): SubfilterListItem {
+        val gson = Gson()
+
+        val mappedItem = if (json.isEmpty()) {
+            MappedItem(type = SITE_ALL.value)
+        } else {
+            gson.fromJson(json, MappedItem::class.java)
```
planarvoid: maybe only initialize Gson when the json is not empty?

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterListItemMapper.kt
```diff
@@ -0,0 +1,92 @@
...
+
+private data class MappedItem(
+    val type: Int,
+    var feedId: Long = 0,
+    var blogId: Long = 0,
+    var tagSlug: String = "",
+    var tagType: Int = 0
+)
+
+class SubfilterListItemMapper @Inject constructor() {
```
planarvoid: I'd add a few unit tests for this class.

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterListItem.kt
develric: Sure 🤦‍♂️! They staied there while playing around with making them nullable 😊. Changed [here](c2838ce62e13f6eb8a80776aa1db1655a434f2f9)

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterListItemMapper.kt
develric: Makes sense 😊. Moved [here](f47b9b2dab27e072d97cd6e34379c39089b9f015)

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterListItemMapper.kt
develric: Agree 👌. Added here 4e937721a5282bab82594ce00be957cbdd2fe88f

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterListItemMapper.kt
```diff
@@ -0,0 +1,94 @@
...
+    private val readerUtilsWrapper: ReaderUtilsWrapper,
+    private val readerBlogTableWrapper: ReaderBlogTableWrapper
+) {
+    fun fromJson(
+        json: String,
+        onClickAction: (filter: SubfilterListItem) -> Unit,
+        isSelected: Boolean
+    ): SubfilterListItem {
+        val mappedItem = if (json.isEmpty()) {
+            MappedItem(type = SITE_ALL.value)
```
khaykov: This is just my personal opinion, but I try to avoid using generic names for non-generic elements. A more specific name, like `MappedSubfilterListItem`  makes it easier to distinguish it from other elements and tells you what is it about.

develric: Agree 👍! Changed here f4da91bb363a762553322e5999f6b3863ca0f0c6

develric:
Hi @khaykov 👋and thanks for the review 😎! I agree on your comment and changed accordingly. Hope I got it right, let me know 🙇‍♂️!

PR #10929, opened 2019-12-10
Replace date locally changed with UTC

planarvoid:
There was a bug that caused date locally changed to be in 2 formats, one ending with `+0000` and the other one with `+00:00`. This makes `equals` always fail and causes unnecessary saves to the database even when there was no actual change.  

To test:
* Try saving posts and check with debug that the save to DB is not called when there is no change

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
@@ -14,7 +14,6 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.preference.PreferenceManager;
 import android.text.TextUtils;
-import android.util.Log;
 import android.view.ContextThemeWrapper;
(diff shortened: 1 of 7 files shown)


malinajirka:
Nice catch @planarvoid! I've been thinking about this solution and I'm not sure it's safe. Some of the methods we use throughout the codebase count on a certain date format. I believe they won't work as expected if we change it. Imho it's especially risky change since afaiu we already have some of the data stored in a persistence storage so existing clients will need to handle this inconsistency.

For example, if you go to `UploadActionUseCaseTest` and you replace all usages of `iso8601FromTimestamp` with `iso8601UTCFromTimestamp` the tests won't work since the tested class is using `timestampFromIso8601Millis` which returns 0 for not-supported formats such as "+00:00. 

Wdyt?

jkmassel:

malinajirka:
Closing this PR since we don't plan to merge it.

PR #10932, opened 2019-12-10
Remove escaped double quotes in translated files

Tug:
This is a hot fix which updates the extra double quotes in gutenberg strings for the following locales:
- `cs`
- `en-rAU`
- `en-rCA`
- `en-rGB`
- `fr`
- `he`
- `id`
- `ko`
- `nl`
- `ru`
- `sv`
- `tr`

This patch was done by running a few search and replace and some manual fixes:
1. Replace `<string name="gutenberg_native_(.*)">\\"(.*)\\"</string>` with `<string name="gutenberg_native_$1">$2</string>`

2. Replace `<string name="gutenberg_native_(.*)">« (.*) »</string>` with `<string name="gutenberg_native_$1">$2</string>`

3. Replace `<string name="gutenberg_native_(.*)">“(.*)”</string>` with `<string name="gutenberg_native_$1">$2</string>`

4. Replace `<string name="gutenberg_native_(.*)">“(.*)\\"</string>` with `<string name="gutenberg_native_$1">$2</string>`

5. Replace `<string name="gutenberg_native_(.*)">\\"(.*)”</string>` with `<string name="gutenberg_native_$1">$2</string>`

After that we'll need to merge https://github.com/wordpress-mobile/gutenberg-mobile/pull/1637 on gutenberg-mobile and update the translations in glotpress. OR keep manually patching the translations until this is fixed.

To test: Run the production app and switch to one of the locale mentioned above, load a page or a post with gutenberg and make sure strings (such as the placeholder for an empty block) do not contain double quotes

PR submission checklist:


hypest:
👋 @Tug , my impression is that the different locale files will get overwritten during the normal release process so, does this replacement here means that step in the process should be skipped? @jkmassel , you probably know better here, what's your recommendation?

Tug:
> In the PR description, can you also elaborate on how come the en-US strings.xml file doesn't need any changes related to the double quotes? Thanks!

Double quotes are not escaped in the original `strings.xml` thus they'll be parsed as wrappers for the string by the android tool (as expected).

This PR just removes double quotes escaped by Glotpress

Tug:
@hypest Updated the original `strings.xml` as discussed on slack

Issue #10918, opened 2019-12-06
Post-Signup Interstitial Master Issue


renanferrari:
This is a master issue for the work around the post-signup interstitial. This screen will be shown once, after signup or login, if the user account has no sites.

## Blueprint
![blueprint-android](https://user-images.githubusercontent.com/830056/70346081-400e3a00-183c-11ea-90e3-bf10645b4777.jpg)

## Tasks
- [x] **Create interstitial screen** ([issue](https://github.com/wordpress-mobile/WordPress-Android/issues/10923), [pull request](https://github.com/wordpress-mobile/WordPress-Android/pull/10933))
  - [x] Create layout
    - [x] Standard
    - [ ]  ~~Dark Mode~~
    - [x] Landscape
    - [x] Tablet
  - [x] Implement basic functionality
    - [x] Make **Create New Site** button open its respective screen
    - [x] Make **Add a Self-Hosted Site** button open its respective screen
    - [x] Make **Not Right Now** button open the main screen
      - [x] Ensure it redirects to the _Reader_ tab
- [x] **Display interstitial screen** ([issue](https://github.com/wordpress-mobile/WordPress-Android/issues/10969), [pull request](https://github.com/wordpress-mobile/WordPress-Android/pull/10976))
  - [x] After signup epilogue screen 
  - [x] After login epilogue screen
  - [x] Ensure it's being shown only once per app install
- [x] **Track interstitial screen events** ([issue](https://github.com/wordpress-mobile/WordPress-Android/issues/10985), [pull request](https://github.com/wordpress-mobile/WordPress-Android/pull/11014))
  - [x] When interstitial is seen 
  - [x] When the **Create New Site** button is clicked
  - [x] When the **Add a Self-Hosted Site** button is clicked
  - [x] When the **Not Right Now** button is clicked / the screen is dismissed

## Notes
- There's an updated version of the blueprint, which I'll update this issue with once I have it in hand.
- I'm going to update the tasks with links for the specific issues and PRs once they're up.

PR #10933, opened 2019-12-10
Post-Signup Interstitial: Create interstitial screen

### LINKED ISSUE #10923: Post-Signup Interstitial: Create interstitial screen
## Designs

<table>
  <thead>
    <tr>
      <th>Standard</th>
      <th>Dark Mode</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>
        
      </td>
      <td>
        
      </td>
    </tr>
  </tbody>
  <thead>
    <tr>
      <th colspan=2>Landscape</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td colspan=2>
        
      </td>
    </tr>
  </tbody>
  <thead>
    <tr>
      <th colspan=2>Tablet</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td colspan=2>
        
      </td>
    </tr>
  </tbody>
</table>

## Tasks
- [x] Create layout
  - [x] Standard
  - [ ] ~~Dark Mode~~
  - [x] Landscape
  - [x] Tablet
- [x] Implement basic functionality
  - [x] Make **Create New Site** button open its respective screen
    - [ ] ~~Ensure the site creation flow completes successfully~~
  - [x] Make **Add a Self-Hosted Site** button open its respective screen
    - [ ] ~~Ensure the site connection flow completes successfully~~
  - [x] Make **Not Right Now** button open the main screen
    - [x] Ensure it redirects to the _Reader_ tab

**Update**: Added new design spec for landscape layout

## Notes
- Master Issue: #10918

renanferrari:
Fixes #10923

### Dark Mode
The dark mode layout has been scraped for now, as it depends on the work being done at the [material-theme](https://github.com/wordpress-mobile/WordPress-Android/tree/feature/material-theme) branch.

#### Notes
There are some slight differences between the layouts implemented here and the ones in the design spec (mainly the status bar color). Most of these differences are due to the fact that I'm reusing existing theme and styles instead of creating new ones just for this screen. It is my understanding that there is an ongoing effort to update the theme of the app as a whole, so I still need to make sure what would be the best approach here.

## To test

To be able to test this, you need to:
- Build and launch a debug version of the app.
- Use `adb` to start the Activity when needed: 
`adb shell am start -n org.wordpress.android/org.wordpress.android.ui.accounts.PostSignupInterstitialActivity`

### Standard layout
1. Start the `PostSignupInterstitialActivity` on a phone in portrait mode.
2. Verify that the layout is compliant with the design spec on #10923.

### Landscape layout
1. Start the `PostSignupInterstitialActivity` on a phone in landscape mode.
2. Verify that the layout is compliant with the design spec on #10923.

### Tablet layout
1. Start the `PostSignupInterstitialActivity` on a tablet.
2. Verifiy both texts and buttons don't expand horizontally more than what's been defined in the design spec on #10923.
3. Rotate the tablet.
4. Verify that the layout doesn't change.

### Create new site button
1. Sign in to an account without any sites.
2. Start the `PostSignupInterstitialActivity` while `WPMainActivity` is on the back stack (more details on this below).
3. Tap on the **Create WordPress.com Site** button.
4. Verifiy that the `SiteCreationActivity` has been opened.
5. Complete the site creation flow.
6. Verify that the `WPMainActivity` is shown and the newly created site is selected.

### Add self-hosted site button
1. Sign in to an account without any sites.
2. Start the `PostSignupInterstitialActivity` while `WPMainActivity` is on the back stack (more details on this below).
3. Tap on the **Add Self-Hosted Site** button.
4. Verifiy that the `LoginActivity` has been opened.
5. Complete the site connection flow.
6. Verify that the `WPMainActivity` is shown and the newly created site is selected.

### Dismissal button
1. Sign in to an account without any sites.
2. Start the `PostSignupInterstitialActivity`.
3. Tap on the **Not Right Now** button.
4. Verify that the `WPMainActivity` has been opened on the Reader tab.

#### Notes
Please, check my comments below for more details on the assumptions I made while working on the buttons functionality and what is currently needed before moving forward.


WordPress/src/main/java/org/wordpress/android/ui/accounts/PostSignupInterstitialActivity.kt
```diff
@@ -0,0 +1,48 @@
...
+         * I'm leaving this here until I get some confirmation on that.
+         **/
+        create_new_site_button.setOnClickListener {
+            /** TODO: Should we listen for the result?
+             * The assumption here is that an instance of WPMainActivity will be on the back stack, which may be true
+             * depending on where this activity is initiated. If this proves to be hard to guarantee, then we can try
+             * listening for the result and then pass it forward to WPMainActivity ourselves. If not,
+             * WPMainActivity.onResume() will handle the new site for us automatically by calling
+             * WPMainActivity.initSelectedSite().
+             **/
```
renanferrari: Wanna give more context on this. The plan is to call `PostSignupInterstitialActivity` from `LoginEpilogueActivity` and `SignupEpilogueActivity`, which, by what I've seen so far, seems to make this assumption true.

WordPress/src/main/java/org/wordpress/android/ui/accounts/PostSignupInterstitialActivity.kt
```diff
@@ -0,0 +1,48 @@
...
+            ActivityLauncher.newBlogForResult(this)
+            finish()
+        }
```
renanferrari: Wanna give more context on this. The plan is to call `PostSignupInterstitialActivity` from `LoginEpilogueActivity` and `SignupEpilogueActivity`, which, by what I've seen so far, seems to make both assumptions true.

WordPress/src/main/res/values/strings.xml
```diff
@@ -2356,6 +2356,14 @@
 
     <string name="google_error_timeout">Google took too long to respond. You may need to wait until you have a stronger internet connection.</string>
 
+    <!-- Post-Signup Interstitial -->
+    <string name="post_signup_interstitial_title">Welcome to WordPress</string>
+    <string name="post_signup_interstitial_subtitle">Whatever you want to create or share, we’ll help you do it right here.</string>
+    <string name="post_signup_interstitial_create_new_site">Create a new site</string>
+    <string name="post_signup_interstitial_add_self_hosted_site">Add a self-hosted site</string>
```
renanferrari: Maybe it's possible to reuse some strings from existing screens. @SylvesterWilmott is currently considering this.

SylvesterWilmott:
Looks great so far!

>There are some slight differences between the layouts implemented here and the ones in the design spec (mainly the status bar color).

I am aware of this, the status bar colour is fine as is in the screenshots

We sometimes use a container width of 512dp max-width in some areas around the app. Can we try applying this constraint to the tablet layout (keeping the existing left and right paddings)?

WordPress/src/main/res/values/strings.xml
SylvesterWilmott: We can reuse the string `Add self-hosted site` but let's keep `Create a new site`. In previous reviews `Create WordPress.com site` was voted too wordy.

renanferrari:
> Looks great so far!

Glad you like it!

> We sometimes use a container width of 512dp max-width in some areas around the app. Can we try applying this constraint to the tablet layout (keeping the existing left and right paddings)?

If I understand correctly, that's something we can try for sure. I would just need to adjust [this](https://github.com/wordpress-mobile/WordPress-Android/pull/10933/files#diff-a38f8955f26ed07cc283ff53e02b57dbR326) to `512dp`.

WordPress/src/main/res/values/strings.xml
SylvesterWilmott: After some further discussion, I think we can also reuse `Create WordPress.com site` to be explicit about the button's function.

So the button strings should be:

`Create WordPress.com site`
`Add self-hosted site`

WordPress/src/main/java/org/wordpress/android/ui/accounts/PostSignupInterstitialActivity.kt
```diff
@@ -0,0 +1,29 @@
...
+
+        create_new_site_button.setOnClickListener {
+            ActivityLauncher.newBlogForResult(this)
+            finish()
+        }
+
+        add_self_hosted_site_button.setOnClickListener {
+            ActivityLauncher.addSelfHostedSiteForResult(this)
+            finish()
+        }
```
renanferrari: The assumption here is that we don't need to listen for the result of any of those calls because `WPMainActivity` would already be on the back stack and would handle the eventual new site automatically by calling [`WPMainActivity.initSelectedSite()`](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java#L1180) inside [`WPMainActivity.onResume()`](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java#L637).

The reason I believe `WPMainActivity` would already be on the back stack is based on the fact that we would be starting `PostSignupInterstitialActivity` from `LoginEpilogueActivity` and `SignupEpilogueActivity`. Those classes are initialized by `LoginActivity` when it calls these methods [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java#L98-L106) and [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java#L108-L118), which starts `WPMainActivity` and pass some arguments to it indicating that the epilogue screen should be shown. The other initialization methods on `ActivityLauncher` seem to either be called by `WPMainActivity` itself or are part of other distinct flows around Jetpack or Magic Link, for example. 

If this assumption proves to be incorrect, then I think we could try following the same approach used by `LoginActivity` I described above, where we would listen for the result ourselves and then call `WPMainActivity` with some arguments to instruct it to start the site creation/connection flow as needed.

Another thing to consider is that this topic of whether or not these flows complete correctly might be more related to the next task in the pipeline for this feature ([**Show interstitial screen**](https://github.com/wordpress-mobile/WordPress-Android/issues/10918)). If we realize this would need a more detailed discussion, I'm in favor of letting it be discussed later on the next PR. That way, this one would be focused only on the layouts and the basic behavior of the screen buttons.

SylvesterWilmott:
The changes look good to me! Thanks @renanferrari

WordPress/src/main/java/org/wordpress/android/ui/accounts/PostSignupInterstitialActivity.kt
maxme: > If this assumption proves to be incorrect, then I think we could try following the same approach used by LoginActivity I described above, where we would listen for the result ourselves and then call WPMainActivity with some arguments to instruct it to start the site creation/connection flow as needed.

I think you're right on that assumption. If that's not correct, please don't duplicate `initSelectedSite` in there, this one is too sensitive. 

> I'm in favor of letting it be discussed later on the next PR. That way, this one would be focused only on the layouts and the basic behavior of the screen buttons.

+1, we'll see how you integrate this in the signup/login flow and depending on that we can discuss what solution is the best.

WordPress/src/main/res/values/strings.xml
renanferrari: I have now updated the strings to reuse existing ones from the **Site Picker** screen. I have also updated the screenshots to reflect this change.

WordPress/src/main/res/drawable/img_illustration_construct_site_190dp.xml
```diff
@@ -0,0 +1,310 @@
...
+        android:strokeWidth="2"
+        android:strokeColor="#FFF"
+        android:strokeLineCap="round"
+        android:strokeLineJoin="round" />
+    <path
+        android:fillColor="#FFF"
+        android:fillType="evenOdd"
+        android:pathData="M103.177,81.402H59.46v-9.947h51.915v1.749a8.198,8.198 0,0 1,-8.198 8.198" />
+    <path
+        android:fillColor="#C3C4C7"
```
maxme: @SylvesterWilmott - in other vector images we use color references (it would be `gray_10` here for example). Any reason we don't do that here? Examples in [img_illustration_cloud_off_152dp.xml](https://github.com/wordpress-mobile/WordPress-Android/blob/f8ca54caed8114a6399fa8d3ef2807f1ec07ab1e/WordPress/src/main/res/drawable/img_illustration_cloud_off_152dp.xml) or [img_illustration_add_media_150dp.xml](https://github.com/wordpress-mobile/WordPress-Android/blob/f8ca54caed8114a6399fa8d3ef2807f1ec07ab1e/WordPress/src/main/res/drawable/img_illustration_add_media_150dp.xml)

WordPress/src/main/res/values/strings.xml
```diff
@@ -2356,6 +2356,14 @@
 
     <string name="google_error_timeout">Google took too long to respond. You may need to wait until you have a stronger internet connection.</string>
 
+    <!-- Post-Signup Interstitial -->
+    <string name="post_signup_interstitial_title">Welcome to WordPress</string>
+    <string name="post_signup_interstitial_subtitle">Whatever you want to create or share, we’ll help you do it right here.</string>
+    <string name="post_signup_interstitial_create_new_site">@string/site_picker_create_wpcom</string>
+    <string name="post_signup_interstitial_add_self_hosted_site">@string/site_picker_add_self_hosted</string>
```
maxme: We should mark these strings `translatable="false"`. Note: I'm fine using `site_picker_create_wpcom` and `site_picker_add_self_hosted` directly from the interstitial if we find more generic names.

maxme:
> I'll create a feature branch, so you can merge this one and target the same feature branch in the upcoming PRs.

Pushed a new branch: `feature/10918-post-signup-interstitial`

WordPress/src/main/java/org/wordpress/android/ui/accounts/PostSignupInterstitialActivity.kt
renanferrari: Thanks for taking the time to review this, @maxme.

> If that's not correct, please don't duplicate `initSelectedSite` in there, this one is too sensitive.

In that case I'd probably try some other, simpler, naive solution first, like just calling `WPMainActivity` the same way I'm doing in [`ActivityLauncher.viewReader()`](https://github.com/renanferrari/WordPress-Android/blob/issue/10923-create-interstitial-screen/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java#L200-L205) and hoping `onResume()` would do its thing then. Only if _that_ didn't work I'd try to go for some more invasive solution and even then I think we probably wouldn't need to touch `initSelectedSite()` in any way, so I wouldn't worry about that. 😉I appreciate you pointing this out anyway.

> +1, we'll see how you integrate this in the signup/login flow and depending on that we can discuss what solution is the best.

Great! 👍 So I guess there isn't much left to do here. We can probably remove the _Do Not Merge_ tag now and wait for further reviews.

WordPress/src/main/res/values/strings.xml
renanferrari: > We should mark these strings `translatable="false"`.

Done. I had those in initially but forgot to add them back after removing them. 🤦‍♂️ Good catch!

> Note: I'm fine using `site_picker_create_wpcom` and `site_picker_add_self_hosted` directly from the interstitial if we find more generic names.

I'm fine with that as well. I was just trying to follow the perceived convention here, but after talking to @theck13 and @malinajirka about it, I now know the convention is actually very loose.

Personally, I would prefer naming them like this:

```
<string name="welcome_to_wordpress_title">Welcome to WordPress</string>
<string name="welcome_to_wordpress_subtitle">Whatever you want to create or share, we’ll help you do it right here.</string>
```
or even
```
<string name="welcome_to_wordpress">Welcome to WordPress</string>
<string name="whatever_you_want_to_create_or_share">Whatever you want to create or share, we’ll help you do it right here.</string>
```

and then rename `site_picker_create_wpcom` and `site_picker_add_self_hosted` to something like `create_wpcom_site` and `add_self_hosted_site`. 

In previous projects, I also used some prefixes to help distinguish different contexts in a more generic way, like `label_`, `action_`, `message_`, `title_`, etc. But I wouldn't use these here.

That being said, I decided to keep the screen name prefixes and leave the existing strings unchanged so I could continue with the non-intrusive approach I've been following so far. As @malinajirka said, every approach here has its downsides and there's probably a larger conversation that needs to take place to discuss this further, so I'd rather leave it that way until then.

renanferrari:
> Having separate layouts is often a source of issues, I wish we could find a simple way to only have one.

I agree. I have actually tried a few things, but none were quite good.

Our first iteration had a single layout, but the user had to scroll to see the buttons on a phone in landscape mode, which is not a good UX. @SylvesterWilmott fixed that with the new landscape layout. 

From there, we experimented reusing the landscape layout on a tablet, but that didn't look good – and even if it did, we would still have at least two layouts: _default_ and _landscape_. 

The only alternatives I could think of that would let us achieve a single layout again would be to either update the [`ConstraintLayout`](https://developer.android.com/jetpack/androidx/releases/constraintlayout) lib to `2.0.0-beta2` so we could leverage its new [Flow](https://developer.android.com/reference/androidx/constraintlayout/helper/widget/Flow) helper or to use the [`FlexboxLayout`](https://github.com/google/flexbox-layout) lib from Google. I just don't know if these alternatives can be considered "simple".

WordPress/src/main/res/values/strings.xml
maxme: Works for me.

> In previous projects, I also used some prefixes to help distinguish different contexts in a more generic way, like label_, action_, message_, title_, etc. But I wouldn't use these here.

We've been discussing having more formal string reference, it could also help with translations (like `action.verb.trash` is easier to translate than just `trash`), but we're not there yet.

WordPress/src/main/res/drawable/img_illustration_construct_site_190dp.xml
SylvesterWilmott: Interesting. @theck13 do you know anything about this? Was it done as part of the colour changes?

theck13: Yes, all vector images should be using `@color/` or `@android:color/` references only.  There are also extra attributes that are unnecessary in this vector file.  All paths should only need the `fillColor` and `pathData` attributes.  Take a look at any of the other `img_illustration_` files for expected format.  It may take some work with the image source file to merge the stroke with the path or separate the stroke into a different path depending on the desired effect.

renanferrari: I think I could fix the color references by doing a quick find and replace, but I'm wondering if there's any other way this is usually done?

theck13: That's the only way I know to do it.

WordPress/src/main/res/layout/post_signup_interstitial_default.xml
```diff
@@ -19,28 +19,30 @@
...
-    <TextView
+    <org.wordpress.android.widgets.WPTextView
         android:id="@+id/title_view"
         android:layout_width="0dp"
         android:layout_height="wrap_content"
         android:layout_marginBottom="@dimen/margin_medium"
         android:text="@string/post_signup_interstitial_title"
         android:textAlignment="center"
         android:textAppearance="@style/TextAppearance.AppCompat.Title"
+        app:fixWidowWords="true"
```
maxme: Well spotted, 👍 I actually forgot about that one ;)

WordPress/src/main/res/drawable/img_illustration_construct_site_190dp.xml
maxme: Vector file was updated in 866c12a

maxme:
> There are some slight differences between the layouts implemented here and the ones in the design spec (mainly the status bar color). Most of these differences are due to the fact that I'm reusing existing theme and styles instead of creating new ones just for this screen. It is my understanding that there is an ongoing effort to update the theme of the app as a whole, so I still need to make sure what would be the best approach here.

Sorry, I just realized that, but I think we could reuse `@style/WordPress.Button.Primary` for the 2 primary buttons. I'll add a comment in the follow up PR.

PR #10945, opened 2019-12-11
Issue/10399 add search for 'Set parent'

yomiyusuf:
### Description
In the 'Set parent' screen, there is no **Search** functionality for the list of possible parents. This is a potential UX challenge for sites with a numerous number of pages. It's might be difficult to scan through a long list of pages.
This PR implements a search feature in the 'Set Parent' screen. 

### Implementation
The main part of this PR is the LiveData `filteredPages`, in `PageParentViewModel.kt ` which uses the `LiveDataUtils.mergeAsyncNotNull` method to merge the Pages list and the query into one Observable. 

`filteredPages` in turn, uses the `filterListByQuery` to filter the list using the supplied query string.

The merging is done on a background thread, this way, even with a large pages list, the UI will remain responsive.

### To test:
Use an account that has a site with a couple of pages
- Go to Home Screen > Pages
- Tap the contextual menu of any of the listed pages and tap 'Set parent'
- Tap the overflow menu icon in the toolbar and tap search
- Enter any query to search the list of possible parent pages

### PR submission checklist:

### ToDo
- Add other translations for new strings used in this PR

Note: @planarvoid


WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageParentViewModel.kt
```diff
@@ -128,4 +153,26 @@ class PageParentViewModel
         }
         return children + grandchildren
     }
+
+    internal fun filterListByQuery(list: List<PageItem>, query: String): List<PageItem> {
+        if (query.isEmpty()) {
+            return list
+        } else {
+            val filtered = list.filter { pageItem ->
+                (pageItem.type != PARENT) || // keep the TOP_LEVEL_PARENT and DIVIDER types
```
planarvoid: Is there a reason why you always show the Top level page on the filtered result list?

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageParentViewModel.kt
```diff
@@ -128,4 +153,26 @@ class PageParentViewModel
...
+                        (pageItem.type == PARENT &&
+                                (pageItem as ParentPage).title.toLowerCase(Locale.getDefault()).contains(
```
planarvoid: `contains` method has an `ignoreCase` parameter which you can use to simplify this expression

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageParentViewModel.kt
```diff
@@ -47,6 +51,12 @@ class PageParentViewModel
     private lateinit var site: SiteModel
     private var isStarted: Boolean = false
     private var page: PageModel? = null
+    private var _isSearchInUse = false
```
planarvoid: We only use the `_` prefix when you have 2 instances of the same live data, one is private and mutable and the other one is public and immutable. I'd just use `isSearchInUse` here.

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageParentViewModel.kt
```diff
@@ -47,6 +51,12 @@ class PageParentViewModel
...
+
+    val filteredPages: LiveData<List<PageItem>> = mergeAsyncNotNull(defaultScope, _pages, _searchQuery) { list, query ->
```
planarvoid: I think I'd still use the original name `pages` here. The View doesn't need to care if the data is filtered, it just shows whatever it gets.

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageParentViewModel.kt
yomiyusuf: I struggled a bit to decide on this one. But my thinking was I want to keep access to 'Top Level' option. I assume the user will not be searching for 'Top Level' but can easily change their mind during the search that they want to use 'Top level'. In that case, they have to cancel search first.

WordPress/src/main/java/org/wordpress/android/ui/pages/PageParentFragment.kt
```diff
@@ -70,12 +75,23 @@ class PageParentFragment : Fragment() {
         setHasOptionsMenu(true)
     }
 
+    @SuppressLint("RestrictedApi")
```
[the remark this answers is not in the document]
yomiyusuf: I have deleted the lint suppress.

>  I also don't think usage of the setOptionalIconsVisible is necessary

But to be clear, are you saying we do not need to show the icons? Because by default, Android won't show it. 
Meanwhile, in EditPostSettingsFragment#showLocationPickerOrPopupMenu (line 1124), I see that we are forcing Android to show the icons.

yomiyusuf:
@planarvoid 
the ktlint check is failing with `Process 'command '/usr/local/openjdk-8/bin/java'' finished with non-zero exit value 1`. I have checked but could not locate the problem. 
Do you have an Idea what the problem might be?

yomiyusuf:
@yomiyusuf

WordPress/src/main/java/org/wordpress/android/ui/pages/PageParentFragment.kt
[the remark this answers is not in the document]
planarvoid: That was one more thing I wanted to mention. Why did you add the overflow menu at all? 
The `EditPostSettingsFragment#showLocationPickerOrPopupMenu` is different, that's not a menu in toolbar. The standard behaviour is to directly show the search icon in the toolbar (as it's done in all the other screens).

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageParentViewModel.kt
planarvoid: I think that once you start searching, it's fine to lose the top level (since normally it's always in the first position). I assume that searching means you just want stuff that matches your query.

WordPress/src/main/java/org/wordpress/android/ui/pages/PageParentFragment.kt
```diff
@@ -135,6 +183,8 @@ class PageParentFragment : Fragment() {
 
     override fun onSaveInstanceState(outState: Bundle) {
         linearLayoutManager?.let { outState.putParcelable(listStateKey, it.onSaveInstanceState()) }
+        previousSearchString = searchView.query.toString()
```
planarvoid: I don't think the solution with a `previousSearchString` value in Fragment that is stored to saved instance state is correct (and clean) because right now you have a live data instance that holds the query and you also have it in the fragment. The advantage of ViewModel is that it survives the rotation. The only case when it's not enough is when the application is destroyed and recreated. In that case it's perfectly fine to lose the query (and search state). There are two options you can take to handle this correctly. The first one is simpler, you can expose `_searchQuery` and observe it. If the query is not empty and the search view is not showing, you'd show the search view and set the query. This would recreate the Search state on rotation. 
The other option (that's little cleaner) is to create a new class that would hold the UI state. This would be something like:
```
data class PageParentUiState(val pages: List<PageItem>, val searchQuery: String); 
```
You'd expose this instead of just `List<PageItem>` and do the logic in the `observe` method.

PR #10955, opened 2019-12-13
Issue/10637 material people management

### LINKED ISSUE #10637: Implement Material theme on People List, Details and Invite screens
We should switch widgets used at People List, Details and Invite screens to their material counterparts, remove color references where possible, and switch to using material `textAppearance` attributes.

I expect issues with the Flow layout we are using for usernames/email entry. Maybe we can switch them to Chips?

khaykov:
Fixes #10637 

This PR moves People Management to Material Theme and adds Dark Theme support. There are some code changes for cosmetic reasons.

While refactoring is not in the scope of this PR, since I touched those files, I cleaned up/optimized some of the redundant code as suggested by AS, so do not be alarmed by the size of the PR. I suggest turning on "Hide whitespace changes" option during the review.

comparison_people

To test:
Confirm the visual integrity and functionality of the  People Management screens in both light and night modes.
The theme can be changed from App Settings.

PR submission checklist:


mattmiklic:
Following up on discussion from Slack, I just looked at how this works in Calypso and I see it does the same thing (turning blue when validated). I think we should leave it as-is for that reason, since it’s really just a visual detail.

PR #10957, opened 2019-12-13
Issue/10691 primary secondary editor actions

### LINKED ISSUE #10691: Editor: Extract primary and secondary actions logic and write tests

malinajirka:
Fixes #10691 

- Moves primary and secondary editor actions into a single injectable/mockable file
- Adds Sentry logging when a user opens a Trashed post in the editor as it shouldn't happen
- Adds Sentry logging when a user without publishing rights opens PUBLISHED/SCHEDULED/PRIVATE posts in the editor as it shouldn't happen
- Adds unit tests for all the primary and secondary actions

To test:
1. Smoke test that the correct primary and secondary actions are visible in the editor (more info about the primary and secondary actions can be found on paCBwp-95-p2)
2. There shouldn't be any user facing changes

PR submission checklist:


WordPress/src/test/java/org/wordpress/android/ui/posts/editor/EditorActionsProviderTest.kt
```diff
@@ -0,0 +1,172 @@
...
+                getPrimaryAndSecondaryActions(PostStatus.PUBLISHED, userCanPublish = true)
+        // Assert
+        assertThat(primaryAction).isEqualTo(PrimaryEditorAction.UPDATE)
+        assertThat(secondaryAction).isEqualTo(SecondaryEditorAction.NONE)
+    }
+
+    @Test
+    fun `Verify an error is logged when the user doesn't have publishing rights and works with a PUBLISHED post`() {
+        // Arrange & Act
+        val (_, _) =
```
jd-alexander: Hi, @malinajirka do we have to create a variable here to store the result if we aren't using it?  The names have also been omitted for brevity, so I want to understand what's the intent of creating it if it's to communicate that a result is being returned. Let me know. Thanks 😄

WordPress/src/test/java/org/wordpress/android/ui/posts/editor/EditorActionsProviderTest.kt
```diff
@@ -0,0 +1,172 @@
...
+                getPrimaryAndSecondaryActions(PostStatus.SCHEDULED, userCanPublish = true)
+        // Assert
+        assertThat(primaryAction).isEqualTo(PrimaryEditorAction.SCHEDULE)
+        assertThat(secondaryAction).isEqualTo(SecondaryEditorAction.PUBLISH_NOW)
+    }
+
+    @Test
+    fun `Verify an error is logged when the user doesn't have publishing rights and works with a SCHEDULED post`() {
+        // Arrange & Act
+        val (_, _) =
```
jd-alexander: Same question as above 😄

WordPress/src/test/java/org/wordpress/android/ui/posts/editor/EditorActionsProviderTest.kt
```diff
@@ -0,0 +1,172 @@
...
+    fun `Verify an error is logged when the user doesn't have publishing rights and works with a PRIVATE post`() {
+        // Arrange & Act
+        val (_, _) =
+                getPrimaryAndSecondaryActions(PostStatus.PRIVATE, userCanPublish = false)
+        // Assert
+        verify(crashLoggingUtilsWrapper, times(2)).log(any<String?>())
+    }
+
+    @Test
+    fun `Verify actions of a UNKNOWN post`() {
```
jd-alexander: English nitpick 😅 unknown is normally preceded by "an" 😄

WordPress/src/test/java/org/wordpress/android/ui/posts/editor/EditorActionsProviderTest.kt
malinajirka: Good point! ;D Removed in 19492ed4c6166f9254f03c4b6a2d79691ca70ff3

WordPress/src/test/java/org/wordpress/android/ui/posts/editor/EditorActionsProviderTest.kt
malinajirka: Fixed in 19492ed4c6166f9254f03c4b6a2d79691ca70ff3

WordPress/src/test/java/org/wordpress/android/ui/posts/editor/EditorActionsProviderTest.kt
malinajirka: ❤️ Finally someone who is willing to correct my mistakes in English - I honestly appreciate it;)!
Fixed in 19492ed4c6166f9254f03c4b6a2d79691ca70ff3

malinajirka:
Thanks for the review @jd-alexander;). It's ready for another round :)

WordPress/src/test/java/org/wordpress/android/ui/posts/editor/EditorActionsProviderTest.kt
jd-alexander: 😄

WordPress/src/main/java/org/wordpress/android/util/CrashLoggingUtilsWrapper.kt
```diff
@@ -0,0 +1,13 @@
+package org.wordpress.android.util
+
+import dagger.Reusable
+import javax.inject.Inject
+
+@Reusable
+class CrashLoggingUtilsWrapper @Inject constructor() {
+    fun log(message: String?) = CrashLoggingUtils.log(message)
+    fun log(exception: Throwable) = CrashLoggingUtils.log(exception)
+    fun logException(tr: Throwable, tag: AppLog.T) = CrashLoggingUtils.logException(tr, tag)
```
jd-alexander: I can see that the parameter of type Throwable is called `tr`  in the `CrashLoggingUtils` function. Do we want to leave this as is or should we rename it to `throwable` or something else? Our parameter names have been pretty consistent with the types they represent.

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/EditorActionsProvider.kt
```diff
@@ -0,0 +1,105 @@
...
+                PostStatus.TRASHED, PostStatus.UNKNOWN -> SecondaryEditorAction.SAVE_AS_DRAFT
+            }
+        } else {
+            // User doesn't have publishing permissions
+            when (postStatus) {
+                PostStatus.DRAFT,
+                PostStatus.PENDING,
+                PostStatus.UNKNOWN -> SecondaryEditorAction.NONE
+                PostStatus.TRASHED -> {
+                    // TODO if this log won't appear in Sentry, we should start throwing IllegalStateException
```
jd-alexander: Another English nitpick 😉 Instead of saying "if this log won't appear" it's better to say "if this log doesn't appear" 
won't = is final
doesn't = it might or might not happen.

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/EditorActionsProvider.kt
```diff
@@ -0,0 +1,105 @@
...
+                    remoteLoggingUtils.log(
+                            "User shouldn't be able to open a trashed post in an editor " +
+                                    "without publishing rights."
+                    )
+                    SecondaryEditorAction.SAVE_AS_DRAFT
+                }
+                PostStatus.PUBLISHED,
+                PostStatus.SCHEDULED,
+                PostStatus.PRIVATE -> {
+                    // TODO if this log won't appear in Sentry, we should start throwing IllegalStateException
```
jd-alexander: Same as above 😉

WordPress/src/test/java/org/wordpress/android/ui/posts/editor/EditorActionsProviderTest.kt
```diff
@@ -0,0 +1,168 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.ArgumentMatchers.any
+import org.mockito.junit.MockitoJUnitRunner
+import org.wordpress.android.fluxc.model.post.PostStatus
+import org.wordpress.android.util.CrashLoggingUtilsWrapper
+
+@RunWith(MockitoJUnitRunner::class)
+class EditorActionsProviderTest {
+    private var crashLoggingUtilsWrapper: CrashLoggingUtilsWrapper = mock()
```
jd-alexander: Is there a reason why this wrapper has to be mutable? If not then we could utilize `val` here. Let me know your thoughts 👍

WordPress/src/main/java/org/wordpress/android/util/CrashLoggingUtilsWrapper.kt
malinajirka: Makes sense 👍 Fixed in 2bd574b13e755a27c0ad280ed27305f757841657

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/EditorActionsProvider.kt
malinajirka: Fixed in 2bd574b13e755a27c0ad280ed27305f757841657

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/EditorActionsProvider.kt
malinajirka: Fixed in 2bd574b13e755a27c0ad280ed27305f757841657

WordPress/src/test/java/org/wordpress/android/ui/posts/editor/EditorActionsProviderTest.kt
malinajirka: No reason - I had `setUp` method but since I removed it, it's better to use `val` now;). Fixed in 2bd574b13e755a27c0ad280ed27305f757841657

malinajirka:
Thanks Joel ;)! I've done all the proposed changes in 2bd574b13e755a27c0ad280ed27305f757841657 ;)

jd-alexander:
Changes look good @malinajirka LGTM 🚢

PR #10959, opened 2019-12-13
Add tests for featured image support in EditPostSettings

### LINKED ISSUE #10688: Write tests for FeaturedImageHelper

malinajirka:
Fixes #10688 

Adds tests for featured image support in the EditPostActivity/PostSettings.

To test:
- I think running the CircleCI tests should be enough

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/posts/FeaturedImageHelper.kt
```diff
@@ -85,6 +85,7 @@ internal class FeaturedImageHelper @Inject constructor(
             media.localPostId = localPostId
         } else {
             AppLog.e(T.MEDIA, "Upload featured image can't be invoked without a valid local post id.")
+            return false
```
develric: This seems fine to me. I was just wondering if the Toast message in the calling function `doMediaUriSelected` (`R.string.file_not_found`) can be misleading in this condition. Maybe we can consider to return an enum instead of a boolean to discriminate on the message to show (maybe better) or change the message to include both cases (maybe worse)? Not sure if it's worth it tough, let me know wdyt 😊.

WordPress/src/test/java/org/wordpress/android/ui/posts/FeaturedImageHelperTest.kt
```diff
@@ -0,0 +1,297 @@
...
+        val result = featuredImageHelper.queueFeaturedImageForUpload(0, createSiteModel(), mock(), "")
+        // Assert
+        assertThat(result).isFalse()
+    }
+
+    @Test
+    fun `queueFeaturedImageForUpload returns false when the provided localPostId is empty`() {
+        // Arrange
+        whenever(fluxCUtilsWrapper.mediaModelFromLocalUri(anyOrNull(), anyOrNull(), anyInt()))
+                .thenReturn(createMediaModel(markedLocallyAsFeatured = false))
```
develric: Extremely super minor np: I would consider to align on using or not the named argument for markedLocallyAsFeatured parameters throughout the file.

malinajirka: Good call! ;) Fixed in 1c2625dcff70cc9551838f796fec36eef9aff073

WordPress/src/main/java/org/wordpress/android/ui/posts/FeaturedImageHelper.kt
malinajirka: The "invalid post" shouldn't happen, unless we have a bug in the app so I kept the same message. But I agree it might be confusing not just for users but especially for the HEs if the user contacts them. Fixed in 215b794638496b88cbe7fb1785c8c010a5187b6e

malinajirka:
Thanks for the review and for the suggestions @develric! I've fixed both issues. It's ready for another round 🙇

PR #10963, opened 2019-12-16
Fastlane: Push current branch only, don't rely on local git configuration

maxme:
Push current branch only, don't rely on local git configuration


maxme:
@loremattei bumped `release-toolkit` to version 0.8.2

maxme:
Note: Wait before #10965 before merging that one.

loremattei:
Hey @maxme! This look good to me, but I think we also need the updated `Gemfile.lock` or the repository will turn out to be dirty after `bundle install`.

maxme:
Thanks @loremattei, You can see I'm not familiar with ruby gems ;) - updated in 08d3cd1

PR #10968, opened 2019-12-16
Accessibility : Login Improvements 

### LINKED ISSUE #10960: Accessibility : Login Screen description should focus first
### Expected behavior
The description of the login screen should gain focus and begin its announcement. 

### Actual behavior
The back button gains focus instead. 

1. Go to Login. 
2. Once TalkBack is enabled you will hear the announcement.

jd-alexander:
Fixes #10960

To test:

1. Go to Login.
2. Once TalkBack is enabled you will hear the announcement of the email login description.

PR submission checklist:

### DIFF
--- a/libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginEmailFragment.java
+++ b/libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginEmailFragment.java
@@ -14,6 +14,7 @@ import android.view.ContextThemeWrapper;
 import android.view.View;
 import android.view.View.OnClickListener;
 import android.view.ViewGroup;
+import android.view.accessibility.AccessibilityEvent;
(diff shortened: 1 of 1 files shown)


malinajirka:
Thanks @jd-alexander!

I'm a bit concerned that this change might potentially have a negative impact on the UX. In my experience, users with visual impairment tend to explore the screen first when they enter it. However, if we force focus a field, they might enter their email address without exploring the screen -> they might miss the self-hosted + Google login options. Wdyt?

jd-alexander:
Yes thanks for this comment Jirka. I had this same discussion with Jayson yesterday and decided I’m going to close the PR because it will also break the way how TalkBack traversal functions 🙏🏾

jd-alexander:
So currently the page is focusing on the back button first and then it announces the activity title.  I guess the onus is on the user to then navigate to the rest of the page.

PR #10976, opened 2019-12-17
Post-Signup Interstitial: Display interstitial screen

### LINKED ISSUE #10969: Post-Signup Interstitial: Display interstitial screen
This is the second task of the Post-Signup Interstitial feature. The goal here is to display the interstitial screen only once per app install, at the end of successful signup and login flows.

## Tasks
- [x] Display interstitial screen after signup epilogue screen 
- [x] Display interstitial screen after login epilogue screen
- [x] Ensure it's being shown only once per app install

## Notes
- Master Issue: #10918

renanferrari:
Fixes #10969

For this PR, I have made the following changes:
- Created a new preference called `SHOULD_SHOW_POST_SIGNUP_INTERSTITIAL` to control whether the screen has already been shown or not.
  - Created a new `UndeletablePrefKey`, since we want the screen to be shown only once per install.
  - It has a default value of `true`, which is changed to `false` when the interstitial is created.
- Created a new method in the `ActivityLauncher` class to launch the interstitial activity.
  - It uses the `TaskStackBuilder` class and the `FLAG_ACTIVITY_CLEAR_TOP` to guarantee the interstitial activity will always be launched on top of `WPMainActivity`.
- Updated `LoginActivity` and `SignupEpilogueActivity` with a logic to check for the need to launch the interstitial activity.
  - Since we need to check if the user has sites, a `SiteStore` instance is now injected into those classes.
  - `LoginActivity` also checks for the newly created preference. 
  - Since this logic is pretty simple, I tried to avoid [early abstractions](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).
- Updated the UI tests for the signup flow, since it needed to account for the interstitial.

| Signup | Login |
|---|---|
| signup | login |

#### 1. Signup
1. Clear app data.
2. On the initial screen, tap _**Sign up for WordPress.com**_ and _**Sign up with email**_.
3. Enter an email address _that is not associated_ with a WordPress.com account.
4. Complete the signup flow.
5. Confirm that the post-signup interstitial screen is shown.

#### 2. Login (account without sites)
1. Clear app data.
2. On the initial screen, tap _**Log in**._
3. Enter an email address that is associated with an account _that doesn't have any sites_.
4. Complete the login flow.
5. Confirm that the login epilogue screen is **not** shown.
6. Confirm that the post-signup interstitial screen is shown instead.

#### 3. Login (account with sites)
1. Clear app data.
2. On the initial screen, tap _**Log in**_.
3. Enter an email address that is associated with an account _that has at least one site_.
4. Complete the login flow.
5. Confirm that the post-signup interstitial screen  is **not** shown.
6. Confirm that the login epilogue screen is shown instead.

#### 4. Once per app install
1. Complete test 2 and sign out.
2. Repeat test 2 _without clearing the app data_.
3. Confirm that the post-signup interstitial screen is **not** shown.

#### 5. Site creation and site connection flows
1. Complete test 1 or 2.
2. On the post-signup interstitial screen, tap either **Create WordPress.com Site** or **Add Self-Hosted Site**.
3. Complete the flow.
4. Confirm that the main screen is shown.

## Notes
- `LoginEpilogueActivity` seems to not only be shown at the end of the login flow itself, but also in [other flows](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java#L248-L254) related to Jetpack and reauthentication.
  - Because of that, I initially thought of an edge case where I would need to make sure an already authenticated user with no sites wouldn't see the interstitial screen from one of these flows.
  - After a bit of investigation, this seems like it wouldn't be needed. The Jetpack flow seems to only happen when the user already has a site and the interstitial screen doesn't seem to be out of place for the reauthentication flow.
- I’m assuming the only behavior I need to guarantee upon completion (successfully or not) of either the site creation or site connection flow is that the user will be redirected to `WPMainActivity`.
- I haven't felt the need to implement a ViewModel for this screen yet. I might do that after we start tracking the screen events as we may need some unit tests then.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginEpilogueActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginEpilogueActivity.java
@@ -9,9 +9,11 @@ import androidx.fragment.app.FragmentTransaction;
 import org.wordpress.android.R;
 import org.wordpress.android.WordPress;
 import org.wordpress.android.fluxc.store.AccountStore;
+import org.wordpress.android.fluxc.store.SiteStore;
 import org.wordpress.android.ui.ActivityLauncher;
 import org.wordpress.android.ui.accounts.login.LoginEpilogueFragment;
 import org.wordpress.android.ui.accounts.login.LoginEpilogueListener;
+import org.wordpress.android.ui.prefs.AppPrefs;
 import org.wordpress.android.util.LocaleManager;
 
 import java.util.ArrayList;
@@ -24,6 +26,7 @@ public class LoginEpilogueActivity extends AppCompatActivity implements LoginEpi
     public static final String ARG_OLD_SITES_IDS = "ARG_OLD_SITES_IDS";
 
     @Inject AccountStore mAccountStore;
+    @Inject SiteStore mSiteStore;
 
     @Override
     protected void attachBaseContext(Context newBase) {
@@ -67,7 +70,12 @@ public class LoginEpilogueActivity extends AppCompatActivity implements LoginEpi
 
     @Override
     public void onContinue() {
-        setResult(RESULT_OK);
+        if (mSiteStore.hasSite()) {
+            setResult(RESULT_OK);
+        } else if (AppPrefs.shouldShowPostSignupInterstitial()) {
+            ActivityLauncher.showPostSignupInterstitial(this);
+        }
+
         finish();
     }
 }
--- a/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginEpilogueActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginEpilogueActivity.java
@@ -9,11 +9,9 @@ import androidx.fragment.app.FragmentTransaction;
 import org.wordpress.android.R;
 import org.wordpress.android.WordPress;
 import org.wordpress.android.fluxc.store.AccountStore;
-import org.wordpress.android.fluxc.store.SiteStore;
 import org.wordpress.android.ui.ActivityLauncher;
 import org.wordpress.android.ui.accounts.login.LoginEpilogueFragment;
 import org.wordpress.android.ui.accounts.login.LoginEpilogueListener;
-import org.wordpress.android.ui.prefs.AppPrefs;
 import org.wordpress.android.util.LocaleManager;
 
 import java.util.ArrayList;
@@ -26,7 +24,6 @@ public class LoginEpilogueActivity extends AppCompatActivity implements LoginEpi
     public static final String ARG_OLD_SITES_IDS = "ARG_OLD_SITES_IDS";
 
     @Inject AccountStore mAccountStore;
-    @Inject SiteStore mSiteStore;
 
     @Override
     protected void attachBaseContext(Context newBase) {
@@ -70,12 +67,7 @@ public class LoginEpilogueActivity extends AppCompatActivity implements LoginEpi
 
     @Override
     public void onContinue() {
-        if (mSiteStore.hasSite()) {
-            setResult(RESULT_OK);
-        } else if (AppPrefs.shouldShowPostSignupInterstitial()) {
-            ActivityLauncher.showPostSignupInterstitial(this);
-        }
-
+        setResult(RESULT_OK);
         finish();
     }
 }
--- a/WordPress/src/main/java/org/wordpress/android/ui/accounts/SignupEpilogueActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/accounts/SignupEpilogueActivity.java
@@ -7,9 +7,15 @@ import androidx.appcompat.app.AppCompatActivity;
 import androidx.fragment.app.FragmentTransaction;
 
 import org.wordpress.android.R;
+import org.wordpress.android.WordPress;
+import org.wordpress.android.fluxc.store.SiteStore;
+import org.wordpress.android.ui.ActivityLauncher;
 import org.wordpress.android.ui.accounts.signup.SignupEpilogueFragment;
 import org.wordpress.android.ui.accounts.signup.SignupEpilogueListener;
+import org.wordpress.android.ui.prefs.AppPrefs;
 import org.wordpress.android.util.LocaleManager;
+
+import javax.inject.Inject;
 
 public class SignupEpilogueActivity extends AppCompatActivity implements SignupEpilogueListener {
     public static final String EXTRA_SIGNUP_DISPLAY_NAME = "EXTRA_SIGNUP_DISPLAY_NAME";
@@ -20,6 +26,8 @@ public class SignupEpilogueActivity extends AppCompatActivity implements SignupE
     public static final String MAGIC_SIGNUP_PARAMETER = "new_user";
     public static final String MAGIC_SIGNUP_VALUE = "1";
 
+    @Inject SiteStore mSiteStore;
+
     @Override
     protected void attachBaseContext(Context newBase) {
         super.attachBaseContext(LocaleManager.setLocale(newBase));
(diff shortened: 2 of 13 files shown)


libs/mocks/WordPressMocks/src/main/assets/mocks/mappings/wpcom/reader/rest_v12_read_following_mine.json
```diff
@@ -1,7 +1,7 @@
 {
     "request": {
         "method": "GET",
-        "urlPattern": "/rest/v1.2/read/following/mine.*"
+        "urlPattern": "/rest/v1.(1|2)/read/following/mine.*"
```
renanferrari: Is it ok to update this mock file like this? [`rest_v11_me_sites.json`](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/libs/mocks/WordPressMocks/src/main/assets/mocks/mappings/wpcom/me/rest_v11_me_sites.json) seems to do the same, but I'm not sure what's the best practice here – specially because these files are prefixed with the API version.

WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginEpilogueActivity.java
```diff
@@ -67,7 +70,12 @@ public void onConnectAnotherSite() {
 
     @Override
     public void onContinue() {
-        setResult(RESULT_OK);
+        if (mSiteStore.hasSite()) {
+            setResult(RESULT_OK);
+        } else if (AppPrefs.shouldShowPostSignupInterstitial()) {
```
maxme: 👍 I think that makes sense to store this in the app prefs and skip the interstitial during sign in. We have some users who regularly sign out and sign in back, and we might not want to show them the interstitial every time.

Note: See my next comment in the SignupEpiloguActivity.

WordPress/src/main/java/org/wordpress/android/ui/accounts/SignupEpilogueActivity.java
```diff
@@ -51,7 +61,12 @@ protected void addSignupEpilogueFragment(String name, String email, String photo
 
     @Override
     public void onContinue() {
-        setResult(RESULT_OK);
+        if (mSiteStore.hasSite()) {
+            setResult(RESULT_OK);
+        } else if (AppPrefs.shouldShowPostSignupInterstitial()) {
```
maxme: Here, I think we shouldn't check for the preference value and always show the interstitial. I can't think of any use case where we shouldn't nudge new users to create a new site.

WordPress/src/main/java/org/wordpress/android/ui/prefs/AppPrefs.java
```diff
@@ -989,6 +992,14 @@ public static boolean getSystemNotificationsEnabled() {
         return getBoolean(UndeletablePrefKey.SYSTEM_NOTIFICATIONS_ENABLED, true);
     }
 
+    public static void setShouldShowPostSignupInterstitial(boolean shouldShow) {
+        setBoolean(UndeletablePrefKey.SHOULD_SHOW_POST_SIGNUP_INTERSTITIAL, shouldShow);
```
maxme: 👍 `UndeletablePrefKey` is the correct choice.

libs/mocks/WordPressMocks/src/main/assets/mocks/mappings/wpcom/reader/rest_v12_read_following_mine.json
maxme: Yep looks good

WordPress/src/main/java/org/wordpress/android/ui/accounts/SignupEpilogueActivity.java
renanferrari: Can we guarantee a new user will never have a site? If so, I think we can remove the `SiteStore.hasSites()` check. Wasn't sure about this at first, because the user can log in with a self-hosted site (but apparently not _sign up_ with one).

The preference check is needed here to comply with the requirement of only showing the interstitial **once per app install**. Maybe @SylvesterWilmott can give us a bit of context about this.

renanferrari:
> I wonder if we should update the button styling, current style looks a bit off to me. Not directly related to this PR, but we could update the code in here. cf. [#10933 (comment)](https://github.com/wordpress-mobile/WordPress-Android/pull/10933#issuecomment-567492968)

The design spec for the button styling used to be like the one from the [original blueprint], but it has been [updated] to the one you see implemented here. I _think_ this change was made in preparation for dark mode, but I could be wrong. Maybe @SylvesterWilmott can give us a bit of context about this as well.

> There is an "end screen" at the end of the sign in step. I think we should remove it and only have the new interstitial for users who don't have any wordpress.com sites. See how weird the UX looks currently: https://bia.is/s/bez8/signin-interstitial.webm

You're right, it does look weird. I think we can skip the `LoginEpilogueActivity` for users that don't have any sites by calling the interstitial [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java#L242-L247).

WordPress/src/main/java/org/wordpress/android/ui/accounts/SignupEpilogueActivity.java
SylvesterWilmott: >Can we guarantee a new user will never have a site? If so, I think we can remove the SiteStore.hasSites() check. Wasn't sure about this at first, because the user can log in with a self-hosted site (but apparently not sign up with one).

I can't think of any scenarios where a new user will already have a site but is there a downside to performing the `hasSites` check on sign up?

>The preference check is needed here to comply with the requirement of only showing the interstitial once per app install. Maybe @SylvesterWilmott can give us a bit of context about this.

I don't have any strong opinions about this. We can show the interstitial per account created.

SylvesterWilmott:
>The design spec for the button styling used to be like the one from the original blueprint, but it has been updated to the one you see implemented here. I think this change was made in preparation for dark mode, but I could be wrong. Maybe @SylvesterWilmott can give us a bit of context about this as well.

The pink colour is reserved for primary actions and should be used for one element on a screen. Since neither of the buttons in the interstitial is a primary action, both use the secondary style.

>There is an "end screen" at the end of the sign in step. I think we should remove it and only have the new interstitial for users who don't have any wordpress.com sites. See how weird the UX looks currently: https://bia.is/s/bez8/signin-interstitial.webm

Agree

WordPress/src/main/java/org/wordpress/android/ui/accounts/SignupEpilogueActivity.java
renanferrari: > I can't think of any scenarios where a new user will already have a site but is there a downside to performing the `hasSites` check on sign up?

I just confirmed _there is_ a way for a new user to already have a site: 
- Log in with a self-hosted site address.
- Go to the _Me_ tab and tap **Log in to WordPress.com**.
- You'll be redirected to the [login prologue screen](https://user-images.githubusercontent.com/830056/71200202-6093b680-2276-11ea-9836-0af22937c0f8.png), where you'll be able to enter the signup flow and create a new account normally.

Do we still want to always show the interstitial in this case or can I leave the `hasSites` check?

renanferrari:
> There is an "end screen" at the end of the sign in step. I think we should remove it and only have the new interstitial for users who don't have any wordpress.com sites. See how weird the UX looks currently: https://bia.is/s/bez8/signin-interstitial.webm

I have replaced `LoginEpilogueActivity` entirely for users that don't have any sites. Here's how it looks now:

| Login (without sites) | Login (with sites) |
|---|---|
| after (no sites) | after (with sites) |

WordPress/src/main/java/org/wordpress/android/ui/accounts/SignupEpilogueActivity.java
maxme: Yes and the API might change, user creation might end up with a site created, so better to keep the `hasSites()` check anyway.

maxme:
> The pink colour is reserved for primary actions and should be used for one element on a screen. Since neither of the buttons in the interstitial is a primary action, both use the secondary style.

Perfect, thanks for confirming.

renanferrari:
We had a bug with this implementation for login flows using **Magic Link**. It was later documented in [this issue](https://github.com/wordpress-mobile/WordPress-Android/issues/11019).

PR #10977, opened 2019-12-18
Issue/10641 material activity log

### LINKED ISSUE #10641: Implement Material theme on Activity Log
We should switch widgets used at Activity Log screen to their material counterparts, remove color references where possible, and switch to using material `textAppearance` attributes.

A pretty recent screen so should be pretty straightforward.

khaykov:
Fixes #10641

This PR moves Activity Log screen to Material Theme and adds Dark Theme support. There are some code changes for cosmetic reasons.

While refactoring is not in the scope of this PR, since I touched those files, I cleaned up/optimized some of the redundant code as suggested by AS, so do not be alarmed by the size of the PR. I suggest turning on "Hide whitespace changes" option during the review.

comparison_activitylog

To test:
- Navigate to Activity Log screen and confirm it's visual and functional integrity in both Light and Dark themes. The theme can be changed from App Settings.

PR submission checklist:


khaykov:
@mattmiklic It's no problem to move it out of card :)
comparison_log2

mattmiklic:
Good to go!

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/EventItemViewHolder.kt
```diff
@@ -36,12 +37,16 @@ class EventItemViewHolder(
         text.text = activity.description
 
         if (activity.isButtonVisible) {
-            ContextCompat.getDrawable(container.context, activity.buttonIcon.drawable)?.let { buttonIcon ->
-                val wrapDrawable = DrawableCompat.wrap(buttonIcon).mutate()
-                DrawableCompat.setTint(wrapDrawable, ContextCompat.getColor(container.context, R.color.primary_40))
-                actionButton.setImageDrawable(DrawableCompat.unwrap(wrapDrawable))
-                actionButton.visibility = View.VISIBLE
-            }
+            ContextCompat.getDrawable(container.context, activity.buttonIcon.drawable)
```
planarvoid: What do you think about extracting this function so we can share it? What we want is to something like:
```
fun ImageView.setDrawableResourceWithTint(@IntRes res: Int, @ColorRes tint: Int? = null) 
```
and we'd default tint to primary color? I think this might be valuable all over the place.

khaykov: I thought I already saw an existing utility method that does this and sure thing, it was `ColorUtils.setImageResourceWithTint` :)

khaykov:
Thanks for the review, @planarvoid , I implemented your suggestion :)

jkmassel:

develric:
Hi @khaykov 😊...jumping in just to complete this review. And I say LGTM! Good job! 👍

PR #10980, opened 2019-12-18
Accessibility : Posting Activity Stats Improvement

### LINKED ISSUE #10975: Accessibility : Stats Posting Activity Improvements
### Expected behavior

You should be able to select each month and navigate through the different days to hear an overview of the activity.

### Actual behavior
The entire component is selected but you aren't able to get any details of what's within it. 

1. Go to my site. 
2. Select Stats 
3. Select the "Posting Activity" section

jd-alexander:
Fixes #10975 

## Findings
The Posting Stats was inaccessible when using TalkBack. This improvement makes it more accessible for visually impaired users who are navigating the stats functionality. 

## Solution
The solution was to loop through each `Box` that represents a day and announce the associated stats. This implementation is similar to [what was done on iOS.](https://github.com/wordpress-mobile/WordPress-iOS/pull/13123)  The `Boxes` didn't have a content description so refactoring was done to generate it based on the month, day and amount of posts. 

Before | After 
--------|-------
[Before Video ](https://drive.google.com/open?id=18ErO8-vpZxLuxucriVecOq218nEcmaSK) |   [After Video    ](https://drive.google.com/open?id=186KK4cugtT2PX-SzyCCeGM8ka3UyzPKR)

## Testing

1. Enable TalkBack. 
2. Navigate to the Stats page.
3. Ensure you are on the Insights Tab. 
4. Select each Month under "Posting Activity" and follow the instructions to hear the announcement. 

Tap To Hear Day Stats | Stat Announced 
--------|-------
        |       

All Stats Announced | No Stats Available 
--------|-------
   |    

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/StatsBlockAdapter.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/StatsBlockAdapter.kt
@@ -16,8 +16,8 @@ import org.wordpress.android.ui.stats.refresh.utils.PostingActivityBlockAnnounce
 import org.wordpress.android.util.image.ImageManager
 
 class StatsBlockAdapter(
-    val imageManager: ImageManager, private val postingActivityBlockAnnouncer:
-    PostingActivityBlockAnnouncer
+    val imageManager: ImageManager,
+    private val postingActivityBlockAnnouncer: PostingActivityBlockAnnouncer
 ) : Adapter<BaseStatsViewHolder>() {
     private var items: List<StatsBlock> = listOf()
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/StatsBlockAdapter.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/StatsBlockAdapter.kt
@@ -15,8 +15,10 @@ import org.wordpress.android.ui.stats.refresh.lists.viewholders.LoadingViewHolde
 import org.wordpress.android.ui.stats.refresh.utils.PostingActivityBlockAnnouncer
 import org.wordpress.android.util.image.ImageManager
 
-class StatsBlockAdapter(val imageManager: ImageManager, private val postingActivityBlockAnnouncer:
-PostingActivityBlockAnnouncer) : Adapter<BaseStatsViewHolder>() {
+class StatsBlockAdapter(
+    val imageManager: ImageManager, private val postingActivityBlockAnnouncer:
+    PostingActivityBlockAnnouncer
+) : Adapter<BaseStatsViewHolder>() {
     private var items: List<StatsBlock> = listOf()
 
     fun update(newItems: List<StatsBlock>) {
@@ -32,8 +34,12 @@ PostingActivityBlockAnnouncer) : Adapter<BaseStatsViewHolder>() {
 
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): BaseStatsViewHolder {
         return when (values()[viewType]) {
-            SUCCESS, ERROR, EMPTY -> BlockListViewHolder(parent, imageManager,postingActivityBlockAnnouncer)
-            LOADING -> LoadingViewHolder(parent, imageManager,postingActivityBlockAnnouncer)
+            SUCCESS, ERROR, EMPTY -> BlockListViewHolder(
+                    parent,
+                    imageManager,
+                    postingActivityBlockAnnouncer
+            )
+            LOADING -> LoadingViewHolder(parent, imageManager, postingActivityBlockAnnouncer)
         }
     }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/StatsBlockAdapter.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/StatsBlockAdapter.kt
@@ -12,9 +12,11 @@ import org.wordpress.android.ui.stats.refresh.lists.StatsBlock.Type.values
 import org.wordpress.android.ui.stats.refresh.lists.viewholders.BaseStatsViewHolder
 import org.wordpress.android.ui.stats.refresh.lists.viewholders.BlockListViewHolder
 import org.wordpress.android.ui.stats.refresh.lists.viewholders.LoadingViewHolder
+import org.wordpress.android.ui.stats.refresh.utils.PostingActivityBlockAnnouncer
 import org.wordpress.android.util.image.ImageManager
 
-class StatsBlockAdapter(val imageManager: ImageManager) : Adapter<BaseStatsViewHolder>() {
+class StatsBlockAdapter(val imageManager: ImageManager, private val postingActivityBlockAnnouncer:
+PostingActivityBlockAnnouncer) : Adapter<BaseStatsViewHolder>() {
     private var items: List<StatsBlock> = listOf()
 
     fun update(newItems: List<StatsBlock>) {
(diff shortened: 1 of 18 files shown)


jd-alexander:
TalkBack is announcing Dec as "Deck" will make the content description of the months the long version.

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
```diff
@@ -233,7 +233,8 @@ sealed class BlockListItem(val type: Type) {
     data class LoadingItem(val loadMore: () -> Unit, val isLoading: Boolean = false) : BlockListItem(LOADING_ITEM)
 
     data class ActivityItem(val blocks: List<Block>) : BlockListItem(ACTIVITY_ITEM) {
-        data class Block(val label: String, val boxes: List<Box>)
+        data class Block(val label: String, val boxItems: List<BoxItem>)
+        data class BoxItem(val box: Box, val contentDescription: String? = null)
```
jd-alexander: Let me know what you think about the naming here. I am not sure if I am adhering to the naming convention being used so I am rethinking it. So the BoxItem could be renamed to `Box` and the enum `BoxStyle` or `BoxType`

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/utils/PostingActivityBlockAnnouncer.kt
```diff
@@ -0,0 +1,115 @@
...
+    private var isStatsInPeriod = false
+
+    fun initialize(activityItem: ActivityItem, vararg blockViews: View) {
+        this.activityItem = activityItem
+        this.blockViews = blockViews.toList()
+
+        setupViewDelegate()
+    }
+
+    private fun setupViewDelegate() {
```
jd-alexander: Most of the logic that does the announcement lives within this function. As reviewers how do you feel about this? I have been thinking and I was wondering if it's worth breaking this up a bit more with some tests as well or since the functionality works and it achieves the goal of announcing each daily stats then it okay. Let me know :)

planarvoid (changes requested):
Thanks for the changes @jd-alexander, I've originally omitted this card for accessibility on purpose because I think its value is very questionable. If you use the app without talkback, the purpose of this card is to generally see how active your site was in the past months. It's impossible to see what the single day boxes mean (find what date they represent is very hard), wha the number of views on these days are, and you can't click them. It's purely a visual component and I still believe it should be omitted. If we really want to implement it, I'd approach it differently. First of all, I think reading all the items on click and saying the number of values is not useful and is a completely different use case. What about reading only the top days? Maybe ordered by views. Something like:
"The days with very high views for January are: the 1st, the 3rd...". On click you'd read "The days with high views for January are: the 5th, the 7th ...". This would reduce the number of clicks the user has to do and make the talk back semi-useful. The actual number of views can be reached through a different use case.
From the code point of view, I think we should build the list of content descriptions for a month in the use case (because building it is business logic). We should pass this list on the `Block` element and not have granular content descriptions on the Boxes (since they are really still not clickable). `PostingActivityBlockAnnouncer` doesn't have to be specific to posting activity any more, you'd just pass in a list of content descriptions and a view they are tied to and it would iterate through them on click. I also don't think it needs to be injected since it can't be tested anyway (and it would be used purely as view logic). I'd generally inject classes for 2 reasons, one is they can be mocked in unit tests and the second one is if their construction is too complex (and needs other injected classes). This one is neither. 

Let me know if all of this makes sense!

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/StatsBlockAdapter.kt
```diff
@@ -12,9 +12,13 @@ import org.wordpress.android.ui.stats.refresh.lists.StatsBlock.Type.values
 import org.wordpress.android.ui.stats.refresh.lists.viewholders.BaseStatsViewHolder
 import org.wordpress.android.ui.stats.refresh.lists.viewholders.BlockListViewHolder
 import org.wordpress.android.ui.stats.refresh.lists.viewholders.LoadingViewHolder
+import org.wordpress.android.ui.stats.refresh.utils.PostingActivityBlockAnnouncer
 import org.wordpress.android.util.image.ImageManager
 
-class StatsBlockAdapter(val imageManager: ImageManager) : Adapter<BaseStatsViewHolder>() {
+class StatsBlockAdapter(
+    val imageManager: ImageManager,
+    private val postingActivityBlockAnnouncer: PostingActivityBlockAnnouncer
```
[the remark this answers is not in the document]
jd-alexander: Makes sense. Thanks 🙏

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
jd-alexander: Okay thanks @planarvoid 👌

jd-alexander:
> Thanks for the changes @jd-alexander, I've originally omitted this card for accessibility on purpose because I think its value is very questionable. If you use the app without talkback, the purpose of this card is to generally see how active your site was in the past months. It's impossible to see what the single day boxes mean (find what date they represent is very hard), wha the number of views on these days are, and you can't click them. It's purely a visual component and I still believe it should be omitted. If we really want to implement it, I'd approach it differently. First of all, I think reading all the items on click and saying the number of values is not useful and is a completely different use case. What about reading only the top days? Maybe ordered by views. Something like:
"The days with very high views for January are: the 1st, the 3rd...". On click you'd read "The days with high views for January are: the 5th, the 7th ...". This would reduce the number of clicks the user has to do and make the talk back semi-useful. The actual number of views can be reached through a different use case.

That's a good idea @planarvoid I originally decided to go with this implementation since it's approach was [merged into the iOS app. ](https://a8c.slack.com/archives/DN82QTN6S/p1576549513005100)

> From the code point of view, I think we should build the list of content descriptions for a month in the use case (because building it is business logic). We should pass this list on the Block element and not have granular content descriptions on the Boxes (since they are really still not clickable). 

Okay understood! 

> PostingActivityBlockAnnouncer doesn't have to be specific to posting activity any more, you'd just pass in a list of content descriptions and a view they are tied to and it would iterate through them on click. I also don't think it needs to be injected since it can't be tested anyway (and it would be used purely as view logic). I'd generally inject classes for 2 reasons, one is they can be mocked in unit tests and the second one is if their construction is too complex (and needs other injected classes). This one is neither.

Good point!

> Let me know if all of this makes sense!

Yes, it does. Overall, what's the decision that's normally taken when there's a variation with an implementation on one platform? In this case, I could do the month content description approach but it would be definitely from that of iOS but it should be fine since it's accessibility right?

planarvoid:
> Overall, what's the decision that's normally taken when there's a variation with an implementation on one platform? In this case, I could do the month content description approach but it would be definitely from that of iOS but it should be fine since it's accessibility right?

@jd-alexander that's a very good question. I personally don't think it's super important to have everything 100% the same. I generally try to think about the changes (if they make sense) and if I think something doesn't look right, I give feedback to the designer (or to the iOS team). Sometimes they overlook a better solution and it's fine to iteratively improve one platform based on the findings of devs on the other. I think this is especially important in the accessibility improvements because even little changes to the usability can make the lives of the users so much more difficult/easy (controlling the app without seeing the screen is hard). In this case I'd suggest this solution to the iOS team (maybe you can also document it in a post) but I also think it's fine to have the solutions slightly different.

jd-alexander:
@planarvoid thanks again for the review and thoughts. They have been very helpful. I implemented the solution you discussed in the comments above. I created another PR since this one was specific to the approach being taken. Let me know what you think 😄 I will be closing this one.

PR #10986, opened 2019-12-18
IA reader self hosted enhancements

develric:
This PR aims to bring some bug fix/improvements to the IA modified reader in logged out WP.com but with self-hosted sites user. 

In more details it does the following:

1. Fix a crash that can happen if a WP.com user with self-hosted sites signs out
2. Create a dummy in memory Following Tag to be used to show the default tags for self-hosted users
3. Introduce an auto update step in the reader in case a user changed is detected

cc @osullivanchris 

### Case 1 - crash fix
- Log in with a WP.com account and also add a pure self-hosted site (no jetpack)
- Log out from WP.com account, you are still kept in the app since you have at least one self-hosted site added
- Check that the app does not crash and you can access all the 3 main pages of the app (My site, Reader, Notifications

### Case 2 - Reader for self-hosted site user
- Verify you are logged out from WP.com account and you have added a pure self-hosted site
- Go to reader and check you have a condition similar to the following (the 3 Tabs and subfilter in Following with "Select Filter" text and bottom sheet with default proposed tags only)
<p align="center">
  
</p>

- Smoke test the reader

### Case 3 - Auto update in case of user changed detected
- Log in with a WP.com account and also add a pure self-hosted site
- Go to the Reader and check the list of posts is automatically loaded and you get a condition similar to the following
<p align="center">
  
</p>

- Smoke test the reader
- Log out from WP.com 
- Go to the Reader and check the list of posts is automatically loaded and you get a condition similar to the following
<p align="center">
  
</p>

- Smoke test the reader

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/models/ReaderTag.java
+++ b/WordPress/src/main/java/org/wordpress/android/models/ReaderTag.java
@@ -69,7 +69,7 @@ public class ReaderTag implements Serializable, FilterCriteria {
         return StringUtils.notNullStr(mEndpoint);
     }
 
-    private void setEndpoint(String endpoint) {
+    public void setEndpoint(String endpoint) {
         this.mEndpoint = StringUtils.notNullStr(endpoint);
     }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/FilteredRecyclerView.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/FilteredRecyclerView.java
@@ -101,7 +101,7 @@ public class FilteredRecyclerView extends RelativeLayout {
         mCurrentFilter = filter;
 
         if (mUseTabsForFiltering) {
-            int position = mFilteredPagerAdapter.getItemPosition(filter);
+            int position = filter == null ? -1 : mFilteredPagerAdapter.getItemPosition(filter);
             if (position > -1 && position != mViewPager.getCurrentItem()) {
                 mViewPager.setCurrentItem(position);
             } else if (position > -1 && position == mViewPager.getCurrentItem()) {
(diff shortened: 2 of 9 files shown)


osullivanchris:
Looks mostly good to me

- The default view for case 2 it seems strange not to see any content. But as discussed this solution helps with all use cases not just this one.
- I like initially saying "Select Filter" as it explains to the user the interaction pattern, rather than "All Sites" which we found in user testing is unclear. However I would change the label from "Select Filter" to "Filter"
- The empty state says "Unable to retrieve posts" - would it be worth considering changing the copy here to something like "Select a tag from the filter to see posts" or is it too complex to have a conditional string here just for this use case?

I still think it feels a bit weird. Maybe we can consider having conditional logic just for case 2 to default to the first tag. But I know it is not beautiful logic

planarvoid:
Hey @develric, while testing this PR I've discovered a bug in the bottom sheet recycler view adapter. It's not correctly removing the selected state. The cause of this bug is in `org.wordpress.android.ui.reader.subfilter.viewholders.SubfilterListItemViewHolder` where you only set the selected color in this scenario:
```
if (filter.isSelected) {
                it.setTextColor(ContextCompat.getColor(parent.context, R.color.primary))
            }
```
However, in a recycler view the views get recycled so it might happen that the view with a selected text color gets chosen again (and it does happen if you scroll down in a list). The solution to this problem is to add something like:

```
if (filter.isSelected) {
   it.setTextColor(ContextCompat.getColor(parent.context, R.color.primary))
} else {
  it.setTextColor(ContextCompat.getColor(parent.context, R.color.default))
}
```

planarvoid (commented):
Looks good 👍 . I have one question. Does it make sense to show the "Unable to retrieve posts" message when actually there's no selected tag? Is it too complicated to reword it in this scenario?

develric:
Closing this in favor of #11270

Issue #10151, opened 2019-07-04
Add Tracks to Post Settings - Featured Image functionality


malinajirka:
We don't track the following
- Featured image picked
- Featured image upload canceled
- Retry clicked
- Remove featured image clicked

PR #10998, opened 2019-12-20
Issue/10151 add tracks to post settings featured image actions

ashiagr:
Attempt to address #10151

This PR adds analytics tracks to Post Settings - Featured Image functionality

To test:

- Create a draft or edit a published post
- Click on the overflow menu in the editor -> Post settings
- Click Set Featured Image, choose a photo, wait till image is successfully uploaded and set to image view
- Check that the console contains the following log:
    - `🔵 Tracked: post_settings_featured_image_picked, Properties: {"post_id":<post id>,"post_status":<post status>}`
- Tap on the featured image, click “Remove featured image”
- Check that the console contains the following log:
    - `🔵 Tracked: post_settings_featured_image_remove_clicked, Properties: {"post_id":<post id>,"post_status":<post status>}`
- Click Set Featured Image, choose a photo, tap on the local image while it is uploading and click “Remove featured image”
- Check that the console contains the following log:
    - `🔵 Tracked: post_settings_featured_image_upload_canceled, Properties: {"post_id":<post id>,"post_status":<post status>}`
- Click Set Featured Image, choose photo, disable network so that upload is cancelled, tap on photo to retry, click retry uploading 
- Check that the console contains the following log:
    - `🔵 Tracked: post_settings_featured_image_upload_retry_clicked, Properties: {"post_id":<post id>,"post_status":<post status>}`

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/posts/FeaturedImageHelper.kt
```diff
@@ -137,6 +157,32 @@ internal class FeaturedImageHelper @Inject constructor(
         return FeaturedImageData(FeaturedImageState.REMOTE_IMAGE_LOADING, photonUrl)
     }
 
+    fun trackFeaturedImageEvent(
+        event: TrackableEvent,
+        post: PostImmutableModel,
+        originType: OriginType
+    ) {
+        val currentStat = if (originType == EDIT_POST_SETTINGS) {
```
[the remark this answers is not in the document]
ashiagr: I was thinking in terms of future capabilities we might have like image transformation by clicking the image from the editor itself and possibly syncing updated copy to the server using helper class, in which case origin type will be different and we might want to use different track events.

As discussed earlier, it seems like an overkill. I’ll simplify it. I like your suggestion.

WordPress/src/main/java/org/wordpress/android/ui/posts/FeaturedImageHelper.kt
```diff
@@ -93,7 +103,13 @@ internal class FeaturedImageHelper @Inject constructor(
         return true
     }
 
-    fun cancelFeaturedImageUpload(site: SiteModel, post: PostImmutableModel, cancelFailedOnly: Boolean) {
+    @JvmOverloads
+    fun cancelFeaturedImageUpload(
+        site: SiteModel,
+        post: PostImmutableModel,
+        cancelFailedOnly: Boolean,
+        originType: OriginType? = null
```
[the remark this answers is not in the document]
ashiagr: Same reason, just wanted to keep it optional till the time we really want to track an event for a new capability. I'll try to refactor it.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostSettingsFragment.java
```diff
@@ -934,6 +943,15 @@ public void updateFeaturedImage(long featuredImageId) {
         postRepository.update(postModel -> {
             postModel.setFeaturedImageId(featuredImageId);
             updateFeaturedImageView();
+
+            if (featuredImageId != 0) {
```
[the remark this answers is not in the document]
ashiagr: Makes sense. Thanks 🙏. Fixed in 0fa6128.

WordPress/src/main/java/org/wordpress/android/ui/posts/FeaturedImageHelper.kt
```diff
@@ -137,6 +149,24 @@ internal class FeaturedImageHelper @Inject constructor(
         return FeaturedImageData(FeaturedImageState.REMOTE_IMAGE_LOADING, photonUrl)
     }
 
+    fun trackFeaturedImageEvent(
+        stat: Stat,
+        post: PostImmutableModel?
+    ) {
+        var properties: HashMap<String, Any>? = null
```
malinajirka: We might want to consider simplifying this method to
```
fun trackFeaturedImageEvent(
        stat: Stat,
        post: PostImmutableModel?
    ) {
       if (post == null) {
            analyticsTrackerWrapper.track(stat)
        } else {
            analyticsTrackerWrapper.track(stat, mapOf(POST_ID_KEY to post.id, POST_STATUS_KEY to post.status))
        }
}
```
Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/FeaturedImageHelper.kt
```diff
@@ -137,6 +149,24 @@ internal class FeaturedImageHelper @Inject constructor(
         return FeaturedImageData(FeaturedImageState.REMOTE_IMAGE_LOADING, photonUrl)
     }
 
+    fun trackFeaturedImageEvent(
+        stat: Stat,
```
malinajirka: I'm a bit confused why we decided to change this parameter from `TrackableEvent` to `Stat`. Can you please elaborate?

It seems `TrackableEvent` enum is now pointless since we lost the type safety it brought us. We might want to consider changing this parameter back to `TrackableEvent` and calling `getLabel()` inside this method. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/FeaturedImageHelper.kt
```diff
@@ -137,6 +149,24 @@ internal class FeaturedImageHelper @Inject constructor(
...
+        post: PostImmutableModel?
```
malinajirka: It feels a bit weird that the post is nullable. Featured image always requires a post. We are passing null just because we don't have an access to the post object from the `PhotoPickerActivity`.  I'd consider replacing `PostImmutableModel?` with a post id to which we have access even from the Activity (unless we know we need to track "post_status"). Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/FeaturedImageHelper.kt
ashiagr: My bad :neutral_face:. Fixed in 20a7fe3.

ashiagr:
Thanks once again for reviewing the changes 🙏.

> SideNote: I also realized, we don't track clicks on "SET FEATURED IMAGE" button. I think it might be useful to know how many users clicked on the button, but didn't pick any image. However, we can add that as part of another PR.

I included it in this PR itself.

PR #10999, opened 2019-12-20
Add media upload completion processor

mkevins:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1629

### Description

This PR is the Android implementation of a new processor which can handle media upload completion events while the Gutenberg editor is closed.

This uses a "hybrid approach", with both RegEx, and the jsoup html parser to parse and splice the post content and update local media ids and urls for various cases. This addresses some of the limitations of the previous approach.

This approach borrows from the previous approach, and also the approach used for iOS: https://github.com/wordpress-mobile/WordPress-iOS/pull/13032.

#### Possible improvements for future iterations:

The implementation in this PR can be improved (and DRYed up) by using a JSON parser to handle the block header contents in a more general way.

#### Note:

The HTML parser used in this implementation has two modes with slightly different serialization outputs. However, post content I've encountered in my tests conforms strictly to neither:

Content source|self-closing-tags|boolean-attributes
-|-|-
`html` mode|``|`checked`
`xml` mode|``|`checked=""`
Post content|``|`checked`

I don't foresee this being a problem, but it's worth keeping in mind while writing any new tests.

### Tests

Parts of this implementation make extensive use of RegEx patterns, which can be difficult to read. I've added javadocs in many areas to mitigate that. Also, I've added a suite of unit tests to cover many scenarios.

#### Copied from the @SergioEstevao 's iOS PR:

It handles 3 possible scenarios for galleries, that depend of the link to value:
 - none: there is no `a` element is in the gallery and we just have img
 - media:  we have a `a` element and it points to the full url of the img
 - attachment: we have a 'a' element and it should point to the attachment page for media. I noticed that the moment the gallery block is not inserting the `a` element because the `link` field for media is not defined. Fixing this should be done on the main gallery block PR because we will need to start to send the link information for image/media to the JS/React side. This will need to change to the bridge code and in the main apps to retrieve that information.

#### To test:
 - Create a new post using Gutenberg
 - Add a gallery block with some images
 - While the images are still uploading press publish
 - Wait outside the editor for the post to be published
 - Open the post again
 - Check that the gallery block is working
 - Switch to HTML mode and see if the link and id are pointing to the server/remote values.

Also, please test this same flow for image and media-text blocks.

#### Video block

This PR may address some of the issues described here: https://github.com/wordpress-mobile/gutenberg-mobile/issues/1512, but this needs further testing.

PR submission checklist:


pinarol:
I started testing this one.

I come across with the below problem while the editor starts so I tested with latest develop on gutenberg-mobile

screenshot_20200106_190832_org wordpress android beta_

So it would be good to update this PR to show latest develop on gutenberg-mobile.

Other than that, here are some problems I noticed:

- [ ] 1. It is allowing to choose only 1 image at a time from device library. I can't choose multiple.
- [ ] 2. I get this crash in the following scenario:
   - Start to upload an image
   - Close/Reopen post when the upload is still in progress
   - Tap on the image
This crash is seen:

screenshot_20200106_192949_org wordpress android beta_

We can address the second issue on a separate issue if that will complicate things. But it would be good to address the first one since it is preventing to test this PR properly.

pinarol:
In addition to that, I realized a trouble with media text:

- Add a media text block
- Start uploading an image
- Close post, wait until upload is finished
- Open post
Result: Media Text tries to show a video and when I tap "play" the browser opens showing the uploaded image, so media-text mistakens the mediaType as video. The result html:

```
<!-- wp:media-text {"mediaId":1058,"mediaType":"video","isStackedOnMobile":false,"className":"is-stacked-on-mobile"} -->
<div class="wp-block-media-text alignwide is-stacked-on-mobile"><figure class="wp-block-media-text__media"><video controls src="https://pinarspremiumtestsite123443.files.wordpress.com/2020/01/img_20200106_192814-8.jpg"></video></figure><div class="wp-block-media-text__content"><!-- wp:paragraph -->
<p></p>
<!-- /wp:paragraph --></div></div>
<!-- /wp:media-text -->
```

pinarol:
Not urgent but it would be good to add a new mapping to [mapping.json here](https://github.com/wordpress-mobile/test-cases/tree/master/config) for the new processor in this PR. We added it for iOS already and mapped **gutenberg/gallery-upload-processor.md**.

mkevins:
Hi @pinarol :wave: :smiley: , thanks for testing this PR!

> I come across with the below problem while the editor starts so I tested with latest develop on gutenberg-mobile
> 
> So it would be good to update this PR to show latest develop on gutenberg-mobile.

Good idea! I've updated the gutenberg-mobile reference in this PR to bring the fix for this issue: https://github.com/wordpress-mobile/gutenberg-mobile/issues/1703

I was unable to reproduce the other issues you described (testing on Pixel 3a - Android 10). Can you tell me what device / etc. you tested with? If it was an emulator, maybe I can find a way to reproduce the issues.

> * [ ]  1. It is allowing to choose only 1 image at a time from device library. I can't choose multiple.

For this one, I wonder did you try long-pressing the image to initiate multi-select? A single tap will select only that one image, even if multi-select is enabled.

> * [ ]  2. I get this crash in the following scenario:
>   
>   * Start to upload an image
>   * Close/Reopen post when the upload is still in progress
>   * Tap on the image
>     This crash is seen:

I remember encountering that red screen at one point while developing, but was unable to reproduce it. I tried the steps you listed, and that didn't reproduce it either. I wonder if it can be reproduced after clearing the gradle cache and doing a `yarn clean:install`, and rebuilding?

I was also unable to reproduce this one:

> In addition to that, I realized a trouble with media text:
> 
> * Add a media text block
> * Start uploading an image
> * Close post, wait until upload is finished
> * Open post
>   Result: Media Text tries to show a video and when I tap "play" the browser opens showing the uploaded image, so media-text mistakens the mediaType as video.

That is definitely strange behavior. Can you tell me which method was used to upload the image? Selecting from device, I was unable to reproduce this, but perhaps this only happens in a specific case somehow?



> Not urgent but it would be good to add a new mapping to [mapping.json here](https://github.com/wordpress-mobile/test-cases/tree/master/config) for the new processor in this PR. We added it for iOS already and mapped **gutenberg/gallery-upload-processor.md**.

Will do, thanks for the suggestion, and for already handling it for iOS!

pinarol:
I am testing on real device Huawei P20 lite with Android 9

> For this one, I wonder did you try long-pressing the image to initiate multi-select? A single tap will select only that one image, even if multi-select is enabled.

I didn't know I needed to long press, is that a common thing on Android? I am able to multi select now.

----------

> I remember encountering that red screen at one point while developing, but was unable to reproduce it. I tried the steps you listed, and that didn't reproduce it either. I wonder if it can be reproduced after clearing the gradle cache and doing a yarn clean:install, and rebuilding?

I tried `yarn clean:install` and `yarn clean:start` but I can still repro the issue, maybe this screen capture would help: https://drive.google.com/file/d/1tYVSqW_1AGWG34rflWuEfYXwsfRcR7Bd/view?usp=sharing 
Maybe you should try with a rather big file to have a long enough time to close/reopen the post while the upload is still in progress?
I also invited you into my test site so that you can try with the same site.
---------
About the media-text issue, I can repro that on develop either. So I opened a [new issue](https://github.com/wordpress-mobile/gutenberg-mobile/issues/1753). 

----------

I noticed another weird behaviour about gallery, and this looks

- [ ] Gallery becomes invalid if I change column count before closing the post

Steps:
- add a gallery block
- start media upload with multiple images 
- change column count
- close post
- wait until uploads complete
- open post (result post I used: https://wordpress.com/block-editor/post/pinarspremiumtestsite123443.wordpress.com/1105)

This validation error occurs:

screenshot_20200107_114725_org559759793227872246

pinarol:
I understood what's going on with media-text and opened a [new PR ](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1756)to fix that, added you as reviewer @mkevins

pinarol:
I tried again but I couldn't repro ` - [ ] Gallery becomes invalid if I change column count before closing the post.` I don't know what is going on. 

I tested below cases and they all worked well:

- [x] Gallery block - Close post with an ongoing image upload - [steps](https://github.com/wordpress-mobile/test-cases/blob/master/test-cases/gutenberg/gallery.md#tc002)
- [x] Video block - Close post with an ongoing video upload - [steps](https://github.com/wordpress-mobile/test-cases/blob/master/test-cases/gutenberg/video.md#tc005)	
- [x] Media Text block - Close post with an ongoing video upload - [steps](https://github.com/wordpress-mobile/test-cases/blob/master/test-cases/gutenberg/media-text.md#tc004-v)
- [x] Image block - Close post with an ongoing image upload - [steps](https://github.com/wordpress-mobile/test-cases/blob/master/test-cases/gutenberg/image.md#tc005)
- [x] Media Text block - Close post with an ongoing image upload - [steps](https://github.com/wordpress-mobile/test-cases/blob/master/test-cases/gutenberg/media-text.md#tc004-i)
- [x] Media Text block - Change media text order during upload and close the post - [steps](https://github.com/wordpress-mobile/test-cases/blob/master/test-cases/gutenberg/media-text.md#tc005-i)

(Copied the test items from test suites [here](https://github.com/wordpress-mobile/test-cases/tree/master/test-suites/gutenberg))


This PR is fixing [this issue](https://github.com/wordpress-mobile/gutenberg-mobile/issues/1512) as well.

The only problem left is the "java.lang.String cannot be cast to java.lang.Double" crash. I am OK with doing that on another PR but the root cause needs to be figured before shipping Gallery upload feature to live. 

I quickly went over the code and it looked OK to me but I'll leave code review part to @marecar3 since this is Android.

Thanks for all the effort @mkevins this looks good to me.

WordPress/src/main/java/org/wordpress/android/ui/posts/MediaUploadCompletionProcessor.java
```diff
@@ -0,0 +1,371 @@
...
+import org.wordpress.android.util.helpers.MediaFile;
+
+import java.util.ArrayList;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static org.wordpress.android.ui.posts.MediaUploadCompletionProcessorPatterns.PATTERN_BLOCK;
+import static org.wordpress.android.ui.posts.MediaUploadCompletionProcessorPatterns.PATTERN_GALLERY_LINK_TO;
+
+public class MediaUploadCompletionProcessor {
```
pinarol: On the second look, I think this file is too big and it'll get bigger in time as we add new blocks that require the same kind of processing. Can we maybe extract out processors that belong to different blocks? 

What I mean is instead of having: processImageBlock, processVideoBlock, processMediaTextBlock, processGalleryBlock all here we could do sth like:

```
//in a different file
interface BlockProcessor {
  process(String block)
}

//in a different file
class ImageBlockProcessor implements BlockProcessor {
    process(String block) {
       ....
    }
}

//in a different file
class VideoBlockProcessor implements BlockProcessor {
    process(String block) {
       ....
    }
}
```

Same with MediaTextBlockBlockProcessor, GalleryBlock etc.

We can even go one step further and define a BlockProcessorFactory that will return the needed processor based on the type, like:

```
class BlockProcessorFactory {

 private ImageBlockProcessor imageProcessor;
 private VideoBlockProcessor imageProcessor
 ...
 public BlockProcessor getProcessor(BlockType type) {
    switch blockType {
      case IMAGE:
         return imageProcessor;
     case VIDEO:
        return videoProcessor;
      ....
    }
 }
}

```
and then processBlock becomes:

    private String processBlock(String block) {
        BlockType type = Helpers.detectBlockType(block);
        blockProcessorFactory.getProcessor(type).process(block);
     }

WordPress/src/test/java/org/wordpress/android/ui/posts/PostUtilsUploadProcessingTest.kt
```diff
@@ -0,0 +1,98 @@
...
+
+@RunWith(MockitoJUnitRunner::class)
+class PostUtilsUploadProcessingTest {
+    private val siteUrl = "https://wordpress.org"
+    private val remoteImageUrl = "https://wordpress.org/gutenberg/files/2018/07/Screenshot-1-1.png"
+    private val remoteVideoUrl = "https://videos.files.wordpress.com/qeJFeNa2/macintosh-plus-floral-shoppe-02-e383aae382b5e38395e383a9e383b3e382af420-e78fbee4bba3e381aee382b3e383b3e38394e383a5e383bc-1_hd.mp4"
+    private val localMediaId = "112"
+    private val remoteMediaId = "97629"
+    private val attachmentPageUrl = "https://wordpress.org?p=97629"
+    private val oldImageBlock = """<!-- wp:image {"id":112,"align":"full"} -->
```
pinarol: Could you also add test cases where id is not the first attribute. Because there's no guarantee that id will be the first attribute. For example in media-text block, when we place the media at right and the text on left id becomes the second attribute. So I think we should add this case for all blocks.

WordPress/src/main/java/org/wordpress/android/ui/posts/MediaUploadCompletionProcessorPatterns.java
```diff
@@ -0,0 +1,214 @@
...
+         * @return A {@link Pattern} with the following capture groups:
+         *
+         * <ol>
+         *     <li>Block header before id</li>
+         *     <li>localId</li>
+         *     <li>Block header after id</li>
+         *     <li>Block contents</li>
+         *     <li>Block closing comment and any following characters</li>
+         * </ol>
+         */
```
pinarol: independent of the previous comment, there's some duplication here, both on comments and function implementations. Can we reduce this?

first thing that pops into my head: Instead of different getXXXBlockPattern functions maybe we can do this:

```

 public static Pattern getPattern(String localId, MediaBlockType type) {
   return Pattern.compile(String.format(getTemplate(type), localId), Pattern.DOTALL);
 }

String getTemplate(MediaBlockType type) {
  // switch type...
}

```

WordPress/src/main/java/org/wordpress/android/ui/posts/MediaUploadCompletionProcessorPatterns.java
```diff
@@ -0,0 +1,214 @@
+package org.wordpress.android.ui.posts;
+
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Arrays;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
```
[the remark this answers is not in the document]
pinarol: If this sounds like an overkill at this point we can consider separating Helpers to a separate file I think.

marecar3:
Hey, @mkevins great job here.
I have tests all scenarios from issue:  https://github.com/wordpress-mobile/gutenberg-mobile/issues/1512
and it seems that this PR has fixed both of them 🎉

WordPress/src/main/java/org/wordpress/android/ui/posts/MediaUploadCompletionProcessor.java
mkevins: Good suggestion! I've updated the PR after extracting the different processors, and implementing the factory pattern.

WordPress/src/test/java/org/wordpress/android/ui/posts/PostUtilsUploadProcessingTest.kt
mkevins: Indeed! I've added test cases for all blocks. It turns out, image was the only one failing this assertion, because the regex had a typo (now fixed). :smiley:

WordPress/src/main/java/org/wordpress/android/ui/posts/MediaUploadCompletionProcessorPatterns.java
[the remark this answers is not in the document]
mkevins: I've extracted these as well, placing them closer to the implementations where they are used and eliminated the inner Helpers class.

WordPress/src/main/java/org/wordpress/android/ui/posts/MediaUploadCompletionProcessorPatterns.java
mkevins: > there's some duplication here, both on comments and function implementations. Can we reduce this?

Absolutely! I've refactored this to DRY up the pattern usage, moving most of the logic to the parent (abstract) `BlockProcessor` class.

pinarol:
What happened here, wpmobilebot has been very productive :) . Is there any way to detect these errors in local? or maybe autocorrect?

mkevins:
> What happened here, wpmobilebot has been very productive :) . Is there any way to detect these errors in local? or maybe autocorrect?

Indeed :joy: I was trying to reproduce the checkstyle errors locally with `./gradlew lint` instead of `./gradlew ktlint` :man_facepalming: (not sure why Android Studio didn't show the errors). Anyway, got the proper ktlint-disable comments now, so this is ready for another pass.

WordPress/src/test/java/org/wordpress/android/ui/posts/PostUtilsUploadProcessingTest.kt
pinarol: wow, so it did help to add some new test cases. I am super happy we have unittests!

mkevins: Me too! Very helpful to have them, especially with regex changes.

mkevins:
Thanks for reviewing and testing again Pinar :smiley: . I'll open an issue for the crash mentioned, and investigate to find a solution.

mkevins:
Also, thanks @marecar3 for reviewing and testing this one. :+1:

PR #11001, opened 2019-12-20
React native/pass http error code

mchowning:
Updates to version of FluxC that returns full error object for failed requests. Takes that full response and extracts the http error code from it and passes that along to mobile gutenberg.

⚠️  Before merging this PR, the [associated FluxC PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1459) should be merged and the FluxC version reference here should be updated.

### Why We're Using a `Bundle` To Pass Error Information

This error information has to be passed from the `WordPress`module to the `WordPressEditor` module and then to the `react-native-gutenberg-bridge` module. In other words, `WordPress` depends on `WordPressEditor`, which depends on `react-native-gutenberg-bridge`. `WordpressEditor`, however, only has `react-native-gutenberg-bridge` as an ["implementation" dependency](https://github.com/wordpress-mobile/WordPress-Android/blob/fd1e8475e6898d906c6ae8887def97c11b821783/libs/editor/WordPressEditor/build.gradle#L82) (I'm guessing this was done for the very good reason of wanting to keep the core app clean of any React Native dependencies). As a result, `WordPress` does not know anything about `react-native-gutenberg-bridge` and `react-native-gutenberg-bridge` knows nothing about either `WordPress` or `WordPressEditor`. 

This means that if we made a class to hold the error data, no matter where we put it, one of either `WordPress` or `react-native-gutenberg-bridge` would not be able to use it. Instead, we would be forced to have one class in `WordPressEditor` for passing the error data from `WordPress` to `WordPressEditor`. Then, inside `WordPressEditor`, we would have to convert that first data class to a new data class from `react-native-gutenberg-bridge`, so that the information could be passed to that module (because `react-native-gutenberg-bridge` would not know about the class from `WordPressEditor`). 

Not only would we need these two, essentially identical, data classes, however. We would also need to convert the data class from `react-native-gutenberg-bridge` into a kind of object that can be sent over the javascript bridge to React Native (`WritableMap`). Fortunately, React Native has a helper method to automatically convert a `Bundle` into a `WritiableMap`. 

FWIW, one other option would be to use a `Map<String, Object>` instead of a `Bundle`. React Native also has a helper to convert that kind of object to a `WritableMap`. I did not go with that approach because that has even less type-safety than a `Bundle`.

If you think there's a better way to handle this, please let me know. It just seemed that using a well-typed data class with these errors would require a substantial amount of boilerplate for little, if any, gain since that data would then just have to be convered into a `WritableMap`, which is essentially the same thing as a `Bundle`.

### To Test

See test steps in https://github.com/wordpress-mobile/gutenberg-mobile/pull/1686

### PR submission checklist


WordPress/src/main/java/org/wordpress/android/ui/posts/reactnative/ReactNativeRequestHandler.kt
```diff
@@ -73,11 +76,35 @@ class ReactNativeRequestHandler @Inject constructor(
...
+            is Error -> {
+                val bundle = Bundle().apply {
+                    response.error.volleyError?.networkResponse?.statusCode?.let {
+                        putInt("code", it)
+                    }
+                    extractErrorMessage(response.error)?.let {
+                        putString("message", it)
+                    }
+                }
+                onError(bundle)
```
planarvoid: I don't know if it's possible but what about sending a `Pair` instead? It seems less error prone than a bundle 🤷‍♂

mchowning: `Pair` is actually what I used for my first implementation of this. 😀

I decided to use a `Bundle` instead because I had to convert the `Pair` to a `Bundle`-like React Native object (`WritableMap`) that can be sent across the javascript bridge to React Native (like with saving instance state for `Activities` you can only send a few special kinds of objects across the javascript bridge). That conversion involves exactly the same kind of `putString(...)` and `putInt(...)` calls that I am doing here, so it didn't seem like `Pair` was gaining me anything, it was just moving the place where I did the "risky" `putX(...)` calls. Using a `Bundle` directly seemed to make since on account of the fact that React Native has a helper method for converting a `Bundle` to a `WritableMap` automatically. 

There is a helper for converting `Map<String, Object` to `WritableMap` as well, but that actually seemed less safe since we would have no type safety to insure that we only put objects into the map that could be sent across the bridge.

Using `Bundle` directly also seemed like a better longterm approach since if/when (most likely when) we end up wanting to pass a third piece of information from the error, I think we would need to convert this back from a `Pair` to a `Bundle` or start nesting `Pair<Pair<>>`. 😱 (

An alternative approach would be to change `WordPress-Editor` to use `react-native-gutenberg-bridge` as an `api` dependency instead of an `implementation` dependency. That would allow us to define a well-typed object for the relevant error information in `react-native-gutenberg-bridge` and use that object here in `WordPress-Android` (transitively, via the Editor module). I feel like this might actually end up being the best longterm solution, but it adds React Native dependencies to the WordPress app itself (as opposed to just being implementation dependencies of the Editor module) and I don't feel comfortable that I've thought through all the implications of doing that yet. wdyt?

PR #11014, opened 2019-12-23
Post-Signup Interstitial: Track interstitial screen events

### LINKED ISSUE #10985: Post-Signup Interstitial: Track interstitial screen events
This is the third task of the Post-Signup Interstitial feature. The goal here is to track all user behavior related to the interstitial screen.

## Tasks
- [x] Track when interstitial is seen 
- [x] Track when the **Create New Site** button is clicked
- [x] Track when the **Add a Self-Hosted Site** button is clicked
- [x] Track when the **Not Right Now** button is clicked / the screen is dismissed

## Notes
- Master Issue: #10918

renanferrari:
Fixes #10985

For this PR, the following changes were made:

- Four new tracking events were created:
  - `POST_SIGNUP_INTERSTITIAL_SHOWN` 
    - Emmited when the `PostSignupInterstitialActivity` is created.
  - `POST_SIGNUP_INTERSTITIAL_CREATE_NEW_SITE_TAPPED`
    - Emmited when the user taps the **Create WordPress.com Site** button.
  - `POST_SIGNUP_INTERSTITIAL_ADD_SELF_HOSTED_SITE_TAPPED`
    - Emmited when the user taps the **Add Self-Hosted Site** button.
  - `POST_SIGNUP_INTERSTITIAL_DISMISSED`
    - Emmited when the user taps either the **Not Right Now** button or the back button.

- For consistency, the interstitial's back button now has the same behavior as the **Not Right Now** button, meaning the screen will be closed and the user will be redirected to the _Reader_ tab.

- A `PostSignupInterstitialViewModel` has been created alongside some local unit tests.
  - Note: the interstitial screen is fairly simple and a view model isn't really necessary for it to work. The only reason it is there is because it does make local unit tests much easier to write and read. That being said, I thought it would be worth giving up a bit of the implementation's simplicity to achieve simplicity in testing, but I could totally be convinced otherwise. 

## To test

For each of the tests below, you first need to:
1. Clear app data.
2. Start `logcat`.
3. Complete the login flow with an account _that doesn't have any sites_.
4. Confirm that the post-signup interstitial screen is shown.

#### 1. Track interstitial shown
1. On `logcat`, confirm that the `post_signup_interstitial_shown` event was tracked.

#### 2. Track Create New Site button tapped
1. On the post-signup interstitial screen, tap **Create WordPress.com Site**.
2. On `logcat`, confirm that the `post_signup_interstitial_button_tapped` event was tracked with the following properties: `{"button":"create_new_site"}`.

#### 3. Track Add Self-Hosted Site button tapped
1. On the post-signup interstitial screen, tap **Add Self-Hosted Site**.
2. On `logcat`, confirm that the `post_signup_interstitial_button_tapped` event was tracked with the following properties: `{"button":"add_self_hosted_site"}`.

#### 4. Track interstitial dismissed
1. On the post-signup interstitial screen, tap either the **Not Right Now** button or the back button.
2. On `logcat`, confirm that the `post_signup_interstitial_dismissed` event was tracked.

---

PR submission checklist:


libs/analytics/WordPressAnalytics/src/main/java/org/wordpress/android/analytics/AnalyticsTrackerNosara.java
```diff
@@ -1698,6 +1704,13 @@ public static String getEventNameForStat(AnalyticsTracker.Stat stat) {
                 return "unpublished_revision_dialog_load_local_version_clicked";
             case UNPUBLISHED_REVISION_DIALOG_LOAD_UNPUBLISHED_VERSION_CLICKED:
                 return "unpublished_revision_dialog_load_unpublished_version_clicked";
+            case POST_SIGNUP_INTERSTITIAL_SHOWN:
+                return "post_signup_interstitial_shown";
```
maxme: Can we use a more generic name since it's used after signup and signin? I'm afraid this can be confusing for someone looking only at the analytics.

renanferrari: I actually think the whole "post-signup interstitial" name could be confusing for anyone looking at this feature for the first time in the future. That being said, I agree it could be even _more_ confusing for someone looking only at analytics events. 

Got any suggestions for a new name? How about just `interstitial_shown`? Or should we have a more descriptive name like `welcome_interstitial_shown` or `no_sites_interstitial_shown`?

renanferrari: We have decided to rename the events prefix from `post_signup_interstitial` to `welcome_no_sites_interstitial`.

maxme: Just a short note: we picked `welcome_no_sites_interstitial` over `welcome_interstitial` because we thought we could add other welcome interstitial screens for other use cases in the future.

PR #11023, opened 2019-12-24
Issue/10635 material comments

### LINKED ISSUE #10635: Implement Material theme on Comments List, Details and Edit screen
We should switch widgets used at Comments List, Details and Edit screens to their material counterparts, remove color references where possible, and switch to using material `textAppearance` attributes.

khaykov:
Fixes #10635

This PR moves Comments List, Detail and Editor screens to Material Theme and adds Dark Theme support. There are some code changes for cosmetic reasons.

Design changes are reviewed before the code review is requested.

@mattmiklic Color of PENDING status was different in List and Details, so I changed one used in the list back to yellow.

comparison_comments

To test:
Navigate to Comments management screen and confirm it's visual and functional integrity in both Light and Dark themes. The theme can be changed from App Settings.

PR submission checklist:


mattmiklic:
Looking good here. Two things I'm thinking about:

- The Spam/Trash/Edit icons and text on the comment detail screen are looking a little bit too high-contrast. Could we make those use the "medium emphasis" variant of the On Surface color?
- Could we remove the horizontal line on the Edit Comment screen? It doesn't seem necessary since there's no equivalent on the "Reply to post" comment screen.

khaykov:
@mattmiklic Sure, I updated the icons and label color. I removed the "medium emphasis" when secondary color is used (for like and approve actions) since they look really washed out.

comparison_comments

The bottom line on Edit Comment magically disappeared in recent version :)

jkmassel:

WordPress/src/main/java/org/wordpress/android/ui/CollapseFullScreenDialogFragment.java
```diff
@@ -257,15 +263,16 @@ private void initBuilderArguments() {
...
         toolbar.setTitle(mTitle);
         toolbar.setNavigationContentDescription(R.string.description_collapse);
         toolbar.setNavigationIcon(ContextCompat.getDrawable(view.getContext(), R.drawable.ic_chevron_down_white_24dp));
-        toolbar.setNavigationOnClickListener(new View.OnClickListener() {
-            @Override
-            public void onClick(View view) {
-                onCollapseClicked();
-            }
-        });
+        toolbar.setNavigationOnClickListener(view1 -> onCollapseClicked());
```
develric: I would consider to not use the numbering in the variable name (something like a simple `v` maybe?). Same in `CommentDetailFragment`. (I think the `view1` naming is the one created by AS to avoid the clash with the function parameter but I would consider to modify it as well, let me know wdyt 😊)

WordPress/src/main/java/org/wordpress/android/ui/comments/EditCommentActivity.java
```diff
@@ -309,22 +309,15 @@ private void cancelEditCommentConfirmation() {
...
-                });
+                (dialog12, whichButton) -> finish());
         dialogBuilder.setNegativeButton(
                 getResources().getText(R.string.no),
-                new DialogInterface.OnClickListener() {
-                    public void onClick(DialogInterface dialog, int whichButton) {
-                        // just close the dialog
-                    }
+                (dialog1, whichButton) -> {
+                    // just close the dialog
```
develric: Same as in a previous comment, maybe I would use `dialog` (instead of `dialog1`)

WordPress/src/main/res/values/dimens.xml
```diff
@@ -434,7 +434,6 @@
     <dimen name="domain_suggestions_intro_horizontal_padding">16dp</dimen>
     <dimen name="domain_suggestions_intro_header_bottom_padding">8dp</dimen>
 
-    <dimen name="keyline_first">16dp</dimen>
     <dimen name="keyline_second">72dp</dimen>
```
develric: Not directly related to this PR, so no action required, but just to share this thought 😊; it sounded strange that we are not using anywhere the `keyline_first`. Probably we ended up using other `16dp` constant instead of this explicit one.

WordPress/src/main/res/layout/comment_edit_activity.xml
```diff
@@ -1,48 +1,68 @@
...
             android:orientation="vertical">
 
-            <EditText
-                android:id="@+id/edit_comment_content"
+            <LinearLayout
+                android:id="@+id/edit_comment_container"
                 android:layout_width="match_parent"
-                android:layout_height="match_parent"
-                android:gravity="top"
-                android:hint="@string/hint_comment_content"
```
develric: I'm pretty sure it is wanted but I prefer to cross check with you. I saw from one of your comment that we wanted to remove the horizontal line, but it seems we are also losing the hint, like in the following image:

image

WordPress/src/main/java/org/wordpress/android/ui/CollapseFullScreenDialogFragment.java
khaykov: Good catch! Yes, AS creates those sometimes, even though it's inside a function. Weird.

WordPress/src/main/res/layout/comment_edit_activity.xml
khaykov: Hints in `TextInputLayout` look a bit different - there is a label that goes to the top when you start typing and it doesn't really match the look. Considering that you will always open edit comment screen with some sort of comment (and there is an error if a comment is empty) I decided that we can get rid of it.

khaykov:
Thanks for taking a look @develric ! I fixed the variable names :)

WordPress/src/main/res/layout/comment_edit_activity.xml
develric: Thanks for the explanation Klym 😊!

PR #11027, opened 2019-12-30
Accessibility : Aztec UI Improvements 

### LINKED ISSUE #10906: Accessibility : Aztec Editor UI Improvements
### Accessibility Scanner audit

Each item corresponds to an outlined area on the attached screenshot below.

* Text contrast
org.wordpress.android:id/title
The item's text contrast ratio is 2.56. This ratio is based on an estimated foreground color of #87A6BC and an estimated background color of #FFFFFF. Consider increasing this item's text contrast ratio to 3.00 or greater.

* Text contrast
org.wordpress.android:id/aztec
The item's text contrast ratio is 2.56. This ratio is based on an estimated foreground color of #87A6BC and an estimated background color of #FFFFFF. Consider increasing this item's text contrast ratio to 3.00 or greater.

screenshot_WordPress_2019-12-02-22_23_22

jd-alexander:
Fixes #10906

## Findings

Title and Aztec content views weren't meeting the contrast guidelines for accessibility. i.e the contrast ratio derived from the foreground and background wasn't high enough to pass the mark that makes the control's color accessible. The only point to note here, is that this change makes it obvious that the hint colors in Gutenberg are different from Aztec when doing the switch.

## Solution
Changed the hint color of the views to one that makes the contrast ratio higher. 

## Testing
1. New post. 
2. Switch to the classic editor. 
3. You will see the controls as shown below. 

Before | After 
--------|-------
        | 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/libs/editor/WordPressEditor/src/main/res/values/colors.xml
+++ b/libs/editor/WordPressEditor/src/main/res/values/colors.xml
@@ -19,7 +19,7 @@
     <color name="code_background">@color/wp_grey_light</color>
     <color name="hint_text">@color/wp_grey_darken_10</color>
     <color name="link">@color/wp_blue_medium</color>
-    <color name="quote">@color/wp_grey_darken_10</color>
+    <color name="quote">@color/wp_grey</color>
     <color name="quote_background">@color/wp_grey_lighten_30</color>
     <color name="text">@color/wp_grey_darken_30</color>
 
--- a/libs/editor/WordPressEditor/src/main/res/values/colors.xml
+++ b/libs/editor/WordPressEditor/src/main/res/values/colors.xml
@@ -17,9 +17,9 @@
     <color name="bullet">@color/wp_grey_darken_30</color>
     <color name="code">@color/text</color>
     <color name="code_background">@color/wp_grey_light</color>
-    <color name="hint_text">@color/wp_grey</color>
+    <color name="hint_text">@color/wp_grey_darken_10</color>
     <color name="link">@color/wp_blue_medium</color>
-    <color name="quote">@color/wp_grey</color>
+    <color name="quote">@color/wp_grey_darken_10</color>
     <color name="quote_background">@color/wp_grey_lighten_30</color>
     <color name="text">@color/wp_grey_darken_30</color>
 


iamthomasbishop:
@jd-alexander Thanks for working on this! Fwiw, we will be implementing a new color system in mobile GB soon, so all placeholder text should pass with _at least_ 3.1:1 contrast ratio, (can’t remember the exact ratio). With that said, I am all for improving it for Aztec as well so thanks for tackling this! I have a couple of thoughts:

* What is the contrast ratio of the new color against the white canvas background?
* Can we make sure the placeholder text color in dark mode also passes? // cc @khaykov 

Can we also create a ticket for this on Aztec for WPiOS and make sure our text is passing contrast tests similarly there? // cc @frosty

jd-alexander:
Hi @iamthomasbishop thanks for the review!

> @jd-alexander Thanks for working on this! Fwiw, we will be implementing a new color system in mobile GB soon, so all placeholder text should pass with at least 3.1:1 contrast ratio, (can’t remember the exact ratio)

Great! So does this mean I shouldn't do any other color changes like this? These are already low priority so I don't mind :) 

> What is the contrast ratio of the new color against the white canvas background?
Screen Shot 2019-12-29 at 8 54 21 PM

> Can we make sure the placeholder text color in dark mode also passes? // cc @khaykov

This sounds like a good idea. 

> Can we also create a ticket for this on Aztec for WPiOS and make sure our text is passing contrast tests similarly there? // cc @frosty

Since Aztec, is going to be retired I am only doing these changes since they seemed simple and could have an immediate impact but if the changes have a larger impact across platforms then it could be saved for Gutenberg. Let me know.

iamthomasbishop:
> does this mean I shouldn't do any other color changes like this? These are already low priority so I don't mind :)

Not at all! If you find things in Aztec, feel free! I would just leave the Gutenberg side as-is, considering those will be changing with  the new color system anyhow.

> Since Aztec, is going to be retired I am only doing these changes since they seemed simple and could have an immediate impact but if the changes have a larger impact across platforms then it could be saved for Gutenberg. Let me know.

To be clear, I don’t think Aztec will be fully retired for a while – it will soon become the _secondary_ editor to Gutenberg which becomes the default. So these changes are still useful 😀

libs/editor/WordPressEditor/src/main/res/values/colors.xml
```diff
@@ -17,9 +17,9 @@
     <color name="bullet">@color/wp_grey_darken_30</color>
     <color name="code">@color/text</color>
     <color name="code_background">@color/wp_grey_light</color>
-    <color name="hint_text">@color/wp_grey</color>
+    <color name="hint_text">@color/wp_grey_darken_10</color>
     <color name="link">@color/wp_blue_medium</color>
-    <color name="quote">@color/wp_grey</color>
+    <color name="quote">@color/wp_grey_darken_10</color>
```
theck13:
```suggestion
    <color name="quote">@color/wp_grey</color>
```

We don't need to change the value for the `quote` color in order to update the hint text.  The `quote` color is used for the vertical line to the left of a quote.  See the screenshots below for illustration.

jd-alexander: Thanks! I reviewed the diff in GitKraken and didn't see this quote change. Must have slipped in somehow. Thanks for the review!

PR #11029, opened 2019-12-31
Fix #10747: Gutenberg rollout phase 2

### LINKED ISSUE #10747: Gutenberg progressive rollout strategy
Implement a progressive rollout strategy to opt more users in the Gutenberg editor.

Phase 1: set the block editor as the default editor for new posts. We'll only target people who didn't already switched back from the block editor to Aztec

Phase 2: set the block editor as the default editor for users who already switched back to Aztec from a previous version of the app.

We'll rollout phase 1 and phase 2 progressively on randomly selected users. We'll target wpcom users first:

- [x] phase 1 - 10% - release 13.7
- [x] phase 1 - 30% - release 13.8
- [x] phase 1 - 100% - release 13.9
- [x] phase 2 - 30% - release 14.0
- [ ] phase 2 - 100% - release 14.1

maxme:
Fixes #10747 

<s>This FluxC PR must be merged first: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1466</s> ✅ FluxC version was bumped to 1.5.6 in 7c586ea 

This PR adds:
- A new rollout phase. This target users who have aztec enabled before the first rollout phase.
- A new popup that shows only the first time the user open Gutenberg after the site has been switched from aztec to gutenberg. Popup text is: `We made big improvements to the block editor and think it\'s worth a try! We enabled it for new posts and pages but if you'd like to change to the classic editor, go to 'My Site' > 'Site Settings'`

Note: there is one downside to this, if the user wasn't part of the first rollout (it means they switched manually to Gutenberg a while ago) and they switch a site to Aztec, the next time they re-open the app this site will be switched back to Gutenberg (this will happen only once, they'll be considered in the rollout group after that).

To test:
- Set `GB_ROLLOUT_PERCENTAGE_PHASE_2` to 100% and build.
- Clear App DB, run the app, go to site settings switch a site to Aztec (this should ensure that you're not part of the first rollout).
- Close the app / re-open the app. 
- Create a new post, Gutenberg should open with the new popup.


hypest:
Removed the PR from the project board since we typically only track tickets on the board.

WordPress/src/main/java/org/wordpress/android/ui/prefs/AppPrefs.java
```diff
@@ -649,13 +650,13 @@ public static boolean shouldShowGutenbergInfoPopupForTheNewPosts(String siteURL)
         return flag;
     }
 
-    public static void setShowGutenbergInfoPopupForTheNewPosts(String siteURL, boolean show) {
+    public static void setBooleanPrefForKeyAndURL(PrefKey key, String siteURL, boolean show) {
```
hypest: Kinda nit but, I think that the method name has gotten a bit too generalized, to the point that is probably not making too much sense. Can we keep the previous name or is there some reason to move away from that one?

WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
```diff
@@ -45,16 +46,47 @@ public static void migrateAppWideMobileEditorPreferenceToRemote(final AccountSto
...
-        // ^ This code should be here.
-
         // If the user is already in the rollout group, we can skip this the migration.
         if (AppPrefs.isUserInGutenbergRolloutGroup()) {
             return;
         }
 
+        // -------- Rollout: Phase 2 -------
+        //
+        // If the user as one Aztec enabled site, we'll migrate all his sites to gutenberg and show
```
hypest: Nit: typo: s/as/has

WordPress/src/main/java/org/wordpress/android/ui/prefs/AppPrefs.java
maxme: Updated in 4f4cdb2278e547207c78f2105bf0c4ce0d9b3e10

WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
maxme: Updated in 4f4cdb2278e547207c78f2105bf0c4ce0d9b3e10

maxme:
Ready for a second round. When this is merged, #11062 will need a review.

WordPress/src/main/java/org/wordpress/android/ui/prefs/AppPrefs.java
```diff
@@ -624,14 +625,14 @@ public static void removeAppWideEditorPreference() {
         remove(DeletablePrefKey.GUTENBERG_DEFAULT_FOR_NEW_POSTS);
     }
 
-    public static boolean shouldShowGutenbergInfoPopupForTheNewPosts(String siteURL) {
+    public static boolean getShowGutenbergInfoPopupForTheNewPosts(PrefKey key, String siteURL) {
```
hypest: I wonder, does the `setShowGutenbergInfoPopupForTheNewPosts()` call in line 646 below need to adopt a "PrefKey" param as well? I'm a bit confused by the codepaths but, it looks to me that when the flag is true, we'll try to clear the Phase 1 flag only.

maxme: Updated in 47d5282d7f

WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
```diff
@@ -45,16 +46,47 @@ public static void migrateAppWideMobileEditorPreferenceToRemote(final AccountSto
...
+                for (SiteModel site : siteStore.getSites()) {
+                    // Show "phase 2" dialog on sites that get switched from aztec to gutenberg
+                    if (TextUtils.equals(site.getMobileEditor(), AZTEC_EDITOR_NAME)) {
+                        AppPrefs.setShowGutenbergInfoPopupPhase2ForNewPosts(site.getUrl(), true);
+                        // Will show the popup again, even if it was displayed in the past
+                        AppPrefs.setGutenbergInfoPopupDisplayed(site.getUrl(), false);
+                    }
+
+                    // Show "phase 1" dialog on sites that get switched from "empty" (no pref) to gutenberg
+                    if (TextUtils.isEmpty(site.getMobileEditor())) {
```
hypest: What do you think about adding a `continue` statement to finish the loop iteration in the earlier conditional, or alternatively, merge the conditionals with an `else if`?

maxme: Yes they're mutually exclusive and a `else if` would work, but I preferred to keep it like this for readability. The `isEmpty` check won't have any performance impact.

hypest: I'd recommend keeping them in separate blocks for readability :nod:, that's why I mentioned using a `continue` statement first. The end goal is to make the intention very clear, and probably even protect from future code modifications that could involuntarily change the execution flow. Not a blocker if you still prefer keeping the code as is though.

hypest: Addressed with 4204171f2d2b1208e2f770c59621b8bf2c8cab9d.

PR #11035, opened 2020-01-03
Replace recursion with inline logic to fix stack overflow

### LINKED ISSUE #11034: Fix StackOverflow in StatsUtils
Sentry issue fe9959eb176f48e48802afa058313f24

There was a stack overflow exception introduced in a fix for stats utils.

planarvoid:
Fixes #11034 
I can't explain how this stack overflow happens. For some reason `-this` on a long doesn't change the value to be negative. However, I've removed even the option of this overflow by replacing the recursive function with a simpler approach where I update a variable instead. The functionality remains unchanged. I would trust tests to check whether everything works as expected.

To test:
* Smoke tests stats on a site with huge number of values. The numbers should still be correctly shortened/negated.

PR submission checklist:


malinajirka (approved):
Thanks @planarvoid! LGTM

PR #11038, opened 2020-01-06
Add tests for RemoveMediaUseCase

malinajirka:
Adds basic tests for RemoveMediaUseCase.

To test:
1. Create a post
2. Turn on airplane mode
3. Add an image
4. Remove the image
5. Press back and make sure `removeMediaIfNotUploading` is invoked and the media is removed (using a breakpoint)

Note: My AS keeps freezing when I try to stop the app on the breakpoint, I'm not sure why.
It seems the logic is run only with Aztec, is this expected?

PR submission checklist:


malinajirka:
>It seems the logic is run only with Aztec, is this expected?

I've tested this behavior on release 13.7 and it seems we haven't changed this behavior. `EditorFragmentListener.onMediaDeleted` isn't being invoked from the Gutenberg Fragment. I'll submit a new issue.

Issue #11012, opened 2019-12-23
IA create FAB: need some kind of tooltip to promote the FAB.


develric:
As we introduced the Post/Page creation FAB in the My Site screen we want to use some kind of tooltip to promote it's usage and availability.

cc @osullivanchris

PR #11042, opened 2020-01-06
QuickStart review for ia changes

### LINKED ISSUE #11011: Review quick start modifications based on the new IA architecture
We need to review quick start modifications based on the new IA architecture. 

PS: at least the quick start item related to the central bottom navigation button must be moved to FAB.

develric:
Fixes #11011 

With the modifications for the IA project we made some layout changes. In the IA project we removed the add post button in the main navigation bottom bar. As a result of this, the majority of the QuickStart tutorials are still ok, but the `PUBLISH_POST_TUTORIAL` that was targeting the removed bottom button is no more ok as it is.

Our working hypothesis is to align the `PUBLISH_POST_TUTORIAL` to the `CREATE_NEW_PAGE` behavior. That is targeting the `row_blog_posts` in the My Site list of sections (in analogy to the `row_pages` targeted by the `CREATE_NEW_PAGE` tutorial). We are thinking to not move the `PUBLISH_POST_TUTORIAL` to the FAB (and in this case possibly also the `CREATE_NEW_PAGE` should be moved there) also because the new added IA FAB in the My Site page will be announced by an onboarding tooltip woking on the #11012 issue. (cc @osullivanchris to review this rationale)

## Notes
Working on this PR we also noticed the following:
- In the `my_site_fragment.xml` the post row is the only one inside a FrameLayout. Based on the following commit comment

> adf72e53fea9caf5df4fe413c4484d8145bd4ef0 -> Switching to FrameLayout so blinking animation works on API 16

I think it was due to this [lines of code](https://github.com/wordpress-mobile/WordPress-Android/blob/adf72e53fea9caf5df4fe413c4484d8145bd4ef0/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.java#L286) that AFAIU are no more used (so we are removing the FrameLayout)

- AFAIU currently the Publish Post tutorial is usually not triggered since it is marked as completed by the EditPostActivity when doing the Create Page tutorial; we are moving this completion logic for Posts into the `MySiteFragment > viewPosts` function

- With the presence of the FAB, it could be possible that in some screen resolution configuration the focus point is not fully visible (see the image below for the VIew Site step). It seems still good enough, but maybe we could reconsider the horizontal placement of the focus point in this or another PR (cc @osullivanchris )

<p align="center">
  
</p>

## To test
The modifications are still all behind the `INFORMATION_ARCHITECTURE_AVAILABLE` feature flag.

Note that to have access to the QuickStart tutorials you need to create a new site (you can do it from the Switch Site on the My Site page). Once done this when coming back to the My Site page you will be prompted if you want to use the QuickStart tutorial to getting started; Accept and you will have access to the tutorials

| Accept dialog | Next Steps tutorials |
|---|---|
| image | image |

- with `vanilla` flavor go through the tutorials and check that nothing changed for the QuickStart
- with the `zalpha` go through the tutorials and check that nothing changed for the QuickStart except the Publish Post for which check the next step
- for the Publish Post you need to get a similar behavior as for the create page tutorial but targeting the blog posts row in My Site. Also some minor changes in the copy of the snackbars are made to account for the Posts (not Pages) scenario naming. Note that in zalpha you are prompted for the create post tutorial since it is not marked as completed while doing the create page tutorial.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PostsListActivity.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PostsListActivity.kt
@@ -25,10 +25,16 @@ import androidx.viewpager.widget.ViewPager.OnPageChangeListener
 import com.google.android.material.floatingactionbutton.FloatingActionButton
 import com.google.android.material.snackbar.Snackbar
 import com.google.android.material.tabs.TabLayout
+import org.greenrobot.eventbus.EventBus
+import org.greenrobot.eventbus.Subscribe
+import org.greenrobot.eventbus.ThreadMode
+import org.wordpress.android.BuildConfig
 import org.wordpress.android.R
 import org.wordpress.android.WordPress
 import org.wordpress.android.fluxc.Dispatcher
 import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.store.QuickStartStore
+import org.wordpress.android.fluxc.store.QuickStartStore.QuickStartTask
 import org.wordpress.android.fluxc.store.SiteStore
 import org.wordpress.android.ui.ActivityId
 import org.wordpress.android.ui.ActivityLauncher
@@ -39,6 +45,7 @@ import org.wordpress.android.ui.posts.BasicFragmentDialog.BasicDialogOnDismissBy
 import org.wordpress.android.ui.posts.BasicFragmentDialog.BasicDialogPositiveClickInterface
 import org.wordpress.android.ui.posts.PostListType.SEARCH
 import org.wordpress.android.ui.posts.adapters.AuthorSelectionAdapter
+import org.wordpress.android.ui.quickstart.QuickStartEvent
 import org.wordpress.android.ui.uploads.UploadActionUseCase
 import org.wordpress.android.ui.uploads.UploadUtilsWrapper
 import org.wordpress.android.ui.utils.UiHelpers
@@ -46,9 +53,11 @@ import org.wordpress.android.ui.utils.UiString
 import org.wordpress.android.ui.utils.UiString.UiStringRes
 import org.wordpress.android.util.AppLog
 import org.wordpress.android.util.LocaleManager
+import org.wordpress.android.util.QuickStartUtils
 import org.wordpress.android.util.SnackbarSequencer
 import org.wordpress.android.util.SnackbarItem
 import org.wordpress.android.util.redirectContextClickToLongPressListener
+import org.wordpress.android.widgets.WPDialogSnackbar
 import javax.inject.Inject
 
 const val EXTRA_TARGET_POST_LOCAL_ID = "targetPostLocalId"
@@ -68,6 +77,7 @@ class PostsListActivity : AppCompatActivity(),
     @Inject internal lateinit var uploadActionUseCase: UploadActionUseCase
     @Inject internal lateinit var snackbarSequencer: SnackbarSequencer
     @Inject internal lateinit var uploadUtilsWrapper: UploadUtilsWrapper
+    @Inject internal lateinit var quickStartStore: QuickStartStore
 
     private lateinit var site: SiteModel
     private lateinit var viewModel: PostListMainViewModel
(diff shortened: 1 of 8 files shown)


WordPress/src/main/java/org/wordpress/android/ui/posts/PostsListActivity.kt
```diff
@@ -470,4 +493,42 @@ class PostsListActivity : AppCompatActivity(),
     private fun updateMenuTitle(title: UiString, menuItem: MenuItem): MenuItem? {
         return menuItem.setTitle(uiHelpers.getTextOfUiString(this@PostsListActivity, title))
     }
+
+    @Subscribe(sticky = true, threadMode = ThreadMode.MAIN)
+    @SuppressWarnings("unused")
+    fun onEvent(event: QuickStartEvent) {
```
planarvoid: Could we maybe move this whole part (subscription, and all the rest) to the view model? And we could use the SnackbarSequencer to show the snackbar?

WordPress/src/main/java/org/wordpress/android/ui/quickstart/QuickStartMySitePrompts.kt
```diff
@@ -43,10 +44,16 @@ enum class QuickStartMySitePrompts constructor(
     ),
     PUBLISH_POST_TUTORIAL(
             QuickStartTask.PUBLISH_POST,
-            R.id.root_view_main,
-            R.id.bottom_nav_new_post_button,
-            R.string.quick_start_dialog_publish_post_message_short,
-            R.drawable.ic_create_white_24dp
+            if (BuildConfig.INFORMATION_ARCHITECTURE_AVAILABLE) R.id.my_site_scroll_view_root else R.id.root_view_main,
```
planarvoid: It might be little more readable to create a new enum item (maybe rename this one to `OLD_PUBLISH_POST_TUTORIAL`?).

develric: AFAIU we get the `QuickStartMySitePrompts` by `QuickStartTask` but I think we would need to introduce some extra logic (not needed when removing the FF) in managing how we scroll the list of tutorials in `QuickStartUtils.getNextUncompletedQuickStartTask`. So if you agree in this case I would sacrifice the readability a bit (since it's time boxed to when we remove the FF) and not change the logic code. Let me know wdyt 😊.

WordPress/src/main/java/org/wordpress/android/ui/posts/PostsListActivity.kt
develric: Yep I think we could move to the ViewModel. Currently I used the same approach and structure used in the `PagesFragment` for the `CREATE_NEW_PAGE` tutorial so I think if we change here we should change there also but maybe this is out of the scope of this PR? Let me know wdyt, if you think it's worth to modify maybe we can open an issue to come back to it in a dedicated PR I guess (not sure we can have it within this code freeze but since it is not a functional issue at least we trace it).

PS: about if to use the snackbar sequencer, I was just writing `Yep! Good idea!` but then realized that here we have a `WPDialogSnackbar` and not the `WPSnackbar` we use in the sequencer and all the QuickStart is using the `WPDialogSnackbar`. At first sight they seem similar with some difference in styling (?), so I guess we could extend the sequencer to manage them also but again maybe it's a wider scope 🤷‍♂️.

develric:
Hi @planarvoid, thanks for the review 😊! 

I have not made code changes for now but followed up to your comments. Let me know wdyt, thanks 🙇‍♂️

WordPress/src/main/java/org/wordpress/android/ui/posts/PostsListActivity.kt
planarvoid: That makes sense 👍. We should definitely merge the two snackbars and move all the logic to view models but let's do it in a separate PR.

WordPress/src/main/java/org/wordpress/android/ui/quickstart/QuickStartMySitePrompts.kt
planarvoid: Makes sense 👍

PR #11045, opened 2020-01-06
Track gutenberg_enabled event during rollout

maxme:
Part fix for https://github.com/wordpress-mobile/gutenberg-mobile/issues/1718
There is another upcoming PR for `develop` but I'm waiting for https://github.com/wordpress-mobile/WordPress-Android/pull/11029 to be merged before opening it (code is ready here: https://github.com/wordpress-mobile/WordPress-Android/tree/issue/1718-track-gutenberg-rollout-v2)

Known issue: it's tracked when the network request is initiated, not when we get the response.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
+++ b/WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
@@ -81,10 +81,10 @@ public class SiteUtils {
                 }
             }
 
-            // Enable Gutenberg for all sites using a single network call
+            // Enable and track Gutenberg for all sites using a single network call
+            trackGutenbergEnabledForNonGutenbergSites(siteStore);
             dispatcher.dispatch(SiteActionBuilder.newDesignateMobileEditorForAllSitesAction(
                     new DesignateMobileEditorForAllSitesPayload(SiteUtils.GB_EDITOR_NAME)));
-            trackGutenbergEnabledForNonGutenbergSites(siteStore);
 
             // After enabling Gutenberg on these sites, we consider the user entered the rollout group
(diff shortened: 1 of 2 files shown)


WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
```diff
@@ -103,6 +104,15 @@ public static void migrateAppWideMobileEditorPreferenceToRemote(final AccountSto
         }
     }
 
+    private static void trackGutenbergEnabledForNonGutenbergSites(final SiteStore siteStore) {
+        for (SiteModel site : siteStore.getSites()) {
+            if (!TextUtils.equals(site.getMobileEditor(), GB_EDITOR_NAME)) {
```
mchowning: I think this is really just a theoretical concern, but _in theory_ if the dispatch call completed and updated the `site` before we got to this line, would it result in analytics _not_ being sent off because the site's mobile editor would have already been updated?

maxme: Yeah, good point, better to move this before we dispatch the event.

maxme:
@mchowning I moved the analytics tracking code. Can you do a second pass on that one?

mchowning:
Change look good to me @maxme !  I might include something in the comment about why we're taking the obviously-less-than-ideal approach of not only sending analytics without checking whether the network call succeeded, but actually sending off the event _before_ we even try to make the network call. Obviously, someone could check the commit behind the change and get back to this PR though, so I don't think updating the comment is critical.

PR #11046, opened 2020-01-06
Accessibility : Free Photo Library Crash Fix

### LINKED ISSUE #11041: Free Photo library crashes in Debug build
### Expected behavior
The app doesn't crash when I select an image in the Free Photo library.

### Actual behavior
The app crashes as soon as I select an image in the Free Photo Library

1. Create a new post
2. Add a new image
3. Select "Choose from Free Photo Library"
4. Type "cat" into the search box
5. Select an image and notice the app crashes

Emulator, API 28

@jd-alexander Could you please look into this issue as it seems related to the recent accessibility changes. Thanks

jd-alexander:
Fixes #11041

To test:

1. Create a new post
2. Add a new image
3. Select "Choose from Free Photo Library"
4. Type "cat" into the search box
5. Select an image and notice the app doesn't crash.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/stockmedia/StockMediaPickerActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stockmedia/StockMediaPickerActivity.java
@@ -635,21 +635,19 @@ public class StockMediaPickerActivity extends AppCompatActivity implements Searc
         }
 
         private void addImageSelectedToAccessibilityFocusedEvent(ImageView imageView, int position) {
-            if (!ViewCompat.hasAccessibilityDelegate(imageView)) {
-                AccessibilityUtils.addPopulateAccessibilityEventFocusedListener(imageView, event -> {
-                    if (isValidPosition(position)) {
-                        if (isItemSelected(position)) {
-                            final String imageSelectedText = imageView.getContext().getString(
-                                    R.string.photo_picker_image_selected);
-                            if (!imageView.getContentDescription().toString().contains(imageSelectedText)) {
-                                imageView.setContentDescription(
-                                        imageView.getContentDescription() + " "
-                                        + imageSelectedText);
-                            }
+            AccessibilityUtils.addPopulateAccessibilityEventFocusedListener(imageView, event -> {
+                if (isValidPosition(position)) {
+                    if (isItemSelected(position)) {
+                        final String imageSelectedText = imageView.getContext().getString(
+                                R.string.photo_picker_image_selected);
+                        if (!imageView.getContentDescription().toString().contains(imageSelectedText)) {
+                            imageView.setContentDescription(
+                                    imageView.getContentDescription() + " "
+                                    + imageSelectedText);
                         }
                     }
-                });
-            }
+                }
+            });
         }
 
         boolean isValidPosition(int position) {
--- a/WordPress/src/main/java/org/wordpress/android/ui/stockmedia/StockMediaPickerActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stockmedia/StockMediaPickerActivity.java
@@ -22,6 +22,7 @@ import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.appcompat.widget.SearchView;
 import androidx.appcompat.widget.Toolbar;
+import androidx.core.view.ViewCompat;
 import androidx.fragment.app.FragmentManager;
 import androidx.recyclerview.widget.GridLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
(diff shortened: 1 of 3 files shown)


jd-alexander:
> addImageSelectedToAccessibilityFocusedEvent is called with a position parameter and that position is then used to determine if it should have a specific content description. With this fix, we'll ignore the position when the same view is recycled and used in a different position which means it'll not have the text added. Am I misinterpreting this?

Yes, that's a good point. I just checked and you are right about recycling. The views weren't recycling for me but after a few scrolls on a test device, I saw this happening.

jd-alexander:
@oguzkocer so the previous solution worked well where each time the view was bound the delegate was being set but based on how the view recycling works is it better to just overwrite the delegate for each position each time? The only drawback that would happen is if another developer in the future is making improvements and they don't see this delegate being set they could overwrite it unintentionally so I am still thinking of a way to make the improvement keep the safety check.

jd-alexander:
I will ping you for another review while I run some checks 👍

oguzkocer:
> @oguzkocer so the previous solution worked well where each time the view was bound the delegate was being set but based on how the view recycling works is it better to just overwrite the delegate for each position each time?

I think that depends on what our goal is about the safety check. If all we want to do is raise awareness about "overwriting" then I think we can just get away with using good & clear names such as `setOrOverwrite`. That'd make sense if overwriting the behavior makes sense in certain situations such as in `RecyclerView`.

However, if we definitely want to keep the safety check because the downside of overwriting is major (I don't know what that downside is, again really no knowledge of accessibility whatsoever, but if you can help me understand it, I might be able to help more) then I think keeping the safety check as is and maybe fixing this crash with something like below would make more sense:

1. In `onBindViewHolder` add `ViewCompat.setAccessibilityDelegate(holder.mImageView, null);`. In general it's important to "reset" a `ViewHolder" and this would be a part of that. Just this change would be enough to fix the crash without the changes in this PR. (or you can add this code under `addImageSelectedToAccessibilityFocusedEvent` but in that case I suggest changing the method name to reflect that it'll remove the existing delegate if necessary, so maybe something like `resetAnd` would work)
2. In the crash message add a warning such as "You can't override the accessibility delegate directly. Please make sure to call `ViewCompat.setAccessibilityDelegate({yourView}, null)` if you want to override. Or we could even have a helper method with a documentation on top about why it's important for us to manually remove this delegate before overriding. (explaining why we used a destructive crash over a simple warning or a specific name)

Hopefully these will help make the decision easier.

jd-alexander:
> I think that depends on what our goal is about the safety check. If all we want to do is raise awareness about "overwriting" then I think we can just get away with using good & clear names such as setOrOverwrite. That'd make sense if overwriting the behavior makes sense in certain situations such as in RecyclerView.

Okay understood. This is a good idea!

> However, if we definitely want to keep the safety check because the downside of overwriting is major (I don't know what that downside is, again really no knowledge of accessibility whatsoever, but if you can help me understand it, I might be able to help more) then I think keeping the safety check as is and maybe fixing this crash with something like below would make more sense:

Yes, so the safety check is just defensive programming in action. Say someone was to come along and create a new `AccessibilityDelegate` that analyzes the `ImageView` and then sets a content description with that data then this `AccessibilityDelegate` that tells if an `ImageView` is selected or not would be overwritten, meaning if they used the default  `ViewCompat.setAccessibilityDelegate` that does no checks; all it does is overwrite the current delegate without a warning. So this util function `setAccessibilityDelegateSafely` is for us to promote good design so when someone utilizes this it will warn if one already exists and they could either modify the existing one or remove it altogether.

> 1. In onBindViewHolder add ViewCompat.setAccessibilityDelegate(holder.mImageView, null);. In general it's important to "reset" a ViewHolder" and this would be a part of that. Just this change would be enough to fix the crash without the changes in this PR. (or you can add this code under addImageSelectedToAccessibilityFocusedEventbut in that case I suggest changing the method name to reflect that it'll remove the existing delegate if necessary, so maybe something likeresetAnd` would work)

Yes, this is a good approach due to the nature of `ViewHolders` and the method name would communicate what's taking place indeed 🤝 

>2. In the crash message add a warning such as "You can't override the accessibility delegate directly. Please make sure to call ViewCompat.setAccessibilityDelegate({yourView}, null) if you want to override. Or we could even have a helper method with a documentation on top about why it's important for us to manually remove this delegate before overriding. (explaining why we used a destructive crash over a simple warning or a specific name)

Yes, modifying the current helper method with this comment will help. 

> Hopefully these will help make the decision easier.

Yes it does! Good discussion! So to summarize to ensure I get a clear understanding, in cases where the delegate has to be reset then the method name should let that be known but in other cases, the safety check util function can be utilized. Overall, there will be an accessibility guideline for the repo so this can be placed in it and that will enforce the importance of how the overriding should work 💯

oguzkocer:
> Yes, so the safety check is just defensive programming in action. Say someone was to come along and create a new `AccessibilityDelegate` that analyzes the `ImageView` and then sets a content description with that data then this AccessibilityDelegate that tells if an `ImageView` is selected or not would be overwritten, meaning if they used the default `ViewCompat.setAccessibilityDelegate` that does no checks; all it does is overwrite the current delegate without a warning. So this util function `setAccessibilityDelegateSafely` is for us to promote good design so when someone utilizes this it will warn if one already exists and they could either modify the existing one or remove it altogether.

I think I understand it a bit better now and although I am a big fan of defensive programming in general, I don't actually see its value in this specific case. What's the difference between setting the accessibility delegate and setting the text of a view? If we think about the `setText` example, there is nothing stopping from another developer setting the text for a second time and overwriting what's already there, so would we want to add a catch for this case as well? Probably not, so we have to trust other devs with basic things that don't require any specific knowledge especially for non-destructive actions.

So, in this case, I think the question is "Does the accessibility delegate needs special attention". If you believe so, these checks are a good thing for us. Otherwise, it's probably not worth introducing the overhead and making things more complicated than it needs to be. It's completely your call and I don't mind one way or another. I am just sharing my thought process.

Thank you for giving me more context about this PR and the current implementation. I really appreciate that and this discussion :)

mchowning:
This is an interesting issue. 😄 

>What's the difference between setting the accessibility delegate and setting the text of a view?

I think that's generally a good point, with the caveat that an issue with the visible text of a view is going to be much much more likely to be noticed compared to an accessibility issue (that needs TalkBack enabled to be noticed). 

For that reason, I think that a good compromise might be to just log it any time we overwrite an accessibility delegate (instead of causing a crash on debug builds). That way we give some higher visibility to the issue. At the same time, we also avoid the confusing “wait we’re using `position` in the accessibility delegate, but we’re not resetting the accessibility delegate if a view is recycled” scenario that @oguzkocer noticed [here](https://github.com/wordpress-mobile/WordPress-Android/pull/11046#pullrequestreview-339466257).

I don't feel strongly about the logging, but I do think that not resetting the accessibility delegate in `onBindViewHolder` when it is using the position is confusing and we should avoid that because it would definitely confuse me if I were looking at this code fresh.

jd-alexander:
So to resolve this issue I simply opted to set the accessibility delegate with `ViewCompat.setAccessibilityDelegate` 0b0cf44

Video of Free Photo Library Working  |
--------|
  [Click To Watch](https://drive.google.com/file/d/19RVFqhMG2UiZ5hmJ9v2g6uAfS9t8zzAw/view?usp=sharing)        |       

All media pickers are working. Accessibility functionality that this code touches still intact.

Gallery Picker | Free Photo Picker 
--------|-------
        |  

WordPress Media Picker |
--------|
       |

PR #11049, opened 2020-01-07
Accessibility : Posting Stats Improvement v2 

### LINKED ISSUE #10975: Accessibility : Stats Posting Activity Improvements

jd-alexander:
Fixes #10975 
## Findings
The Posting Stats was inaccessible when using TalkBack. This improvement makes it more accessible for visually impaired users who are navigating the stats functionality. 

## Solution
The solution for this is to highlight the views by grouping them by the days that share the same ranking([This solution was discussed in detail on this PR](https://github.com/wordpress-mobile/WordPress-Android/pull/10980#pullrequestreview-334602495)). Once this is done, the user is able to hear the days with the highest views as soon as they interact with the component and then can continue to cycle through the rest of activity with subsequent clicks. 

Before | After 
--------|-------
[Before Video ](https://drive.google.com/open?id=18ErO8-vpZxLuxucriVecOq218nEcmaSK) |   [After Video    ](https://drive.google.com/file/d/19AwFLbvKmtyl16pL9nG3BLGEDjRnKA_h/view?usp=sharing)

## Testing

1. Enable TalkBack. 
2. Navigate to the Stats page.
3. Ensure you are on the Insights Tab. 
4. Select each Month under "Posting Activity" and follow the instructions to hear the announcement.

Tap To Hear Day Stats | Stat Announced 
--------|-------
        |       

All Stats Announced | No Stats Available 
--------|-------
   |    

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
@@ -237,7 +237,7 @@ sealed class BlockListItem(val type: Type) {
             val label: String,
             val boxes: List<Box>,
             val contentDescription: String,
-            val activityContentDescriptions: List<String>? = null
+            val activityContentDescriptions: List<String>
         )
 
         enum class Box {
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
@@ -240,9 +240,7 @@ sealed class BlockListItem(val type: Type) {
             val activityContentDescriptions: List<String>? = null
         )
 
-        data class Box(val boxType: BoxType, val day: Int? = null)
-
-        enum class BoxType {
+        enum class Box {
             INVISIBLE, VERY_LOW, LOW, MEDIUM, HIGH, VERY_HIGH
         }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
@@ -236,6 +236,7 @@ sealed class BlockListItem(val type: Type) {
         data class Block(
             val label: String,
             val boxes: List<Box>,
+            val contentDescription: String,
             val activityContentDescriptions: List<String>? = null
         )
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
@@ -233,8 +233,15 @@ sealed class BlockListItem(val type: Type) {
     data class LoadingItem(val loadMore: () -> Unit, val isLoading: Boolean = false) : BlockListItem(LOADING_ITEM)
 
     data class ActivityItem(val blocks: List<Block>) : BlockListItem(ACTIVITY_ITEM) {
-        data class Block(val label: String, val boxes: List<Box>)
-        enum class Box {
+        data class Block(
+            val label: String,
+            val boxes: List<Box>,
+            val activityContentDescriptions: List<String>? = null
+        )
+
+        data class Box(val boxType: BoxType, val day: Int? = null)
+
+        enum class BoxType {
             INVISIBLE, VERY_LOW, LOW, MEDIUM, HIGH, VERY_HIGH
         }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/insights/usecases/PostingActivityMapper.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/insights/usecases/PostingActivityMapper.kt
@@ -57,7 +57,13 @@ class PostingActivityMapper
                                     Calendar.SHORT,
                                     localeManagerWrapper.getLocale()
                             ),
-                            boxes
+                            boxes,
+                            firstDayOfMonth.getDisplayName(
+                                    Calendar.MONTH,
+                                    Calendar.LONG,
+                                    localeManagerWrapper.getLocale()
+                            )
+
                     )
             )
         }
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/insights/usecases/PostingActivityMapper.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/insights/usecases/PostingActivityMapper.kt
@@ -1,16 +1,18 @@
 package org.wordpress.android.ui.stats.refresh.lists.sections.insights.usecases
 
+import org.wordpress.android.R
 import org.wordpress.android.fluxc.model.stats.insights.PostingActivityModel.Month
 import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ActivityItem
 import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ActivityItem.Block
 import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ActivityItem.Box
-import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ActivityItem.BoxType.HIGH
-import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ActivityItem.BoxType.INVISIBLE
-import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ActivityItem.BoxType.LOW
-import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ActivityItem.BoxType.MEDIUM
-import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ActivityItem.BoxType.VERY_HIGH
-import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ActivityItem.BoxType.VERY_LOW
+import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ActivityItem.Box.HIGH
+import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ActivityItem.Box.INVISIBLE
+import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ActivityItem.Box.LOW
+import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ActivityItem.Box.MEDIUM
+import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ActivityItem.Box.VERY_HIGH
+import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ActivityItem.Box.VERY_LOW
 import org.wordpress.android.util.LocaleManagerWrapper
+import org.wordpress.android.viewmodel.ResourceProvider
 import java.util.Calendar
 import javax.inject.Inject
 
@@ -20,7 +22,10 @@ private const val MEDIUM_LEVEL = 0.25
 private const val LOW_LEVEL = 0
 
 class PostingActivityMapper
-@Inject constructor(private val localeManagerWrapper: LocaleManagerWrapper) {
+@Inject constructor(
+    private val localeManagerWrapper: LocaleManagerWrapper,
+    private val resourceProvider: ResourceProvider
+) {
     fun buildActivityItem(months: List<Month>, max: Int): ActivityItem {
         val blocks = mutableListOf<Block>()
         val veryHighLimit = (max * VERY_HIGH_LEVEL).toInt()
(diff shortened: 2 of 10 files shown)


WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/insights/usecases/PostingActivityMapper.kt
```diff
@@ -55,7 +57,13 @@ class PostingActivityMapper
                                     Calendar.SHORT,
                                     localeManagerWrapper.getLocale()
                             ),
-                            boxes
+                            boxes,
```
jd-alexander: I needed to create this content description for the Block because the label is the short version of the month so "Dec" is being read as "Deck" so this content description allows me to pass the long version of the month.

WordPress/src/main/java/org/wordpress/android/util/ContentDescriptionListAnnouncer.kt
```diff
@@ -0,0 +1,67 @@
...
+                    override fun onInitializeAccessibilityNodeInfo(
+                        host: View?,
+                        info: AccessibilityNodeInfoCompat?
+                    ) {
+                        super.onInitializeAccessibilityNodeInfo(host, info)
+
+                        clickActionText?.let {
+                            info?.addAction(
+                                    AccessibilityNodeInfoCompat.AccessibilityActionCompat(
+                                            ACTION_CLICK, host?.context?.getString(
```
jd-alexander: In this class, I utilize the `context` of the host view so I don't have to inject the `ResourceProvider` since this class isn't testable and doesn't need to be right now since it's just a view related one. What do you think?

WordPress/src/main/java/org/wordpress/android/util/ContentDescriptionListAnnouncer.kt
```diff
@@ -0,0 +1,67 @@
...
+class ContentDescriptionListAnnouncer constructor(
+    @StringRes private val emptyListText: Int,
+    @StringRes private val endOfListText: Int,
+    @StringRes private val clickActionText: Int? = null,
+    private val contentDescriptions: List<String>,
+    targetView: View
+) {
+    private var currentIndex = 0
+
+    init {
```
jd-alexander: I am using the `init` to wire up the behavior. What I noticed is that the class is just instantiated and the functionality works so I am wondering if that's fine from a readable/architecture standpoint or should there be a function that's called so that the action being taken when the class is created is easily expressed.

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/insights/usecases/PostingActivityUseCase.kt
```diff
@@ -55,11 +61,43 @@ class PostingActivityUseCase
...
+    private fun addBlockContentDescriptions(activityItem: ActivityItem): ActivityItem {
+        val blocks = mutableListOf<Block>()
+
+        activityItem.blocks.forEach { block ->
+            val descriptions = mutableListOf<String>()
+            block.boxes.filter { box -> box.boxType != BoxType.INVISIBLE && box.boxType != BoxType.VERY_LOW }
+                    .sortedByDescending { box -> box.boxType }
+                    .groupBy { box -> box.boxType }
+                    .forEach { entry ->
+                        val readableBoxType = entry.key.name.replace(
```
jd-alexander: I just realized this isn't localized so I will switch this out with an implementation that is.

jkmassel:

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
```diff
@@ -233,8 +233,16 @@ sealed class BlockListItem(val type: Type) {
...
-        data class Block(val label: String, val boxes: List<Box>)
-        enum class Box {
+        data class Block(
+            val label: String,
+            val boxes: List<Box>,
+            val contentDescription: String,
+            val activityContentDescriptions: List<String>? = null
+        )
+
+        data class Box(val boxType: BoxType, val day: Int? = null)
```
malinajirka: I'm not familiar with this part of the app. However, I'm wondering whether it makes sense to introduce `day` field in the `Box` class => the field isn't being used on the UI. It's being used only when we are building the ActivityItem. We might want to consider reverting some of the changes in `ActivityItem` and moving `buildActivityItem` to `PostingActivityMapper`, wdyt?

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/insights/usecases/PostingActivityMapper.kt
```diff
@@ -29,33 +29,38 @@ class PostingActivityMapper
...
             }
             blocks.add(
                     Block(
-                            firstDayOfMonth.getDisplayName(
-                                    Calendar.MONTH,
-                                    Calendar.SHORT,
-                                    localeManagerWrapper.getLocale()
-                            ),
-                            boxes
+                            getMonthDisplayName(Calendar.SHORT),
```
malinajirka: We might want to consider settings the content description to something like "Posting Activity for $getMonthDisplayName". Wdyt?

WordPress/src/main/java/org/wordpress/android/util/ContentDescriptionListAnnouncer.kt
malinajirka: Yep, I agree that it's ok => the class isn't testable anyway and making it testable would require more changes.

WordPress/src/main/res/values/strings.xml
```diff
@@ -884,9 +884,19 @@
...
-    <string name="stats_posting_activity_graph">Graph of posting activity.</string>
     <string name="stats_list_item_description">%1$s: %2$s, %3$s: %4$s</string>
 
+    <string name="stats_box_type_low">low</string>
+    <string name="stats_box_type_medium">medium</string>
+    <string name="stats_box_type_high">high</string>
+    <string name="stats_box_type_very_high">very high</string>
+
+    <string name="stats_posting_activity_action">navigate all stats for this period</string>
+    <string name="stats_posting_activity_content_description">"The days with %1$s views for %2$s are : %2$s %3$s."</string>
```
malinajirka: We might want to consider ending this announcement with ", tap for more". Or is it a common practice that clicking multiple times on the same view keeps going more into detail?

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/ActivityViewHolder.kt
```diff
@@ -50,6 +51,23 @@ class ActivityViewHolder(val parent: ViewGroup) : BlockListItemViewHolder(
...
+        setupBlocksForAccessibility(item)
+    }
+
+    private fun setupBlocksForAccessibility(item: ActivityItem) {
+        val blocks = listOf(firstBlock, secondBlock, thirdBlock)
+
+        blocks.forEachIndexed { index, block ->
+            block.label.contentDescription = item.blocks[index].contentDescription
+
+            ContentDescriptionListAnnouncer(
```
malinajirka: It feels a bit weird that we create an Object and we don't do anything with it. I know the class isn't testable anyway, but putting logic into constructor/init methods is a code smell. I'd consider moving the logic into a method. Wdyt?

malinajirka (approved):
Really great job @jd-alexander! I haven't found any issues and the code looks good.

I've left a few comments for consideration. Especially one of them might result in more changes, but I don't feel strongly about it. It just feels a bit weird. Let me know what you think;)

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/insights/usecases/PostingActivityMapper.kt
jd-alexander: That sounds good! I will look into it.

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
jd-alexander: Hi Jirka! Thanks for this observation. It's something we can think of modifying if it makes sense. The reason I put it here is the `contentDescription` is somewhat a UI element, it's just a different kind of UI for other users so I thought it made sense to include it.

WordPress/src/main/res/values/strings.xml
jd-alexander: This is a good observation. I will include this change. Thanks!

jd-alexander: Resolved 40cd081

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/insights/usecases/PostingActivityMapper.kt
jd-alexander: Resolved 40cd081

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/ActivityViewHolder.kt
jd-alexander: Resolved 962663d

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
malinajirka: >contentDescription is somewhat a UI element, it's just a different kind of UI for other users so I thought it made sense to include it.

I absolutely agree with this statement. Having `contentDescription` field in `Block` is ok. 
However, it seems `ActivityItem` and `Box` are UIStates and `day` is not part of their UIState = it isn't being used in the view layer. We use it only to store some data we need to use while we are building the UIState.

It's not a big deal, it just feels weird.

WordPress/src/main/res/values/strings.xml
```diff
@@ -893,6 +893,7 @@
 
     <string name="stats_posting_activity_action">navigate all stats for this period</string>
     <string name="stats_posting_activity_content_description">"The days with %1$s views for %2$s are : %2$s %3$s, tap for more."</string>
+    <string name="stats_posting_activity_label_content_description">"Posting Activity for %1$s</string>
```
malinajirka: The string starts with `"`, is it intentional?

WordPress/src/main/res/values/strings.xml
```diff
@@ -884,9 +884,20 @@
...
-    <string name="stats_posting_activity_graph">Graph of posting activity.</string>
     <string name="stats_list_item_description">%1$s: %2$s, %3$s: %4$s</string>
 
+    <string name="stats_box_type_low">low</string>
+    <string name="stats_box_type_medium">medium</string>
+    <string name="stats_box_type_high">high</string>
+    <string name="stats_box_type_very_high">very high</string>
+
+    <string name="stats_posting_activity_action">navigate all stats for this period</string>
+    <string name="stats_posting_activity_content_description">"The days with %1$s views for %2$s are : %2$s %3$s.Tap for more."</string>
```
malinajirka: I have missed this during the review, but it seems there are at least 3 strings starting or/and ending with `"`. Is it intentional?

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
jd-alexander: Okay! I totally get what you mean now when you connect it to the concept of `UIState`. Thanks for the explanation. So are you saying I should move this logic to the `UseCase` then?

WordPress/src/main/res/values/strings.xml
jd-alexander: Thanks for catching this 😅

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
malinajirka: `PostingActivityMapper.buildActivityItem` returns an instance of `ActivityItem`, we pass that `ActivityItem` into `addBlockContentDescriptions` and we create a different instance of `ActivityItem` there, which also feels a bit weird. 
I think the best place for the contentDescription logic would be in `PostingActivityMapper` =>`PostingActivityMapper.buildActivityItem` would return a complete `ActivityItem`. Wdyt?

WordPress/src/main/res/values/strings.xml
jd-alexander: Resolved b5f6951

WordPress/src/main/res/values/strings.xml
jd-alexander: Resolved b5f6951

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
jd-alexander: Resolved 50224c5

jd-alexander:
Hey @malinajirka I made all the changes we discussed and I even included unit tests for the content description logic f077555 😉

malinajirka:
Thanks @jd-alexander ;)! Really glad to see the unit tests.

I noticed the days read by the TalkBack don't match the days one the UI. To be more specific, it works for the first month, but the same day is read for the other months.

posting-activity
- Notice it keeps reading "25." for all the months

jd-alexander:
Thanks for the review of the whole component. This issue was resolved acf5db7 and the tests were updated 0f7afe1 to ensure that the data that corresponds to each month is accurate.

malinajirka:
Thanks @jd-alexander !

I've tested the changes and everything seems to work as expected now. Thank you for extending the unit test so it covers even the scenario which wasn't working.

SideNote:
I don't want to block this PR anymore, but I'll leave a brief feedback about the unit test here. Unit tests have several goals. Probably the two main goals are to test the logic (facilitate refactoring/modifications) and serve as a documentation. I think the unit test in this PR somewhat meets the first goal, however I feel like it doesn't server as a documentation. I believe the main reason for this is that we are testing everything within one test method. In general, it's better to have each scenario tested in a separate method so each test has only a single reason to fail.

PR #11050, opened 2020-01-07
Move ReattachUploadMedia to a usecase

malinajirka:
Moves `reattachUploadingMediaForAztec` to a usecase.

To test:
1. Create a new draft in Aztec
2. Attach an image (You might need to slow down the internet connection)
3. Rotate your device before the upload completes
4. Make sure `uploadServiceFacade.getUploadProgressForMedia(..)` gets invoked (Eg. put breakpoint/log [here](https://github.com/wordpress-mobile/WordPress-Android/blob/2223b16c80727424d79fde464c06f759031f39f5/WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/ReattachUploadingMediaUseCase.kt#L32))
-------------------
1. Run the same test, just add multiple images in step 2 and make sure the method gets called for all of them.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/ReattachUploadingMediaUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/ReattachUploadingMediaUseCase.kt
@@ -11,31 +11,28 @@ class ReattachUploadingMediaUseCase @Inject constructor(
 ) {
     fun reattachUploadingMediaForAztec(
         editPostRepository: EditPostRepository,
-        isAztec: Boolean,
         editorMediaUploadListener: EditorMediaUploadListener
     ) {
-        if (isAztec) {
-            // UploadService.getPendingMediaForPost will be populated only when the user exits the editor
-            // But if the user doesn't exit the editor and sends the app to the background, a reattachment
-            // for the media within this Post is needed as soon as the app comes back to foreground,
-            // so we get the list of progressing media for this Post from the UploadService
-            val uploadingMediaInPost = editPostRepository.getPendingMediaForPost()
-            val allUploadingMediaInPost = ArrayList<MediaModel>(uploadingMediaInPost)
+        // UploadService.getPendingMediaForPost will be populated only when the user exits the editor
+        // But if the user doesn't exit the editor and sends the app to the background, a reattachment
+        // for the media within this Post is needed as soon as the app comes back to foreground,
+        // so we get the list of progressing media for this Post from the UploadService
+        val uploadingMediaInPost = editPostRepository.getPendingMediaForPost()
+        val allUploadingMediaInPost = ArrayList<MediaModel>(uploadingMediaInPost)
 
-            editPostRepository.getPendingOrInProgressMediaUploadsForPost().forEach { media1 ->
-                // add them to the array only if they are not in there yet
-                if (uploadingMediaInPost.none { media2 -> media1.id == media2.id }) {
-                    allUploadingMediaInPost.add(media1)
-                }
+        editPostRepository.getPendingOrInProgressMediaUploadsForPost().forEach { media1 ->
+            // add them to the array only if they are not in there yet
+            if (uploadingMediaInPost.none { media2 -> media1.id == media2.id }) {
+                allUploadingMediaInPost.add(media1)
             }
+        }
 
-            // now do proper re-attachment of upload progress on each media item
-            allUploadingMediaInPost.forEach { mediaModel ->
-                editorMediaUploadListener.onMediaUploadReattached(
-                        mediaModel.id.toString(),
-                        uploadServiceFacade.getUploadProgressForMedia(mediaModel)
-                )
-            }
+        // now do proper re-attachment of upload progress on each media item
+        allUploadingMediaInPost.forEach { mediaModel ->
+            editorMediaUploadListener.onMediaUploadReattached(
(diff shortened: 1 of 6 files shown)


WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/ReattachUploadingMediaUseCase.kt
```diff
@@ -0,0 +1,41 @@
+package org.wordpress.android.ui.posts.editor.media
+
+import org.wordpress.android.editor.EditorMediaUploadListener
+import org.wordpress.android.fluxc.model.MediaModel
+import org.wordpress.android.ui.posts.EditPostRepository
+import org.wordpress.android.ui.uploads.UploadServiceFacade
+import javax.inject.Inject
+
+class ReattachUploadingMediaUseCase @Inject constructor(
+    private val editPostRepository: EditPostRepository,
```
planarvoid: I don't think this will work because the injected edit post repository is not initialized with the post. What we do in the `EditPostActivity` is that we inject the repository and load the post in `onCreate`. We do not share the same instance of the repository because there might be more instances of the activity. I think we have to pass the repository as a parameter in `reattachUploadingMediaForAztec`.

malinajirka: Oh, my bad. You are right. I'm wondering now, what the code actually does - because I tested the functionality after the changes and it worked :D.

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/ReattachUploadingMediaUseCase.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.wordpress.android.ui.posts.EditPostRepository
+import org.wordpress.android.ui.uploads.UploadServiceFacade
+import javax.inject.Inject
+
+class ReattachUploadingMediaUseCase @Inject constructor(
+    private val uploadServiceFacade: UploadServiceFacade
+) {
+    fun reattachUploadingMediaForAztec(
+        editPostRepository: EditPostRepository,
+        isAztec: Boolean,
```
planarvoid: I'm thinking about this approach and it just doesn't feel right - passing a boolean field and not do anything if it's false. I'd move this check to the EditorMedia and maybe renamed this use case to include `Aztec` so no one is tempted to use it with GB?

malinajirka: Yeah I agree it feels weird. I wanted to emphasise it can be removed as soon as we remove Aztec,  but it doesn't make sense. 
I've moved it to EditorMedia as you suggested. We are still passing a boolean and do nothing when it's false though. Didn't you want me to move it to the EditPostActivity?

malinajirka:
@jd-alexander Could you please finish the review of this PR, since Vojta won't be available any time soon. Thanks!

jd-alexander:
Hi @malinajirka 

Have you ever seen this errror? 

```
   --------- beginning of crash
2020-01-21 21:01:48.788 22119-22337/org.wordpress.android.beta E/AndroidRuntime: FATAL EXCEPTION: create_react_context
    Process: org.wordpress.android.beta, PID: 22119
    java.lang.UnsatisfiedLinkError: couldn't find DSO to load: libreactnativejni.so
        at com.facebook.soloader.SoLoader.doLoadLibraryBySoName(SoLoader.java:738)
        at com.facebook.soloader.SoLoader.loadLibraryBySoName(SoLoader.java:591)
        at com.facebook.soloader.SoLoader.loadLibrary(SoLoader.java:529)
        at com.facebook.soloader.SoLoader.loadLibrary(SoLoader.java:484)
        at com.facebook.react.bridge.ReactBridge.staticInit(ReactBridge.java:31)
        at com.facebook.react.bridge.NativeMap.<clinit>(NativeMap.java:19)
        at com.facebook.react.jscexecutor.JSCExecutorFactory.create(JSCExecutorFactory.java:25)
        at com.facebook.react.ReactInstanceManager$5.run(ReactInstanceManager.java:949)
        at java.lang.Thread.run(Thread.java:919)
```

It seems to be coming from React Native. When I try to create a new post it happens.

malinajirka:
Hi @jd-alexander, nope I haven't. Is it crashing only on this branch or even on develop?

I think it's either an issue with build cache or with the submodule.
You might want to try running
```
./gradlew --stop
./gradlew clean
 ./gradlew  --no-build-cache assembleWasabiDebug
```

jd-alexander:
Hi @malinajirka thanks for your suggestions 😄 
It seems the error was coming from a needed kotlin dependency update that was merged in to develop https://github.com/wordpress-mobile/WordPress-Android/pull/11094
Once I merged this into my local branch, everything started working well again.

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/ReattachUploadingMediaUseCase.kt
jd-alexander: I somehow prefer the `if` being in the `EditPostActivity` because it's one less method parameter and it makes the intent clear since it will be obvious that `mEditorMedia.reattachUploadingMediaForAztec` shouldn't be called unless the current editor is Aztec. It's a small change though, so this can still be pushed as is.

PR #11052, opened 2020-01-07
Gutenberg: update to v1.20.0

Tug:
Gutenberg mobile "master" PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1752

This PR updates the gutenberg-mobile reference to point to the latest release, `v1.20.0`. No wpandroid integration changes introduced.

The `libs/gutenberg-mobile` reference is currently pointing to the `release/1.20.0` branch on https://github.com/wordpress-mobile/gutenberg-mobile.

When that gets merged to gutenberg-mobile master, I'll update the ref here to point to the merge commit.

## Changes:
- Fix displaying placeholder for images in some cases when the image cannot be downloaded (https://github.com/wordpress-mobile/gutenberg-mobile/pull/1631)
- Introduce new Stepper UI component, used in the Spacer block to select the block's height (https://github.com/wordpress-mobile/gutenberg-mobile/pull/1640)
- Add a testID to the inserter button so in can be used for AT (https://github.com/wordpress-mobile/gutenberg-mobile/pull/1681)
- Hide image size selection options when image is a url (https://github.com/wordpress-mobile/gutenberg-mobile/pull/1669)
- Add experimental page template selector (`__DEV__`) (https://github.com/wordpress-mobile/gutenberg-mobile/pull/1576)
- Fix crash on undo (https://github.com/wordpress-mobile/gutenberg-mobile/pull/1701)
- Style fixes on navigation UI (https://github.com/wordpress-mobile/gutenberg-mobile/pull/1754)

To test:
The block editor should work as normal.

Update release notes:

- [x] If there are user facing changes, I have added an item to RELEASE-NOTES.txt.


etoledom:
I've found these two small issues so far:

1- The strikethrough icon is not highlighted when the cursor is over a strikethrough-ed text.
2- The Layout looks a bit messy on the Images source selection bottom sheet.

Tested on Device Mi A2 Lite, Android 9, using the APK from CircleCI

android

Tug:
@etoledom it seems you strike-through text is also underlined? I think that might be one of aztec<-> gutenberg limitations when formatting text, aztec might just produce other tags that the one gutenberg format-library supports.

2 is indeed a problem, and more generally for translated text that might be longer. I have also noticed on my (iOS) device that we were showing the block selection bottom sheet on 2 columns which crops some of the blocks. So in my opinion screen sizes looks like a more general problem we should address, but probably not as part of this release, wdyt?

etoledom:
> aztec might just produce other tags that the one gutenberg format-library supports.

Yes that's it! Aztec creates a `<del>` tag, and Gutenberg expects an `<s>` tag. 👍

etoledom:
About 2: I remember have seen this issue before, and @marecar3 fixed it (If I'm not being confused). I'm trying to find that ticket but haven't found it yet.

@marecar3 maybe you remember?

etoledom:
Just tested the version from the store (Beta) and it still happens there, so I guess is not a regression 👍 

> So in my opinion screen sizes looks like a more general problem we should address, but probably not as part of this release, wdyt?

Agreed!

marecar3:
> About 2: I remember have seen this issue before, and @marecar3 fixed it (If I'm not being confused). I'm trying to find that ticket but haven't found it yet.
> 
> @marecar3 maybe you remember?

Hey @etoledom, I can't remember that I was working on that one 😕

etoledom:
Thank you @marecar3 ! I got confused then 👍 

**There's one more:** Captions on Images and Gallery have issues with multiline when pressing enter.
@pinarol any idea? If this seems important enough, we might want to target the release branch with a hot-fix. 

caption

etoledom:
Everything else seems to work well. I'm happy to ✅ if all the issues described are not blockers for this PR.

Tug:
It's not a regression but it's still pretty serious as caption might have line breaks (at least it's possible on the web) and we might lose content when editing on mobile.

etoledom:
> It's not a regression but it's still pretty serious as caption might have line breaks (at least it's possible on the web) and we might lose content when editing on mobile.

Tested on iOS and it works well. It might be related to Aztec Android?

hypest:
The removal of the newlines in captions on Android is a known issue:
https://github.com/wordpress-mobile/gutenberg-mobile/issues/1651

hypest:
✅ Fix displaying placeholder for images in some cases when the image cannot be downloaded (wordpress-mobile/gutenberg-mobile#1631)
🆗 (not exposed to any user-facing UI yet) Introduce new Stepper UI component, used in the Spacer block to select the block's height (wordpress-mobile/gutenberg-mobile#1640)
⌛️ Add a testID to the inserter button so in can be used for AT (wordpress-mobile/gutenberg-mobile#1681)
⌛️ Hide image size selection options when image is a url (wordpress-mobile/gutenberg-mobile#1669)
⌛️ Add experimental page template selector (__DEV__) (wordpress-mobile/gutenberg-mobile#1576)
⌛️ Fix crash on undo (wordpress-mobile/gutenberg-mobile#1701)
⌛️ Style fixes on navigation UI (wordpress-mobile/gutenberg-mobile#1754)

RELEASE-NOTES.txt
```diff
@@ -1,5 +1,9 @@
 14.0
 -----
+* Block Editor: Hide image size selection options when image is a url
```
etoledom: The image size selector is still under dev flag, so this should not affect the release build.

pinarol:
>There's one more: Captions on Images and Gallery have issues with multiline when pressing enter.
@pinarol any idea? If this seems important enough, we might want to target the release branch with a hot-fix.

As @hypest said it is a known issue. We [tried](https://github.com/WordPress/gutenberg/pull/19118#issuecomment-571939480) fixing that for the individual tiles but this time an interesting bug happens where the first word of caption can not be styled.

hypest:
> We tried fixing that for the individual tiles but this time an interesting bug happens where the first word of caption can not be styled.

Just to be sure @pinarol , do you mean that there's a new issue in this PR? Or was it an issue that happened in the attempts to fix the removing caption newlines issue?

pinarol:
> Just to be sure @pinarol , do you mean that there's a new issue in this PR? Or was it an issue that happened in the attempts to fix the removing caption newlines issue?

The "First word not stylable" problem is currently reproducible only on individual Gallery tiles because it uses its own RichText and not the Caption component. The existing Caption component was suitable only for blocks so we couldn't use it on tiles. In this PR, newline issue does not exist on tiles because we solved it without noticing the "First word not stylable" regression.

We are about to [merge a PR](https://github.com/WordPress/gutenberg/pull/19118) to develop that refactors Caption component to make it suitable for both tiles and blocks. After that one is merged, both blocks and tiles will be using the same Caption component so they will have the same bug which is the newline issue.

hypest:
The strings update (1294de9) looks good to me.

PR #11056, opened 2020-01-07
Tooltip onboarding announcement for IA create FAB

### LINKED ISSUE #11012: IA create FAB: need some kind of tooltip to promote the FAB.

develric:
Fixes #11012

This PR adds an announcing tooltip for the create FAB in the My Site page.

The logic is the following:

- When this new app version is installed (or updated from previous older installation) the tooltip is shown near to the FAB

<p align="center">
  
</p>

- The tooltip is presented near the FAB until the user first tap on the FAB itself
- Once tapped on the FAB the tooltip is not shown any more and there is no way to have it back unless the app storage data is cleared or the user fully uninstall and reinstall the app

## Notes

- The tooltip work with RTL languages

<p align="center">
  
</p>

- Current implementation of `WPTooltipView` is pretty basic. Mainly to cover this specific scenario. More work/rework is needed to make it a full custom view tooltip component. I explored a bit also a more dynamic approach with the PopupWindow but messing around only with simple layouts seemed more convenient in this simple scenario. Also the placement and interactions (like the coordinator layout behavior with snackbars like in this case) comes at not cost in this way. (Nonetheless the PopupWindow approach can be something to further explore later and I have something basic working also there) 
- I used Chris original position and copy of the tooltip but we can change it (ping @osullivanchris)

## To test
Again all the feature is still behind feature flag so:

- First smock test with vanilla to check that no evident issues were introduced in the My Site and switching between pages in the `WPMainActivity`
- Compile zalpha flavor and check that you get the tooltip near the FAB
- Use the app without using the FAB (for example create a post from Blog Posts) and check that the tooltip is always presented when you come back to the My Site page
- Use the FAB and check that after that the tooltip is no more visualized. Also close and open the app to confirm.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/WPTooltipView.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/WPTooltipView.kt
@@ -66,7 +66,7 @@ class WPTooltipView @JvmOverloads constructor (
         inflate(getContext(), position.layout, this)
         val root = findViewById<LinearLayout>(R.id.root_view)
         val tvMessage = findViewById<TextView>(R.id.tooltip_message)
-        val arrow = findViewById<ImageView>(R.id.tooltip_arrow)
+        val arrow = findViewById<View>(R.id.tooltip_arrow)
 
         if (messageId > 0) {
             tvMessage.setText(messageId)
--- a/WordPress/src/main/java/org/wordpress/android/ui/WPTooltipView.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/WPTooltipView.kt
@@ -4,7 +4,6 @@ import android.content.Context
 import android.util.AttributeSet
 import android.view.Gravity
 import android.view.View
-import android.widget.ImageView
 import android.widget.LinearLayout
 import android.widget.TextView
 import androidx.annotation.LayoutRes
--- a/WordPress/src/main/java/org/wordpress/android/ui/WPTooltipView.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/WPTooltipView.kt
@@ -1,0 +1,95 @@
+package org.wordpress.android.ui
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.View
+import android.widget.FrameLayout
+import android.widget.ImageView
+import android.widget.LinearLayout
+import android.widget.TextView
+import androidx.annotation.LayoutRes
+import androidx.core.view.ViewCompat
+import org.wordpress.android.R
+import org.wordpress.android.ui.WPTooltipView.TooltipPosition.ABOVE
+import org.wordpress.android.ui.WPTooltipView.TooltipPosition.BELOW
+import org.wordpress.android.ui.WPTooltipView.TooltipPosition.LEFT
+import org.wordpress.android.ui.WPTooltipView.TooltipPosition.RIGHT
+import org.wordpress.android.util.RtlUtils
+import java.lang.IllegalArgumentException
+import java.util.jar.Attributes
+
+/**
+ * Partially based on https://stackoverflow.com/a/42756576
+ *
+ * NOTE: this is a very basic implementation of a tooltip component
+ * mainly used to cover the need of onboarding/announcing in the IA Project main create FAB.
+ * More work/rework is needed to make it a full custom view tooltip component.
+ * A different and more dynamic approach that can be used is with PopupWindow taking care of behaviors like
+ * CoordinatorLayout behavior (as in this scenario with FAB and snackbars)
+ */
+
+class WPTooltipView @JvmOverloads constructor (
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : LinearLayout(context, attrs, defStyleAttr) {
+    var position = LEFT
+    var messageId = 0
+
+    init {
+        attrs?.also {
+            val a = context.theme.obtainStyledAttributes(
+                    attrs,
+                    R.styleable.WPTooltipView,
+                    0,
+                    0
+            )
+
+            try {
+                position = TooltipPosition.fromInt(a.getInt(R.styleable.WPTooltipView_wpTooltipPosition, 0))
+                messageId = a.getResourceId(R.styleable.WPTooltipView_wpTooltipMessage, 0)
+            } finally {
+                a.recycle()
+            }
+        }
+
+        inflate(getContext(), position.layout, this)
+        val tvMessage = findViewById<TextView>(R.id.tooltip_message)
+
+        if (RtlUtils.isRtl(context)) {
+            val arrow = findViewById<ImageView>(R.id.tooltip_arrow)
+            when (position) {
+                LEFT -> arrow.rotation = -90f
+                RIGHT -> arrow.rotation = 90f
+                ABOVE, BELOW -> {}
+            }
+        }
+
+        if (messageId > 0) {
+            tvMessage.setText(messageId)
+        }
+    }
+
+    enum class TooltipPosition(val value: Int, @LayoutRes val layout: Int) {
+        LEFT(0, R.layout.tooltip_left),
+        RIGHT(1, R.layout.tooltip_right),
+        ABOVE(2, R.layout.tooltip_above),
+        BELOW(3, R.layout.tooltip_below);
+
+        companion object {
+            fun fromInt(value: Int): TooltipPosition =
+                    values().firstOrNull() { it.value == value }
+                            ?: throw IllegalArgumentException("TooltipPosition wrong value $value")
+        }
+    }
+
+    fun show() {
+        this.postDelayed({
+            this.visibility = View.VISIBLE
+        }, 600)
+    }
+
+    fun hide() {
+        this.visibility = View.GONE
+    }
+}
--- a/WordPress/src/main/java/org/wordpress/android/ui/WPTooltipView.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/WPTooltipView.kt
@@ -27,6 +27,8 @@ import java.lang.IllegalArgumentException
  * CoordinatorLayout behavior (as in this scenario with FAB and snackbars)
  */
 
+private const val HIDE_ANIMATION_DURATION = 50L
+
 class WPTooltipView @JvmOverloads constructor (
     context: Context,
     attrs: AttributeSet? = null,
@@ -130,7 +132,7 @@ class WPTooltipView @JvmOverloads constructor (
         this.apply {
             animate()
                     .alpha(0f)
-                    .setDuration(animationDuration.toLong())
+                    .setDuration(HIDE_ANIMATION_DURATION)
                     .setListener(object : AnimatorListenerAdapter() {
                         override fun onAnimationEnd(animation: Animator?) {
                             super.onAnimationEnd(animation)
--- a/WordPress/src/main/java/org/wordpress/android/ui/WPTooltipView.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/WPTooltipView.kt
@@ -1,5 +1,7 @@
 package org.wordpress.android.ui
 
+import android.animation.Animator
+import android.animation.AnimatorListenerAdapter
 import android.content.Context
 import android.util.AttributeSet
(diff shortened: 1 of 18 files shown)


WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
```diff
@@ -386,17 +388,24 @@ public boolean isGooglePlayServicesAvailable(Activity activity) {
...
         mFloatingActionButton = findViewById(R.id.fab_button);
+        mFabTooltip = findViewById(R.id.fab_tooltip);
 
         mViewModel = ViewModelProviders.of(this, mViewModelFactory).get(WPMainActivityViewModel.class);
 
         if (BuildConfig.INFORMATION_ARCHITECTURE_AVAILABLE) {
             // Setup Observers
-            mViewModel.getShowMainActionFab().observe(this, shouldShow -> {
-                if (shouldShow) {
+            mViewModel.getShowMainActionFab().observe(this, showUiState -> {
```
planarvoid: maybe call it `fabUiState` instead of `showUiState`?

WordPress/src/main/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModel.kt
```diff
@@ -3,21 +3,24 @@ package org.wordpress.android.viewmodel.main
...
 import javax.inject.Inject
 
-class WPMainActivityViewModel @Inject constructor() : ViewModel() {
+class WPMainActivityViewModel @Inject constructor(private val appPrefsWrapper: AppPrefsWrapper) : ViewModel() {
     private var isStarted = false
 
-    private val _showMainActionFab = MutableLiveData<Boolean>()
-    val showMainActionFab: LiveData<Boolean> = _showMainActionFab
+    private val _showMainActionFab = MutableLiveData<MainFabUiState>()
+    val showMainActionFab: LiveData<MainFabUiState> = _showMainActionFab
```
planarvoid: I'd also rename this field 👍

WordPress/src/main/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModel.kt
```diff
@@ -28,11 +31,16 @@ class WPMainActivityViewModel @Inject constructor() : ViewModel() {
     private val _isBottomSheetShowing = MutableLiveData<Event<Boolean>>()
     val isBottomSheetShowing: LiveData<Event<Boolean>> = _isBottomSheetShowing
 
+    private var mainFabTooltipDisabled = true
```
planarvoid: I'd rather not have this state here as a field. Maybe we can create a method `isMainFabTooltipEnabled()` instead? I think it's safe to always go to app prefs to get the value. Keeping this field makes us keep the state in 2 places and increases chance of bugs.

WordPress/src/main/res/layout/tooltip_above.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:orientation="vertical"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    android:gravity="center_horizontal">
+
+    <TextView
+        android:id="@+id/tooltip_message"
```
planarvoid: Could you maybe extract all the common properties into a style? These 4 layouts are very similar.

WordPress/src/test/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModelTest.kt
```diff
@@ -18,22 +20,25 @@ class WPMainActivityViewModelTest {
...
         viewModel.onPageChanged(true)
-        assertThat(viewModel.showMainActionFab.value).isEqualTo(true)
+        assertThat(viewModel.showMainActionFab.value?.isFabVisible).isEqualTo(true)
     }
 
     @Test
     fun `fab hidden when asked`() {
         viewModel.onPageChanged(false)
-        assertThat(viewModel.showMainActionFab.value).isEqualTo(false)
+        assertThat(viewModel.showMainActionFab.value?.isFabVisible).isEqualTo(false)
```
planarvoid: You can add 2 more tests here for the tooltip field :-)

WordPress/src/main/java/org/wordpress/android/ui/WPTooltipView.kt
```diff
@@ -0,0 +1,92 @@
...
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : LinearLayout(context, attrs, defStyleAttr) {
+    var position = LEFT
+    var messageId = 0
+
+    init {
+        attrs?.also {
+            val a = context.theme.obtainStyledAttributes(
```
planarvoid: maybe give this field a bit better name? for example `stylesAttributes`?

WordPress/src/main/java/org/wordpress/android/ui/WPTooltipView.kt
```diff
@@ -0,0 +1,92 @@
...
+            try {
+                position = TooltipPosition.fromInt(a.getInt(R.styleable.WPTooltipView_wpTooltipPosition, 0))
+                messageId = a.getResourceId(R.styleable.WPTooltipView_wpTooltipMessage, 0)
+            } finally {
+                a.recycle()
+            }
+        }
+
+        inflate(getContext(), position.layout, this)
+        val tvMessage = findViewById<TextView>(R.id.tooltip_message)
```
planarvoid: you can move this initialization into the `if (messageId > 0)` block

WordPress/src/main/java/org/wordpress/android/ui/WPTooltipView.kt
```diff
@@ -0,0 +1,92 @@
...
+
+    enum class TooltipPosition(val value: Int, @LayoutRes val layout: Int) {
+        LEFT(0, R.layout.tooltip_left),
+        RIGHT(1, R.layout.tooltip_right),
+        ABOVE(2, R.layout.tooltip_above),
+        BELOW(3, R.layout.tooltip_below);
+
+        companion object {
+            fun fromInt(value: Int): TooltipPosition =
+                    values().firstOrNull() { it.value == value }
```
planarvoid: You can remove the `()` here

osullivanchris:
FAO editorial team. We requested a copy review for this new tooltip (style/position might change slightly). We have introduced this new shortcut button, and want to highlight it the first time a user encounters it. Its quite hard to write without the possibility of ambiguity/misinterpretation. Here are some options considered and the flaws we noticed:

- Your new shortcut to create content [flaw: it is not a new shortcut if the user is new to the app]
- Add new content to your site [flaw: is it a blog or site?]
- Add new content [flaw: is it clear what it means to add 'content'?]
- Create a post or page [flaw: we might later add other things (e.g. media). Should it be generic (e.g. content) to cover all future use cases?)]

I think "Create a post or page" is quite descriptive of what it does.

[Clickthrough prototype of the feature here](https://automattic.invisionapp.com/share/A2V1SS0B8DR#/396965081_My_Site)

kristastevens:
Hello @osullivanchris, thanks for the ping! I agree wholeheartedly, that of the options you shared above, "Create a post or page" is the strongest option. It's clear, succinct, and uses simple, declarative, action-oriented language.

WordPress/src/main/java/org/wordpress/android/ui/WPTooltipView.kt
develric: Done here d5bb53872c114f53a917d58ef346a4406db1132e

WordPress/src/main/java/org/wordpress/android/ui/WPTooltipView.kt
develric: Yep. Changed here d5bb53872c114f53a917d58ef346a4406db1132e

WordPress/src/main/java/org/wordpress/android/ui/WPTooltipView.kt
develric: Auto completion issue 😅...changed here d5bb53872c114f53a917d58ef346a4406db1132e

WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
develric: Yep better naming...see here 95381cb8653731e40440a61be244ed2eeafd7664

WordPress/src/main/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModel.kt
develric: Correct 👍...see here 95381cb8653731e40440a61be244ed2eeafd7664

WordPress/src/main/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModel.kt
develric: The goal was as you guessed to not access the shared every time, but agree maybe it's not worth it in this case and we get the extra plus of having a single state point 👍...changed here 95381cb8653731e40440a61be244ed2eeafd7664

WordPress/src/main/res/layout/tooltip_above.xml
develric: Good point, done here 48a3e9960bde6726ac891445a9f5d58d24cdf0eb

WordPress/src/test/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModelTest.kt
develric: Correct 😊....see here 5511a9a541dc4c8e8bafcad7bc8ddc3a40f996a1

develric:
Hi @khaykov , thanks so much for the review 🙇‍♂️!

>  I think the visibility change of tooltip is a bit sudden...

Good point. I added a simple alpha animation (both in and out) and left the delay for the show so that the tooltip appearance is in sequence to that of the FAB. Hope it's better, changes here 200f85621d7c2d846eec5ffe4719369ff204518d

> I assume the following things are intentional, but...

Yep they are intentional and shared with Chris, but let's see if he wants to add/rethink something 😊.

If tomorrow it will be more convenient, I can maybe ask Jirka to follow up on this. But stay tuned, I can possibly still need your help 😎!

osullivanchris:
Thanks @khaykov those are good points to raise!

> Tapping on tooltip does not dismiss it. If I don't want to make a post, but also don't want a giant tooltip on my main screen then I'm out of luck.

I think this is a fair point. Perhaps we should allow dismissing the tooltip by tapping either the FAB or the tooltip. What do you think @develric ? The only downside I can see is that the user might mistakingly tap the tooltip wondering what it is, and not learn about the FAB button. But that seems edge-casey.

> If I navigate to Post screen, tap fab there and make a post I will still see the tooltip without any way to dismiss it.

Interesting point. I am not so worried about this one. The new feature allows the user to get a shortcut to create content from the My Site screen. And if they navigate through to the post screen they might not have made that learning. It may have been clearer when we used language in the tooltip that specifically called it a "shortcut". I think I had something like "Your shortcut to create content" which explains how its different to the FAB on the deeper level. In any case I don't feel that there's a big issue here to take action on it.

develric:
Hi @khaykov and @osullivanchris 👋! I pushed modifications here 22b615017506b12a958b7b4f75f26c102fa1e7e5 

This should:
1. add the tooltip dismiss on tap: this action will only hide the tooltip but not marked the tooltip announce as disabled; so the tooltip will be shown again after a page switch and still disabled only after using the main FAB.
2. reduced the hiding animation duration (to totally remove it was not good since the FAB is still animated)

@khaykov ready for another review pass, thanks 🙇‍♂️

PR #11059, opened 2020-01-08
Accessibility : Bar Chart Improvements

### LINKED ISSUE #10987: Accessibility : Stats Graph Improvement
This graph was made accessible in a previous version of its implementation. More details #10285

## Expected behavior
Each bar in the graph should be accessible and when focused should be able to tell the user something. 

## Actual behavior
The whole view the graph resides in is announced. 

1. Go Stats within the app. 
2. Select either Day or Weeks and you will see the graph. 
3. Attempt to interact with the columns and you will see that's not possible. 

jd-alexander:
Fixes #10987

## Findings
The bar chart component wasn't accessible so this improvement seeks to make it easier for visually impaired users to navigate its contents. 

## Solution

`ExploreByTouchHelper` was utilized to make the bars within the chart selectable which would then trigger the announcement of a bar's content description. 

Before | After 
--------|-------
[Before Video ](https://drive.google.com/file/d/19dDnYoiueb1S5NTwNTH8POlAQvb65-vH/view?usp=sharing) |   [After Video    ](https://drive.google.com/file/d/19m6iZKmCrKoBehhI0k3YAVGDHlpxbloB/view?usp=sharing)

## Testing
1. Turn on TalkBack
2. Open Stats → Days
3. Slide finger over each bar to hear the content description. 
4. Open other tabs and the bar chart will update. 
5. Tap to make a bar highlighted and the stats will be updated as well. 

Days - Visitors | Days - Visitors (Bar Selected) 
--------|-------
        |   

Days - Views | Days - Views (Bar Selected) 
--------|-------
        | 

Days - Likes | Week - Views 
--------|-------
        |       

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/granular/usecases/OverviewMapper.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/granular/usecases/OverviewMapper.kt
@@ -225,7 +225,7 @@ class OverviewMapper
             null
         }
 
-        val contentDescriptions = statsUtils.getEntryContentDescriptions(
+        val contentDescriptions = statsUtils.getBarChartEntryContentDescriptions(
                 entryType,
                 chartItems,
                 overlappingType,
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/granular/usecases/OverviewMapper.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/granular/usecases/OverviewMapper.kt
@@ -12,6 +12,10 @@ import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.Colum
 import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.Columns.Column
 import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ValueItem
 import org.wordpress.android.ui.stats.refresh.lists.sections.BlockListItem.ValueItem.State
+import org.wordpress.android.ui.stats.refresh.lists.sections.granular.usecases.OverviewMapper.SelectedType.Comments
+import org.wordpress.android.ui.stats.refresh.lists.sections.granular.usecases.OverviewMapper.SelectedType.Likes
+import org.wordpress.android.ui.stats.refresh.lists.sections.granular.usecases.OverviewMapper.SelectedType.Views
+import org.wordpress.android.ui.stats.refresh.lists.sections.granular.usecases.OverviewMapper.SelectedType.Visitors
 import org.wordpress.android.ui.stats.refresh.utils.ContentDescriptionHelper
 import org.wordpress.android.ui.stats.refresh.utils.MILLION
 import org.wordpress.android.ui.stats.refresh.utils.StatsDateFormatter
@@ -32,6 +36,17 @@ class OverviewMapper
             R.string.stats_likes,
             R.string.stats_comments
     )
+
+    enum class SelectedType(val value: Int) {
+        Views(0),
+        Visitors(1),
+        Likes(2),
+        Comments(3);
+
+        companion object {
+            fun valueOf(value: Int): SelectedType? = values().find { it.value == value }
+        }
+    }
 
     fun buildTitle(
         selectedItem: PeriodData,
@@ -99,11 +114,11 @@ class OverviewMapper
     private fun PeriodData.getValue(
         selectedPosition: Int
     ): Long? {
-        return when (selectedPosition) {
-            0 -> this.views
-            1 -> this.visitors
-            2 -> this.likes
-            3 -> this.comments
+        return when (SelectedType.valueOf(selectedPosition)) {
+            Views -> this.views
+            Visitors -> this.visitors
+            Likes -> this.likes
+            Comments -> this.comments
             else -> null
         }
     }
@@ -166,11 +181,11 @@ class OverviewMapper
         selectedItemPeriod: String
     ): List<BlockListItem> {
         val chartItems = dates.map {
-            val value = when (selectedType) {
-                0 -> it.views
-                1 -> it.visitors
-                2 -> it.likes
-                3 -> it.comments
+            val value = when (SelectedType.valueOf(selectedType)) {
+                Views -> it.views
+                Visitors -> it.visitors
+                Likes -> it.likes
+                Comments -> it.comments
                 else -> 0L
             }
             Bar(
@@ -197,10 +212,10 @@ class OverviewMapper
             result.add(ChartLegend(R.string.stats_visitors))
         }
 
-        val entryType = when (selectedType) {
-            1 -> R.string.stats_visitors
-            2 -> R.string.stats_likes
-            3 -> R.string.stats_comments
+        val entryType = when (SelectedType.valueOf(selectedType)) {
+            Visitors -> R.string.stats_visitors
+            Likes -> R.string.stats_likes
+            Comments -> R.string.stats_comments
             else -> R.string.stats_views
         }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/granular/usecases/OverviewMapper.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/granular/usecases/OverviewMapper.kt
@@ -198,11 +198,10 @@ class OverviewMapper
         }
 
         val entryType = when (selectedType) {
-            0 -> R.string.stats_views
             1 -> R.string.stats_visitors
             2 -> R.string.stats_likes
             3 -> R.string.stats_comments
-            else -> -1
+            else -> R.string.stats_views
         }
 
         val overlappingType = if (shouldShowVisitors) {
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/granular/usecases/OverviewMapper.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/granular/usecases/OverviewMapper.kt
@@ -202,7 +202,7 @@ class OverviewMapper
             1 -> R.string.stats_visitors
             2 -> R.string.stats_likes
             3 -> R.string.stats_comments
-            else -> 0
+            else -> -1
         }
 
         val overlappingType = if (shouldShowVisitors) {
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/granular/usecases/OverviewMapper.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/granular/usecases/OverviewMapper.kt
@@ -196,13 +196,36 @@ class OverviewMapper
         if (shouldShowVisitors) {
             result.add(ChartLegend(R.string.stats_visitors))
         }
+
+        val entryType = when (selectedType) {
+            0 -> R.string.stats_views
+            1 -> R.string.stats_visitors
+            2 -> R.string.stats_likes
+            3 -> R.string.stats_comments
+            else -> 0
+        }
+
+        val overlappingType = if (shouldShowVisitors) {
+            R.string.stats_visitors
+        } else {
+            null
+        }
+
+        val contentDescriptions = statsUtils.getEntryContentDescriptions(
+                entryType,
+                chartItems,
+                overlappingType,
+                overlappingItems
+        )
+
         result.add(
                 BarChartItem(
                         chartItems,
                         overlappingEntries = overlappingItems,
                         selectedItem = selectedItemPeriod,
                         onBarSelected = onBarSelected,
-                        onBarChartDrawn = onBarChartDrawn
+                        onBarChartDrawn = onBarChartDrawn,
+                        entryContentDescriptions = contentDescriptions
                 )
         )
         return result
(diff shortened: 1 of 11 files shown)


WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/utils/BarChartAccessibilityHelper.kt
```diff
@@ -0,0 +1,82 @@
...
+    private val barChart: BarChart,
+    private val contentDescriptions: List<String>,
+    private val hasOverlappingEntries: Boolean
+) : ExploreByTouchHelper(barChart) {
+    interface BarChartAccessibilityEvent {
+        fun onHighlight(index: Int, hasOverlappingEntries: Boolean)
+    }
+
+    private val dataSet = barChart.data.dataSets.first()
+    var accessibilityEvent: BarChartAccessibilityEvent? = null
```
malinajirka: Does this field need to be `var` and nullable? Could we move it into the constructor and change it to `val`?

WordPress/src/main/res/values/strings.xml
```diff
@@ -886,6 +885,8 @@
     <string name="stats_item_collapsed">Item collapsed</string>
     <string name="stats_posting_activity_graph">Graph of posting activity.</string>
     <string name="stats_list_item_description">%1$s: %2$s, %3$s: %4$s</string>
+    <string name="stats_bar_chart_accessibility_entry">%1$s, %2$d %3$s</string>
+    <string name="stats_bar_chart_accessibility_overlapping_entry">&#160; &amp; %1$d %2$s</string>
```
malinajirka: I'm not familiar with the code, could you please help me understand what is an "overlapping entry"? Can I somehow test this behavior?

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/utils/StatsUtils.kt
```diff
@@ -132,6 +134,36 @@ class StatsUtils
             number
         }
     }
+
+    fun getEntryContentDescriptions(
```
malinajirka: It seems this method is made for chart entries - we might want to consider renaming it so the intention/usage is more clear. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/BarChartViewHolder.kt
```diff
@@ -42,6 +46,24 @@ class BarChartViewHolder(parent: ViewGroup) : BlockListItemViewHolder(
...
+                        chart,
+                        contentDescriptions = item.entryContentDescriptions,
+                        hasOverlappingEntries = item.overlappingEntries != null
+                )
+
+                accessibilityHelper.accessibilityEvent = object : BarChartAccessibilityEvent {
+                    override fun onHighlight(index: Int, hasOverlappingEntries: Boolean) {
+                        chart.highlightColumn(index, hasOverlappingEntries)
+                        val entry = chart.data.dataSets.first().getEntryForIndex(index)
+                        val value = entry?.data as? String
```
malinajirka: Btw can't the caller pass us the value so we don't need to look it up and access the `chart.data.dataSet` from here?

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/granular/usecases/OverviewMapper.kt
```diff
@@ -196,13 +196,35 @@ class OverviewMapper
         if (shouldShowVisitors) {
             result.add(ChartLegend(R.string.stats_visitors))
         }
+
+        val entryType = when (selectedType) {
+            1 -> R.string.stats_visitors
```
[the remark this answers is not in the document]
jd-alexander: Thanks! That sounds like a good idea! I did this here because the same approach was being used above to determine the `selectedType` so I thought doing this would make it easier to spot the relation.

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/BarChartViewHolder.kt
```diff
@@ -42,6 +46,24 @@ class BarChartViewHolder(parent: ViewGroup) : BlockListItemViewHolder(
         GlobalScope.launch(Dispatchers.Main) {
             delay(50)
             chart.draw(item, labelStart, labelEnd)
+            chart.post {
```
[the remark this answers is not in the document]
jd-alexander: I invoked post here because it would initiate the helper when the chart is drawn and that would allow all the virtual views that are needed by the accessibility framework to be created. I am going to check to see if I can use another way to accomplish this that doesn't involve mixing threads which could cause issues or side effects.

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/BarChartViewHolder.kt
```diff
@@ -42,6 +46,24 @@ class BarChartViewHolder(parent: ViewGroup) : BlockListItemViewHolder(
...
+                accessibilityHelper = BarChartAccessibilityHelper(
+                        chart,
+                        contentDescriptions = item.entryContentDescriptions,
+                        hasOverlappingEntries = item.overlappingEntries != null
+                )
+
+                accessibilityHelper.accessibilityEvent = object : BarChartAccessibilityEvent {
+                    override fun onHighlight(index: Int, hasOverlappingEntries: Boolean) {
+                        chart.highlightColumn(index, hasOverlappingEntries)
+                        val entry = chart.data.dataSets.first().getEntryForIndex(index)
```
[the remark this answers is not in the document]
jd-alexander: Yes, I should add a comment indeed. I called first here because the chart will always have at least one dataset that does the default rendering i.e the values that the chart is trying to represent. Based on my exploration, it seems three datasets can get applied: main bars being represented, overlapping bars, and highlights and since the main set of values are always added first it will always be the first dataset present. 

Eg. 
main values = Views
overlapping values = Visitors
highlights = selected bar(s) 

I didn't see a unique identifier I could identify each dataset which would make the solution more direct so I opted for this approach.

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/BarChartViewHolder.kt
```diff
@@ -42,6 +46,24 @@ class BarChartViewHolder(parent: ViewGroup) : BlockListItemViewHolder(
...
+                        val value = entry?.data as? String
+                        item.onBarSelected?.invoke(value)
```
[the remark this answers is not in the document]
jd-alexander: No we don't. Thanks 👏

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/utils/BarChartAccessibilityHelper.kt
jd-alexander: Yes, this a good observation. Will do the refactor!

WordPress/src/main/res/values/strings.xml
jd-alexander: An overlapping entry is when two sets of stats are being represented on a single chart. So when you are able to see both "Views and Visitors" at the same time one of those values is overlapping the other. In this case, it would be the visitors that are overlapping entries. Let me know if this explanation helps!

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/utils/StatsUtils.kt
jd-alexander: Good observation! I will rename it!

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/BarChartViewHolder.kt
jd-alexander: Yes, I will move this to the constructor.

jkmassel:

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/granular/usecases/OverviewMapper.kt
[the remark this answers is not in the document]
malinajirka: Yeah, I saw it and I understand your decision. However, I thought introducing an enum might still be worth it and would make the code clearer.

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/BarChartViewHolder.kt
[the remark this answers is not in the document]
malinajirka: Thanks ;) - I'd just try the approach I suggested or anything else that comes from the top of your head -> if it doesn't work, don't worry about it.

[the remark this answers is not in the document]
jd-alexander: I tried the coroutines approach but it didn't work out well. My coroutine knowledge isn't as advanced so that could probably be the reason so, for now, I think it's okay to leave it once there are no obvious side effects in testing.

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/BarChartViewHolder.kt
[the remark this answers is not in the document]
jd-alexander: Resolved in 2b51423

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/utils/BarChartAccessibilityHelper.kt
jd-alexander: resolved in 4c473f4

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/utils/StatsUtils.kt
jd-alexander: resolved in 9b5f428

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/BarChartViewHolder.kt
jd-alexander: Resolved in 8bab45d dfbab1b

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/BarChartViewHolder.kt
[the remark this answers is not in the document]
malinajirka: 👍 Thanks for giving it a try! ;)

malinajirka:
Thanks @jd-alexander! 
I've reviewed the changes and they look great overall. I've left one comment.

I've also looked into [the issue](https://github.com/wordpress-mobile/WordPress-Android/pull/11059#pullrequestreview-341778905) I'm experiencing on my hw device. It took me a while, but I believe I found the culprit. We sometimes "cut" some entries/columns depending on the width of the screen - [here](https://github.com/wordpress-mobile/WordPress-Android/blob/dfbab1bb6b9a75014f1071e320d42ea20a67b5c0/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/BarChartViewHolder.kt#L302). However, we don't adjust the list of contentDescriptions (`item.entryContentDescriptions`) accordingly. Therefore the indices between graph-columns and contentDescriptions are off by n (in my case off by 1).

I was able to reproduce the bug on an emulator - Pixel 3, API 28. I assume you'll be able to reproduce it as well. If you won't be able to reproduce it, we can look into it together.

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/BarChartViewHolder.kt
jd-alexander: Resolved df98ba8

malinajirka:
I'm still experiencing the same offset issue as before even after the fix. Does the fix work for you?

I think the culprit is `take(..)`. `take` takes first n elements and [`cutEntries`](https://github.com/wordpress-mobile/WordPress-Android/blob/df98ba859acb4a36476f0f69a73158391c2b230f/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/BarChartViewHolder.kt#L305) takes last n elements.

It also feels a bit weird we cut the contentDescription in the helper even though the logic isn't related to it at all. We might want to consider passing only the relevant contentDescription to the helper (or write a comment with explanation at least). Wdyt?

jd-alexander:
> I'm still experiencing the same offset issue as before even after the fix. Does the fix work for you?

> I think the culprit is take(..) - it takes first n elements and cutEntries takes last n elements.

> It also feels a bit weird we cut the contentDescription in the helper even though the logic isn't related to it at all. We might want to consider passing only the relevant contentDescription to the helper (or write a comment with an explanation at least). Wdyt?

Thanks for a productive pair programming session @malinajirka! The fixes we came up with for these issues were really good! I just reviewed the code again and I think all is well now so this can be merged!

PR #10300, opened 2019-07-30
Add additional properties to the Block Editor activation events

daniloercoli:
Add additional properties to  `gutenberg_enabled` , and `gutenberg_disabled` analytics events:
The property `source` can now have the following values:
- via-site-settings
- on-block-post-open
- on-site-creation

Test 1:
- Flip the block editor setting in Site settings and make sure the property is added correctly

Test 2:
- Create a new site, and make sure the property is added with the correct value
- Tap on new post, and make sure nothing is tracked

Test 3:
- Clear your editor settings on the web
- Re-install the app
- open a block based post
- the popup should appear, and the correct property value bumped with analytics

cc @etoledom Since we need the iOS counterpart of this PR

Update release notes:

- [ x ] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

PR #11060, opened 2020-01-08
Set gutenbergEnabled to false if all sites use Aztec

mchowning:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1714

This fixes an issue that began occurring with the `user_info_gutenberg_enabled` boolean property to our analytics events. In particular, we switched from sending `false` if a user was only using Aztec to sending `null`. iOS did not make this change and continued sending `false` in that scenario (as did Android until recently). This PR fixes fixes that by setting the `gutenbergEnabled` metadata to be false if all of a user's sites use Aztec.

### Background

This issue occurs because we made a change in July to check to see whether the `gutenbergEnabled` metadata has been set and only set that property on analytics events [if it has been set](https://github.com/wordpress-mobile/WordPress-Android/pull/10300/files#diff-b746b023ad22dfe67f64acc153c207deR527-R531). At the same time, we also made a change to only update the metadata's `gutenbergEnabled` property to be `true` [if _at least one_ of the user's sites were using gutenberg](https://github.com/wordpress-mobile/WordPress-Android/pull/10300/files#diff-1f0db6b056d7a64bd9a2584db14f7006R119-R122). **We do not do anything, however, if none of the user's sites use gutenberg.** As a result, the `gutenbergEnabled` metadata property remained unset, and a `null` value was sent if the user used Aztec on all their sites.

With the changes in this PR, we now only send `null` if the user has no sites. Otherwise we send `true` if any of the user's sites use gutenberg (this behavior is unchanged) and we send `false` if all the user's sites use Aztec (before this change, this would have been sent as `null`). This should reverse the "swap" that occurred between `null` and `false` values in August 2019.

Screen Shot 2020-01-08 at 2 05 43 PM

#### Scenario 1: No Sites
1. Log into an account with no sites
2. Verify that the `user_info_gutenberg_enabled` property on the analytics events is set to `null`

#### Scenario 2: All sites using Aztec
1. Log into an account where all the associated sites use the Aztec editor
2. Verify that the `user_info_gutenberg_enabled` property on the analytics events is set to `false`

#### Scenario 3: At least one site using Gutenberg
1. Log into an account with at least one site using the gutenberg editor (i.e., log into the account used in test scenario 2, and change one site to use gutenberg by default).
2. Verify that the `user_info_gutenberg_enabled` property on the analytics events is set to `false`

#### PR submission checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/util/analytics/AnalyticsUtils.java
+++ b/WordPress/src/main/java/org/wordpress/android/util/analytics/AnalyticsUtils.java
@@ -9,6 +9,7 @@ import android.text.TextUtils;
 import android.webkit.MimeTypeMap;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.VisibleForTesting;
 
 import org.json.JSONException;
 import org.json.JSONObject;
@@ -37,6 +38,7 @@ import org.wordpress.android.util.VideoUtils;
 import java.io.File;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


mchowning:
:wave: @hypest ! You may want to give this a look since you and @daniloercoli were most closely connected to the earlier changes to this code.

mchowning:
Thanks for the review @cameronvoell ! 🙇 

>I wasn't able to figure out how to execute Scenario 1 from your testing steps. wordpress-mobile/gutenberg-mobile#1714 seems to focus on the wpandroid_editor_session_end event, but I don't think the WP Android app lets you start (or end) an editor session if you do not have any sites (conditions for Scenario 1), so I couldn't trigger this event to test. 

That is a tricky one to test for sure. :smile: The way I validated this was through both (a) checking manual logs I added to the app to verify the property value was `null`; and (b) checking tracks to insure that the events that were sent off (i.e., `wpandroid_application_closed`) did not have an unexpected gutenberg enabled value (so in this scenario, that the gutenberg value was `null`).

>  Is there another relevant event that would send user_info_gutenberg_enabled as null? If not it may be useful to know that no events should be sent with user_info_gutenberg_enabled as null when a user is running an app with this change going forward.

Events like `wpandroid_application_closed`, `wpandroid_application_opened`, and others do get the `gutenbergEnabled` property attached to them as well, and those events will be sent off even if an account has no sites associated with it (and therefore, the `gutenbergEnabled` property would be `null`.

cameronvoell:
> Events like `wpandroid_application_closed`, `wpandroid_application_opened`, and others do get the `gutenbergEnabled` property attached to them as well, and those events will be sent off even if an account has no sites associated with it (and therefore, the `gutenbergEnabled` property would be `null`.

Thanks for clarifying that Scenario for me. Ready to merge! :shipit:

PR #11062, opened 2020-01-09
Track gutenberg rollout for phase 2

maxme:
Waiting for #11029 and #11045 to be merged.


WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
```diff
@@ -103,6 +139,15 @@ public static void migrateAppWideMobileEditorPreferenceToRemote(final AccountSto
         }
     }
 
+    private static void trackGutenbergEnabledForNonGutenbergSites(final SiteStore siteStore,
+                                                                  final BlockEditorEnabledSource source) {
+        for (SiteModel site : siteStore.getSites()) {
+            if (!TextUtils.equals(site.getMobileEditor(), GB_EDITOR_NAME)) {
```
hypest: I know this PR is still marked as Draft but, went ahead and gave it a look anyway.

Won't the editor settings have changed already by the time we reach this code? If this is valid code, can you add a comment to explain how it works in conjunction with https://github.com/wordpress-mobile/WordPress-Android/pull/11062/files#diff-cb360ad2878e66ffe65d8c54967921d6R75 for example?

maxme: Hey @hypest, you're right, actually fixed in the other PR #11045 and now in this one - cc @mchowning

Issue #11013, opened 2019-12-23
IA create FAB: make more easily understandable difference between page/post


develric:
As we introduced the create post/page FAB in the My Site screen, we want to use some subheading in the bottom sheet to explain post/page difference and change title hint in the opened editor to "Add post title" and "Add page title" to make more easily understandable the action flow.

cc @osullivanchris

develric:
This is still current (tested with current develop 16.1-rc-1 on Pixel 3 Emu API 28), so leaving it opened but removing it from IA GH Project that is not currently active.

PR #11064, opened 2020-01-10
IA create content bottom sheet changes

develric:
Partial fix for #11013 

This PR does the following:

- Changes the copy of the actions in the main bottom sheet
- Adds a title to the bottom sheet

Below a comparison for before/after bottom sheet:

| Before | After |
|---|---|
| image | image |

## To test
- Compile and run `vanilla flavor`. Smoke test the app (switching between main sections in the bottom nav bar and using the My Site features). Check no evident abnormalities are triggered.

- Compile and run the `zalpha flavor`
- Select the FAB and check that a bottom sheet is opened and is similar to the After image above. The Editor is opened when selecting the Blog Post/Site Page action from the bottom sheet. 
- Check that you can create a new Post/Page and it is Published/Drafted
- Check that no action is taken tapping on the bottom sheet `Add new` title

NOTE: RELEASE NOTES not modified yet since still bihind feature flag.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/ActionListItemViewHolder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/ActionListItemViewHolder.kt
@@ -1,6 +1,7 @@
 package org.wordpress.android.ui.main
 
 import android.view.LayoutInflater
+import android.view.View
 import android.view.ViewGroup
 import android.widget.ImageView
 import android.widget.TextView
@@ -15,11 +16,24 @@ class ActionListItemViewHolder(
         val actionIcon: ImageView = this.itemView.findViewById(R.id.action_icon)
         val actionTitle: TextView = this.itemView.findViewById(R.id.action_title)
 
-        actionIcon.setImageResource(action.iconRes)
-        actionTitle.setText(action.labelRes)
+        if (action.iconRes > 0) {
+            actionIcon.setImageResource(action.iconRes)
+            actionIcon.visibility = View.VISIBLE
+        } else {
+            actionIcon.visibility = View.GONE
+        }
 
-        this.itemView.setOnClickListener {
-            action.onClickAction?.invoke(action.actionType)
+        if (action.labelRes > 0) {
+            actionTitle.setText(action.labelRes)
+            actionTitle.visibility = View.VISIBLE
+        } else {
+            actionTitle.visibility = View.GONE
+        }
+
+        action.onClickAction?.let {
+            this.itemView.setOnClickListener {
+                action.onClickAction.invoke(action.actionType)
+            }
         }
     }
 }
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/ActionListItemViewHolder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/ActionListItemViewHolder.kt
@@ -5,8 +5,10 @@ import android.view.View
 import android.view.ViewGroup
 import android.widget.ImageView
 import android.widget.TextView
+import androidx.core.widget.TextViewCompat
 import androidx.recyclerview.widget.RecyclerView.ViewHolder
 import org.wordpress.android.R
+import org.wordpress.android.ui.main.MainActionListItem.ActionType.NO_ACTION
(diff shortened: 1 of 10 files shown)


WordPress/src/main/java/org/wordpress/android/ui/main/ActionListItemViewHolder.kt
```diff
@@ -15,11 +16,24 @@ class ActionListItemViewHolder(
         val actionIcon: ImageView = this.itemView.findViewById(R.id.action_icon)
         val actionTitle: TextView = this.itemView.findViewById(R.id.action_title)
 
-        actionIcon.setImageResource(action.iconRes)
-        actionTitle.setText(action.labelRes)
+        if (action.iconRes > 0) {
+            actionIcon.setImageResource(action.iconRes)
```
malinajirka: This isn't exactly part of this PR, so feel free to ignore this comment.

We try to load all images, both local and remote, using `ImageManager` -> `imageManager.load(actionIcon, action.iconRes)`. It uses glide underneath and supports various types of optimisations/caches.

WordPress/src/main/java/org/wordpress/android/ui/main/MainActionListItem.kt
```diff
@@ -7,6 +7,7 @@ sealed class MainActionListItem {
     abstract val actionType: ActionType
 
     enum class ActionType {
+        NO_ACTION,
```
malinajirka: Out of curiosity: Have you considered creating a different child of the sealed class instead of reusing `CreateAction`. It seems we are losing some of the advantages of using a sealed class by setting the values to "null/0/NO_ACTION". Moreover, it feels a bit weird to have CreateAction which doesn't perform any actions. Wdyt?

osullivanchris:
@malinajirka designer here 👋haha

> Was the "Add new" title designed by a designer. Tbh it seems a bit "cheap" to me. I'd expect the text to be colored/bold/sized or something like that.

It was me! I followed the type styles that material suggests, and there was no medium/bold indicated. I'm using similar to these mocks they provide also
Screenshot 2020-01-10 at 10 47 33

However in my mocks we have a slightly more sophisticated bottom sheet with corners and drag handles. I agree looking at the build here it looks a little bit 'plain' before we have those details. So I think we should make the heading medium. Here's my original mock, and changed to medium now, for reference. Thanks for calling it out!
regular medium

> Is the overlay over the navigation bar (back, home,..) expected/desired?

that's a known issue which we're looking into!

malinajirka:
>designer here 👋haha
:D

@osullivanchris Reading my comment again, I didn't pick the best words. I'm sorry if it sounded a bit harsh. It wasn't my intention at all. 

>I'm using similar to these mocks they provide also

Looking at the mocks, the title color seems at least a bit different than the list item text color.

>So I think we should make the heading medium. Here's my original mock, and changed to medium now, for reference.

👍 Perfect, looks great now! Thanks!

osullivanchris:
@malinajirka  no worries at all, I thought it was funny! And actually kinda agree once I saw it. Glad you agree its better now :) And it will look better further in future when we improve the sheet.

WordPress/src/main/java/org/wordpress/android/ui/main/ActionListItemViewHolder.kt
develric: Thanks for sharing. Changed here 7ca16a944a46c9731a804b19c7e969e3ca239e3f

WordPress/src/main/java/org/wordpress/android/ui/main/MainActionListItem.kt
develric: yep, used this approach in the reader bottom sheet that's more complex one (with item divider, tags, sites type etc). I just tried to stay simple here and it worked fine when having only Actions 😊. I don't think this code is final anyway since I expect we will come to the bottom sheet as Chris say so if you agree for now I would keep this one, let me know 🙇‍♂️

develric:
Hi @jirka and @osullivanchris , I should have followed up to the items above. Ready for a second pass, thanks 🙇‍♂️!

PR #11071, opened 2020-01-11
Issue/10639 material site settings

### LINKED ISSUE #10639: Implement Material theme on Site Settings screen
We should switch widgets used at Site Settings screen to their material counterparts, remove color references where possible, and switch to using material `textAppearance` attributes.

Bunch of weirdness there, so we need to be careful.

khaykov:
Fixes #10639

This PR moves Site Settings to Material Theme and adds Dark Theme support.

Site Settings look pretty simple, but they are super-old, and there are a lot of scattered changes. Luckily it's pretty easy to confirm that everything looks ok.

There is a huge amount of PRs in this feature, that covers all the screens within the app, so I'm requesting reviews from a wide variety of people.

comparison_settings

To test:
- Ideally, navigate to Site Settings and confirm the visual and functional integrity of each element in both Light and Dark themes.
The theme can be changed from App Settings.

PR submission checklist:


khaykov:
@mattmiklic I found some extra code that applied custom hardcoded styling to preference category headers and removed it, so they will follow our style:
comparison_settings

malinajirka:
Thanks @khaykov.

I tried to review the code - there is a lot of changes so it was more of a scan than actual review.

I tested the app and it works as expected. However, I noticed the dialogs don't look as good as they used to in the landscape mode.
site-settings-dark
Is there anything we can do about it, wdyt?

khaykov:
@malinajirka Thanks for taking a look at this one! And sorry for the length of the PR - style changes tend to explode the PR quite a lot, even though not much is actually changed (and I shouldn't have done the AS suggested refactoring, in retrospect).

About the dialogs - looks like the activity had an `android:configChanges="orientation|screenSize"` flag in manifest, which messed up with the top and bottom inset of dialogs after orientation change. I set the inset manually, so they will fit more content and will look uniform in both portrait and landscape orientations. 

The input field hint text size is OK (it's default one) 👍

PR #11089, opened 2020-01-15
Create FAB tooltip announce improvements

### LINKED ISSUE #11088: Create FAB tooltip announce improvements
### Expected behavior
In the IA project we added a tooltip to announce the new create FAB feature. We want the tooltip to be dismissed forever both if user taps on the FAB or on the Tooltip. Also we want to add a persistent long-pressing toast bound to the FAB, showing same text of the tooltip.

### Actual behavior
Currently the tooltip is dismissed only if user taps on the FAB. Tapping on the tooltip temporarily dismiss the tooltip that appears again if users goes for example to Reader and comes back to My Site. Also the long-pressing toast is currently not present. 

1 - Tap on the tooltip and check it is dismissed
2 - Go to Reader, come back to My Site and check the tooltip is there again
3 - Long-press the fab anche check no toast is shown

develric:
Fixes #11088

| image | image |
|---|---|

### Test Case 1: FAB still dismisses the Tooltip forever
- Check that the tooltip is shown above the FAB (if not, clear the storage of the app and open it again)
- Tap on the FAB anche check the tooltip is dismissed
- Go to Reader, come back to My Site and check the FAB is shown but the tooltip is not shown any more
- Close and open the app again and check the FAB is shown but the tooltip is not shown any more

### Test Case 2: tooltip is dismissed forever on tap and we have the long press toast
- Clear the storage of the app, open the app again and check the tooltip is shown
- Tap on the tooltip and check it is dismissed
- Go to Reader, come back to My Site and check the FAB is shown but the tooltip is not shown any more
- Close and open the app again and check the FAB is shown but the tooltip is not shown any more
- Long-press the fab anche check that a toast is shown with the message `Create a post or page`

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModel.kt
@@ -82,6 +82,8 @@ class WPMainActivityViewModel @Inject constructor(private val appPrefsWrapper: A
     }
 
     fun onTooltipTapped() {
+        appPrefsWrapper.setMainFabTooltipDisabled(true)
+
         val oldState = _fabUiState.value
         oldState?.let {
             _fabUiState.value = MainFabUiState(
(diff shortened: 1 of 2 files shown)


WordPress/src/main/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModel.kt
```diff
@@ -82,6 +82,8 @@ class WPMainActivityViewModel @Inject constructor(private val appPrefsWrapper: A
     }
 
     fun onTooltipTapped() {
+        appPrefsWrapper.setMainFabTooltipDisabled(true)
```
malinajirka: I don't want to block this PR, however, I'd consider adding a simple unit test for this behavior. Wdyt?

PR #11096, opened 2020-01-16
Navigation Bar shadowed by bottom sheet

develric:
Fixes #11018 

This PR is based on the solution elaborated in #11051 (credits to @ashiagr , @theck13 and @malinajirka to having come to this 🙇‍♂️).

In the following the relevant screenshots (FYI cc @osullivanchris ).

| API Level/Device type | Portrait | Landscape |
|---|---|---|
| API 26/Phone | image | image |
| API 26/Tablet | image | image |
| API 28/Phone| image | image |
| API 28/Tablet | image | image |
| API 29/Phone | image | image |
| API 29/Tablet | image | image |

 

## To test
- Use an emulator for each of the above API level
- Open the bottom sheet tapping on the FAB
- Check that the system navigation bar is not shadowed both in portrait and landscape 

PR submission checklist:

- [x] I have considered adding unit tests where possible.

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #11107, opened 2020-01-17
Navigation Bar shadowed by bottom sheet

### LINKED ISSUE #11018: System navigation buttons are shadowed by bottom sheets.
### Expected behavior
System navigation buttons are expected to not being shadowed when bottom sheet is showing.

### Actual behavior
System navigation buttons are shadowed when showing bottom sheets (for example the create FAB and Reader filter bottom sheets shown below):

|  |  |
|---|---|

- With the alpha-200 zalpha release (Information Architecture modifications available) open the create FAB or the Reader filter bottom sheet
- Observe the shadowed system navigation buttons

develric:
Fixes #11018 

This PR is based on the solution elaborated in #11051 (credits to @ashiagr , @theck13 and @malinajirka to having come to this 🙇‍♂️).

In the following the relevant screenshots (FYI cc @osullivanchris ).

| API Level/Device type | Portrait | Landscape |
|---|---|---|
| API 26/Phone | image | image |
| API 26/Tablet | image | image |
| API 28/Phone| image | image |
| API 28/Tablet | image | image |
| API 29/Phone | image | image |
| API 29/Tablet | image | image |

 

## To test
- Use an emulator for each of the above API level
- Open the bottom sheet tapping on the FAB
- Check that the system navigation bar is not shadowed both in portrait and landscape 

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/MainBottomSheetFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/MainBottomSheetFragment.kt
@@ -24,6 +24,11 @@ class MainBottomSheetFragment : BottomSheetDialogFragment() {
     @Inject lateinit var viewModelFactory: ViewModelProvider.Factory
     private lateinit var viewModel: WPMainActivityViewModel
 
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setStyle(STYLE_NORMAL, R.style.LoginTheme_BottomSheetDialogStyle)
+    }
+
     override fun onCreateView(
         inflater: LayoutInflater,
         container: ViewGroup?,
--- a/WordPress/src/main/res/layout/add_content_bottom_sheet.xml
+++ b/WordPress/src/main/res/layout/add_content_bottom_sheet.xml
@@ -4,7 +4,8 @@
     android:orientation="vertical"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:animateLayoutChanges="true">
+    android:animateLayoutChanges="true"
+    android:fitsSystemWindows="true">
 
     <androidx.recyclerview.widget.RecyclerView
         android:id="@+id/create_actions_recycler_view"
--- a/WordPress/src/main/res/values-v27/styles.xml
+++ b/WordPress/src/main/res/values-v27/styles.xml
@@ -4,4 +4,11 @@
         <item name="android:navigationBarColor">@android:color/white</item>
         <item name="android:windowLightNavigationBar">true</item>
     </style>
+
+    <style name="LoginTheme.BottomSheetDialogStyle" parent="Theme.MaterialComponents.Light.BottomSheetDialog">
+        <item name="android:statusBarColor">@android:color/transparent</item>
+        <item name="android:navigationBarColor">?android:attr/colorBackground</item>
+        <item name="android:windowIsFloating">false</item>
+        <item name="bottomSheetStyle">@style/Widget.Design.BottomSheet.Modal</item>
+    </style>
 </resources>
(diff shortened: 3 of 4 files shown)


theck13:
I don't think we should settle for the full-width solution.  Restricting the bottom sheet to a maximum width is by design and has accessibility advantages as well.  I haven't seen another app using a full-width bottom sheet on a large screen (e.g. tablet in landscape orientation).  There are a few alternatives to leaving the bottom sheet full-width.
1. Make bottom sheet full-width _only for API 28 and large screen_.
2. Make style of navigation bar dark _only for API 28 and large screen_.
3. Find out why the navigation bar behavior is different on API 28 only.

osullivanchris:
I agree a max-width makes sense especially looking at tablet. There's a couple of visual changes we want to make to the sheets adding corner radius and a drag handle on both iOS and Android. So we could add max-width as one of the visual improvements to follow.

Here is the existing mock
styles

and here is how it could look on tablet
table1

Or I saw @iamthomasbishop do something where the sheet was not fixed to the bottom on tablet but had all rounded corners. Something like this
tablet2

When it comes to defining a max-width do you know of any keyline or guideline to follow @theck13 ? Here I've somewhat arbitrarily made it 64dp wider than the phone mock that I had. One thing @develric pointed out that **is** working in the current solution is that we've padded the content inside the sheet more, so that it lines up with the list items underneath

iamthomasbishop:
Looking good, pals! For background, what we ended up doing on Mobile Gutenberg for tablets is set a max-width on sheets, pretty much like Chris' last mockups. I believe the max-width was roughly 580, but for something like this (a quick list of options, the equivalent of an ActionSheet), it could be narrower.

> Or I saw @iamthomasbishop do something where the sheet was not fixed to the bottom on tablet but had all rounded corners. Something like this

I think we just ended up with the sheet flush against the edge of the screen, like your second to last mockup above @osullivanchris, but I don't dislike how the "floating" one looks. The main thing is if we go with the "flush" version, we want to leave a slight bit of safe area when the user has gesture-based system navigation enabled (the equivalent to the home indicator on iOS).

theck13:
> When it comes to defining a max-width do you know of any keyline or guideline to follow @theck13?

Unfortunately, I don't see anything in the Material guidelines for that specifically.  It looks like WordPress uses [`300dp`](https://github.com/wordpress-mobile/WordPress-Android/blob/9dbd9ed1da0f4cb2b29d26af1c59b06cdc32dacf/WordPress/src/main/res/values-sw600dp/dimens.xml#L10) for screens wider than `600dp` while Simperium (i.e. Simplenote) uses [`360dp`](https://github.com/Simperium/simperium-android/blob/4c4b361685d0ee06ea44cb9b55bc1c6162b7cbc3/Simperium/src/main/res/values-large/dimens.xml#L6) for large screens and [`600dp`](https://github.com/Simperium/simperium-android/blob/4c4b361685d0ee06ea44cb9b55bc1c6162b7cbc3/Simperium/src/main/res/values-large-land/dimens.xml#L6) for larges screens in landscape orientation.  I checked a couple apps I know have bottom sheets.  Even though they definitely have a maximum width, there doesn't appear to be a set standard even among Google apps.  See the screenshots and table below for illustration.

bottom_sheet_portrait

bottom_sheet_landscape

App|Portrait|Landscape
-|-|-
Drive|420dp|420dp
Photos|512dp|512dp
Simplenote|360dp|600dp

develric:
Thanks for all the contributions above 🙇‍♂️! Since @osullivanchris approved to proceed this PR to the hotfix with full-width bottom sheet (see this [comment](https://github.com/wordpress-mobile/WordPress-Android/pull/11107#pullrequestreview-345468150)) and having already discussed this with @theck13 on slack (thanks @theck13 for all the feedbacks 😊) I would try to finalize this PR asking @ashiagr if she can review it. Hi Ashita, since the code finally didn't change this is the same as the #11096 , so I hope it's a quick task for you. Let me know, thanks 🙇‍♂️.

PR #11113, opened 2020-01-20
IA Reader: adding scroll flags to reader toolbar.

### LINKED ISSUE #10885: Collapse reader toolbar when scrolling posts list up
With [Reader navigation revised](https://href.li/?https://github.com/wordpress-mobile/WordPress-Android/pull/10792) as part of the Information Architecture project, we introduced a 2 level filter capability (tabs filtering and subfilter component). This reduces the vertical spaces available for posts visualization.

### Expected behavior
We would like the toolbar to collapse when we scroll up the reader posts list. This is the same behavior we have already in Notifications screen.

### Actual behavior
Currently when scrolling up the reader posts list the toolbar is not collapsed and this makes the vertical space available to posts not optimal especially in the _Following_ tab where also the subfilter component is visible.

develric:
Fixes #10885 

With Reader navigation revised as part of the Information Architecture project, we introduced a 2 level filter capability (tabs filtering and subfilter component). This reduces the vertical space available for posts visualisation. So we want to make the toolbar collapsable with the same behavior as the Notifications screen toolbar. Please note that reader feature is available in zalpha flavor so you will need to use that one to test.

## To test
- Compile and run zalpha flavor
- go to reader and scroll down and check the toolbar collapses
- Scroll up and check the toolbar is visible again
- To ensure the previous behavior is still valid in vanilla flavor, compile and run vanilla
- Go to reader and check the toolbar does not collapse/uncollapse on scroll down/up.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
@@ -395,6 +395,9 @@ public class ReaderPostListFragment extends Fragment
         mViewModel = ViewModelProviders.of((FragmentActivity) getActivity(), mViewModelFactory)
                                        .get(ReaderPostListViewModel.class);
 
+        AppLog.d(T.READER, "DBEUG VIEWMODEL - " + mViewModel);
+
+
         if (BuildConfig.INFORMATION_ARCHITECTURE_AVAILABLE && mIsTopLevel) {
             mViewModel.getCurrentSubFilter().observe(this, subfilterListItem -> {
                 if (ReaderUtils.isFollowing(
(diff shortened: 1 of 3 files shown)


khaykov (approved):
Works as expected! The toolbar is getting hidden on a scroll in an empty search view, so I wanted to confirm if the intended behavior :)

[Image from Gyazo](https://gyazo.com/b862e8fe80cf1ab47d8af6e2dda8af9a)

develric:
> The toolbar is getting hidden on a scroll

Good catch @khaykov 😎! Sorry I missed it!

I think we want the Toolbar to collapse only when we are in the tabbed view (not in search or search result views). The need comes from the presence of the subfilter but I also think it's not good to limit the scroll only to the `Following` tab (where the subfilter is present) since it's not great to remove the collapse state while moving into different tabs. I already moved the logic related to the scroll flag into the View Model but maybe better to ping @osullivanchris to get a confirmation of my understanding before you go with your review 😊.

Hey Chris just ping me if you need a zalpha apk to check what's the effect of the above 👍.

osullivanchris:
Thanks for the ping @develric 

Yep good spot @khaykov we shouldn't have it in search. 

I also agree with @develric's decision to have the collapsing happen in all the tabs on the main reader screen. Its what I expected.

develric:
Thanks @osullivanchris for checking 🙇‍♂️!

Hey @khaykov , should be ready for your review now, thanks 😊.

khaykov (approved):
Works well! Thanks for the changes :)

PR #11117, opened 2020-01-20
[Gutenberg] Update to v1.21.0

etoledom:
Gutenberg-mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1788

Changes:
* Reduced padding around text: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1560
* Fixed intermittent crash that occurred with very long posts: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1560#issuecomment-573123343

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -3,7 +3,8 @@
 * Disable the option to remove a Jetpack site from site picker
 * Block Editor: Reduced padding around text on Rich Text based blocks.
 * Block Editor: Improved stability on very long posts.
- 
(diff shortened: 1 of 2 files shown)


maxme:
> I think we should add a line to 14.1 even if we're one release late: "We were so focused on improving the app that we forgot to tell you a new shortcode block was introduced in 14.0"

Actually shortcode was added after the 14.0 branch cut, so we're fine here. I'll update the release notes.

etoledom:
Thank you @maxme for the review and for adding the missing release notes 🙏

PR #11134, opened 2020-01-22
Avoiding update subfilter status while we are in search mode.

### LINKED ISSUE #10952: Reader search screen not correctly managed when rotating device
This is relevant to the Information Architecture zalpha flavor of the app.

### Expected behavior
While in the Reader, when selecting the search icon you should stay on the search functionality when rotating the device.

| Correct | Correct |
|---|---|
| image | image |

### Actual behavior
When in the Reader selecting the search icon and then rotating the device makes the background posts list visible.

| Correct | Wrong |
|---|---|
| image | image |

- Open the reader
- Select the Following tab
- Select the Search icon
- Rotate the device and see the posts list visible in the background

develric:
Fixes #10952 

### Vanilla
This is to confirm we have still the correct behavior in vanilla
- Open the reader
- Select the Followed Sites filter
- Select the Search icon
- Rotate the device and see the posts list is not visible in the background and you have something similar to below

| Correct | Correct |
|---|---|
| image | image |

- smoke test the search and reader

### Zalpha
This is to confirm we now have the correct behavior also in zalpha
- Open the reader
- Select the Following tab
- Select the Search icon
- Rotate the device and see the posts list is not visible in the background and you have something similar to below

| Correct | Correct |
|---|---|
| image | image |

- smoke test the search and reader

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
@@ -396,11 +396,11 @@ public class ReaderPostListFragment extends Fragment
 
         if (BuildConfig.INFORMATION_ARCHITECTURE_AVAILABLE && mIsTopLevel) {
             mViewModel.getCurrentSubFilter().observe(this, subfilterListItem -> {
-                if (ReaderUtils.isFollowing(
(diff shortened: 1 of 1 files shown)


khaykov (approved):
Confirmed the issue and that the fix is working!

Issue #11140, opened 2020-01-23
Introduce a new module for media editing


malinajirka:
Goal of this task is to introduce a skeleton/structure for the media editing module.

PR #11159, opened 2020-01-24
Issue/11140 introduce a new module for media editing

ashiagr:
This PR Introduces a new module for media editing #11140.

**Things to note**

- With focus on image editing, module is named `ImageEditor`.  We can add it as a sub module to a parent `MediaEditor` module when we support other media formats.
- Dependencies common to the module and main app are currently using same version to avoid duplicate files. Module dependencies will be later updated to newer versions when we extract the module to a separate repository.
- Kotlin directory is as kept as source root in android section (build.gradle) as source code will be written in Kotlin.
- Package name starts with `org.wordpress.android` considering we'll extract the library to [wordpress-mobile](https://github.com/wordpress-mobile) repo similar to other shared libraries.
- Added [Jetpack's navigaton-component](https://developer.android.com/guide/navigation) to help us follow [Principles of navigation](https://developer.android.com/guide/navigation/navigation-principles ).
- Added skeletal activity, fragment.

PR submission checklist:


ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/ImageEditor.kt
```diff
@@ -1,29 +1,8 @@
...
-    ) {
-        // TODO
-        // Temporarily goes to edit image activity
-        val intent = Intent(context, EditImageActivity::class.java)
-        intent.putExtra(MainImageFragment.ARG_MEDIA_CONTENT_URI, contentUri)
-
-        EditImageActivity.startIntent(context, intent)
-    }
-}
+@Singleton
```
malinajirka: I'm not sure what exactly will we put into this class, but I'm not sure we'll want to keep it as a singleton. I guess we can keep it for now, but we need to be careful about what we add.

ImageEditor/build.gradle
```diff
@@ -27,30 +29,40 @@ android {
...
-        release.java.srcDirs += 'src/release/kotlin'
-        debug.java.srcDirs += 'src/debug/kotlin'
         test.java.srcDirs += 'src/test/kotlin'
         androidTest.java.srcDirs += 'src/androidTest/kotlin'
     }
 }
 
 dependencies {
-    def nav_version = "2.0.0"
+    def navComponentVersion = "2.0.0"
```
malinajirka: We might want to consider using the `ext { .. }` approach which here (we use in the other modules). Wdyt?

ImageEditor/build.gradle
```diff
@@ -27,30 +29,40 @@ android {
...
     implementation fileTree(dir: 'libs', include: ['*.jar'])
+    // Kotlin
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation 'androidx.appcompat:appcompat:1.0.2'
 
-    implementation 'com.github.bumptech.glide:glide:4.10.0'
-    implementation ('com.github.yalantis:ucrop:2.2.2') {
-        exclude group: 'com.squareup.okhttp3'
-    }
+    // Dagger
```
malinajirka: If we ever plan to move it to another repo and use it as a library I think adding unnecessary external dependencies, such as Dagger, might discourage some potential users. Wdyt?

ImageEditor/build.gradle
ashiagr: I considered using `def` here as navigation component was used just in this module. I find a good explanation [here](https://stackoverflow.com/a/56969539/193545).
> it lets you click on the variable and points exactly where it is defined in the file 

But for consistency, I'll convert it to `ext { .. }` approach.

malinajirka:
@loremattei @jkmassel Will we need to update the script for localization (or anything else)?

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/EditImageActivity.kt
```diff
@@ -1,75 +1,35 @@
...
         super.onCreate(savedInstanceState)
 
         setContentView(R.layout.edit_image_activity)
 
-        if (savedInstanceState != null) {
-            contentUri = savedInstanceState.getString(MainImageFragment.ARG_MEDIA_CONTENT_URI)
-        } else {
-            contentUri = intent.getStringExtra(MainImageFragment.ARG_MEDIA_CONTENT_URI)
-        }
+        hostFragment = supportFragmentManager
```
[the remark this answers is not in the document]
ashiagr: Done [25fd346](https://github.com/wordpress-mobile/WordPress-Android/pull/11159/commits/25fd3467f708f2488a263889b3b9753468b0a04b)

ImageEditor/build.gradle
ashiagr: Changed here 386b270

ImageEditor/build.gradle
ashiagr: Removed here af5488a

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/ImageEditor.kt
ashiagr: Removed unused class here af5488a

ImageEditor/build.gradle
malinajirka: 👍 Thanks, I think it's a good call. If we keep consistency, finding the declaration shouldn't be an issue.

malinajirka:
LGTM, thanks! 

We need to make sure localization works before we merge the master branch into develop. But this PR is good to be merged now ;). 🚢

jkmassel:
👋 Howdy @malinajirka – I think you're right – this probably needs some localization attention. If it hasn't been handled yet, @oguzkocer and I can take a look at it as part of this release cycle. 

Sorry about the delay here!

PR #11170, opened 2020-01-27
Merge 14.0 final into develop

jkmassel:
Includes updated metadata translations and a couple of last-minute beta fixes


tools/release-checks.sh
```diff
@@ -74,7 +74,7 @@ function printVersion() {
   echo "$BUILDFILE version $gradle_version"
 }
 
-checkNewLanguages
+#checkNewLanguages
```
maxme: I think we should re-enable it. This check is useful and actually it found that the app is translated in Urdu but we're not bundling it. Urdu is supported in Android apps (but not in the Play Store).

maxme: BTW, the `checkENStrings` function took ±45mins on my laptop. I think we should revisit that one.

jkmassel: I'm hoping to rebuild this check during this sprint as part of our Fastlane tooling – unfortunately, if there's a failure (in this case, the Urdu one), the script will never finish`checkENStrings` (at least on my machine – and I've confirmed this across multiple checkouts of the repo).

Because `checkENStrings` already takes as long as it does, it's very difficult to know whether the build is stuck permanently, or just taking it's usual slow amount of time, but I've had this script run for >3 hours before when there's a new language available (I cancelled it at that point). I _think_ it's something to do with emitting errors in this script but I'm not entirely certain. 

The task to fix this script has been in our backlog for a while, but because it keeps biting us (and should be a pretty quick change), I'm ok running without this safety check briefly while we sort out the issue.

maxme: I'd probably drop `checkENStrings` and keep `checkNewLanguages` in that case. Anyway, It's fine if you're dropping this and add it to fastlane.

I'm not sure if it's in your backlog, but we should open a ticket to add Urdu to a future release.

jkmassel: Totally agree – I was thinking that would be a good test case for the new scripting, then I can open the PR for it right after, if that works for you? That would put it in `14.2`.

maxme: Yep sure, I'll merge this one, ping me on the follow up PR.

PR #11179, opened 2020-01-28
Initiate upload starter on Pull-to-refresh and opening post list

### LINKED ISSUE #11130: Run UploadStarter when user opens or pulls-to-refresh PageList
We want to run the UploadStarter when the user
- [x] Pulls-to-refresh ([see implementation on Post List](https://github.com/wordpress-mobile/WordPress-Android/blob/27489cf1a114916ee7ab56f06199172906686077/WordPress/src/main/java/org/wordpress/android/viewmodel/posts/PostListViewModel.kt#L312))
- [x] Open Page List ([see implementation on Post List](https://github.com/wordpress-mobile/WordPress-Android/blob/3167057339349b76bf2571db6917777d8e29f17c/WordPress/src/main/java/org/wordpress/android/ui/posts/PostListMainViewModel.kt#L255))

malinajirka:
Fixes #11130 

Initiates the UploadStarter whenever the user opens the page list screen or when they pull to refresh.

To test:

Ideally comment out [this line](https://github.com/wordpress-mobile/WordPress-Android/blob/a89450d78b4052fc355b455f91498accdc38ecdb/WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadStarter.kt#L93
) so the upload starter isn't initiated when the device is back online
1. Open Pages list
2. Turn on the airplane mode
3. Create a new page with title "1" and publish it
4. Go to "Published tab" and find the page (it should be one of the top items)
5. Turn off the airplane mode
6. Make sure the upload isn't initiated automatically (shouldn't happen if you commented out the line)
7. Pull to refresh and notice the page is uploading - notification is shown
----------------
1. Open Pages list
2. Turn on the airplane mode
3. Create a new page with title "2" and publish it
4. Go to "Published tab" and find the page (it should be one of the top items)
5. Turn off the airplane mode
6. Make sure the upload isn't initiated automatically (shouldn't happen if you commented out the line)
7. Press back
8. Open the pages list again and notice the page is uploading - notification is shown

PR submission checklist:

### DIFF
--- a/WordPress/src/test/java/org/wordpress/android/viewmodel/pages/PagesViewModelTest.kt
+++ b/WordPress/src/test/java/org/wordpress/android/viewmodel/pages/PagesViewModelTest.kt
@@ -3,6 +3,8 @@ package org.wordpress.android.viewmodel.pages
 import androidx.arch.core.executor.testing.InstantTaskExecutorRule
 import com.nhaarman.mockitokotlin2.any
 import com.nhaarman.mockitokotlin2.mock
+import com.nhaarman.mockitokotlin2.times
+import com.nhaarman.mockitokotlin2.verify
 import com.nhaarman.mockitokotlin2.whenever
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.ExperimentalCoroutinesApi
@@ -25,6 +27,7 @@ import org.wordpress.android.fluxc.store.PostStore.OnPostChanged
 import org.wordpress.android.fluxc.store.PostStore.OnPostUploaded
 import org.wordpress.android.test
 import org.wordpress.android.ui.uploads.PostEvents
+import org.wordpress.android.ui.uploads.UploadStarter
 import org.wordpress.android.util.NetworkUtilsWrapper
 import org.wordpress.android.viewmodel.pages.PageListViewModel.PageListState
 import org.wordpress.android.viewmodel.pages.PageListViewModel.PageListState.DONE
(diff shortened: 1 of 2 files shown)


WordPress/src/test/java/org/wordpress/android/viewmodel/pages/PagesViewModelTest.kt
```diff
@@ -183,6 +188,26 @@ class PagesViewModelTest {
         assertThat(viewModel.arePageActionsEnabled).isTrue()
     }
 
+    @Test
+    fun `Auto-upload is initiated when the user enters the screen`() {
+        // Given
+        // When
+        viewModel.start(site)
+        // Then
+        verify(uploadStarter).queueUploadFromSite(site)
```
jd-alexander: I noticed was when I run this test it passes but there is an exception because the page store is being triggered and it's not set up so we could call `setUpPageStoreWithEmptyPages()` in both tests so the exception is no longer shown. Wdyt?

WordPress/src/test/java/org/wordpress/android/viewmodel/pages/PagesViewModelTest.kt
```diff
@@ -183,6 +188,26 @@ class PagesViewModelTest {
...
+    }
+
+    @Test
+    fun `Auto-upload is initiated when the user pulls to refresh`() {
+        // Given
```
jd-alexander: Same as above 👍

WordPress/src/test/java/org/wordpress/android/viewmodel/pages/PagesViewModelTest.kt
malinajirka: Ohh, I missed that. Great idea! Thanks!
Fixed in 143453475e547212b89d1991f7ecf1cca98e2bec

WordPress/src/test/java/org/wordpress/android/viewmodel/pages/PagesViewModelTest.kt
malinajirka: Fixed in 143453475e547212b89d1991f7ecf1cca98e2bec

malinajirka:
Thanks for the review @jd-alexander!

I've fixed the tests ;). It's ready for another round.

Issue #10962, opened 2019-12-16
Information Architecture modification -> Vision Accessibility audit


develric:
We need to Audit all the modifications introduced with IA project to ensure we correctly support Vision Accessibility.

Reporting below some note by @shiki as a starting point:

> - Reader: When the filter control is selected, TalkBack just says the filter name (e.g. “All sites”). It doesn’t say that it’s a button or that it’s a filter. It’s not obvious what happens after tapping, basically. It might be better for TalkBack to say something like “Filter, , Button”.
> - Reader: The Settings button is announced as “Unlabeled, Button”.
> - Reader → Filter List: The Sites and Tags are not marked as headings. A heading attribute is useful if you’re [navigating by heading ](https://href.li/?https://support.google.com/accessibility/android/answer/6006598?hl=en)with TalkBack.
> - Reader → Filter List: The currently selected filter is not announced by TalkBack as “selected”.
> - My Site: The “Stats”, “Pages”, “Posts”, and “Media” buttons are spoken as “Unlabeled button”.
> - My Site: The “Stats”, “Pages”, “Posts”, and “Media” labels are unnecessarily accessible. Since the buttons (icons) are accessible, these don’t need to be.

Issue #10887, opened 2019-12-02
Implement the newly designed bottom sheet component


develric:
As part of the Information Architecture project we introduced two bottom sheets. One for the [create new post/page](https://href.li/?https://github.com/wordpress-mobile/WordPress-Android/pull/10720) and the other for the [Reader navigation subfiltering](https://href.li/?https://github.com/wordpress-mobile/WordPress-Android/pull/10792).

We need to replace this basic implementation of the subfilter with a new component as defined in the design work in `pauD4L-vh-p2`

Note: when start working on this item review the above description to get rid of the internal document reference.

develric:
The design in particular for the Reader navigation subfiltering is further evolved. Adding below images for reference (for more details see `pauD4L-B3-p2`)

![image](https://user-images.githubusercontent.com/47797566/73141969-e3840a00-4089-11ea-8c8b-b798625f2c5b.png)

develric:
Closed by #11183, #11184 and #11193

PR #11184, opened 2020-01-29
New design reader bottom sheet - step 2

### LINKED ISSUE #11009: IA Reader filter bottom sheet: manage untitled sites visualization and selection
In the filter bottom sheet we introduced in the IA reader modifications we present the sites with their name and place `(untitled)` label as placeholder when the site has no name.

We need to change this behavior adding at least the site URL and (eventually) associated image.

develric:
Fixes #11009 and partially #10887

**NOTE: this requires PR #11183 merged first into develop**

| image | image |
|---|---|

## Notes
The feature is active in the zalpha flavor.

This PR covers the following:
1. Modifies the `SITES` section to present the list items with both the site title and a URL. In this way it's easier to manage cases where the site title is empty or equal since the match and selection is done on both.
2. Fixes a possible issue with the recycler view recycling (see [this comment ](https://github.com/wordpress-mobile/WordPress-Android/pull/10986#issuecomment-570216861)for more context)

### Test case 1: SITES tab shows site title and URL
- Follow some sites and tags if not already 
- Open the filter bottom sheet
- Check that the list of SITES shows both site title and url
- Check that the list of TAGS shows only tag title
- Smoke test the two lists checking the correct filter is selected and correct posts are loaded

### Test case 2: SITES tab shows and select Untitled sites correctly
- Follow at least a couple of sites with empty title (maybe not the only way to do this but you can set it empty from the web page settings)
- Open the bottom sheet and select one of the untitled site
- check that the correct posts load in the reader
- open the bottom sheet again and check that only the desired untitled site is highlighted as selected whereas the other one is not selected
- switch between the two untitled sites and check posts are correctly loaded
- smoke test selecting other sites and tags
### Test case 3
Since we made modifications to the `add_content_bottom_sheet` that is used also in the main create fab bottom sheet, makes sense to tap on the main FAB in My Site and smoke test the bottom sheet.

PR submission checklist:
note: not adding accessibility here since there will be a dedicated audit of the feature to close ticket #10962

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/viewholders/SiteViewHolder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/viewholders/SiteViewHolder.kt
@@ -1,18 +1,32 @@
 package org.wordpress.android.ui.reader.subfilter.viewholders
 
+import android.view.View
 import android.view.ViewGroup
 import android.widget.TextView
 import org.wordpress.android.R
 import org.wordpress.android.ui.reader.subfilter.SubfilterListItem.Site
 import org.wordpress.android.ui.utils.UiHelpers
+import org.wordpress.android.util.UrlUtils
 
 class SiteViewHolder(
     parent: ViewGroup
 ) : SubfilterListItemViewHolder(parent, R.layout.subfilter_list_item) {
     private val itemTitle = itemView.findViewById<TextView>(R.id.item_title)
+    private val itemUrl = itemView.findViewById<TextView>(R.id.item_url)
 
     fun bind(site: Site, uiHelpers: UiHelpers) {
         super.bind(site, uiHelpers)
         this.itemTitle.text = uiHelpers.getTextOfUiString(parent.context, site.label)
+        this.itemUrl.visibility = View.VISIBLE
+
+        val blog = site.blog
+
+        this.itemUrl.text = if (blog.hasUrl()) {
+            UrlUtils.getHost(blog.url)
+        } else if (blog.hasFeedUrl()) {
+            UrlUtils.getHost(blog.feedUrl)
+        } else {
+            ""
+        }
     }
 }
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/viewholders/SiteViewHolder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/viewholders/SiteViewHolder.kt
@@ -21,12 +21,10 @@ class SiteViewHolder(
 
         val blog = site.blog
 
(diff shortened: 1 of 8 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/viewholders/SubfilterListItemViewHolder.kt
```diff
@@ -23,9 +24,13 @@ open class SubfilterListItemViewHolder(
         }
         val itemText: TextView? = this.itemView.findViewById(R.id.item_title)
         itemText?.let {
-            if (filter.isSelected) {
-                it.setTextColor(ContextCompat.getColor(parent.context, R.color.primary))
-            }
+            it.setTextColor(ContextCompat.getColor(
+                    parent.context,
+                    if (filter.isSelected)
```
khaykov: This is super minor, but can we add curly brackets to if/else here? It could prevent some silly issues in the future :)

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/viewholders/SiteViewHolder.kt
```diff
@@ -1,18 +1,32 @@
...
+    private val itemUrl = itemView.findViewById<TextView>(R.id.item_url)
 
     fun bind(site: Site, uiHelpers: UiHelpers) {
         super.bind(site, uiHelpers)
         this.itemTitle.text = uiHelpers.getTextOfUiString(parent.context, site.label)
+        this.itemUrl.visibility = View.VISIBLE
+
+        val blog = site.blog
+
+        this.itemUrl.text = if (blog.hasUrl()) {
```
khaykov: Maybe we can switch to `when` for better readability here?

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/viewholders/SubfilterListItemViewHolder.kt
develric: 👍changed here f97bbf90c05ca39e67825776a94c71d15142de1c

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/viewholders/SiteViewHolder.kt
develric: TIL! I was actually not aware of this when without parameter form 😬, thanks 😊! Changed here f97bbf90c05ca39e67825776a94c71d15142de1c

develric:
Hey @khaykov ! Thanks for looking into this one 🙇‍♂️! I made the changes based on your comments. Let me know if it's ok 😊.

PR #11193, opened 2020-01-30
New design reader bottom sheet - step3

### LINKED ISSUE #11016: IA Reader: filter bottom sheet sites/tags lists special cases.
In the IA reviewed reader filter bottom sheet, we need to better define what to present to the user in cases where Sites/Tags lists can be empty.

Examples of this are brand new users (no tags present) or users who does not follow anything (All sites label currently available but no posts can be shown, and Tags list empty).

cc @osullivanchris

develric:
Fixes #11016 

**NOTE: this requires PR #11183 and #11184  merged first into develop**

| image | image |
|---|---|

## Notes
- The feature is active in the zalpha flavor.
- While working on this one found an issue that can happen mostly with sites which number of subscribers can vary fastly (like for example `longreads.com`). Since the method `ReaderBlog.isSame`compares also number of subscribers by default, it could happen that the subfilter in the bottom sheet could remain not selected even if it was the current selected subfilter (since a changing in the number of subscribers would made the `isSame` test to fail). Introduced an `isSame`method that allows comparison skipping number of subscribers.
- Moved the logic to refresh/load the subfilters from the bottomsheet tabs fragments (`SubfilterPageFragment`) to the root bottomsheet fragment (`SubfilterBottomSheetFragment`) since it seems a more suitable place.

## Test case 1: Empty state not visible when following sites and tag
- Follow some sites and tags if not already
- Open the filter bottom sheet and check both `SITES`and `TAGS` shoes the list of followed tags and sites and not the empty states

## Test case 2: Empty state visible for SITES when no sites are followed
- Follow some sites and tags if not already
- Open the filter bottom sheet and check both `SITES`and `TAGS` shoes the list of followed sites and tags and not the empty states
- Unfollow all sites and check that the `SITES` tab shows the empty state as per above image
- Tap on the `FOLLOW A SITE` cta and check that the followed sites management tab gets opened

## Test case 3: Empty state visible for TAGS when no tags are followed
- Follow some sites and tags if not already
- Open the filter bottom sheet and check both `SITES`and `TAGS` shoes the list of followed sites and tags  and not the empty states
- Unfollow all tags and check that the `TAGS` tab shows the empty state as per above image
- Tap on the `ADD A TAG` cta and check that the followed tags management tab gets opened

## Test case 4
Since we made modifications to the add_content_bottom_sheet that is used also in the main create fab bottom sheet, makes sense to tap on the main FAB in My Site and smoke test the bottom sheet.

## Test case 5
Compile and run vanilla flavor, smoke test the standard reader and the main create fab bottom sheet.

PR submission checklist:
note: not adding accessibility here since there will be a dedicated audit of the feature to close ticket #10962


develric:
> type size looks a little small - can you confirm you're using the text button component for this?

You have good eyes @osullivanchris 😊! You are right I was using a simple TextView actually like the `SWITCH SITE` one in `My Site`. Changed to Borderless Button and the result is as below. Even if I need to say it looks pretty similar dimension-wise (using `style="@style/Widget.AppCompat.Button.Borderless"` that I think defaults text size to 14sp (the size I was using for the TextView also), eventually we can increase it). In any case makes sense to use the button and I will change that one once we decide on the styling/size.

| Before | After |
|---|---|
| image | image |

osullivanchris:
Looks pretty much the same to me but semantically more correct to use a button so lets do it!

develric:
Hi @osullivanchris, changed the `TextView`into a `Button`. Here below how it appears for final reference:

<p align="center">
  
</p>

osullivanchris:
LGTM!

WordPress/src/main/java/org/wordpress/android/ui/reader/SubfilterBottomSheetFragment.kt
```diff
@@ -46,6 +55,9 @@ class SubfilterBottomSheetFragment : BottomSheetDialogFragment() {
 
         viewModel = ViewModelProviders.of(requireActivity(), viewModelFactory).get(ReaderPostListViewModel::class.java)
 
+        performUpdate()
```
malinajirka: We might want to consider invoking something like `viewModel.onSubFilterBottomSheetShown()` and moving both the lines there. It's technically a business logic. Wdyt?

Side question: Do we want to call it whenever the view is created (even after rotation?).

WordPress/src/main/java/org/wordpress/android/ui/reader/SubfilterBottomSheetFragment.kt
```diff
@@ -89,4 +101,41 @@ class SubfilterBottomSheetFragment : BottomSheetDialogFragment() {
         super.onCancel(dialog)
         viewModel.setIsBottomSheetShowing(false)
     }
+
+    @Subscribe(threadMode = ThreadMode.MAIN)
+    fun onEventMainThread(event: ReaderEvents.FollowedTagsChanged) {
```
malinajirka: We might want to consider moving all these methods into a VM. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterListItem.kt
```diff
@@ -23,7 +23,7 @@ sealed class SubfilterListItem(val type: ItemType) {
         return if (type == otherItem.type) {
             when (type) {
                 SECTION_TITLE -> label == otherItem.label
-                SITE -> (this as Site).blog.isSameAs((otherItem as Site).blog)
+                SITE -> (this as Site).blog.isSameAs((otherItem as Site).blog, false)
```
malinajirka: Do we want to find out:
a) the two objects are referencing the same blog
b) the two objects are equals

If the answer is `a`, can't we just compare their ids?
Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterPageFragment.kt
```diff
@@ -87,41 +80,39 @@ class SubfilterPageFragment : DaggerFragment() {
...
-    private fun performUpdate() {
-        performUpdate(EnumSet.of(
-                UpdateTask.TAGS,
-                UpdateTask.FOLLOWED_BLOGS
-        ))
-    }
-
-    private fun performUpdate(tasks: EnumSet<UpdateTask>) {
-        if (!NetworkUtils.isNetworkAvailable(activity)) {
+    private fun manageEmptyView(isEmpty: Boolean, category: SubfilterCategory) {
```
malinajirka: This all seems like a new business logic. We might want to consider moving it into a VM and writing unit tests for it. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
```diff
@@ -281,6 +284,11 @@ class ReaderPostListViewModel @Inject constructor(
         _filtersMatchCount.postValue(currentValue)
     }
 
+    fun onBottomSheetActionClicked(selectedTabIndex: Int) {
+        _isBottomSheetShowing.postValue(Event(false))
```
malinajirka: This might be a personal preference, but "isBottomSheetShowing" evokes me that the LiveData hold the current state. However, it seems we are sending an Event there, so they are not holding the current state, but they emit an event when we want to show/hide the bottom sheet.
I'd consider renaming it to something like `changeBottomSheetVisibility` or similar. Wdyt?

WordPress/src/main/res/layout/add_content_bottom_sheet.xml
```diff
@@ -1,19 +1,52 @@
...
+        android:layout_height="wrap_content"
         android:layout_width="match_parent"
         android:clipToPadding="false"
         android:descendantFocusability="beforeDescendants"
         android:scrollbars="vertical"
         android:layout_marginTop="@dimen/margin_extra_large"
         android:layout_marginBottom="@dimen/margin_extra_large"/>
 
-</LinearLayout>
+    <LinearLayout
```
malinajirka: I'm not sure if it's applicable, but have you considered using `ActionableEmptyView`?

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -450,6 +450,14 @@ public void onCreate(Bundle savedInstanceState) {
                     return null;
                 });
             });
+
```
malinajirka: NitPick: Could we please remove the extra empty line :D.

WordPress/src/main/java/org/wordpress/android/ui/reader/SubfilterBottomSheetFragment.kt
develric: Correct 😊, also agree with the side question, I don't think we should trigger the update on configuration changed or similar. Changed here feef4bceedb37674d9b5ef01706059c0837e6f9a

WordPress/src/main/java/org/wordpress/android/ui/reader/SubfilterBottomSheetFragment.kt
develric: yep, same refactor commit as above here feef4bceedb37674d9b5ef01706059c0837e6f9a

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterListItem.kt
develric: so , we want `a`. This said at a gut level I like more the multifield match 😊. Jokes apart at a logical level I agree the ids should be enough, not have enough context to say if that is really unique among blogs/feeds. I think the multifield comparison should be safe but I'm sure you have a deeper understanding of the overall architecture/infrastructure so I'm happy to change if you think it's more readable/efficient (and works of course 😬).

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterPageFragment.kt
develric: Good point! Changed here 2930454a5828ace28737502fedfc77dbd862914f 👍!

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
develric: 👍Changes here 19fb6dd30d5831e4162b311fbf9d6680913fbf87

WordPress/src/main/res/layout/add_content_bottom_sheet.xml
develric: My first implementation was using the `ActionableEmptyView` but @osullivanchris thought a `lighter` at least button and no image was fitting better in the context of the bottom sheet. Probably we could use the `ActionableEmptyView` styling in the wanted way the button of the view but for now I stayed with the simple layout as you have seen. Let me know if you feel we should change that in this PR 🙇‍♂️.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
develric: Right 😊. Removed here 0cdd939e4f1c9acff807ef2728f259f5ffcfc281

develric:
Hey @malinajirka 👋! Should be ready for a second pass! Let me know, thanks 😊.

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterListItem.kt
malinajirka: I don't have enough context either. It just feels very weird to be comparing equality when we want to mark a blog as selected if it matches the selected blog. But if you copied it from the other part of the reader, let's just leave it as it is.

SideNote: Even more weird is that the method is called `isSameAs` even though it's not `isSame` comparison but `isEquals` comparison :D. But I understand you didn't introduce it.

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
```diff
@@ -131,7 +144,7 @@ class ReaderPostListViewModel @Inject constructor(
         newsTracker.trackNewsCardExtendedInfoRequested(READER, item.version)
     }
 
-    fun loadSubFilters() {
+    fun onLoadSubFilters() {
```
malinajirka: It seems the method actually loads the subfilters, we might want to consider renaming it back to `loadSubFilters`.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -405,7 +405,7 @@ public void onCreate(Bundle savedInstanceState) {
                         BuildConfig.INFORMATION_ARCHITECTURE_AVAILABLE && mIsTopLevel,
                         mRecyclerView
                 ) || ReaderUtils.isDefaultTag(mCurrentTag)) && getPostListType() != ReaderPostListType.SEARCH_RESULTS) {
-                  mViewModel.applySubfilter(subfilterListItem, true);
```
malinajirka: The whole `mViewModel.getCurrentSubFilter().observe` seems a bit weird. The VM informs the view layer that `currentSubFilter` has changed -> the view layer does some business level comparisons and informs the VM back that  the currentSubFilter has changed. Is there any reason why we observe this? Couldn't we simply move the business level comparisons to the VM and call `onSubfilterChanged` there?

Again something not introduced in this PR, so feel free to ignore this comment.

osullivanchris:
@malinajirka  thanks for testing and for the positive feedback! Glad you are enjoying the feature. 

> The fully expanded state is lost during rotation.

thanks for spotting, sounds like a technical one

> Possible UX issue: When a site/tag is selected and I open the bottom sheet again and I click on the selected site/tag, I'd expect to de-select it.

That's interesting. I wonder is it the selected state. I actually tried a lot of different models for this but ended up trying not to add too much functionality too it. At the moment we're dealing with 'clearing the selection' with the x in the filter bar itself. 

I think there's two ways we could adapt it:
1. Treat it as a check box. Show a selected state. Tapping on the selected state unselects it. But in this case, I think the sheet should stay open. Currently the sheet closes on any selection, so its more like a quick switcher between sites. 
2. Removing the selected indicator makes it feel less like a checkbox/radio button and more like a shortcut.

Its not 100% clear to me that this is a big issue. But if I was to make a change I would actually do the second one, and not show a selected indicator. That might make it clearer how the thing is working, rather than changing how it works.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -2628,6 +2639,17 @@ private void purgeDatabaseIfNeeded() {
         }
     }
 
+    private void performTagsAndSitesUpdate() {
```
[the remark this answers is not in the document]
develric: Agree, changes here 092d6fd903f100bda334b1f66854a0a81606ab00

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterPageFragment.kt
```diff
@@ -64,65 +65,51 @@ class SubfilterPageFragment : DaggerFragment() {
...
+        actionButton = emptyStateContainer.findViewById(R.id.action_button)
+
+        readerViewModel = ViewModelProviders.of(requireActivity(), viewModelFactory)
+                .get(ReaderPostListViewModel::class.java)
 
-        viewModel.subFilters.observe(this, Observer {
+        readerViewModel.subFilters.observe(this, Observer {
             (recyclerView.adapter as? SubfilterListAdapter)?.let { adapter ->
                 val items = it?.filter { it.type == category.type } ?: listOf()
+                viewModel.onManageEmptyView(items.isEmpty())
```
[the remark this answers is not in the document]
develric: Thanks for this Jirka 🙇‍♂️, I think I got the general point but not sure I got the full picture here so for now didn’t change, sorry ! I guess this means that the ReaderPostListFragment should have a reference to the SubfilterPageViewModel? Is this correct? If this applies, since the child VM is per SubfilterPageFragment this would mean to deliver (at least as it is now of course) the events to both fragments view model (probably we can combine them in some way but I would need to better think about that).

Meanwhile changed naming to `onSubfilterPageUpdated` here 89dab3793dbf1c6bdaa270ee12d007a3532b8f46

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterPageFragment.kt
```diff
@@ -64,65 +65,51 @@ class SubfilterPageFragment : DaggerFragment() {
...
                 adapter.update(items)
-                viewModel.updateTabTitle(category, items.size)
+                readerViewModel.onUpdateTabTitleCount(category, items.size)
```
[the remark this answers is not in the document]
develric: As above changed the name method to `onSubfilterPageUpdated` here 89dab3793dbf1c6bdaa270ee12d007a3532b8f46

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
develric: Changed here 89dab3793dbf1c6bdaa270ee12d007a3532b8f46

develric:
Hi @malinajirka and thanks for the usual insights you deliver in your review 🙇‍♂️!

> 1. The fully expanded state is lost during rotation.

I should have fixed this one here 0efe7352afb405b6a83f3363e52b0f63018fdb52. Please note that the solution seems to work well but it's a bit naif 😜! Let me know wdyt.

> 2. Possible UX issue

@osullivanchris , @malinajirka , it seems there are no strong opinions on this for now, can we maybe skip it for this PR? Let me know what do you think, thanks.

> It seems we weren't on the same page when we discussed changing the method names 

Thanks for this, I've already printed it and put on my wall for future reference 😜! (not kidding!)
I agree with your view and description of the pattern (that was also the one by @khaykov ); I think you got it right also that I was basically somewhere in between in being forced to manage the method naming more as `commands` (since a lot of logic is/was in the fragment) and more like `events` on which the VM can act and execute his logic after being notified.

I need to admit that trying to align the methods naming to the `event` convention even when they were currently commands is even worse, so I tried to revert the names I thought could make sense. Let me know wdyt, thanks Jirka 🙇‍♂️.

Having worked a bit now on the reader I think it would deserve a dedicated effort to refactor in the direction you highlighted since it is an important part of the app. Basically something similar to the project you were deeply involved about the `EditPostActivity`. Something to think of in next project/tasks proposal I guess.

osullivanchris:
> @osullivanchris , @malinajirka , it seems there are no strong opinions on this for now, can we maybe skip it for this PR? Let me know what do you think, thanks.

I think its ok to leave it for now. Hope you don't mind @malinajirka !

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -988,14 +988,14 @@ public void onClick(View view) {
...
-                mViewModel.onChangeBottomSheetVisibility(true);
+                mViewModel.onSubFiltersListButtonClicked();
             });
 
             mSubFilterTitle = mSubFilterComponent.findViewById(R.id.selected_filter_name);
 
             mRemoveFilterButton = mSubFilterComponent.findViewById(R.id.remove_filter_button);
             mRemoveFilterButton.setOnClickListener(v -> {
-                mViewModel.onSetDefaultSubfilter();
+                mViewModel.setDefaultSubfilter();
```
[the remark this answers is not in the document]
develric: I was thinking about it but the `setDefaultSubfilter` method is currently used also in other places that are not related to the remove filter click. Also in this case that method is both a `command` and an `event` (this is a problematic one sorry 😅); cannot find a name that fits al the places tbh, but don't know if it makes sense to create separated methods. Let me know if you prefer me to change it anyway and I will just do it 👍.

develric:
Hi @malinajirka 😊, thanks for the quick follow up 🙇‍♂️!

> Yep I agree. Having said that, we might want to create a new GitHub issue so we don't forget about it. Wdyt?

Good point 👍, created it [here](https://github.com/wordpress-mobile/WordPress-Android/issues/11232) 

> Yeah, it deserves a huge refactor. Unfortunately, even the EditPostActivity refactoring is far from being done.

I'm sure `EditPostActivity` is going to be great 😎!

> Is there any reason why we need to set the default/initial state manually? Can't we simply remove
behavior.state = BottomSheetBehavior.STATE_HALF_EXPANDED
and let the system handle the configuration change?

🤦‍♂️! You found a blind spot for me! I was so focusing on managing the state that didn't realize that just setting the `peekHeight` was enough! Thanks for highlighting that! 🙇‍♂️

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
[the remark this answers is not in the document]
malinajirka: Ohh, I missed that. I agree, thanks for clarification!

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterPageFragment.kt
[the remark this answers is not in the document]
malinajirka: I'd need to dive deep into the code to be sure it's even feasible. Let's not worry about it for now. Sorry for bringing it up without verifying it's feasible... I think you are right that it wouldn't be as simple as I thought.

develric:
Hi @malinajirka 😊. Thanks for catching the above issue with the bottom sheet. If it's fine for you I have opened the following [issue](https://github.com/wordpress-mobile/WordPress-Android/issues/11242) to track it and address in a separate PR.

While looking into it I noticed also a small thing in this PR I would like to address that is I think it's better to register/unregister the EventBus in the `ReaderPostListViewModel` only if `INFORMATION_ARCHITECTURE_AVAILABLE` is true. Going to open a dedicated quick PR for that one. Hope it's fine with you, thanks 🙇‍♂️.

develric:
As confirmed in slack with @malinajirka we are merging this one and moving the minor enhancement on the `EventBus` registration discussed here [above](https://github.com/wordpress-mobile/WordPress-Android/pull/11193#issuecomment-582372512) in this dedicated [PR](https://github.com/wordpress-mobile/WordPress-Android/pull/11243) .

PR #11198, opened 2020-01-30
11195 Illegal argument exception fix 

### LINKED ISSUE #11195: IllegalArgumentException: Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkParameterIsNo...
Sentry Issue: [WORDPRESS-ANDROID-7GY](https://sentry.io/organizations/a8c/issues/1386975461/?referrer=github_integration)

```
IllegalArgumentException: Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkParameterIsNotNull, parameter object
    at org.wordpress.android.ui.FilteredPagerAdapter.getItemPosition
    at org.wordpress.android.ui.FilteredRecyclerView.setCurrentFilter(FilteredRecyclerView.java:104)
    at org.wordpress.android.ui.FilteredRecyclerView$2.onFilterCriteriasLoaded(FilteredRecyclerView.java:257)
    at org.wordpress.android.ui.reader.ReaderPostListFragment$LoadTagsTask.onPostExecute(ReaderPostListFragment.java:2672)
    at org.wordpress.android.ui.reader.ReaderPostListFragment$LoadTagsTask.onPostExecute(ReaderPostListFragment.java:2646)
...
(9 additional frame(s) were not displayed)

Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkParameterIsNotNull, parameter object
```

develric:
Fixes #11195 

Was not able to reproduce. 

The fix basically adds a null check outside of the call into the `mFilteredPagerAdapter.getItemPosition` and so before the `kotlin.jvm.internal.Intrinsics.checkParameterIsNotNull` can take care of it. We are basically following the same logic that is inside `mSpinnerAdapter.getIndexOfCriteria`.

Just as a note, AFAIU the issue (and the fix) currently affects only the alpha internal so we should be safe enough.

## To test:
Not able to reproduce myself so not sure how to test it. I can just report that @frosty (thanks for this 🙇‍♂️) was having the issue on his android device and was able to confirm the patched apk removed the crash.

cc @maxme (as anticipated can make sense to update the current alpha with this patch when reviewed/approved)

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/FilteredRecyclerView.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/FilteredRecyclerView.java
@@ -101,7 +101,7 @@ public class FilteredRecyclerView extends RelativeLayout {
         mCurrentFilter = filter;
 
         if (mUseTabsForFiltering) {
-            int position = mFilteredPagerAdapter.getItemPosition(filter);
+            int position = filter == null ? -1 : mFilteredPagerAdapter.getItemPosition(filter);
             if (position > -1 && position != mViewPager.getCurrentItem()) {
                 mViewPager.setCurrentItem(position);
             } else if (position > -1 && position == mViewPager.getCurrentItem()) {


develric:
Hi @oguzkocer 😊!
First of all thanks for looking into it deeply, much appreciated 🙇‍♂️!

I should have probably articulated more about this PR (my bad for this!). The intention of this PR was to be a temporary fix. We have an open PR [here](https://github.com/wordpress-mobile/WordPress-Android/pull/10986) that was addressing some issues I think can be partly related to this one. I already declared that one as `not ready` since some more investigation was needed (and also to separate that one in more isolated PRs tbh). Since @frosty confirmed that the quick fix was solving the issue for him my plan was to make this reach the alpha branch so to not prevent some alpha users to give additional feedback on the release and then come to the wider issue. 

But since you made a good analysis [here](https://github.com/wordpress-mobile/WordPress-Android/pull/11206) makes sense to take advantage of it (that hopefully it's not far from the spirit of groundskeeping also 😊), so moving there to follow up 👍!

develric:
Closing in favor of #11206

PR #11207, opened 2020-01-31
Fix update gravatar feature

### LINKED ISSUE #11196: [GlobalStep] The feature to change the Profile Photo does not function
### Description
When the user attempts to change the Profile Photo, after selecting or taking a new photo the application returns to the profile screen but the Profile Photo remains unchanged.

### Reproduction Rate
4/4 100%

### Expected behavior
The Profile Photo should update to the selected photo.

### Actual behavior
The Profile Photo remains unchanged.

1. Install WordPress 14.1
2. Log in to an account.
3. Press the “Me” button on the top right of the screen.
4. Press “Change Photo”.
5. Select an existent photo or take a new one.
##### Tested on the following
Samsung Galaxy S6 (7.0)
Samsung Galxay S7 (6.0.1)

#### Please see the attached video for more information
[AndroidProfilePhoto.zip](https://github.com/wordpress-mobile/WordPress-Android/files/4133838/AndroidProfilePhoto.zip)

### Submitted by:
Luis Pimenta

malinajirka:
Fixes #11196

This PR fixes an issue with `startActivityForResult` being called on an activity, but the result being expected in a fragment. 
This issue was introduced when we moved "Me" section from the bottom navigation bar into the toolbar (it's own activity). In the previous location the WPMainActivity would manually delegate the `onActivityResult` to the fragment. 
Imo it's better to let the system handle the delegation, so I decided to call `fragment.startActivityForResult` instead of adding something similar to [this](https://github.com/wordpress-mobile/WordPress-Android/blob/f1663e6a9b49ad6226a60dc210d6bcf20ae15eba/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java#L950) to `MeActivity`. Wdyt?

To test:
1. Click on "Me" section in the toolbar
2. Click on "Switch photo"
3. Choose an image and confirm
4. Edit Image screen is opened (this didn't work before)
5. Confirm again and notice the gravatar gets uploaded and set
------------------------
1. Create a new post
2. Open EditPostSettings
3. Click on "Set featured image"
4. Select an image and make sure the image gets uploaded and set

PR submission checklist:

cc @maxme @jkmassel This bug was introduced in 14.1. It'd be great if we could release a new beta version with the fix. Thanks!

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
@@ -128,7 +128,23 @@ public class ActivityLauncher {
                                                 @NonNull MediaBrowserType browserType,
                                                 @Nullable SiteModel site,
                                                 @Nullable Integer localPostId) {
-        Intent intent = new Intent(activity, PhotoPickerActivity.class);
+        Intent intent = createShowPhotoPickerIntent(activity, browserType, site, localPostId);
+        activity.startActivityForResult(intent, RequestCodes.PHOTO_PICKER);
+    }
+
+    public static void showPhotoPickerForResult(Fragment fragment,
+                                                @NonNull MediaBrowserType browserType,
+                                                @Nullable SiteModel site,
+                                                @Nullable Integer localPostId) {
+        Intent intent = createShowPhotoPickerIntent(fragment.getContext(), browserType, site, localPostId);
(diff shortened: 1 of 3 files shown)


maxme:
> This bug was introduced in 14.1. It'd be great if we could release a new beta version with the fix. Thanks!

malinajirka:
I think one reviewer is enough for this change. Thank you @develric for the review and thanks @maxme for releasing a new version;)!

PR #11218, opened 2020-02-01
Fix Reader tab bar not retaining its scroll position

### LINKED ISSUE #11010: IA Reader: tabs bar not retaining scroll position under certain circumstances
### Expected behavior
Tablayout should retain the scroll position when user switch away from the reader screen and comes back.

### Actual behavior
The scroll position is not retained and the selected tab risk to go out of current visualization.

1. Select an item off the right side of the tab bar. It becomes centred, as you’d expect:
<p align="center">
  
</p>
2. Switch to the Notifications section of the app and then back to Reader.
3. Notice that the tab is now off the right side of the screen again; the bar has scrolled back to its default position:
<p align="center">
  
</p>

renanferrari:
Fixes #11010

This should be considered a temporary fix. A larger rework is probably required to make the **Reader** take full advantage of the `TabLayout`. Please read the comments on the issue linked above for more details.

To test:

0. Make sure you're using the build variant thas has enabled IA flags.
1. Open the **Reader**.
2. Select an item off the right side of the tab bar. 
3. Notice that the tab becomes [centered](https://user-images.githubusercontent.com/47797566/71355040-18c6a500-257e-11ea-92af-82ad0ed19b0e.png).
4. Switch to the **Notifications** section of the app and then go back to the **Reader**.
5. Notice that the tab remains centered.

PR submission checklist:


develric:
Hi @renanferrari 👋! Thanks for this PR . I run the steps for testing above and it worked as described 👍. The code overall looks good; I did not complete the review yet, since I found one issue that I hope you can look into before to complete 🙏!

### Page refresh triggered on configuration change or coming from My Site or Notifications
Switching from My Site/Notifications to Reader page or on device screen rotation, a refresh is triggered. We would like to avoid it.

I suspect this could have something to do with some logic in the `initFilterAdapter` related to the boolean `mSelectingRememberedFilterOnCreate` (used in the spinner case). I would also check if it could be related (I'm not sure this is the case tbh) to possible `mCurrentFilter` being `null` in `int currentFilterPosition = mFilterCriteriaOptions.indexOf(getCurrentFilter());`

Let me know wdyt 😊.

renanferrari:
Hey @develric 👋Thanks a lot for those observations. You're right, the `TabLayout` was not making use of the `mSelectingRememberedFilterOnCreate` as I thought that was there only to handle `Spinner` reselections, but I now see that it was there to prevent unnecessary refreshes. This is fixed now.

I have also removed that piece of code that was calling `getCurrentFilter()`. It wasn't causing any problems, but I realized it was simply not necessary there.

PR #11222, opened 2020-02-03
Issue/11142 show low resolution image on the preview screen

### LINKED ISSUE #11142: Show low resolution image on the Hub
Goal of this task is to show a low resolution version of the image which is about to be opened in uCrop. Use `ImageManager` to load the image into an ImageView

ashiagr:
Fixes #11142

This PR loads a low resolution image into `ImageEditor` module's ImageView using main app's image loader. 

P.S. Low resolution image url is generated using Photon service and device's width. This url needs to be created explicitly as unlike iOS we don't yet have a common image loader between WPAndroid and ReactNative editor.

**Steps to test**
- Select a site having a post with an image and using block editor
- Select the post and click the image on the post
- Low res image should display on the next screen

Test for
- Image resolution and loading time (we might need to figure out optimal resize factor).
- Image rendering on configuration and app/ fragment lifecycle changes like device rotation, app moving from background to foreground, screen lock/ unlock.
- Image rendering on Photon capable site and other types of sites I might not be aware of.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -1647,11 +1650,35 @@ public void onImageSettingsRequested(EditorImageMetaData editorImageMetaData) {
     @Override public void onImagePreviewRequested(String mediaUrl) {
 //        MediaPreviewActivity.showPreview(this, null, mediaUrl);
         // TODO: Temporarily open image editor at this point
-        ActivityLauncher.openImageEditor(this, mediaUrl);
+        String resizedImageUrl = StringUtils.notNullStr(mediaUrl);
```
ashiagr: I considered moving this logic to the `PhotonUtil` class inside `WordPressUtils` lib.  Wondering if we can also move `ReaderUtils`  to the lib and reuse `ReaderUtils.getResizedImageUrl`?  I guess it will be out of scope of this PR.

ImageEditor/src/main/AndroidManifest.xml
```diff
@@ -4,6 +4,7 @@
     <application>
         <activity
             android:name=".EditImageActivity"
+            android:configChanges="orientation|screenSize"
```
malinajirka: Can you please elaborate why we want to handle the configuration changes manually?

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
```diff
@@ -4,29 +4,87 @@ import android.os.Bundle
...
     override fun onCreateView(
         inflater: LayoutInflater,
         container: ViewGroup?,
         savedInstanceState: Bundle?
-    ): View? = inflater.inflate(R.layout.fragment_preview_image, container, false)
+    ): View? = inflater.inflate(layout.fragment_preview_image, container, false)
+
+    override fun onSaveInstanceState(outState: Bundle) {
+        val intent = activity?.intent
+        if (intent != null) {
```
malinajirka: Do we really need to copy the values from intent into the bundle? I'd expect the system will always recreate the Activity with the same intent it was started with. Wdyt?

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
```diff
@@ -4,29 +4,87 @@ import android.os.Bundle
...
+        }
+        super.onSaveInstanceState(outState)
+    }
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+
+        val nonNullIntent = checkNotNull(requireActivity().intent)
+
+        lowResImageUrl = if (savedInstanceState == null) {
```
malinajirka: We might want to consider passing the URL into the VMs using the `vm.start(..)` approach we use in other VMs. 
I think the view layer shouldn't be responsible for keeping the data (it should be responsible only for the state of its views). Wdyt?

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
```diff
@@ -4,29 +4,87 @@ import android.os.Bundle
...
-        const val ARG_IMAGE_URL = "arg_image_url"
+    override fun onPause() {
+        wasPaused = true
+        super.onPause()
+    }
+
+    override fun onResume() {
+        super.onResume()
+
+        if (wasPaused) {
```
malinajirka: Do we really need to manually handle "wasPaused" state? Can't we load the image in "onCreateView" (ideally from vm) and let the system handle the state of the view after a configuration change?

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
malinajirka: I'd consider moving the method we need from ReaderUtils to a utility class located in WordPressUtils lib. Wdyt?

SideNote: I'm also not sure why we need to introduce something like this here - we are loading images in the reader, Aztec, Gutenberg, comments. How come we don't have this logic somewhere already?

ashiagr:
> We might want to consider passing the URL into the VMs using the vm.start(..) approach we use in other VMs.

Thank you for the directions @malinajirka ! I'll go through other VMs and update the code.  Hopefully it will fix other issues as well.

ashiagr:
As discussed, noting down some of the findings and decisions we took in our call for this PR: 

1. **Configuration change**  
Method `loadImage` is called multiple times on configuration change (like rotation) because of an active observer. We'll load the image from the cache right now and consider using `Event`  in an upcoming PR.
https://github.com/wordpress-mobile/WordPress-Android/blob/b8346ea30a19e1124895ed84701397a7b54c34fc/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt#L47-L49

2. **Resized image url** 
We need to test for different types of sites if the resized (low res) image url is generated properly. Photon url generation logic is kept similar to the one we use for _PostListViewModel -> convertToPhotonUrlIfPossible_  with an assumption it works for different types of sites. We tested resized image url for photon capable private site and private post -> [video link](https://www.dropbox.com/s/vdcmd7smtvsdgk9/private_site_post.webm?dl=0)

https://github.com/wordpress-mobile/WordPress-Android/blob/59778060ca0dd6b716ba60ea8a727fa6a0d4137c/WordPress/src/main/java/org/wordpress/android/viewmodel/posts/PostListViewModel.kt#L385-L391

3. **Network handling** 
We’ll display default error placeholder used in the `ImageManager` method in case of no network 
https://github.com/wordpress-mobile/WordPress-Android/blob/35ec1333ad125819724d599f56d8ffa7461c91c0/WordPress/src/main/java/org/wordpress/android/util/image/ImageManager.kt#L75-L76

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
ashiagr: Fixed here acd9e6a

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
ashiagr: Changed here b8346ea

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
ashiagr: Fixed here acd9e6a

ImageEditor/src/main/AndroidManifest.xml
ashiagr: Fixed here acd9e6a

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
ashiagr: Changed here d151f4f

jkmassel:

PR #10738, opened 2019-11-05
Add Encryption Functionality and Tests

ravenstewart:
This PR makes progress towards a solution for issue #10698.

### Description
I added an EncryptionUtils static class which currently has one function:
`public static String getEncryptedAppLog(Context context)`

This function returns a String containing JSON in this format:
```
        {
            "keyedWith": "v1",
            "encryptedKey": "$key_as_base_64",  // The encrypted AES key
            "header": "base_64_encoded_header", // The xchacha20poly1305 stream header
            "messages": []                      // the stream elements, base-64 encoded
        }
```

The `messages` value is a JSON array containing encrypted elements that represent lines in the logs from AppLog.  The last element in the JSON array is an empty string with the `FINAL` tag.

Calling this function will also write out two test output files in the Android device storage:
- `app_log.txt `: the original log text from AppLog
- `app_log_encrypted.json`: JSON with encrypted log data

### Testing Instructions

**Steps:**
1. Call this function from the App after AppLog has accumulated some logs.  (For my tests, the function call was added to the  `void copyAppLogToClipboard()` function in `AppLogViewerActivity.java`.  This way I can view the logs in the app before I generate the test files.)
2. Build the Android app, run the app in the emulator, and trigger the function call to generate the test files.
3. Run the command `adb pull /sdcard/<test file name> <target path>` to copy the files to your hard drive
4. Run the command `curl --data "@<file name>.json" https://log-encryption-testing.herokuapp.com`
5. Compare the output from the test endpoint with the contents of `app_log.txt`

**Expected Results:**
The contents from the test endpoint should be the same as the contents of `app_log.txt`.  (Currently, there is an extra new line at the end of the output from the test endpoint.)

### Next Steps
- Use this function to create encrypted log data associated with Sentry events
    - Need to associate an UUID with the log data and send the UUID with the Sentry Event
    - Where should we store the log data and how will the team access it?
- Use this function to create encrypted log data used for Zendesk ticket creation
    - How will the team decrypt the data once they retrieve it from Zendesk?

### Additional Notes

- I chose the [libsodium-JNI](https://github.com/joshjdevl/libsodium-jni) Java (Android) binding because it had the simplest steps for integration out of all of the bindings for Java (Android) that had the features that we needed.

PR #11230, opened 2020-02-04
Issue/10698 encryption utils cleanup

claucookie:
Making progress towards this issue https://github.com/wordpress-mobile/WordPress-Android/issues/10698
## Description
This PR intends to simplify and make the `EncryptionUtils.java` easy to follow.

The following tasks have been performed:
- Remove Android Studio warnings.
- Export some logic into methods to improve readability.
- Rename some methods to clarify responsibility.
- Rename `encryptStringData()` into `generateJSONEncryptedLogs()`
- Update modifiers to avoid exposing variables or methods that might not be needed outside the class.
- Update the method signature for `generateJSONEncryptedLogs()` to receive a list of "messages" instead of a formatted string split by `\n` to avoid formatting errors.

## To test:
- Run the testing class `EncryptionUtilsTest.java`
- Or follow the testing instructions from this PR https://github.com/wordpress-mobile/WordPress-Android/pull/10738

## Next Steps
- Look into the testing instructions, the api https://log-encryption-testing.herokuapp.com is returning an error when I send the json file (I checked also on develop, it seems to return the same result).
- Integrate this with Sentry (the crash logging system) 
- Look into performance for big log files.
- Move class and tests to the library https://github.com/wordpress-mobile/WordPress-Analytics-Android
- Look into persisting the logs between sessions for the WPAndroid.

## PR submission checklist:


WordPress/src/androidTest/java/org/wordpress/android/util/EncryptionUtilsTest.java
```diff
@@ -4,58 +4,62 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
 import static junit.framework.Assert.fail;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 
 @RunWith(AndroidJUnit4.class)
 public class EncryptionUtilsTest {
```
claucookie: Side note: At the moment this class sits inside `androidTest folder, which makes it part of the Instrumentation tests, I think this would fit better inside `test` as part of unit tests. 
Thoughts?

WordPress/src/main/java/org/wordpress/android/util/EncryptionUtils.java
```diff
@@ -5,118 +5,116 @@
...
+    private static final int BASE64_FLAGS = Base64.DEFAULT;
+    private static final String KEYED_WITH = "v1";
+    private static final byte[] STATE = new byte[STATEBYTES];
+
+    /**
+     * This method is in charge of encrypting logs using the symmetric cypher
+     * XCHACHA20 and the cryptographic message authentication code (MAC) POLY1305.
+     * For more information about the encryption process you can find it here
+     * https://libsodium.gitbook.io/doc/secret-key_cryptography/secretstream.
+     * @param publicKeyBase64 The public key encoded using Base64.
```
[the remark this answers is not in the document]
claucookie: Very good point 👏 , I didn't notice that. Now I look at it, it certainly makes sense to pass it as bytes. Thanks

WordPress/src/androidTest/java/org/wordpress/android/util/EncryptionUtilsTest.java
claucookie: My bad, moving the tests to the `test` folder can't be done. The library is an `AAR` file (Android ARchive) which means it needs to be run on an emulator/device. 🤦‍♀ 
I noticed when I tried to move the library and I ran the tests, which displayed the following error: 
`java.lang.UnsatisfiedLinkError: no sodiumjni in java.library.path` 

It is funny, there happen to be person who reported this, it is not a library issue though. 
https://github.com/joshjdevl/libsodium-jni/issues/95

claucookie:
@jkmassel good news! Here is an update about the pending stuff for this PR :) 

>  Ensure that the code works correctly with the API endpoint (we'll need to encrypt messages with the endpoint's certificates – provided via Slack)

✅ Done, solved on Slack. https://a8c.slack.com/archives/C0HTU7HV3/p1581003101096500?thread_ts=1581001073.092400&cid=C0HTU7HV3

> Follow your suggestion to move this over to the Unit Tests

🚫 This was a very good idea until I realized that **It can't be done**. For more details feel free to go to the explanation in this comment https://github.com/wordpress-mobile/WordPress-Android/pull/11230#discussion_r375334800

> Consider avoiding an encode/decode cycle on the public key.

✅ Done

Thank you so much for your assistance, it is really appreciated it.

PR #11240, opened 2020-02-05
Fix #10747: Bump phase 2 rollout to 100%

### LINKED ISSUE #10747: Gutenberg progressive rollout strategy

maxme:
This is the last step of the Gutenberg rollout - fixes #10747.

Note: this targets 14.1, we were waiting for 14.0 feedback before making that changes.

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
+++ b/WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
@@ -27,7 +27,7 @@ public class SiteUtils {
     public static final String GB_EDITOR_NAME = "gutenberg";
     public static final String AZTEC_EDITOR_NAME = "aztec";
     private static final int GB_ROLLOUT_PERCENTAGE_PHASE_1 = 100;
-    private static final int GB_ROLLOUT_PERCENTAGE_PHASE_2 = 30;
+    private static final int GB_ROLLOUT_PERCENTAGE_PHASE_2 = 100;
 
     /**
      * Migrate the old app-wide editor preference value to per-site setting. wpcom sites will make a network call


mchowning:
This change looks fine to me, but should we go ahead and do some cleanup of the phase 1 rollout code? Now that Phase 2 will be at 100%, the Phase 1 code is just dead code (i.e., `(accountStore.getAccount().getUserId() % 100 >= (100 - GB_ROLLOUT_PERCENTAGE_PHASE_2)` will now always be `true`).

maxme:
> This change looks fine to me, but should we go ahead and do some cleanup of the phase 1 rollout code?

mchowning:
> Some users might migrate from an older version of the app (eg. from 13.0 to 14.1) and in that case, the mobile editor settings won't be set, and they will enter in the "phase 1" group (atLeastOneSiteHasAztecEnabled will be false).

You're right, I saw the `atLeastOneSiteHasAztecEnabled` check for skipping Phase 1 as well, but missed that it was reversed.

PR #11241, opened 2020-02-05
Issue/11143 show loading indicator and blur effect on the preview screen

### LINKED ISSUE #11143: Show loading indicator and blur effect on the Hub
Show an overlay, progress bar and "Loading media" label over the ImageView with the low resolution version of the image.

Consider using ImageManager.loadWithResultListener + it's `thumbnail` parameter.

ashiagr:
Fixes #11143

This PR 
- Loads low resolution image using the thumb attribute of main module's `ImageManager.loadWithResultListener` [as per our discussion on slack to use main app's image request listener

https://github.com/wordpress-mobile/WordPress-Android/blob/35ec1333ad125819724d599f56d8ffa7461c91c0/WordPress/src/main/java/org/wordpress/android/util/image/ImageManager.kt#L39

 in the `ImageEditorInitializer` and delegating the calls to the listener in the `ImageEditor` module.]
- Displays a progress indicator while the image loads
- Uses `uiState` as an initial attempt to control visibility of the progress bar similar to how we use in `FeaturedImageHelper`
https://github.com/wordpress-mobile/WordPress-Android/blob/040032a58c921b40611a991acf722ee740a2902b/WordPress/src/main/java/org/wordpress/android/ui/posts/FeaturedImageHelper.kt#L157

P.S. As discussed with our designer @mbshakti , we decided to just display the low resolution image as a placeholder adding no further blur effect and to display a normal progress loader in this first version.

In order to speed things up, as part of this PR, I want to discuss:
1.  `uiState` approach: Is the right one to show/hide progress bar in this case? I just tried to be consistent with an existing class in the main app and would like to know other alternatives I should rather focus on.
2. Noticed we're using same request listener for both - thumbnail and main image in `loadWithResultListener` method:
https://github.com/wordpress-mobile/WordPress-Android/blob/35ec1333ad125819724d599f56d8ffa7461c91c0/WordPress/src/main/java/org/wordpress/android/util/image/ImageManager.kt#L142-L158
It causes the progress bar to hide as soon as the thumbnail loads. We would want to display the progress bar to display till the time full image loads. Any suggestions?

PR submission checklist:


ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
```diff
@@ -37,19 +42,38 @@ class PreviewImageFragment : Fragment() {
...
         viewModel = ViewModelProvider(this).get(PreviewImageViewModel::class.java)
         setupObservers()
-        viewModel.start(lowResImageUrl)
+        viewModel.start(lowResImageUrl, highResImageUrl)
     }
 
     private fun setupObservers() {
-        viewModel.loadImageFromUrl.observe(this, Observer {
+        viewModel.loadImageFromData.observe(this, Observer {
             loadImage(it)
```
malinajirka: It seems we are ignoring the UI state the vm has sent us, is this intentional?

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
```diff
@@ -6,19 +6,36 @@ import androidx.lifecycle.ViewModel
...
+        highResImageUrl = hiResImageUrl
+
+        _loadImageFromData.value = ImageData(
+            uiState = ImageUiState.IMAGE_LOAD_IN_PROGRESS,
+            lowResImageUrl = lowResImageUrl,
+            highResImageUrl = highResImageUrl
+        )
+    }
+
+    data class ImageData(val uiState: ImageUiState, val lowResImageUrl: String, val highResImageUrl: String)
```
malinajirka: Thinking about it a bit more, I'd even consider changing direction of the dependency from ImageUiState ->ImageData, to ImageData->ImageUiState. I basically think, only `IMAGE_LOAD_IN_PROGRESS` needs to have access to the ImageData, right? (we'd need to replace enum with a sealed class).

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
```diff
@@ -6,19 +6,36 @@ import androidx.lifecycle.ViewModel
 
 class PreviewImageViewModel : ViewModel() {
     private lateinit var lowResImageUrl: String
```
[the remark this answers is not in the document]
ashiagr: Removed here 7ba47d2

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
ashiagr: Replaced here 2a39251, 30f2166

ashiagr:
Thanks for detailed suggestions @malinajirka! Appreciate it 🙏. 

> Does at least glide support differentiating between these two cases? If it does, we might need to update the ImageManager so it can handle this scenario. I'd consider doing such change in a separate PR.

Yes, changes made [here](https://github.com/wordpress-mobile/WordPress-Android/pull/11329) in `ImageManager` and f737b85 to include model param that provides the url of the image being loaded. 

Currently included ui states just for the low res image. Can you take a look at these changes first?

ImageEditor/src/test/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModelTest.kt
```diff
@@ -20,8 +24,44 @@ class PreviewImageViewModelTest {
...
+
+    @Test
+    fun `progress bar shown on low res image load success`() {
+        initViewModel()
+        viewModel.onImageLoadSuccess(TEST_LOW_RES_IMAGE_URL)
+        assertThat(requireNotNull(viewModel.uiState.value).progressBarVisible).isEqualTo(true)
+    }
+
+    @Test
+    fun `progress bar hidden on low res image load failed`() {
```
malinajirka: I'm wondering if we really want to hide the progress when the low res image fails to load? The high res image can still succeed right?

ImageEditor/src/test/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModelTest.kt
```diff
@@ -20,8 +24,44 @@ class PreviewImageViewModelTest {
...
+        assertThat(viewModel.uiState.value).isInstanceOf(ImageInLowResLoadSuccessUiState::class.java)
+    }
+
+    @Test
+    fun `low res image failed ui shown on low res image load failed`() {
+        initViewModel()
+        viewModel.onImageLoadFailed(TEST_LOW_RES_IMAGE_URL)
+        assertThat(viewModel.uiState.value).isInstanceOf(ImageInLowResLoadFailedUiState::class.java)
+    }
+
```
malinajirka: The tests are truly awesome! I'm just wondering, since we are already loading even the high res image, we might want to consider adding similar tests for the high res image, wdyt?

For example
`progress bar hidden on high res image load success`
`progress bar hidden on high res image load failed`
`high res image success ui shown on high res image load success`
`high res image failed ui shown on high res image load failed`

Wdyt?

ashiagr:
Thanks! 
> I've left two in-line comments about the tests, but I think if we decide to implement some changes, we can do so in a different PR

Agree. Let's do it in a different PR.

ImageEditor/src/test/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModelTest.kt
malinajirka: Another test we might want to consider adding is
`high res image success ui shown when low res image loads after high res image`

Issue #11260, opened 2020-02-07
Offline Support Pages - Update Page Status causes null post when fetching upload status


jd-alexander:
### Expected behavior
A post should be available in the `PageItemUploadProgressHelper`s `getProgressStateForPage` at all times so that the `UploadStatus` can be used to determine the `ProgressState`.

### Actual behavior
when the page status is being changed and  `PageItemUploadProgressHelper`s `getProgressStateForPage` is called it returns null at times. 
This will require investigation. 

1. The current null check might be hiding a potential issue. 
2. `The database just returned us a list of pages and a few ms later it seems the page is not in the database anymore. Do we remove the page from the database before we move it to draft/publish it?` an important question posed by @malinajirka which should help the investigation. 

### Steps to reproduce the behavior
1. Go the `Page List` and change the page status of either a `Published` or `Draft` page multiple times until there is a crash. 

Tested on Google Pixel 4 XL, Android 10, WPAndroid https://github.com/wordpress-mobile/WordPress-Android/pull/11191

malinajirka:
AFAIK this issue will be fixed in https://github.com/wordpress-mobile/WordPress-Android/issues/11271. We won't load the model from database and we'll interact with the model which we already have and can't be null, right?  If I'm not missing something, please feel free to close this issue @jd-alexander . Thanks!

Issue #11177, opened 2020-01-28
Update snackbar messages


malinajirka:
Update SnackBar messages to be in-line with page list item labels. I think we'll be able to "reuse" all the messages from posts (we just need to replace "post" with "page").
Also enable actions which are suppressed in offline mode - eg. Publish Now in the page list displays "Your device is offline" message. It should enqueue the post and show something like "Page will be uploaded when your device is back online".

PR #11266, opened 2020-02-07
Offline Support Pages : Cancel Auto Upload

### LINKED ISSUE #11150: Add "cancel auto-upload" action to PageList items

jd-alexander:
Fixes #11150 

## Findings
The `Post List` has logic within `PostListItemUiStateHelper` that helps to determine which buttons should be shown based on several different conditions. This logic was adapted to create a similar behavior for the `Page List` cancel auto-upload functionality. 

## Solution
The `PageItems` had the actions hardcoded within its constructor. The actions were removed from them and added at instantiation so that logic could be run that determines if the `Cancel` button should be shown or not. 

## Issues Noticed 
* There's a crash that occurs at times when trying to fetch the post to determine if a page can be canceled or not, so a null check has to be done 0a035ad6de396287865228b0acd0382da4389669. I think this crash is related to the one taking place here https://github.com/wordpress-mobile/WordPress-Android/issues/11260. I think the refactor [here](https://github.com/wordpress-mobile/WordPress-Android/issues/11271) should resolve this. 

* Sometimes when a `Scheduled` post is saved, it doesn't show the `Cancel` action or the title doesn't get updated. I don't think that's related to this PR but it might need investigation.

* I added a `localId` and `remoteId` to the `PageItem` because there are times when either is required. I renamed the `id` in `PageItem` to `localId` because it's often confusing when trying to determine which type of id should be used for a query 4caff1c. A further enhancement would be to simply change the types to `LocalId` and `RemoteId` so that that the compiler's type system enforces this across the featureset. I opted to leave this out of this PR. Should I create an issue for this or it can remain? since the name is already conveying the intent without changing the type. 

## Testing
1. Put the device in airplane mode.
2. Create a new page. Either a draft, published or scheduled page. Repeat all the steps below for each page type.
3. Once the page has been created, publish it or save it. 
4. Click on the more button and you should see a “Cancel” button. 
5. Once you press cancel you should see a snackbar message saying that the changes won’t be published. 
6. Go back online and ensure that the page isn’t uploaded. 
7. Also, check other pages that haven’t been changed to ensure the “Cancel” action doesn’t show for them. 

Before | After 
--------|-------
        |    

## Reviewing 

## Submitter Checklist


jd-alexander:
Hi @malinajirka this PR is almost ready for review! Quick question, so I am seeing where clicking "Publish Now" on the `Page List Item` won't be triggered in airplane mode since no internet connectivity is available. I checked the `Post List` and realized the action is allowed, as expected 😉 so I am wondering if this should be included in this PR as well. Since, testing this behavior is related to the toggling of the `auto-upload cancellation`. 

Before | After 
--------|-------
        |

malinajirka:
>"Publish Now" on the Page List Item won't be triggered in airplane mode since no internet connectivity is available.

I was planning to include it in https://github.com/wordpress-mobile/WordPress-Android/issues/11177. You can test the cancel action by publishing the page from the editor. Thanks for checking!

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageItemUiStateHelper.kt
```diff
@@ -117,6 +132,38 @@ class PageItemUploadProgressHelper @Inject constructor(
...
+            DRAFTS -> mutableSetOf(VIEW_PAGE, SET_PARENT, PUBLISH_NOW, MOVE_TO_TRASH).apply {
+                if (canCancelPendingAutoUpload(pageId, site)) {
+                    add(CANCEL_AUTO_UPLOAD)
+                }
+            }
+            TRASHED -> setOf(MOVE_TO_DRAFT, DELETE_PERMANENTLY)
+        }
+    }
+
+    @VisibleForTesting(otherwise = PRIVATE)
```
jd-alexander: How do you feel about using this annotation since I am making the method available for testing?

WordPress/src/test/java/org/wordpress/android/viewmodel/pages/PageListViewModelTest.kt
```diff
@@ -296,14 +300,95 @@ class PageListViewModelTest : BaseUnitTest() {
...
+
+        // Assert
+        assertThat((result[0].first[0] as DraftPage).actions).contains(CANCEL_AUTO_UPLOAD)
+    }
+
+    @Test
+    fun `CANCEL_AUTO_UPLOAD is not added to DraftPage if auto upload is not pending`() {
+        // Arrange
+        whenever(pageItemUiStateHelper.canCancelPendingAutoUpload(any(), any())).thenReturn(false)
+        whenever(pageItemUiStateHelper.setupPageActions(any(), any(), any())).thenCallRealMethod()
```
jd-alexander: I am using the `thenCallRealMethod()` method so that I can verify that the logic within the helper is generating the appropriate actions depending on the `canCancelPendingAutoUpload`. I did a search of the codebase and realized it isn't used anywhere else so I am just bringing it your attention to hear your thoughts 😄

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageItemUiStateHelper.kt
```diff
@@ -117,6 +132,38 @@ class PageItemUploadProgressHelper @Inject constructor(
...
+    fun canCancelPendingAutoUpload(pageId: LocalId, site: SiteModel): Boolean {
+        val post = postStore.getPostByLocalPostId(pageId.value)
```
malinajirka: I'd suggest adding a TODO here, since this is yet another place which we'll need to update as the current implementation would have significant performance impact.

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageItemUiStateHelper.kt
```diff
@@ -117,6 +132,38 @@ class PageItemUploadProgressHelper @Inject constructor(
...
+
+        post?.let {
+            val uploadUiState = createUploadUiState(
```
malinajirka: We might want to consider adding a TODO here as well. I think we'll want to pass the uploadState from the VM when https://github.com/wordpress-mobile/WordPress-Android/pull/11278 is merged. Wdyt?

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageItemUiStateHelper.kt
malinajirka: I'm not a fan of using these types of annotations. The fact that you need something like this is a code smell - flawed design. 
Moreover, in this case the code smell isn't just a false positive. This class does more than it should. It breaks the SRP. I understand that the main reason is that you were duplicating the design from the PostList. However, I think we might want to consider breaking it down. I actually already broke it down a bit in https://github.com/wordpress-mobile/WordPress-Android/pull/11278. I'd suggest waiting until https://github.com/wordpress-mobile/WordPress-Android/pull/11278 is merged before doing any further changes on this PR. Wdyt?

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/SearchListViewModel.kt
```diff
@@ -88,38 +91,62 @@ class SearchListViewModel
...
                     progressBarUiState = progressState.first,
                     showOverlay = progressState.second
             )
             PageStatus.SCHEDULED -> ScheduledPage(
                     remoteId,
+                    pageId,
                     title,
                     date,
+                    actions = pageItemUiStateHelper.setupPageActions(
+                            PUBLISHED,
```
malinajirka: Shouldn't we pass `SCHEDULED` here?

WordPress/src/test/java/org/wordpress/android/viewmodel/pages/PageListViewModelTest.kt
malinajirka: If we call the real method it won't be a unit tests anymore. Since it'll test two separate units. 

>so that I can verify that the logic within the helper is generating the appropriate actions depending on the canCancelPendingAutoUpload

This would ideally be tested in unit tests for the helper. If it's tested there we can rely on it here and we can just mock it.

`thenCallRealMethod` is usually used for utility methods about which we are 99.99% sure they work as expected. Eg. when we use java Date methods etc. Chances that our code discovers a bug in Java is so low that it's ok to use `thenCallRealMethod` for such cases.

jd-alexander:
Thanks for the thoughtful review @malinajirka . Your suggestion to go on a call to figure out the best way to do the merge due to the conflicts are more than welcome 😄 Once that's done the comments and changes here will be addressed!

WordPress/src/test/java/org/wordpress/android/viewmodel/pages/PageListViewModelTest.kt
jd-alexander: Okay, thank you for this! I knew there was something I was missing and your explanation helps me to understand this clearly!

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageItemUiStateHelper.kt
jd-alexander: > I'm not a fan of using these types of annotations. The fact that you need something like this is a code smell - flawed design.

Indeed! I am glad I pointed it out because there was something about it that didn't feel right 🙇  

> Moreover, in this case the code smell isn't just a false positive. This class does more than it should. It breaks the SRP. I understand that the main reason is that you were duplicating the design from the PostList. However, I think we might want to consider breaking it down. I actually already broke it down a bit in #11278. I'd suggest waiting until #11278 is merged before doing any further changes on this PR. Wdyt?

Yes. I really love this approach. Looking at the `usecases` you created for the labels and colors has solidified the approach that should be taken for components so they are more maintainable and testable.

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageItemUiStateHelper.kt
```diff
@@ -1,27 +1,42 @@
...
+import org.wordpress.android.viewmodel.pages.PageListViewModel.PageListType.SCHEDULED
+import org.wordpress.android.viewmodel.pages.PageListViewModel.PageListType.TRASHED
 import org.wordpress.android.viewmodel.posts.PostListItemUploadStatus
+import org.wordpress.android.viewmodel.uistate.ProgressBarUiState
 import javax.inject.Inject
 
 typealias ShouldShowOverlay = Boolean
 
-class PageItemUploadProgressHelper @Inject constructor(
+class PageItemUiStateHelper @Inject constructor(
```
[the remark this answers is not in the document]
jd-alexander: Yes, that makes sense for sure. It also makes it easier to have tests and behavior constrained to one use case. I must admit, I was using the component in the `PostList` as inspiration for the changes but now after looking at the work being done and the approaches being taken I realize even it's design might need to be changed in the future for it to be more maintainable.

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageItemUiStateHelper.kt
jd-alexander: Yes, that's a really good idea. Thanks!

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/SearchListViewModel.kt
jd-alexander: Exactly so. Thanks for spotting this.

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageItemUiStateHelper.kt
jd-alexander: Removed when I moved this to it's own use case 
aa8231a

WordPress/src/test/java/org/wordpress/android/viewmodel/pages/PageListViewModelTest.kt
jd-alexander: Created separate unit tests for the actions use case adfffae and one for the page list 3cbea5b 

N.B in a future commit the `uploadUiState` is being passed to the component so the most updated tests are here e036fa0

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageItemUiStateHelper.kt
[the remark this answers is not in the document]
jd-alexander: Resolved here 9da40ef

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageItemUiStateHelper.kt
jd-alexander: Resolved here e036fa0

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageItemUiStateHelper.kt
jd-alexander: Resolved here e036fa0

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/SearchListViewModel.kt
jd-alexander: Resolved in 78dda37

jd-alexander:
Hi @malinajirka this PR is ready for another review from a code standpoint! As discussed, we will need to resolve the scoping issue #11303 before this can be merged so I added the do not merge label just to be safe!

jd-alexander:
I merged #11315 into this and it works perfectly. Thanks!

WordPress/src/test/java/org/wordpress/android/viewmodel/pages/PageListViewModelTest.kt
malinajirka: 👍 Awesome! Btw thank you for asking these great questions! ;)

WordPress/src/test/java/org/wordpress/android/viewmodel/pages/CreatePageListItemActionsUseCaseTest.kt
```diff
@@ -0,0 +1,98 @@
...
+
+        // Act
+        val trashedActions = useCase.setupPageActions(TRASHED, mock())
+
+        // Assert
+        assertThat(trashedActions).isEqualTo(expectedActions)
+    }
+
+    @Test
+    fun `verify that SCHEDULED & PUBLISHED actions are the same`() {
```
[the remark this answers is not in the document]
jd-alexander: Yes, thanks for pointing this out. I would say that the actions being the same aren't necessarily a coincidence since a `Scheduled` page is just a page being `Published` at a later date 😄 Nonetheless, it's still good to have them separate since the business requirements could be different for both so I resolved this here 4e6dc61

WordPress/src/test/java/org/wordpress/android/viewmodel/pages/PageListViewModelTest.kt
```diff
@@ -317,14 +333,42 @@ class PageListViewModelTest : BaseUnitTest() {
         assertThat((result[0].first[1] as Page).showOverlay).isEqualTo(false)
     }
 
+    @Test
+    fun `verify PageListItemActionsUseCase generates correct Menu Actions for PublishedPage`() {
+        // Arrange
+        val actions = setOf(
```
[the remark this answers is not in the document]
jd-alexander: > We might want to consider replacing the actions with setOf(mock<PageItem.Action>()). This test is a bit confusing as it seems we are testing the specific actions but at the same time we are mocking the useCase so it returns the actions.

Yes, that's true. The `CreatePageListItemActionsUseCaseTest` verifies this so a mocked set can be returned. Thanks for spotting this. 

> I'd also consider renaming the name of the tests -> we are mocking what the usecase returns, so we are not verifying if PageListItemActionsUseCase generates correct actions.

Wdyt?

Yes, good idea. I mocked the actions and renamed the method! All is well now 4e6dc61

malinajirka:
Thanks for the changes! ;) 🚢

Issue #11119, opened 2020-01-21
Add support for search to "set a parent to a page"


malinajirka:
Goal of this task is to add search for "set page parent" screen.

1. Click on "My Site" in the bottom bar
2. Click on "Site Pages"
3. Click on the "three vertical dots" on one of the pages
4. Click on "Set Parent"
5. Notice the "Set Parent" screen doesn't have search

<img src="https://user-images.githubusercontent.com/2261188/72802725-b4663680-3c4c-11ea-9f17-f18e4146c59d.gif" width="300" />

PR #11269, opened 2020-02-09
Issue/11119 search in set parent page

zwarm:
**Fixes #**:  https://github.com/wordpress-mobile/WordPress-Android/issues/11119

**Description**:
This PR adds search capabilities to the Set Parent page. It is 95% complete. Following along with the
pattern in Site Pages, this implementation separates the search into a `PageParentSearchFragment`
and `PageParentSearchVieModel`, both of which have been added to the necc. dagger modules. A new layout file (`page_parent_fragment.xml`) was introduced for `PageParentFragment` to hold the FrameLayout for the search and page_parent_menu.xml was expanded to hold a new SearchView. The searching takes advantage of LiveData, ViewModels and observability to provide the user with a clean UI experience. 

While this PR is 95% complete, there are still a few lingering questions that may require action. 
1 - Does the text need to be adjusted when showing the suggestion or empty results? Right now that 
text is shared as ... 
    <string name="pages_empty_search_result">No pages matching your search</string>
    <string name="pages_search_suggestion">Search pages</string>
If changes are necessary, two new string would be required + any language translations required.

2 - As of this PR submittal, the search is locally performed on the parent page options during page load. It is unclear if the search needs to be performed against the database each and every time. If so, then the search logic will need to be updated to reflect that.

**Testing instructions:**  
Test starting prompt is shown 
- Navigate to Site pages
- Tap on a action menu (3 vertical dots) for a page and Select "Set Parent"
- Tap on the search icon 
- "Search hint" is available on the toolbar
- "Search pages" is visible in the page_parent_fragment

Test search text found
- Navigate to Site pages
- Tap on a action menu for a page and Select "Set Parent"
- Tap on the search icon 
- Enter search text (jot dowan a title snippet from the pages view & search for that)
- A list of possible parent pages is shown

Test search text found
- Navigate to Site pages
- Tap on a action menu for a page and Select "Set Parent"
- Tap on the search icon 
- Enter jibberish as the search text
- "No pages matching your search" is shown

Screenshot_1581281732
Screenshot_1581281746
Screenshot_1581281755
Screenshot_1581281775

search-set-page-parent

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageParentViewModel.kt
```diff
@@ -128,4 +152,74 @@ class PageParentViewModel
...
+        } else {
+            clearSearch()
+        }
+    }
+
+    private suspend fun search(
+        searchQuery: String
+    ): List<PageItem> = withContext(defaultDispatcher) {
+        val tempPages = _pages.value
+        if (tempPages != null) {
```
malinajirka: NitPick: We might want to consider using `let` here.

```
_pages.value?.let {
...
}
```
Wdyt?

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageParentViewModel.kt
```diff
@@ -128,4 +152,74 @@ class PageParentViewModel
...
+            if (tempPages.isNotEmpty()) {
+                return@withContext tempPages
+                        .filterIsInstance(ParentPage::class.java)
```
malinajirka: AFAIK unless there is a bug, this shouldn't filter out anything, right? If that's the case we might want to consider using `.also` and logging an exception if the list contains a non-parent page. The current implementation might just hide a bug which would be more difficult to find, wdyt? 

SideNote: Same thing probably applies to `it.id != 0L`.

WordPress/src/main/java/org/wordpress/android/ui/pages/PageParentFragment.kt
```diff
@@ -76,10 +86,63 @@ class PageParentFragment : Fragment() {
...
+                } else {
+                    viewModel.onSearch(newText)
+                }
+                return true
+            }
+        })
+
+        val searchEditFrame = searchAction?.actionView.findViewById<LinearLayout>(R.id.search_edit_frame)
+        (searchEditFrame.layoutParams as LinearLayout.LayoutParams)
+                .apply { this.leftMargin = DisplayUtils.dpToPx(activity, -8) }
```
malinajirka: We might want to consider moving the value into dimens.xml, wdyt?

SideNote: If you decide to move it to dimens.xml feel free to update even PagesFragment. Thanks!

zwarm:
@malinajirka 

> Have you considered filtering the items directly in PageParentViewModel/PageParentFragment? I'm not saying the approach you took isn't good. I understand you decided to prefer consistency. I'd just love to know your thoughts about it.

Yes I did, I had coded both solutions and in the end decided to PR this one to maintain consistency. I can just as easily kill this PR & submit the other instead. lmk

WordPress/src/test/java/org/wordpress/android/viewmodel/pages/PageParentSearchViewModelTest.kt
```diff
@@ -0,0 +1,69 @@
...
+import org.wordpress.android.ui.pages.PageItem.Empty
+import org.wordpress.android.viewmodel.ResourceProvider
+import java.util.Date
+
+@RunWith(MockitoJUnitRunner::class)
+class PageParentSearchViewModelTest {
+    @Rule
+    @JvmField val rule = InstantTaskExecutorRule()
+
+    @Mock lateinit var resourceProvider: ResourceProvider
```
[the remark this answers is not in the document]
zwarm: @malinajirka - Good catch, don't need them for the tests

WordPress/src/main/java/org/wordpress/android/ui/pages/PageParentSearchViewModel.kt
```diff
@@ -0,0 +1,60 @@
+package org.wordpress.android.viewmodel.pages
```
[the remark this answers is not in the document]
zwarm: Moved & pushed

WordPress/src/main/java/org/wordpress/android/ui/pages/PageParentFragment.kt
zwarm: @malinajirka - didn't know the extent of my reach. I'll make this change.

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageParentViewModel.kt
```diff
@@ -128,4 +152,74 @@ class PageParentViewModel
...
+                                            Locale.ROOT
+                                    )
+                            )
+                        }
+            }
+        }
+        return@withContext mutableListOf<PageItem>()
+    }
+
+    private fun <T> MutableLiveData<T>.postOnUi(value: T) {
```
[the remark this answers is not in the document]
zwarm: It is my understanding is that setValue() must be called from the main thread whereas postValue() is used when you want to set a message from a background thread. In this case I want to ensure that the change is posted on the main thread and observers are notified accordingly.

WordPress/src/main/java/org/wordpress/android/ui/pages/PageParentFragment.kt
```diff
@@ -53,7 +60,10 @@ class PageParentFragment : Fragment() {
         } else if (item.itemId == R.id.save_parent) {
             viewModel.onSaveButtonTapped()
             return true
+        } else if (item.itemId == R.id.action_search) {
+            return true
```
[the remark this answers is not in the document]
zwarm: This line is not needed. Missed setting a TODO and inadvertently left it in the code. Fixed and pushed

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageParentViewModel.kt
zwarm: NitPick accepted :) Changed & pushed

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageParentViewModel.kt
zwarm: Actually I have that filter in there to get rid of any "Divider" lines which are not ParentPage types. The Top Level Parent item is a ParentPage, but can be identified by id of "0". I had made the assumption that neither of those items should be searchable, so I exclude them. So it is possible for the items to contain non ParentPages and that doesn't mean it's a bug.

WordPress/src/main/java/org/wordpress/android/ui/pages/PageParentFragment.kt
zwarm: Actually I take that back, this is not a dimension per se, but an int interpretation. The DisplayUtils.dpToPx() takes an int value and uses the display metrics for calculation. I am leaving this one as is, but I am a bit rusty on converting dpToPx, so please advise.

zwarm:
I need some clarification around expected behavior upon searching .... Thanks!

> Save button state gets lost on configuration change (This might not be related to this PR. In case it isn't related, please create a new issue on GitHub)
> Open Page list
> Click on Set Page
> Select one of the pages -> notice "Save" button appears
> Click on Search icon
> Click on the Up button
> Notice "Save" button is invisible even though the page is still selected


**Scenario #1**
- Tap "set parent" to show view
- Initial parent is shown as selected => Save is GONE
- Select a different parent => Save is VISIBLE
- Tap search icon => **Should Save be VISIBLE or GONE during search?**
- Tap UP => Save should be VISIBLE

**Scenario #2** 
- Tap "set parent" to show view
- Initial parent is shown as selected => Save is GONE
- Tap search icon => Save is GONE
- Tap UP => Save should be GONE

**Scenario #3** 
- Tap "set parent" to show view
- Initial parent is shown as selected => Save is GONE
- Tap search icon => Save is GONE
- Enter search text => Save is GONE
- Select a parent => Save is VISIBLE
- "X" is tapped to clear search
   -- **What happens here? Is the selected item cleared? Is Save visible or Gone**
- Tap UP => **Should save be visible or not?**

Scenario #4 
- Tap "set parent" to show view
- Initial parent is shown as selected => Save is GONE
- Tap search icon => Save is GONE
- Enter search text => Save is GONE
- "X" is tapped to clear search
- Tap back => Save is GONE

Scenario #5 (Happens in production)
- Tap "set parent" to show view
- Initial parent is shown as selected => Save is GONE
- Tap item to select as parent => Save is VISIBLE
- Tap initial parent again => **Save is still VISIBLE (I would think that save should be gone if selection matches initial value)**

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageParentViewModel.kt
```diff
@@ -128,4 +152,74 @@ class PageParentViewModel
...
+        searchQuery: String
+    ): List<PageItem> = withContext(defaultDispatcher) {
+        val tempPages = _pages.value
+        if (tempPages != null) {
+            if (tempPages.isNotEmpty()) {
+                return@withContext tempPages
+                        .filterIsInstance(ParentPage::class.java)
+                        .filter { it.id != 0L }
+                        .filter {
+                            it.title.toLowerCase(Locale.ROOT).contains(
```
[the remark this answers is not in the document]
zwarm: I think you make a very good point and the safer solution is not to convert to lowercase, but to use contains with ignoreCase = true. Pushed.

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageParentViewModel.kt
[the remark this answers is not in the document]
malinajirka: Hmm, interesting. My understanding was that `setValue` needs to be called on the main thread as it's a blocking operation requiring UI thread. 
Whereas `postValue` doesn't need to be performed on the main thread as it's a non-blocking operation. 
However, even `postValue` eventually sets the value on the main thread, since LiveData are observed from the UI and we can't change the UI from a background thread.
So I'm still not sure  if our custom `postOnUi` actually does anything special or it's just a custom implementation of `postValue`. Am I missing something?

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageParentViewModel.kt
malinajirka: ohh, I see. Thanks for the explanation!

WordPress/src/main/java/org/wordpress/android/ui/pages/PageParentFragment.kt
malinajirka: I think replacing `DisplayUtils.dpToPx(activity, -8)` with `resources.getDimensionPixelSize(R.dimens.xyz)` where `xyz = -8dp` would have the same result. All `dp` values take into account display metrics. Having said all that, let's not make this change now. You have a good point that it's not just a matter of replacing `-8` with `R.dimens.xyz`.

malinajirka:
Thanks for all the improvements @zwarm ! The changes look great. I haven't tested the code as I wasn't sure if you plan to make changes to the Save button behavior.

>Yes I did, I had coded both solutions and in the end decided to PR this one to maintain consistency. I can just as easily kill this PR & submit the other instead. lmk

Ohh nice ;). I don't want you to replace the solution, I like it. I was just curious how did you make the decision. It's all clear now :). Thanks

--------------------------

Thanks for listing the different scenarios! ;)

>Scenario #1 - Should Save be VISIBLE or GONE during search?

I think Save should be gone.

>Scenario #3 -- What happens here? Is the selected item cleared? Is Save visible or Gone

I think Save should be gone.

>Scenario #3  Tap UP => Should save be visible or not?

I think Save should be gone.

>Scenario #5 Tap initial parent again => Save is still VISIBLE (I would think that save should be gone if selection matches initial value)

I agree with you. Could you please fill a bug report. Thanks!

To summarize it. I think Save should be visible when a non-initial item is selected and visible on the screen (Save should be also visible when the selected item is not visible due to a different scroll position).

-------------------------
SideNote:
A couple of tips on how to make the PR easier to review
- In general, we prefer to leave all the in-code discussions unresolved and let the reviewer resolve them. It's a bit cumbersome when the reviewer needs to open all the resolved discussions to double check they have been resolved as they expected and the fixes don't contain any bugs.
- I believe it's really useful when you leave a commit hash of the fix in the reply to an in-code comment -> example https://github.com/wordpress-mobile/WordPress-Android/pull/11228#discussion_r377274591. (If you just paste a plain commit hash, GitHub will turn it into a short-hash with a link.)

Wdyt?

zwarm:
@malinajirka - Still have one open question regarding "text" ... leave as is or needs to be something custom?
```
1 - Does the text need to be adjusted when showing the suggestion or empty results? Right now that
text is shared as ...
No pages matching your search
Search pages
If changes are necessary, two new string would be required + any language translations required.
```

zwarm:
```
SideNote:
A couple of tips on how to make the PR easier to review

In general, we prefer to leave all the in-code discussions unresolved and let the reviewer resolve them. It's a bit cumbersome when the reviewer needs to open all the resolved discussions to double check they have been resolved as they expected and the fixes don't contain any bugs.
I believe it's really useful when you leave a commit hash of the fix in the reply to an in-code comment -> example #11228 (comment). (If you just paste a plain commit hash, GitHub will turn it into a short-hash with a link.)
```
👍 - thanks for letting me know, will definitely follow this guideline from this point forward.

This will be adjusted as part of the Save button visibility changes for searching - can't really avoid fixing it if the logic is changing. Do you still want a separate ticket for it? 
`Scenario #5 Tap initial parent again => Save is still VISIBLE (I would think that save should be gone if selection matches initial value)`

malinajirka:
> Still have one open question regarding "text" ... leave as is or needs to be something custom?

Ohh sorry,  I forgot about that. I've discussed it with a designer and it's ok to keep it as it is.

>This will be adjusted as part of the Save button visibility changes for searching - can't really avoid fixing it if the logic is changing. Do you still want a separate ticket for it?

Nope, no need to create a separate ticket. If you expect it'll require more changes, feel free to open a new PR against this PR.

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageParentViewModel.kt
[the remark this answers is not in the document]
zwarm: Actually I didn't have the need for this code any longer, it has been removed.

zwarm:
@malinajirka 
Updated the save button behavior in [e9b4991](https://github.com/wordpress-mobile/WordPress-Android/pull/11269/commits/e9b499103c4abc9d450a6688c734c61f42212567)

I think should take care of everything you wanted done. Please let me know if there is a scenario I have not accounted for. Thanks.

For quick reference, this is what was implemented for the behavior

**Scenario #1**
Tap "set parent" to show view
Initial parent is shown as selected => Save is GONE
Select a different parent => Save is VISIBLE
Tap search icon => Save is GONE
Tap UP => Save is VISIBLE

**Scenario #2**
Tap "set parent" to show view
Initial parent is shown as selected => Save is GONE
Tap search icon => Save is GONE
Tap UP => Save is GONE

**Scenario #3**
Tap "set parent" to show view
Initial parent is shown as selected => Save is GONE
Tap search icon => Save is GONE
Enter search text => Save is GONE
Select a parent => Save is VISIBLE
"X" is tapped to clear search => Save is GONE
Tap UP => SAVE is VISIBLE 

**Scenario #4**
Tap "set parent" to show view
Initial parent is shown as selected => Save is GONE
Tap search icon => Save is GONE
Enter search text => Save is GONE
"X" is tapped to clear search
Tap UP => Save is GONE

**Scenario #5**  
Tap "set parent" to show view
Initial parent is shown as selected => Save is GONE
Tap item to select as parent => Save is VISIBLE
Tap initial parent again => Save is GONE 

**Note: If a value is selected while in search, that value remains selected when search is closed.

WordPress/src/main/java/org/wordpress/android/ui/pages/PageParentFragment.kt
```diff
@@ -222,12 +226,13 @@ class PageParentFragment : Fragment() {
...
         frameSearch.visibility = View.GONE
-        if (saveButton?.isVisible == true) {
-            saveButton?.isVisible = false
-        }
         if (myActionMenuItem.isActionViewExpanded) {
             myActionMenuItem.collapseActionView()
         }
+        /**Force the recyclerview to redraw if selection has changed while in search mode*/
+        if (viewModel.currentParent != viewModel.initialParent) {
+            recyclerView.adapter?.notifyDataSetChanged()
```
malinajirka: I wanted to propose moving this logic into the viewModel. Updating _pages with the updated list should force the view to redraw only the changed items, since the PageParentAdapter.update(..) is using DiffUtils. However, it unfortunately won't work. We currently share the list of PageItems between the VM and the adapter -> the PageItems are mutable and whenever we change their state (eg. isSelected field) in the vm it gets changed in the adapter (since they share the same instance). Therefore, the DiffUtils won't find any difference.

Having said all that, I don't think it's worth fixing at the moment as we'd need to fix it in both regular search + parent search. I just wanted to share another great example of why using mutable classes is risky and can easily cause unexpected side-effects.

zwarm: @malinajirka absolutely agree with you. DiffUtils won't work in this case.  It's a larger issue that needs to be addressed and I intentionally did not dive into (hence why you see this line for the trial). This list should be immutable. I may have even attempted a more reactive pattern to handle state changes more gracefully.

zwarm:
```
Select a parent (eg "Test page")
Tap on search
Type "Test"
Notice "Test Page" is visible and selected, but the SAVE button isn't (Should it be selected? Should the Save button be visible? Wdyt?)

Tap on search
Type something so there is at least one result
Rotate your device
Click on "Clear search"
Notice the results didn't disappear
I don't expect a real user will encounter either of the scenarios, wdyt?
```
@malinajirka - #1 - If an item is selected, it should stay selected. I can go either way on the save button. It sounds like that would be a designer/behavior decision. #2 Not a scenario I had thought of, but if you did then it's absolutely possible that customer may.

malinajirka:
Thanks for the reply ;)!

>#1 - If an item is selected, it should stay selected. I can go either way on the save button. It sounds like that would be a designer/behavior decision. #2 Not a scenario I had thought of, but if you did then it's absolutely possible that customer may.

I don't wont to delay the trial on such minor issues about which we are not even sure how to proceed. Could you please create GitHub issues for them. I'll add a low priority label to both of them and "needs design" label to the second one. Thank you 🙇 

The only remaining thing before I merge this PR is this comment - https://github.com/wordpress-mobile/WordPress-Android/pull/11269/files#r377068726. I'd rather avoid changing the Project.xml, wdyt?


SideNot:
Btw using #[number] syntax is reserved for referencing other issues/PRs on GitHub. Since issues with numbers 1-5 are quite old, it doesn't really matter. But I'd rather avoid using this syntax anyway as it redirects the user to a completely unrelated issue. Thanks!

zwarm:
Oops on the "pound" sign - didn't even cross my mind it would link out. I edited my post. I see the other comment, I'll revert it back.

zwarm:
@malinajirka [#d4000c0](https://github.com/wordpress-mobile/WordPress-Android/pull/11269/commits/d4000c0b24de40ee4d77129c37f192aee80a2fa4) Reverted back the .idea/codeStyles/Project.xml file to the commit before mine.

Obviously I needed to make a new commit reverting the file back to that state. I can’t just change history because it’s already been pushed. You know, editing history throws off everyone else’s stuff.

malinajirka:
>Oops on the "pound" sign - didn't even cross my mind it would link out. I edited my post. I see the other comment, I'll revert it back.

No worries ;)

> #d4000c0 Reverted back the .idea/codeStyles/Project.xml file to the commit before mine.
>Obviously I needed to make a new commit reverting the file back to that state. I can’t just change history because it’s already been pushed. You know, editing history throws off everyone else’s stuff.

👍 Yeah, revert is fine. Thanks!


LGTM 🚢

PR #11301, opened 2020-02-13
Fixing reader bottom sheet RTL support.

### LINKED ISSUE #11300: Reader bottom sheet RTL support.
### Expected behavior
When accessing to the reader filter bottom sheet we should get the `SITES` and `TAGS` lists mirrored and aligned to the view start. This should be true also for the title of the bottom sheet.

### Actual behavior
both the title of the bottom sheet and the list of SITES and TAGS are left aligned.

- With the zalpha flavor of the app, set an RTL language from `Me > App Settings > Interface Language` (for example Hebrew). 
- Open the reader bottom sheet and check you get something similar to the following

| image | image |
|---|---|

develric:
Fixes #11300

This PR fixes an issue with RTL support as per below images.

| Description | Wrong 👎 | Correct 👍 |
|---|---|---|
| `Following` title alignement and `SITES` list | image | image |
| `Following` title alignement and `TAGS` list | image | image |

## NOTES
- Some strings for Hebrew language are not set so we get english defaults but the alignements should be correct with this patch.

## To test
- Run the zalpha flavor of the app and check first with English that you get left alignment in the reader bottom sheet like below 

| image | image |
|---|---|

- Change now the language to a RTL language (like Hebrew) and check you get something similar to the `Correct 👍` column above.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/res/layout/subfilter_bottom_sheet.xml
+++ b/WordPress/src/main/res/layout/subfilter_bottom_sheet.xml
@@ -16,8 +16,7 @@
         android:id="@+id/title"
         style="@style/SiteTagBottomSheetTitle"
         android:layout_width="match_parent"
-        android:text="@string/reader_filter_main_title"
-        tools:text="Following"/>
+        android:text="@string/reader_filter_main_title"/>
 
     <LinearLayout
         android:layout_width="match_parent"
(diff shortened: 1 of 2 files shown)


develric:
Hey @renanferrari ! Thanks for reviewing this 😊!

> I was a bit confused if the ViewPager scrolling direction was correct...

Correct, I also needed to check it with behavior in other places across the app to confirm it was aligned.

> this was only addressed recently with ViewPager2, which we should definitely consider adopting in the future.

I was not aware that ViewPager2 was introducing RTL support, so good to know and thanks for sharing 🙇‍♂️! I agree we should consider to use it and opened an issue (#11328) to track it 👍.

PR #11302, opened 2020-02-13
Update Sentry sdk to v2.0

claucookie:
**Title:** Update Sentry SDK to v2.0 Release Candidate

**Description:** 
We want to be able to link Sentry events (unhandled or caught exceptions) to future Encrypted logs we are gonna submit to Wordpress API. 

This PR allows us to attach additional information to any Sentry event before it gets sent to Sentry server. 

- _Implemented already_: Logic to encrypt the logs #11230 #10738 
- _Not implemented yet_: The logic to generate the UUID and sending the encrypted log. (Next PR)

_Why not using Sentry v1.X?_  Because there is a race condition when sending the event for caught exceptions. If we try to attach additional data or tags or anything before the event is sent, it seems the logic is not waiting for the info to be attached and sends the event anyway without it.

_Why V2.0?_ With Sentry v2.0, we gain the ability to update the event with the future UUID we are going to generate in the next PR. E.g.
```
SentryAndroid.init(context.applicationContext) {
            it.dsn = BuildConfig.SENTRY_DSN
            it.beforeSend = BeforeSendCallback { event, hint ->
                // Todo: Update event with a UUID from the Logs we want to encrypt.
                return@BeforeSendCallback event
            }
        }
```

**Testing instructions:** 
```
# CrashloggingUtils.kt
        }

        return !hasUserOptedOut
    }
```
Build the project, deploy the app nd check out the logcat to see if the following messages are displayed:
```
2020-02-13 15:31:26.847 6970-6970/org.wordpress.android D/Sentry: Auto-init: false
2020-02-13 15:31:26.847 6970-6970/org.wordpress.android I/Sentry: Retrieving auto-init from AndroidManifest.xml
```
If there are no other errors, then it should be fine.

If we attach info to events, how is it gonna look like? Let's see a dummy `Logs` id sent as Additional Data
Screenshot 2020-02-12 at 13 14 11

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/util/CrashLoggingUtils.kt
+++ b/WordPress/src/main/java/org/wordpress/android/util/CrashLoggingUtils.kt
@@ -41,7 +41,6 @@ class CrashLoggingUtils {
         }
 
         @JvmStatic fun stopCrashLogging() {
-            Sentry.clearBreadcrumbs()
             Sentry.close()
         }
 
--- a/WordPress/src/main/java/org/wordpress/android/util/CrashLoggingUtils.kt
+++ b/WordPress/src/main/java/org/wordpress/android/util/CrashLoggingUtils.kt
@@ -7,6 +7,7 @@ import io.sentry.core.Breadcrumb
 import io.sentry.core.Sentry
 import io.sentry.core.SentryOptions.BeforeSendCallback
 import org.wordpress.android.BuildConfig
+import org.wordpress.android.R
 
 class CrashLoggingUtils {
     fun shouldEnableCrashLogging(context: Context): Boolean {
--- a/WordPress/src/main/java/org/wordpress/android/util/CrashLoggingUtils.kt
+++ b/WordPress/src/main/java/org/wordpress/android/util/CrashLoggingUtils.kt
@@ -1,14 +1,15 @@
 package org.wordpress.android.util
 
+import android.content.Context
 import android.preference.PreferenceManager
-import io.sentry.Sentry
-import io.sentry.android.AndroidSentryClientFactory
-import io.sentry.event.BreadcrumbBuilder
+import io.sentry.android.core.SentryAndroid
+import io.sentry.core.Breadcrumb
+import io.sentry.core.Sentry
+import io.sentry.core.SentryOptions.BeforeSendCallback
 import org.wordpress.android.BuildConfig
-import org.wordpress.android.R
 
 class CrashLoggingUtils {
-    fun shouldEnableCrashLogging(context: android.content.Context): Boolean {
+    fun shouldEnableCrashLogging(context: Context): Boolean {
         if (PackageUtils.isDebugBuild()) {
             return false
         }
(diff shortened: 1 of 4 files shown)


WordPress/src/main/java/org/wordpress/android/util/CrashLoggingUtils.kt
```diff
@@ -33,7 +41,7 @@ class CrashLoggingUtils {
         }
 
         @JvmStatic fun stopCrashLogging() {
-            Sentry.clearContext()
+            Sentry.clearBreadcrumbs()
```
jkmassel: I'm not sure we need to clear the breadcrumbs if we're closing down Sentry – in the older version clearing out the context would tear down Sentry completely so that we could re-init it later.

WDYT?

claucookie: That's a very good question. When I was looking for an alternative to `.clearContext()` I couldnt find any equivalent method in the v2.0.
Then, I checked what was doing inside to see if I could find a similar logic and I bumped into the implementation, which is calling this method :  
```

**Comment by claucookie** [comment]
@jkmassel Answering your other questions here :) 
>I'm not seeing any events coming in Sentry when I build and run the app in release mode, even if I override the shouldEnableCrashLogging callback. It seems to work fine in development mode on-device though if I do the same. Are you seeing the same issue?
I am going to run the release build and have a look.

>Does this configuration support multi-dex? I noted that the config suggestion from the docs wasn't present here – will that matter?
I missed that bit in the docs, not sure how. Anyway, We definitely use multidex in the project, but we haven't set up any rules for the release yet using:
`multiDexKeepProguard file(...)`
I think it would be a good idea to create a rules file and include the rules to keep the classes they mention.

**Comment by claucookie** [comment]
> @jkmassel Answering your other questions here :)
> 
> > I'm not seeing any events coming in Sentry when I build and run the app in release mode, even if I override the shouldEnableCrashLogging callback. It seems to work fine in development mode on-device though if I do the same. Are you seeing the same issue?
> 
> I am going to run the release build and have a look.
> 
> > Does this configuration support multi-dex? I noted that the config suggestion from the docs wasn't present here – will that matter?
> 
> I missed that bit in the docs, not sure how. Anyway, We definitely use multidex in the project, but we haven't set up any rules for the release yet using:
> `multiDexKeepProguard file(...)`
> I think it would be a good idea to create a rules file and include the rules to keep the classes they mention.

Sorry for my late response, regarding events coming in Sentry when building the release. Yes, I do see the  log messages from Sentry when running the release build:
```
2020-02-17 16:23:15.364 29448-29448/? D/Sentry: Auto-init: false
2020-02-17 16:23:15.364 29448-29448/? I/Sentry: Retrieving auto-init from AndroidManifest.xml
2020-02-17 16:25:56.729 29779-29779/? D/Sentry: Auto-init: false
2020-02-17 16:25:56.729 29779-29779/? I/Sentry: Retrieving auto-init from AndroidManifest.xml
2020-02-17 16:27:42.992 30892-30892/? D/Sentry: Auto-init: false
2020-02-17 16:27:42.992 30892-30892/? I/Sentry: Retrieving auto-init from AndroidManifest.xml
``` 

What's the configuration on your side? Did you update the `wp.sentry.dsn` property inside the `gradle. properties`? It might still have a dummy value in your local. 🤔

WordPress/src/main/java/org/wordpress/android/util/CrashLoggingUtils.kt
jkmassel: Sounds great!

jkmassel:
>What's the configuration on your side? Did you update the wp.sentry.dsn property inside the gradle. properties? It might still have a dummy value in your local. 🤔

It's working now! I think it might've been something to do with our production credentials system – I re-applied them and was able to get it consistently working. Sorry about that!!

PR #11325, opened 2020-02-15
IA Reader add dedicated tracking for time spend in sub-filtered streams.

### LINKED ISSUE #11324: Adding IA specific time spent in reader track.
We already added some tracking figures to measure time spent in reader (see #10956). 

We would like to add a dedicated figure to track the time spent in the reader while Sites/Tags sub-filtering is active.

develric:
Fixes #11324

This is an addition to the already merged PR #10956 and adds the `time_in_subfiltered_list` property to the already available `application_closed` event.

This is a specific event of the IA revised reader (zalpha flavor) and reports the seconds spent in the main reader `FOLLOWING` tab when a `SITE` or `TAG` is selected in the sub-filter.

## To test
- Unit tests must pass
- Be sure you have tracking turned on (Me > App Settings > Privacy Settings > Collect information must be active)
- you can observe logcat while making following steps, filtering with this `(TRACK READER|Tracked: application_closed)` and Regex active
- Open the zalpha flavor of the app
- Go to Following and select a subfilter Site/Tag (stay there for something like 5 seconds); 
- Move to a different tab (like Discover) and stay there for other 5 seconds
- Open a post from Discover list and stay there for other 5 seconds 
- Come back to Following and wait 5 seconds
- Clear the filter and wait 5 seconds
- Put the app in background
- Check that you get some message like 
`Tracked: application_closed, Properties: {"time_in_main_reader":20,"last_visible_screen":"Reader","time_in_reader_paged_post":5,"time_in_app":34,"time_in_subfiltered_list":10,"time_in_reader_filtered_list":0}`
and the reported timings are somewhat similar to the expected
- Smoke test the app selecting other features different from reader and check the reader timing does not accumulate

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
```diff
@@ -197,7 +200,24 @@ class ReaderPostListViewModel @Inject constructor(
...
+                readerTracker.start(ReaderTrackerType.SUBFILTERED_LIST)
+            } else if (!getCurrentSubfilterValue().isTrackedItem &&
+                    readerTracker.isRunning(ReaderTrackerType.SUBFILTERED_LIST)) {
+                AppLog.d(T.READER, "TRACK READER ReaderPostListFragment > STOP Count SUBFILTERED_LIST")
+                readerTracker.stop(ReaderTrackerType.SUBFILTERED_LIST)
+            }
+        } else if (readerTracker.isRunning(ReaderTrackerType.SUBFILTERED_LIST)) {
+            AppLog.d(T.READER, "TRACK READER ReaderPostListFragment > STOP Count SUBFILTERED_LIST")
+            readerTracker.stop(ReaderTrackerType.SUBFILTERED_LIST)
+        }
```
renanferrari: This is kind of hard to read at first glance. What do you think about extracting a couple of local variables to the top of the function to make it a bit easier to parse? 

I'm thinking something like this:

```kotlin
val isCurrentItemTracked = getCurrentSubfilterValue().isTrackedItem
val isTrackerReaderRunning = readerTracker.isRunning(ReaderTrackerType.SUBFILTERED_LIST)
if (show) {
    if (isCurrentItemTracked && !isTrackerReaderRunning) {
        // ...
    } else if (!isCurrentItemTracked && isTrackerReaderRunning) {
        // ...
    }
} else if (isCurrentItemTracked) {
        // ...
}
```

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
```diff
@@ -244,10 +264,19 @@ class ReaderPostListViewModel @Inject constructor(
...
         requestNewerPosts: Boolean
     ) {
+        if (subfilterListItem.isTrackedItem &&
+                            !readerTracker.isRunning(ReaderTrackerType.SUBFILTERED_LIST)) {
+            AppLog.d(T.READER, "TRACK READER ReaderPostListFragment > START Count SUBFILTERED_LIST")
+            readerTracker.start(ReaderTrackerType.SUBFILTERED_LIST)
+        } else if (!subfilterListItem.isTrackedItem && readerTracker.isRunning(ReaderTrackerType.SUBFILTERED_LIST)) {
+            AppLog.d(T.READER, "TRACK READER ReaderPostListFragment > STOP Count SUBFILTERED_LIST")
+            readerTracker.stop(ReaderTrackerType.SUBFILTERED_LIST)
+        }
```
renanferrari: My previous comment about the code being a bit hard to read could also be applied here.

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
```diff
@@ -295,6 +324,14 @@ class ReaderPostListViewModel @Inject constructor(
         isFirstLoad = false
     }
 
+    fun onSubfilterChanged(subfilterListItem: SubfilterListItem) {
+        changeSubfilter(subfilterListItem, true)
+    }
+
+    fun manageSubfilter() {
+        changeSubfilter(getCurrentSubfilterValue(), false)
+    }
```
renanferrari: It took me a while to understand what `manageSubfilter()` actually does and I think a more descriptive name could help.

How about renaming these functions to something like:
-  `onSubfilterChanged` to `onSubfilterSelected`
-  `manageSubfilter ` to `onSubfilterReselected`

WordPress/src/main/java/org/wordpress/android/ui/reader/tracker/ReaderTracker.kt
```diff
@@ -45,6 +45,12 @@ class ReaderTracker @Inject constructor(private val dateProvider: DateProvider)
         }
     }
 
+    fun isRunning(type: ReaderTrackerType): Boolean {
+        return trackers[type]?.let { trackerInto ->
+            trackerInto.startDate != null
+        } ?: false
```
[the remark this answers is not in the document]
develric: Correct. Changes [here](7369d1a3224faf9d1555984dff7d8d46eba10d1b)

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
develric: Agree 😊! Changed it [here](56f01bceb31845b0c9a545c26dedf0a39fd75fc7)

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
develric: Same as above [here](56f01bceb31845b0c9a545c26dedf0a39fd75fc7)

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
develric: Makes sense. Changed it [here](09f76fb894bbe9e6dc16f762f0d44e01d16c4ae8) and [here](ac146f1ff94197769160de6644e4c2a3d16de2c6)  👍

develric:
Hi @renanferrari, thanks for the review and suggestions 😊. I should have addressed your comments. I also merged back develop and resolve some conflicts so I would appreciate if you could make a quick re-test run on this also from your side. Thanks in advance 🙇‍♂️!

PR #11332, opened 2020-02-18
Issue/11271 add post to page model

### LINKED ISSUE #11271: Offline Support - Pages : Add PostModel reference to PageModel
### Expected behavior
The `PostModel` for each page should be available across [PagesViewModel](https://github.com/wordpress-mobile/WordPress-Android/blob/59778060ca0dd6b716ba60ea8a727fa6a0d4137c/WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PagesViewModel.kt) and [PagesListViewModel](https://github.com/wordpress-mobile/WordPress-Android/blob/2f8c4efb962da6c91f2dfe85fbdb93d0f877eab0/WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageListViewModel.kt) so that functionality and components that need to be shared between the `Pages` and `Posts` can be done in an efficient manner since the APIs would contain `PostModel` as a parameter in some cases. 

### Actual behavior
Currently, `PageModel` is being used to create the UI and behavior for the Page List functionality. It's created by mapping a `PostModel` to its properties. The issue with this approach is that this model worked for the current implementation but when further details of the post are needed for certain APIs or functionality a call has to be made to the `PostStore` which isn't performant in many cases. 

There are several related issues where `PostModel` fetching is affecting performance and at times causing `NullPointers` such as https://github.com/wordpress-mobile/WordPress-Android/issues/11233 , https://github.com/wordpress-mobile/WordPress-Android/issues/11260 , https://github.com/wordpress-mobile/WordPress-Android/pull/11266

malinajirka:
Fixes #11271

This PR fixes several TODOs. PageModel didn't have reference to the PostModel. We decided to temporarily load PostModel from database whenever we needed it in order to utilize existing methods which worked only with PostModel. However, this approach had significant performance impact.
We added PostModel field into PageModel in [FluxC's PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1496) so we don't need to load the PostModel from database anymore.

Note:
It seems the progress state and item's ui is sometimes temporarily de-synced (there is +- 3 second delay). So when I for example publish a post, the progress indicator disappears as soon as the post gets published, but the PageItem gets updated after a delay. Therefore "Post will be published when your device is back online" label is shown even though the post has already been published. @jd-alexander If I recall correctly we've already been dealing with this issue before. Any ideas what is causing the delay? (I've looked into the issue, and the root cause seems to be that we update `_pages` livedate multiple times within a short time period. The UI receives only the latest update which might unfortunately contain out-dated values, if `invalidatePages` finished after `refreshPages`. Either case I believe the issue isn't related to this PR.)

To test:
We are merging this PR into a feature branch. Since this branch will be heavily tested before it's merged, we don't need to perform any deep tests in this PR. However, I think it'd be great if you smoke tested the PageList. For example try editing, trashing, restoring, publishing a post. Also try to scroll the page list.

PR submission checklist:


jd-alexander:
Thanks for these changes Jirka! I did testing on the functions below and more and I didn't notice any issues. 
> editing, trashing, restoring, publishing a post

> Note:
It seems the progress state and item's ui is sometimes temporarily de-synced (there is +- 3 second delay). So when I for example publish a post, the progress indicator disappears as soon as the post gets published, but the PageItem gets updated after a delay. Therefore "Post will be published when your device is back online" label is shown even though the post has already been published. @jd-alexander If I recall correctly we've already been dealing with this issue before. Any ideas what is causing the delay? (I've looked into the issue, and the root cause seems to be that we update _pages livedate multiple times within a short time period. The UI receives only the latest update which might unfortunately contain out-dated values, if invalidatePages finished after refreshPages. Either case I believe the issue isn't related to this PR.

Thanks. I am going to be investigating this. I noticed this again as well.

LGTM 🚢since it addresses the issue we have been having for a while now!

Issue #11129, opened 2020-01-22
Enqueue page to UploadWorker when the user tries to upload it while offline


malinajirka:
When the user tries to upload a post in offline, we enqueue the post to UploadWorker which is run periodically when certain conditions are met. Make sure the same behavior works for pages.

Note: I think it might start working automatically when #11126 is implemented. The logic for posts is [here](https://github.com/wordpress-mobile/WordPress-Android/blob/76df20a836872dc5ed186ec4d5f98d679368368b/WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadUtils.java#L169) and if I'm not wrong, the same code is executed for pages.

jd-alexander:
You were right. I ran a test and the page was enqueued. I also utilized the debugger to ensure that this [line](https://github.com/wordpress-mobile/WordPress-Android/blob/76df20a836872dc5ed186ec4d5f98d679368368b/WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadUtils.java#L169) was hit. 

These were the steps I followed. 
1. Opened the app while it was in airplane mode. 
2. On the `My Site` tab I clicked the create FAB, selected `Site Page`.
3. I created a new page and clicked publish. 
4. Once I re-enabled the internet, the page was uploaded.

malinajirka:
Re-opening this issue, since the page doesn't get enqueued when it's created/published/saved from PagesList.

jd-alexander:
Oh, thanks for that. I never realized this was supposed to be done on the `PagesList` as well 🙏

PR #11335, opened 2020-02-18
Issue/11177 update snackbar msgs 1

### LINKED ISSUE #11177: Update snackbar messages

malinajirka:
Fixes https://github.com/wordpress-mobile/WordPress-Android/issues/11129, partially fixes #11177

Review instructions
1. ~Review and merge https://github.com/wordpress-mobile/WordPress-Android/pull/11332~
2. ~Update target branch to `feature/master-pages-offline-support`~
3. Review and merge this PR

To test:
1. My Site
2. Page List
3. Turn on airplane mode
4. Open an existing page
5. Make changes to its content
6. Press back
7. Notice "Device is offline. Paged saved locally" snackbar message

Repeat the above steps but in step 6 click on "Publish/Update/Save/Schedule". Notice "We'll publish/schedule/save .... " message.

Repeat the testing steps and make sure [this line](https://github.com/wordpress-mobile/WordPress-Android/blob/69bd622c1f8a914a15a49d9c05b940117a0060b6/WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadUtils.java#L178) gets hit.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/pages/PagesFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/pages/PagesFragment.kt
@@ -48,6 +48,9 @@ import org.wordpress.android.ui.posts.PreviewStateHelper
 import org.wordpress.android.ui.posts.ProgressDialogHelper
 import org.wordpress.android.ui.posts.RemotePreviewLogicHelper
 import org.wordpress.android.ui.quickstart.QuickStartEvent
+import org.wordpress.android.ui.uploads.UploadActionUseCase
+import org.wordpress.android.ui.uploads.UploadUtils
+import org.wordpress.android.ui.uploads.UploadUtilsWrapper
 import org.wordpress.android.ui.utils.UiHelpers
 import org.wordpress.android.util.DisplayUtils
 import org.wordpress.android.util.QuickStartUtils
@@ -85,6 +88,8 @@ class PagesFragment : Fragment() {
     @Inject lateinit var remotePreviewLogicHelper: RemotePreviewLogicHelper
     @Inject lateinit var previewStateHelper: PreviewStateHelper
     @Inject lateinit var progressDialogHelper: ProgressDialogHelper
+    @Inject internal lateinit var uploadActionUseCase: UploadActionUseCase
+    @Inject internal lateinit var uploadUtilsWrapper: UploadUtilsWrapper
     private var quickStartEvent: QuickStartEvent? = null
     private var progressDialog: ProgressDialog? = null
 
@@ -119,8 +124,6 @@ class PagesFragment : Fragment() {
 
     override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
         if (requestCode == RequestCodes.EDIT_POST && resultCode == Activity.RESULT_OK && data != null) {
-            val pageId = data.getLongExtra(EditPostActivity.EXTRA_POST_REMOTE_ID, -1)
-
             if (EditPostActivity.checkToRestart(data)) {
                 ActivityLauncher.editPageForResult(data, this@PagesFragment, viewModel.site,
                         data.getIntExtra(EditPostActivity.EXTRA_POST_LOCAL_ID, 0), false)
@@ -129,8 +132,9 @@ class PagesFragment : Fragment() {
                 return
             }
 
-            if (pageId != -1L) {
-                viewModel.onPageEditFinished()
+            val localPageId = data.getIntExtra(EditPostActivity.EXTRA_POST_LOCAL_ID, -1)
+            if (localPageId != -1) {
+                viewModel.onPageEditFinished(localPageId, data)
             }
         } else if (requestCode == RequestCodes.PAGE_PARENT && resultCode == Activity.RESULT_OK && data != null) {
             val parentId = data.getLongExtra(EXTRA_PAGE_PARENT_ID_KEY, -1)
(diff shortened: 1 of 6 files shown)


WordPress/src/test/java/org/wordpress/android/viewmodel/pages/PagesViewModelTest.kt
```diff
@@ -215,6 +219,24 @@ class PagesViewModelTest {
         verify(uploadStarter, times(2)).queueUploadFromSite(site)
     }
 
+    @Test
+    fun `postUploadAction invoked on edit post activity result`() = test {
+        // Given
+        val localPageId = 999
+        val intent = mock<Intent>()
+        val postModel = PostModel().apply { setStatus(PostStatus.PUBLISHED.toString()) }
+        val siteModel = SiteModel()
```
jd-alexander: Hi Jirka! Quick question. Why did you use a different `SiteModel` here instead of using the `site` that's already created? Let me know your thoughts. Thanks 😄

WordPress/src/main/java/org/wordpress/android/ui/pages/PagesFragment.kt
```diff
@@ -85,6 +88,8 @@ class PagesFragment : Fragment() {
     @Inject lateinit var remotePreviewLogicHelper: RemotePreviewLogicHelper
     @Inject lateinit var previewStateHelper: PreviewStateHelper
     @Inject lateinit var progressDialogHelper: ProgressDialogHelper
+    @Inject internal lateinit var uploadActionUseCase: UploadActionUseCase
+    @Inject internal lateinit var uploadUtilsWrapper: UploadUtilsWrapper
```
jd-alexander: Quick question Jirka. What was your thought process for using the `internal` modifier here? Are these objects gonna be used in other places within this feature set implementation or it has something to do with Dagger? Let me know. 😀

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PagesViewModel.kt
```diff
@@ -250,9 +254,14 @@ class PagesViewModel
...
+    fun onPageEditFinished(localPageId: Int, data: Intent) {
         launch {
             refreshPages() // show local changes immediately
+            withContext(defaultDispatcher) {
+                pageStore.getPageByLocalId(pageId = localPageId, site = site)?.let {
+                    _postUploadAction.postValue(Triple(it.post, it.site, data))
+                }
+            }
         }
     }
```
jd-alexander: I am still getting a solid idea of how the `pageStore` component workss. Since the `refreshPages` was called above could we utilize that `PagedModel` instead of making another call to the `pageStore`. Something like this 

```kotlin
fun onPageEditFinished(remotePageId: Long, data: Intent) {
        launch {
            refreshPages() // show local changes immediately
            withContext(defaultDispatcher) {
                pageMap[remotePageId]?.let {
                    _postUploadAction.postValue(Triple(it.post, it.site, data))
                }
            }
        }
    }
```
Or is there a reason we need to fetch it from the `pageStore`. Let me know if it makes sense or not 😅

WordPress/src/test/java/org/wordpress/android/viewmodel/pages/PagesViewModelTest.kt
malinajirka: Hi Joel :). 

> Why did you use a different SiteModel here instead of using the site that's already created?

:D Simple answer, because I overlooked it:P. Fixed in a3bf1776a423d8b30d6e0211e180608a032f4970

WordPress/src/main/java/org/wordpress/android/ui/pages/PagesFragment.kt
malinajirka: I copy pasted the method and didn't notice we don't use "internal" in this fragment.

Dagger doesn't support injection into `private` fields. Best practice is to use the most strict modifier available, which in Kotlin means `internal`. I don't think it matter much since this is not a public API and I don't assume anyone will touch Fragment's fields anyway. But I believe that's the reason why we somewhere use `internal`. Removed in a3bf1776a423d8b30d6e0211e180608a032f4970

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PagesViewModel.kt
malinajirka: Great question! I tried the approach you suggested and right after I committed the changes I remembered why I didn't use that approach yesterday :D. LocalDrafts don't have remoteId since they weren't uploaded to the server yet. However, we want to emit the postUploadAction even for those. We could try to get the PageModel from `pageMap` with remoteId and if it's not there fetch it from database, but it seems like a super minor optimisation which isn't worth it imo. If we were loading the data from database while scrolling or something like that, it'd make total sense. But we hit the database just once on a background thread when the user closes the editor.

First attempt committed in a3bf1776a423d8b30d6e0211e180608a032f4970. Partial revert committed in f9b6a52168c8406b7292cc592eaecf81164b4621.

malinajirka:
Thanks for the great review @jd-alexander ;)! I've tried to answer to all your questions. It's ready for another round :).

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PagesViewModel.kt
jd-alexander: > Great question! I tried the approach you suggested and right after I committed the changes I remembered why I didn't use that approach yesterday :D. LocalDrafts don't have remoteId since they weren't uploaded to the server yet.

Thanks for this! Yes, that's true. I got a `NullPointerException` when the `remoteId` was missing while working on a PR and had to switch to the `localId` 🙏

WordPress/src/test/java/org/wordpress/android/viewmodel/pages/PagesViewModelTest.kt
jd-alexander: Awesome 😉

WordPress/src/main/java/org/wordpress/android/ui/pages/PagesFragment.kt
jd-alexander: > Dagger doesn't support injection into private fields. Best practice is to use the most strict modifier available, which in Kotlin means internal. I don't think it matter much since this is not a public API and I don't assume anyone will touch Fragment's fields anyway. But I believe that's the reason why we somewhere use internal. Removed in a3bf177

Yes, good point! Thanks for this explanation!

PR #11336, opened 2020-02-18
Send encrypted logs to Wordpress API

claucookie:
Note: This PR is a Draft until this PR https://github.com/wordpress-mobile/WordPress-Android/pull/11302 gets merged.

**Title:** Send encrypted logs to Wordpress API

**Description:** 
Apart from Sending the Logs Id to Sentry, We want to send the actual encrypted Logs to our Wordpress API.

In this PR we have made some progress towards this and a working implementation is able to send the encrypted logs to the current Testing Wordpress API:

**Summary of What's done in this PR:**
- ✅ Implemented`LogEncryptionTestingActions.sendEncryptedLogsFile(...)` as a POST request to send the JSON object containing the encrypted logs information.  
- ✅ Updated `EncryptionUtils.java` to include the field `logsId` populated using a UUID V3 method https://en.wikipedia.org/wiki/Universally_unique_identifier#Versions_3_and_5_(namespace_name-based). The idea of using a namespace is to avoid collisions when generating the UUIDs before sending them to Wordrpess API and Sentry.
- ✅ Created `EncryptionUtils.getLogsUUID(...)` which returns the value of the `logsId`, this is used when we mutate the Sentry event just before it gets sent.
- ✅ Updated `CrashLoggingUtils.enableCrashLogging(...)` to perform the following operations before the Sentry event gets sent:
   - Get recent logs and encrypt them.
   - Send encrypted logs to WordPress API.
   - Attach the `LogsId` to the Sentry event as additional information or `extra`.
Screenshot 2020-02-18 at 16 29 19

- ✅ Moved the Encryption related classes to a separate package called `Encryption`.

**Pending stuff for future PRs:**
- 🚧 Update backend logic to read the new field `logsId`.
- 🚧 Update endpoint url. @jkmassel At the moment We are using the testing endpoint, so when the real one gets implemented we might need to update this.
- 🚧 Solve queue issue with Volley. When an unhandled exception closes the app the Volley request that sends the encrypted logs get successfully started but the response is not received. 
🤔 We might need a caching strategy to avoid losing requests when the app gets closed. Or use an intent service to send the request? WDYT?

```kotlin
// LogEncryptionTestingActions.kt

fun sendEncryptedLogsFile(encryptedLogsJson: JSONObject) {
     val jsonRequest = JsonStringRequest(
             Request.Method.POST,
             ENCRYPTION_LOGS_SERVER_URL,
             encryptedLogsJson,
             Listener {   // <-- This success listener never gets called for unhandled exception
AppLog.i(AppLog.T.API, "Encrypted logs response successfully received by Wordpress API") 
             },
             ErrorListener { AppLog.e(AppLog.T.API, "An error occurred when sending logs. " + it.message) }
     )
     WordPress.sRequestQueue.add(jsonRequest)
     AppLog.i(AppLog.T.API, "Sending encrypted logs to Wordpress API") // <-- I see this in the logcat
    }
```

**Testing instructions:** 
```kotlin
// CrashloggingUtils.kt
...
        }

        return !hasUserOptedOut
    }
```
Update the `gradle.properties` file with your Sentry DSN and the base64 public key:
```
wp.sentry.dsn=https://000000000000@sentry.io/0000000
wp.encryption.public_key = XXXXXXXXXXXXXXXXXXXXXXXXXXXXX
```
Log an exception inside the Login or HelpActivity by adding this line:
```kotlin
// HelpActivity.kt

CrashloggingUtils.log(RuntimeException("This is a new exception"))
```

Build the project, deploy the app and go to the Activity where you added the exception.

*PR submission checklist:*

### DIFF
--- a/WordPress/src/androidTest/java/org/wordpress/android/util/encryption/EncryptionUtilsTest.java
+++ b/WordPress/src/androidTest/java/org/wordpress/android/util/encryption/EncryptionUtilsTest.java
@@ -11,7 +11,6 @@ import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.libsodium.jni.NaCl;
-import org.libsodium.jni.Sodium;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -59,7 +58,7 @@ public class EncryptionUtilsTest {
     public void setup() {
         mPublicKey = new byte[BOX_PUBLIC_KEY_BYTES];
         mSecretKey = new byte[BOX_SECRET_KEY_BYTES];
-        Sodium.crypto_box_keypair(mPublicKey, mSecretKey);
+        NaCl.sodium().crypto_box_keypair(mPublicKey, mSecretKey);
     }
 
     @Test
@@ -92,7 +91,7 @@ public class EncryptionUtilsTest {
         assertNotNull(header);
 
         final byte[] state = new byte[EncryptionUtils.STATEBYTES];
-        final int initPullReturnCode = Sodium.crypto_secretstream_xchacha20poly1305_init_pull(
+        final int initPullReturnCode = NaCl.sodium().crypto_secretstream_xchacha20poly1305_init_pull(
                 state,
                 header,
                 dataSpecificKey);
@@ -131,7 +130,7 @@ public class EncryptionUtilsTest {
             final byte[] decryptedKey = new byte[EncryptionUtils.KEYBYTES];
             final String encryptedKeyBase64 = encryptionDataJson.getString("encryptedKey");
             final byte[] encryptedKey = Base64.decode(encryptedKeyBase64, BASE64_DECODE_FLAGS);
-            final int returnCode = Sodium.crypto_box_seal_open(
+            final int returnCode = NaCl.sodium().crypto_box_seal_open(
                     decryptedKey,
                     encryptedKey,
                     EncryptionUtils.KEYBYTES + EncryptionUtils.BOX_SEALBYTES,
@@ -182,7 +181,7 @@ public class EncryptionUtilsTest {
         final byte[] additionalData = new byte[0]; // opting not to use this value
         final int additionalDataLength = 0;
         final int[] decryptedLineLengthOutput = new int[0]; // opting not to get this value
-        final int returnCode = Sodium.crypto_secretstream_xchacha20poly1305_pull(
+        final int returnCode = NaCl.sodium().crypto_secretstream_xchacha20poly1305_pull(
                 state,
                 decryptedLine,
                 decryptedLineLengthOutput,
--- a/WordPress/src/androidTest/java/org/wordpress/android/util/encryption/EncryptionUtilsTest.java
+++ b/WordPress/src/androidTest/java/org/wordpress/android/util/encryption/EncryptionUtilsTest.java
@@ -15,10 +15,14 @@ import org.libsodium.jni.NaCl;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.UUID;
 
 import static junit.framework.Assert.fail;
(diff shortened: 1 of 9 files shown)


WordPress/src/androidTest/java/org/wordpress/android/util/encryption/EncryptionUtilsTest.java
```diff
@@ -77,6 +80,29 @@ public void testCharacterSampleEncryptionResultIsValid() {
...
 
+    @Test
+    public void testUUIDAreDifferent() {
+        String source1 = "v1" + "android" + System.currentTimeMillis();
+        String uuid1 = UUID.nameUUIDFromBytes(source1.getBytes()).toString();
+
+        String source2 = "v1" + "android" + System.currentTimeMillis();
+        String uuid2 = UUID.nameUUIDFromBytes(source2.getBytes()).toString();
+
+        assertFalse(uuid1.equals(uuid2));
```
jkmassel: It seems to me these two tests are testing the implementation of `UUID.nameUUIDFromBytes` more than they are testing the functionality of this class. WDYT?

WordPress/src/main/java/org/wordpress/android/util/encryption/EncryptionUtils.java
```diff
@@ -43,16 +54,21 @@ public static String generateJSONEncryptedLogs(final byte[] publicKey,
                                                    final List<String> logMessages) throws JSONException {
         // Schema version
         JSONObject encryptionDataJson = new JSONObject();
-        encryptionDataJson.put("keyedWith", KEYED_WITH);
+        encryptionDataJson.put(FIELD_KEYED_WITH, KEYED_WITH);
+
+        // Generate Logs UUID
+        String source = KEYED_WITH + ANDROID_PLATFORM + System.currentTimeMillis();
+        byte[] bytes = source.getBytes();
+        encryptionDataJson.put(FIELD_LOGS_ID, UUID.nameUUIDFromBytes(bytes));
```
jkmassel: This looks like it produces a solid UUIDv3, and it's exceedingly unlikely we'd ever call this method twice in the same millisecond. Out of curiosity though, why not use a UUIDv4 and avoid the concatenation and byte wrangling?

claucookie: Good question! I was reading about V3 and V4 and I ended up in this blog post where they explain what are the chances of creating unique codes https://towardsdatascience.com/are-uuids-really-unique-57eb80fc2a87

They mention it is quite good for practical purposes, which is true, but since we are potentially using this logic in other apps and platforms I thought it was a good idea to have more control over it and remove the tiny bit of chance of getting a duplicate.

Does it make sense? :)

WordPress/src/androidTest/java/org/wordpress/android/util/encryption/EncryptionUtilsTest.java
claucookie: Yes, you are totally right. :) Good spot! 

Actually, I created them while playing with UUID to be able to test the result from `UUID.nameUUIDFromBytes` running multiple times.  I should remove them and update the existing tests to accommodate the new logic.

claucookie:
@jkmassel I'm closing this draft and opening a new one https://github.com/wordpress-mobile/WordPress-Android/pull/11341

The reason is that I forgot to sync my fork with Wordpress develop before cherry-picking a pair of commits from yesterday and the history was a bit dirty. 

Sorry about that, since you already added comments on this one.

claucookie:
@jkmassel I'm closing this draft and opening a new one https://github.com/wordpress-mobile/WordPress-Android/pull/11341

The reason is that I forgot to sync my fork with Wordpress develop before cherry-picking a pair of commits from yesterday and the history was a bit dirty. 

Sorry about that, since you already added comments on this one.

PR #11340, opened 2020-02-19
Fix page list upload state

### LINKED ISSUE #11263: Offline Support : Pages - PostUploadComplete logic issue
### Expected behavior
When the page is uploaded and it needs to be refreshed to show the changes that took place, it should simply run a refresh from the database since it was updated and we are getting the completed event. 

https://github.com/wordpress-mobile/WordPress-Android/blob/1adbfd718608a7d15290d70f73f6bb788b5b6912/WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PagesViewModel.kt#L210

The function above could be renamed to `fetchPagesFromServerAndRefreshUi` to make it clear what it does, since it's possible that this bug was caused by `reloadPages` and `refreshPages` being close in name and intent. 

### Actual behavior
The entire page is basically reloaded; the page is re-fetched from the server which isn't necessary in this case. 

1. Switch to this branch https://github.com/wordpress-mobile/WordPress-Android/pull/11191
2. Trigger an upload and if you put a breakpoint in the function below you will see that the pages are being fetched and the "local changes" if it exists will have a delay with being refreshed.
https://github.com/wordpress-mobile/WordPress-Android/blob/1adbfd718608a7d15290d70f73f6bb788b5b6912/WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PagesViewModel.kt#L210

### LINKED ISSUE #11265: Offline Support : Pages - PageListEventListener uploadStatus not cleared on error
### Expected behavior
When an update is being made on a post and an error occurs the `ProgressState` should be updated to reflect this. 

### Actual behavior
This should be invoked whether an error has took place or not since it will update the status.
https://github.com/wordpress-mobile/WordPress-Android/blob/1adbfd718608a7d15290d70f73f6bb788b5b6912/WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageListEventListener.kt#L102
At the moment, the invocation only occurs if the event doesn't have an error. 

Similar behavior in `PostListEventListener`
https://github.com/wordpress-mobile/WordPress-Android/blob/be1eff3ff6af9dbd60a79e83364f533e4b4c13a4/WordPress/src/main/java/org/wordpress/android/ui/posts/PostListEventListener.kt#L125

1. Update/Upload a post.
2. Create some form of error by manipulating the environment eg. disabling the internet during an upload.

malinajirka:
Fixes #11263
Fixes #11265

**Review instructions**
1. Review this PR
2. Wait until https://github.com/wordpress-mobile/WordPress-Android/pull/11335 gets merged
3. Update target branch to `feature/master-pages-offline-support`
4. Merge this PR

------------------

This PR should fix all the weird issues with upload state on the page list items.
- progress which never disappears
- local changes label remains visible even after the page gets uploaded
- upload error isn't displayed
- 3-5s delay after the upload finishes and the ui state gets updated
- ...

I tried to simplify the logic in PageListEventListener. 
- call `uploadStatusChanged` whenever we receive an event related to upload - no matter if it's a success or an error
- removed `handlePostUploadedWithoutError` and replaced it with `handlePageUpdated` which is invoked whenever we receive `UpdatePost` or `RemoteAutoSave` events - no matter if it's a success or an error

To test:
Anything related to upload you can think of
- remote auto save
- media upload
- failed upload
- failed media upload
- offline/online
We'll test all the different scenarios before we merge this whole feature into develop, so don't worry about testing all the scenarios.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PagesViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PagesViewModel.kt
@@ -203,7 +203,7 @@ class PagesViewModel
                 postStore = postStore,
                 eventBusWrapper = eventBusWrapper,
                 site = site,
-                handlePostUploadedWithoutError = this::handlePostUploadedWithoutError,
+                handlePageUpdated = this::handlePageUpdated,
                 invalidateUploadStatus = this::handleInvalidateUploadStatus,
                 handleRemoteAutoSave = this::handleRemoveAutoSaveEvent,
                 handlePostUploadedStarted = this::postUploadStarted
@@ -248,10 +248,6 @@ class PagesViewModel
 
     private suspend fun refreshPages() {
         pageMap = pageStore.getPagesFromDb(site).associateBy { it.remoteId }
-    }
-
-    private fun invalidatePages() {
-        pageMap = pageMap
     }
 
     fun onPageEditFinished(localPageId: Int, data: Intent) {
@@ -750,7 +746,7 @@ class PagesViewModel
     private fun hasRemoteAutoSavePreviewError() = _previewState.value != null &&
             _previewState.value == PostListRemotePreviewState.REMOTE_AUTO_SAVE_PREVIEW_ERROR
 
-    fun handlePostUploadedWithoutError(remotePostId: RemoteId) {
+    fun handlePageUpdated(remotePostId: RemoteId) {
         var id = 0L
         if (!pageUpdateContinuations.contains(id)) {
             id = remotePostId.value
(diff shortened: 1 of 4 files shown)


WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PagesViewModel.kt
```diff
@@ -772,16 +768,15 @@ class PagesViewModel
 
     private fun handleInvalidateUploadStatus(ids: List<LocalId>) {
         launch {
-            _invalidateUploadStatus.postValue(ids)
-            invalidatePages()
+            _invalidateUploadStatus.value = ids
+            refreshPages()
```
malinajirka: This line will keep hitting the database on each media upload progress state change, which isn't optimal. But I didn't come up with a different reliable solution. I haven't noticed any performance side-effects, so I guess we can keep it unless you can think of a better solution.

jd-alexander: Yes, we will have to keep it based on my observation.

PR #11341, opened 2020-02-19
Send encrypted logs to Wordpress API

claucookie:
Note: This PR is a Draft until this PR https://github.com/wordpress-mobile/WordPress-Android/pull/11302 gets merged.

**Title:** Send encrypted logs to Wordpress API

**Description:** 
Apart from Sending the Logs Id to Sentry, We want to send the actual encrypted Logs to our Wordpress API.

In this PR we have made some progress towards this and a working implementation is able to send the encrypted logs to the current Testing Wordpress API:

**Summary of What's done in this PR:**
- ✅ Implemented`LogEncryptionTestingActions.sendEncryptedLogsFile(...)` as a POST request to send the JSON object containing the encrypted logs information.  
- ✅ Updated `EncryptionUtils.java` to include the field `logsId` populated using a UUID V3 method https://en.wikipedia.org/wiki/Universally_unique_identifier#Versions_3_and_5_(namespace_name-based). The idea of using a namespace is to avoid collisions when generating the UUIDs before sending them to Wordrpess API and Sentry.
- ✅ Created `EncryptionUtils.getLogsUUID(...)` which returns the value of the `logsId`, this is used when we mutate the Sentry event just before it gets sent.
- ✅ Updated `CrashLoggingUtils.enableCrashLogging(...)` to perform the following operations before the Sentry event gets sent:
   - Get recent logs and encrypt them.
   - Send encrypted logs to WordPress API.
   - Attach the `LogsId` to the Sentry event as additional information or `extra`.
Screenshot 2020-02-18 at 16 29 19

- ✅ Moved the Encryption related classes to a separate package called `Encryption`.

**Pending stuff for future PRs:**
- 🚧 Update backend logic to read the new field `logsId`.
- 🚧 Update endpoint url. @jkmassel At the moment We are using the testing endpoint, so when the real one gets implemented we might need to update this.
- 🚧 Solve queue issue with Volley. When an unhandled exception closes the app the Volley request that sends the encrypted logs get successfully started but the response is not received. 
🤔 We might need a caching strategy to avoid losing requests when the app gets closed. Or use an intent service to send the request? WDYT?

```kotlin
// LogEncryptionTestingActions.kt

fun sendEncryptedLogsFile(encryptedLogsJson: JSONObject) {
     val jsonRequest = JsonStringRequest(
             Request.Method.POST,
             ENCRYPTION_LOGS_SERVER_URL,
             encryptedLogsJson,
             Listener {   // <-- This success listener never gets called for unhandled exception
AppLog.i(AppLog.T.API, "Encrypted logs response successfully received by Wordpress API") 
             },
             ErrorListener { AppLog.e(AppLog.T.API, "An error occurred when sending logs. " + it.message) }
     )
     WordPress.sRequestQueue.add(jsonRequest)
     AppLog.i(AppLog.T.API, "Sending encrypted logs to Wordpress API") // <-- I see this in the logcat
    }
```

**Testing instructions:** 
```kotlin
// CrashloggingUtils.kt
...
        }

        return !hasUserOptedOut
    }
```
Update the `gradle.properties` file with your Sentry DSN and the base64 public key:
```
wp.sentry.dsn=https://000000000000@sentry.io/0000000
wp.encryption.public_key = XXXXXXXXXXXXXXXXXXXXXXXXXXXXX
```
Log an exception inside the Login or HelpActivity by adding this line:
```kotlin
// HelpActivity.kt

CrashloggingUtils.log(RuntimeException("This is a new exception"))
```

Build the project, deploy the app and go to the Activity where you added the exception.

*PR submission checklist:*

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/util/encryption/LogsEncryptionActions.kt
+++ b/WordPress/src/main/java/org/wordpress/android/util/encryption/LogsEncryptionActions.kt
@@ -12,7 +12,7 @@ import org.json.JSONObject
 import org.wordpress.android.WordPress
 import org.wordpress.android.util.AppLog
 
-class LogEncryptionTestingActions {
+class LogsEncryptionActions {
     /**
      * Sends a POST request with the encrypted logs and useful fields
      * for the server to decrypt them.
@@ -22,7 +22,7 @@ class LogEncryptionTestingActions {
                 Request.Method.POST,
                 ENCRYPTION_LOGS_SERVER_URL,
                 encryptedLogsJson,
-                Listener { AppLog.i(AppLog.T.API, "Encrypted logs response successfully received by Wordpress API") },
+                Listener { AppLog.i(AppLog.T.API, "Encrypted logs response successfully received from Wordpress API") },
                 ErrorListener { AppLog.e(AppLog.T.API, "An error occurred when sending logs. " + it.message) }
         )
         WordPress.sRequestQueue.add(jsonRequest)
@@ -34,6 +34,8 @@ private const val ENCRYPTION_LOGS_SERVER_URL = "https://log-encryption-testing.h
 
 /**
  * This class represents a Request which response is a String.
+ * We need it because at the moment the testing endpoint returns
+ * a String with the decrypted logs.
  */
 class JsonStringRequest(
     method: Int,
--- a/WordPress/src/main/java/org/wordpress/android/util/encryption/LogsFileProvider.kt
+++ b/WordPress/src/main/java/org/wordpress/android/util/encryption/LogsFileProvider.kt
@@ -1,19 +1,17 @@
 package org.wordpress.android.util.encryption
 
 import android.content.Context
-import android.util.Base64
-import org.wordpress.android.BuildConfig
 import org.wordpress.android.util.AppLog
-import java.io.File
-import java.io.FileOutputStream
 
 class LogsFileProvider {
 
-    fun getMostRecentLogsFile(context: Context): File? {
-        return null
+    fun getMostRecentLogs(context: Context): List<String> {
+        // TODO: Implement a specific method to retrieve logs from a file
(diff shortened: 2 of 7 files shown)


WordPress/src/main/java/org/wordpress/android/util/encryption/LogsFileProvider.kt
```diff
@@ -0,0 +1,11 @@
+package org.wordpress.android.util.encryption
+
+import android.content.Context
+import org.wordpress.android.util.AppLog
+
+class LogsFileProvider {
+    fun getMostRecentLogs(context: Context): List<String> {
+        // TODO: Implement a specific method to retrieve logs from a file
```
jkmassel: Can we implement this method? Generally it's preferable not to commit a `TODO` to the repo – even if this were a separate PR that implemented `LogsFileProvider`, then this PR just called the completed method, that would be ideal :)

WordPress/src/main/java/org/wordpress/android/util/encryption/LogsEncryptionActions.kt
```diff
@@ -0,0 +1,52 @@
...
+     */
+    fun sendEncryptedLogsFile(encryptedLogsJson: JSONObject) {
+        val jsonRequest = JsonStringRequest(
+                Request.Method.POST,
+                ENCRYPTION_LOGS_SERVER_URL,
+                encryptedLogsJson,
+                Listener { AppLog.i(AppLog.T.API, "Encrypted logs response successfully received from Wordpress API") },
+                ErrorListener { AppLog.e(AppLog.T.API, "An error occurred when sending logs. " + it.message) }
+        )
+        WordPress.sRequestQueue.add(jsonRequest)
```
jkmassel: This call tightly binds this class to the `WordPress` one. This could make it difficult to implement goal 4 (migrating this to a shared library later).

PR #11347, opened 2020-02-20
Enable publish page action on page list in offline

### LINKED ISSUE #11177: Update snackbar messages

malinajirka:
Fixes #11177

Enables "Publish Now" action on page list in offline. This new approach doesn't support "Undo" action, but it's not supported even on the post list.

Also shows snackbar messages when upload finished (both success and fail).

To test:
1. Open page list
2. Select "drafts" tab
3. Turn on airplane mode
4. Click on "Publish now" menu item on one of the pages
5. Notice the page is moved to Published tab and "We'll publish the page when your device is back online." label is shown
6. Turn off airplane mode
7. Notice the page gets published and a snackbar is shown

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/pages/PagesFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/pages/PagesFragment.kt
@@ -49,7 +49,6 @@ import org.wordpress.android.ui.posts.ProgressDialogHelper
 import org.wordpress.android.ui.posts.RemotePreviewLogicHelper
 import org.wordpress.android.ui.quickstart.QuickStartEvent
 import org.wordpress.android.ui.uploads.UploadActionUseCase
-import org.wordpress.android.ui.uploads.UploadUtils
 import org.wordpress.android.ui.uploads.UploadUtilsWrapper
 import org.wordpress.android.ui.utils.UiHelpers
 import org.wordpress.android.util.DisplayUtils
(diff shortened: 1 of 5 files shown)


WordPress/src/main/java/org/wordpress/android/ui/pages/PagesFragment.kt
```diff
@@ -388,6 +388,25 @@ class PagesFragment : Fragment() {
                 )
             }
         })
+
+        viewModel.publishAction.observe(this, Observer {
+            it?.let {
+                UploadUtils.publishPost(activity, it.post, it.site, dispatcher)
```
jd-alexander: Hi Jirka! This is probably a nitpick but I was thinking we could add this function to the `UploadUtilsWrapper` since it's being used in this class. A function above this that wasn't in this diff is also using `UploadUtils` as well. Wdyt?

jd-alexander:
I wasn't able to see the failure message due to this bug https://github.com/wordpress-mobile/WordPress-Android/issues/11357

WordPress/src/main/java/org/wordpress/android/ui/pages/PagesFragment.kt
malinajirka: Good call, thanks! Fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/11347/commits/5f5e4a593b902f08ac081c2239a7eb9efbda04d6

jd-alexander: Awesome thanks!

PR #11355, opened 2020-02-21
Centering vertically the items of reader bottom sheet.

develric:
This PR is related to issue 3 found during #11343 review (see [here](https://github.com/wordpress-mobile/WordPress-Android/pull/11343#pullrequestreview-361948620) for more context)

### To test
- Open the Reader bottom sheet both in SITES and TAGS pages and check that the items are vertically aligned (here below two images for reference)

| Screenshot_1582244689 | Screenshot_1582244714 |
|---|---|

PR submission checklist:


malinajirka:
Thanks @develric!

I've noticed this PR changes the height of the items. The items look more condensed. I'm not sure we want to make such change. Could we perhaps add top and bottom padding to all the items?

I also noticed that the top margin before the first item in the list cuts-off the item. It's not related to this PR, but we might want to fix it. 

| Before  | After |
| ------------- | ------------- |
| Screenshot_1582276689 | Screenshot_1582276636 |

malinajirka:
@renanferrari I've talked to Riccardo and it'd be great if you could take over review of this PR. I'm afk on Monday and it's almost EOD for me today and I'd rather not block the progress. Thank you so much!

develric:
Hi @malinajirka thanks for your time and all the valuable feedbacks in this and the other PRs in last days 🙇‍♂️!

Hi @renanferrari I pushed some modifications:

1. Removed the extra top margin in the recycler view

<p align="center">
  
</p>

2. Center the sites and tags list items. As discussed (cc @osullivanchris ) I tried to stay to 64dp for sites and 48 for tags

| Screenshot_1582319086 | Screenshot_1582319266 |
|---|---|



I should not have messed up with styles, but if you could have a special eye that I did not change things inherited by other styles I would appreciate 😊!

Thanks for looking into it Renan 🙇‍♂️!

WordPress/src/main/res/layout/subfilter_page_fragment.xml
```diff
@@ -13,7 +13,6 @@
         android:clipToPadding="false"
         android:descendantFocusability="beforeDescendants"
         android:scrollbars="vertical"
-        android:layout_marginTop="@dimen/margin_extra_large"
         android:layout_marginBottom="@dimen/margin_extra_large"/>
```
renanferrari: You might want to switch from `layout_marginBottom` to `paddingBottom` here, so that the `clipToPadding` has effect and the list won't cut-off the items at the bottom.

develric: Good point 👍! Changed it here 6347458

PR #11373, opened 2020-02-25
Issue/material stats widget configuration

khaykov:
This PR moves Stats widget configuration to the material theme and adds dark mode support.

We are using Subtitle 1 text appearance for the previously bold labels.

comparison_st

Only one designer and one engineer are required to review this PR, but any is welcome to join.

To test:
- Make sure that configuration activities for all the widgets look ok in both light and dark themes.

PR submission checklist:


khaykov:
Ah, good catch, @develric ! I hope you don't mind that I updated the look of that widget in this PR. It's super old, so I had to change couple of things, but I tried no to get too involved :)

comparison_qp

Since widget screens are pretty simple, I think we can skip design review, for now, to speed up things a bit. We can address any potential issue during the final check.

WordPress/src/main/res/layout/quick_press_widget_configure_list_row.xml
```diff
@@ -0,0 +1,43 @@
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:background="?android:selectableItemBackground"
+    android:orientation="vertical"
```
develric: super minor np: I guess this can be here from some copy/paste? wdyt?

khaykov: Yep, this not supposed to be there :)

khaykov:
Thank you for review, @develric !

PR #11374, opened 2020-02-25
Issue/material log view activity

khaykov:
This PR moves Application Log activity to material theme and adds dark mode support.

This one is pretty simple - we removed the different colors for logs since they are supposed to be shared with developers, and will be shared as a plain text.  Maintaining their colors for both light and dark modes in the Utility submodule is not worth the effort.

comparison_cr_land

To test:
- Check out the Application Log activity in both light and dark modes and make sure it looks ok.

PR submission checklist:


khaykov:
Thanks for the review, @theck13 ! Personally, I like having those, but as you said, It's not related to the theme, so we can address them, if necessary, at a later time :)

Issue #11146, opened 2020-01-23
Replace the original image with the edited image in Gutenberg editor


malinajirka:
Goal of this task is to accept the result from uCrop and replace the original image with the edited image in the editor.

ashiagr:
Fixed in #11403.

PR #11376, opened 2020-02-25
Issue/11145 start ucrop with high res image

### LINKED ISSUE #11145: Start uCrop with the high resolution version of the image
Start uCrop with the high resolution version of the image.

Consider introducing the navigation component.

ashiagr:
Fixes #11145

This PR
- Starts `uCrop` from the preview fragment with the high res image file path.
- Returns to the original post on the `Gutenberg` editor skipping preview fragment on back action.

P.S.
Few observations related to `uCrop` fragment which may require separate tasks (adding them as  a checklist): 

- [X] On back action from the app bar on the `uCrop` fragment, we want to go to the original post on the `Gutenberg` editor skipping preview fragment. 
    Using `popUpToInclusive` property for this purpose  doesn't pop the start destination (preview fragment) off the back stack  in a module.
     On doing so, [activity finishes](https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-master-dev/navigation/navigation-runtime/src/main/java/androidx/navigation/NavController.java#386) and then a failed attempt is made to reopen the start destination. As a workaround, explicitly invoking back action for the `uCrop` fragment. Will further investigate it when we pass edited image to the `Gutenberg` editor.

    Related Android issue [here](https://issuetracker.google.com/issues/147312109).

    **UPDATE**: As discussed in comments, workaround looks ok. 

- [ ] Using temporary options (aspect ratios, crop frame/ grid) and styling as shown in this screenshot. Will update with actual options after a discussion with the designer.
cc @mbshakti 

device-2020-02-25-140701

- [ ] The cropping rectangle in the `uCrop` fragment is currently shown behind the toolbar. As a result we cannot drag upper corners. Propagating events to the view behind (by setting `clickable`, `focusable` to `false`) doesn't seem to be working in case of [Toolbar](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/widget/Toolbar.java#1585) as it eats all touch events, regardless of the `clickable` attribute and `@null` background.

```
@Override
public boolean onTouchEvent(MotionEvent ev) {
  // Toolbars always eat touch events
```
Will take care of this issue in a separate PR.

- [X] `uCrop` doesn't seem to be saving image state and restoring it on orientation change (at least with the version we're using`2.2.2`). Will investigate it more in a separate task. 

    Related `uCrop` issue [here](https://git.io/JvE7P).
    
   **UPDATE**: As discussed in comments, leaving this issue to the end of the project. Added a note in the Media Editing project.

- [X] Yet to decide output file path.

    **UPDATE**: As discussed in comments, output file location is ok.

- [X] Noticed a bug, low res(LR) image was saved to file and sent to `uCrop` on moving the app to background while LR image was loading and bringing the app to foreground after it was loaded but before HR image was loaded. 

    It is probably happening because `viewModel.onLoadIntoImageViewSuccess(url)` is called again due to the [LiveData fact](https://developer.android.com/topic/libraries/architecture/livedata#the_advantages_of_using_livedata):

    > If a lifecycle becomes inactive, it receives the latest data upon becoming active again. For example, an activity that was in the background receives the latest data right after it returns to the foreground.

    Will create an issue and fix it in a separate PR.
    
    **UPDATE**: Will take care of it in this issue: https://github.com/wordpress-mobile/WordPress-Android/issues/11377.

LR_image

PR submission checklist:

### DIFF
--- a/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
+++ b/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
@@ -10,13 +10,16 @@ import android.widget.ImageView.ScaleType.CENTER
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.Observer
 import androidx.lifecycle.ViewModelProvider
+import androidx.navigation.fragment.findNavController
 import kotlinx.android.synthetic.main.fragment_preview_image.*
 import org.wordpress.android.imageeditor.ImageEditor
 import org.wordpress.android.imageeditor.ImageEditor.RequestListener
+import org.wordpress.android.imageeditor.R
 import org.wordpress.android.imageeditor.R.layout
 import org.wordpress.android.imageeditor.preview.PreviewImageViewModel.ImageData
 import org.wordpress.android.imageeditor.preview.PreviewImageViewModel.ImageLoadToFileState.ImageStartLoadingToFileState
 import org.wordpress.android.imageeditor.preview.PreviewImageViewModel.ImageUiState.ImageDataStartLoadingUiState
+import org.wordpress.android.imageeditor.utils.UCropUtil
 import org.wordpress.android.imageeditor.utils.UiHelpers
 import java.io.File
 
@@ -103,5 +106,14 @@ class PreviewImageFragment : Fragment() {
         )
     }
 
-    private fun startUCrop(inputFilePath: String) { }
+    private fun startUCrop(inputFilePath: String) {
+        findNavController().navigate(
+            R.id.action_previewFragment_to_ucropFragment,
+            UCropUtil.getUCropOptionsBundle(
+                File(inputFilePath),
+                // TODO: output file location
+                File(requireContext().cacheDir, UCropUtil.IMAGE_EDITOR_OUTPUT_IMAGE_FILE_NAME)
+            )
+        )
+    }
 }
--- a/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
+++ b/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
@@ -68,8 +68,8 @@ class PreviewImageFragment : Fragment() {
             }
         })
 
-        viewModel.startCrop.observe(this, Observer { filesInfo ->
-            startCrop(filesInfo)
+        viewModel.navigateToCropScreenWithFilesInfo.observe(this, Observer { filesInfo ->
+            navigateToCropScreenWithFilesInfo(filesInfo)
         })
     }
 
@@ -106,7 +106,7 @@ class PreviewImageFragment : Fragment() {
         )
     }
 
-    private fun startCrop(filesInfo: Pair<File, File>) {
+    private fun navigateToCropScreenWithFilesInfo(filesInfo: Pair<File, File>) {
         findNavController().navigate(
             R.id.action_previewFragment_to_cropFragment,
             CropUtil.getCropInfoBundleWithFilesInfo(filesInfo)
--- a/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
+++ b/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
@@ -16,7 +16,6 @@ import org.wordpress.android.imageeditor.ImageEditor
 import org.wordpress.android.imageeditor.ImageEditor.RequestListener
 import org.wordpress.android.imageeditor.R
 import org.wordpress.android.imageeditor.R.layout
-import org.wordpress.android.imageeditor.preview.PreviewImageViewModel.Companion.IMAGE_EDITOR_OUTPUT_IMAGE_FILE_NAME
 import org.wordpress.android.imageeditor.preview.PreviewImageViewModel.ImageData
 import org.wordpress.android.imageeditor.preview.PreviewImageViewModel.ImageLoadToFileState.ImageStartLoadingToFileState
 import org.wordpress.android.imageeditor.preview.PreviewImageViewModel.ImageUiState.ImageDataStartLoadingUiState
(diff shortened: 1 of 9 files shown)


ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
```diff
@@ -98,4 +105,15 @@ class PreviewImageFragment : Fragment() {
...
         )
     }
+
+    private fun startUCrop(inputFilePath: String) {
+        findNavController().navigate(
+            R.id.action_previewFragment_to_ucropFragment,
+            UCropUtil.getUCropOptionsBundle(
+                File(inputFilePath),
+                // TODO: output file location
+                File(requireContext().cacheDir, UCropUtil.IMAGE_EDITOR_OUTPUT_IMAGE_FILE_NAME)
```
malinajirka: I think we might want to consider moving this little piece of logic into the VM. We could pass the cacheDir into `vm.onCreateView` and pass both `inputFilePath` and `outputFilePath` into the LiveData. Wdyt?

>UCropUtil.IMAGE_EDITOR_OUTPUT_IMAGE_FILE_NAME

How would you feel about moving this constant into the VM? I feel like it's too specific to live in a utility class, wdyt?

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
```diff
@@ -41,7 +44,10 @@ class PreviewImageViewModel : ViewModel() {
             else -> ImageInHighResLoadSuccessUiState
         }
 
-        if (newState == ImageInHighResLoadSuccessUiState && currentState != ImageInHighResLoadSuccessUiState) {
+        if (newState == ImageInHighResLoadSuccessUiState &&
+                currentState != ImageInHighResLoadSuccessUiState &&
+                loadIntoFile.value !is ImageLoadToFileSuccessState
```
malinajirka: I feel like this condition is starting to be difficult to reason about. How would you feel about something like this?

```
val highResImageJustLoadedIntoView = newState != currentState && newState == ImageInHighResLoadSuccessUiState
val imageNotLoadedIntoFile  = loadIntoFile.value !is ImageLoadToFileSuccessState
if (highResImageJustLoadedIntoView && imageNotLoadedIntoFile) {
...
```

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
```diff
@@ -98,4 +105,15 @@ class PreviewImageFragment : Fragment() {
             }
         )
     }
+
+    private fun startUCrop(inputFilePath: String) {
+        findNavController().navigate(
+            R.id.action_previewFragment_to_ucropFragment,
+            UCropUtil.getUCropOptionsBundle(
+                File(inputFilePath),
+                // TODO: output file location
```
malinajirka: I think the cacheDir is an ok location or do you have any concerns?

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
ashiagr: Agree. Changed here dd96921

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
ashiagr: Refactored here 7552d63 and 87fb32a

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
ashiagr: No concerns. Just wanted to confirm if it is ok.

ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
```diff
@@ -64,6 +67,10 @@ class PreviewImageFragment : Fragment() {
                 loadIntoFile(fileState.imageUrl)
             }
         })
+
+        viewModel.startUCrop.observe(this, Observer { filePath ->
```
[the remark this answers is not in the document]
ashiagr: Good suggestion! Removed "u" from "uCrop" to generalise it 😀. Let me know if you can think of a better name.

ashiagr:
Great review @malinajirka ! I've addressed most of your comments. Answered some here:

> We have a FrameLayout there which draws the toolbar on top (z-axis) of the fragment instead of drawing it above (y-axis). 

Actually that was intentional. I initially planned to not include free style crop rectangle and just provide aspect ratios similar to iOS implementation. As we're not displaying any title on the app bar, so thought of utilising the space behind it by extending the `ImageView` behind it. It could have been benefited us in the landscape mode where we have really less space.

device-2020-02-26-095712

Then thought of setting app bar background to `@null `and allow touch events to reach the `ImageView` behind it and have a free style crop rectangle. I can discuss the design with @mbshakti and update it accordingly. 
 
> Is the FrameLayout even necessary? 

Removed, it was an oversight 🤦‍♀️.

> I'd suggest leaving this issue to the end of the project after everything else is done. Wdyt?

Agree

> Imo this workaround is ok, especially since we'll want to have a different behavior when we introduce support for editing multiple images. Wdyt?

Agree

>  I noticed there isn't any "Done" button in the uCrop fragment. Is that something we plan to add?

Yes, I planned to include a check mark on the app bar on the right corner as part of https://github.com/wordpress-mobile/WordPress-Android/issues/11146. That's why I mentioned to investigate more about `popToInclusive` in that task.
I plan to use the `UCropResult` received here: 
https://github.com/wordpress-mobile/WordPress-Android/blob/fa4471a3e8aa70da6ffc91a8a7d9a0052588afdf/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/EditImageActivity.kt#L61 I will create a separate task for it.

PR #11385, opened 2020-02-26
Remove the Gutenberg dialog on new sites

maxme:
Remove the Gutenberg dialog that was showning after creating a new site. 
Also change the default for freshly added self hosted sites.

Test 1:
- Create a new site.
- Create a new post, make sure the block editor is showing and there is no dialog about it.

Test 2:
- Create a new user.
- Create a new site.
- Create a new post, make sure the block editor is showing and there is no dialog about it.

Test 3:
- Add a self hosted site.
- Create a new post, make sure the block editor is showing and there is no dialog about it.

PR submission checklist:


cameronvoell:
> Test 1:
> * Create a new site.
> * Create a new post, make sure the block editor is showing and there is no dialog about it.

In this case, for both new and and existing users, I am still seeing the dialog. The code enabling it is here: https://github.com/wordpress-mobile/WordPress-Android/blob/d5687d59a6d4f417e7d4242aef7225dea406772a/WordPress/src/main/java/org/wordpress/android/ui/main/SitePickerActivity.java#L223-L227

> Test 2:
> - Create a new user.
> - Create a new site.
> - Create a new post, make sure the block editor is showing and there is no dialog about it.

In this case I am also still seeing the dialog.  It seems to get enabled here: https://github.com/wordpress-mobile/WordPress-Android/blob/d5687d59a6d4f417e7d4242aef7225dea406772a/WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java#L118

> Test 3:
> - Add a self hosted site.
> - Create a new post, make sure the block editor is showing and there is no dialog about it.

In this specific case I am not seeing the dialog. However, the first time I click on editing an existing post, the dialog does show. This is caused by code on this line. https://github.com/wordpress-mobile/WordPress-Android/blob/d5687d59a6d4f417e7d4242aef7225dea406772a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L1408

cameronvoell:
I think this change was made after it was pointed out,
> "We display the message: “Block editor enabled: You are using the new editor…” to a brand new user. They don’t need to know we have another editor, that message can be removed now for new users. "

What is unclear for me is what are the cases for non-new users that we still want to show the dialog?

maxme:
> What is unclear for me is what are the cases for non-new users that we still want to show the dialog?

We should show the dialog only to users that we're migrating to the block editor.

maxme:
> > Test 2:
> > 
> > * Create a new user.
> > * Create a new site.
> > * Create a new post, make sure the block editor is showing and there is no dialog about it.
> 
> In this case I am also still seeing the dialog. It seems to get enabled here:

I can't reproduce that one, but it might happen due to the delay on that call after creating the site: https://github.com/wordpress-mobile/WordPress-Android/blob/99065d89b21393cb9e3e790c15f592dc100d4ac6/WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java#L181

maxme:
I fixed case 1 and 3 in 4ff75c3773edcc8f442ca866ba44d72ae668e14c

In 4f01138 - I also updated the enable/disable functions to update the site model locally, I'm not sure why it wasn't done originally.

PR #11421, opened 2020-03-10
Site Creation Remove Verticals and Site Info Steps - Master Issue

ashiagr:
Fixes #11420

Master PR - Site Creation Remove Verticals and Site Info Steps

- [x] Remove Analytics Tracking for Verticals (Step 1) and Site Info (Step 2) Site Creation steps 
    - [x] Inside app
    - [ ] Inside Tracks (Mark keys invalid - do later)
- [x] Remove Step 1 (Verticals)
    - [x] Remove Verticals Suggestions, corresponding FluxC `VerticalModel`
    - [x] Remove Header and Search Input Segment, corresponding FluxC `SegmentPromptModel`
    - [x] Remove Verticals Step screen
    - [x] Update new site requests to not include `verticalId`
- [x]  Remove Step 2 (Site Info)
    - [x] Remove Site Info Step screen
    - [x] Update new site requests to not include site title and tag line
- [x]  Other changes like modifying single domain step title (Step 1 of 1)

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #11420, opened 2020-03-10
Remove verticals and site info steps on site creation


ashiagr:
Remove below steps on site creation:
- What will your blog/site be about? (Verticals)
- Title and tagline. (Site Info) 

| Verticals | Site Info | 
| ------ | ------------- | 
| ![device-2020-03-10-104341](https://user-images.githubusercontent.com/1405144/76282082-bb104200-62bc-11ea-9a14-6fee1ff761ca.png) | ![device-2020-03-10-104351](https://user-images.githubusercontent.com/1405144/76282109-c19eb980-62bc-11ea-87dc-2891583f3d18.png)  | 

These steps were felt unnecessary when testing new user flow on WordPress apps.

Internal Ref: pbArwn-an-p2

PR #11429, opened 2020-03-11
[Site Creation] Step 1: Remove verticals suggestions display

ashiagr:
This PR removes verticals suggestions display (including custom category) from the verticals site creation step (Step 1 of 3) as part of https://github.com/wordpress-mobile/WordPress-Android/issues/11420 and https://github.com/wordpress-mobile/WordPress-Android/pull/11421. 

PR in FluxC (to remove `VerticalModel`) - https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1509
(Temporarily pointing to commit hash for above FluxC PR which needs to be reverted before merge.) 

device-2020-03-11-163712

To test:
- Go to My site
- Click Switch Site 
- Click "+" button to launch Add new site 
- Click Create WordPress.com site on Add new site 
- Choose a segment Blog/Business/Professional/Blank Canvas to open Step 1 of 3 
- Search for some query in Step 1 of 3 
- Verify no results are displayed

verticals_suggestions

PR submission checklist:


ashiagr:
> During testing if I type in a vertical, there will be no way to proceed since verticals are not available anymore. I am guessing this is expected

Right, we cannot select a vertical now. In case we want to proceed, we can try skipping the screen. But as these changes are part of a master PR which includes additional changes for verticals screen removal, it's probably best to wait to proceed till we remove the screen completely. Thanks for the note, appreciate it 🙏.

> it's probably best not to commit it and then revert back. Again, this is just me being nitpicky. 🙈

Completely valid point! I'll include it as a testing step instead in the future.

PR #11430, opened 2020-03-11
Sharing: Consistent messaging in Publicize to Facebook Page Error

### LINKED ISSUE #10278: iOS and Android: Consistent messaging in Publicize to Facebook
This covers both iOS and Android but we're mainly tracking it here in this Android issue.

## Problem

In p7cLQ7-DY-p2, we found that there are inconsistencies in how the mobile apps present information regarding the requirements for successfully connecting a Facebook account.

Connecting a Facebook account can be achieved by navigating to a site and tapping on Sharing → Facebook.

Here is a connection flow diagram that shows the inconsistencies:

Facebook@2x(1)

- iOS does not show any requirements information on the first page but shows it after the user fails to connect.
- Android shows some requirements information on the first page but the link is pointing to an old blog post, not the support page.

### Connection Requirements

The known requirements to successfully connect a Facebook account are:

1. The user should have a Facebook Page.
2. The user should select a Facebook Page when going through the connection flow.
3. The Facebook Page should be set to [**Published**](https://www.facebook.com/help/184605634921611).

This information is displayed on our support page. However, there is no mention of it in the apps.

## Solution

@thehenrybyrd, @rfaile313, and I propose to make these changes.

### Copy changes

On the first page, the description can be modified slightly to describe the requirement for Pages to be published.

> Connect to automatically share your blog posts to a Published Facebook Page

The error message shown after failing to connect can be made more concise and mention the requirements.

> We couldn’t connect this Facebook account because it doesn’t have access to any Published Facebook Pages. Facebook allows automatic sharing only to Pages which are Published.

### Android Flow

The first page on Android shows outdated information. We propose to change the flow on Android to match iOS instead. This wireframe shows the desired flow as well as the proposed copy changes.

Android Flow@2x(1)

## Editorial Team

We'd love to have your input on the copy changes we're proposing.

## Design Team

We'd love to hear your input on this as well. Does the existing iOS design still match our current standards? Or is there something that should be changed?

## Tasks

- [x] Wait for input from Editorial and Design 
- [x] Create an issue for iOS https://github.com/wordpress-mobile/WordPress-iOS/issues/12228
- [x] Implement on iOS. Done in https://github.com/wordpress-mobile/WordPress-iOS/pull/12337
- [ ] Implement on Android

jd-alexander:
Fixes #10278
## Findings
The messaging of the error that occurs when trying to publish to Facebook wasn't consistent on both platforms. This PR addresses this since the messaging on iOS was already updated to the new version. 

## Solution
I removed the obsolete messaging and learn more button from the publicize page flow. I then created a dialog specific to this since the existing dialog just supported the `message` of the dialog being set via an `EventBus` event that gets a reason message propagated from an exception that occurs within `PublicizeActions`. 

## Testing
1. I used a Facebook account that that has no pages. ( Having an existing Facebook account with no public pages should work as well.) 
2. Navigate to My Site. 
3. Select Sharing.
4. Click Facebook.
5. Click Connect. 
6. Follow through with the Facebook authentication flow. 
7. You will then be greeted with a Pages flow, which you should just go through and once it's done you should see the error dialog. 
8. To ensure no regressions occurred I tested the flow with a Facebook account that has public pages and it allowed me to go to the dialog where I see the page selected and I can connect it so posts are shared automatically. 

Step 1 | Step 2  
--------|-------
 |  
     

Step 3 | Step 4 
--------|-------
    |   

Result (Android)  | Comparison (iOS) 
--------|-------
 |       

* I capitalized the C in "Not connected" but decided not to update screenshot since it's a subtle change 60e3b85
## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/publicize/PublicizeErrorFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/publicize/PublicizeErrorFragment.kt
@@ -1,0 +1,64 @@
+package org.wordpress.android.ui.publicize
+
+import android.app.AlertDialog
+import android.app.Dialog
+import android.os.Bundle
+import android.view.ContextThemeWrapper
+import androidx.annotation.StringRes
+import androidx.fragment.app.DialogFragment
+import org.wordpress.android.R
+import org.wordpress.android.R.style
+import org.wordpress.android.ui.WPWebViewActivity
+import org.wordpress.android.ui.publicize.PublicizeConstants.PUBLICIZE_FACEBOOK_SHARING_SUPPORT_LINK
+import java.lang.NullPointerException
+
+class PublicizeErrorFragment : DialogFragment() {
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        arguments?.let {
+            it.getInt(REASON_RES_ID).let { reasonResId ->
+                // the custom dialog behavior here is specific to a user who is trying to share to Facebook
+                // but does not have any available Facebook Pages.
+                if (reasonResId == R.string.sharing_facebook_account_must_have_pages) {
+                    return AlertDialog.Builder(
+                            ContextThemeWrapper(
+                                    activity,
+                                    style.Calypso_Dialog
+                            )
+                    ).apply {
+                        setTitle(R.string.dialog_title_sharing_facebook_account_must_have_pages)
+                        setMessage(R.string.sharing_facebook_account_must_have_pages)
+                        setPositiveButton(R.string.ok) { dialog, _ -> dialog.dismiss() }
+                        setNegativeButton(R.string.learn_more) { _, _ ->
+                            WPWebViewActivity.openURL(
+                                    activity,
+                                    PUBLICIZE_FACEBOOK_SHARING_SUPPORT_LINK
+                            )
+                        }
+                    }.create()
+                } else {
+                    return AlertDialog.Builder(
+                            ContextThemeWrapper(
+                                    activity,
+                                    style.Calypso_Dialog
+                            )
+                    ).apply {
+                        setMessage(reasonResId)
+                        setPositiveButton(R.string.ok) { dialog, _ -> dialog.dismiss() }
+                    }.create()
+                }
+            }
+        } ?: throw NullPointerException("Missing argument. Utilize newInstance() to instantiate.")
+    }
+
+    companion object {
+        const val REASON_RES_ID = "REASON_RES_ID"
+        const val TAG = "publicize_error_fragment"
+
+        @JvmStatic
+        fun newInstance(@StringRes reasonId: Int) = PublicizeErrorFragment().apply {
+            arguments = Bundle().apply {
+                putInt(REASON_RES_ID, reasonId)
+            }
+        }
+    }
+}
--- a/WordPress/src/main/java/org/wordpress/android/ui/publicize/PublicizeListActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/publicize/PublicizeListActivity.java
@@ -26,6 +26,7 @@ import org.wordpress.android.fluxc.model.SiteModel;
 import org.wordpress.android.fluxc.store.SiteStore;
 import org.wordpress.android.models.PublicizeConnection;
 import org.wordpress.android.models.PublicizeService;
+import org.wordpress.android.ui.WPWebViewActivity;
 import org.wordpress.android.ui.publicize.PublicizeConstants.ConnectAction;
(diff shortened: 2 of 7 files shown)


WordPress/src/main/java/org/wordpress/android/ui/publicize/PublicizeListActivity.java
```diff
@@ -384,4 +394,15 @@ private void showAlertForFailureReason(int reasonResId) {
         builder.setPositiveButton(R.string.ok, (dialog, which) -> dialog.dismiss());
         builder.show();
     }
+
+    private void showAlertForFacebookPageFailureReason() {
```
jd-alexander: This dialog and the one below it isn't adhering to the new material theme. Should I create an issue to track this @khaykov or it's something you plan to work on? Let me know the best course of action. Thanks 🙏

khaykov: No worries! I'll make a mental note to update it :)

jd-alexander: Awesome! Also if you want a Facebook account to test with I could always set up a temp password for the one I used because I know it can be a bit annoying to do this stuff with social networks. Let me know

WordPress/src/main/java/org/wordpress/android/ui/publicize/PublicizeErrorFragment.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.wordpress.android.ui.publicize.PublicizeConstants.PUBLICIZE_FACEBOOK_SHARING_SUPPORT_LINK
+import java.lang.NullPointerException
+
+class PublicizeErrorFragment : DialogFragment() {
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        arguments?.let {
+            it.getInt(REASON_RES_ID).let { reasonResId ->
+                // the custom dialog behavior here is specific to a user who is trying to share to Facebook
+                // but does not have any available Facebook Pages.
+                if (reasonResId == R.string.sharing_facebook_account_must_have_pages) {
```
jd-alexander: I flipped the condition here to `!=` to test the other dialog to verify it works as expected.

jd-alexander:
@khaykov I made changes so the dialogs survive configuration change. Ready for another review!

WordPress/src/main/java/org/wordpress/android/ui/publicize/PublicizeErrorFragment.kt
```diff
@@ -0,0 +1,64 @@
...
+                        setMessage(R.string.sharing_facebook_account_must_have_pages)
+                        setPositiveButton(R.string.ok) { dialog, _ -> dialog.dismiss() }
+                        setNegativeButton(R.string.learn_more) { _, _ ->
+                            WPWebViewActivity.openURL(
+                                    activity,
+                                    PUBLICIZE_FACEBOOK_SHARING_SUPPORT_LINK
+                            )
+                        }
+                    }.create()
+                } else {
```
jd-alexander: I put both dialogs here since they are both displaying an error/failure. Is this approach fine? Double-checking to ensure I am not somehow putting too much responsibility in the class. I just thought it wouldn't make sense to have two of these since the main difference is just the contents of the dialog being displayed.

khaykov: Yeah, that's fine in my opinion :)

WordPress/src/main/java/org/wordpress/android/ui/publicize/PublicizeErrorFragment.kt
```diff
@@ -0,0 +1,64 @@
...
+import android.view.ContextThemeWrapper
+import androidx.annotation.StringRes
+import androidx.fragment.app.DialogFragment
+import org.wordpress.android.R
+import org.wordpress.android.R.style
+import org.wordpress.android.ui.WPWebViewActivity
+import org.wordpress.android.ui.publicize.PublicizeConstants.PUBLICIZE_FACEBOOK_SHARING_SUPPORT_LINK
+import java.lang.NullPointerException
+
+class PublicizeErrorFragment : DialogFragment() {
```
khaykov: Can we rename this one to `PublicizeErrorDialogFragment` for consistency with other dialog fragments?

jd-alexander: Sure! Thanks for spotting. Resolved a773842

khaykov: Perfect, thank you!

PR #11431, opened 2020-03-12
Me/ Profile - Change Photo  -  Image Preview Fix

### LINKED ISSUE #11314: [GlobalStep] Tapping and holding on images in the profile picture album shows ‘Unable to load media’.
### Description
If the user goes on to the ‘Me’ profile section and selects the profile picture, when the album loads it is possible to tap and hold on an image which will always result in a black screen with the error ‘Unable to load media’.

### Reproduction Rate
4/4 100%

### Expected behavior
Tap and hold has a function or else can’t be done.

### Actual behavior
Tap and hold shows ‘Unable to load media’.

1.	Install WordPress 14.2
2.	Login.
3.	Select the ‘Me’ profile icon.
4.	Tap the user profile picture.
5.	Tap and hold on an image in the album.

##### Tested on the following
Samsung Galaxy S2 8.0” (7.0)
LG G3 (5.0)

### Submitted by:
Oliver Howe

jd-alexander:
Fixes #11314

## Findings
The `MediaPreviewFragment` primarily uses a URL to load images from the web. A content URI was being passed to the `ImageManager`'s image loading function which only knew how to deal with Photon based URLs.

## Solution
An overload was created in the `ImageManager` that loads a content URI and a check is being done so that the appropriate function is utilized based on the image URI type. 

I had to do a bit of code duplication to support the additional type. Let me know if you have any ideas to simplify the fix. I was going to use generics but that would involve making a change to a function being used in other places so for simplicity, I just utilized method overloading.
## Testing
1. Go to Me page. 
2. Click on "Change Photo"
3. Long press on an image in "Choose Photo" Picker. 
4. See the image preview. 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/media/MediaPreviewFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/media/MediaPreviewFragment.java
@@ -285,36 +285,35 @@ public class MediaPreviewFragment extends Fragment implements MediaController.Me
         }
         showProgress(true);
 
-        mImageManager
-                .loadWithResultListener(mImageView, ImageType.IMAGE, Uri.parse(mediaUri), ScaleType.CENTER, null,
-                        new RequestListener<Drawable>() {
-                            @Override
-                            public void onResourceReady(@NonNull Drawable resource, @Nullable Object model) {
-                                if (isAdded()) {
-                                    // assign the photo attacher to enable pinch/zoom - must come before
-                                    // setImageBitmap
-                                    // for it to be correctly resized upon loading
-                                    PhotoViewAttacher attacher = new PhotoViewAttacher(mImageView);
-                                    attacher.setOnViewTapListener((view, x, y) -> {
-                                        if (mMediaTapListener != null) {
-                                            mMediaTapListener.onMediaTapped();
-                                        }
-                                    });
-                                    showProgress(false);
+        mImageManager.loadWithResultListener(mImageView, ImageType.IMAGE, Uri.parse(mediaUri), ScaleType.CENTER, null,
+                new RequestListener<Drawable>() {
+                    @Override
+                    public void onResourceReady(@NonNull Drawable resource, @Nullable Object model) {
+                        if (isAdded()) {
+                            // assign the photo attacher to enable pinch/zoom - must come before
+                            // setImageBitmap
+                            // for it to be correctly resized upon loading
+                            PhotoViewAttacher attacher = new PhotoViewAttacher(mImageView);
+                            attacher.setOnViewTapListener((view, x, y) -> {
+                                if (mMediaTapListener != null) {
+                                    mMediaTapListener.onMediaTapped();
                                 }
-                            }
+                            });
+                            showProgress(false);
+                        }
+                    }
 
-                            @Override
-                            public void onLoadFailed(@Nullable Exception e, @Nullable Object model) {
-                                if (isAdded()) {
-                                    if (e != null) {
-                                        AppLog.e(T.MEDIA, e);
-                                    }
-                                    showProgress(false);
-                                    showLoadingError();
-                                }
+                    @Override
+                    public void onLoadFailed(@Nullable Exception e, @Nullable Object model) {
+                        if (isAdded()) {
+                            if (e != null) {
+                                AppLog.e(T.MEDIA, e);
                             }
-                        });
+                            showProgress(false);
+                            showLoadingError();
+                        }
+                    }
+                });
     }
 
     /*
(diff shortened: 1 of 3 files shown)


libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/UrlUtils.java
```diff
@@ -49,6 +49,10 @@ public static String getHost(final String urlString) {
         return "";
     }
 
+    public static boolean checkIfContentUri(String uri) {
```
khaykov: I feel like `checkIfContentUri` name is a bit ambiguous.  Maybe we could rename it to something like `isContentUri`? WDYT?

WordPress/src/main/java/org/wordpress/android/ui/media/MediaPreviewFragment.java
```diff
@@ -277,43 +278,74 @@ private void loadImage(String mediaUri) {
...
-                    public void onResourceReady(@NonNull Drawable resource, @Nullable Object model) {
-                        if (isAdded()) {
-                            // assign the photo attacher to enable pinch/zoom - must come before setImageBitmap
-                            // for it to be correctly resized upon loading
-                            PhotoViewAttacher attacher = new PhotoViewAttacher(mImageView);
-                            attacher.setOnViewTapListener(new PhotoViewAttacher.OnViewTapListener() {
+
+        if (UrlUtils.checkIfContentUri(mediaUri)) {
+            mImageManager
+                    .loadWithResultListener(mImageView, ImageType.IMAGE, Uri.parse(mediaUri), ScaleType.CENTER, null,
```
khaykov: I wonder, can we utilize `Uri.parse(mediaUri)` for non-content URI? I briefly tried it with remote URLs from the free photo library, and it worked ok. If there is any issue with this, do you think we can extract `RequestListener` from here and use it for both `loadWithResultListener` calls?

libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/UrlUtils.java
jd-alexander: Nice. Thanks!

WordPress/src/main/java/org/wordpress/android/ui/media/MediaPreviewFragment.java
jd-alexander: > I wonder, can we utilize Uri.parse(mediaUri) for non-content URI? I briefly tried it with remote URLs from the free photo library, and it worked ok.

Super cool! Thanks for spotting this @khaykov Eliminated all the other concerns I was having and the duplicated code is now gone. I resolved it here 0c024ad I am hiding that it's being converted to a `Uri` in the `ImageManager` because it's still accepting a `String` url. That's fine right? All other methods are accepting `String` values. My only concern with my approach is that I am wondering if someone would use this for a content URI and it works and then attempt to do the same with another method in the `ImageManager` and the behavior is different and the issue isn't easily noticeable. Let me know your thoughts.

khaykov: Yeah, I agree that it might be confusing. We should have the same behavior in other methods too. Changing `imgUrl` from String to Uri without overloading will result in quite a lot of changes, so URI parsing in other methods sounds like the most reasonable approach to me.

jd-alexander: Hey @khaykov I was going to do this but decided against it since there would be alot of functionality to test/double check to be sure no regression occurs. ( I know it would work but would be worth retesting since this impacts a lot of core functionality) So in the interest of time, I simply created an overload that utilized the `URI` type for the `MediaPreviewFragment` so the rest of the functionality in the `ImageManager` remains untouched. Did that here 1e7df64

khaykov: @jd-alexander Makes sense! Thank you :)

WordPress/src/main/java/org/wordpress/android/ui/media/MediaPreviewFragment.java
```diff
@@ -277,43 +278,43 @@ private void loadImage(String mediaUri) {
...
-                            PhotoViewAttacher attacher = new PhotoViewAttacher(mImageView);
-                            attacher.setOnViewTapListener(new PhotoViewAttacher.OnViewTapListener() {
-                                @Override
-                                public void onViewTap(View view, float x, float y) {
-                                    if (mMediaTapListener != null) {
-                                        mMediaTapListener.onMediaTapped();
-                                    }
+
+        mImageManager
+                .loadWithResultListener(mImageView, ImageType.IMAGE, Uri.parse(mediaUri), ScaleType.CENTER, null,
```
khaykov: Just one small thing - can we fix an indentation of this method?

jd-alexander: Of course! That one slipped in 😅 67977f2

khaykov: Perfect, thank you!

PR #11432, opened 2020-03-12
Fixing material theme on older API's and removing inconsistencies

khaykov:
This PR fixes drawable tinting on older devices (API 21-23) and removes some minor inconsistencies with styling around the app.

What changed:
- Bumped appcompat to `1.2.0-alpha03`. This was necessary because of the nasty bug in appcompat 1.1.0 (yes, another bug) that crashed WebView.
- Updated tinting logic to be older-api friendly:
    - Using `app:tint` instead of `android:tint` (this is 85% of changes)
    - Using alpha values inside selectors instead of using nested selectors with alpha inside them.
    - Using `AppCompatResources` to get `ColorStateList`
- Changed `accentColor` from secondary to primary.
- Normalized icon color through the app. Now, most icons should use only one shade of gray and not 50.
- Global style value for material cards - since we are using Bridge version of the material theme, it's not automatically picked up by cards, and without style, they are too elevated.

To test:
- Launch the app on device with API 21 and do a brief look around. Apart from icon color, card elevation and accent color you should see no changes in appearance. 

If you see any icons that look like this, let me know:
[Image from Gyazo](https://gyazo.com/8f8c125e3f2c0b21feeb2109604e834b)

- Launch the app on modern devices with API 27, 28 or 29. and do a brief look around. Apart from icon color, card elevation and accent color you should see no changes in appearance. 

- On any device, navigate to site creation flow, and make sure that the domain picker field works as expected.

PR submission checklist:


khaykov:
Looks like I'll have to backpedal on bumping appcompat to `1.2.0-alpha03` to fix the WebView crash. The fix on Google side was quite radical, and might potentially introduce another issue. I'll work on an alternative solution.

khaykov:
Ok, so I reverted bump of appcompat to `1.2.0-alpha03`. It had a bunch of good fixes, but it's still in alpha, and it's not worth risking. Because of this and the fact that appcompat 1.1.0 is flaming mess, I had to fix some of the issues on older APIs manually:
- Added a custom WebView with an overloaded context to combat crash on API's 21 and 23.
- Changed the logic of Locale change - instead of modifying the base context of activity and injecting new locale in it, we only modify Configuration in `applyOverrideConfiguration`, which is one of the approaches suggested by google engineers [here](https://issuetracker.google.com/issues/141869006#comment9).

Two new things to test:
- WebView on devices with API 21 or 21. Preview and Reader should not crash.
- Locale change on devices with API from 21 to 25. Try setting the locale to RTL language and make sure it works and change persists after the configuration change.

One thing to note:
- Some UI elements on older devices might not look "material" or might look slightly different comparing to newer APIs (eg. button colors in some preference dialogs). This is expected.

planarvoid:
hey @khaykov , thanks for the changes. I was testing the changes and there is one thing I've noticed. The grey background is now gone. I'm not sure if it's caused by your changes but I've compared the code to develop (on the same devices) and it looks like this:
Screenshot_1584365722
This is develop:
Screenshot_1584366218

Don't mind the tablet setup, it was a device I already had installed with the correct API version :-)

planarvoid:
There is one more thing I got from testing. I got weird behaviour from switching my language to Czech. Steps to repro are:
* Switch your language in App settings to Czech (or probably any other language that's different from your device's language)
* Go back to "My site"
* Go to Reader
* Go back to "My site"
* Notice that the icon is missing and the Quick start steps are there with empty text

Check this video to see for yourself
lanuage_issue

khaykov:
@planarvoid @renanferrari Thank you so much for the review! It looks like my change in locale updating logic affected the devices with API 26+, so I brought back `attachBaseContext`. `applyOverrideConfiguration` is called on API 21 to 25 and `attachBaseContext` called on all API's but does nothing on 21 to 25. 

@planarvoid We got rid of the grey background in the app since it was not working too well with dark mode.

Regarding the other issue, you found. That was an interesting one to debug! We recreate the main activity after locale change by calling `recreate()` from `onActivityResult`. Not sure why, but maybe because it's called right before activity's `onResume` it prevents `onResume` of fragments from being called when they are replaced in the fragment container. Weird stuff. Calling `recreate()` outside of lifecycle events does not produce this issue. I tried to call it from `onPostResume` but it did not fix the issue.

Posting it to UI thread, even without any delay makes it work without any issues so that the fix I went with.

planarvoid:
thanks for the fix @khaykov , sounds like a fun issue to investigate :-). It works well now 👍 

> We got rid of the grey background in the app since it was not working too well with dark mode.

That makes sense but I think we should implement some UI changes to the stats because they look pretty bad and confusing now in the tablet mode. One approach I could think of is maybe a stroke around the cards? I think it's very hard to see where one card ends and the other begins (especially in the tablet mode). Let me know what you think :-) 

I'm not generally a huge fan of inheritance but maybe we can consider pulling this into a base activity? We can totally do it in a separate PR or issue though. 
```
@Override
    public void applyOverrideConfiguration(Configuration overrideConfiguration) {
        super.applyOverrideConfiguration(LocaleManager.updatedConfigLocale(getBaseContext(), overrideConfiguration));
    }
```

khaykov:
@planarvoid Yes, I agree about stats on a tablet. It's a bit hard to tell blocks apart. I'll direct the question to @mattmiklic , but I think we should address this in separate PR (as well as base activity, which is a good idea for all the locale things we are doing).

mattmiklic:
Thanks for the comment re: the stats cards on tablet -- totally agreed, and after talking with @SylvesterWilmott I have an idea of how to address it. I will share a visual with @khaykov when I have it and we can create a new issue to tackle that.

renanferrari:
Here is the issue and the PR regarding the stats cards on tablet:
- **Issue:** #11468
- **PR:** #11470

PR #11433, opened 2020-03-12
Comments : Prevent empty comment submission

### LINKED ISSUE #11137: Prevent empty looking comments from being submitted
When creating or editing a comment, it's currently possible to submit just space or newline character. I believe we could implement some sort of trimming logic to prevent this.

jd-alexander:
Fixes #11137
## Findings
The `EditText`s being used for the comments was only checking if text length was greater than 0 and not considering whitespace.

## Solution
`.trim()` is utilized to remove whitespace so that an empty comment can't be submitted.

## Testing
1. Go to Notifications.
2. Go to Comments.
3. Click on a comment.
4. In the Reply to section enter a whitespace character and realize the submit button isn't enabled. 
5. Also, press the up arrow in the comment box to expand the other comment component and test for the same behavior.
--------------
1. Go to Reader.
2. Go to Discover.
3. Click on the comments icon (between save post and like button).
4. In the Reply to section enter a whitespace character and realize the submit button isn't enabled. 
5. Also, press the up arrow in the comment box to expand the other comment component and test for the same behavior.

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/CommentFullScreenDialogFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/CommentFullScreenDialogFragment.kt
@@ -30,7 +30,6 @@ class CommentFullScreenDialogFragment : Fragment(), CollapseFullScreenDialogCont
     @Inject lateinit var siteStore: SiteStore
     private lateinit var dialogController: CollapseFullScreenDialogController
     private lateinit var reply: SuggestionAutoCompleteText
-    private lateinit var siteModel: SiteModel
 
     override fun onCreateView(
         inflater: LayoutInflater,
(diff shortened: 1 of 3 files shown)


jd-alexander:
Thanks for the review @ashiagr 😄 
I will look into these issues and make the necessary changes.

jd-alexander:
> Landscape mode scenario - send button got enabled after rotation to landscape orientation and tapping to write.

@ashiagr based on my investigation, this UI being displayed is that of the keyboard and we are injecting the comment's `EditText` to make the experience more streamlined. Unfortunately, the`SEND` button is not accessible so I am not able to enable/disable it. An approach we could take is showing a `Toast` if we determine that the value in the `EditText` isn't a valid comment, which is what I will be adding in a commit. 

Another issue I noticed was that there are times when the `SEND` button doesn't work when text is entered in the expanded fragment. I will create another issue to track this.

ashiagr:
LGTM, 🚢. Thanks!

PR #11447, opened 2020-03-16
Use event wrapper for single actions on crop screen

ashiagr:
This PR adds `Event` wrapper class for single actions on crop screen as part of https://github.com/wordpress-mobile/WordPress-Android/pull/11408 otherwise cropping error once displayed, gets displayed every time on orientation change because of an active observer.

PR submission checklist:


ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/viewmodel/Event.kt
```diff
@@ -0,0 +1,30 @@
+package org.wordpress.android.imageeditor.viewmodel
+
+/**
+ * Used as a wrapper for data that is exposed via a LiveData that represents an event.
+ */
+open class Event<out T>(private val content: T) {
```
malinajirka: I think it'd be nice to start thinking about extracting these common classes into a module so we don't need to create copies. It's out of scope for now, but we should keep it in mind so we don't forget about it.

malinajirka:
@ashiagr Please feel free to merge this when the conflict is resolved. Thanks!

Issue #11377, opened 2020-02-25
Low res(LR) image is saved to file on moving the app to background and bringing it back to foreground when images are loading


ashiagr:
### Expected behavior
High res(HR) image should be saved to file on moving the app to background and bringing it to foreground again while images are loading. 

### Actual behavior
Low res(LR) image is saved to file on moving the app to background while LR image is loading and bringing the app to foreground after it is loaded but before HR image is loaded. 

For more details, see https://github.com/wordpress-mobile/WordPress-Android/pull/11376#issue-379506529.

### Steps to reproduce the behavior
1. Clear the app cache.
2. Open preview image screen from the Gutenberg editor so that LR image loading starts. 
3. Click device home button to move the app to background.
4. Wait for a while and bring the app to foreground from the recent apps list. 
5. Notice that LR image is saved in the cache.

malinajirka:
Fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/11459

PR #11459, opened 2020-03-18
Handle image load failure on preview screen

### LINKED ISSUE #11346: Handle image load failure on preview screen
Goal of this task is to display an error placeholder in case of image load failure on the image preview screen as mentioned in the https://github.com/wordpress-mobile/WordPress-Android/pull/11222#issuecomment-584078153

ashiagr:
Fixes #11346

This PR displays a retry button when image to be edited fails to load on the preview screen.

PR submission checklist:


malinajirka:
Thanks @ashiagr!

The code looks good. However, I've encountered one bug and I'm also not entirely sure about the design. Having a basic button in the middle of the screen seems a bit weird. Wdyt?

1. Add an image into the editor
2. Turn on airplane mode
3. Click on "Edit" image
4. Notice retry button is shown
5. Click on retry
6. Turn off airplane mode
7. Click on retry
8. Notice the low resolution image is opened in uCrop

ashiagr:
I've updated retry UI and tried to fix the bug. Ready for another pass @malinajirka , thanks 🙏.

malinajirka:
Thanks @ashiagr ! The bug fix seems good. I have two minor notes to the design update:
1. The text is "grey" is that intentional? I'd expect a white text, since the icon is white. Wdyt?


2. Clicking on the retry icon doesn't do anything. I need to click on the text to invoke the refresh.

ashiagr:
> The text is "grey" is that intentional? I'd expect a white text, since the icon is white. Wdyt?

Fixed here d1ce2ed. Updated base theme to have white accent and `textColor` color so we don't  have to worry about it everytime.

> Clicking on the retry icon doesn't do anything. I need to click on the text to invoke the refresh.

Fixed here e3690b7. Clicking anywhere on the image view will trigger retry. Retry icon was set to `ImageButton` which interfered.

> The bug fix seems good. 

It seems to have this issue as well 😄 https://github.com/wordpress-mobile/WordPress-Android/issues/11377.

Ready for another look @malinajirka !

malinajirka:
>Fixed here d1ce2ed. Updated base theme to have white accent and textColor color so we don't have to worry about it everytime.

Perfect solution, thank you!

>Fixed here e3690b7. Clicking anywhere on the error layout will trigger retry. Retry icon was set to ImageButton which interfered.

Yeah, that was my guess. Works like a charm now!

>It seems to have fixed this issue as well 😄 #11377.

Tested and you are right, awesome! 

Looks great, thank you!

PR #11497, opened 2020-03-23
Adding reader_tag_loaded event when user selects a tag in subfilter.

### LINKED ISSUE #11496: reader_tag_loaded not triggered when changing tag.
### Expected behavior
When selecting/changing a tag in the Reader we expect the event reader_tag_loaded to be triggered.

### Actual behavior
When selecting/changing a tag in the Reader the event reader_tag_loaded is not triggered.

Select or change a tag in the reader (select it from the filter bottom sheet) and check that the reader_tag_loaded is not triggered in Logcat

develric:
Fixes #11496 

With IA reader changes (version 14.3) we moved the way we select tags from being a spinner in the action bar to being a list item in a bottom sheet. In this way we are not triggering the reader_tag_loaded.

This PR reintroduces that event when user select a tag from the bottom sheet.

## To test
Make sure events are enabled in the app settings. Then you can check the event is triggered or not looking in Logcat and filtering with `Tracked: ` string.

- Open the reader and remove any filter in Following
- Move between tabs and check the event is not triggered
- Select a Site from the subfilter and check the event is not triggered (also moving between Discover/Likes etc... tabs)
- Select a tag and check the event is triggered
- Move between tabs and check that the event is not triggered moving away from the Following tab but is triggered when coming back to Following (I think this is correct since you are accessing again to the posts of that stream, it is similar to selecting again that tag stream from the spinner as we were doing before)
- Remove the filter and check the event is not triggered
- Opening a post from a tag or another stream should not trigger the event

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
@@ -518,6 +518,10 @@ public class ReaderPostListFragment extends Fragment
                               || (mCurrentBlogId != readerModeInfo.getBlogId())
                               || (mCurrentFeedId != readerModeInfo.getFeedId())
                               || (readerModeInfo.isFirstLoad());
+
+            if (changesDetected && !readerModeInfo.isFirstLoad()) {
+                trackTagLoaded(readerModeInfo.getTag());
+            }
         }
 
         if (onlyOnChanges && !changesDetected) return;


malinajirka:
Just for future reference:
We are aware of two inconsistencies introduced by this change, but at least the first one isn't imo solvable.
1. If the user selects a tag (eg. "Nature") -> they select a discovery tab -> they want to see "City" tag now -> they have to go back to the "following" tab and the event for "Nature" is triggered. We changed the UI and it's not possible to change the tag from any other subscreen than `Reader-Following`. I don't think there is a solution for making the event 100% consistent with the previous version.
2. When the app is restarted the previous version of the app would trigger the event. The new version does't trigger the event in this scenario. However, I think the new behavior makes more sense. Since we are not able to make the events 100% consistent (due to issue 1) we decided not to worry about this minor difference.

PR #11510, opened 2020-03-24
Add analytics to image editor

### LINKED ISSUE #10984: Track usage
We want to track the usage of the Media Editor to:

1. Discover how many users are using it
2. Discover what operations the user is doing with the image

For that we will:

1. Fire `wpandroid_media_editor_show` when the Media Editor is shown
2. Fire `wpandroid_media_editor_used` with property `actions` (more details below) when the user edits the image

### Actions property

We will send the below actions:

1. `crop`: when the user cropped the image
2. `rotate`: when the user rotates the image
3. `crop, rotate`: when the user rotates and crop the image

ashiagr:
Fixes #10984

This PR adds given analytics tracks events to track usage.

To test:
- Ensure `Collect Information` option is enabled in user's App's Settings -> Privacy settings
- Launch image editor by clicking edit option for an image in the Gutenberg editor (need to enable edit option) 
- Check that the console contains the following log:
    - `🔵 Tracked: media_editor_shown`
- Click Done (check) button on Crop screen
- Check that the console contains the following log:
   - `🔵 Tracked: media_editor_used, Properties: {"actions":"crop"}`

P.S. `uCrop` doesn't send any data in crop result to differentiate between `crop`, `rotate`, `scale` actions. As per internal conversation, tracking just `crop` action on done button click is sufficient based on iOS tracks analysis to avoid any hack.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/posts/editor/ImageEditorInitializer.kt
```diff
@@ -64,5 +70,21 @@ class ImageEditorInitializer {
             } else {
                 throw(IllegalArgumentException(IMAGE_STRING_URL_MSG))
             }
+
+        private fun trackData(): (Pair<TrackableEvent, TrackableAction?>) -> Unit = { data ->
```
malinajirka: NitPick: We could simplify this by using kotlin's feature: 
Replace
```
data -> 
           val event = data.first
            val action = data.second
```
with
`(event, action) -> 
`

Wdyt?

malinajirka (commented):
Thanks @ashiagr!

I haven't done a full review yet. I might be missing something, but how would you feel about tracking the events [here](https://github.com/wordpress-mobile/WordPress-Android/blob/4c72fc2938c9d27b5edb1d28bd9aecbd31cee2d3/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L1681) and [here](https://github.com/wordpress-mobile/WordPress-Android/blob/4c72fc2938c9d27b5edb1d28bd9aecbd31cee2d3/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L2274)? We wouldn't need to touch the library at all and afaict it'd track exactly the same events. Moreover, our tracking wouldn't be part of a general-purpose library. Wdyt?

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/ImageEditor.kt
```diff
@@ -46,20 +47,36 @@ class ImageEditor private constructor(
         loadIntoFileWithResultListener.invoke(imageUrl, listener)
     }
 
+    fun trackEventWithActionData(data: Pair<TrackableEvent, TrackableAction?>) {
+        trackData(data)
```
malinajirka: How would you feel about using a sealed class instead of two enums? Using two enums is in general fine if all the combinations are valid. However, the current interface supports for example sending `TrackableAction-CROP with TrackableEvent-MEDIA_EDITOR_SHOWN` which AFAICT doesn't make sense. Wdyt?

ashiagr:
Thanks @malinajirka ! As discussed, created a new PR https://github.com/wordpress-mobile/WordPress-Android/pull/11516 to track events inside `WPAndroid` freeing the library from any analytics tracking 🙂.

malinajirka:
@ashiagr I think we can close this PR, right? Just make sure NOT to delete the branch :).

ashiagr:
Yes we can close. Will not delete the branch.

Issue #10930, opened 2019-12-10
Login: The Email Address error is not preserved once the screen is rotated


yesunch:
### Expected behavior

When writing the email in a wrong format error message appears. When rotating the screen the message should be preserved.

### Actual behavior

When rotating the screen the message is not preserved.

### Steps to reproduce the behavior
1. Press Log In
2. Enter Email address and press Next. An error message will be shown
3. Rotate the screen. The message is not shown anymore

##### Tested on Samsung S10, Android 9, WPAndroid 13.7
![WP-login-issue](https://user-images.githubusercontent.com/39059254/70522031-a3e87980-1b40-11ea-92f5-fdf24a22fb60.gif)

jd-alexander:
Behavior here is similar to this [issue](https://github.com/wordpress-mobile/WordPress-Android/issues/10361) so a fix could be applied to both.

FernandaCG:
Hi! can I work on this?

jd-alexander:
@FernandaCG Sure! definitely can!

FernandaCG:
Hi! Can you check if my solution solve both issues? :D

FernandaCG:
Hi @jd-alexander, any update?

jd-alexander:
Hi @FernandaCG I will have a look soon!

namanrana16:
Hey, Can I work on this issue?

jd-alexander:
Hi @namanrana16 Thank you for your interest in contributing to WordPress Android! This issue was already resolved in https://github.com/wordpress-mobile/WordPress-Android/pull/11537 so I will be closing this!

PR #11537, opened 2020-03-26
Fix/10930 email error dissapears on rotation - continued

jd-alexander:
Fixes #11003

## Findings
This PR is a continuation of https://github.com/wordpress-mobile/WordPress-Android/pull/11003 where the email error wasn't being preserved when a configuration change took place.

## Testing
1. Press Log In
2. Enter an invalid email address and press Next. An error message will be shown
3. Rotate the screen.
4. Notice the error is still shown. 

## Reviewing 
2. Once approved, close #11003

## Submitter Checklist

### DIFF
--- a/libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginEmailFragment.java
+++ b/libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginEmailFragment.java
@@ -64,7 +64,7 @@ public class LoginEmailFragment extends LoginBaseFormFragment<LoginListener> imp
     private static final String KEY_IS_SOCIAL = "KEY_IS_SOCIAL";
     private static final String KEY_OLD_SITES_IDS = "KEY_OLD_SITES_IDS";
     private static final String KEY_REQUESTED_EMAIL = "KEY_REQUESTED_EMAIL";
-    private static final String VALIDITY_EMAIL = "KEY_VALIDITY_EMAIL";
+    private static final String KEY_EMAIL_ERROR_RES = "KEY_EMAIL_ERROR_RES";
     private static final String LOG_TAG = LoginEmailFragment.class.getSimpleName();
     private static final int GOOGLE_API_CLIENT_ID = 1002;
     private static final int EMAIL_CREDENTIALS_REQUEST_CODE = 25100;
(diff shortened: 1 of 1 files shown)


ashiagr (commented):
👋@jd-alexander 😊!  I tested your changes, code-wise it looks good and it should have worked but it seems Android behaves differently 😖and `TextWatcher's onTextChanged()` is called again when phone orientation is changed which alters the behaviour. 

If it might help, tested on 
Pixel 3a, API level 29
Pixel 3, API level 27 

preserve_email

jd-alexander:
Thanks @ashiagr will check it out!

jd-alexander:
I removed the clearing of the `mIsValidEmail` 3eb51e8 Let me know what you think. Do you think it has to be preserved for all states?

ashiagr:
> I removed the clearing of the mIsValidEmail

Thanks @jd-alexander ! 

I was wondering if we can delay adding text changed listener till the screen renders after rotation as described [here](https://stackoverflow.com/a/29449717/193545) while retaining `mIsValidEmail` inside `onTextChanged`. How about something like this?

```
mEmailInput.getViewTreeObserver().addOnGlobalLayoutListener(new ViewTreeObserver.OnGlobalLayoutListener() {
    @Override public void onGlobalLayout() {
        mEmailInput.getViewTreeObserver().removeOnGlobalLayoutListener(this);
        mEmailInput.addTextChangedListener(LoginEmailFragment.this);
    }
});
```

This way we can avoid redisplay of error text on rotation once it is cleared after text change.

email-addr

Also since we're trying to preserve email address error, I was thinking to also preserve another email error `email_not_registered_wpcom`.  We can create a new issue if you feel so. 

device-2020-04-01-102545

jd-alexander:
> I was wondering if we can delay adding text changed listener till the screen renders after rotation as described here while retaining mIsValidEmail inside onTextChanged. How about something like this?

Nice @ashiagr Knew you would have something cool I didn't remember 😉 

I used `.post()` instead and it works the same. Let me know if that's fine 26cef97

> Also since we're trying to preserve email address error, I was thinking to also preserve another email error email_not_registered_wpcom. We can create a new issue if you feel so.

Makes sense! Good eye 😄 Since we are here right now, I just decided to support all the email errors in this class. 
Done in 709c35e

ashiagr (approved):
Perfect @jd-alexander 🎉! LGTM 🚢.

P.S. As discussed on slack, we should also push the changes to the login lib repo.

hypest:
👋 @jd-alexander , does this PR also address this other ticket? https://github.com/wordpress-mobile/WordPress-Android/issues/10930. Let's close it if it does, thanks!

jd-alexander:
Yes, it does @hypest Closing!

PR #11540, opened 2020-03-27
Add reblog to reader post list

### LINKED ISSUE #11532: Add reblog functionality to Reader posts: Add reblog action to post cards in stream view
Master issue #10934 

A reblog button should be added to the reader post card actions in a stream view.

If the user has more that one sites the button action should trigger a site picker.

picker

When the users selects a site the editor opens with the following information prefilled:
* Title (with source)
* Image (if the post has a featured)
* Quote
* Citation

editor

When the user publishes the article the flow returns to the stream view. A snackbar is shown with a view post action.

antonis:
Fixes #11532 

## Description

This feature adds reblog action to post cards in stream view

A reblog button has been added in the actions of the post cards in stream view.
If the user has more that one sites the button triggers the existing Site Picker.

| Post Details |Site Picker (unchanged)|
|--- |---|
|device-2020-03-27-082924||

If the user has only one site the editor is triggered directly.
At the moment handling users with no site is not supported.

When the users selects a site he is redirected to the editor with the following information prefilled:
* Title (with source)
* Image (if the post has a featured)
* Quote
* Citation

| Gutenberg editor with featured image| Gutenberg editor without image |
| --- | --- |
|device-2020-03-22-114527||

| Aztec editor with featured image| Aztec editor without image |
| --- | --- |
|device-2020-03-22-114630|device-2020-03-22-114609|

When the user taps the Publish button and confirms he returns to the original context

| Uploading State| Post Published |
| --- | --- |
|device-2020-03-27-083017|device-2020-03-27-083046|

## Test

0. Open the App and Login if needed
1. Select the Reader tab
4. If prompted select a site from the picker
5. Notice that the editor opens with content from the original post: Title (with source), Image (if the post has a featured), Quote and Citation link
6. Optionally edit the post
7. Press the Publish button
8. Confirm by pressing the Publish Now button in the dialog
9. Notice that you return to the original context (post list view)
10. Notice the snackbar with the VIEW button
11. Tap on the VIEW button
12. Examine the reblogged post in the browser

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListActivity.java
@@ -1,5 +1,6 @@
 package org.wordpress.android.ui.reader;
 
+import android.app.Activity;
 import android.content.ActivityNotFoundException;
 import android.content.Context;
 import android.content.Intent;
@@ -17,15 +18,33 @@ import androidx.fragment.app.Fragment;
 
 import com.google.android.material.appbar.AppBarLayout;
 
+import org.greenrobot.eventbus.Subscribe;
+import org.greenrobot.eventbus.ThreadMode;
 import org.wordpress.android.R;
+import org.wordpress.android.WordPress;
 import org.wordpress.android.analytics.AnalyticsTracker;
 import org.wordpress.android.analytics.AnalyticsTracker.Stat;
 import org.wordpress.android.datasets.ReaderBlogTable;
+import org.wordpress.android.fluxc.Dispatcher;
+import org.wordpress.android.fluxc.model.PostModel;
+import org.wordpress.android.fluxc.model.SiteModel;
+import org.wordpress.android.fluxc.store.PostStore;
+import org.wordpress.android.fluxc.store.PostStore.OnPostUploaded;
+import org.wordpress.android.fluxc.store.SiteStore;
 import org.wordpress.android.models.ReaderBlog;
 import org.wordpress.android.models.ReaderTag;
+import org.wordpress.android.ui.ActivityLauncher;
+import org.wordpress.android.ui.RequestCodes;
+import org.wordpress.android.ui.posts.EditPostActivity;
+import org.wordpress.android.ui.prefs.AppPrefs;
 import org.wordpress.android.ui.reader.ReaderTypes.ReaderPostListType;
+import org.wordpress.android.ui.uploads.UploadActionUseCase;
+import org.wordpress.android.ui.uploads.UploadUtils;
+import org.wordpress.android.ui.uploads.UploadUtilsWrapper;
 import org.wordpress.android.util.LocaleManager;
 import org.wordpress.android.util.ToastUtils;
+
+import javax.inject.Inject;
 
 /*
  * serves as the host for ReaderPostListFragment when showing blog preview & tag preview
@@ -33,6 +52,12 @@ import org.wordpress.android.util.ToastUtils;
 public class ReaderPostListActivity extends AppCompatActivity {
     private ReaderPostListType mPostListType;
     private long mSiteId;
+
+    @Inject SiteStore mSiteStore;
+    @Inject PostStore mPostStore;
+    @Inject Dispatcher mDispatcher;
+    @Inject UploadActionUseCase mUploadActionUseCase;
+    @Inject UploadUtilsWrapper mUploadUtilsWrapper;
 
     @Override
     protected void attachBaseContext(Context newBase) {
@@ -42,6 +67,8 @@ public class ReaderPostListActivity extends AppCompatActivity {
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        ((WordPress) getApplication()).component().inject(this);
+
         setContentView(R.layout.reader_activity_post_list);
 
         Toolbar toolbar = findViewById(R.id.toolbar);
@@ -105,6 +132,19 @@ public class ReaderPostListActivity extends AppCompatActivity {
         super.onResumeFragments();
         // this particular Activity doesn't show filtering, so we'll disable the FilteredRecyclerView toolbar here
         disableFilteredRecyclerViewToolbar();
+    }
+
+    @Override
+    protected void onResume() {
+        super.onResume();
+        // We register the dispatcher in order to receive the OnPostUploaded event and show the snackbar
+        mDispatcher.register(this);
+    }
+
+    @Override
+    protected void onPause() {
+        super.onPause();
+        mDispatcher.unregister(this);
     }
 
     /*
@@ -263,5 +303,56 @@ public class ReaderPostListActivity extends AppCompatActivity {
             return null;
         }
         return ((ReaderPostListFragment) fragment);
+    }
+
+    @Override
+    protected void onActivityResult(int requestCode, int resultCode, Intent data) {
+        super.onActivityResult(requestCode, resultCode, data);
+
+        if (requestCode == RequestCodes.EDIT_POST && resultCode == Activity.RESULT_OK
+            && data != null && !isFinishing()) {
+            int localId = data.getIntExtra(EditPostActivity.EXTRA_POST_LOCAL_ID, 0);
+            final SiteModel site = (SiteModel) data.getSerializableExtra(WordPress.SITE);
+            final PostModel post = mPostStore.getPostByLocalPostId(localId);
+
+            if (EditPostActivity.checkToRestart(data)) {
+                ActivityLauncher.editPostOrPageForResult(data, ReaderPostListActivity.this, site,
+                        data.getIntExtra(EditPostActivity.EXTRA_POST_LOCAL_ID, 0));
+                // a restart will happen so, no need to continue here
+                return;
+            }
+
+            if (site != null && post != null) {
+                mUploadUtilsWrapper.handleEditPostResultSnackbars(
+                        this,
+                        findViewById(R.id.coordinator),
+                        data,
+                        post,
+                        site,
+                        mUploadActionUseCase.getUploadAction(post),
+                        new View.OnClickListener() {
+                            @Override
+                            public void onClick(View v) {
+                                UploadUtils.publishPost(ReaderPostListActivity.this, post, site, mDispatcher);
+                            }
+                        });
+            }
+        }
+    }
+
+    @SuppressWarnings("unused")
+    @Subscribe(threadMode = ThreadMode.MAIN)
+    public void onPostUploaded(OnPostUploaded event) {
+        int siteLocalId = AppPrefs.getSelectedSite();
+        SiteModel site = mSiteStore.getSiteByLocalId(siteLocalId);
+        if (site != null && event.post != null) {
+            mUploadUtilsWrapper.onPostUploadedSnackbarHandler(
+                    this,
+                    findViewById(R.id.coordinator),
+                    event.isError(),
+                    event.post,
+                    null,
+                    site);
+        }
     }
 }
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
@@ -2901,16 +2901,18 @@ public class ReaderPostListFragment extends Fragment
     @Override
     public void reblog(ReaderPost post) {
         List<SiteModel> sites = mSiteStore.getVisibleSites();
+        int selectedSiteId = AppPrefs.getSelectedSite();
+        SiteModel selectedSite = mSiteStore.getSiteByLocalId(selectedSiteId);
         switch (sites.size()) {
             case 0:
                 ToastUtils.showToast(getActivity(), R.string.reader_no_site_to_reblog);
                 break;
             case 1:
-                ActivityLauncher.openEditorForReblog(getActivity(), getSelectedSite(), post);
+                ActivityLauncher.openEditorForReblog(getActivity(), selectedSite, post);
                 break;
             default:
                 this.mPostToReblog = post; // Stores the post to be handled in the onActivityResult after site selection
-                ActivityLauncher.showSitePickerForResult(this, getSelectedSite());
+                ActivityLauncher.showSitePickerForResult(this, selectedSite);
                 break;
         }
     }
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
@@ -2078,10 +2078,10 @@ public class ReaderPostListFragment extends Fragment
                     context,
                     getPostListType(),
                     mImageManager,
-                    mIsTopLevel,
-                    this
+                    mIsTopLevel
             );
             mPostAdapter.setOnFollowListener(this);
+            mPostAdapter.setReblogActionListener(this);
             mPostAdapter.setOnPostSelectedListener(this);
             mPostAdapter.setOnPostPopupListener(this);
             mPostAdapter.setOnDataLoadedListener(mDataLoadedListener);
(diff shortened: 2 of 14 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -2888,4 +2897,30 @@ protected void onPostExecute(ReaderTagList tagList) {
...
+
+    @Override
+    public void reblog(ReaderPost post) {
+        List<SiteModel> sites = mSiteStore.getVisibleSites();
+        switch (sites.size()) {
+            case 0:
+                ToastUtils.showToast(getActivity(), R.string.reader_no_site_to_reblog);
+                break;
+            case 1:
+                ActivityLauncher.openEditorForReblog(getActivity(), getSelectedSite(), this.mPostToReblog);
```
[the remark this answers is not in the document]
antonis: You are write. Fixed with 96de3a8d7a3650aa305ecad3cecbb50ca3e816af

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -2070,7 +2078,8 @@ private ReaderPostAdapter getPostAdapter() {
                     context,
                     getPostListType(),
                     mImageManager,
-                    mIsTopLevel
+                    mIsTopLevel,
+                    this
             );
             mPostAdapter.setOnFollowListener(this);
```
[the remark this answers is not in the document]
antonis: Fixed with e0891a014f91e8395394fa40be20013f3560ef6a

antonis:
> I think this is due to how we recover the current site in `getSelectedSite`relaying on WPMainActivity being the activity of the fragment that is not the case in the above list. 

You are right. Fixed with 9caa59a811213738bd4dc349b964acb73c329a04

antonis:
>  I think the reblog button spacing in the detailed view (see right below) is a bit too near respect what we get in the main reader (see left)

I missed the margin. Fixed with e779301f2c6e6cb55429115a93254b34dc7ba88f

antonis:
> I'm also reporting here (for tracing purpose but no action for now, we can dig it later) that I'm getting the following error when reblogging:
> 
> ```
> An error occurred while fetching History revisions
> ```
> 
> I think it can have to do with the fact that we are triggering the fetch of the revisions but the post is still present in remote so cannot have a history (we will need to dig further about this one).

Ok. I open #11563 to track this.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -2888,4 +2897,30 @@ protected void onPostExecute(ReaderTagList tagList) {
...
+                this.mPostToReblog = post; // Stores the post to be handled in the onActivityResult after site selection
+                ActivityLauncher.showSitePickerForResult(this, getSelectedSite());
+                break;
+        }
+    }
+
+    @Override public void onActivityResult(int requestCode, int resultCode, Intent data) {
+        super.onActivityResult(requestCode, resultCode, data);
+        if (requestCode == RequestCodes.SITE_PICKER && resultCode == Activity.RESULT_OK) {
+            int siteLocalId = data.getIntExtra(SitePickerActivity.KEY_LOCAL_ID, -1);
```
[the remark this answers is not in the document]
antonis: Refactored this with e5be5a22fca9edee16fc4f37c1c038a8f380bf24

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
```diff
@@ -249,6 +261,37 @@ class ReaderPostListViewModel @Inject constructor(
...
+            else -> _reblogAction.value = Event(SitePicker(selectedSite, post))
+        }
+    }
+
+    /**
+     * Handles site selection
+     *
+     * @param site selected site to reblog to
+     */
+    fun selectedSiteToReblog(siteLocalId: Int) {
```
develric: Wdyt on keeping the same `event style` naming here? Something like `onReblogSiteSelected`?

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
```diff
@@ -249,6 +261,37 @@ class ReaderPostListViewModel @Inject constructor(
...
     }
 
+    /**
+     * Handles reblog button action
+     *
+     * @param post post to reblog
+     */
+    fun onReblogButtonClicked(post: ReaderPost) {
+        val selectedSiteId = appPrefsWrapper.getSelectedSite()
+        val selectedSite: SiteModel = siteStore.getSiteByLocalId(selectedSiteId)
```
develric: `getSiteByLocalId ` can return a null value (see previous comment in ReblogState); I think we should take care of this possibility.

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
```diff
@@ -88,6 +97,9 @@ class ReaderPostListViewModel @Inject constructor(
     private val _updateTagsAndSites = MutableLiveData<Event<EnumSet<UpdateTask>>>()
     val updateTagsAndSites: LiveData<Event<EnumSet<UpdateTask>>> = _updateTagsAndSites
 
+    private val _reblogAction = MutableLiveData<Event<ReblogState>>()
+    val reblogAction: LiveData<Event<ReblogState>> = _reblogAction
```
develric: Since it is a state, maybe we could name it `_reblogState`? wdyt?

WordPress/src/test/java/org/wordpress/android/viewmodel/reader/ReaderPostListViewModelTest.kt
```diff
@@ -351,6 +360,90 @@ class ReaderPostListViewModelTest {
...
+        whenever(appPrefsWrapper.getSelectedSite()).thenReturn(siteId)
+        whenever(siteStore.getSiteByLocalId(siteId)).thenReturn(site)
+        whenever(siteStore.visibleSites).thenReturn(visibleSites)
+
+        viewModel.onReblogButtonClicked(post)
+        viewModel.selectedSiteToReblog(siteId)
+
+        val state = viewModel.reblogAction.value?.peekContent()
+        assert(state is PostEditor)
+        assertThat(state?.site).isEqualTo(site)
```
develric: This can be a personal preference 😊; in general I like the assertThat readability more than the pure assert function. Wdyt in using them also for things like `assert(state is PostEditor)`? Something like `assertThat(state).isInstanceOf(PostEditor::class.java)`?

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
```diff
@@ -249,6 +261,37 @@ class ReaderPostListViewModel @Inject constructor(
...
+     * @param site selected site to reblog to
+     */
+    fun selectedSiteToReblog(siteLocalId: Int) {
+        val currentState = _reblogAction.value?.peekContent()
+        val selectedPost = currentState?.post
+        if (currentState is SitePicker && selectedPost != null) {
+            val site = siteStore.getSiteByLocalId(siteLocalId)
+            _reblogAction.value = Event(PostEditor(site, selectedPost))
+        } else {
+            _reblogAction.value = Event(ReblogError)
```
develric: Good to see you thought to some error handling here 👍! In cases like this I think we are dealing with some condition we do not expect (possibly a bug), so maybe worth to enforce fail fast at least for Debug configurations adding also something like this

```
                if (BuildConfig.DEBUG) {
                    throw IllegalStateException("Error Message")
                } else {
                    AppLog.e(T.READER, "Error Message")
                }
```

Wdyt? Let me know if I got your general intention wrong of course 😊.

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
antonis: I agree. Fixed 074a7a0e1cdb0cb89e06c885b76adfda130fb11c

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
antonis: I actually encountered this in the no site use case and handled it in #11553 [here](https://github.com/wordpress-mobile/WordPress-Android/pull/11553/files#diff-4d1136c584e903ce48a0094d9d581b44R269).
I will leave it as is to avoid a conflict for now.

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
antonis: I was in a dilemma on this. It's a state but also an event/action.
I will agree with you and change the name.
Fixed 7743337c0c59a0e22c21c982a28fa87176f5d935

WordPress/src/test/java/org/wordpress/android/viewmodel/reader/ReaderPostListViewModelTest.kt
antonis: Tbh I don't like the verbosity of the `::class.java` suffix.
It would be more concise with the rest of the tests though.
Fixed cf3017354faf1278810179b5955dcbf3b01d6a9e

WordPress/src/main/java/org/wordpress/android/ui/reader/reblog/ReblogState.kt
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Represents the No Site View State
+ */
+object NoSite : ReblogState()
+
+/**
+ * Represents the Error State
+ */
+object ReblogError : ReblogState()
```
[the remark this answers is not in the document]
antonis: > Having this last consideration, since NoSite and ReblogError are not keeping state, what do you think about moving site and post only to SitePicker and PostEditor? 

You are right, The NoSite/ReblogError cases do not keep any state. I've found it useful in the past to have some common properties in the top level sealed class (even if they are not used in some cases) to be able to handle states in a uniform way or to avoid ugly casting especially in java. Type safety is an undesired tradeoff of course.

> side note: in general I would also consider to create a null object state here like Unknown or similar, but in this case we are going to use it in a live variable in kotlin code so maybe not much value added

My intention was to use the ErrorState as Unknown. Probably my naming is bad. I'm not sure if both can fit in our case.

I made fixes based on your feedback with 490ae8aab850f4f0e2f9c1e6c38b54d8a80eac07

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
antonis: Also tackled with the state refactoring 490ae8aab850f4f0e2f9c1e6c38b54d8a80eac07

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
antonis: Fixed with 490ae8aab850f4f0e2f9c1e6c38b54d8a80eac07

WordPress/src/main/java/org/wordpress/android/ui/reader/reblog/ReblogState.kt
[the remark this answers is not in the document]
develric: > ...avoid ugly casting especially in java

Agree; just to share maybe could make sense in cases like this to have a type enum in the base sealed class and manage the two with state classes as a data class (discriminated by the type that should work in Java also)

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
develric: Sharing some thoughts below, I'm not sure if I'm missing something. 

> val selectedSiteId: Int? = appPrefsWrapper.getSelectedSite()

I think selectedSiteId should not be nullable (it can be eventually -1)

> val selectedSite: SiteModel?

I agree selectedSite can be nullable 👍

```
        when (siteStore.visibleSites.size) {
            0 -> _reblogState.value = Event(NoSite)
            1 -> _reblogState.value = selectedSite?.let { Event(PostEditor(it, post)) } ?: Event(Unknown)
            else -> _reblogState.value = selectedSite?.let { Event(SitePicker(it, post)) } ?: Event(Unknown)
        }
```

Maybe we can make it a bit more readable using the when as an expression, something like:

```
       _reblogState.value = when (siteStore.visibleSites.size) {
            0 -> Event(NoSite)
            1 -> if (selectedSite == null) Event(PostEditor(selectedSite, post)) else Event(Unknown)
            else -> if (selectedSite == null) Event(SitePicker(selectedSite, post)) else Event(Unknown)
        }
```

Also `onReblogSiteSelected` I think can have the same issue with `getSiteByLocalId`

wdyt?

antonis: I agree. Fixed with c9b69578f905b9e8b9bf8e6bd1a8bd1cac57f2d7

antonis:
When reblogging to a site that is not the currently selected one the _post published_ snackbar was not showing. I fixed this by removing the check that the site of the post published is the same with the currently selected site.
`event.post.getLocalSiteId() == site.getId()`
Commit 030f6594738dc8d25336f12553c4e91d9c112447
I tested this both in publishing a post and when pressing back to save the post in the drafts. It seems to have no side effects but I'm not sure if the check was there for security reasons or to handle another state I'm not testing.
@develric Please let me know if you have any feedback on this.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostPagerActivity.java
```diff
@@ -1051,7 +1051,7 @@ public void onPositiveClicked(@NotNull String instanceTag) {
     public void onPostUploaded(OnPostUploaded event) {
         int siteLocalId = AppPrefs.getSelectedSite();
         SiteModel site = mSiteStore.getSiteByLocalId(siteLocalId);
-        if (site != null && event.post != null && event.post.getLocalSiteId() == site.getId()) {
+        if (site != null && event.post != null) {
```
develric, line 1054: This is a good point. I agree that check is an explicit guard condition to avoid to manage posts not related to the current site. I also didn't find an immediate issue removing it so I'm accepting your solution for now. One thing we could maybe do is to check if the event siteId is from a site we are allowed to post to.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListActivity.java
```diff
@@ -264,4 +304,55 @@ private ReaderPostListFragment getListFragment() {
...
+                        post,
+                        site,
+                        mUploadActionUseCase.getUploadAction(post),
+                        new View.OnClickListener() {
+                            @Override
+                            public void onClick(View v) {
+                                UploadUtils.publishPost(ReaderPostListActivity.this, post, site, mDispatcher);
+                            }
+                        });
+            }
```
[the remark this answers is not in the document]
antonis: Yes. It is actually the 3rd place that repeats this logic since it was already in WPMainActivity.
It the logic should be delegated to a common component or use the Activity Result APIs introduced in AndroidX https://developer.android.com/training/basics/intents/result
I was thinking of proceeding with the PRs and open another issue for this since it could also touch the main activity.

PR #11541, opened 2020-03-27
Cover block upload completion processor

mkevins:
Fixes: https://github.com/wordpress-mobile/gutenberg-mobile/issues/1739

## Description
This PR implements the upload completion processor for the Cover block. 

#### Related PRs:

* `gutenberg`: TBD
* `gutenberg-mobile`: TBD
* `WordPress-iOS`: https://github.com/wordpress-mobile/WordPress-iOS/pull/13928

##### Unit tests:

Run all unit tests in `WordPress/src/test/java/org/wordpress/android/ui/posts/mediauploadcompletionprocessors/`

##### Manual tests:

- [ ] Close/Re-open post with an ongoing image upload - [steps](https://github.com/wordpress-mobile/test-cases/blob/master/test-cases/gutenberg/cover.md#tc010)
- [ ] Close post with an ongoing image upload - [steps](https://github.com/wordpress-mobile/test-cases/blob/master/test-cases/gutenberg/cover.md#tc011)

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/posts/mediauploadcompletionprocessors/CoverBlockProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.regex.Pattern;
+
+public class CoverBlockProcessor extends BlockProcessor {
+    /**
+     * Template pattern used to match and splice cover inner blocks
+     */
+    private static final Pattern PATTERN_COVER_INNER = Pattern.compile(new StringBuilder()
+                    .append("(^.*?<div class=\"wp-block-cover__inner-container\">\\s*)")
+                    .append("(.*)") // inner block contents
+                    .append("(\\s*</div>\\s*</div>\\s*<!-- /wp:cover -->.*)").toString(), Pattern.DOTALL);
```
dmsnell, line 19: take a look at `blocks.php` for some inspiration on how to process blocks. with a full parser we can remove the entirety of the need to parse the document here and focus on the structural form of the block

dmsnell, line 19: oops. I had left a comment here that was wrong. you can take some inspiration from `hpq` which @aduth wrote. something like [jsoup](https://jsoup.org) (random internet search) could automate the inference of the block attributes as defined in the block implementation, rather than falling back to attempts at parsing hTML with RegExp

mkevins: Hi @dmsnell :wave: :smiley: , thanks for taking a look at this one. I'm using jsoup already, which is really helpful. IIRC, I encountered some minor differences in output in certain scenarios, possible because it parses to an AST and not a parse tree, however, it is probably worth revisiting, as the variations might still be acceptable to pass validation. :+1:

WordPress/src/main/java/org/wordpress/android/ui/posts/mediauploadcompletionprocessors/BlockProcessor.java
```diff
@@ -146,4 +120,33 @@ String processBlock(String block) {
...
+     * {@link MediaUploadCompletionProcessor#processContent(String)} for block types that have media-containing blocks
+     * within their inner content.<br>
+     * <br>
+     * The default implementation provided is a NOOP that leaves the content of the block unchanged.
+     *
+     * @param block The raw block contents
+     * @return A string containing content with ids and urls replaced
+     */
+    String processInnerBlock(String block) {
+        return block;
```
dmsnell, line 150: Is there a reason to not make the default processor process inner blocks? It will come surprisingly often that the blocks we care about are nested, for example, when people build with the full-site-editing layout blocks, or the Group block, or the Quote block, or the Gallery block…

WordPress/src/main/java/org/wordpress/android/ui/posts/mediauploadcompletionprocessors/CoverBlockProcessor.java
dmsnell: > I encountered some minor differences in output in certain scenarios, possible because it parses to an AST and not a parse tree

Do you remember what those differences were? Any difference would have to be a bug since we're dealing with HTML/CSS selectors…

dmsnell: also it looks like this entire pattern shouldn't be necessary if we process the inner blocks using a spec-compliant parser

WordPress/src/main/java/org/wordpress/android/ui/posts/mediauploadcompletionprocessors/CoverBlockProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Pattern to match background-image url in cover block html content
+     */
+    private static final Pattern PATTERN_BACKGROUND_IMAGE_URL = Pattern.compile("background-image:url\\([^\\)]+\\)");
```
dmsnell, line 24: so definitely there's an issue with blocks defining their attributes in the "JSON attributes" vs. inside HTML, but instead of hard-coding a string-matching pattern we can go back to `jsoup` as you were doing and hard-code the [known attribute definition](https://github.com/WordPress/gutenberg/blob/master/packages/block-library/src/cover/block.json), which would then end up looking like this…

```java
JSONAttribute backgroundImageUrl = new JSONAttribute("url", JSONAttribute.String);
```

I'm sure there's a million ways to do this. The main point I was trying to make is that we can rely on the structural form of the data instead of processing the HTML string as if it were plain text. It's quite possible to have `background-image:url` inside the block in a way that isn't meant to indicate a different background, and this string-matching might end up paving a road of headaches to debug 😄

WordPress/src/main/java/org/wordpress/android/ui/posts/mediauploadcompletionprocessors/CoverBlockProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+    @Override boolean processBlockContentDocument(Document document) {
+        // select cover block div
+        Element targetDiv = document.select(".wp-block-cover").first();
+
+        // if a match is found, proceed with replacement
+        if (targetDiv != null) {
+            // replace background-image url in style attribute
+            String style = PATTERN_BACKGROUND_IMAGE_URL.matcher(targetDiv.attr("style"))
+                    .replaceFirst(String.format("background-image:url(%1$s)", mRemoteUrl));
+            targetDiv.attr("style", style);
```
dmsnell, line 70: hm…I'm seeing the cover block actually store the background imagine inside the JSON attributes. this will deifnitely update the style, but won't the newly-uploaded image get wiped out after someone loads the post in the editor again and there's a mismatch between what the attributes indicate and what HTML has been generated?

we probably need to update both places.

WordPress/src/main/java/org/wordpress/android/ui/posts/mediauploadcompletionprocessors/BlockProcessor.java
mkevins: > Is there a reason to not make the default processor process inner blocks?

For the current implementation of the processors, they are not full-blown parsers, but part of their behavior resembles a parser. Instead, the processors are only concerned with a subset of block types (`image, video, etc.). To hopefully clarify: when `processContent` encounters a Group block, it does not match on that block, but if that block's inner contents contains a media-containing block, that will match and be processed further (via `processBlock`), while all content preceeding that match is treated as a raw string. This is a way to produce the desired output without the need to implement any serialization logic for all potential block types.

WordPress/src/main/java/org/wordpress/android/ui/posts/mediauploadcompletionprocessors/CoverBlockProcessor.java
mkevins: > Do you remember what those differences were? Any difference would have to be a bug

Off-hand, I recall that jsoup was adding closing tags, which was problematic during at least one of the iterations toward the current implementation. It's likely this is resolved now, as the implementation no longer processes units of non-self-contained HTML, so probably worth revisiting. To be sure, this was not a bug in jsoup, but rather a limitation in a particular iteration of this implementation, in that it required serialization of the parse tree, while jsoup is designed to serialize abstract trees.

> also it looks like this entire pattern shouldn't be necessary if we process the inner blocks using a spec-compliant parser

Indeed, using a spec-compliant parser would obviate the need for this, as we'd be able to handle inner blocks an a much more general way, albeit for _all_ block types.

WordPress/src/main/java/org/wordpress/android/ui/posts/mediauploadcompletionprocessors/CoverBlockProcessor.java
mkevins: > It's quite possible to have `background-image:url` inside the block in a way that isn't meant to indicate a different background, and this string-matching might end up paving a road of headaches to debug

Ah, indeed! I wonder if there is a misunderstanding about how this pattern is currently used. This only matches on the `style` attribute [within the parsed HTML of a matching block](https://github.com/wordpress-mobile/WordPress-Android/pull/11541/files/e6c78af481ba7d6d361d9ff526933494bac0ee32#diff-8c06e672b915270d7915261d8e7e367cR68), and so the string replacement is restricted to that particular inline style. In other words, it will _not_ replace similar text within the content of HTML (e.g. a caption, a paragraph, etc.). Perhaps the Javadoc could be made more specific.

WordPress/src/main/java/org/wordpress/android/ui/posts/mediauploadcompletionprocessors/CoverBlockProcessor.java
mkevins: > we probably need to update both places

The Json attributes are updated in an [earlier step](https://github.com/wordpress-mobile/WordPress-Android/pull/11541/files/e6c78af481ba7d6d361d9ff526933494bac0ee32#diff-5cc438e830db3e4f6b106dca92c585c8R91) via the abstract method `processBlockJsonAttributes`. For Cover block, the `id` is used to check for a match first, and the `url` [is mutated](https://github.com/wordpress-mobile/WordPress-Android/pull/11541/files/e6c78af481ba7d6d361d9ff526933494bac0ee32#diff-8c06e672b915270d7915261d8e7e367cR54) a match with the local id is found. The replacement within the `style` attribute only occurs subsequently, if that condition is met.

WordPress/src/test/java/org/wordpress/android/ui/posts/mediauploadcompletionprocessors/TestContent.kt
```diff
@@ -357,6 +360,92 @@ object TestContent {
...
+<div class="wp-block-cover has-background-dim" style="background-image:url($remoteImageUrl)">
+  <div class="wp-block-cover__inner-container">
+    <!-- wp:paragraph {"align":"center","placeholder":"Write title…"} -->
+    <p class="has-text-align-center"></p>
+    <!-- /wp:paragraph -->
+  </div>
+</div>
+<!-- /wp:cover -->
+"""
+    const val oldCoverBlockWithNestedImageBlock = """<!-- wp:cover {"url":"$localImageUrl","id":$localMediaId} -->
```
chipsnyder, line 386: Hey, @mkevins Just want to confirm this flow. This is this data supposed to represent a new cover block and a new image block within it or is it supposed to be an existing cover block with a new image block? If it is supposed to be new at both layers in practice these will have different IDs correct? 

Not a big deal but want to make sure I understand correctly for the way I approach it on iOS

pinarol:
Friendly reminder: We should merge into `gutenberg/after_1.26.0` instead of `develop` until Monday. After Monday we can target `develop` again.

mkevins:
Thanks for reviewing and testing :+1: .

PR #11552, opened 2020-03-28
Remove extra mocks files

jkmassel:
**To test:**
I think all of our UI tests are disabled right now anyway, so this may not matter. But these files shouldn't affect them. We could try re-enabling a test we know to work, or if this breaks something, we could fix it later?

Either way, these seem incorrectly located.

WDYT @rachelmcr?

**PR submission checklist:**


rachelmcr:
Oh, interesting. These are definitely in the wrong location.

However, both of these mocks relate to tests that we've had flaky results from (tests using magic links and editor tests). Before we remove these mocks entirely, it might be worth moving them to the correct location and seeing if they resolve that flakiness.

PR #11553, opened 2020-03-29
Handle user with no site

### LINKED ISSUE #11533: Add reblog functionality to Reader posts: Handle users that do not have a site
Master issue #10934

When the reblog action is triggered (either from the post cards in a stream view or from the actions in a reader post detail view) and the user has no site the site creation flow should be triggered. 

The user should be prompted to select either to cancel the action or to create a site and proceed with the reblog action.

The site creation flow would be the same with the one triggered from the site picker.

antonis:
Fixes #11533 

## Description

This addition handles the case when the user has no site to reblog to.

A new screen appears when the user triggers the reblog action and has no sites. The screen design is in sync with the iOS implementation and has an actionable button directing the user to the _My site_ screen where the _Add New Site_ action is available.

The new screen is show when the reblog action is triggered in the post details or in the post cards stream view.

| Post details action | Stream view action |
| --- | --- |
|detail|list|

### Reblog action on post cards actions in stream view

0. Open the App and Login with a user having no sites
1. Select the Reader tab
4. Notice the no sites available screen
5. Press the _Manage Sites_ button
6. Verify that you are directed to the _My Sites_ page
7. Notice the _Add New Site_ button

### Reblog action on post details actions

0. Open the App and Login with a user having no sites
1. Select the Reader tab
4. Notice the no sites available screen
5. Press the _Manage Sites_ button
6. Verify that you are directed to the _My Sites_ page
7. Notice the _Add New Site_ button

### Note

~~The **site creation flow** after redirection to the _My Sites_ page **has not been tested**. The default site creation flow fails with an HTTP 403/Forbidden error:
`Unexpected response code 403 for https://public-api.wordpress.com/rest/v1.1/sites/new/?locale=en_US`
This is error is probably related with the restrictions on credentials created with [applications manager](https://developer.wordpress.com/apps/)~~.
The site creation flow after redirection to the _My Sites_ page has been tested with credentials authorised to perform the site creation action.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/res/layout/reader_no_site_to_reblog.xml
+++ b/WordPress/src/main/res/layout/reader_no_site_to_reblog.xml
@@ -5,7 +5,7 @@
     android:layout_height="match_parent">
 
     <org.wordpress.android.ui.ActionableEmptyView
-        android:id="@+id/no_site_to_reblog_view"
+        android:id="@+id/noSiteToReblogView"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
         android:layout_marginTop="@dimen/toolbar_height"
--- a/WordPress/src/main/res/layout/reader_no_site_to_reblog.xml
+++ b/WordPress/src/main/res/layout/reader_no_site_to_reblog.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <org.wordpress.android.ui.ActionableEmptyView
+        android:id="@+id/no_site_to_reblog_view"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:layout_marginTop="@dimen/toolbar_height"
+        app:aevButton="@string/reader_no_site_to_reblog_action"
+        app:aevImage="@drawable/img_illustration_empty_results_216dp"
+        app:aevSubtitle="@string/reader_no_site_to_reblog_detail"
+        app:aevTitle="@string/reader_no_site_to_reblog_title" />
+
+</RelativeLayout>
--- a/WordPress/src/main/res/layout/reader_no_site_to_reblog.xml
+++ b/WordPress/src/main/res/layout/reader_no_site_to_reblog.xml
@@ -11,6 +11,7 @@
         android:layout_marginTop="@dimen/toolbar_height"
         app:aevButton="@string/reader_no_site_to_reblog_action"
         app:aevImage="@drawable/img_illustration_empty_results_216dp"
+        app:aevImageHiddenInLandscape="true"
         app:aevSubtitle="@string/reader_no_site_to_reblog_detail"
         app:aevTitle="@string/reader_no_site_to_reblog_title" />
 
--- a/WordPress/src/main/AndroidManifest.xml
+++ b/WordPress/src/main/AndroidManifest.xml
@@ -451,7 +451,7 @@
             android:theme="@style/CalypsoTheme"
             android:windowSoftInputMode="adjustResize|stateHidden" />
         <activity
-            android:name=".ui.reader.ReaderPostNoSiteToReblogActivity"
+            android:name=".ui.reader.NoSiteToReblogActivity"
             android:theme="@style/CalypsoTheme" />
         <activity
             android:name=".ui.AppLogViewerActivity"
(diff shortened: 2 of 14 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostNoSiteToReblogActivity.kt
```diff
@@ -0,0 +1,36 @@
...
+/*
+ * Serves as an intermediate screen where the user is informed that a site is needed for the reblog action
+ */
+class ReaderPostNoSiteToReblogActivity : AppCompatActivity() {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.reader_no_site_to_reblog)
+
+        supportActionBar?.setDisplayShowTitleEnabled(false)
+        supportActionBar?.setDisplayHomeAsUpEnabled(true)
```
develric, line 20: I think we can maybe make it more synthetic using `supportActionBar?.apply {...}`; wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostNoSiteToReblogActivity.kt
```diff
@@ -0,0 +1,36 @@
...
+
+        no_site_to_reblog_view.button.setOnClickListener {
```
develric, line 22: Maybe just a personal preference here: I usually prefer the old plain `findViewById` to avoid using xml `_` notation in naming; but where we use it in the codebase I think we usually let the kotlin notation to win (also in the xml) like `noSiteToReblogView`

WordPress/src/main/res/layout/reader_no_site_to_reblog.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <org.wordpress.android.ui.ActionableEmptyView
```
develric, line 7: It was not shared before so you could not know but I'm reporting since I noticed while testing this PR: I think we should manage the image of this `ActionableEmptyView` to hide it in landscape so to avoid the action button to not be visible. See the notifications case below:

| Screenshot_1586276298 | Screenshot_1586276303 |
|---|---|

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostNoSiteToReblogActivity.kt
```diff
@@ -0,0 +1,36 @@
...
+import android.view.MenuItem
+import androidx.appcompat.app.AppCompatActivity
+import kotlinx.android.synthetic.main.reader_no_site_to_reblog.*
+import org.greenrobot.eventbus.EventBus
+import org.wordpress.android.R
+
+/*
+ * Serves as an intermediate screen where the user is informed that a site is needed for the reblog action
+ */
+class ReaderPostNoSiteToReblogActivity : AppCompatActivity() {
```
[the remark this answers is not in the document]
antonis: Fixed with 8eb14af7c180569716f5aaf8e86f6b55f5ca78e5

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostNoSiteToReblogActivity.kt
antonis: Fixed with 8eb14af7c180569716f5aaf8e86f6b55f5ca78e5

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostNoSiteToReblogActivity.kt
antonis: Fixed with 8eb14af7c180569716f5aaf8e86f6b55f5ca78e5

WordPress/src/main/res/layout/reader_no_site_to_reblog.xml
antonis: I tried a more generic approach on this since it seems to be an issue in other screens too.
Fix: cab7a8d4c0ca69c4d3ebe9f06406a28ee7dc9f59
I tested that this can also work in cases like the NotificationsListFragmentPage. It fails though in other pages (e.g. MySitesFragment) probably due to the way configuration change is handled. Wdyt?
I can revert this and fix it in NoSiteToReblogActivity only.

develric: I agree with your solution here 👍! In the My Site case currently we are conditioning based on the screen height being >= 600 that makes sense to me. I think we need some more digging here to generalize the cases but we can eventually improve the logic by which we hide the image in a dedicated task! For now I'm accepting the modification since it can be generalized and currently has an effect only in the screen we are covering in this PR 👍.

PR #11555, opened 2020-03-29
Customise Site Picker for Reblog Action

### LINKED ISSUE #11554: Add reblog functionality to Reader posts: Customise Site Picker for Reblog Action
Master issue #10934

When the reblog action is triggered (either from the post cards in a stream view or from the actions in a reader post detail view) and the user has more than one sites the default Site Picker is triggered.

In that case the unneeded functions should be removed and leave only the back button and the search.

antonis:
Fixes #11554

## Description

Displays only the Search and Back action on the top toolbar when site picker is opened for reblog

The custom sitepicker is show when the reblog action is triggered in the post details or in the post cards stream view.

| Post details action | Stream view action |
| --- | --- |
|detail|list|

### Reblog action on post cards actions in stream view

0. Open the App and Login with a user that has more than one sites
1. Select the Reader tab
4. Verify that the site picker opens
5. Verify that the site picker top toolbar has only the search action
6. Long tap on any list item and verify that no action is triggered

### Reblog action on post details actions

0. Open the App and Login with a user that has more than one sites
1. Select the Reader tab
4. Verify that the site picker opens
5. Verify that the site picker top toolbar has only the search action
6. Long tap on any list item and verify that no action is triggered

### Default site picker without customization

0. Open the App and Login with a user that has more than one sites
1. Press the _Switch Site_ button on the My Sites page
2. Verify that all the actions are displayed in the top toolbar: Search, Add and Edit
3. Long tap on any list item and verify that the list enters edit mode

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
@@ -43,6 +43,7 @@ import org.wordpress.android.ui.history.HistoryDetailContainerFragment;
 import org.wordpress.android.ui.history.HistoryListItem.Revision;
 import org.wordpress.android.ui.main.MeActivity;
 import org.wordpress.android.ui.main.SitePickerActivity;
+import org.wordpress.android.ui.main.SitePickerActivity.SitePickerMode;
 import org.wordpress.android.ui.main.WPMainActivity;
 import org.wordpress.android.ui.media.MediaBrowserActivity;
 import org.wordpress.android.ui.media.MediaBrowserType;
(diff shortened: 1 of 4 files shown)


WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
```diff
@@ -126,31 +126,34 @@ public static void showMainActivityAndSignupEpilogue(Activity activity, String n
      * @param site     the preselected site
      */
     public static void showSitePickerForResult(Activity activity, SiteModel site) {
-        Intent intent = createSitePickerIntent(activity, site);
+        Intent intent = createSitePickerIntent(activity, site, false);
```
develric, line 129: Even if we have currently only a 2 state discrimination here I would maybe use an enum (like `SitePickerMode`). Wdyt?

antonis: You are right Riccardo, this will make the code more clear. I fixed it with 54adab18f52343bd8a095ad6b35565d2ed5976ef

antonis:
Tank you @develric. I can investigate further the site picker issue

PR #11556, opened 2020-03-30
Feature/support multiple auth headers in gb

khaykov:
This PR addresses changes in GB side to allow multiple headers to be passed to editor's `OkHttpHeaderInterceptor.

GB sitde changes are [here](https://github.com/wordpress-mobile/gutenberg-mobile/pull/2080).

To test:
- Using a private site, navigate to post that contains an image using GB or Aztec.
- Notice that the image is correctly loaded.

PR submission checklist:


libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/AztecEditorFragment.java
```diff
@@ -147,6 +147,8 @@ public AztecLoggingException(Throwable originalException) {
     private static final String TEMP_VIDEO_UPLOADING_CLASS = "data-temp-aztec-video";
     private static final String GUTENBERG_BLOCK_START = "<!-- wp:";
 
+    private static final String AUTHORIZATION_HEADER_NAME = "Authorization";
```
mchowning, line 150: Would it make sense to reuse the `AUTHORIZATION_HEADER_NAME` from `AuthenticationUtils` here instead of declaring a new variable?

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -2663,14 +2664,15 @@ public void onVideoPressInfoRequested(final String videoId) {
...
-        String authHeader = "";
+    public Map<String, String> onAuthHeaderRequested(String url) {
+        Map<String, String> authHeaders = new HashMap<>();
+
         String token = mAccountStore.getAccessToken();
         if (mSite.isPrivate() && WPUrlUtils.safeToAddWordPressComAuthToken(url)
             && !TextUtils.isEmpty(token)) {
-            authHeader = "Bearer " + token;
+            authHeaders.put(AuthenticationUtils.AUTHORIZATION_HEADER_NAME, "Bearer " + token);
         }
```
mchowning, line 2674: In `AuthenticationUtils` when `safeToAddWordPressComAuthToken` is `false` we try to use HTTP auth credentials, and it's not clear to me why we don't do that here. Similarly, we're checking `mSite.isPrivate()` here, but we don't check that in the `AuthenticationUtils`. 

I know this isn't anything you changed, I'm just curious if you know why there's the difference. If you do know why there is the difference, it might even be useful to move this logic into a method in `AuthenticationUtils` with some explanation about why the two methods are different. This is not anything that needs to block this PR.

mchowning:
I think this looks good @khaykov . 👍 I just had a couple of small comments, and they may not even require any changes. It's not clear to me why we're making this change though, would you mind leaving a comment just explaining the motivation behind why we're making this change?

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/AztecEditorFragment.java
khaykov: Yes, but `AztecEditorFragment` is part of the Editor submodule, so It has no access to `AuthenticationUtils`. This is not super elegant, but my understanding is that we slowly but surely switch towards block editor, so I decided not to add any new logic related to it. But maybe there is some simple way to fix it, that I'm missing?

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
khaykov: I'm not 100% sure about the difference there, but as far as I'm aware, `AuthenticationUtils` is mostly used with image loading, and requests that actually get to it are subject to `isPrivate()` check since it's related to Photon/https (I have only surface level of knowledge there).

I'm not sure why we are not using HTTP auth credentials in the block editor, but since it's not broken I would rather not touch it :)

khaykov:
@mchowning Thanks for taking a look at this one! I'm doing it so we can add an extra header for the upcoming private Atomic sites support. A cookie is used there to retrieve private media, but in an unorthodox way (same as a token, without using Set-Cookie header, so we can't use a cookie jar).

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
mchowning: > I'm not sure why we are not using HTTP auth credentials in the block editor, but since it's not broken I would rather not touch it :)

I think that makes sense. Obviously it would be nice to figure out the cause of the inconsistency, but I don't think it's something we have to do now.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/AztecEditorFragment.java
mchowning: > Yes, but `AztecEditorFragment` is part of the Editor submodule, so It has no access to `AuthenticationUtils`. 

Right, I forgot about that minor detail. 🤦‍♂

mchowning:
Just a reminder that we need to update the gutenberg-mobile submodule ref now that the GB-side changes have been merged.

khaykov:
Updated the GB reference 👍

mchowning:
Double checked with the update, and everything is working great!

PR #11573, opened 2020-04-01
Issue/11152 add horizontal recycler view on preview screen

### LINKED ISSUE #11152: Show all images in a horizontal recycler view on the Hub
Goal of this task is to add support for showing thumbnails of multiple local images on the Hub.

- [x] Introduce horizontal RecyclerView below the ImageView and load thumbnails of all the  images
- [x] Make the items clickable and load the selected image into the main ImageView

It should look similarly to the following design. The "Edit, Insert, Cancel" actions will be added as part of a different task. Goal of this task is to only show all the images in a recycler view.
Screenshot 2020-01-23 at 18 50 54

ashiagr:
Partially fixes #11152

This PR just

- introduces horizontal `RecyclerView` below the  preview `ImageView`
- attempts to load thumbnails from a dummy data list using [ListAdapter](https://developer.android.com/reference/kotlin/androidx/recyclerview/widget/ListAdapter) so that we don’t have to fully implement [DiffUtil](https://developer.android.com/reference/android/support/v7/util/DiffUtil.html) or call [notifyDataSetChanged](https://developer.android.com/reference/android/support/v7/widget/RecyclerView.Adapter.html#notifyDataSetChanged()):

    > If you are writing an adapter it will always be more efficient to use the more specific change events if you can. Rely on notifyDataSetChanged() as a last resort.

thumbnails

P.S.
- Dummy data will be replaced with actual data from custom image picker later (along with changes to any method signatures and any logic that depends on number of images) 
- Temporarily commented: navigate to next screen 
- Click action to be fully added in another PR (there's a commented `onThumbnailClick` method just to see `ListAdapter` refresh animations)
- Trying to use one common `UiState` in another PR #11574. Not very sure if this is how we should do it.

PR submission checklist:


ashiagr:
Closing in favor of #11608

PR #11585, opened 2020-04-03
Merge gutenberg strings in the main strings.xml file

maxme:
Merge gutenberg strings in the main `strings.xml` file.


maxme:
The script generating the file but be modified, there is a lint error coming from the string file:
```
/home/circleci/project/WordPress/src/main/res/values/strings.xml:2682: Error: Replace "-" with an "en dash" character (–, &#8211;) ? [TypographyDashes]
    <string name="gutenberg_native_555_555_1234" tools:ignore="UnusedResources">(555)555-1234</string>
```
I'll update the file manually here and open another PR to fix the script.

maxme:
Please hold on reviewing that one, it seems there are some extra strings in the generated file, that might come from a missing `gutenberg.pot` file which should be excluded. I'll investigate and fix or comment here.

maxme:
The patch looks better after 878eada

Note after a coulpe of problems here, I opened:
- https://github.com/wordpress-mobile/gutenberg-mobile/pull/2107
- https://github.com/wordpress-mobile/gutenberg-mobile/pull/2109

maxme:
BTW, I guess the next step is to merge `release/14.5` to `master` so the new strings get picked up by the GlotPress script. cc @loremattei

WordPress/src/main/res/values/strings.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version='1.0' encoding='utf-8'?>
```
Tug, line 1: So far I've tried to manually revert this change each time. It's annoying but I haven't found a way to fix it yet

maxme: Is this important? Or just annoying because it's different from the default android format for  xml files?

Tug: Not important afaik

PR #11590, opened 2020-04-06
Feature/support private at sites

khaykov:
This PR adds support for loading media from private Atomic sites.

Companion [Flux-C](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1531) PR needs to be merged first.

There are two main changes:
- If the site is private atomic, we load [Photon](https://developer.wordpress.com/docs/photon/) media using different base URL (media proxy). Most of the changes are related to this method.
- We use a cookie to load media in Editor, Reader post details and site/post preview.

Since the use case for this is relatively small, I wanted to avoid adding too much new logic, and instead integrated the behavior into existing flow with (hopefully) minimal amount of interruption.

There are a couple of steps to test the changes:
- Using a8c account and wpcalypso.wordpress.com (important) create a site with a business plan, perform AT by installing a plugin and set the site to private.
- Using App, switch to the abovementioned site.

Site preview:
- From MySite screen, preview the site on the web by tapping on the site name, or View Site button. You should be able to view the site.

Editor:
- Create a new post, set a featured image to it and add an image to the post content. Publish the post.
- View post from post list. It should be displayed correctly.
- Open post in Block editor. The image in the post should be displayed correctly.
- Open post in Aztec. The image in the post should be displayed correctly.

Reader:
- Navigate to the reader, and select Following tab.
- From subfilter select the private AT site we are testing.
- Notice that the featured image in the post is displayed correctly in the reader post list.
- Open the post details. Notice that the image in the post is displayed correctly.

When testing cookie authentication, you might encounter cookie loading with a modal dialog. 
Cookie loading injects into normal content loading flow and interrupts it if necessary. Loading will continue as normal after the cookie is retrieved. A cookie Progress dialog is cancellable, and when canceled the normal flow of content render will resume. 

PR submission checklist:


develric:
Hi @khaykov , was looking into this PR and was able to get the described scenarios to work correctly 👍; while looking into it I noticed the following and I thought to leave this quick note for you to double check.

In the Posts list, if I ask for a VIEW on a published post or for a PREVIEW on a draft post I get something like the following:

<p align="center">
  
</p>

Let me know if I can help in reproducing in case it is setup specific 😊

khaykov:
Thank you, @develric ! A mistake on my part - I assumed Post preview is launched in the same way as site preview, but it's not. Added a private site id to post preview launcher, so it should work.

Also fixed issue where site preview was always behaving thinking that the site is private.

khaykov:
There is one hard to tackle issue - when you play with calypso settings while the app is open, the app can (and will) get out of sync with calypso. This affects all parts of our app, and I don't think there is much we can do at this point.

PR #11598, opened 2020-04-07
Using Gravatar in me menu.

### LINKED ISSUE #10511: Gravatar vs Icon in the Me Section menu.
As part of the **Information Architecture project** we are moving the access of the Me section from Bottom Navigation bar to a menu in My Site section (see  #10495).

A dedicated PR has been created for this (see #10498). Note that in this PR we used the "ic_user_circle_white_24dp" icon for the My site section menu.

This issue has been created as a reminder to track eventual further discussion about the convenience or not to use the user Gravatar instead of the "ic_user_circle_white_24dp" icon in the menu.

cc @planarvoid and @osullivanchris .

develric:
Fixes #10511

This PR adds the Gravatar to the Me menu.

**NOTE 1:** I opened a related issue #11599 documenting an accessibility issue I noticed while working on this; I think we can manage it in a dedicated task PR.

**Note 2:** working on this PR I saw some alignment issue we have in various places where we use/show the gravatar; when it is changed in the app not all places get updated if they were previously accessed (see for example the My Site > Comments section if you have commented on your site and visualized the comment with Gravatar before to change it). AFAIU and if I'm not missing something obvious (in that case there is not actual issue 😜) it is due to caching behaviours (at Glide level and at API side level; also tried to use the `signature` to invalidate the glide cache but many times one of the Glide cache level seems to be used anyway [see [here](https://bumptech.github.io/glide/doc/caching.html) for more glide caching details]). 
For now I tried to align the two gravatar behavior in Me section and in Me menu (also not sure it is only something limited to gravatar). 

**NOTE 3:** I think currently we are not robust in catching an update in the Gravatar from the web. So for example if the app is opened and already showed a section with the gravatar (so it is cached) an update of the gravatar is not triggered (at least this is what I got). If I'm not missing something and reviewer confirms the above, probably make sense to face all this wider task in dedicated PRs effort. Using something like the `MeGravatarLoader` maybe can help but does not manage the sync with a web Gravatar change (I needed to clean the cache to get it updated and synched along the app).

### WP.com and JP site
- check you get the `not set` or your set gravatar in the Me menu and Me section

| Not set Gravatar | Set Gravatar |
|---|---|
| image | image |

- Change/set the gravatar and check you get the correct updated one in both Me menu and Me section

### Self-hosted site added by address
- Check you get the generic user icon like below in the menu

<p align="center">
  
</p>

### Smoke test
Smoke test the menu functionality to access the Me Section.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/MeFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/MeFragment.java
@@ -48,6 +48,7 @@ import org.wordpress.android.networking.GravatarApi;
 import org.wordpress.android.ui.ActivityLauncher;
 import org.wordpress.android.ui.RequestCodes;
 import org.wordpress.android.ui.accounts.HelpActivity.Origin;
+import org.wordpress.android.ui.main.utils.MeGravatarLoader;
 import org.wordpress.android.ui.media.MediaBrowserType;
 import org.wordpress.android.ui.photopicker.PhotoPickerActivity;
 import org.wordpress.android.ui.photopicker.PhotoPickerActivity.PhotoPickerMediaSource;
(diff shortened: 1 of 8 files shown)


develric:
To ease a bit the design review adding this image below hoping it helps (each grid square is 8dp wide)

<p align="center">
  
</p>

PR #3754, opened 2016-02-15
Revisiting Learn More preference

tonyr59h:
Addresses #3678, #3677, #3638, #3705.

cc @mattmiklic

Edit: Much of the work was actually done on the web page itself. Content was added [here](https://en.support.wordpress.com/settings/discussion-settings/#default-article-settings) that the mobile client uses for the dialog.

Issue #3678, opened 2016-01-30
Site Settings: "Learn More" popup can follow links in browser to odd pages


v18:
The "Learn More" pop up displays the links present in the webpage as usual, and therefore it's possible to go to unintuitive places from there. Maybe it would be better to open the support page using a new browser tab instead of as a pop up?

Here are some screenshots to illustrate:

![odd-pages-1](https://cloud.githubusercontent.com/assets/11487924/12691898/52123d42-c6a5-11e5-9847-c066b943bc58.png)
![odd-pages-2](https://cloud.githubusercontent.com/assets/11487924/12691899/52147044-c6a5-11e5-875a-af7e502710b9.png)
![odd-pages-3](https://cloud.githubusercontent.com/assets/11487924/12691901/52157a70-c6a5-11e5-9bdd-b86fe37a47c0.png)
![odd-pages-4](https://cloud.githubusercontent.com/assets/11487924/12691900/521539ca-c6a5-11e5-9803-1e86771ccb9a.png)

This is related to #3638

tonyr59h:
> Maybe it would be better to open the support page using a new browser tab instead of as a pop up?

Other parts of the app will generally open a WebView in-app (rather than in the browser) so the UX isn't split between two apps. My first thought would be to intercept link clicks and ignore them. @mattmiklic any thoughts?

tonyr59h:
Addressed in #3754.

PR #11600, opened 2020-04-07
Commenting out the override method shouldOverrideUrlLoading() fixes the issue

gibtang:
Fixes #
Fixed issue where the URL for the popup does not load due to shouldOverrideUrlLoading()

To test:
In the WordPress app, go to My Sites > Settings > More > Learn More > Website loads correctly

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/prefs/LearnMorePreference.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/prefs/LearnMorePreference.java
@@ -213,11 +213,11 @@ public class LearnMorePreference extends Preference implements View.OnClickListe
     }
 
     private class LearnMoreClient extends WebViewClient {
-        @Override
-        public boolean shouldOverrideUrlLoading(WebView webView, String url) {
-            // prevent loading clicked links
-            return !mUrl.equals(url) && !SUPPORT_CONTENT_JS.equals(url) && !CONTENT_PADDING_JS.equals(url);
-        }
+//        @Override
+//        public boolean shouldOverrideUrlLoading(WebView webView, String url) {
+//            // prevent loading clicked links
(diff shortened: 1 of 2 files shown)


WordPress/src/main/java/org/wordpress/android/ui/prefs/LearnMorePreference.java
```diff
@@ -213,11 +213,11 @@ public SavedState createFromParcel(Parcel in) {
...
-        @Override
-        public boolean shouldOverrideUrlLoading(WebView webView, String url) {
-            // prevent loading clicked links
-            return !mUrl.equals(url) && !SUPPORT_CONTENT_JS.equals(url) && !CONTENT_PADDING_JS.equals(url);
-        }
+//        @Override
+//        public boolean shouldOverrideUrlLoading(WebView webView, String url) {
+//            // prevent loading clicked links
+//            return !mUrl.equals(url) && !SUPPORT_CONTENT_JS.equals(url) && !CONTENT_PADDING_JS.equals(url);
+//        }
```
aerych, line 220: A couple of thoughts.  

We prefer to just remove code that is no longer needed rather than comment it out. It keeps things nice and tidy that way.

Removing this overridden method may have some side effects we probably don't want. If you check the git blame and history you can see where the [code was added](https://github.com/wordpress-mobile/WordPress-Android/pull/3754) and the [issue it was trying to solve](https://github.com/wordpress-mobile/WordPress-Android/issues/3678). 

 Maybe do a little more investigation to see how the issue could be addressed without removing this method.

gibtang: Done. It was a URL redirect of https://en.support.wordpress.com/settings/discussion-settings/#default-article-settings to https://wordpress.com/support/settings/discussion-settings/#default-article-setting that will trigger the calling of the method. So I changed the URL in site_settings.xml to https://wordpress.com/support/settings/discussion-settings/#default-article-settings instead and the issue is fixed

PR #11601, opened 2020-04-07
Mark some string references untranslatable

maxme:
We missed to mark some string references untranslatable. Translators were getting asked to translate them in GlotPress.

### DIFF
--- a/WordPress/src/main/res/values/strings.xml
+++ b/WordPress/src/main/res/values/strings.xml
@@ -827,7 +827,7 @@
     <string name="app_theme_entry_value_dark" translatable="false">dark</string>
     <string name="app_theme_entry_value_default" translatable="false">default</string>
     <string name="app_theme_entry_value_default_key" translatable="false">@string/app_theme_entry_value_light</string>
-    <string-array name="app_theme_entries" tools:ignore="InconsistentArrays">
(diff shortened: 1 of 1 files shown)


khaykov:
Hey @maxme 👋 Regarding `app_theme_entries`, it contains translatable strings in it it as a reference. I assume this doesn't work well with GlotPress? If so, there is another version of this array in [v28](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/res/values-v28/strings.xml).

maxme:
Hi @khaykov - we should mark all the references as untranslatable, but the referenced strings themselves should be translatable. In that case, the string array is using references only, it should be marked untranslatable (we don't want the references to land as a string to be translated) but the referenced strings should be translated (and obviously not marked untranslatable).

This looks good to me (and this what we're doing in other places with other `string-array` that are displayed to the users).

```
<string name="app_theme_light">Light</string>
<string name="app_theme_dark">Dark</string>

[...]

<string-array name="app_theme_entries" translatable="false" tools:ignore="InconsistentArrays">
  <item>@string/app_theme_light</item>
  <item>@string/app_theme_dark</item>
</string-array>
```

Everything in `v28/strings.xml` should also be marked untranslatable, but since it's not imported in GlotPress, there is no impact.

PR #11608, opened 2020-04-08
Support multiple images on preview screen

### LINKED ISSUE #11152: Show all images in a horizontal recycler view on the Hub

ashiagr:
Partially fixes #11152

We decided to use `ViewPager2` (for swipeable ImageViews) and `TabLayout` + `TabLayoutMediator` (for thumbnails) instead of having a static `ImageView` and displaying thumbnails in a `RecyclerView`.

This PR just
- loads images into `ViewPager2` from a dummy data list using [ListAdapter](https://developer.android.com/reference/kotlin/androidx/recyclerview/widget/ListAdapter) so that we don’t have to fully implement [DiffUtil](https://developer.android.com/reference/android/support/v7/util/DiffUtil.html) or call [notifyDataSetChanged](https://developer.android.com/reference/android/support/v7/widget/RecyclerView.Adapter.html#notifyDataSetChanged()). 
- observes changes to the above adapter inside `PagerAdapterObserver` and loads thumbnails using `TabLayoutMediator`.
- refactors code to use existing preview image states we used for single image editing

(This PR only supports image load success states. Load failed states are being supported in [this branch](https://github.com/wordpress-mobile/WordPress-Android/tree/issue/viewpager2-image-load-failed) because of some image display issues, can create separate PR or merge in this one later. )

viewpager

Pending action items related to this PR:

- [ ] Dummy data - replace with actual data from custom image picker later (along with changes to any method signatures and any logic that depends on number of images).
- [ ] Thumbnails - manage select index on orientation change.
- [ ] Thumbnails - do we need to manage thumbnail image failed states?
- [ ] Existing test cases - update once changes are finalised.
- [ ] Temporarily commented: navigate to next screen, file saving logic. Handle in upcoming PRs when we add done button to preview screen.

PR submission checklist:


libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
```diff
@@ -36,22 +39,57 @@ class PreviewImageFragment : Fragment() {
...
+
+        val tabConfigurationStrategy = TabLayoutMediator.TabConfigurationStrategy { tab, position ->
+            if (tab.customView == null) {
+                val customView = LayoutInflater.from(context).inflate(layout.preview_image_thumbnail, null)
+                tab.customView = customView
+            }
+            val imageView = (tab.customView as FrameLayout).findViewById<ImageView>(R.id.thumbnailImageView)
+            val pagerAdapter = previewImageViewPager.adapter as PreviewImageAdapter
+            val imageData = pagerAdapter.currentList[position].data
+            loadIntoImageView(imageData.lowResImageUrl, imageView)
```
ashiagr, line 72: I'm thinking to pass selected position to the VM and let VM pass image data to the fragment to load the thumbnail. This selected position can also be used to manage selected position on orientation change.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
```diff
@@ -36,22 +39,57 @@ class PreviewImageFragment : Fragment() {
...
+        val previewImageAdapter = PreviewImageAdapter(
+            loadIntoImageViewWithResultListener = { imageData, imageView, position ->
+                loadIntoImageViewWithResultListener(imageData, imageView, position)
+            }
+        )
+        previewImageViewPager.adapter = previewImageAdapter
+
+        val tabConfigurationStrategy = TabLayoutMediator.TabConfigurationStrategy { tab, position ->
+            if (tab.customView == null) {
+                val customView = LayoutInflater.from(context).inflate(layout.preview_image_thumbnail, null)
```
malinajirka, line 66: Do we really want to pass null as root? Have you considered passing `root = tablayout` and `attachToParent = false`?

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
```diff
@@ -36,22 +39,57 @@ class PreviewImageFragment : Fragment() {
...
+                tab.customView = customView
+            }
+            val imageView = (tab.customView as FrameLayout).findViewById<ImageView>(R.id.thumbnailImageView)
+            val pagerAdapter = previewImageViewPager.adapter as PreviewImageAdapter
```
malinajirka, line 70: Nitpick: Could we use `previewImageAdapter` and remove this line?

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
malinajirka: >This selected position inside VM can also be used to select the right thumbnail on orientation change.

Re-selection of the correct tab should be handled by the system. 

> let VM pass image data to the fragment to load the thumbnail

What do you mean by that? Thumbnail loading already works, right?

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
```diff
@@ -119,10 +157,17 @@ class PreviewImageFragment : Fragment() {
...
+//        val inputFilePath = fileInfo.first
+//        val outputFileExtension = fileInfo.second
+//        findNavController().navigate(
+//            PreviewImageFragmentDirections.actionPreviewFragmentToCropFragment(inputFilePath, outputFileExtension)
+//        )
+    }
+
+    override fun onDestroyView() {
+        super.onDestroyView()
+        previewImageViewPager.adapter?.unregisterAdapterDataObserver(pagerAdapterObserver)
```
malinajirka, line 170: I'm just wondering if we can set these fields simply to `null` -> they are no longer valid? AFAIK gc will be able to collect both the fields and the views. Wdyt?

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewHolder.kt
```diff
@@ -0,0 +1,42 @@
...
+    private val errorLayout = itemView.findViewById<ConstraintLayout>(R.id.errorLayout)
+
+    fun onBind(uiState: ImageUiState) {
+        loadIntoImageViewWithResultListener.invoke(uiState.data, previewImageView, adapterPosition)
+        UiHelpers.updateVisibility(progressBar, uiState.progressBarVisible)
+        UiHelpers.updateVisibility(errorLayout, uiState.retryLayoutVisible)
+    }
+
+    companion object {
+        fun create(
```
malinajirka, line 30: Have you considered using [this approach](https://github.com/wordpress-mobile/WordPress-Android/blob/2b681ae6d87f837c2eab29a99a428ce2a7a875fc/WordPress/src/main/java/org/wordpress/android/ui/main/ActionListItemViewHolder.kt#L18) instead of the companion object? Is there any advantage?

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageAdapter.kt
```diff
@@ -0,0 +1,41 @@
...
+        (holder as PreviewImageViewHolder).onBind(uiState)
+    }
+}
+
+private class PreviewImageDiffCallback : DiffUtil.ItemCallback<ImageUiState>() {
+    override fun areItemsTheSame(oldItem: ImageUiState, newItem: ImageUiState): Boolean {
+        return oldItem.data.lowResImageUrl == newItem.data.lowResImageUrl
+    }
+
+    override fun areContentsTheSame(oldItem: ImageUiState, newItem: ImageUiState): Boolean {
```
malinajirka, line 35: I'm wondering if this implementation is what we actually want. AFAIK we want to return `false` when we need to re-render the UI right?

For example I believe we don't need to re-render anything for the following states, right? 
ImageDataStartLoadingUiState
ImageInLowResLoadSuccessUiState
ImageInLowResLoadFailedUiState

SideNote: It raises a question if it make sense to have 3 uiState for something which doesn't result in a change on the UI.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewHolder.kt
ashiagr: I was used to companion object approach from [Android Architecture Blueprints sample](https://github.com/android/architecture-samples/blob/master/app/src/main/java/com/example/android/architecture/blueprints/todoapp/tasks/TasksAdapter.kt#L53-L60) and it just creeped in. There was no added advantage so used suggested approach here: bf8f911.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
ashiagr: Yes we can, replaced and removed here: 342099b. Thanks :).

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
ashiagr: Inflated layout’s root element wrapped a fixed size `ImageView`, so even if root element's layout params were ignored, the view would render fine. But I agree, we should avoid passing null. Fixed here:  83af0d7

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
ashiagr: `previewImageViewPager.adapter` is set to null inside `TabLayoutMediator` `detach()` called in the following line. I've also set `pagerAdapterObserver` to `null` in ea85194. Hope this is what you meant by 

> these fields

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PagerAdapterObserver.kt
```diff
@@ -0,0 +1,63 @@
+package org.wordpress.android.imageeditor.preview
+
+import androidx.recyclerview.widget.RecyclerView
+import androidx.viewpager2.widget.ViewPager2
+import com.google.android.material.tabs.TabLayout
+import com.google.android.material.tabs.TabLayoutMediator.TabConfigurationStrategy
+
+/**
```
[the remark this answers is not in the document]
ashiagr: > I'm new to the ViewPager2. Moreover, I haven't used any ViewPager with a Tablayout for a long time. I'm sorry if the questions don't make sense.

I'm new to `ViewPager2` as well :D. Please don't feel sorry, all thoughts are welcome that can improve the implementation 🙌. 

I think flickering was because of all tabs getting removed and then re-added in the default adapter observer whenever adapter refreshed. I tried to keep the tabs and just update the url to bring a little improvement. It can be further improved, but I'm not sure if we should do it now or later after we've finished higher priority tasks. Utilising `onItemRangeChanged` and `DiffUtil` seems like a promising solution. Thanks for sharing your thoughts 🙏.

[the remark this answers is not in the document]
malinajirka: Thanks for the answer! ;) I agree that we can get back to optimizations later.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
ashiagr: I think it was premature to comment on it. Please ignore it. 

> Re-selection of the correct tab should be handled by the system.

Currently last selected tab doesn't get reselected after orientation change. It is in my pending action items list, might need some changes in the VM.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
malinajirka: Sorry, my comment wasn't clear. What I meant is if we can set both the fields to null instead of calling "deatch" and "unregisterAdapterDataObserver". It seems weird that we'd need to manually unregister from a UI component - or have you find this information somewhere in the docs? Wdyt?

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageAdapter.kt
malinajirka: I'd consider fixing this. Even though it's just an optimization, it feels like a bug.
Perhaps, something like this?
`
return oldItem.data != newItem.data
           || oldItem.progressBarVisible != newItem.progressBarVisible
           || oldItem.retryLayoutVisible != newItem.retryLayoutVisible;
`

wdyt?

ashiagr: >  AFAIK we want to return false when we need to re-render the UI right?

Correct

> For example I believe we don't need to re-render anything for changes between the following states, right?
ImageDataStartLoadingUiState
ImageInLowResLoadSuccessUiState
ImageInLowResLoadFailedUiState

> SideNote: It raises a question if it make sense to have 3 uiState for something which doesn't result in a change on the UI.

I'll think more on it. I just tried to refactor and fit in existing logic and states for single image to multiple images in the first iteration. There might still be a case when we will have just one image in the list and all states might be meaningful, however I haven't thought much about it yet.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
ashiagr: So I checked `TabLayoutMediator` source code and observed inside `detach()` it [unregisters the adapter observer](https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/tabs/TabLayoutMediator.java#L150-L153) if `autoRefresh` is enabled. Since we haven't enabled `autoRefresh`, I considered unregistering it explicitly in an appropriate lifecycle method:

`onViewCreated -> registerAdapterDataObserver`
`onDestroyView -> unregisterAdapterDataObserver`

I guess there's no harm in doing so?

malinajirka: Ok, lets keep it. It makes the code more complicated, but I guess it's better to play it safe in this case. Thanks for looking into it.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageAdapter.kt
malinajirka: >I'll think more on it. I just tried to refactor and fit in existing logic and states for single image to multiple images in the first iteration. There might still be a case when we will have just one image in the list and all states might be meaningful, however I haven't thought much about it yet.

I see. Yeah make sense to keep it since we still use it for the single image scenario. I didn't realize that. Thanks for the clarification!

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
```diff
@@ -80,18 +129,61 @@ class PreviewImageViewModel : ViewModel() {
...
 
-    private fun createImageDataStartLoadingUiState(
-        loResImageUrl: String,
-        hiResImageUrl: String
-    ): ImageDataStartLoadingUiState = ImageDataStartLoadingUiState(
-        ImageData(loResImageUrl, hiResImageUrl)
-    )
+    private fun createViewPagerItemsInitialUiStates(
+        data: List<ImageData>
+    ): List<ImageUiState> {
```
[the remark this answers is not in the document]
ashiagr: Changed here 7cbdb37

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
```diff
@@ -100,29 +192,48 @@ class PreviewImageViewModel : ViewModel() {
 
     data class ImageData(val lowResImageUrl: String, val highResImageUrl: String)
 
+    data class UiState(
+        val viewPagerItemsUiState: ViewPagerItemsUiState
+    ) {
+        sealed class ViewPagerItemsUiState(
```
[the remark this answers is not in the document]
ashiagr: Changed here 7cbdb37

malinajirka:
LGTM! Thanks for the changes.

I'm merging the PR even though lint and strings-check are not passing, as it goes only to a feature branch.

PR #11611, opened 2020-04-08
Issue/11518 media picker actions

### LINKED ISSUE #11518: Replace system image picker with custom image picker in gutenberg
### Expected behavior
When the user inserts image/gallery/video block and they click on "Choose from device" a custom media picker is shown. User can select single or multiple images depending on the block they are inserting the media into.

### Actual behavior
When the user inserts image/gallery/video block and they click on "Choose from device" a OS default media picker is shown. The UI can significantly differ across devices. Moreover, we can't modify its UI. It's blocking our progress in [MediaEditing project](https://github.com/orgs/wordpress-mobile/projects/54).

- [x] Re-route gutenberg's "Choose from device" button into our custom media picker for image and gallery blocks (fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/11520)
- [x] Add support for selecting multiple images (fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/11520)
- [x] Re-route gutenberg's "Choose from device" button into our custom media picker for video and text&media blocks (fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/11527)
- [x] Make sure the picker displays only image files for image/gallery blocks and only video files for video block (fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/11527)
- [x] Hide "Take photo/WPLibrary" in the custom media picker bottom bar (https://github.com/wordpress-mobile/WordPress-Android/pull/11611)
- [x] Make sure the custom media picker bottom bar has two different buttons for selecting an image from the local storage and selecting a video from the local storage (https://github.com/wordpress-mobile/WordPress-Android/pull/11611)
- [x] Verify the app requests permissions when necessary
- [x] Make sure to review the custom picker in unrelated flows before we merge the changes into develop (site avatar, user avatar, WPMedia library picker, Media Library, ..?)

malinajirka:
Fixes #11518

Final PR for the issue #11518 
- Hides "Pick from WP library" option in the bottom bar in the custom picker
- Select from device pop up with "select photo" and "select video" options is shown only for media&text block as the other blocks support either image or video not both.
- Enables multi-selection in the os default picker for blogs which support multi-selection

To test:
Gifs with all the flows https://cloudup.com/cjA9rSxls0a (cc @mbshakti could you please review if these flows look ok?)

Ideally test all the media blocks
- [x] Image block custom picker
- [x] Image block os picker
- [x] Image block take photo
- [x] Video block custom picker
- [x] Video block os picker
- [x] Video block take video
- [x] Media&Text block custom picker
- [x] Media&Text block os picker
- [x] Media&Text block take photo 
- [x] Media&Text block take video
- [x] Gallery block custom picker
- [x] Gallery block os picker
- [x] Gallery block take photo

PR submission checklist:

cc @hypest @mkevins @marecar3 Could one of you please review the gifs if all the flows look as expected? Thanks! (if you want to check all the code changes compare `issue/11518-media-picker-actions` against `develop` - https://github.com/wordpress-mobile/WordPress-Android/compare/develop...issue/11518-media-picker-actions)


WordPress/src/main/java/org/wordpress/android/ui/media/MediaBrowserType.java
```diff
@@ -23,20 +23,39 @@ public boolean isBrowser() {
...
-               || this == GUTENBERG_SINGLE_VIDEO_PICKER || this == GUTENBERG_SINGLE_MEDIA_PICKER
+        return this == EDITOR_PICKER
+               || this == AZTEC_EDITOR_PICKER
+               || this == GUTENBERG_VIDEO_PICKER
+               || this == GUTENBERG_SINGLE_VIDEO_PICKER
+               || this == GUTENBERG_SINGLE_MEDIA_PICKER
+               || this == GUTENBERG_MEDIA_PICKER;
+    }
+
+    public boolean isGutenbergPicker() {
```
malinajirka, line 53: This method is new the rest is just formatted.

hypest:
👋 thanks for pinging for reviewing the flows @malinajirka .

I haven't yet looked at all of them but the one that we actually capture a new photo in Gutenberg caught my attention: https://cloudup.com/iIP5b_K5aVO. It looks like we'll need to perhaps remove the individual "Take a photo" option in Gutenberg's media options menu? Is there a plan for that?

Also, pinging @iamthomasbishop to have a look at the flows as well since those relate to the design side of things.

malinajirka:
>I haven't yet looked at all of them but the one that we actually capture a new photo in Gutenberg caught my attention: https://cloudup.com/iIP5b_K5aVO. It looks like we'll need to perhaps remove the individual "Take a photo" option in Gutenberg's media options menu? Is there a plan for that?

We discussed this with Shakti and we thought it wouldn't hurt to have a duplicated path to achieve the same goal. I definitely wouldn't remove the individual "Take a photo" option in Gutenberg's media options menu as it's shorter. Wdyt?

malinajirka:
Just to clarify a bit more. We won't support editing the media before it gets uploaded in the "take a photo" and "OS default picker" paths. The media-editing will be supported only when the user selects a media in the custom picker.

ashiagr (approved):
Nice work @malinajirka ! I tested different media blocks, they work as described :).
Feel free to merge if flows are approved design-wise.

iamthomasbishop:
@malinajirka Thanks for the gifs as examples, helps a lot to visualize.

A few things that stand out to me while reviewing:

- When the device picker opens, it took me a minute to realize that there were footer actions. I think we should either move these actions/tabs to sit under the App Bar or perhaps add text labels to them, in an effort to improve discoverability? (something like this: 

image

- I can’t help but wonder if we should present the dialog (after you select “choose from devices “) as a bottom sheet of sorts, kind of like Aztec’s initial picker. Thoughts?

- I was thinking about what @hypest mentioned about removing the “take a...” options, and it doesn’t bother me all that much to have multiple routes to an action, but I don’t know if the “take a...” options with GB are used enough to justify such prominent placement. So maybe if we improve the discoverability of that from the “choose photo/video” step, we can remove from the initial source selection sheet in GB. Does that make sense? // cc @hypest I don’t think we have any data on this, but would be useful — my hunch tells me not many people are taking images directly into the editor, but I don’t have data to back it up.

malinajirka:
Thanks for the reviews ;)!

>When the device picker opens, it took me a minute to realize that there were footer actions. I think we should either move these actions/tabs to sit under the App Bar or perhaps add text labels to them, in an effort to improve discoverability? (something like this:

>I was thinking about what @hypest mentioned about removing the “take a...” options, and it doesn’t bother me all that much to have multiple routes to an action, but I don’t know if the “take a...” options with GB are used enough to justify such prominent placement. So maybe if we improve the discoverability of that from the “choose photo/video” step, we can remove from the initial source selection sheet in GB. Does that make sense? // cc @hypest I don’t think we have any data on this, but would be useful — my hunch tells me not many people are taking images directly into the editor, but I don’t have data to back it up.


We might want to consider leaving both these decisions for the Media Consolidation project (cc @frosty ). Goal of this task is to replace the OS default picker with a custom picker. This PR is re-using an existing picker which definitely could be improved, but at the same time is imo much better than the OS default picker. Unless you feel one of these issues is vital, I'd prefer to limit the scope of this task as much as possible. Wdyt @iamthomasbishop ?



>I can’t help but wonder if we should present the dialog (after you select “choose from devices “) as a bottom sheet of sorts, kind of like Aztec’s initial picker. Thoughts?

This is the default OS picker and we can't change its UI. It can look differently on different devices. We want to keep it so the user can choose photos from a custom folders as our custom picker shows only content from default media folders.

hypest:
I'd prefer it if we don't have multiple paths to do the same thing in flows so close together as these, unless it's pretty much necessary. I also agree that we don't want to turn this ticket/PR into a Media Consolidation project one.

So, let me ask, can we instead hide the "camera" tab of the custom picker when its client hints it only wants to _choose_ a photo? Would that help unblock the MediaEditing project as #11518 requires?

malinajirka:
>So, let me ask, can we instead hide the "camera" tab of the custom picker when its client hints it only wants to choose a photo? Would that help unblock the MediaEditing project as #11518 requires?

Yes, we could do that and it'd unblock us. @mbshakti If I recall correctly, you preferred having two buttons in the bottom bar, right?

mbshakti:
When we made the decision to keep the camera button in the bottom bar, the reasoning was that it wouldn't hurt the user, and in case the user changes their mind, they can always use the camera. I don't think having multiple paths to do the same thing in flows is problematic, but I understand if @hypest believes so, especially since the placement is too prominent as @iamthomasbishop mentioned.

We can go ahead and hide it from this view. The only weird thing there is that we would just have one button in the bottom bar then, but I don't mind that.

malinajirka:
>>I can’t help but wonder if we should present the dialog (after you select “choose from devices “) as a bottom sheet of sorts, kind of like Aztec’s initial picker. Thoughts?

>This is the default OS picker and we can't change its UI. It can look differently on different devices. We want to keep it so the user can choose photos from a custom folders as our custom picker shows only content from default media folders.

@iamthomasbishop I'm sorry, I misread your comment. We could show the picker (after you select "choose from device") in a bottom sheet. However, I'd make this change in the Media Consolidation project as well - or perhaps even a different project as it's technically not related to Media Consolidation.

malinajirka:
@ashiagr I've remove the "take photo/video" option from the custom picker in gutenberg. Could you please review https://github.com/wordpress-mobile/WordPress-Android/pull/11611/commits/ed6b137438685663619261e43d91c498a6cf1cc6 and merge the PR if you are ok with the changes. Thanks!

ashiagr:
Thanks @malinajirka. I tested and found camera option hidden from the custom picker in gutenberg. Merging these changes 👍.

PR #11636, opened 2020-04-10
Post & Page List : Changed auto upload text from cancel to cancel upload

### LINKED ISSUE #11515: Post+Page list: Rename Cancel action in collapsed view
When the user edits a post/page and confirms the changes while offline, a label is displayed (eg. "We'll publish this post/page when your device is back online"). We also display a "Cancel" action on the item. The goal of this task is to rename the "Cancel" action in the collapsed view on Post List and regular view on Page List (we want to keep "cancel" on the expanded view on post list). 
We are not sure about the new text. One of the proposals is "Cancel publishing". However, I think we support the cancel action even when the user explicitly saves a draft in offline. In that scenario "cancel publishing" doesn't make sense.

jd-alexander:
Fixes #11515

## Solution
Updates the `Cancel` action on the Page & Post list item to `Cancel upload`. 

## Testing
1. Turn on airplane mode.  
2. Create a post or page and publish it. 
3. An auto-upload label is displayed on the list item (eg. "We'll publish this post/page when your device is back online").
3. You will see that `Cancel upload` is shown on both the Page/Post List list item option menu. 

Page List (Compact)  | Post List (Compat)     | Post List (Normal) 
--------|-------  |  -------  
      |    |  

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/widgets/PostListButtonType.kt
+++ b/WordPress/src/main/java/org/wordpress/android/widgets/PostListButtonType.kt
@@ -30,7 +30,7 @@ enum class PostListButtonType constructor(
             R.drawable.ic_trash_white_24dp,
             R.attr.colorOnSurface
     ),
-    BUTTON_CANCEL_PENDING_AUTO_UPLOAD(14, R.string.cancel, R.drawable.ic_undo_white_24dp, R.attr.wpColorWarningDark);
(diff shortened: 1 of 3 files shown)


malinajirka:
@jd-alexander Thanks for working on this task! ;) However, AFAIU we were supposed to change the title just in the expanded view as per the description of the issue `The goal of this task is to rename the "Cancel" action in the collapsed view on Post List and regular view on Page List (we want to keep "cancel" on the expanded view on post list).` Wdyt?

jd-alexander:
Thanks for pointing that out @malinajirka I misread it. I will address that and send in another PR to correct it. Thanks.

Also thanks for the review @ashiagr

PR #11642, opened 2020-04-10
Add mobile Gutenberg localization to code freeze

loremattei:
This PR adds a step in the `code_freeze` lane to import strings from the Gutenberg submodule into the main `strings.xml`.
As a first step, the script will stop and ask for a manual commit if any change to the main `strings.xml` is done (the script will modal-wait on the current terminal, but the changes can be reviewed and committed in another terminal). 
We can automate this step when we are confident that everything works as expected.  

To test:
1. Run `bundle install`.
2. Checkout a new test branch from this one.
3. Update the Gutenberg submodule in order to include some changes in the strings.
4. Run `bundle exec fastlane localize_gutenberg` (it's the same sub-lane called during the code freeze).
5. Verify that the strings have been imported. 

PR submission checklist:

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -67,7 +67,9 @@ ENV["HAS_ALPHA_VERSION"]="true"
     setbranchprotection(repository:GHHELPER_REPO, branch: "release/#{new_version}")
     setfrozentag(repository:GHHELPER_REPO, milestone: new_version)
     
+    localize_gutenberg(skip_module_update: true)
     localize_libs()
+    ensure_git_status_clean()
     android_tag_build()
     get_prs_list(repository:GHHELPER_REPO, start_tag:"release/#{old_version}", report_path:"#{File.expand_path('~')}/wpandroid_prs_list_#{old_version}_#{new_version}.txt")
   end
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -67,7 +67,9 @@ ENV["HAS_ALPHA_VERSION"]="true"
     setbranchprotection(repository:GHHELPER_REPO, branch: "release/#{new_version}")
     setfrozentag(repository:GHHELPER_REPO, milestone: new_version)
     
+    localize_gutenberg(skip_module_update: true)
```
maxme, line 70: Just a note to say that's great we're calling `gutenberg_dep_check()`  before that, that should ensure the bundle, and bundled strings are up to date.

fastlane/Fastfile
```diff
@@ -319,11 +321,37 @@ ENV["HAS_ALPHA_VERSION"]="true"
     {library: "Image Editor", strings_path: "./libs/image-editor/ImageEditor/src/main/res/values/strings.xml", exclusions: []}
   ]
 
+  private_lane :commit_strings do | options |
+    if (options[:auto_commit]) then
+       sh("cd .. && git add #{main_strings_path}")
+       sh("git commit -m 'Update strings for translation'")
+       sh("git push")
```
maxme, line 328: Can we do `git push origin HEAD` here? Just to make sure the local git config doesn't interfere?

loremattei: Ah! Thank you for catching this @maxme! 
t seems it's hard for me to get rid on this old habit! I'm sorry for having missed it again!
This is now updated!

PR #11648, opened 2020-04-13
Post/Page compact cancel auto upload action improvement

### LINKED ISSUE #11515: Post+Page list: Rename Cancel action in collapsed view

jd-alexander:
Fixes #11515
This PR is a continuation of https://github.com/wordpress-mobile/WordPress-Android/pull/11636 
It addresses the notion that`Cancel Upload` should only be visible on compact actions view of both the page and post list. 

To test:
1. Turn on airplane mode.
2. Create a post or page and publish it.
3. An auto-upload label is displayed on the list item (eg. "We'll publish this post/page when your device is back online").
4. You will see that Cancel upload is shown on both the compact Page/Post List list item option menu.
5. For the post list, the default/normal view will only contain "Cancel"

Default | Compact 
--------|-------
        |   

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/viewmodel/posts/PostListItemUiStateHelper.kt
+++ b/WordPress/src/main/java/org/wordpress/android/viewmodel/posts/PostListItemUiStateHelper.kt
@@ -46,6 +46,7 @@ import org.wordpress.android.viewmodel.posts.PostListItemType.PostListItemUiStat
 import org.wordpress.android.viewmodel.uistate.ProgressBarUiState
 import org.wordpress.android.widgets.PostListButtonType
 import org.wordpress.android.widgets.PostListButtonType.BUTTON_CANCEL_PENDING_AUTO_UPLOAD
+import org.wordpress.android.widgets.PostListButtonType.BUTTON_CANCEL_PENDING_AUTO_UPLOAD_COMPACT
 import org.wordpress.android.widgets.PostListButtonType.BUTTON_DELETE
 import org.wordpress.android.widgets.PostListButtonType.BUTTON_DELETE_PERMANENTLY
 import org.wordpress.android.widgets.PostListButtonType.BUTTON_EDIT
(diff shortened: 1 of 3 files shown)


WordPress/src/main/java/org/wordpress/android/viewmodel/posts/PostListItemUiStateHelper.kt
```diff
@@ -89,16 +90,28 @@ class PostListItemUiStateHelper @Inject constructor(
         val onButtonClicked = { buttonType: PostListButtonType ->
             onAction.invoke(post, buttonType, POST_LIST_BUTTON_PRESSED)
         }
-        val buttonTypes = createButtonTypes(
+        val defaultButtonTypes = createButtonTypes(
```
malinajirka, line 93: I feel like it might be a bit difficult to read when we call the same method (which unfortunately needs a lot of parameters) twice in a row and we change just one of the parameters. How would you feel about something like this

```
val getButtonTypes: (isCompact: Boolean) -> List<PostListButtonType> = { isCompact ->
            createButtonTypes(
                    postStatus = postStatus,
                    isLocalDraft = post.isLocalDraft,
                    isLocallyChanged = post.isLocallyChanged,
                    uploadUiState = uploadUiState,
                    siteHasCapabilitiesToPublish = capabilitiesToPublish,
                    statsSupported = statsSupported,
                    isCompactActions = isCompact
            )
        }
        val defaultActions = createDefaultViewActions(getButtonTypes(false), onButtonClicked)
        val compactActions = createCompactViewActions(getButtonTypes(true), onButtonClicked)
```

WordPress/src/main/java/org/wordpress/android/widgets/PostListButtonType.kt
```diff
@@ -31,6 +31,12 @@ enum class PostListButtonType constructor(
             R.attr.colorOnSurface
     ),
     BUTTON_CANCEL_PENDING_AUTO_UPLOAD(
+            14,
+            R.string.cancel,
+            R.drawable.ic_undo_white_24dp,
+            R.attr.wpColorWarningDark
+    ),
+    BUTTON_CANCEL_PENDING_AUTO_UPLOAD_COMPACT(
```
malinajirka, line 39: We might want to consider refactoring `PostListButtonType` into a sealed class instead of introducing a new type

`data class BUTTON_CANCEL_PENDING_AUTO_UPLOAD(val isCompact: Boolean) : PostListButtonType(14, if (isCompact) R.string.cancel else R.string.pages_and_posts_cancel_auto_upload, R.drawable.ic_undo_white_24dp, R.attr.wpColorWarningDark)`

I'm not sure about this. I'd like to hear your thoughts. :)

jd-alexander: This sounds like a good idea!

jd-alexander:
Closed since this doesn't have to be changed right now.

PR #11665, opened 2020-04-15
Issue/11153 support editing multiple images: open selected image on crop screen

### LINKED ISSUE #11153: Add support for editing multiple local images from the Hub before they are uploaded
A horizontal RecyclerView on the Hub is added in https://github.com/wordpress-mobile/WordPress-Android/issues/11152. Goal of this task is to add support for editing the images shown in the RecyclerView and appending them to Gutenberg. 

- [x] Make sure clicking on Edit opens the selected image
- [x] Make sure the ImageView and thumbnails gets udpated after the user edits the image
- [x] Make sure all images are added to the editor when the user clicks on "Insert"

This is an overall design of the flow (Android specific design will be added later):
Screenshot 2020-01-23 at 19 04 39

~Note: Gutenberg currently doesn't support selecting multiple images from the local storage. The flow added in this ticket might not be reachable by the user until the support is added to Gutenberg.~
This is supported now in Gallery block.

ashiagr:
Partially fixes #11153

This PR 
- Displays crop menu
- Manages crop menu enabled state for currently displayed image
     - [x] **High res image loaded image ui state** - enabled
     - [x] **Other  image ui states** - disabled (might want to change the tint color)
- Manages thumbnails layout visibility state 
     - [x] **Multiple images** - visible
     - [x] **Single image** - not visible
- Navigation to crop screen 
     - [x]  **Multiple images** - displayed image is passed to crop screen when crop menu is clicked
     - [x]  **Single image** - crop screen auto opens with displayed image 
- Back navigation from crop screen
     - [x]  **Multiple images** - returns to preview screen
     - [x]  **Single image** - returns to gutenberg editor

Note:
- Navigation on crop (crop screen) currently disabled for multiple images.
- Test cases not final, updating in a [separate branch ](https://github.com/wordpress-mobile/WordPress-Android/commits/issue/11153-update-tests)

 | Multiple Images (crop opened on menu click)| Single Image (auto open crop) |
|--------------|----|
|navtocrop| singlecrop|

PR submission checklist:


libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
```diff
@@ -3,27 +3,34 @@ package org.wordpress.android.imageeditor.preview
...
 import org.wordpress.android.imageeditor.preview.PreviewImageViewModel.ImageUiState.ImageInHighResLoadSuccessUiState
 import org.wordpress.android.imageeditor.preview.PreviewImageViewModel.ImageUiState.ImageInLowResLoadFailedUiState
 import org.wordpress.android.imageeditor.preview.PreviewImageViewModel.ImageUiState.ImageInLowResLoadSuccessUiState
+import org.wordpress.android.imageeditor.viewmodel.Event
 
 class PreviewImageViewModel : ViewModel() {
-    private val _uiState: MutableLiveData<UiState> = MutableLiveData()
-    val uiState: LiveData<UiState> = _uiState
+    private val _uiState: MutableLiveData<Event<UiState>> = MutableLiveData()
+    val uiState: LiveData<Event<UiState>> = _uiState
```
malinajirka, line 19: Can you please elaborate why we need to use Event for the UiState. 

`Event` is usually used for events we want to consume just once (eg. snackbar msgs, navigation) to make sure we don't perform the same navigation or show the same snackbar msg more than once (eg. after rotation).

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
```diff
@@ -35,14 +42,24 @@ class PreviewImageViewModel : ViewModel() {
...
         )
-        // TODO: handle file save on done button action to be added later
-//        val highResImageJustLoadedIntoView = newState != currentState && newState == ImageInHighResLoadSuccessUiState
-//        val imageNotLoadedIntoFile = loadIntoFile.value !is ImageLoadToFileSuccessState
-//        if (highResImageJustLoadedIntoView && imageNotLoadedIntoFile) {
-//            updateLoadIntoFileState(ImageStartLoadingToFileState(currentUrl))
-//        }
-        val currentUiState = uiState.value as UiState
-        updateUiState(currentUiState.copy(viewPagerItemsStates = newImageUiStates))
+        val newImageState = newImageUiStates[currentPosition]
```
malinajirka, line 45: `currentPosition` seems a bit misleading name for the parameter. It's not the position of the currently selected item but a position of any item which has just finished loading, right? I'd consider renaming it to `position`, wdyt?

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
```diff
@@ -35,14 +42,24 @@ class PreviewImageViewModel : ViewModel() {
...
+            updateLoadIntoFileState(
+                ImageStartLoadingToFileState(
+                    imageUrl = currentImageState.data.highResImageUrl,
+                    position = currentPosition
+                )
+            )
+        }
+        updateUiState(currentUiState.copy(
+                viewPagerItemsStates = newImageUiStates,
+                editActionsEnabled = shouldEnableEditActionsForImageState(newImageState)
```
malinajirka, line 60: Should we really change `shouldEnableEditActionsForImageState(newImageState)` here? Are we sure the image that has just finished loading is the selected image? AFAIU more than one image can be loading when we are for example in the middle of a swipe, right?

I think I managed to reproduce this issue - I changed the url of the second dummy image to a non-existing URL so it'd keep failing -> this is what I recorded https://cloudup.com/co9XP0q-HrR -> notice the crop button is sometimes disabled and sometimes enabled.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
```diff
@@ -168,19 +181,27 @@ class PreviewImageFragment : Fragment() {
...
+                UiHelpers.updateVisibility(thumbnailsTabLayout, state.thumbnailsTabLayoutVisible)
+            }
         })
 
-        /*viewModel.loadIntoFile.observe(this, Observer { fileState ->
-            if (fileState is ImageStartLoadingToFileState) {
-                loadIntoFile(fileState.imageUrl)
+        viewModel.loadIntoFile.observe(this, Observer { fileStateEvent ->
+            fileStateEvent?.getContentIfNotHandled()?.let { fileState ->
+                if (fileState is ImageStartLoadingToFileState) {
```
malinajirka, line 194: I know this was here before, but it feels weird that we are propagating different states to the UI even though the UI cares only about a single state - ImageStartLoadingToFileState. I think we might want to consider separating the state from the "start loading an image into a file" action. Wdyt?

libs/image-editor/ImageEditor/src/main/res/navigation/nav_graph.xml
```diff
@@ -21,6 +19,10 @@
                 android:name="outputFileExtension"
                 app:argType="string"
                 app:nullable="true"/>
+            <argument
```
malinajirka, line 22: Hmm interesting, do we really need to define the same argument twice? Or is the first definition input and the second definition output?

malinajirka (commented):
Thanks @ashiagr! Wonderful job :)

I've left a couple of questions and comments. Let me know what you think ;).

libs/image-editor/ImageEditor/src/main/res/navigation/nav_graph.xml
ashiagr: I referred [source frag nav graph args](https://github.com/android/architecture-samples/blob/master/app/src/main/res/navigation/nav_graph.xml#L30-L35), [dest frag nav graph args](https://github.com/android/architecture-samples/blob/master/app/src/main/res/navigation/nav_graph.xml#L64-L71) and [args fetch in the dest frag](https://github.com/android/architecture-samples/blob/master/app/src/main/java/com/example/android/architecture/blueprints/todoapp/addedittask/AddEditTaskFragment.kt#L42) in the architecture blueprints sample. 

So yes, they seem to be input and output args. 

The only difference is args are defined outside actions in the source frag in the sample. We can do that if we have multiple actions for a fragment needing the same args.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
ashiagr: Renamed here: 2867882

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
ashiagr: I noticed active observer count (`hasActiveObservers()`) increased every time for `uiState` `LiveData` on returning to preview from crop screen. To prevent excessive UI re-rendering I decided to use `Event`.

I didn't get a chance to investigate it but [Architecture Navigation Component : onCreateView gets called every time](https://stackoverflow.com/questions/54879581/architecture-navigation-component-oncreateview-gets-called-every-time) could be a potential reason. However I might be wrong. Let me know what you think. 

Quoting an [explanation](https://issuetracker.google.com/issues/109856764) from a linked issue: 

> Google doesn't want you to stack a bunch of fragments on top of each other using show/hide because it promotes bad app architecture and wastes available memory. .... Nobody is stopping you from keeping views in memory manually, but the whole point of these jetpack components is to stop you from doing these mistakes in the first place.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
```diff
@@ -156,7 +169,7 @@ class PreviewImageFragment : Fragment() {
         // Set adapter data before the ViewPager2.restorePendingState gets called
         // to avoid manual handling of the ViewPager2 state restoration.
         viewModel.uiState.value?.let {
-            (previewImageViewPager.adapter as PreviewImageAdapter).submitList(it.viewPagerItemsStates)
+            (previewImageViewPager.adapter as PreviewImageAdapter).submitList(it.peekContent().viewPagerItemsStates)
```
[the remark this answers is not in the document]
ashiagr: As we're not observing any changes to `uiState` here, and [view model sets the value](https://github.com/wordpress-mobile/WordPress-Android/blob/c8a7c019a0eb4547a719f17b5fc97e2698d92e8a/libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt#L29) before we reach this point, I thought to just use `peekContent`.

We still need to decide if we should use `Event` for `uiState`.  Will reconsider fixing it after we've made a decision.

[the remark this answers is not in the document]
malinajirka: My understanding is that `peekcontent` is used for cases where the event isn't handled. Whereas `getContentIfNotHandled` is used for cases where we handle the event and we don't need to handle it again - which is afaiu this scenario. The current implementation handles the same event twice within 1ms - once here and once in the `observe` method, right?

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
malinajirka: Thanks for the clarification :)

>I noticed active observer count (hasActiveObservers()) increased every time for uiState LiveData on returning to preview from crop screen. To prevent excessive UI re-rendering I decided to use Event.

ohh, that'd mean we are leaking the fragment. Which is a high priority issue anyway, but I agree we can fix it after the project issues. ;)

libs/image-editor/ImageEditor/src/main/res/navigation/nav_graph.xml
malinajirka: It seems we don't use the output argument (shouldReturnToPreviewScreen) though, right?

ashiagr: Sorry if I misunderstood your comment. So the source frag args are passed in the `PreviewImageFragmentDirections` action: actionPreviewFragmentToCropFragment

```
PreviewImageFragmentDirections.actionPreviewFragmentToCropFragment(
                    inputFilePath,
                    outputFileExtension,
                    shouldReturnToPreviewScreen
                )
```

and they're received  by the `CropFragment` in `CropFragmentArgs`. We do use `shouldReturnToPreviewScreen` in the `CropFragment`.

Am I missing something?

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
malinajirka: I'm not sure I follow the issue from the StackOverFlow. When a fragment is put on backstack we always want to invoke `onDestroyView` and whenever the fragment is poped from a backstack we always want to invoke `onCreateView`. This is imo expected behavior no matter if we use NavComp or not. As per the [fragment's lifecycle diagram](https://blog.avenuecode.com/hs-fs/hubfs/The%20activity%20and%20fragment%20lifecycles%20(1).png?width=640&name=The%20activity%20and%20fragment%20lifecycles%20(1).png).


I've checked `viewModel.uiState.mActiveCount` in the `setupObservers` method after I return back from the Crop fragment and the count is == 0 (even when we don't use `Event`). If we want to reduce the number of times the observer gets notified with a UI change, we might want to consider introducing a throttle on the uiState LiveData.  

Wdyt? Is there something I'm missing?

libs/image-editor/ImageEditor/src/main/res/navigation/nav_graph.xml
malinajirka: Ohh, I see. That's a very weird API that we need to declare the same thing twice :D. I haven't noticed this until now. Thanks for the clarification!

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
malinajirka: I think it's safe to use the `Event`. It just something we don't usually do, but I don't see any downsides except of inconsistency. So feel free to keep it if you want. I'm glad that it seems we are not leaking the fragment :) - I don't like leaks:P.

ashiagr: I really want to research bit more on this but later. Till that time let's keep `Event`.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/crop/CropFragment.kt
```diff
@@ -139,9 +147,13 @@ class CropFragment : Fragment(), UCropFragmentCallback {
     }
 
     private fun navigateBackWithCropResult(cropResult: CropResult) {
-        activity?.let {
-            it.setResult(cropResult.resultCode, cropResult.data)
-            it.finish()
+        if (navArgs.shouldReturnToPreviewScreen) {
```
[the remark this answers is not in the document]
ashiagr: Starting another fragment for result is not properly supported in the jetpack nav component. We might end up using some boilerplate code, or view model sharing or use this nav component fix  [release 2.3.0-alpha05](https://issuetracker.google.com/issues/79672220#comment55). 

As discussed in the call, we’d continue using current approach for now.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/crop/CropFragment.kt
```diff
@@ -106,6 +107,13 @@ class CropFragment : Fragment(), UCropFragmentCallback {
     override fun onOptionsItemSelected(item: MenuItem): Boolean = if (item.itemId == R.id.menu_done) {
         viewModel.onDoneMenuClicked()
         true
+    } else if (item.itemId == android.R.id.home) {
+        if (navArgs.shouldReturnToPreviewScreen) {
```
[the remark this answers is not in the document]
ashiagr: Same as above.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
ashiagr: Fixed here: 39b3b44

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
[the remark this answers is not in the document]
ashiagr: Tried using `getContentIfNotHandled` instead of `peekContent`. It resulted in a blank screen. So reverted to `peekContent` here: 955b690

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageViewModel.kt
ashiagr: Added a [note](https://github.com/orgs/wordpress-mobile/projects/54#card-36418894) in the project to investigate it later.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
ashiagr: I myself want to revisit this logic that came from single image editing and added a TODO in the view model. As this is non-blocking added a note in the project.

malinajirka (approved):
LGTM, thanks!

PR #11688, opened 2020-04-17
Prepublishing Nudges : Fetch tags on first launch 

### LINKED ISSUE #11687: Prepublishing Nudges : Tags screen empty on first use.
### Expected behavior
That the screen will show the tags when the user opens it for the first time.

### Actual behavior
The tags aren't shown. They only get shown if you go into Post Settings - Tags since a fetch action is called to refresh them. 

1. Check out this branch https://github.com/wordpress-mobile/WordPress-Android/pull/11656 and ensure that you currently have tags within the selected site. 
2. Create a new post. 
3. Press publish.
4. Select the Tags field.
5. See the empty tags screen. 

Tested on Google Pixel 4 XL, Android 10, WPAndroid `tags_navigation` ()

jd-alexander:
Fixes #11687 

To test:

1. Create a new post.
2. Press publish.
3. Select the Tags field.
4. See the tags screen and the list should be populated.

### Review 
2. Ensure that #11634 is merged.
3. Set this base branch to feature/prepublishing_nudges
4. Remove `Not Ready for Merge`
5. Merge :)

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingViewModel.kt
```diff
@@ -196,4 +200,11 @@ class PrepublishingViewModel @Inject constructor() : ViewModel() {
         val formattedTags = TextUtils.join(",", tags)
         return StringEscapeUtils.unescapeHtml4(formattedTags)
     }
+
+    /**
+     * Fetches the tags so that they will be available when the Tags action is clicked
+     */
+    private fun fetchTags() {
+        dispatcher.dispatch(TaxonomyActionBuilder.newFetchTagsAction(site))
```
malinajirka, line 208: Don't we need to check if the action has completed? Or what happens when the action doesn't complete and the user clicks on the Tags action?

jd-alexander: Thanks for pointing this out @malinajirka 🙇 So to answer the first part of your question, we don't check if the action completes because it seems like a fire & forget approach, especially since we are on a screen that doesn't even need these tags. I am doing this by following the approach used in the `EditPostSettingsFragment`
https://github.com/wordpress-mobile/WordPress-Android/blob/31ef563ac86efeb9a78ccf914fa30a5fd4359f4e/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostSettingsFragment.java#L579

This expects that by the time the user tries to edit the tags this operation would be completed in the background and they would see a preloaded list of previously added tags.

For the second part of your question, if the user clicks on the tags action and the action didn't complete then they won't see the preloaded list but it seems once they have been fetched once all is well. In terms of errors,  I simply logged it as seen here 5c8e893

Let me know what you think. Thanks 🙏

malinajirka:
LGTM, thanks!

PR #11691, opened 2020-04-17
Set My site fragment title to 'My site' instead of using site title

frosty:
IANAAD (I Am Not An Android Developer), so please let me know if I did anything wrong! This PR is a small tweak the title of the My Site screen to always read "My site" instead of duplicating the site name, which is shown in the screen header anyway.

site-title-android

To test:

* Build and run, and ensure the My Site fragment has the title "My site"
* Try selecting a different site and ensure the title stays the same

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.java
@@ -377,7 +377,7 @@ public class MySiteFragment extends Fragment implements
         setupClickListeners(rootView);
 
         mToolbar = rootView.findViewById(R.id.toolbar_main);
-        mToolbar.setTitle(mToolbarTitle);
+        mToolbar.setTitle(R.string.my_site_section_screen_title);
 
         mToolbar.inflateMenu(R.menu.my_site_menu);
 
@@ -955,9 +955,6 @@ public class MySiteFragment extends Fragment implements
         } else {
             mQuickActionButtonsContainer.setWeightSum(75f);
         }
-
-        // Refresh the title
-        setTitle(site.getName());
     }
 
     private void toggleAdminVisibility(@Nullable final SiteModel site) {
@@ -1009,12 +1006,9 @@ public class MySiteFragment extends Fragment implements
 
     @Override
     public void setTitle(@NonNull final String title) {
-        if (isAdded()) {
-            mToolbarTitle = (title.isEmpty()) ? getString(R.string.wordpress) : title;
-
-            if (mToolbar != null) {
-                mToolbar.setTitle(mToolbarTitle);
-            }
+        mToolbarTitle = title;
+        if (mToolbar != null) {
+            mToolbar.setTitle(title);
         }
     }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
@@ -813,12 +813,8 @@ public class WPMainActivity extends LocaleAwareActivity implements
     }
 
     private void updateTitle(PageType pageType) {
-        if (pageType == PageType.MY_SITE && mSelectedSite != null) {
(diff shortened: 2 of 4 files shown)


WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.java
```diff
@@ -1009,12 +1006,9 @@ public void onStart() {
 
     @Override
     public void setTitle(@NonNull final String title) {
-        if (isAdded()) {
```
planarvoid, line 1012: Did you remove the `isAdded` check because we no longer need context when setting the title?

WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
```diff
@@ -813,12 +813,8 @@ private void updateTitle() {
     }
 
     private void updateTitle(PageType pageType) {
```
planarvoid, line 815: the `updateTitle()` method is no longer used so you can remove it

WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.java
frosty: I removed the `isAdded` check because I simplified this implementation using the `Me` fragment as reference – I have no idea whether it's necessary, so happy to take your lead here.

WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
frosty: Isn't it still used from `onPageChanged()`?

WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
```diff
@@ -813,12 +813,8 @@ private void updateTitle() {
...
-        if (pageType == PageType.MY_SITE && mSelectedSite != null) {
-            ((MainToolbarFragment) mBottomNav.getActiveFragment()).setTitle(mSelectedSite.getName());
-        } else {
-            ((MainToolbarFragment) mBottomNav.getActiveFragment())
-                    .setTitle(mBottomNav.getTitleForPageType(pageType).toString());
-        }
+        ((MainToolbarFragment) mBottomNav.getActiveFragment())
+                .setTitle(mBottomNav.getTitleForPageType(pageType).toString());
```
planarvoid, line 817: It's out of scope of this PR but this whole logic is strange (and wrong). We don't actually need to update the title anywhere else than in the Bottom nav component since this is a callback triggered by the bottom nav. What we do here is that the bottom nav tells us that a tab was selected and we react to it by asking bottom nav for the active fragment and setting its title from the bottom nav.

WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.java
```diff
@@ -1009,12 +1006,9 @@ public void onStart() {
...
-            mToolbarTitle = (title.isEmpty()) ? getString(R.string.wordpress) : title;
-
-            if (mToolbar != null) {
-                mToolbar.setTitle(mToolbarTitle);
-            }
+        mToolbarTitle = title;
```
planarvoid, line 1009: I'm actually wondering if we need the body of this method at all since the toolbar title is a constant and we set it in the `onCreate` method

WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.java
planarvoid: I think it isn't necessary :-).

WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
planarvoid: I meant the one above this one, without parameters. I couldn't comment there :-)

frosty: Done 👍

PR #11686, opened 2020-04-17
Issue/11153 (multiple images) support "Insert" action

ashiagr:
This PR just adds Insert button as part of #11153

PR submission checklist:

- [X] I have considered adding unit tests where possible.
- [X] I have considered adding accessibility improvements for my changes.
- [X] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #11695, opened 2020-04-19
Issue/11153 insert multiple edited images into gutenberg

### LINKED ISSUE #11153: Add support for editing multiple local images from the Hub before they are uploaded

malinajirka:
Partially fixes #11153

Merge instructions:
1. Review the code 
2. Make sure the issues in the `notes` section are fixed
3. Test the changes
4. Make sure https://github.com/wordpress-mobile/WordPress-Android/pull/11686 is merged
5. Update target branch to `issue/11153-update-image-thumbnail-after-crop`
6. Merge the PR

Notes:
The feature still has a vital issues and the flow doesn't fully work yet. However, I didn't want to make even more changes to this PR. I think the best approach is to fix the issue in another PRs branched from this PR.
- [ ] When the PreviewImage flow is started with images saved in local storage, Glide's `asFile()` method returns an error - nothing happens after you click on "Crop" action in the Toolbar. We probably shouldn't even call `asFile()` on local images, wdyt? Moreover, we probably should show a snackbar and show an error to the user when `asFile` fails.

To test:
- Test after the issue mentioned above is fixed

1. Add an ImageBlock
2. Select "Add an image from local storage"
3. Select an image and click on "Edit"
4. Crop the image
5. Notice the cropped image gets added to the imageBlock
-------------------------------
1. Add a Gallery block
2. Select "Add an image from local storage"
3. Select multiple images and click on "Edit"
4. Crop some of the images
5. Click on "Insert"
6. Verify all images are added and the cropped images are actually cropped

PR submission checklist:

- [] I have considered adding accessibility improvements for my changes.


WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoPickerFragment.java
```diff
@@ -679,4 +684,15 @@ private void trackSelectedOtherSourceEvents(AnalyticsTracker.Stat stat, boolean
         properties.put("is_video", isVideo);
         AnalyticsTracker.track(stat, properties);
     }
+
+    private ArrayList<EditImageData.InputData> createListOfEditImageInputData(List<Uri> uris) {
+        ArrayList<EditImageData.InputData> inputData = new ArrayList<>(uris.size());
+        for (Uri uri : uris) {
+            String stringUri = uri.toString();
+            String outputFileExtension = MimeTypeMap.getFileExtensionFromUrl(stringUri);
+            // TODO Add support for nullable lowResImgUrl
```
malinajirka, line 693: I'm passing null as thumbnail URL because the same file should be used for both thumbnail and full image. However, the logic in the PreviewImageViewModel didn't work as expected when I passed the same value for both low and high res.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -2286,15 +2292,9 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
                     }
                     break;
                 case RequestCodes.IMAGE_EDITOR_EDIT_IMAGE:
-                    if (data.hasExtra(UCrop.EXTRA_OUTPUT_URI)) {
```
malinajirka, line 2289: I've consolidated the results from the ImageEditor and decided to always use the same path. The result contains either a list with a single item or a list with multiple items.

WordPress/src/main/java/org/wordpress/android/util/WPMediaUtils.java
```diff
@@ -486,6 +491,26 @@ public static boolean fetchMediaAndDoNext(Context context, Uri mediaUri, MediaFe
         }
     }
 
+    public static List<Uri> retrieveImageEditorResultAndTrackEvent(Intent data) {
+        if (data != null && data.hasExtra(PreviewImageFragment.ARG_EDIT_IMAGE_DATA)) {
+            Map<String, String> properties = new HashMap<>();
+            properties.put("actions", "crop");
+            AnalyticsTracker.track(Stat.MEDIA_EDITOR_USED, properties);
```
malinajirka, line 498: I'm not sure about having the track event here. However, I wanted to re-use the same code in EditPostActivity.onActivityResult and PhotoPickerActivity.onActivityResult and I couldn't think of a better place. Any ideas?

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
```diff
@@ -189,33 +140,27 @@ class PreviewImageFragment : Fragment() {
...
         }
     }
 
     private fun initializeViewModels(nonNullIntent: Intent) {
         viewModel = ViewModelProvider(this).get(PreviewImageViewModel::class.java)
         setupObservers()
-        // TODO: replace dummy list
-        viewModel.onCreateView(imageDataList)
+        val inputData = nonNullIntent.getParcelableArrayListExtra<EditImageData.InputData>(ARG_EDIT_IMAGE_DATA)
+        numberOfImages = inputData.size
```
malinajirka, line 157: I had to change the way we retrieve the number of images. I think we might want to move the logic into the VM as it seems inconsistent now. I run out of time though.

WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoPickerFragment.java
ashiagr: Fixed here: 94ca331

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/preview/PreviewImageFragment.kt
ashiagr: Moved here: cfe8cdf

WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoPickerFragment.java
```diff
@@ -679,4 +684,15 @@ private void trackSelectedOtherSourceEvents(AnalyticsTracker.Stat stat, boolean
         properties.put("is_video", isVideo);
         AnalyticsTracker.track(stat, properties);
     }
+
+    private ArrayList<EditImageData.InputData> createListOfEditImageInputData(List<Uri> uris) {
+        ArrayList<EditImageData.InputData> inputData = new ArrayList<>(uris.size());
+        for (Uri uri : uris) {
+            String stringUri = uri.toString();
+            String outputFileExtension = MimeTypeMap.getFileExtensionFromUrl(stringUri);
```
ashiagr, line 692: `MimeTypeMap.getFileExtensionFromUrl(stringUri)` is returning a blank value here and results in transparency loss for png images. We can probably create a function like this that checks for `SCHEME_CONTENT` when determining the file extension:

```
   public String getFileExtension(Uri uri) {
        String fileExtension;
        if (uri.getScheme() != null && uri.getScheme().equals(ContentResolver.SCHEME_CONTENT)) {
            ContentResolver cr = requireContext().getContentResolver();
            String mimeType = cr.getType(uri);
            fileExtension = MimeTypeMap.getSingleton().getExtensionFromMimeType(mimeType);
        } else {
            fileExtension = MimeTypeMap.getFileExtensionFromUrl(uri.toString());
        }
        return fileExtension;
    }
```

WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoPickerFragment.java
malinajirka: Fixed in 39074c2c389829057318343b297a3d7d364d469b

WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoPickerFragment.java
malinajirka: Nice catch! ;) Fixed in 39074c2c389829057318343b297a3d7d364d469b

malinajirka:
Thanks for the review @ashiagr. It's ready for another round ;)

WordPress/src/main/java/org/wordpress/android/util/WPMediaUtils.java
ashiagr: We can move it to a better place when we start tracking edit action for individual images. I compared with iOS implementation, it too doesn't capture individual image edit action. So we're good for now. It's nice that you've included "track" in the method name.

ashiagr:
Thanks for the changes @malinajirka ! Nice work  👍.
Feel free to merge after [this discussion](https://github.com/wordpress-mobile/WordPress-Android/pull/11695#discussion_r410844619) is marked resolved.

Above issue and this one are taken care of in the PR #11696.

> When the PreviewImage flow is started with images saved in local storage, Glide's asFile() method returns an error - nothing happens after you click on "Crop" action in the Toolbar. We probably shouldn't even call asFile() on local images, wdyt? Moreover, we probably should show a snackbar and show an error to the user when asFile fails.

PR #11709, opened 2020-04-21
Reblog feature: small customization of site picker

develric:
This PR tries to further enhance the #11555 by adding a little step in site picker selection before to continue the reblog flow when a user has more than one site.

| NOTE | IMAGE |
|---|---|
| Adding `CONTINUE` menu disabled at beginning | image |
| Enabling `CONTINUE` menu after first site selection  | image |
| Search will allow site filtering and on filtered site selection user will be again in a  screen like the first one above  | image  |

cc @osullivanchris and @iamthomasbishop 

## To test
Be sure to be logged in with a WP.com account having more than one site (you should have enough sites to need to scroll in the site picker so to manual test the scroll to selected scenario)

### Feature Test
- From a reader post select the reblog button
- Check you get access to the site picker like below without any selected site

<p align="center">
  
</p>

- Check single site is selected if you tap on it, and you get an action bar like below

<p align="center">
  
</p>

- Check you can select other sites and still remain in the same screen as above with an updated action bar title set to the site title
- Check you can go back and come to the initial screen with no site selected
- Long press on a site and check nothing happens (hiding sites capability should not be available)
- Activate search and type some search string to get a filtered list of sites
- Select a site that is at the lower end of the main list
- Check the search view is dismissed and the sites list is scrolled to the selected site
- Tap on the Checkmark icon and check you can complete the reblog flow.

### Smoke Test previous available scenarios
Site Picker is used in `My Site > Switch Site`, `Share to WP`, `Login Epilogue`; Smock test these scenarios to check for eventual regressions. In particular check the Site Hide feature in `Switch Site` picker works as usual.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/SitePickerActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/SitePickerActivity.java
@@ -142,19 +142,16 @@ public class SitePickerActivity extends LocaleAwareActivity
                                 if (site != null) {
                                     mReblogActionMode.setTitle(site.getBlogNameOrHomeURL());
                                 }
-
                                 break;
                             case TO_NO_SITE_SELECTED:
                                 mSitePickerMode = SitePickerMode.REBLOG_SELECT_MODE;
                                 getAdapter().clearReblogSelection();
                                 break;
                         }
-
                         break;
                     case CONTINUE_REBLOG_TO:
                         SiteRecord siteToReblog = ((ContinueReblogTo) action).getSiteForReblog();
                         selectSiteAndFinish(siteToReblog);
-
                         break;
                     case ASK_FOR_SITE_SELECTION:
                         if (BuildConfig.DEBUG) {
@@ -168,7 +165,6 @@ public class SitePickerActivity extends LocaleAwareActivity
                             );
                             ToastUtils.showToast(this, R.string.site_picker_ask_site_select);
                         }
-
                         break;
                 }
                 return null;
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/SitePickerActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/SitePickerActivity.java
@@ -401,6 +401,13 @@ public class SitePickerActivity extends LocaleAwareActivity
                     @Override
                     public void onAfterLoad() {
                         showProgress(false);
+                        if (mSitePickerMode == SitePickerMode.REBLOG_CONTINUE_MODE && !isInSearchMode) {
+                            mAdapter.findAndSelect(mCurrentLocalId);
+                            int scrollPos = mAdapter.getItemPosByLocalId(mCurrentLocalId);
+                            if (scrollPos > -1 && mRecycleView != null) {
+                                mRecycleView.scrollToPosition(scrollPos);
+                            }
+                        }
                     }
                 },
                 mSitePickerMode);
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/SitePickerActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/SitePickerActivity.java
@@ -5,6 +5,7 @@ import android.app.Dialog;
 import android.app.DialogFragment;
 import android.content.Intent;
 import android.os.Bundle;
+import android.util.TypedValue;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
@@ -96,6 +97,7 @@ public class SitePickerActivity extends LocaleAwareActivity
     private SitePickerMode mSitePickerMode;
     private Debouncer mDebouncer = new Debouncer();
     private SitePickerViewModel mViewModel;
+    private float mDisabledOpacity;
 
     @Inject AccountStore mAccountStore;
     @Inject SiteStore mSiteStore;
@@ -121,6 +123,10 @@ public class SitePickerActivity extends LocaleAwareActivity
         }
 
         if (mSitePickerMode.isReblogMode()) {
+            TypedValue disabledAlpha = new TypedValue();
+            this.getResources().getValue(R.dimen.material_emphasis_disabled, disabledAlpha, true);
+            mDisabledOpacity = disabledAlpha.getFloat();
+
             mViewModel.getOnReblogActionTriggered().observe(
                     this,
                     unitEvent -> unitEvent.applyIfNotHandled(reblogAction -> {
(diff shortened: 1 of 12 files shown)


iamthomasbishop:
@develric I haven't had a chance to download a test build, but looking at the screenshots a couple of things stand out to me, some of which @osullivanchris has already mentioned.

* The `Continue` button def feels a little off. I think a `checkmark` icon might be a bit more balanced. The trade-off is that it might be _slightly_ less obvious, although translations on that `Continue` label could also make it feel less balanced if longer. Another option for the label (to keep it short) could be `Go`.
* I think the selection state on the list row could be a bit stronger. I'm not sure if that's the standard/generic style that we are inheriting or a custom, but I wonder if we should make the highlight stronger or add an icon/indicator when a list item is selected?

develric:
Hi @osullivanchris and @iamthomasbishop , thanks for the follow up 😊!

> Perhaps we could scroll to the position of the selected site?

Agree and modified the code (you should be able to check using the new APK from the above link)

> The Continue button def feels a little off...

Below how it results in both version; I still feel `continue` sounds clearer at least in english but fully get your concerns and probably would go with the checkmark icon; let me know wdyt, thanks 😊.

| image | image |
|---|---|
| image | image |


> I'm not sure if that's the standard/generic style that we are inheriting or a custom...

Yep, I used (only customized some behavior) the same site picker we have when switching site in My Site (see below image), also because this one has the search already in place like in iOS. 

<p align="center">
  
</p>

My understanding was we were trying to minimize modifications here but let me know 😊 (some site picker revisiting could be a good thing to tackle anyway, maybe in a dedicated task? 🤷‍♂️)

osullivanchris:
Thanks for the fixes @develric ! The checkmark looks better, and that's great you were able to do the scrolling thing easily enough.

> My understanding was we were trying to minimize modifications here but let me know 😊 (some site picker revisiting could be a good thing to tackle anyway, maybe in a dedicated task? 🤷‍♂️)

I agree this is just a generic/global thing we're inheriting here, so I don't think it needs to be addressed as part of this task

develric:
Thanks @osullivanchris , confirming final status as per below images:

| image | image |
|---|---|

iamthomasbishop:
@develric @osullivanchris Checkmark definitely feels better, although now that I see it I have a suggestion — and I apologize for not catching this earlier in the process, I should've thought of it earlier. Would it make more sense to perhaps morph the app bar into a "selection mode" app bar — in other words, something like this:

<a href="https://user-images.githubusercontent.com/1200432/80003535-cb543900-8486-11ea-9dc0-272e24108307.jpeg"></a>

| **No selection** | **Site selected** |
| --------------- | --------------- |
| `X` `Choose site` `🔍` | `<-` `[Site name]` `√` |

I think this would make selection even clearer and better align with Material. I do think the solution proposed in the last comment is workable, but I think this would be clearer, cleaner, and achieve the same result.

osullivanchris:
That is very nice @iamthomasbishop 

are we using that pattern anywhere else?

iamthomasbishop:
> are we using that pattern anywhere else?

@osullivanchris Good question. I hadn't even realized this, but we're using the same pattern already (diff actions, but same concept) on the site picker that shows when you tap on the `Switch site` button on the My Site view (when you get to the picker, long-press on a site cell).

develric:
Hi @iamthomasbishop and @osullivanchris 👋! I updated the code and now it should look (and hopefully behave 😊) as described by Thomas. You should be able to download the apk, below some screenshots for quick reference.

| image | image |
|---|---|




**Note for the code reviewer:** I updated the test steps in the PR description but not the initial images for history purposes. In the code I tried to move logic into the View Model I added but still some logic is in between the Activity/Adapter and the VM. I'm not trilled with that but I was trying to minimize the changes (as said previously probaby a review of the site pickers could make sense, for example screen rotation is without some memory of the status; I tried to cover it at least in this reblog scenario). Let me know 😊.

osullivanchris:
Thanks for the steps in the description the top, really helpful to make sure I'm covering everything.

This works really well, I really like it, great job. Glad we were able to make some nice improvements here.

The only tiny thing I can think of is
> Check you can go back and come to the initial screen with no site selected

I wonder should tapping the selected site again deselect it?

Its such a minor thing so I'm fine without it also! Thanks @iamthomasbishop  for the great suggestion on the action bar

develric:
Following up after a quick chat on slack with @osullivanchris (thanks for your time 🙇‍♂️)

> I wonder should tapping the selected site again deselect it?

Our current understanding is that it's not worth changing (I also think we could aim to some review of the app pickers) and we can proceed with the code review (pinging @khaykov and @planarvoid to let them know).

iamthomasbishop:
> I wonder should tapping the selected site again deselect it?

@osullivanchris I think I agree, although it should prob be done as a separate PR as @develric suggested.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -226,7 +226,7 @@
         HistoryListFragment.HistoryItemClickInterface,
         EditPostSettingsCallback,
         PrivateAtCookieProgressDialogOnDismissListener {
-    public static final String ACTION_REBLOG = "reblogAction";		
+    public static final String ACTION_REBLOG = "reblogAction";
```
planarvoid, line 229: I think there are no real changes to this file so you can revert it :-)

develric: yep no real changes, but AS was complaining about trailing tabs so I removed them; if you agree I would keep this, let me know.

WordPress/src/main/java/org/wordpress/android/ui/main/SitePickerActivity.java
```diff
@@ -116,6 +120,60 @@ public void onCreate(Bundle savedInstanceState) {
...
+                                    "SitePickerActivity > Selected site was null while attempting to reblog"
+                            );
+                        } else {
+                            AppLog.e(
+                                    AppLog.T.READER,
+                                    "SitePickerActivity > Selected site was null while attempting to reblog"
+                            );
+                            ToastUtils.showToast(this, R.string.site_picker_ask_site_select);
+                        }
+
```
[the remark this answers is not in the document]
develric: ehm, not actually 😊..removed.

develric:
Thanks for the review and spotting that issue @planarvoid 🙇‍♂️. Should have fixed it (bad copy paste error 🤦‍♂️).

This PR started as a "small customisation" of the site picker (as the title still says 😊) but ended up being more complex. Coming back I would probably create a dedicated picker for the reblog task but as said maybe could make sense to review the site picker component instead outside of this task. Let me know wdyt 🙇‍♂️.

PR #11711, opened 2020-04-21
Introduce kotlin to LoginFlow and bump libraries

planarvoid:
PR moved from the LoginFlow repo to WPAndroid - https://github.com/wordpress-mobile/WordPress-Login-Flow-Android/pull/36

To test:
- No need to test these changes

PR submission checklist:

### DIFF
--- a/libs/login/WordPressLoginFlow/build.gradle
+++ b/libs/login/WordPressLoginFlow/build.gradle
@@ -5,10 +5,14 @@ buildscript {
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:3.5.1'
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }
 
 apply plugin: 'com.android.library'
+apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-kapt'
+apply plugin: 'kotlin-android-extensions'
 
 repositories {
     google()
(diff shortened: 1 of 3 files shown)


libs/login/WordPressLoginFlow/build.gradle
```diff
@@ -58,21 +65,21 @@ dependencies {
     }
 
     implementation 'com.github.bumptech.glide:glide:4.10.0'
-    annotationProcessor 'com.github.bumptech.glide:compiler:4.10.0'
+    kapt 'com.github.bumptech.glide:compiler:4.10.0'
 
     // Dagger
-    implementation 'com.google.dagger:dagger:2.22.1'
-    annotationProcessor 'com.google.dagger:dagger-compiler:2.22.1'
+    implementation "com.google.dagger:dagger:$daggerVersion"
```
renanferrari, line 71: Where does `$daggerVersion` come from?

renanferrari, line 71: Where does `$daggerVersion` come from? Shouldn't we include this in the library?

libs/login/WordPressLoginFlow/build.gradle
```diff
@@ -31,17 +35,20 @@ android {
...
+    implementation 'androidx.vectordrawable:vectordrawable-animated:1.1.0'
+    implementation 'androidx.media:media:1.1.0'
     implementation 'androidx.legacy:legacy-support-v13:1.0.0'
     implementation 'androidx.gridlayout:gridlayout:1.0.0'
     implementation 'com.google.android.material:material:1.1.0'
 
+    implementation "androidx.core:core-ktx:$kotlin_ktx_version"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+
     api 'com.google.android.gms:play-services-auth:15.0.1'
```
renanferrari, line 52: I see we're not bumping `play-services-auth` anymore. This is probably the right call for now and I think we can try bumping it again once we merge the changes that affect the Google Sign-In method, so we can properly test it. What do you think?

planarvoid: I agree. I decided not to merge it after your suggestion but we can test it later.

libs/login/WordPressLoginFlow/build.gradle
planarvoid: I think it pulls the version from the default `build.gradle`. Thinking about this, it might not work cause it doesn't solve the Woo using the same library. I'll add it to the buildscript of the login library

renanferrari:
LGTM! :shipit:

PR #11721, opened 2020-04-22
Smart lock from prologue

### LINKED ISSUE #11712: Trigger smart lock from the Prologue screen during login
We want to trigger the smart lock from the Prologue screen. Since the screen will now contain only one button, we can directly open the smart lock in order to log user in (or sign up the user).

planarvoid:
Fixes #11712

This PR introduces smart lock from prologue.

I've created a flag in the BuildConfig file to hide the functionality. We can use this flag for our development. Right now it's enabled for all the versions but we should disable it once we decide to merge our code.

To test:
- Use a device that has smart lock set up
- Open the freshly installed app
- The smart lock is triggered automatically
- Select your account
- You're automatically logged in

To test:
- Use a device that has smart lock set up
- Open the freshly installed app
- The smart lock is triggered automatically
- Dismiss smart lock
- The app stays on the Prologue screen

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
@@ -177,7 +177,6 @@ public class LoginActivity extends LocaleAwareActivity implements ConnectionCall
 
         outState.putString(KEY_SMARTLOCK_HELPER_STATE, mSmartLockHelperState.name());
         outState.putBoolean(KEY_SIGNUP_FROM_LOGIN_ENABLED, mIsSignupFromLoginEnabled);
-        outState.putBoolean(KEY_SIGNUP_FROM_LOGIN_ENABLED, mIsSignupFromLoginEnabled);
     }
 
     private void showFragment(Fragment fragment, String tag) {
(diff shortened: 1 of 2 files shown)


planarvoid:
thanks for the check @renanferrari. The original PR contained 2 things. One is that when the user opens the activity without a login mode set (`FULL`) or only with wpcom login (`WPCOM_LOGIN_ONLY` - I'm pretty sure that never happens though), the app opens the smart lock. I think that part is safe.

The other part is what happens when the smart lock is dismissed. This was previously automatically starting login. I've disabled that when the unified login flow was enabled and that was the mistake (because you can get to that place even from other places than from the `FULL` login mode). The changes I'm making to fix this are introducing a flag that is true when the app opens the smart lock from the prologue screen. I only skip the previous behaviour when this flag is set. Since this flag is not set from any other login flow, everything else should work as expected. I've tested the share and jetpack sources, not sure how to do a deeplink but overall it should work now well with all the login modes.

WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
```diff
@@ -171,6 +177,7 @@ public void onSaveInstanceState(Bundle outState) {
 
         outState.putString(KEY_SMARTLOCK_HELPER_STATE, mSmartLockHelperState.name());
         outState.putBoolean(KEY_SIGNUP_FROM_LOGIN_ENABLED, mIsSignupFromLoginEnabled);
+        outState.putBoolean(KEY_SIGNUP_FROM_LOGIN_ENABLED, mIsSignupFromLoginEnabled);
```
renanferrari, line 180: This looks like a copy/paste typo maybe?

planarvoid: ah yeap, I'll remove it, thanks for the catch 👍

PR #11758, opened 2020-04-27
Issue/track media editor events

ashiagr:
This PR adds/ modifies tracks for Media Editor as follows:

1. Adds new property `number_of_images` to `media_editor_shown` to match with iOS (https://github.com/wordpress-mobile/WordPress-iOS/pull/13745) and moves tracking to `ImageEditorInitializer`'s `trackEditorAction`

2. Property `actions` in `media_editor_used` includes actions made per session to match with iOS (with an exception to track just distinct actions) and moves tracking to `ImageEditorInitializer`'s `trackEditorAction`

3. ~~Tracks new events~~   (See notes)

    ~~- General events~~
    ~~`media_editor_cancelled`~~

    ~~- Preview screen events~~
    ~~`media_editor_preview_image_selected`~~
    ~~`media_editor_preview_insert_images_clicked`~~
    ~~`media_editor_preview_crop_menu_clicked`~~

    ~~- Crop screen events~~
    ~~`media_editor_crop_opened`~~
    ~~`media_editor_crop_done_menu_clicked`~~
    ~~`media_editor_crop_successful`~~

Notes: 
- We decided on Slack to track only the events we track for iOS (with fix for distinct actions), and consider tracking other events if we decide to include them for iOS in the future.

- As we don't need to track different actions within the library, it might be better to close this PR and make simple changes to existing tracking logic as done[ in this branch](https://github.com/wordpress-mobile/WordPress-Android/tree/issue/fix_tracks_media_editor_phase_2).

UPDATE: We decided on Slack to keep this PR.

- This PR is still in draft as we need to change source to media editor when images are inserted from media editor. Currently when images are inserted from the media editor, "add media" tracking is done from inside `optimizeMediaAndTrackEvent` called from `addNewMediaToEditorAsync`

https://github.com/wordpress-mobile/WordPress-Android/blob/0387bc62b2d8f57a256313c31f24f72721a72ed9/WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/OptimizeMediaUseCase.kt#L59

The source is calculated in this method based on whether media is a photo/video and if it is freshly taken.

https://github.com/wordpress-mobile/WordPress-Android/blob/8614ce8452134ecfc0e9e67e86fa4bdcd2253af9/WordPress/src/main/java/org/wordpress/android/ui/posts/editor/EditorTracker.kt#L28-L55

Wondering if we should introduce another flag `isFromMediaEditor` through chain of methods till this point or remove `isNew`, `isVideo` flags and pass exact `Stat` source impacting several files.

UPDATE: We decided on Slack to track `media_editor` source in addition to source tracked above. Changes included here: 317adbd

PR submission checklist:


libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/crop/CropFragment.kt
```diff
@@ -97,7 +104,8 @@ class CropFragment : Fragment(), UCropFragmentCallback {
                     is ImageCropAndSaveFailedState -> {
                         showCropError(state.errorMsg, state.errorResId)
                     }
-                    is ImageCropAndSaveSuccessState -> { // Do nothing
+                    is ImageCropAndSaveSuccessState -> {
+                        ImageEditor.instance.onEditorAction(CropSuccessful(state.cropResult))
```
malinajirka, line 108: I think ideally all the actions would be invoked from a VM (it won't be possible for some of them, but it should hopefully work for most). I'd suggest passing ImageEditor into the `vm.start()` method so we don't access static class from a VM.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/EditImageActivity.kt
```diff
@@ -55,4 +56,11 @@ class EditImageActivity : AppCompatActivity() {
             navController.navigateUp(appBarConfiguration) || super.onSupportNavigateUp()
         }
     }
+
+    override fun onBackPressed() {
+        super.onBackPressed()
+        if (hostFragment.childFragmentManager.backStackEntryCount == 0) {
```
malinajirka, line 62: I think this is one of the exceptions which won't be easy to move to a VM. I'll keep thinking about other possible solution as this feel a bit like hack.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/ImageEditor.kt
```diff
@@ -52,22 +57,56 @@ class ImageEditor private constructor(
...
+    sealed class Action(val label: String) {
+        object Crop : Action("crop")
+    }
+
     companion object {
         private lateinit var INSTANCE: ImageEditor
 
         val instance: ImageEditor get() = INSTANCE
 
+        // The actions that were made in this session.
```
malinajirka, line 94: I'm wondering if we could move the responsibility of keeping track of the actions to the client. We already emit `CropSuccessful` so the client can easily track the actions, right? Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/ImageEditorInitializer.kt
```diff
@@ -70,5 +88,70 @@ class ImageEditorInitializer {
...
+
+        private fun trackAddMediaFromMediaEditor(outputDataList: List<OutputData>) {
+            for (outputData in outputDataList) {
+                val properties = AnalyticsUtils.getMediaProperties(
+                    WordPress.getContext(),
+                    false,
+                    Uri.parse(outputData.outputFilePath),
+                    null
+                )
+                properties["via"] = AnalyticsTracker.track(Stat.EDITOR_ADDED_PHOTO_VIA_MEDIA_EDITOR, properties)
```
malinajirka, line 153: I'm either unfamiliar with this syntax or this code does something else than you intended. I'd expect it saves `Unit` into `properties["via"]` after we track `EDITOR_ADDED_PHOTO_VIA_MEDIA_EDITOR`. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/ImageEditorInitializer.kt
```diff
@@ -70,5 +88,70 @@ class ImageEditorInitializer {
...
+                AnalyticsTracker.track(stat, properties)
+            }
+
+            // Also track "media_editor" source with media properties
+            if (action is EditorFinishedEditing) {
+                trackAddMediaFromMediaEditor(action.outputDataList)
+            }
+        }
+
+        private fun trackAddMediaFromMediaEditor(outputDataList: List<OutputData>) {
```
malinajirka, line 145: Do we want to track this event even when someone uses the ImageEditor in a different flow - such as "Choose Avatar" in the "Me" section? I'd consider tracking this event within the EditPostActivity as it seems unrelated to the ImageEdtior itself. Wdyt?

malinajirka (changes requested):
Thanks @ashiagr! I really like the OnEditorAction approach - I think it'll be useful for different things than just tracking in the future. I haven't tested the app yet, since I've left some comment which might result in bigger changes. Let me know what you think. Thank you!

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/ImageEditorInitializer.kt
```diff
@@ -70,5 +88,70 @@ class ImageEditorInitializer {
             } else {
                 throw(IllegalArgumentException(IMAGE_STRING_URL_MSG))
             }
+
+        private fun onEditorAction(): (EditorAction) -> Unit = { action ->
+            trackEditorAction(action)
```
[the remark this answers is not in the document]
ashiagr: Sure, will do that 👍.

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/ImageEditorInitializer.kt
ashiagr: My bad, will fix.

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/ImageEditorInitializer.kt
ashiagr: Very futuristic of you! Will do as suggested 😀.

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/crop/CropFragment.kt
ashiagr: As discussed, will try it out.

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/ImageEditorInitializer.kt
```diff
@@ -70,5 +88,70 @@ class ImageEditorInitializer {
...
+                is CropDoneMenuClicked,
+                is CropSuccessful -> null
+            }
+
+            val noEditActionPerformed = stat == Stat.MEDIA_EDITOR_USED && properties == null
+            if (stat == null || noEditActionPerformed) {
+                return
+            }
+
+            if (properties == null) {
```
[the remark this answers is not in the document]
ashiagr: Changed here: 81ea437

libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/ImageEditor.kt
ashiagr: Done here: 4760c7f

ashiagr:
Thanks for a great review @malinajirka 🙏.
I've tried to incorporate your feedback except for these two which I plan to do in a separate PR.

> I'd consider extracting the trackEditorAction into its own injectable class

> Do we want to track this event even when someone uses the ImageEditor in a different flow - such as "Choose Avatar" in the "Me" section?

Ready for another pass!

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/ImageEditorInitializer.kt
```diff
@@ -70,5 +88,70 @@ class ImageEditorInitializer {
...
+            }
+
+            val properties = when (action) {
+                is EditorShown -> mapOf(NUMBER_OF_IMAGES to action.numOfImages)
+                is EditorFinishedEditing -> {
+                    val distinctActions = action.actions.distinct()
+                    if (distinctActions.isNotEmpty()) {
+                        mapOf(ACTIONS to distinctActions.map { it.label })
+                    } else {
+                        null
```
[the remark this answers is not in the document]
ashiagr: Changed here: b423fdd

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/ImageEditorInitializer.kt
malinajirka: Tbh I usually don't plan for something that might happen but probably never will. But in this case it felt more like fixing the SRP than planning for future. ;)

malinajirka:
Thanks @ashiagr! The PR looks good overall.

We just need to implement these changes in a separate PR:
- [ ] I'd consider extracting the trackEditorAction into its own injectable class

- [ ] Do we want to track this event even when someone uses the ImageEditor in a different flow - such as "Choose Avatar" in the "Me" section?

We also need to decide if we want to keep tracking `editor_media_added` twice with different `via` parameter, when the media editor is used.

PR #11766, opened 2020-04-28
Bump libraries to support remote config

planarvoid:
This PR bumps three libraries in order to implement firebase remote config library. I've checked the change log of these libraries and everything should be in order. The only breaking change was moving to AndroidX (which we did already) and a different way to handle InstanceID (which I implemented in this PR). 

To test:
- Smoke test the Google login and the app

PR submission checklist:


malinajirka (approved):
LGTM, thanks!

PR #11767, opened 2020-04-28
Prepublishing Nudges : Home Screen UI Update 

### LINKED ISSUE #11697: Prepublishing Nudges : Adjust home screen ui to latest design
Update the clickable actions in the UI to this design. 
 
Current  | Android Design (Updated today)
--------|-------
|    Screen Shot 2020-04-13 at 10 00 57 PM

jd-alexander:
Fixes #11697

## Findings
The skeleton bottom sheet that was created at the start of this project contained the core components required to continue the UI improvements so it was extended to support the header & publish button as seen in the design here. 
Screen Shot 2020-04-28 at 3 43 19 PM

## Solution

In this solution the main changes that were made are 

1. The Header UI was added and it is powered by data from the `SiteModel` associated with the `EditPostRepository`. Within the design above, the site image is rounded, so the `ImageManager` was modified to include functionality that performs rounded corners transformation on images. 

2. The Publish Button was added  5f4adffc4cd38bfc820e43179cdcfb05c863c569  2f7c8c8ccd697a3690f1e2c57fee791f50a91dab  8a999e6bdbc993fd3a3cd211b84c2c500f4c5287 . Currently, it doesn't actually do anything as that will be done in another PR https://github.com/wordpress-mobile/WordPress-Android/issues/11604

3. The `BottomSheet` height was made dynamic since it needs to adapt to several different screen sizes and the UI designs stipulate that it takes up only a portion of the screen real estate so there's logic that calculates the height based on the screen size & orientation. 386a146c4517a8e87cbc0d50f961de17bc9157f0

## Testing

1. Create a new post on press publish. 
2. You should see the bottom sheet. 
3. Ensure that you utilize multiple device types and orientations. 

#### Portrait 

Light Mode | Dark Mode 
--------|-------
|   

#### Landscape

Light Mode | 
--------| 
Screenshot_1588107460
Dark Mode |  
Screenshot_1588107448

### Portrait

Light Mode | Dark Mode 
--------|-------
Screenshot_1588107873| Screenshot_1588105842

### Landscape

Light Mode | 
--------| 
Screenshot_1588108109
Dark Mode | 
Screenshot_1588108119

### Portrait

Light Mode | Dark Mode 
--------|-------
Screenshot_1588108360| Screenshot_1588108372

### Landscape

Light Mode | 
--------| 
Screenshot_1588110686
Dark Mode | 
Screenshot_1588110696

## Reviewing 

## Submitter Checklist


WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingBottomSheetFragment.kt
```diff
@@ -80,6 +84,54 @@ class PrepublishingBottomSheetFragment : WPBottomSheetDialogFragment(),
         }
     }
 
+    private fun setupFragmentContainerHeightBasedOnDevice() {
```
jd-alexander, line 87: @malinajirka At first, this layout had no defined height. I utilized `wrap_content` for the `fragment container` and it worked well until I began navigating to other screens that ended up having different heights from the home screen. So navigating from one screen to the next would include a height change and cause a flicker and design inconsistency. 

1. I decided to utilized the drawn height of the `fragment container` when the home screen is drawn and set that as the permanent height overriding the `wrap_content` this worked well but when the device rotated/ app killed it was lost, so I decided to persist this height. Overall, I didn't like that solution and there were some times when the height would still be reset due to different view configuration changes so I opted against it. 

2. I then tried simply setting a single height on the bottom sheet but this caused the UI to look okay on some screens and not properly sized on others so I decided against it. 

3. So below, is the solution that seemed most workable for the different screen sizes. The height of the container is calculated dynamically based on the orientation and device. The main issue I have noticed is that sometimes there's a bit of extra padding at the bottom because the formula might give a bottom sheet a bit more height than it needs. 

4. Also, I was wondering if I should be utilizing screen-specific layouts to host this logic as that would make it clearer how the UI adapts to each screen.  I opted from doing this, to gain feedback. 

Let me know what your thoughts on this are.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
```diff
@@ -24,17 +29,18 @@ class PrepublishingHomeViewModel @Inject constructor(
...
+        setupHomeUiState(editPostRepository, site)
     }
 
     // TODO remove hardcoded Public with live data from the EditPostRepository / user changes.
-    private fun setupHomeUiState(editPostRepository: EditPostRepository) {
+    private fun setupHomeUiState(editPostRepository: EditPostRepository, site: SiteModel) {
         val prepublishingHomeUiStateList = listOf(
-                PrepublishingHomeUiState(
+                HeaderUiState(UiStringText(site.name), site.iconUrl?.let { it } ?: run { "" }),
+                HomeUiState(
```
jd-alexander, line 43: Let me know if this logic is clear or it needs to be reworked for clarity. Basically, if the icon URL is `null` I am passing an empty string so that the `ImageManager` can fallback to a placeholder.

WordPress/src/main/java/org/wordpress/android/util/image/ImageManager.kt
```diff
@@ -148,6 +151,27 @@ class ImageManager @Inject constructor(private val placeholderManager: ImagePlac
...
+        imgUrl: String,
+        roundingRadius: Pixels
+    ) {
+        val context = imageView.context
+        if (!context.isAvailable()) return
+        GlideApp.with(context)
+                .load(imgUrl)
+                .addFallback(imageType)
+                .addPlaceholder(imageType)
+                .transform(RoundedCorners(roundingRadius))
```
jd-alexander, line 170: Something I noticed while doing this implementation, is that the `rounded corners` transformation is not applied to placeholders. I I checked the `Glide` github repo and they say it's not supported. Should I utilize an XML approach to perform the rounding required based on the designs? 

I just realized that elsewhere in the app in particular `MySiteFragment`, a material card view is being utilized to accomplish this so I might opt for that approach instead. 

```xml
<com.google.android.material.card.MaterialCardView
                        android:layout_width="@dimen/my_site_blavatar_sz"
                        android:layout_height="@dimen/my_site_blavatar_sz"
                        android:layout_margin="@dimen/margin_small"
                        app:cardCornerRadius="@dimen/my_site_blavatar_corner_radius"
                        app:cardElevation="0dp">
```

Let me know 😉

jd-alexander:
In terms of the Large Tablet UI, I noticed that the portrait mode's bottom sheet looks skinny in comparison to landscape. I wanted to get feedback on the sizing that should take place if it should be similar to landscape since it has the screen space. Let me know.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeListItemViewHolder.kt
```diff
@@ -3,30 +3,88 @@ package org.wordpress.android.ui.posts
...
-    fun bind(uiState: PrepublishingHomeItemUiState) {
-        uiState as PrepublishingHomeUiState
-
-        actionType.text = uiHelpers.getTextOfUiString(itemView.context, uiState.actionType.textRes)
-        uiState.actionResult?.let { resultText ->
-            actionResult.text = uiHelpers.getTextOfUiString(itemView.context, resultText)
+private const val ROUNDING_RADIUS_DP = 6
+
+sealed class PrepublishingHomeViewHolder(internal val parent: ViewGroup, @LayoutRes layout: Int) :
+        RecyclerView.ViewHolder(
```
malinajirka, line 23: NitPick: The class declarations in this file have inconsistent spacing/line-breaks. Could you please make them consistent - ideally put them on one/two lines (they should all fit on max two lines with the 120 chars line limit)

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
malinajirka: I wouldn't mind this logic, but since we have a utility method just for this purpose, I'd consider using it - StringUtils.notNullStr(). Imo it's okay to use a static access for this purpose since we don't want to mock this type of utility method anyway.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
```diff
@@ -43,16 +49,18 @@ class PrepublishingHomeViewModel @Inject constructor(
...
-                PrepublishingHomeUiState(
+                HomeUiState(
                         actionType = TAGS,
-                        actionResult = getPostTagsUseCase.getTags(editPostRepository)?.let { UiStringText(it) },
+                        actionResult = getPostTagsUseCase.getTags(editPostRepository)?.let { UiStringText(it) }
+                                ?: run { UiStringRes(R.string.prepublishing_nudges_home_tags_not_set) },
                         onActionClicked = ::onActionClicked
-                )
+                ),
+                PublishButtonUiState(UiStringRes(R.string.prepublishing_nudges_home_publish_button), {})
```
malinajirka, line 63: I assume we are passing an empty onClickListener since this logic isn't implemented yet, right?

WordPress/src/main/java/org/wordpress/android/util/image/ImageManager.kt
malinajirka: I'd consider one of the following approaches
1. Utilize the MaterialCardView - I personally don't see any downside to it and it seems like the simplest solution.
2. Add a new optional sealed class paramater to `fun load(imageView: ImageView, imageType: ImageType, imgUrl: String = "", scaleType: ScaleType = CENTER)`.  Similarly to `ScaleType`. Unfortunately, as you mentioned this won't work for placeholders :(.

Wdyt?

WordPress/src/main/res/values/strings.xml
```diff
@@ -2745,9 +2745,13 @@
     <string name="prepublishing_nudges_publish_action">Publish</string>
     <string name="prepublishing_nudges_visibility_action">Visibility</string>
     <string name="prepublishing_nudges_tags_action">Tags</string>
+    <string name="prepublishing_nudges_home_header_publishing_to">PUBLISHING TO</string>
```
malinajirka, line 2748: I'd consider using `android:textAllCaps="true"` in the xml layout. This approach might not work as expected with GlotPress. Wdyt?

malinajirka (commented):
Thanks @jd-alexander! Really nice changes overall - both UI and the code ;).

I've left some comments. I'm still not sure what is the best solution for the bottomSheet height. However, my testing device looked like this, so I'm a bit worried the custom solution might be a bit risky to use. Wdyt? 



SideNotes: 
1. Shouldn't the bottom sheet have rounded corners or do you plan to solve it in another PR?
2. The bottomSheet can't be fully expanded and always prefers scrolling the content. 
This is comparison of the behavior between the editor and reader

| First Header  | Second Header |
| ------------- | ------------- |
| editor | reader |

3. Last thing - this part scrolls with the content. Is this desired behavior?
Screenshot 2020-04-30 at 15 33 24

mbshakti:
Overall the UI is looking fantastic! 

I've got some nits, documented below:

- Can we reduce the space between [publish button and the bottom to 16pt](https://drive.google.com/file/d/10RqYa0cHjVUHDCMpu1ypvlYVZfNcTHAb/view?usp=sharing)

- Can we add some padding to the [top of the bottom sheet header to 16pt](https://drive.google.com/file/d/1UBf4rjjRbGC1I7pABmVyaNGlHxPvBJxL/view?usp=sharing). I know in the Figma designs this padding doesn't exist, but after playing around with it, I think it would nice to have more padding up there.

- For the input box on the add tags screen, can we auto focus on the input box with the keyboard showing?

- Lastly, a bug report: When I tap visibility, the app crashes for me. [Screenshot here.](https://drive.google.com/file/d/1eDTsZPh6guoHX0nHrX8oUnrQlKgvOYYB/view?usp=sharing)

Let me know if I can clarify anything.

jd-alexander:
Thanks for the review @mbshakti I will make these changes. 

> Lastly, a bug report: When I tap visibility, the app crashes for me. Screenshot here.

Oh yes, this is because the feature wasn't implemented in this PR and is available here https://github.com/wordpress-mobile/WordPress-Android/pull/11796 I didn't realize I didn't mention 😅

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeListItemViewHolder.kt
jd-alexander: Okay, thanks! Is there any way we can modify the auto-formatting to do this? I will take note of this now though but sometimes when working, I just autoformat and move on.

jd-alexander: Thanks! I had missed this 🙏

jd-alexander:
Resolving the action order issue https://github.com/wordpress-mobile/WordPress-Android/issues/11775 here since work is being done the home screen.

jd-alexander:
Thanks for the review @malinajirka I implemented all the changes you requested. I wasn't able to get the shared element transitions working well so we could probably jump on a call to go through that exercise to find a solution to the height issue 🙏

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingBottomSheetFragment.kt
jd-alexander: Thanks! I tried the shared element and I didn't record a gif because it was really bad 😅 For now, I removed this logic and set a minimum height 2deda1d c2c9837

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
jd-alexander: Oh nice! Never knew that existed 🙏 
7759cac

WordPress/src/main/java/org/wordpress/android/util/image/ImageManager.kt
jd-alexander: Thanks! I think the `CardView` is the simplest approach so I opted for it.
51ac8d9

WordPress/src/main/res/values/strings.xml
jd-alexander: Good eye 😉 
532bac6

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
jd-alexander: Yes, this is being done here https://github.com/wordpress-mobile/WordPress-Android/pull/11805

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeListItemViewHolder.kt
```diff
@@ -3,30 +3,88 @@ package org.wordpress.android.ui.posts
...
-    private val actionResult: TextView = itemView.findViewById(R.id.action_result)
-    private val actionLayout: View = itemView.findViewById(R.id.action_layout)
-
-    fun bind(uiState: PrepublishingHomeItemUiState) {
-        uiState as PrepublishingHomeUiState
-
-        actionType.text = uiHelpers.getTextOfUiString(itemView.context, uiState.actionType.textRes)
-        uiState.actionResult?.let { resultText ->
-            actionResult.text = uiHelpers.getTextOfUiString(itemView.context, resultText)
+private const val ROUNDING_RADIUS_DP = 6
```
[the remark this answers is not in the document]
jd-alexander: Removed this. 🙏

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeFragment.kt
```diff
@@ -68,7 +69,16 @@ class PrepublishingHomeFragment : Fragment() {
             }
         })
 
-        viewModel.start(getEditPostRepository())
+        viewModel.start(getEditPostRepository(), getSite())
+    }
+
+    private fun getSite(): SiteModel {
+        val editPostActivityHook = requireNotNull(getEditPostActivityHook()) {
+            "This is possibly null because it's " +
```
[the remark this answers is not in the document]
jd-alexander: No, it should never crash, you are right. Thanks for spotting this. I updated the message.
85153ba

WordPress/src/main/java/org/wordpress/android/util/image/ImageManager.kt
malinajirka: 👍I believe we can remove `loadWithRoundedCorners` from the ImageManager now, right? Wdyt?

malinajirka (approved):
Thanks @jd-alexander for the changes! I agree with all of your comments. I'm approving this PR. Could you please turn the minor remaining issues into github tickets. Thanks!

WordPress/src/main/java/org/wordpress/android/util/image/ImageManager.kt
jd-alexander: yes, I thought I had removed that 😅

Issue #11769, opened 2020-04-28
What's new - Implement full screen dialog with list of new features


khaykov:
We can split this task into the dialog and adapter counterparts.

Should look something like this + progress indicator.

![whats-new-i3-android](https://user-images.githubusercontent.com/728822/80548267-63628e80-896f-11ea-9c78-fd0d2bda7dc2.png)

develric:
closed by #11770 and #11772

PR #11772, opened 2020-04-29
Issue/11769 feature announcement dialog list

khaykov:
Part of #11769
Based on #11770 which needs to be merged first.

Logic in `FeatureAnnouncementProvider` is pretty basic for now. 

I need to polish this a bit more for code review, but the design could be reviewed right now.

[Image from Gyazo](https://gyazo.com/c73e5f150c9eb23550318981a2bbb4cc)

[Image from Gyazo](https://gyazo.com/eed5605883b019fd9ae8fd2b8b2b4685)

To test:
- Open Feature Announcement dialog from App Settings.
- Make sure the list looks ok in light/dark modes and in portrait/landscape orientations.
- Make sure `FIND OUT MORE` buttons opens wp blog.

PR submission checklist:


khaykov:
@osullivanchris Thanks for the quick review! 

> I had the subtitle of each item at medium weight rather than regular. It doesn't actually offer medium as an option for Subtitle1 here. So do you think we should change it per the mock, or is it not good practice?

Personally, I'm for a cohesive look of the app before anything, and not a fan of similar parts of the app (list elements) looking differently through it :) I think that If we want Subtitle1 were to use medium in lists, we should change it in all the lists. But it's up to you!

> @mattmiklic said he's working towards not using the shadow on the primary buttons anymore. But not sure if this is a component change or something we have to do 1 by 1. But just flagging it to be changed at some point

This is done on the component level, so no 1 by 1 changes :)

mattmiklic:
> @mattmiklic said he's working towards not using the shadow on the primary buttons anymore. But not sure if this is a component change or something we have to do 1 by 1. But just flagging it to be changed at some point

Yeah, I will see if we can implement this while we're doing the other visual refresh items included in Unified Login & Signup.

osullivanchris:
> Personally, I'm for a cohesive look of the app before anything, and not a fan of similar parts of the app (list elements) looking differently through it :) I think that If we want Subtitle1 were to use medium in lists, we should change it in all the lists. But it's up to you!

That's a fair point! I kinda freewheeled this one and didn't use a list component as a starting point. So it went over my head that I basically designed a list item! In that case, I agree with you. Lets keep it as you have it.

> This is done on the component level, so no 1 by 1 changes :)

Sweet! Ok we don't need to address it specifically in this PR then!

osullivanchris:
One small thing after getting my hands on the build. I can't tell for sure, as the content isn't long enough to scroll. But the glow indicating the scrollable area seems to indicate to me that just the list is scrollable, and the title/header might be fixed as well as the button/footer? I don't think we need to fix the header. On small devices if the header didn't scroll away, it would just make the list squeezed into a small space.

On the other hand, I may be misinterpreting. Maybe both components have separate scroll areas, and the header isn't fixed. 

Let me know! 😄

mattmiklic:
>  I don't think we need to fix the header. 

I agree; I just noticed that it looks like it's fixed, which requires a lot of unnecessary scrolling in landscape. This would be functionally similar to the new epilogue, where the large title area (avatar + display name + username) scrolls along with the list of sites.

WordPress/src/main/java/org/wordpress/android/ui/whatsnew/FeatureAnnouncementViewModel.kt
```diff
@@ -1,46 +1,73 @@
...
-            _uiModel.value = _uiModel.value?.copy(isProgressVisible = false)
+            _currentFeatureAnnouncement.value = featureAnnouncementProvider.getLatestFeatureAnnouncement()
         }
     }
 
     fun onCloseDialogButtonPressed() {
         _onDialogClosed.call()
     }
 
+    fun onFindMoreButtonPressedPressed() {
```
develric, line 65: is this double `Pressed` intentional or a copy/paste thing?

khaykov: This is Copy/PastePaste thing :) Fixed!

PR #11780, opened 2020-04-30
Unified Login & Sign-Up: login with site address from the Prologue screen

### LINKED ISSUE #11736: Unified Login & Sign-Up: Make it possible to login with a site address from the Prologue screen

With the unified flow, we won't need to have a dedicated **Sign Up** button anymore, so we want to remove it from the Prologue screen and replace it with a button that make it possible for the user to login with a site address.

Besides that, we need to make sure the option to enter with a site address is still available from the Email screen when coming from other flows where the Prologue screen is not shown.

renanferrari:
Fixes #11736

### Instructions
- Make sure the `UNIFIED_LOGIN_AVAILABLE` flag is set to `true`.

### Scenarios

**Site Address From Prologue**

0. Clear app data.
1. Open the app and notice the Prologue Screen.
2. Notice how the top button label is **Continue with WordPress.com**.
3. Notice how the bottom button label is **Enter your site address**.
4. Notice how the top button has the primary style.
5. Tap **Enter your site address**.
6. Notice the Site Address Screen.
7. Complete the login flow normally.

**Email Screen Without Site Option**

0. Clear app data.
1. On the Prologue Screen, tap **Continue with WordPress.com**.
2. On the Email Screen, notice how the option to login with a site address is not available.

**Landscape Orientation**

0. Clear app data.
1. On the Prologue Screen, rotate the device into landscape orientation.
2. Notice how the left button label is **Enter your site address**.
3. Notice how the right button label is **Continue with WordPress.com**.
4. Notice how the right button has the primary style.
5. Tap **Continue with WordPress.com**.
6. On the Email Screen, notice how the option to login with a site address is not available.

**Share Intent Flow**

0. Clear app data.
1. From outside the app, try to share some media with WordPress.
2. If the Smart Lock dialog appears, dismiss it.
3. On the Email Screen, notice how the option to login with a site address is still available.

## Notes
- With the `UNIFIED_LOGIN_AVAILABLE` flag set to `false`, everything should behave as before.
- Once we get rid of the feature flag, we won't need to maintain both `login_prologue_bottom_buttons_container_default` and `login_prologue_bottom_buttons_container_unified` layouts.

---

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginPrologueFragment.java
```diff
@@ -25,19 +28,32 @@
...
-                    mLoginPrologueListener.showEmailLoginScreen();
-                }
+        MaterialCardView bottomButtonsCard = view.findViewById(R.id.bottom_buttons);
+
+        if (BuildConfig.UNIFIED_LOGIN_AVAILABLE) {
+            bottomButtonsCard.removeAllViews();
+            inflater.inflate(R.layout.login_prologue_bottom_buttons_container_unified, bottomButtonsCard);
+        }
+
+        MaterialButton primaryButton = view.findViewById(R.id.login_button);
```
planarvoid, line 38: What do you think about changing the IDs of these two buttons to something more generic? I understand `primary` and `secondary` doesn't really work here since the order is switched between the two designs but maybe something about the order? `first_button` and `second_button`? It might make it a little more readable for the future.

planarvoid (approved):
Thanks for the changes, it looks good and works well 👍 Consider my suggestion about naming of the buttons. I think the current state is a bit confusing

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginPrologueFragment.java
renanferrari: Good point! Since these IDs had not been updated since the creation of that screen, I thought it might have some not-so-obvious reason for that and decided not to touch them. Upon closer inspection, I couldn't find any good reason not to change them, so I'll just go with your suggestion 🙂

Thanks for reviewing this! 🙇‍♂️

renanferrari: > maybe something about the order? `first_button` and `second_button`?

Oh, I just realized that this could make things look weird on the landscape layout, since the order kind of changes there (the top button becomes the right button and the bottom button becomes the left one), meaning we would have the `first_button` on the right and the `second_button` on the left. This might still be a better alternative though. Let me know what you think.

renanferrari: After talking to @mattmiklic about this, we reached the conclusion that it's fine to have the natural order of the buttons on landscape for the new layouts, instead of having them reversed. So I'm renaming the buttons to what you've sugggested and updating the landscape layout for the new layouts, but I'm keeping the old ones as they were. That means we will still have one layout where the `second_button` is coming first, but that will be solved sooner than later, when we get rid of the feature flag and stop maintaining the old flow.

renanferrari: Here's a visual before and after:

**Before**
| Portrait | Landscape |
|---|---|
| old-portrait | old-landscape |

**After**
| Portrait | Landscape |
|---|---|
| new-portrait | new-landscape |

mattmiklic: Looks good to me!

PR #11788, opened 2020-04-30
Adding reblog dedicated tracking events.

develric:
This PR adds specific tracking events for reblog.

The dedicated events added are:

**reader_article_reblogged** -> triggered when reblogging from a list of posts in the reader (main unfiltered reader or a filtered list obtained by using the subfilter bar or tapping on a post site title)
**reader_article_detail_reblogged** -> triggered when reblogging from a reader opened post details

We were already traking in `editor_post_created` event `"created_post_source_detail":"post-from-reader-reblog"` property and we added a dedicated `"created_post_source_detail":"post-from-reader-detail-reblog"` to discriminate between the actions described above.

NOTE: I've set the `SiteModel site` parameter in the `openEditorForReblog` to @Nullable since it can be null for example in `ReaderPostDetailFragment > onActivityResult` when we do `siteStore.getSiteByLocalId(siteLocalId)`; managed it the same as a nullable post in the function.

## To test
- Activate tracking event in app settings
- Trigger a reblog from a list of post (unfiltered or filtered)
- Check in logcat you get something like

```
Tracked: reader_article_reblogged, Properties: {"target_blog_id":157782556,"source_post_id":14871,"source_blog_id":23519412}
Tracked: editor_post_created, Properties: {"blog_id":157782556,"created_post_source_detail":"post-from-reader-reblog","post_type":"post","created_post_source":"post-list","is_jetpack":false}
```

- Trigger a reblog from a post detail
- Check in logcat you get something like

```
Tracked: reader_article_detail_reblogged, Properties: {"target_blog_id":157782556,"source_post_id":2682331778,"source_blog_id":50586720}
Tracked: editor_post_created, Properties: {"blog_id":157782556,"created_post_source_detail":"post-from-reader-detail-reblog","post_type":"post","created_post_source":"post-list","is_jetpack":false}
```

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
```diff
@@ -302,19 +304,42 @@ public static void openEditorForSiteInNewStack(Context context, @NonNull SiteMod
...
+
+        if (site == null) {
+            ToastUtils.showToast(activity, R.string.blog_not_found, ToastUtils.Duration.SHORT);
+            return;
+        }
+
+        Map<String, Object> properties = new HashMap<>();
+        properties.put("source_post_id", post.postId);
+        properties.put("source_blog_id", post.blogId);
+        properties.put("target_blog_id", site.getSiteId());
```
planarvoid, line 326: the `AnalyticsUtils` class has some of the fields used here. Do you think it might make sense to create a utility method there and reuse some of the properties? Maybe it doesn't but I'd still consider extracting it there and using the constants instead of these strings.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -2921,7 +2922,12 @@ private void handleReblogStateChanges() {
                     );
                 } else if (state instanceof PostEditor) {
                     PostEditor peState = (PostEditor) state;
-                    ActivityLauncher.openEditorForReblog(getActivity(), peState.getSite(), peState.getPost());
```
planarvoid, line 2924: I'd rename the `peState` to something better :-)

develric: tbh was pretty in doubt on how to rename it without changing the variable semantic 😅...ended up with `postEditorStateData` 😬.

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
develric: Good suggestion thanks 🙇‍♂️! I moved to the Utils but kept the "source_*" and "target_*" prefix to be more clear on the intended usage; hope it's fine.

develric:
Hey 👋! Made few changes based on your review, thanks for it and the good suggestion 👍! Ready for another pass 🙇‍♂️!

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
planarvoid: sounds good :D it's just that I haven't seen these single letter variables since the times of C++

PR #11795, opened 2020-04-30
Reblog to wpcom sites only.

develric:
This PR does the following:

1. Limits the site that can be selected to reblog to and those that are presented and searchable in the site picker while reblogging to be those that are accessed through the WP.com API (WP.com/JP)
2. NoSiteToReblogActivity: adds back the toolbar wrongly missing after a merge from develop and aligns copy to iOS.

## To test
Setup an user that has in the visible sites at least 1 pure self-hosted site (added by address), at least 1 JP and at least 1 WP.com (also 1 Atomic could be good 😊)
Note: all the below reblog tests should be made from post lists (filtered or from main reader) and from a post details.

- keep all the sites available visible (let's assume something as below with the last one being the self-hosted)
<p align="center">
  
</p>

- reblog and check the site picker is presented as below without the self-hosed

<p align="center">
  
</p>

- try to search and check the self-hosted is not present in the searched list

<p align="center">
  
</p>

- select a site and check the reblog completes correctly.
- Now let's make not visible 1 of the not self-hosted sites

<p align="center">
  
</p>

- Even if the self-hosted site is the currently selected one, when you reblog you should be brought in the editor reblogging to the only visible not self-hosted site. Complete the reblog and check it is processed correctly.

- Now let's keep visible only the self-hosted sites

<p align="center">
  
</p>

- When you reblog you should be brought to the following screen

<p align="center">
  
</p>

- Check the screen has a working toolbar and the copy is as reported in the image
- Pressing the `Manage Sites` button you should be brought to the `My Site` page

### Test case 2 - Smoke test
Smoke test the Reblog functionality; also note the Site Picker is used in `My Site > Switch Site`, `Share to WP`, `Login Epilogue` so Smock test these scenarios to check for eventual regressions. In particular check the Site Hide feature and search in Switch Site picker work as usual.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/modules/ViewModelModule.java
+++ b/WordPress/src/main/java/org/wordpress/android/modules/ViewModelModule.java
@@ -282,7 +282,6 @@ abstract class ViewModelModule {
     @ViewModelKey(SitePickerViewModel.class)
     abstract ViewModel sitePickerViewModel(SitePickerViewModel viewModel);
 
-
     @Binds
     abstract ViewModelProvider.Factory provideViewModelFactory(ViewModelFactory viewModelFactory);
 }
--- a/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
@@ -265,6 +265,12 @@ public class ActivityLauncher {
         context.startActivity(intent);
     }
 
+    public static void viewMySiteInNewStack(Context context) {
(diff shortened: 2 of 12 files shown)


develric:
Hey @khaykov , good catch, thanks 🙇‍♂️! I removed `ReaderEvents.SiteManagerTriggered` and introduced the `viewMySiteInNewStack` to trigger a restart of the WPMainActivity. In this way we should better manage also another case I was spotting while retesting it now, that's when we try to reblog from an inner post list or post detail and happen to get to this Manage Sites scenario. Let me know wdyt 😊.

PR #11796, opened 2020-04-30
Prepublishing Nudges : add visibility option

### LINKED ISSUE #11605: Prepublishing Nudges: Add Visibility option
Add the ability to select either public, private or password protected on a post when it's being published.

jd-alexander:
Fixes #11605

## Findings
Visibility changes allow the user to make their post private, public or password protected. This behavior is connected to the `PostStatus` and the post's `password` field. 

## Solution
The business logic around this implementation goes as follows : 

### Getting Visibility 

1. First I checked to see if the password isn't empty. If it isn't then the `Visibility` is set to `PASSWORD_PROTECTED`. 
2. I then checked to see if the `PostStatus` is equal to `PRIVATE` and if so the `Visibility` is `PRIVATE`. 
3. Based on the design, my assumption was that any other `PostStatus` that is currently set would default to the `PUBLIC` visibility since the post will be accessible with them. 

### Saving Visibility 
1. To transform from a `Visibility` to a `PostStatus` I checked to see if the `Visibility` is set to `PRIVATE` and if so the `PostStatus` would be the same. 
2. For the `PUBLIC` `Visibility` I simply set the `PostStatus` to `DRAFT`. 

### Updating Password

1. To update the password, I utilized the `editPostRepository` and passed the value received from the password dialog used to get the user's input. 
2. In the implementation & tests, I am using an empty string to represent no password, and this is also used to clear it.

### Concerns

1. I noticed that changing the `Visibility`/`PostStatus` will lead to the `PrimaryAction` in the editor being updated and when this happens the action there doesn't trigger the bottom sheet. 

2. The password dialog doesn't have any validation on the `EditText`That is outside the scope of this PR, but something that could be fixed. 

## Testing

1. Create a new post. 
2. Click on `Visibility` in the bottom sheet. 
3. Toggle the private & public visibility options in the list. 
4. Ensure that this is reflected in the `Post Settings` Status.

Public | Private 
--------|-------
<kbd></kbd>   |  <kbd></kbd>
------------- 

1. Clicking on the password option should trigger a password dialog. 
2. Once you set a password then the `Password Protected` option should be selected.  
3. Go to the `Post Settings` and ensure that the password value was persisted. 
4. You can tap on the `Password Protected` option and clear the password and the visibility will return to `Public`
5. You can verify that the password was successfully cleared in the Post Settings. 

Before | After 
--------|-------
 <kbd></kbd>  | <kbd></kbd>
 ------------- 

## Reviewing 

## Submitter Checklist


jd-alexander:
@ashiagr I changed the base branch to https://github.com/wordpress-mobile/WordPress-Android/pull/11767 so that I would be able to merge the changes there into this PR so I can do a demo. So once this is approved, I will merge once https://github.com/wordpress-mobile/WordPress-Android/pull/11767 is good 😄

WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/visibility/PrepublishingVisibilityFragment.kt
```diff
@@ -0,0 +1,140 @@
...
+
+        viewModel.toolbarUiState.observe(this, Observer { uiString ->
+            toolbarTitle.text = uiHelpers.getTextOfUiString(
+                    requireContext(),
+                    uiString
+            )
+        })
+
+        closeButton.setOnClickListener { viewModel.onCloseButtonClicked() }
+        backButton.setOnClickListener { viewModel.onBackButtonClicked() }
```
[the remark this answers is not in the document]
jd-alexander: f94eb56

WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/visibility/PrepublishingVisibilityFragment.kt
```diff
@@ -0,0 +1,140 @@
...
+        return inflater.inflate(R.layout.prepublishing_visibility_fragment, container, false)
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        visibility_recycler_view.layoutManager = LinearLayoutManager(requireActivity())
+        visibility_recycler_view.adapter = PrepublishingVisibilityAdapter(requireActivity())
+        toolbarTitle = view.findViewById(R.id.toolbar_title)
+        closeButton = view.findViewById(R.id.close_button)
+        backButton = view.findViewById(R.id.back_button)
```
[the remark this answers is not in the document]
jd-alexander: 0ddb26b

WordPress/src/test/java/org/wordpress/android/ui/posts/prepublishing/visibility/PrepublishingVisibilityViewModelTest.kt
```diff
@@ -0,0 +1,193 @@
...
+        val password = "password"
+        post.setPassword(password)
+        editPostRepository.set { post }
+
+        // act
+        viewModel.start(editPostRepository)
+        invokeVisibilityOnItemTapped(PRIVATE)
+
+        // assert
+        verify(updatePostPasswordUseCase).updatePassword(any(), any(), any())
```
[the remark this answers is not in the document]
jd-alexander: 2e321c9

jd-alexander:
I just noticed when the visibility is set to `PRIVATE` the `Publish Date` is showing the date instead of `Immediately` due to shared logic being used from the `PostSettingsUtil`. I will address this.

jd-alexander:
> I just noticed when the visibility is set to PRIVATE the Publish Date is showing the date instead of Immediately due to shared logic being used from the PostSettingsUtil. I will address this.

I am wondering if this should be addressed since the existing logic in the Post Settings uses it and it's basically the same view and logic being used in the prepublishing bottom sheet. 

Before | After 
--------|-------
 <kbd></kbd>    |       <kbd></kbd>

For consistency, I am thinking to leave it as is.

WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/visibility/PrepublishingVisibilityFragment.kt
```diff
@@ -47,7 +47,7 @@ class PrepublishingVisibilityFragment : Fragment(), PostSettingsInputDialogListe
 
     override fun onResume() {
         super.onResume()
-        viewModel.setEditPostRepository(getEditPostRepository())
+        viewModel.setEditPostRepositoryAfterConfigChanges(getEditPostRepository())
```
jd-alexander, line 50: @ashiagr let me know what you think about this approach. The reason I had to do this is, when you are entering the password in the dialog and rotation occurs the `ViewModel` retains the `editPostRepository` but it's instance is now different from the one in `EditPostActivity` so when the password is on the retained instance and you press back the state is lost so here I am setting it so that it's consistent. Let me know if this approach is okay or if anything can be simplified. For more clarity, I am passing the `editPostRepository` in `start` as well. Thanks 😄

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
```diff
@@ -50,9 +50,12 @@ class PrepublishingHomeViewModel @Inject constructor(
                 HomeUiState(
                         actionType = PUBLISH,
                         actionResult = editPostRepository.getPost()?.let { postImmutableModel ->
-                            UiStringText(
-                                    postSettingsUtils.getPublishDateLabel(postImmutableModel)
-                            )
+                            val label = postSettingsUtils.getPublishDateLabel(postImmutableModel)
```
jd-alexander, line 53: I added this here because it's related to the visibility. When you press publish and the contents of the editor is empty and you change the visibility to `Private` the `Publish Date` is blank so this is to ensure that the default value of `Immediately` is set.

WordPress/src/test/java/org/wordpress/android/ui/posts/prepublishing/visibility/usecases/UpdatePostPasswordUseCaseTest.kt
```diff
@@ -0,0 +1,50 @@
...
+        // arrange
+        var passwordUpdated = false
+
+        // act
+        updatePostPasswordUseCase.updatePassword("password", editPostRepository) {
+            passwordUpdated = true
+        }
+
+        // assert
+        assertThat(passwordUpdated).isTrue()
```
ashiagr, line 36: We can drop `()` from `isTrue()` to use property access syntax in Kotlin.

WordPress/src/test/java/org/wordpress/android/ui/posts/prepublishing/visibility/usecases/UpdatePostPasswordUseCaseTest.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.junit.Before
+import org.junit.Test
+import org.wordpress.android.BaseUnitTest
+import org.wordpress.android.TEST_DISPATCHER
+import org.wordpress.android.fluxc.model.PostModel
+import org.wordpress.android.ui.posts.EditPostRepository
+
+class UpdatePostPasswordUseCaseTest : BaseUnitTest() {
+    lateinit var editPostRepository: EditPostRepository
+    lateinit var updatePostPasswordUseCase: UpdatePostPasswordUseCase
```
ashiagr, line 15: IDE is complaining these fields could be private.

WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/visibility/PrepublishingVisibilityFragment.kt
ashiagr: As start method is called every time our fragment is recreated (which happens even when orientation changes), how about we move `editPostRepository` assignment above `isStarted` check to avoid calling `setEditPostRepositoryAfterConfigChanges` from fragment's `onResume`? 

```
fun start(editPostRepository: EditPostRepository) {
    this.editPostRepository = editPostRepository // <- here
    if (isStarted) return
    isStarted = true

    setToolbarUiState()
    updateUiState()
}
```

jd-alexander: Yes, thanks! This is a good approach. I was wondering if it would be okay to put code above the `isStarted` if that would break it's purpose so thanks for clearing that up.

WordPress/src/test/java/org/wordpress/android/ui/posts/prepublishing/visibility/usecases/UpdatePostPasswordUseCaseTest.kt
jd-alexander: Hi @ashiagr I had tried to do that and it didn't work due to some type inference issue so I left it as is 🙏

WordPress/src/test/java/org/wordpress/android/ui/posts/prepublishing/visibility/usecases/UpdatePostPasswordUseCaseTest.kt
jd-alexander: Whoops. Didn't see these warnings. I will address 🙏

jd-alexander: 6514af3

jd-alexander:
Thanks again for the review @ashiagr 
I created an issue to track this https://github.com/wordpress-mobile/WordPress-Android/issues/11823

jd-alexander:
Thanks for the review @ashiagr 

The concerns are being tracked as follows : 

> I noticed that changing the Visibility/PostStatus will lead to the PrimaryAction in the editor being updated and when this happens the action there doesn't trigger the bottom sheet.

https://github.com/wordpress-mobile/WordPress-Android/issues/11822

> The password dialog doesn't have any validation on the EditTextThat is outside the scope of this PR, but something that could be fixed.

https://github.com/wordpress-mobile/WordPress-Android/issues/11835

PR #11810, opened 2020-05-04
Show video player on long press on video item in media list

### LINKED ISSUE #11735: [GlobalStep] Android – video preview screen in the media picker is blank
### Description
The user will observe that upon adding a video block, selecting the "Choose from device option" and then previewing any of the videos, the preview does not function and a black screen is displayed instead.

### Expected behaviour
When viewing the preview of a video it plays as expected

### Actual behaviour
The video preview is a black screen.

1.	Install WordPress 14.7-rc-1
2.	Login to a valid account.
3.	Create a new post.
4.	Add a video block.
5.	Tap the “ADD VIDEO” button and select “Choose from device”.
6.	Preview any of the videos.

##### Tested on the following:
Huawei P10 (7.0)
Samsung Galaxy S8+ (8.0)

#### Please see the attached video for reference.
[Android Video preview issue reference.zip](https://github.com/wordpress-mobile/WordPress-Android/files/4522167/Android.Video.preview.issue.reference.zip)

#### Submitted by:
Nicholas Brown

planarvoid:
Fixes #11735

In the current code we're not doing anything in the `MediaPreviewFragment` when we receive a video URL as a parameter with `autoPlay` parameter set to false. This results in an empty screen when a user is trying to see a video preview. My solution opens the full screen player in this case but doesn't trigger auto play. I think this behaviour makes sense. Let me know what you think!

To test:
- Open a post in gutenberg
- Add a video block
- Click on "Add Video" 
- Click on "Choose from device"
- Long press on an item
- You see a full screen paused player (instead of blank screen visible previously)
- Start the video
- It starts playing

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/media/MediaPreviewActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/media/MediaPreviewActivity.java
@@ -271,7 +271,7 @@ public class MediaPreviewActivity extends LocaleAwareActivity implements MediaPr
         if (media != null) {
             fragment = MediaPreviewFragment.newInstance(mSite, media, true);
         } else {
-            fragment = MediaPreviewFragment.newInstance(mSite, mContentUri);
+            fragment = MediaPreviewFragment.newInstance(mSite, mContentUri, true);
         }
         getSupportFragmentManager().beginTransaction()
                 .replace(R.id.fragment_container, fragment, MediaPreviewFragment.TAG)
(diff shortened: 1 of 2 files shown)


planarvoid:
thanks for the check @ashiagr, this is a good idea. I've implemented the suggestion and ended up turning on "auto play" when checking the video preview because it matches the behaviour of when we pass the Media object to the same function. This means that now the video preview starts automatically playing the video and I think that's actually good 👍

PR #11812, opened 2020-05-04
Issue/11760 reader empty view ui

### LINKED ISSUE #11760: Update UI of the empty view on the following tab
Update the UI of the empty view on the following tab.

Consider if we need to make any changes to the design when
- the user is logged in with a self hosted site
- the user is logged in with .com account
- the user is logged in with both self-hosted site and .com account

Note: The following taken from Figma
Screen Shot 2020-05-01 at 11 26 38 AM

zwarm:
Fixes #11760 
This PR changes the Reader UI for empty views on the Following tab and the Followed Site tabs. The copy has been updated to reflect the requested design changes. 

Screenshot_20200504-095759_WordPress Beta
Screenshot_20200504-095855_WordPress Beta
Screenshot_20200504-095916_WordPress Beta
Screenshot_20200504-095922_WordPress Beta

To Test
(A) Light mode
- Log into an account that is not following any sites
- Set app to light mode
- Navigate to the reader tab
- Tab on the Following tab 
- The updated empty view should be visible
- Tap on the  Manage Tags & Sites Settings Cog
- Tap on the Followed Sites tab
- The updated empty view should be visible

(B) Set app to dark mode and repeat above steps

PR submission checklist:


zwarm:
@mbshakti - if you have time, can you take a peek and see if it's what you expected. Thanks.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderBlogFragment.java
```diff
@@ -90,7 +90,7 @@ private void checkEmptyView() {
 
         if (hasBlogAdapter() && getBlogAdapter().isEmpty()) {
             actionableEmptyView.setVisibility(View.VISIBLE);
-            actionableEmptyView.image.setImageResource(R.drawable.img_illustration_empty_results_216dp);
+            actionableEmptyView.image.setImageResource(R.drawable.img_illustration_empty_results_162dp);
```
malinajirka, line 93: The previous illustration was a general purpose empty view illustration used throughout the app. The new illustration seems specific to the following tab so I think we might want to consider using a more specific name. Wdyt?

WordPress/src/main/res/drawable-night/img_illustration_empty_results_162dp.xml
```diff
@@ -0,0 +1,31 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="196dp"
+    android:height="162dp"
+    android:viewportWidth="196"
+    android:viewportHeight="162">
+  <path
+      android:pathData="M98.5,88.5m-60.5,0a60.5,60.5 0,1 1,121 0a60.5,60.5 0,1 1,-121 0"
+      android:fillColor="#F1F8FF"
```
malinajirka, line 8: We might want to consider replacing these colors with references to the colors.xml, wdyt? If these colors don't exist in the colors.xml I think we might want to discuss it with the designers - I believe the colors.xml should contain our complete color pallete and afaik all the illustrations should use only colors from this pallete. But I'm not 100% sure about this.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderBlogFragment.java
malinajirka: I also think the dimension value in the name usually corresponds to max(width, height). Can you please double check it, thanks!

zwarm: Good suggestion @malinajirka on the name. Changing the remainder of the empty views are slated for a future project. I'm stumped on the size, it appears to use the pattern of max width specified in the vector itself. I've changed to match this pattern, along with renaming the file to reflect the following feature.

WordPress/src/main/res/drawable-night/img_illustration_empty_results_162dp.xml
zwarm: @mbshakti 
> If these colors don't exist in the colors.xml I think we might want to discuss it with the designers

I was able to switch out only the white color, but the other colors don't exist within the app yet. Can you provide direction? Thanks.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderBlogFragment.java
zwarm: Updated in 8d07bf2

mbshakti:
We discussed this briefly on slack, but there seems to be a weird pixelated edge on the illustration. I increased the stroke width on it and exported a new illustration, hopefully this helps! 

- [new illustration](https://drive.google.com/file/d/1B1u1T-Xfal5UQvN3B3ApG1qJtwXpBUwW/view?usp=sharing)

- [dark mode](https://drive.google.com/file/d/1LHQudmzfRlAIb1qi7ZzMFabu3EA17KOv/view?usp=sharing)

also per @malinajirka's comment, I've updated the colors in the illustration to make sure they're from our palette. [Here's the color specifications](https://drive.google.com/file/d/1j6h74DvhX1jbRrEtEVE5J2VgjK9Pdi1N/view?usp=sharing). Thanks a lot @zwarm

WordPress/src/main/res/drawable-night/img_illustration_empty_results_162dp.xml
zwarm: Colors update in commit 2638d9b
@malinajirka - changed to one vector drawable and using colors.xml

zwarm:
Updated screenshots with vector drawables adjusted.

Light | Dark 
--------|-------
<kbd></kbd>  |   <kbd></kbd>

mbshakti:
Thanks so much @zwarm!

WordPress/src/main/res/values-night/colors.xml
```diff
@@ -38,4 +38,12 @@
 
     <color name="post_history_underlying_background">@color/gray_100</color>
 
+    <!-- Reader Following empty view -->
```
malinajirka, line 41: @theck13 @khaykov Quick question: What is the best approach to creating a vector drawable in light&dark modes? I noticed we mostly use two copies of the drawable and the only difference between them is the color they reference (this color is from the SEMENATIC or STUDIO sections of the colors.xml. 
The approach in this PR is to introduce new entries in the colors.xml and night/colors.xml and reference SEMANTIC or STUDIO from them. How do you feel about it?

malinajirka:
Thanks @zwarm! I've tested the app and it looks awesome;). The code looks good. I think this PR is ready to be merged as soon as we receive a reply for the question regarding the colors - related post paqN3M-1N-p2. I personally like your approach and it's an approach I used in other apps (before dark mode).

WordPress/src/main/res/values-night/colors.xml
zwarm: @malinajirka - Tyler and I discussed the approach on slack before implementation.

malinajirka: Ohh perfect, thanks! ;)

PR #11825, opened 2020-05-05
Prepublishing Nudges : Toolbar Improvements

### LINKED ISSUE #11698: Prepublishing Nudges : UI improvements - master issue
This issue is tracking UI issues discovered while working on this feature in https://github.com/wordpress-mobile/WordPress-Android/pull/11656 

- [x] Hide keyboard sometimes works but sometimes doesn't - usually it doesn't work when the post content is empty, but I don't think it's a rule

- [x]  The home screen doesn't have "x" button in the top right corner whereas the Tags screen has it. It feels a bit weird since the button hides the bottom sheet.

- [x]  I guess this is not related to these changes, but I noticed I can type anything I want into the tags input  eg "Android,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,Cars,,,,, , , , , , ,         ,,,,,,,,,,". Should we perhaps allow only [a-zA-Z,] and single whitespace?

- [x] Touch area of the close & back button needs to be increased. 

 - [x] Investigate a way to leverage the screen height to show the contents of bottom sheet on the different screens. eg the tags screen could use some more space to show the tags on screens that can allow. This should be discussed since it's better if the height is consistent for all screens so the navigation doesn't look weird. The height will be adjusted when the the UI is implemented especially the addition of the header and publish button, so this might solve the issue. 

- [x] [ If you press the publish button quickly in succession it can cause multiple bottom sheets to be shown. This means we should check if the fragment is visible before showing it. 
](https://github.com/wordpress-mobile/WordPress-Android/pull/11719)
- [x] Very icon behavior and UI in dark mode.

jd-alexander:
Fixes #11698

## Solution

1. Increased the touch target of the back and close button that is within the `prepublishing toolbar` by wrapping them in a layout with the default touch target size. 

2. Since the clicks are now being done on the layout I had to replace the `ImageView` references with `View` and in other instances, I simply utilized the auto-generated ktx view references.  

3. Fixed the constraints of the buttons and the fixed the title positioning. 

## Testing
1. Create a new post. 
2. Click publish.
3. Click on both Tags & Publish Date. 
4. Ensure the back and close buttons for both have a good touch area; they should be easy to click. 

<kbd></kbd>

## Reviewing 

## Submitter Checklist


WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/PrepublishingPublishSettingsFragment.kt
```diff
@@ -41,8 +41,8 @@ class PrepublishingPublishSettingsFragment : PublishSettingsFragment() {
     }
 
     override fun setupContent(rootView: ViewGroup, viewModel: PublishSettingsViewModel) {
-        val closeButton = rootView.findViewById<ImageView>(R.id.close_button)
-        val backButton = rootView.findViewById<ImageView>(R.id.back_button)
+        val closeButton = rootView.findViewById<View>(R.id.close_button)
+        val backButton = rootView.findViewById<View>(R.id.back_button)
         val toolbarTitle = rootView.findViewById<TextView>(R.id.toolbar_title)
```
jd-alexander, line 46: I would have used the ktx approach here but then I realized because I am getting these views from the base fragment so it's not possible.

zwarm:
Looks good. No issues tapping back or close. Tested on samsung s20 with rounded edge and it worked well.

jd-alexander:
Thanks for the review @zwarm 🙇‍♂️

PR #11834, opened 2020-05-06
Issue/11817 read announcement content from json

### LINKED ISSUE #11817: What's new - reader announcement content from JSON
For v1 we should be able to read feature announcement content from JSON, so it could be edited by anyone.

For now content can look something like this:
```
{
  "announcements": [
    {
      "version": "14.7",
      "versionCode": 850,
      "detailsUrl": "http://wordpress.org",
      "features": [
        {
          "title": "Super Publishing",
          "subtitle": "Publish amazing articles using the power of your mind! Concentrate on what you want to post, and we will do the rest!",
          "gridiconName": "ic_posts_white_24dp"
        },
        {
          "title": "Great cats and superb dogs are right behind you!",
          "subtitle": "That's right! They are right in the app! They require pets right now. Are you going to look for them or what?",
          "gridiconName": "ic_plans_white_24dp"
        },
        {
          "title": "We like long feature announcements that why this one is going to be extra long and span multiple lines",
          "subtitle": "Here we run out of budget.",
          "gridiconName": "ic_themes_white_24dp"
        }
      ]
    }
  ]
}
```

khaykov:
Fixes #11817

This PR moves local feature announcements into FEATURE_ANNOUNCEMENTS.json file, so they can be edited by anyone. I also updated the feature announcement model to use URL for icons, instead of resource names.

To test:
- Check that feature announcement dialog looks ok and reflects content of json file.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/whatsnew/FeatureAnnouncementProvider.kt
```diff
@@ -1,37 +1,47 @@
...
+        return getFeatureAnnouncements().firstOrNull()
+    }
+
+    fun getFeatureAnnouncements(): List<FeatureAnnouncement> {
+        val featureAnnouncements = arrayListOf<FeatureAnnouncement>()
+
+        var featureAnnouncementFileContent: String? = null
+        try {
+            featureAnnouncementFileContent = WordPress.getContext().assets
+                    .open("FEATURE_ANNOUNCEMENTS.json")
```
develric, line 27: Looking here I was think that maybe can come handy to place a sort of json template file like we do for `gradle.properties-example` and placing a sort of empty state instead of removing the file like

```
{
  "announcements": [
    {
    }
  ]
}
```

No need to decide in this PR (we have few others upcoming so we can better think how to manage this), but wanted to mention.

PS: out of curiosity, is there a reason for the all capital letters?

WordPress/src/test/java/org/wordpress/android/viewmodel/whatsnew/FeatureAnnouncementViewModelTest.kt
```diff
@@ -29,23 +29,23 @@ class FeatureAnnouncementViewModelTest : BaseUnitTest() {
             FeatureAnnouncementItem(
                     "Test Feature 1",
                     "Test Description 1",
-                    1
+                    "http://wordpress.org"
```
develric, line 32: minor np: I know it does not make any functional difference and in AS the hint `iconUrl` clarifies, but while looking into the pure code here in GH I was thinking could help to place something like `http://wordpress.org/icon1.png`...but feel free to ignore this 😊.

WordPress/src/main/java/org/wordpress/android/ui/whatsnew/FeatureAnnouncementProvider.kt
khaykov: Good idea! I'll add it in subsequent PR since I also wanted to work on making some elements optional. I can reflect those rules in example file.

All caps to mimic RELEASE-NOTES.txt. It doesn't have to be like this tho :)

WordPress/src/main/assets/FEATURE_ANNOUNCEMENTS.json
```diff
@@ -0,0 +1,26 @@
+{
+  "announcements": [
+    {
+      "version": "14.7",
+      "versionCode": 1,
```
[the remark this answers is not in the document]
khaykov: Good idea! Updated :)

PR #11836, opened 2020-05-07
Issue/11816 feature announcement analytics

### LINKED ISSUE #11816: What's new - add analytics
We should add analytics to track:

- When you see feature announcement with a parameter that indicates how it's accessed - on app upgrade or from App Settings.
- Time spent on the Feature Announcement screen.
- When you tap on the "FIND OUT MORE" button.

khaykov:
Fixes #11816

This PR adds analytics to Feature Announcement.

We are tracking:
- When the feature announcement is shown on app upgrade.
- When the feature announcement is shown from app settings.
- When find out more button is tapped.
- When the continue button is tapped an how much time user spent on the screen before tapping it.

To test:
- Ideally, place breakpoints near the events and make sure they are called when expected.
- Pay special attention to the time tracking part, and confirm that it matches the time spent on screen.

PR submission checklist:


khaykov:
I found a problem with time tracking so marking this as not ready yet.

khaykov:
Ok, analytics should track the time correctly now, accounting for minimized or killed app.

khaykov:
@develric Idea is that `totalSessionsLength` continues to accumulate as long as you spend time on the feature announcement dialog. After the user dismissed the dialog `onPause` calculates the session duration and it's logged in `onDestroy`. If the user temporarily leaves the dialog (minimizes the app or goes to Find Out More) we pause the session and resume it after he comes back.

Regarding `onDestroy`, as far as I'm aware, yes it will not be called when the ap process is killed by OS. This is some catastrophic stuff, so in my opinion, it's ok to skip the logging in this case :)  But maybe I'm missing some other cases when it might happen?

khaykov:
Updated the scope of VM! Total session length should now reset correctly :)

PR #11838, opened 2020-05-07
Show fragments correctly in bottom navigation

### LINKED ISSUE #11830: Stop fragments on bottom navigation change
We're not correctly switching fragments on bottom navigation changes. We're calling `fragmentManager.replace` which makes the fragment inactive in the fragment manager but at the same time we're keeping the reference to the fragment in the bottom navigation component. This makes it impossible to have a ViewPager with fragments in one of the main fragments.

planarvoid:
Fixes #11830 

Previously we were using `FragmentManager.replace` to replace fragments on bottom navigation change. This call sets the Fragment as not active. However, we were still keeping and using the fragments in the sparse array. Overall this approach works but it causes some unnecessary fragment recreation and it's buggy with child fragments. If the parent is not active, the child fragments are also not active (even if the parent is actually visible). I'm replacing this functionality with `show/hide` calls instead. The fragments get initialized the first time you open the app and when you switch the tabs (or rotate the device), the fragments are retrieved from the fragment manager and shown/hidden. 

To test:
- Smoke test the bottom navigation
- Rotate the device
- The scroll position is kept
- Debug the code to see that the fragments are not unnecessarily created

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainNavigationView.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainNavigationView.kt
@@ -280,15 +280,12 @@ class WPMainNavigationView @JvmOverloads constructor(
     }
 
     private inner class NavAdapter {
-        private val mFragments = mutableMapOf<PageType, Fragment>()
-
         private fun createFragment(pageType: PageType): Fragment {
             val fragment = when (pageType) {
                 MY_SITE -> MySiteFragment.newInstance()
                 READER -> ReaderPostListFragment.newInstance(true)
                 NOTIFS -> NotificationsListFragment.newInstance()
             }
-            mFragments[pageType] = fragment
             fragmentManager.beginTransaction()
                     .add(R.id.fragment_container, fragment, getTagForPageType(pageType))
                     .hide(fragment)
@@ -298,9 +295,8 @@ class WPMainNavigationView @JvmOverloads constructor(
 
         internal fun init() {
             for (pageType in pages()) {
-                if (mFragments[pageType] == null) {
-                    mFragments[pageType] = fragmentManager.findFragmentByTag(getTagForPageType(pageType))
-                            ?: createFragment(pageType)
+                if (fragmentManager.findFragmentByTag(getTagForPageType(pageType)) == null) {
+                    createFragment(pageType)
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


WordPress/src/main/java/org/wordpress/android/ui/main/WPMainNavigationView.kt
```diff
@@ -281,31 +288,45 @@ class WPMainNavigationView @JvmOverloads constructor(
     }
 
     private inner class NavAdapter {
-        private val mFragments = SparseArray<Fragment>(numPages())
+        private val mFragments = mutableMapOf<PageType, Fragment>()
```
malinajirka, line 283: Do we really need to manually keep references to the fragments? I'm not sure it's safe - what if we unintentionally re-use a destroyed fragment. It won't go though it's regular restoration process and we might end up in an undefined state. I'd personally rather rely on the FragmentManager implementation for the fragment caching, wdyt?

WordPress/src/main/java/org/wordpress/android/ui/main/WPMainNavigationView.kt
```diff
@@ -165,15 +164,15 @@ class WPMainNavigationView @JvmOverloads constructor(
...
+        val previousFragment = navAdapter.getFragment(prevPosition)
         if (fragment != null) {
-            fragmentManager
-                    .beginTransaction()
-                    .replace(R.id.fragment_container, fragment, getTagForPosition(position))
-                    // This is used because the main activity sometimes crashes because it's trying to switch fragments
-                    // after `onSaveInstanceState` was already called. This is the related issue
-                    // https://github.com/wordpress-mobile/WordPress-Android/issues/10852
-                    .commitAllowingStateLoss()
+            if (previousFragment != null) {
```
malinajirka, line 169: AFAICT the changes won't change when the `updateCurrentPosition` gets invoked. So if it was invoked after `onSaveInstanceState` before these changes, it'll still be getting invoked after `onSaveInstanceState` now. I'm not a fan of using `commitAllowingStateLoss`, but I think not using it will re-introduce the crash. Wdyt?

Having said that it might be better to reintroduce the crash and fix it at its root rather than don't know about users who see empty fragments when they click on the bottom navigation. 🤷‍♂️

planarvoid: In theory you're right and this still might happen. I agree I'd keep the crash and see if we can fix it properly when it happens. I think it was very rare even before.

WordPress/src/main/java/org/wordpress/android/ui/main/WPMainNavigationView.kt
planarvoid: Good point! I've removed the map and simplified the adapter. I think now it's much better.

PR #11841, opened 2020-05-07
Hide audio tab when the user has no audio

### LINKED ISSUE #11361: Remove the "Audio" tab in the media library if user has no uploaded audio files
It's currently impossible to upload audio from the app #4842. When browsing the audio tab in the media library, the empty screen shows an "Upload Media" button. This button is a disappointment to our users since it will show the image or video upload menu.

screenshot_20200219-191449

planarvoid:
Fixes #11361 

This PR hides the audio tab in the Media activity when the user has no audio. The activity now listens to the event and refreshes the tab layout style if there is a change to the number of audio files.

To test:
- Go to a site without Audio media
- Go to the My Site/Media
- Check that the "Audio" tab is not visible
- Go to Calypso and add an audio file to the site media library
- Pull to refresh in the app
- The "Audio" appears
- Click on the "Audio" tab
- The audio file is visible on the screen
- Remote the audio file in Calypso
- Pull to refresh in the app
- The "Audio" tab disappears and the "All" tab is automatically selected

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/media/MediaBrowserActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/media/MediaBrowserActivity.java
@@ -56,6 +56,7 @@ import org.wordpress.android.fluxc.model.SiteModel;
 import org.wordpress.android.fluxc.store.MediaStore;
 import org.wordpress.android.fluxc.store.MediaStore.CancelMediaPayload;
 import org.wordpress.android.fluxc.store.MediaStore.OnMediaChanged;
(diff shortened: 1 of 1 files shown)


ashiagr:
Works as described 👍.

Just that "Upload Media" is visible again in the Audio tab if we delete audio file from the app and return to the media screen. Not sure if we want to select "All" tab in this scenario.

audio-tab

planarvoid:
thanks for the check @ashiagr !

I've implemented the fix in 882f665ad8bc8aa452808abd64e97bb9fa142863

PR #11848, opened 2020-05-07
Reader move tab layout to parent part 2

malinajirka:
Partially fixes https://github.com/wordpress-mobile/WordPress-Android/issues/11849
This is a second PR in a series of PRs which is trying to introduce a regular TabLayout+ViewPager2 in the main Reader view. The current implementation uses FilteredRecyclerView component (which is not a RecyclerView :P) that has some downsides (eg. swipe gesture doesn't work).

- Removed the "loadTags" async task from the PostListFragment. The tags are loaded in the new ReaderFragment so it wasn't necessary anymore.
- Removed the TabLayout from the FilteredRecyclerView(FRV) component as it wasn't used anymore
- Removed "useTabsForFiltering" and introduced "HideTabLayout" in the FRV. We use it to hide the whole appbar layout as it'll now live in the ReaderFragment (will be introduced in future PRs)
- Fix bug where the SubFilterComponent wasn't being shown on the following tab (the component doesn't work yet, as we need to make some other changes in order to fix its behavior. Will be introduced in future PRs)

To test:
- This https://github.com/wordpress-mobile/WordPress-Android/pull/11840#pullrequestreview-407342819 bug should be fixed now.
- Make sure you can swipe left-right and that the correct content is shown on all the tabs
- This PR breaks search and following-subfilter component. This is expected and will be fixed later

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/FilteredRecyclerView.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/FilteredRecyclerView.java
@@ -249,18 +249,16 @@ public class FilteredRecyclerView extends RelativeLayout {
         }
     }
 
-    private void manageFilterSelection(int position) {
+    private void onSpinnerItemSelected(int position) {
+        if (mHideAppBarLayout) {
+            throw new IllegalStateException("Developer error: Spinner shouldn't be visible when the appbar is hidden.");
+        }
         if (mSelectingRememberedFilterOnCreate) {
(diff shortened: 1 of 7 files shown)


malinajirka:
Thanks @planarvoid! Both issues should be fixed now;). Btw I'm listing existing issues in https://github.com/wordpress-mobile/WordPress-Android/issues/11849.

PR #11866, opened 2020-05-07
Prepublishing Nudges : Track events

### LINKED ISSUE #11672: Prepublishing Nudges : Events
#### editor_post_published_tapped

Fired when the user taps "Publish" or "Submit for review" at the top-right corner of the editor.

#### editor_post_scheduled

Fired when the user schedule a post. It should be tracked from the Post Settings and the Nudges View.

When scheduled from post settings we submit a `via: post_settings` property. When in Nudges view, we submit a `via: prepublishing_nudges` property.

#### editor_post_visibility_changed

Fired when the user changes the visibility of a post. It should be tracked from the Post Settings and the Nudges View.

When scheduled from post settings we submit a `via: post_settings` property. When in Nudges view, we submit a `via: prepublishing_nudges` property.

#### editor_post_tags_added

Fired when the user changes the tags of a post. It should be tracked from the Post Settings and the Nudges View.

When scheduled from post settings we submit a `via: post_settings` property. When in Nudges view, we submit a `via: prepublishing_nudges` property.

We should remove the currently `post_settings_tags_added`.

#### editor_post_publish_now_tapped

Fired when the user taps "Publish Now" or "Schedule Now" in the Prepublishing Nudges view.

jd-alexander:
Fixes #11672

## Solution

For all the events below you will need to create a post with content before triggering them.

- `editor_post_publish_tapped`
  - Click Publish button in the editor to trigger this event. 

- `editor_post_visibility_changed` 
  - Click Publish. When the bottom sheet appears, change the visibility to trigger this event. 
  - Go to the Post Settings of a newly created post. Change the `PostStatus` using the `Status` settings control and this event should be triggered. 
  N.B I am not sure if this is the right approach. Currently, we are somewhat mixing the `visibility` and `status` of a post so this might make sense to connect the two. 

- `editor_post_schedule_changed` 
    - Click Publish Now. Click Publish Date and schedule a new date for the post to be published and the event will be triggered. 
    - Click the Post Settings. Click the Publish Date and schedule a new date for the post to be published and the event will be triggered.   

- `editor_post_publish_now_tapped` 
  - Click Publish. When the bottom sheet is shown Click Publish Now and this event will be triggered.

- `editor_post_password_changed`
  - Click Publish. When the bottom sheet is shown click the Visibility option. Once the Visibility screen is visible, then press the Password Protected option where you enter a password and save it. This event will be triggered upon it's persistence. 
  - Click Post Settings. Click the Password option and enter a password. Once it's saved the event is triggered. 

- `editor_post_tags_changed`
  - Click Publish. Click Tags on the bottom sheet and add tags. Once they are added this event is triggered.
  - Click the Post Settings option. Click tags and add some. Once they are added the event is triggered. 

- `editor_post_categories_added`
  - Click Post Settings. Click the Categories option and add some. The event should be triggered once they are added. 

- `editor_post_format_changed`
   - Click Post Settings. Click the Post Format and change it. The event should be triggered once the changes are made. 

- `editor_post_slug_changed`
  - Click Post Settings. Click the Slug option and add it. The event should be triggered once it's saved. 

- `editor_post_excerpt_changed`
  - Click Post Settings. Click the Excerpt option and add it. The event should be triggered once it's saved. 

- `editor_post_location_changed`
  - Click the Post Settings. Click Location. ( My emulator is giving issues to verify this.)

## Reviewing 
2. Ensure that https://github.com/wordpress-mobile/WordPress-Android/pull/11805 is merged. 
5. Merge :) 

## Submitter Checklist


malinajirka:
This is still marked as "Not ready for review" - is this intentional @jd-alexander ?

jd-alexander:
Whoops! it's not intentional @malinajirka Let me remove it 🙏

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostSettingsFragment.java
```diff
@@ -219,6 +222,7 @@ public void onFetchError(Exception error) {
                     public void onSettingsUpdated() {
                         // mEditPostActivityHook will be null if the fragment is detached
                         if (getEditPostActivityHook() != null) {
+                            mAnalyticsTrackerWrapper.track(Stat.EDITOR_POST_FORMAT_CHANGED);
```
jd-alexander, line 225: Do we track events that aren't from the user? I am now reconsidering this placement due to the fact that the default post format is being set here. Let me know.

jd-alexander:
I have some questions about tracking events. 

- Does it matter if we track before or after an action takes place? Meaning, most of the events I know are just modifying a property on the current post so once I put the event within the same block of code it should be fine, so just verifying. 

- Should iOS and Android events have consistent names? I did it that way and also utilized the same `via prepublishing_nudges/settings` approach done there.

libs/analytics/WordPressAnalytics/src/main/java/org/wordpress/android/analytics/AnalyticsTrackerNosara.java
```diff
@@ -784,10 +784,32 @@ public static String getEventNameForStat(AnalyticsTracker.Stat stat) {
...
+                return "editor_post_visibility_changed";
+            case EDITOR_POST_TAGS_ADDED:
+                return "editor_post_tags_added";
+            case EDITOR_POST_PUBLISH_NOW_TAPPED:
+                return "editor_post_publish_now_tapped";
+            case EDITOR_POST_PASSWORD_CHANGED:
+                return "editor_post_password_changed";
             case EDITOR_UPDATED_POST:
                 return "editor_post_updated";
             case EDITOR_SCHEDULED_POST:
```
jd-alexander, line 801: I realized that this event exists after I added the `EDITOR_POST_SCHEDULED` above. The difference is that `EDITOR_SCHEDULED_POST` is being tracked in the `UploadService` when an upload actually takes place. It seems like a duplicate so I want to hear some thoughts about it. Thanks.

ashiagr (commented):
Thanks for elaborate testing instructions @jd-alexander 👌.

I tested all events, just found one issue with the `editor_post_tags_added` event. Noticed it was fired multiple times when I typed a long tag with pauses while typing. Let me know if you can reproduce it else I can share a video. Also it got fired when I removed a tag. 

I've tried to answer your questions below, hope they make sense :)

> Does it matter if we track before or after an action takes place? Meaning, most of the events I know are just modifying a property on the current post so once I put the event within the same block of code it should be fine, so just verifying.

I've seen at other places in the code where the event is tracked just before the action takes place. So I think we can do that, let's just make sure tracking and action are not too much apart. 

> Should iOS and Android events have consistent names? I did it that way and also utilized the same via prepublishing_nudges/settings approach done there.

Ideally yes. I've been following the same convention.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostSettingsFragment.java
ashiagr: Ideally we should track events due to user action. Let's try to think of a better placement.

malinajirka: There are certain special events which are not related to user actions. However, `EDITOR_POST_FORMAT_CHANGED` feels like a user action, so I agree with Ashita's comment. 
One example of an event which isn't directly tied to user action would be for example `ENHANCED_SITE_CREATION_BACKGROUND_SERVICE_UPDATED`.

jd-alexander: Thank you both for clearing this up :) I will make the necessary changes.

jd-alexander:
> Thanks for elaborate testing instructions @jd-alexander 👌

🙇‍♂️ 

> I tested all events, just found one issue with the editor_post_tags_added event. Noticed it was fired multiple times when I typed a long tag with pauses while typing. Let me know if you can reproduce it else I can share a video.

Thanks! I am going to be looking into this. So what is happening is the tags are persisted once a delay has passed since the values are being saved once we are sure the user has made an input. I am going to be sharing an idea of how this issue could be resolved.  One approach would be to create a flag that marks when a tag has been persisted and when the screen is being navigated from we check that flag and then trigger the event. Wdyt? 

>  Also, it got fired when I removed a tag.

I am thinking that it should be fired when a tag is removed, but only if we utilize an approach similar to the one detailed above where the removal of a tag is actually persisted. 

> I've tried to answer your questions below, hope they make sense :)

Thank you so much for answering them :) 

> I've seen at other places in the code where the event is tracked just before the action takes place. So I think we can do that, let's just make sure tracking and action are not too much apart.

That's good advice!

> Ideally yes. I've been following the same convention.

Makes sense :)

libs/analytics/WordPressAnalytics/src/main/java/org/wordpress/android/analytics/AnalyticsTrackerNosara.java
jd-alexander: On iOS, I am seeing a class that contains the recent events that were added https://github.com/wordpress-mobile/WordPress-iOS/blob/c9f0dbaca94a117ed003e07216ca9510705aa069/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift

I am not seeing any other post published events. 

`EDITOR_POST_SCHEDULE_CHANGED` 
I was thinking of using this name as it implies the post's schedule was changed and the existing `EDITOR_POST_SCHEDULED` is like the terminal operation; i.e the post is actually scheduled when the scheduling action takes place. Let me know your thoughts.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostSettingsFragment.java
jd-alexander: Removed 9d41fee

ashiagr (commented):
> One approach would be to check if the tags were changed and then track this before it's persisted. 

👍. Tested changes, LGTM.

> I am thinking that it should be fired when a tag is removed

Isn't `EDITOR_POST_TAGS_ADDED` event misleading if a tag is removed?

> EDITOR_POST_SCHEDULE_CHANGED
I was thinking of using this name as it implies the post's schedule was changed and the existing EDITOR_POST_SCHEDULED is like the terminal operation; i.e the post is actually scheduled when the scheduling action takes place. Let me know your thoughts.

Sounds good! Let's add and inform in the project channel.

WordPress/src/main/java/org/wordpress/android/ui/posts/TagsFragment.java
```diff
@@ -180,6 +180,14 @@ public void onTagSelected(@NonNull String selectedTag) {
         mTagsEditText.setSelection(mTagsEditText.length());
     }
 
+    boolean wereTagsChanged() {
+        if (mTags != null) {
+            return !mTags.equals(mTagsEditText.getText().toString());
+        } else {
+            return !mTagsEditText.getText().toString().isEmpty();
+        }
+    }
```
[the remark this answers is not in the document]
jd-alexander: > Isn't `EDITOR_POST_TAGS_ADDED` event misleading if a tag is removed?

Nice catch @ashiagr I like the way you think. I didn't spot that 😄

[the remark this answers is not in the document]
jd-alexander: Thanks for the NP. I like them 😄 I will make that change 🙏

libs/analytics/WordPressAnalytics/src/main/java/org/wordpress/android/analytics/AnalyticsTrackerNosara.java
jd-alexander: Hi @ashiagr I discussed this with the rest of the team and we agreed on changing this. Resolved here 86c98c0

WordPress/src/main/java/org/wordpress/android/ui/posts/TagsFragment.java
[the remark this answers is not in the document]
jd-alexander: The tags event is now changed instead of added so it now suits removal as well 6d8624e

jd-alexander:
Ready for another round @ashiagr Thanks for the reviews thus far 😄

ashiagr (approved):
Thanks Joel! Great work 🎉.

PR #11867, opened 2020-05-08
Reader move tab layout to parent part 3

### LINKED ISSUE #11849: Reader Master Issue: Enable horizontal swipe gesture in the Reader

Created PRs

Remaining issues

malinajirka:
Partially fixes #11849

This is a third PR in a series of PRs which is trying to introduce a regular TabLayout+ViewPager2 in the main Reader view. The current implementation uses FilteredRecyclerView component (which is not a RecyclerView :P) that has some downsides (eg. swipe gesture doesn't work).

This PR moves NewsCard from the ReaderPostListViewModel to NewsCardViewModel.

To test:
Prerequisities
- Open LocalNewsItem.kt and increment the `version` field
- Build the app locally

1. Open the app on the Reader tab
2. Make sure the NewsCard is shown only on the firstly selected tab
3. Restart the App and make sure the card is still being shown
4. Click on the "Read More" button and verify it opens a webbrowser
5. Dismiss the card and make sure it disappears (it doesn't re-appear after rotation or restart)

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/modules/ViewModelModule.java
+++ b/WordPress/src/main/java/org/wordpress/android/modules/ViewModelModule.java
@@ -10,6 +10,7 @@ import org.wordpress.android.ui.posts.EditPostPublishSettingsViewModel;
 import org.wordpress.android.ui.posts.PostListMainViewModel;
 import org.wordpress.android.ui.posts.editor.StorePostViewModel;
 import org.wordpress.android.ui.reader.ReaderCommentListViewModel;
+import org.wordpress.android.ui.reader.viewmodels.NewsCardViewModel;
 import org.wordpress.android.ui.reader.viewmodels.ReaderPostListViewModel;
 import org.wordpress.android.ui.reader.viewmodels.ReaderViewModel;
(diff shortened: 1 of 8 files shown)


malinajirka:
Thanks for the review @planarvoid !

>When I open the app, I see the news card. When I switch tabs to discover and back to following, the news card disappears.

As it turned out the card is still being shown but on some devices the list is scrolled to the first post item instead of the NewsCard. Either case this issue should be fixed in the last 2 commits.

It's ready for another round. Thanks!

planarvoid:
@malinajirka I've tried the changes and the news card is now shown correctly. I have two issues (which you can fix in another PR). Neither of the issues has to be fixed on this PR:

- the news card and the subfilters don't work in the dark mode
- somehow I was able to get the news card in both following and discover tabs: 
duplicitni_news_card

malinajirka:
Thanks @planarvoid !

>the news card and the subfilters don't work in the dark mode

I've added this to the list of existing issues.

>somehow I was able to get the news card in both following and discover tabs:

I've tried to reproduce this without success. Tbh I wouldnt' worry about it as it seems like an edge case scenario. Even when it happens, it doesn't really matter. Wdyt?

PR #11872, opened 2020-05-08
Reader move tab layout to parent part 4 - subfilter component

### LINKED ISSUE #11849: Reader Master Issue: Enable horizontal swipe gesture in the Reader

malinajirka:
Partially fixes #11849

Merge instructions:
1. ~Make sure - https://github.com/wordpress-mobile/WordPress-Android/pull/11867 - is merged~
1. Make sure - https://github.com/wordpress-mobile/WordPress-Android/pull/11881 - is merged
2. Remove the "Not Ready for merge" label
3. Update target branch to `master-feature-branch/move-reader-tablayout-to-parent`

This is a fourth PR in a series of PRs which is trying to introduce a regular TabLayout+ViewPager2 in the main Reader view. The current implementation uses FilteredRecyclerView component (which is not a RecyclerView :P) that has some downsides (eg. swipe gesture doesn't work).

Goal of this PR is to move SubFilter component displayed on the Following tab into it's own ViewModel. This PR also initialized the subfilter views and vm only on the Following tab.

I'm sorry for the size of this PR, but 95% of the code was copied. The only thing I updated was the logic when the subfilter component is shown and interacted with from the ReaderPostListFragment.

This PR introduces an issue with how the "time spent in reader" tracking works. This will be fixed in a follow-up PR (the issue is tracked in #11849)

To test:
- Verify the component works the same way it worked before

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubFilterViewModel.kt
```diff
@@ -0,0 +1,338 @@
...
+    fun setDefaultSubfilter() {
+        updateSubfilter(
+                SiteAll(
+                        onClickAction = ::onSubfilterClicked,
+                        isSelected = true
+                ))
+    }
+
+    fun onSubFiltersListButtonClicked() {
+        _updateTagsAndSites.value = Event(EnumSet.of(
```
malinajirka, line 186: Do we really need to fetch the tags from the server whenever the user clicks on the subfilter button? Wdyt?

develric: I think this was a sort of easy win to get the list updated when you follow a site and come back immediately to the filter bottom sheet (IIRC sometimes it could take a bit to get the list updated if not triggered explicitly), but yep I agree it can be too much so I think it's fine to change.

malinajirka: ok, I think we can update that since I believe the event bus event is emitted when you start following a site. So the list should get updated automatically. However, let's keep this out of scope of this PR or even this refactoring.

malinajirka:
>The tabs names are different from before (not sure if it is intended), I get the below instead of the Following, Discover, Likes, Saved names

I noticed the previous tabs weren't translated so I used a different field in the ReaderTab class. However, I didn't notice the change in the names - good catch. I've added an item into the issue #11849.

>Checked the empty states (when no tags/sites followed) in the bottom sheet and it works. Only thing to check is that when no sites followed, “Discover site” button does not bring to Discover tab (possibly something similar to the issue with the Saved Post empty state).

This is already being tracked in the issue and we are working on it.


>For a fresh install both with wp.com/jp and pure self-hosted, the reader was intended to be presented in the Discover tab (not sure if it still applies so maybe something to check with design eventually)

Ohh, I didn't realize that. I've also added it to the list of issues. Thanks for the info;).

>In my emulator I miss the Reader title and the toolbar is not collapsing (probably the toolbar behavior is still in progress since I see an issue open for the search, so sorry in case it's redundant 😅)

This was implemented yesterday in a different PR ;).

>forgot to mention this (again not directly in scope with this PR, wonder if I could add these points in the main issue, sorry 😜); when you sign up with a new user and you have no sites the empty states and the content can overlap like this below:

This happened to me once as well but I wasn't able to replicate it. Thanks for the steps;). I've added it to the list of known issues.


Thank you so much @develric for the great review! It seems none of the issues is directly related to this PR. I've added them to the list of known issues and I'll work on fixes in different PRs. If you agree of course ;)!

PR #11875, opened 2020-05-08
Issue/example feature announcement

khaykov:
This PR adds an example feature announcement file.

In addition, I also made a detailed URL option, and when it's missing the Find Out More button will not be shown.

To test:
- Remove `detailsUrl` from `FEATURE_ANNOUNCEMENTS.json` or set it to an empty string.
- Open feature announcements from App Settings. Confirm that the Find Out More button is not visible.

PR submission checklist:

### DIFF
--- a/WordPress/FEATURE_ANNOUNCEMENTS.json-example
+++ b/WordPress/FEATURE_ANNOUNCEMENTS.json-example
@@ -1,4 +1,5 @@
-// remove comments from actual announcements
+// FEATURE_ANNOUNCEMENT.json file should be located in WordPress/src/main/assets
+// comments in this file are ignored by Gson, so you can either remove or leave them
 
 {
   "announcements": [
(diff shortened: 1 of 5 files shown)


WordPress/FEATURE_ANNOUNCEMENTS.json-example
```diff
@@ -0,0 +1,28 @@
+// remove comments from actual announcements
```
develric, line 1: wdyt if we add in comment also the asset expected target folder?

I was also thinking to a way to avoid the need to remove the comments from the template, the only thing I could think of is to include extra fields like:

```
"_template_note": "app version name that announcement targets. Used for display purposes only.",
"appVersionName": "14.7", 
```

This would be good to avoid the need of removing comments but it's also not great since it introduces extra fields and maybe can overcomplicate things with GSON (or even making it not scalable?). Not feeling strong about it so let me know if you think it's worth or you have a better fit, otherwise I'm fine with keeping the comments 😊

khaykov: Good idea on including file location 👍 

You actually don't need to remove comments, GSON is smart enough to ignore them :) I also don't have a strong opinion on this. Either way, I think that people probably going to remove comments.

PR #11878, opened 2020-05-11
Remove init method from bottom nav

planarvoid:
This PR removes the `init` method that was introduced to preload all the fragments on app start. This is actually not necessary and might introduce slight slowdown on the app start. 

To test:
- Open the app
- Click through all the bottom nav items
- Rotate the device
- Check that the fragments are always there

PR submission checklist:


malinajirka:
Thanks @planarvoid! The code changes look good. However, I've noticed a weird behavior.

The `NavAdapter.createFragment()` method gets always called twice for MySite fragment when you start the app. Even when the initial tab is `Notifications`, two instances of MySite fragment are created.

However, I've checked the behavior on `develop` and it's not ideal there either. It's a bit different though. Two instances of a fragment are created for the initially selected tab. 🤷

PR #11881, opened 2020-05-11
Reader move tab layout to parent part 3.5 - bug fixes

### LINKED ISSUE #11849: Reader Master Issue: Enable horizontal swipe gesture in the Reader

malinajirka:
Partially fixes #11849

This is a 3.5th PR in a series of PRs which is trying to introduce a regular TabLayout+ViewPager2 in the main Reader view. The current implementation uses FilteredRecyclerView component (which is not a RecyclerView :P) that has some downsides (eg. swipe gesture doesn't work).

- The uiState is propagated only when the tags list isn't empty
- The logic related to fetching the tags from the server was moved to the ReaderViewModel
- LoadReaderTabsUseCase manually adds "following" tab when the user is logged in to a self-hosted site (this logic was copied from `develop`)
- Fixes crash when the onPageChange listener was registered multiple times if the tags were loaded more than once

To test:
- Clear app data
1. Log into a self-hosted site and make sure you see "following, discover, saved" tabs
-----------------
- Clear app data
1. Add Thread.sleep(5000) to https://github.com/wordpress-mobile/WordPress-Android/blob/c1a021ff0f2ef2f1345d09db8b2380f6129a14d0/WordPress/src/main/java/org/wordpress/android/ui/reader/services/update/ReaderUpdateLogic.java#L147
2. Start the app
3. Log in
4. Go to "Reader" tab
5. Notice only "following" is shown
6. Wait for 5s - the remaining tabs should appear
7. Click on any of the tabs and make sure the app doesn't crash

PR submission checklist:


WordPress/src/test/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModelTest.kt
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    private fun <T> testWithEmptyTags(block: suspend CoroutineScope.() -> T) {
+        test {
+            whenever(loadReaderTabsUseCase.loadTabs()).thenReturn(ReaderTagList())
+            block()
+        }
+    }
+
+    private fun createNonEmptyReaderTagList(): ReaderTagList {
```
planarvoid, line 120: What do you think about making this a `testWithNonEmptyTags` function like the one above for consistency?

malinajirka: Good point;), fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/11881/commits/88b554dc090c97b5f2a9bd7e4af0d0c702347225

PR #11887, opened 2020-05-11
Prepublishing Nudges : Refactored PostSettingsUtilsTest

### LINKED ISSUE #11884: Prepublishing Nudges : Refactor PostSettingsUtilsTest so it's deterministic
Currently, within the `PostSettingsUtilsTest` there are references to the current date eg. `new Date()`. This makes the test non-determinstic as the values aren't the same each the tests is run so it's better to mock all date APIs.

jd-alexander:
Fixes #11884

## Solution
Remove the usage of `Calendar.getIntance()` and `Date()` from the tests so that mocked dates can be utilized. This allows the test to be deterministic; the dates will always be the same.

## Testing
1. Run the tests in `PostSettingsUtilsTest`.
2. Smoke test the `Status` and `Publish Date` functionality within `Post Settings` to ensure that they are still working as expected.

## Reviewing 
2. Ensure that https://github.com/wordpress-mobile/WordPress-Android/pull/11805 is merged. 
5. Merge :) 

## Submitter Checklist

### DIFF
--- a/WordPress/src/test/java/org/wordpress/android/ui/posts/PostSettingsUtilsTest.kt
+++ b/WordPress/src/test/java/org/wordpress/android/ui/posts/PostSettingsUtilsTest.kt
@@ -16,7 +16,6 @@ import org.wordpress.android.ui.stats.refresh.utils.DateUtils
 import org.wordpress.android.util.DateTimeUtils
 import org.wordpress.android.viewmodel.ResourceProvider
 import java.util.Calendar
-import java.util.Date
 
 class PostSettingsUtilsTest : BaseUnitTest() {
     @Mock lateinit var resourceProvider: ResourceProvider
@@ -26,6 +25,7 @@ class PostSettingsUtilsTest : BaseUnitTest() {
     private lateinit var postUtilsWrapper: PostUtilsWrapper
 
     private val dateCreated = "2019-05-05T14:33:20+0200"
+    private val currentDate = "2019-05-05T20:33:20+0200"
     private val formattedDate = "5. 5. 2019"
     private lateinit var postModel: PostModel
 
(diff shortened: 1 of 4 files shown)


WordPress/src/test/java/org/wordpress/android/ui/posts/PostSettingsUtilsTest.kt
```diff
@@ -119,9 +124,11 @@ class PostSettingsUtilsTest : BaseUnitTest() {
     @Test
     fun `returns "published on" for local draft when date is not set`() {
         postModel.setIsLocalDraft(true)
-        val calendar = Calendar.getInstance()
-        calendar.add(Calendar.MINUTE, -5)
-        postModel.setDateCreated(DateTimeUtils.iso8601FromDate(calendar.time))
+
```
malinajirka, line 127: - The name of this tests says "published on" but the assertion checks for "publish on"
- The name of this tests says the date is not set, but it seems we set the date to "2019-05-05T20:28:20+0200", right?

jd-alexander: Thanks for pointing this out. I will check this out.

jd-alexander: Exactly. I don't understand this test or the purpose of it so I removed it 9d8e5de but I am now thinking that since the `Publish on` logic exists, a test should verify the conditions that can cause it. I just don't understand it's placement.

malinajirka: I definitely wouldn't remove it just because we don't understand it. 
1. I'd check the history if the name/content was changed after the original author created it
2. I'd ask the original author what they think about it

My guess is that the name should be "returns "publish on" for local draft when date within the next 30 minutes".

jd-alexander: Thanks for this @malinajirka these are solid tips that I will keep in my checklist for knowing how to navigate these scenarios.

> My guess is that the name should be "returns "publish on" for local draft when date within the next 30 minutes".

Nice! This is it. It all makes sense now 🙏

jd-alexander: Resolved here 2ef44ee

PR #11892, opened 2020-05-11
WP Stories integration - save event snackbars

mzorz:
This is WIP.

This PR builds on top of #11930 

This PR implements snackbars on the MySiteFragment for:
- start Story saving process
- Story saving process ending successfully
- Story saving process ending in error

This PR uses a similar structure for event posting with EventBus as per the Media and Post uploaders in the `UploadService`, with some analogies for example the `StorySaveEvents` class holds all possible events for the Saving process now (similarly to how `PostEvents` holds events for Post uploading), and hence we can now listen to these in a more structured way.

1. moved start and results events into its own separate class so it's easier to read (`StorySaveEvents`)
2. added the EventBus listeners to MySite fragment as per UploadErrorEvent
3. to handle snackbar through the sequencer, we're using the  `UploadUtilsWrapper` which was already there serving the same purpose. Given both the `FrameSaveService` and `UploadService` are going to be tightly connected (one chaining into the other), makes sense to reuse the same Utils class for snackbars instead of duplicating code unnecessarily.

To test:
- TBD

PR submission checklist:


mzorz:
closing in favor of https://github.com/wordpress-mobile/WordPress-Android/pull/11931

PR #11893, opened 2020-05-11
Pulling avoid-jitpack changes into gutenberg release branch

mchowning:
Pulling changes to build gutenberg-mobile locally without JitPack from the release/14.8 branch into the `gutenberg/integrate_release_1.28.0` so that branch can successfully build on CI.

To test:

Build debug and release builds locally with `BUILD_GUTENBERG_FROM_SOURCE` set to `false`. Also verify that debug builds can be built with `BUILD_GUTENBERG_FROM_SOURCE` set to `true` (so 3 combinations total).

PR submission checklist:

- [X] I have considered adding unit tests where possible.
- [X] I have considered adding accessibility improvements for my changes.
- [X] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #11896, opened 2020-05-12
Gutenberg/embedded build no jitpack

SergioEstevao:
Stop using jitpack

To test:

- 1
  - `wp.BUILD_GUTENBERG_FROM_SOURCE` is false
  - Run `./gradlew assembleVanillaRelease` from the command line
- 2
  - `wp.BUILD_GUTENBERG_FROM_SOURCE` is true
  - Run `./gradlew assembleWasabiDebug` from the command line

PR submission checklist:


.circleci/config.yml
```diff
@@ -11,22 +11,80 @@ commands:
...
+    steps:
+      - run:
+          name: Checkout submodules
+          command: git submodule update --init --recursive
+  checkout-gutenberg-mobile-submodule-only:
+    steps:
+      - run:
+          name: Checkout gutenberg-mobile submodule (no recursive)
+          command: git submodule update --init
+  yarn-bundle-android:
```
dratwas, line 40: Why do we need that? For generating assets?

settings.gradle
```diff
@@ -9,6 +9,12 @@ include ':libs:login:WordPressLoginFlow'
 include ':WordPressMocks'
 project(':WordPressMocks').projectDir = new File(rootProject.projectDir, properties.getOrDefault('wp.wordpress_mocks_path', 'libs/mocks') + '/WordPressMocks')
 
+include ':react-native-aztec'
+project(':react-native-aztec').projectDir = new File(rootProject.projectDir, 'libs/gutenberg-mobile/react-native-aztec/android')
+
+include ':react-native-gutenberg-bridge'
+project(':react-native-gutenberg-bridge').projectDir = new File(rootProject.projectDir, 'libs/gutenberg-mobile/react-native-gutenberg-bridge/android')
+
 if (properties.getOrDefault('wp.BUILD_GUTENBERG_FROM_SOURCE', false).toBoolean()) {
```
dratwas, line 18: Do we still need the `wp.BUILD_GUTENBERG_FROM_SOURCE` flag?

dratwas, line 18: Do we still need the wp.BUILD_GUTENBERG_FROM_SOURCE flag?

.circleci/config.yml
mchowning: I think that's right. We're generating a fresh JS bundle so the ci build includes any changes.

dratwas: Hmmm, so we still will have the bundle in `gutenberg-mobile` and we will create the fresh one on the CI? So if you run locally WordPress-Android app w/o packager then you will have the bundle from `gutenberg-mobile/bundle` but the CI will generate a new one? Which JS bundle will be used in the APK generated by peril?

mchowning: I believe Android is never relying on having the bundle committed to gutenberg-mobile. Instead, we're always generating a new bundle if we're not using the packager, both locally and on CI. Obviously this is a different approach to what WPiOS does.

dratwas: Ok got it :) So, in that case, it's pointless to have the android bundle here - https://github.com/wordpress-mobile/gutenberg-mobile/tree/develop/bundle/android, right?

mchowning: I think that's right.

mchowning:
:wave: @cameronvoell  and @chipsnyder . I think I've resolved the issues you both observed in doing debug builds with the avoid-jetpack changes in #11893 . Would you care to give this PR a try?

chipsnyder:
Ran through the following builds with no problem 🥳 🚀 Nice work!

`BUILD_GUTENBERG_FROM_SOURCE `set to `false`:
- [x] ./gradlew assembleVanillaDebug 
- [x] ./gradlew assembleVanillaRelease 
- [x] ./gradlew assembleWasabiDebug 
- [x] ./gradlew assembleWasabiRelease 

`BUILD_GUTENBERG_FROM_SOURCE `set to `true`:
- [x] ./gradlew assembleVanillaDebug 
- [x] ./gradlew assembleVanillaRelease 
- [x] ./gradlew assembleWasabiDebug 
- [x] ./gradlew assembleWasabiRelease

cameronvoell:
> 👋 @cameronvoell and @chipsnyder . I think I've resolved the issues you both observed in doing debug builds with the avoid-jetpack changes in #11893 . Would you care to give this PR a try?

Same results as Chip here https://github.com/wordpress-mobile/WordPress-Android/pull/11896#issuecomment-628120065 Looks good!

mchowning:
:wave: @hypest ! I added a couple of changes on top of yours here. Would appreciate it if you could take a look since you're the most familiar with this. 
1. We were having issues building from source with these changes because we need to not build the JSBundle when building from source, so in aca31b1 I updated the `buildGutenbergMobileJSBundle` variable to insure that it was never true if `wp.BUILD_GUTENBERG_FROM_SOURCE` is true.
2. There were some CI failures on the "Ensure Assets Folder Exists` stage because we were trying to create the assets folder, but CI reported that the folder already existed. I updated the commands from `mkdir` to `mkdir -p` to avoid this in 2e91423.

hypest:
I think the no-jitpack solution was merged to develop via https://github.com/wordpress-mobile/WordPress-Android/pull/11921 so, not sure if this PR is needed anymore?

SergioEstevao:
I don't think we need this PR now that the release branch was merged back to develop.
Closing it for now, if we need to have something from this branch we can cherry-pick later on.

PR #11902, opened 2020-05-12
Select tab on button tap

### LINKED ISSUE #11849: Reader Master Issue: Enable horizontal swipe gesture in the Reader

zwarm:
Fixes #11849 

This PR is part 5.5 of a larger PR that moves the tab layout to a parent.

The approach here was to pass the `ReaderViewModel` to the `ReaderPostListViewModel` and when an empty view button is clicked, send the request back to the observer (`ReaderFragment`) who will then change the selected index of view_pager. This way the normal viewpager handler will take over as if the tab itself was tapped. Note that we do save change the selectedTag value before posting the button tap to the viewmodel.

This may or may not be the best approach, but without knowing what else will change, I felt it was the safest option at the moment. We can revisit the further we get with the tab layout transformation.

Merge Instructions
1. Review this PR
2. Make sure #11890 is merged
3. Change the target branch to master-feature-branch/move-reader-tablayout-to-parent
4. Remove the "Not Ready for merge" label
5. Merge this PR

To test:
1. Log in as a user who is not following any sites
2. Select "Following" tab in the Reader
3. Tap on the Discover button from within the empty slate
4. Notice the "Discover" tab is selected

1. Log in as a user who has not saved any sites
2. Select "Saved" tab in the Reader
3. Tap on the Following button from within the empty slate
4. Notice the "Following" tab is selected
PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
```diff
@@ -44,6 +44,9 @@ class ReaderViewModel @Inject constructor(
     private val _selectTab = MutableLiveData<Event<TabPosition>>()
     val selectTab: LiveData<Event<TabPosition>> = _selectTab
 
+    private val _forceTabSelection = MutableLiveData<Event<ReaderTag>>()
```
malinajirka, line 47: What if we reused the existing `_selectTab = MutableLiveData<Event<TabPosition>>()` ? The nice thing would be that we'd move the responsibility of mapping `tab -> position` from the view into the VM. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
```diff
@@ -99,8 +102,14 @@ class ReaderViewModel @Inject constructor(
         return now - lastUpdated > UPDATE_TAGS_THRESHOLD
     }
 
+    fun emptyStateTabChange(tag: ReaderTag) {
+        onTagChanged(tag)
```
malinajirka, line 106: Do we need to manually call `onTagChanged`? I'd expect when the tab is changed it'll invoke the `onPageChangeCallback` and the tag gets eventually changed. I haven't tested it, so I might be wrong. Let me know what you think ;).

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -412,6 +414,8 @@ public void onCreate(Bundle savedInstanceState) {
                                        .get(ReaderPostListViewModel.class);
         mNewsCardViewModel = ViewModelProviders.of(requireActivity(), mViewModelFactory)
                                                .get(NewsCardViewModel.class);
+        mReaderViewModel = ViewModelProviders.of(getParentFragment(), mViewModelFactory)
```
malinajirka, line 417: The ReaderPostListFragment is used even on different screens than just the Reader entry point - eg. Open Reader -> Click on a blog title of one of the posts in the discovery tab -> notice ReaderPostListActivity + ReaderPostListFragment gets created. 
I believe the `getParentFragment` might return null as there isn't any parent fragment in this scenario. Even if the parent fragment existed it wouldn't be `ReaderFragment` and ReaderViewModel wouldn't exist so this would create a duplicate of `ReaderViewModel` which would never be used.
One solution is to wrap this into `mIsTopLevel` - it's a flag indicating we are in the top level (Reader entry point) screen. We can discuss other possible solution on a call/slack.

Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
zwarm: > What if we reused the existing _selectTab = MutableLiveData<Event<TabPosition>>() ?
I had considered that, but decided against it until the majority of the refactor work was complete. I could just as easily look up the `ReaderViewModel` state in the `ReaderPostListViewModel` to grab the tag tab index.

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
zwarm: > Do we need to manually call onTagChanged?

Nope, the callback will take care of it. Going to remove it.

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
zwarm: > P.S. I'd consider writing a simple unit test for the changes made in ReaderViewModel. Wdyt?

Agreed, just wanted to put out a PR with where I was before continuing. :)

zwarm: > The other possible approach I was considering would be to use Android navigation 

The main reason I didn't go in this direction was that I wanted the animation to remain the same as if the tab was tapped directly. Just feels smoother. I can go either direction if you want it to change.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
zwarm: > We can discuss other possible solution on a call/slack.

That would be good because I lack the 'reader' background and your insight is greatly appreciated.

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
malinajirka: >The main reason I didn't go in this direction was that I wanted the animation to remain the same as if the tab was tapped directly. Just feels smoother. I can go either direction if you want it to change.

I'm not sure what you mean. You can play the animation the same way you are playing it now. The only difference would be from where you receive the information that you should change the tab. Can you please elaborate. Thanks!

>I had considered that, but decided against it until the majority of the refactor work was complete. 

I'm just curious what your thoughts on this were. I might be missing something but it seems to me we are creating a duplicate and the only difference is that one LiveData accept ReaderTab and the other position. Right?

>I could just as easily look up the ReaderViewModel state in the ReaderPostListViewModel to grab the tag tab index.

We could also find it within `fun emptyStateTabChange(tag: ReaderTag)` right? So the ReaderViewModel would do derive the position from its internal state, wdyt?

malinajirka: We've discussed all these points during the call, so I'm resolving this discussion ;).

zwarm:
> I'd consider writing a simple unit test for the changes made in ReaderViewModel. 

Added unit test in commit 13a422d

PR #11903, opened 2020-05-13
Remove unnecessary log - gravatar api timeout

malinajirka:
Fixes https://github.com/wordpress-mobile/WordPress-Android/issues/9843

This PR removes unnecessary `Gravatar api timeout` log. The main reason we are removing this is that it's cluttering up Sentry with unimportant logs. Moreover, the error is being tracked into Tracks so we won't lose the data in case we ever need them.

To test:
- I don't think this PR needs any manual tests

PR submission checklist:


malinajirka:
@oguzkocer Sounds good to me ;). Thanks

PR #11912, opened 2020-05-14
[Fix] Gutenberg - honor allowMultipleSelection js flag in StockMediaPicker

mkevins:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/2261

### Description
When the user selects "Free photos library" from the Cover block, it allows multiple photos to be selected. This is inconsistent with the other upload options (which only allow a single image to be selected). Additionally, these selected images are all uploaded to the media library upon confirmation, but only the first selected image is displayed in the post.

This PR adds a new request code for starting the StockMediaPicker activity for single select mode. This is necessary because the existing request code used for this purpose (`RequestCodes.STOCK_MEDIA_PICKER_SINGLE_SELECT`) is assumed to be used for the featured image picker.

**Edit:**

It turns out the issue that this PR fixes also exists for other blocks, and this PR fixes the issue for those blocks as well (e.g. Image, and Media & Text).

**To test:**

1. Create a new Gutenberg post
2. Add a cover block
3. Tap "Add image or video"
4. Tap "Choose from Free Photo Library" 
5. Type a search term

**Expected behavior:**

Only 1 image can be selected.

**Steps:**

6. Select an image
7. Tap "Use this photo"

**Expected behavior**
Selected image becomes the background of the Cover block

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 14.9
 -----
 * Fix issue with Preview post not working after switching to classic editor from inside the post
+* [**] Block editor: Fix bug in Free Photo Library which allowed selecting multiple images but only inserted one
  
 14.8
 -----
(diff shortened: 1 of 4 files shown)


mkevins:
Thanks for reviewing and testing!

PR #11920, opened 2020-05-14
WP Stories integration phase 1 - lint fixes not coming from Stories source code

mzorz:
This PR builds on top of #11864 

Note: the instructions for building are in the base PR #11864.

This PR fixes a lot of lint errors that appeared only after adding the `portkey-android` repo as a `git submodule` in #11864, and that are not strictly related to the submodule being added, but are all part of the code available in the base `develop` branch. I have been trying to understand why these issues are reported by the linter only after adding the submodule (and these errors do _not_ appear in the submodule's code), so after struggling for a couple of days and verifying all the libraries used are the same versions, etc as in the target codebase, decided to go ahead and fix the issues with this PR.
In this sense, while this PR needs to be in the chain of a feature branch for development, is not strictly related to the functionality that the feature branch is going to end up having, but only fixes problems in the target codebase to give the new submodule a proper welcome.

The issues that appear here were of 2 kinds:
- `UseRequireInsteadOfGet`, in order to actually fix this we'd need to change most of the `context!!` and `activity!!` declarations, which I consider this is a nice to have to make code nicer but not really a behavior change, hence I added the flag as a `warning` class in the `lint.xml` configuration in 531d608.
- use `lifeCycleOwner` in all observe() calls where a Fragment or Activity was being added for `ViewModel` changes listening. I believe this change needed to be done as it was the source of many other issues that have been otherwise fixed as they manifested themselves (see for example this one here https://github.com/wordpress-mobile/WordPress-Android/pull/10954 by @develric, I've seen other comments by @planarvoid on other PRs but couldn't find their reference again). This is fixed in a889ee4.

For the second one, I found this  SO answer here to be useful https://stackoverflow.com/questions/59521691/use-viewlifecycleowner-as-the-lifecycleowner

From the documentation of Fragment's `getLifecycleOwner` source code itself it seems this should be the way to go anyway, and is safe to change to the recommended way:
>      * Get a {@link LifecycleOwner} that represents the {@link #getView() Fragment's View}
>     * lifecycle. In most cases, this mirrors the lifecycle of the Fragment itself, but in cases
>     * of {@link FragmentTransaction#detach(Fragment) detached} Fragments, the lifecycle of the

To test:
- build and smoke test the app. Please follow the instructions in  #11864  for initialization of the submodule
- verify the changes in this PR address all of the lint errors that can be seen in the base branch PR (#11864)

PR submission checklist:

### DIFF
--- a/WordPress/lint.xml
+++ b/WordPress/lint.xml
@@ -6,6 +6,7 @@
     <!-- WARNING -->
     <issue id="RtlSymmetry" severity="warning" />
     <issue id="UseSparseArrays" severity="warning" />
+    <issue id="UseRequireInsteadOfGet" severity="warning" />
 
     <issue id="NewApi">
         <warning path="src/main/res/values/reader_styles.xml" />
--- a/WordPress/src/main/java/org/wordpress/android/ui/CommentFullScreenDialogFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/CommentFullScreenDialogFragment.kt
@@ -49,7 +49,7 @@ class CommentFullScreenDialogFragment : Fragment(), CollapseFullScreenDialogCont
             }
         })
 
-        viewModel.onKeyboardOpened.observe(this, Observer {
+        viewModel.onKeyboardOpened.observe(viewLifecycleOwner, Observer {
             it?.applyIfNotHandled {
                 GlobalScope.launch {
                     val imm = activity?.getSystemService(Context.INPUT_METHOD_SERVICE) as? InputMethodManager
--- a/WordPress/src/main/java/org/wordpress/android/ui/JetpackRemoteInstallFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/JetpackRemoteInstallFragment.kt
@@ -44,7 +44,7 @@ class JetpackRemoteInstallFragment : Fragment() {
             val source = intent.getSerializableExtra(TRACKING_SOURCE_KEY) as JetpackConnectionSource
             val retrievedState = savedInstanceState?.getSerializable(VIEW_STATE) as? JetpackRemoteInstallViewState.Type
             viewModel.start(site, retrievedState)
-            viewModel.liveViewState.observe(this, Observer { viewState ->
+            viewModel.liveViewState.observe(viewLifecycleOwner, Observer { viewState ->
                 if (viewState != null) {
                     if (viewState is JetpackRemoteInstallViewState.Error) {
                         AppLog.e(AppLog.T.JETPACK_REMOTE_INSTALL, "An error occurred while installing Jetpack")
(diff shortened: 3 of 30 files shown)


malinajirka:
Thanks @mzorz!

Unfortunately, I'm not able to build the app ("Could not resolve project :libs:portkey-android:stories"). However, I reviewed the code and tested the app on the CircleCI build and it LGTM.

>I believe this change needed to be done as it was the source of many other issues that have been otherwise fixed as they manifested themselves (see for example this one here #10954 by @develric, I've seen other comments by @planarvoid on other PRs but couldn't find their reference again). This is fixed in a889ee4.

I think the issues you pointed out are related to a bit different issue. The mentioned issues are dealing with the difference between having the observer Activity or Fragment. Whereas this PR is dealing with the difference between having as the observer Fragment or FragmentLifeCyclerOwner. Anyway, I agree this is a good change as it fixes some issues when dealing with fragment transactions and LiveData.

As I mentioned on the p2 post I think we should also go through the release notes of `fragments v1.2.0` and make sure none of the changes are breaking changes.

mzorz:
> Hey @mzorz. The code itself looks good but I was unable to test it because I can't get it to build. > I'm getting this error:

> ERROR: Unable to resolve dependency for > ':WordPress@wasabiDebugUnitTest/compileClasspath': Could not resolve project :libs:portkey-> android:stories.
> Am I missing some step?

@0nko The instructions are in the base PR, should have made that clear 😅 - look here  please https://github.com/wordpress-mobile/WordPress-Android/pull/11864 - thanks for checking!

mzorz:
Thanks for taking the time to check this @malinajirka !

> The mentioned issues are dealing with the difference between having the observer Activity or Fragment. Whereas this PR is dealing with the difference between having as the observer Fragment or FragmentLifeCyclerOwner. 

I think I wanted to say it's about getting the _right_ lifecycleOwner, which is obtained through the `getLifecycleOwner`  in a Fragment (which as you say is true, all the changes in this PR only touch on Fragments) but it's also applicable to Activity (both Fragment or Activity can be lifecycleOwners). Was trying to show how using `getActivity()` or `this` directly in the case of the referenced PR is a common  source of problems, that can be solved by encapsulating the decision using  `getLifecycleOwner()`.

> Anyway, I agree this is a good change as it fixes some issues when dealing with fragment transactions and LiveData.

Right, agreed 👍🏻

> As I mentioned on the p2 post I think we should also go through the release notes of fragments v1.2.0 and make sure none of the changes are breaking changes.

As commented elsewhere as well, it'd be great if we could isolate this change on the target `develop` codebase alone if at all possible, but also happy to do it here in the context of this Stories feature branch chained PRs if it turns out to be the case the only things that need to upgrade belong to this functionality (I don't think so - or couldn't see it with my own eyes only, so I very much appreciate your help here 🙏🏻🙌🏻❤️)

mzorz:
Going to move this one to target `develop` soon  as discussed, so, switching back to draft for the time being 👍

mzorz:
Closing this one in favor of https://github.com/wordpress-mobile/WordPress-Android/pull/11934 which targets `develop`, thank you!

PR #11929, opened 2020-05-15
Set discover tab selection

### LINKED ISSUE #11909: Reader: After the user logs in the default selected tab should be discover
After the user logs in, we currently display the "Following" tab. However, before the refactoring the "discover" tab was selected as default tab.

Note: Keep the behavior implemented in https://github.com/wordpress-mobile/WordPress-Android/pull/11890

zwarm:
Fixes #11909 

This PR sets the selected tab to Discover when a reader tag is not available in app preferences and `isTopLevel=true`. This should only happen upon first entry into the reader. 

To test:

**Test that the discover tab is selected upon first entry into reader**

1. Login in to the app with an account that is not following any sites
2. Navigate to Reader
3. Ensure that “Discover” is the selected tab

**Test that the last selected tab remains selected upon app close** 

1. Tap on the "Followed Sites" tab
2. Close the app
3. Open the app
4. Ensure that “Followed Sites” is the selected tab
5. Repeat for My Likes & Saved Posts tabs

**Test that the last selected tab remains selected navigate away from Reader** 
1. Tap on the "Followed Sites" tab
2. Tap on My site from the navigation bar
3. Tap on the Reader tab
4. Ensure that “Followed Sites” is the selected tab
5. Repeat for My Likes & Saved Posts tabs

PR submission checklist:

- [X ] I have considered adding unit tests where possible.
- [X ] I have considered adding accessibility improvements for my changes.
- [X ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
@@ -18,7 +18,6 @@ import org.wordpress.android.ui.reader.usecases.LoadReaderTabsUseCase
 import org.wordpress.android.ui.reader.utils.DateProvider
 import org.wordpress.android.util.distinct
 import org.wordpress.android.viewmodel.Event
-import org.wordpress.android.viewmodel.ResourceProvider
 import org.wordpress.android.viewmodel.ScopedViewModel
 import javax.inject.Inject
 import javax.inject.Named
@@ -32,8 +31,7 @@ class ReaderViewModel @Inject constructor(
     @Named(BG_THREAD) private val bgDispatcher: CoroutineDispatcher,
     private val appPrefsWrapper: AppPrefsWrapper,
     private val dateProvider: DateProvider,
-    private val loadReaderTabsUseCase: LoadReaderTabsUseCase,
-    private val resourceProvider: ResourceProvider
+    private val loadReaderTabsUseCase: LoadReaderTabsUseCase
 ) : ScopedViewModel(mainDispatcher) {
     private var initialized: Boolean = false
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
@@ -8,17 +8,14 @@ import kotlinx.coroutines.withContext
 import org.greenrobot.eventbus.EventBus
 import org.greenrobot.eventbus.Subscribe
 import org.greenrobot.eventbus.ThreadMode.MAIN
-import org.wordpress.android.R.string
 import org.wordpress.android.models.ReaderTag
 import org.wordpress.android.models.ReaderTagList
 import org.wordpress.android.modules.BG_THREAD
 import org.wordpress.android.modules.UI_THREAD
-import org.wordpress.android.ui.prefs.AppPrefs
 import org.wordpress.android.ui.prefs.AppPrefsWrapper
 import org.wordpress.android.ui.reader.ReaderEvents
 import org.wordpress.android.ui.reader.usecases.LoadReaderTabsUseCase
 import org.wordpress.android.ui.reader.utils.DateProvider
-import org.wordpress.android.ui.reader.utils.ReaderUtils
 import org.wordpress.android.util.distinct
 import org.wordpress.android.viewmodel.Event
 import org.wordpress.android.viewmodel.ResourceProvider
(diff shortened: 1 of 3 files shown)


zwarm:
@malinajirka - 

> What do you think about moving the decision about which fragment should be selected to an else branch here?

Too funny, I was on the fence about adding the logic to that spot. let me rework it and see what testing shows. Sometimes I should just go with my first instinct, but learning my way around this particular fragment has been nerve-wracking. Thanks for the vote of confidence.

malinajirka:
>Too funny, I was on the fence about adding the logic to that spot. let me rework it and see what testing shows. Sometimes I should just go with my first instinct, but learning my way around this particular fragment has been nerve-wracking.

On one side, I'm sorry I didn't share some of my thoughts with you on the ticket before you dove into the implementation. On the other side, I'm glad you went through these steps as I believe/hope it was more valuable for you.

>Sometimes I should just go with my first instinct, 

I keep telling this to myself all the time:D.

>but learning my way around this particular fragment has been nerve-wracking. 

Yeah, tell me about it:D. But all the knowledge we gather will come handy during the project itself ;).

zwarm:
@malinajirka 
> On the other side, I'm glad you went through these steps as I believe/hope it was more valuable for you.

Yes, 100% agree.

I went ahead and moved logic to the `restoreTabSelection` in commit e30d539. It works as expected, same testing steps apply.

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
```diff
@@ -75,11 +80,20 @@ class ReaderViewModel @Inject constructor(
 
     private suspend fun restoreTabSelection(tagList: ReaderTagList) {
         withContext(bgDispatcher) {
-            appPrefsWrapper.getReaderTag()?.let {
-                val index = tagList.indexOf(it)
-                if (index != -1) {
-                    _selectTab.postValue(Event(index))
+            val index  = if (AppPrefs.getReaderTag() == null) {
```
malinajirka, line 83: We might want to consider finding the correct tag in the `tagList` using `isDiscover()`.

Something like the following 
```
withContext(bgDispatcher) {
            val selectTab = { it: ReaderTag ->
                val index = tagList.indexOf(it)
                if (index != -1) {
                    _selectTab.postValue(Event(index))
                }
            }
            appPrefsWrapper.getReaderTag()?.let {
                selectTab.invoke(it)
            } ?: tagList.find { it.isDiscover }?.let {
                selectTab.invoke(it)
            }
        }
```

Wdyt?

malinajirka: I'd also consider renaming the method as it doesn't just restore the selection anymore. Wdyt?

malinajirka:
I've reviewed the changes and left some notes. If you decide not to refactor the method, I'd at least consider replacing `AppPrefs` with `appPrefsWrapper`, wdyt? It'd be also great to introduce a unit test for the new behavior.

SideNote: I noticed `ktlint` is failing.

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
zwarm: @malinajirka 
> We might want to consider finding the correct tag in the tagList using isDiscover().

I hadn't seen isDiscover() used in the original code, so did not know it existed. Thanks for the tip.

> I'd also consider renaming the method as it doesn't just restore the selection anymore. Wdyt?  

Feels like `initializeTabSelection` may be a better name, although I feel that the method would never directly be called to set the discover tab. To further this, wdyt on wrapping the default so it can be changed in the future? 

`public boolean isDefault() {
        return isDiscover();    
    }`

zwarm:
> I've reviewed the changes and left some notes. 

I refactored in 132c51c
 
> SideNote: I noticed `ktlint` is failing.

Already taken care of

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
malinajirka: >I hadn't seen isDiscover() used in the original code, so did not know it existed. Thanks for the tip.

Yeah, that's what I assumed ;). 

>To further this, wdyt on wrapping the default so it can be changed in the future?

Love the idea! ;)

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
```diff
@@ -31,7 +32,8 @@ class ReaderViewModel @Inject constructor(
     @Named(BG_THREAD) private val bgDispatcher: CoroutineDispatcher,
     private val appPrefsWrapper: AppPrefsWrapper,
     private val dateProvider: DateProvider,
-    private val loadReaderTabsUseCase: LoadReaderTabsUseCase
+    private val loadReaderTabsUseCase: LoadReaderTabsUseCase,
+    private val resourceProvider: ResourceProvider
```
malinajirka, line 36: I believe we can remove the resourceProvider now, right? It should fix the failing tests.

zwarm: > Agreed, I already did, it's in the next commit with the updated unit tests. Wanted opinion on default fun first.

zwarm:
> I have left one minor note and it'd be great if we added a simple unit test. I'm approving this PR, feel free to merge it when the tests pass on the CI.

Actually had to change one existing test and add another. :) In 07fe169

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
malinajirka: Ohh I see. I usually assume when someone uses a top-level comment (such as this one https://github.com/wordpress-mobile/WordPress-Android/pull/11929#issuecomment-630794673) that the changes are ready for another round. My bad, I should have waited for an explicit comment ;).

zwarm: My bad. Apologies.

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
zwarm: All ready to go in ed0474d.

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
malinajirka: Really no reason to apologize :). You didn't do anything wrong.

PR #11930, opened 2020-05-15
WP Stories integration - Activity launch

mzorz:
This PR builds on top of #11864

This PR adds the Activity launching code so it's accessible from the FAB.

The basic Activity linking functionality implemented also in this PR is tapping on the main screen FAB should show a third option to create a Story and trigger the composing functionality in the Stories library.

#### Tracks
Tracking will be added on a later PR. 
`// TODO WPSTORIES add TRACKS`

#### To test:
1. checkout the branch and run 
```
git submodule init 
git submodule update
```
or for a fresh clone, you can also do `git clone --recurse-submodules git@github.com:wordpress-mobile/WordPress-Android.git`

2. build
3. run
4. once you land on the main screen and provided you have a site, tap on the FAB and then choose the third option (Story).
5. the camera capture mode from the Stories library should open up, and permissions will be requested accordingly.

To test:

PR submission checklist:


oguzkocer (approved):
Looks good :shipit: The testing instructions work as expected. After the camera is opened, I can't really do anything and the emulator can go a little crazy, but I assume most of the feature is not implemented yet (considering this will land in a feature branch) and my emulator might also be a little 🤯 so I think we are all good. Not merging this yet so it doesn't affect #11864.

WordPress/src/main/res/values/strings.xml
```diff
@@ -1874,6 +1874,7 @@
     <string name="my_site_bottom_sheet_title">Add new</string>
     <string name="my_site_bottom_sheet_add_post">Blog post</string>
     <string name="my_site_bottom_sheet_add_page">Site page</string>
+    <string name="my_site_bottom_sheet_add_story">Story</string>
```
maxme, line 1877: I wish we could add a comment for this one, I have no idea how translators will handle that. And I'm not even sure how to translate it in French (the English word "Story" is used in the French translation of Instagram Stories). Maybe you @aerych you thought about that? @mzorz what would be the Spanish translation?

WordPress/src/main/res/drawable/ic_story_icon_24dp.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+  <path
+      android:pathData="M6,3H14V17H6L6,3ZM4,3C4,1.8954 4.8954,1 6,1H14C15.1046,1 16,1.8954 16,3V17C16,18.1046 15.1046,19 14,19H6C4.8954,19 4,18.1046 4,17V3ZM18,5C19.1046,5 20,5.8954 20,7V21C20,22.1046 19.1046,23 18,23H10C8.8954,23 8,22.1046 8,21H18V5Z"
+      android:fillColor="#2C3338"
```
maxme, line 8: We should use a referenced color like in other drawables.

WordPress/src/main/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModel.kt
```diff
@@ -89,16 +98,9 @@ class WPMainActivityViewModel @Inject constructor(private val appPrefsWrapper: A
...
-        // Currently this bottom sheet has only 2 options.
-        // We should evaluate to re-introduce the bottom sheet also for users without full access to content
-        // if user has at least 2 options (eventually filtering the content not accessible like pages in this case)
-        // See p5T066-1cA-p2/#comment-4463
-        if (hasFullAccessToContent) {
-            _isBottomSheetShowing.value = Event(true)
-        } else {
-            _createAction.postValue(CREATE_NEW_POST)
-        }
+        _isBottomSheetShowing.value = Event(true)
```
maxme, line 103: Problem solved ;) 👍

WordPress/src/main/AndroidManifest.xml
```diff
@@ -214,6 +214,15 @@
                 <action android:name="android.intent.action.MAIN" />
             </intent-filter>
         </activity>
+
+        <!-- Story composer activity -->
+        <activity
+            android:name=".ui.stories.StoryComposerActivity"
+            android:label="@string/app_name"
+            android:screenOrientation="portrait"
```
maxme, line 222: I'm a bit worried by this because it's the only screen that forces portrait mode, but I doubt there is a short term solution for that. I'm just curious if you have long a term plan to support landscape mode?

maxme (changes requested):
Looks good, I'm worried by the string "Story" and how it can be translated. Added a minor comment about the drawable color.

WordPress/src/main/res/values/strings.xml
maxme: Just for reference, the Instagram documentation in French:

Screenshot 2020-06-04 at 10 34 45

mzorz: This is such an interesting point you made @maxme! ❤️ In spanish it'd be `Historia` and just checked and that's how they label it on Instagram, Fb, WhatsApp apps.  I think we could probably add the comment (not sure how to get the comment to translators?) so they understand this should be just like “instagram stories” for them to translate. @loremattei @oguzkocer do you know of a way to add such a comment in a way that we can tell translatros something like `Please use the same word other services such as Instagram use for Stories`? 😄  (not joking - I think this is the best indicator of how we want this translated here)

WordPress/src/main/AndroidManifest.xml
mzorz: Good question, we've discussed that in the team but the Stories format (stories composition UI) seems to have been largely adopted as a portrait-mode only kind of experience on phones. Deferring to @megsfulton for further info.
But apart from that,  do you see any technical problems I may be overseeing @maxme ?

WordPress/src/main/res/drawable/ic_story_icon_24dp.xml
mzorz: 👍 addressed in f3ad439492 (also noted, searched a bit and found lots of hardcoded `fillColor` values in icons in the app, would be great to normalize those)

WordPress/src/main/AndroidManifest.xml
maxme: >  do you see any technical problems I may be overseeing @maxme ?

maxme:
@mzorz I'm fine merging that one if we keep the i18n problem in mind.

mzorz:
>  @mzorz I'm fine merging that one if we keep the i18n problem in mind.

Understood - yes I added another point in my project checklist to discuss this with @loremattei (also discussed internally with @megsfulton and agrees it should be treated as you suggested)

Thank you @maxme ! 🙇

loremattei:
Oh! Sorry, I missed the first ping 😅. Answered above

WordPress/src/main/res/values/strings.xml
loremattei: AFAIK, Android XML format supports developer's comment: any standard XML comment should be attributed to the next string. 
But, I think that GlotPress doesn't support importing comments from XML files. I checked WPAndroid strings on  `translate.wordpress.org` and I don't see comments where I expected them to be. 
Do you know anything more about this @akirk ?

akirk: Correct, it looks like GlotPress currently ignores these comments. We can pass on this comment in our hand off to translators, though. I added this for our next week's kit: p4DzFg-1eM-p2

maxme: Thanks @akirk!

loremattei: Thank you Alex!

mzorz: Awesome, thanks a lot @akirk !

maxme (approved):

WordPress/src/main/res/values/strings.xml
mzorz: question @akirk given the feature won't land in `develop` until a few more  weeks / sprints, should I ping you again once we're approaching ? say as soon as we land it in `develop` for example?

mzorz: Tracking in my todo notes, will do that later 👍

gokih (commented):
https://instagram.com/htctrkkk?igshid=1ivvwvxv8ddch

gokih (commented):
instagram[](https://instagram.com/htctrkkk?igshid=1k750ss2waksv)

PR #11931, opened 2020-05-15
WP Stories integration - save event snackbars

mzorz:
This PR builds on top of #11930 

This PR implements snackbars on the MySiteFragment for:
- start Story saving process
- Story saving process ending successfully
- Story saving process ending in error

This PR uses a similar structure for event posting with EventBus as per the Media and Post uploaders in the `UploadService`, with some analogies for example the `StorySaveEvents` class holds all possible events for the Saving process now (similarly to how `PostEvents` holds events for Post uploading), and hence we can now listen to these in a more structured way.

1. moved start and results events into its own separate class so it's easier to read (`StorySaveEvents`)
2. added the EventBus listeners to MySite fragment as per UploadErrorEvent
3. to handle snackbar through the sequencer, we're using the  `UploadUtilsWrapper` which was already there serving the same purpose. Given both the `FrameSaveService` and `UploadService` are going to be tightly connected (one chaining into the other), makes sense to reuse the same Utils class for snackbars instead of duplicating code unnecessarily.

To test:

Successful path
1. tap on the FAB
2. select Story
3. when the camera preview appears, take a picture
4. optionally, add emoji  / text
5. tap on the `PUBLISH` buttton on the top-right corner of the screen
6. verify the app returns to the previous (main) screen and both he start and end snackbars are shown.

Error path
1. introduce an error in saving (i.e. artificial exception in mp4composerengine such as the one int [this gist](https://gist.github.com/mzorz/da69f0dea9918877defefeb0573b560d))
2. create a video and add some text or emoji
3. tap publish
4. observe the error notification and error snackbar appear
5. tap on MANAGE action on the error snackbar
6. observe the error handling view appears
7. tap RETRY,   see it still fails
8. tap on the Cross to exit the screen, and when prompted about discarding the Story, tap DISCARD
9. observe no snackbar appears (meaning, the sticky event added by the Service has been successfully removed - see commit 5397a25fbeba391f1d795aba429129300d8391fe in Stories repo)
10. observe the error notification (originally created and appearing in step 4 above) is not there anymore in the system dashboard (meaning it has been correctly cleard when discarding the Sotry on step 8 above). (see related commit bbce9a03e53b7c8cb0da3ded5b8da65bdabfd1da in Stories repo, and de7d5c50b2b6e68d203f048c987cabf1ca36cd73 in WPAndroid repo)

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
[...]

PR #11939, opened 2020-05-18
WP Stories integration - Media Picker part 1: device media

mzorz:
This builds on top of #11931 

This PR implements the connections necessary for the WP Stories Activity to interact with the host app (WPAndroid) media picker and be fed the media selection properly.

This PR only works for device media, replicating the existing functionality in the portkey demo app on the WordPress Android app.  The WPMedia and Stock photos sources will be added in subsequent PRs.

To test:
1. run the app
2. select a site
3. tap the FAB button on the "Me" screen
4. select Story
5. observe the Story creation experience appears (camera preview)
6. tap on the upload icon to the left of the camera shutter
7. observe the WPAndroid media picker gets launched
8. select some media items and tap on the checkmark on the top toolbar to indicate you're done selecting
9. observe these media items are added as background for as many Story pages as media items were selected.

Also test:
1. tapping on the camera FAB in the media picker should bring you back to the Stories capture mode.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
```diff
@@ -1,28 +1,56 @@
...
 
 class StoryComposerActivity : ComposeLoopFrameActivity(), SnackbarProvider, MediaPickerProvider {
+    private var site: SiteModel? = null
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setSnackbarProvider(this)
         setMediaPickerProvider(this)
+
+        if (savedInstanceState == null) {
```
malinajirka, line 22: We do this all over the app, so feel free to keep it as is. 

Storying the whole model into a bundle is considered a bad practice (not mentioning that we are using serializable and not parcelable:P). We might want to consider passing just an id of the site in the Intent. ViewModel could load this site from the database and cache it. Wdyt?

malinajirka:
I forgot to leave one more note: We might want to consider using the "ImageEditor" path so the user can crop/rotate the image before it gets added to the story. I'm not sure this is planned or we perhaps plan to enable the editing with-in the story editor. I don't have enough context, so take this comment with a grain of salt:P.

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
mzorz: That's a good idea and I recall it having come up in the past - I purposedly tried to be a "good citizen" with the target codebase and write code that looks along the lines of what already existed :) , but you have a point in this. I have been thinking for a while and I think the SiteModel itself is not as convoluted as say a PostModel can be, so I think it'd be preferable to keep the code consistent for now and bring the Site id passing improvements as part of a different refactor. Thanks for the suggestion! 🙇

mzorz:
Thanks for your review @malinajirka ! 🙇 

> Let me know if the FAB is expected and I'll approve the PR.

Yes there are a few other little UI problems there that we decided to leave as is for our alpha1 and tackle later, especially knowing that the Media Picker is going to change at some point. Thanks for pointing that out! I totally forgot to mention that as a "known issue" in the PR description 🤕 

This one should be ready for merge then 🎉 thanks!

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
malinajirka: I believe we already use both approaches in the code base. Having said that I'm completely fine keeping it as is, since it's a very minor "issue". ;)

mzorz:
Merging as both PRs have been approved

PR #11948, opened 2020-05-19
Issue/11905 tab names reader

### LINKED ISSUE #11905: Reader: Tab names are different from before
The tab names used to be "Following, Discover, Likes, Saved" before the refactoring, but now are "Followed sites", "Discover", "My Likes", "Saved".
Check what terminology we use in Calypso and iOS and fix the titles accordingly. Verify translations work as expected.

The line which we'll probably need to update is [here](https://github.com/wordpress-mobile/WordPress-Android/blob/2ef68c45f4f4b7b81f0cbe223d2360aaab2b8dda/WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt#L52)

malinajirka:
Fixes #11905 

Revert back to using shorter `tagDisplayName` instead of `tagTitle`. I thought `tagDisplayName` doesn't work in other languages. But it seems it's just a missing translation for Czech. 

To test:
1. Open the Reader
2. Notice the tabs contain shorter version of the tags - eg "Following" instead of "Following sites"

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
@@ -65,7 +65,7 @@ class ReaderViewModel @Inject constructor(
             val tagList = loadReaderTabsUseCase.loadTabs()
             if (tagList.isNotEmpty()) {
                 _uiState.value = ReaderUiState(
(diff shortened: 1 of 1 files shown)


jd-alexander (approved):
Thanks for this change @malinajirka 😄 
Tested and works as expected. LGTM 🚢

PR #11955, opened 2020-05-20
Prepublishing Nudges : Page editor doesn't display tags in bottom sheet

### LINKED ISSUE #11891: Prepublishing Nudges : Disable tags in the bottom sheet for Pages
Tags can't be added to pages so it's being disabled for pages.

jd-alexander:
Fixes #11891 

## Solution

We are able to check if a `post` is a `page`. Once we are able to do that in the `ViewModel` we can utilize it to toggle visibility of the `Tags` option. 

## Testing
1.  Create a new post. 
2. Click Publish Now. 
3. Ensure you see `Tags` in the bottom sheet. 
-----------
1.  Create a new page. 
2. Click Publish Now. 
3. Ensure you don't see `Tags` in the bottom sheet. 

Page - No Tags | Post - Tags 
--------|-------
   <kbd></kbd>    |   <kbd></kbd>

## Reviewing 

## Submitter Checklist


WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingBottomSheetFragment.kt
```diff
@@ -97,6 +98,21 @@ class PrepublishingBottomSheetFragment : WPBottomSheetDialogFragment(),
...
         }
+        setupMinimumHeightForFragmentContainer()
+    }
+
+    private fun setupMinimumHeightForFragmentContainer() {
+        val isPage = checkNotNull(arguments?.getBoolean(IS_PAGE)) {
+            "arguments can't be null."
+        }
+
+        if (isPage) {
```
jd-alexander, line 109: Let me know your thoughts on this approach. I had to do this because the default minimum height that was set so the bottom sheet container remains at a good height no matter the fragment, was causing the page editor's sheet to have too much padding; the minimum height was too high so I lessened it for when it's a page vs a page. I realize this is some form of logic. I didn't run it through the `ViewModel` because it seemed a bit trivial. Nonetheless, happy to hear thoughts on this 🙏

malinajirka: >I didn't run it through the ViewModel because it seemed a bit trivial. Nonetheless, happy to hear thoughts on this 🙏

Knowing that you considered running it through the VM and decided against it is enough ;)!

>was causing the page editor's sheet to have too much padding; the minimum height was too high so I lessened it for when it's a page vs a page. 

I'll test this on low end devices when I'm done reviewing the code ;).

jd-alexander:
> Thanks @jd-alexander! Great work. I really like how the existing implementation allowed us to easily change what we display (or not display) on the screen! Both the code and the behavior is perfect!

Thanks for the review and comments @malinajirka I appreciate it 🙇 

> SideNot: The min height looks ok. However, we might still want to look into a way how to show the bottom sheet fully expanded. The current implementation sometime cuts off the button on smaller devices.

Good point indeed! This is something I have been meaning to address. Tackling this here https://github.com/wordpress-mobile/WordPress-Android/issues/11964

PR #11970, opened 2020-05-20
Fix: show camera icon in media picker for Gutenberg media blocks

mzorz:
Fixes an issue where the path to adding media to Gutenberg media blocks (video, images, text & media, etc) doesn't have access to create a new video / image using the device's default camera app because the icon to access it is hidden.

The feature and the problem were introduced in this commit https://github.com/wordpress-mobile/WordPress-Android/commit/f6b8bac7cbaf4ccbd2a3e762a53d79b1b140d666, in PR #11611 

For example, take this video in particular https://cloudup.com/iQB35zbOIZQ linked in that PR, shows the user adding a video block, then tapping Add video, then tapping choose from device, and right there the picker shows two buttons at the bottom bar: the one on the right being the camera icon which should allow you to launch the device’s default camera app.

#### Before the fix
beforefix

#### After the fix
afterfix

#### To test
0. make sure to have the Block Editor set as default for your test site.
1. create a new post
2. add a Video block
3. tap on the block's Add video
4. tap on "choose from device"
5. tap the camera button on the right, the device's camera app should fire up
6. record a new video with the camera app
7. once you're done tap on OK to use it
8. verify it gets loaded into the GB video block and uploads

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 15.0
 -----
- 
+* [**] Block Editor: Fix bug in media picker, now you can add new pictures and videos directly to your video, image and media & text blocks.
+
 14.9
 -----
 * [*] Fix issue with Preview post not working after switching to classic editor from inside the post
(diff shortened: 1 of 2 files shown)


malinajirka:
Hi @mzorz, this actually wasn't introduced as bug but as desired behavior. Since the user can select "Take a photo/video" directly from Gutenberg we decided to hide that option from the media picker. 

Moreover, it seems the fix will hide the camera icon from all the other places where we use the media picker, right? Eg "Change user avatar". Wdyt?

mzorz:
Closing in favor of https://github.com/wordpress-mobile/WordPress-Android/pull/11980

PR #11975, opened 2020-05-21
Update build.instructions with update submodules git command

malinajirka:
This PR updates the build instruction with `update submodules` command (p9ugOq-16q-p2).


adityabhaskar:
Hi @malinajirka 

Thanks for the additional instructions. It may also help to add an additional step on starting the Gutenberg react server, probably with a link to [gutenberg-mobile's setup and run guide](https://github.com/wordpress-mobile/gutenberg-mobile#set-up).

Cheers!

malinajirka:
@adityabhaskar Thanks for the suggestion! The Metro server shouldn't be required for the build. It's a bug that it's required for the `jalapeno` build flavor - switching to for example `wasabi` flavor and cleaning the project should resolve the error. We are working on a proper fix.

PR #11986, opened 2020-05-21
Issue/11985 downgrade to viewpager1

### LINKED ISSUE #11957: Reader: Unrelated content overwritten by Discover
**NOTE:** We might want to look into this issue after https://github.com/wordpress-mobile/WordPress-Android/issues/11849 is merged into `develop`.

Note2: I believe this issue is caused by this line - https://github.com/wordpress-mobile/WordPress-Android/blob/a9d79fd56d32ed9d2f541f47eb6d2767c97f4c91/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java#L604. We should at least check we update the tag only when `isTopLevel == true`. However, I'd even consider removing this whole block all together as each top level instance of ReaderPostListFragment displays one of "Following, discover, liked, saved" so there is no need to update it manually in onResume. If we want to request newer posts for Discover everytime, we can simply modify this condition https://github.com/wordpress-mobile/WordPress-Android/blob/bca7d138750581c3ab38698f21b75d5979fc3a04/WordPress/src/main/java/org/wordpress/android/datasets/ReaderTagTable.java#L330.

### Expected behavior
When I open a post detail from a list of posts for a tag, and I go back I should see the list of posts for the tag.

### Actual behavior
When I open a post detail from a list of posts for a tag, and I go back I see "discover.com" and a list of posts for "discover.com".

1 - Tap Reader
2 - Tap Discover
3 - Tap post on list
4 - Shown the detail view
5 - Tapped on one of the tags (if there aren't any, just open a different post detail)
6 - Shown a new list of sites related to that tag (with title on view set to the tag)
7 - Tapped on a post from list
8 - Shown the detail view
9 - Tapped Back
10 - Should be shown the tag list.
- Title matches tag
- Contents were the original Discover list not the tag list <- this is not expected

Emulator, API 29

### LINKED ISSUE #11985: Reader: Content is sometimes empty after switching tabs
### Expected behavior
When I switch between tabs the content of the newly selected fragment gets rendered.

### Actual behavior
When I switch between tabs which are 2 tabs apart, the newly selected fragment is empty - when I try to scroll down, the content appears.

1. Open Reader on "Following"
2. Switched to "Saved"
3. Switch back to "Following" and notice the content is empty

following-tab

malinajirka:
Fixes #11985 
Fixes #11957

This PR downgrades ViewPager2 in the ReaderFragment to ViewPager. The ViewPager2 causes issues when the user switches between tabs which are 2 tabs apart - more details in the issue description.

I also had to remove the manual `updateCurrentTag` in onResume - when the user switched from "discover" to "following" the "following tab" sometimes displayed "discover" content. (It was also causing a semi-related issue -> https://github.com/wordpress-mobile/WordPress-Android/issues/11957)

To test:
- it seems this scenario didn't work only when there were more than 4 tabs -> use `a8c` account to test this PR.
1. Open Reader on "Following"
2. Switched to "Saved"
3. Switch back to "Following" and notice the content is NOT empty

---------------
1 - Tap Reader
2 - Tap Discover
3 - Tap post on list
4 - Detail view is shown
5 - Tap on one of the tags (if there aren't any, just open a different post detail)
6 - List of posts related to that tag is shown (with title on view set to the tag)
7 - Tap on a post from the list
8 - Detail view is shown
9 - Tap Back and verify the list from step 6 is shown (before this fix the content of "discover" was shown)

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderFragment.kt
@@ -8,17 +8,16 @@ import android.view.MenuItem
 import android.view.View
 import androidx.appcompat.app.AppCompatActivity
 import androidx.fragment.app.Fragment
+import androidx.fragment.app.FragmentManager
+import androidx.fragment.app.FragmentStatePagerAdapter
 import androidx.lifecycle.Observer
 import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.ViewModelProviders
-import androidx.viewpager2.adapter.FragmentStateAdapter
-import androidx.viewpager2.widget.ViewPager2
-import com.google.android.material.tabs.TabLayoutMediator
+import androidx.viewpager.widget.ViewPager
 import kotlinx.android.synthetic.main.reader_fragment_layout.*
 import org.wordpress.android.R
 import org.wordpress.android.R.string
 import org.wordpress.android.WordPress
-import org.wordpress.android.models.ReaderTagList
 import org.wordpress.android.ui.WPWebViewActivity
 import org.wordpress.android.ui.reader.ReaderTypes.ReaderPostListType
 import org.wordpress.android.ui.reader.services.update.ReaderUpdateLogic.UpdateTask.FOLLOWED_BLOGS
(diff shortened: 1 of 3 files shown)


malinajirka:
We've decided to close this PR and solve this issue with a workaround - we'll be invoking `notifyDataSetChanged` in `onResume`, which forces the view to be redrawn.

PR #11987, opened 2020-05-21
WP Stories integration: Media Picker part 3 - WP Media Library

mzorz:
This PR builds on top of #11953 

This PR implements the necessary connections to pass selected WP Media LIbrary `mediaId`s from the `MediaStore` through to `EditorMedia` so it can feed the Stories Composer with such selected media URIs.
- https://github.com/wordpress-mobile/WordPress-Android/pull/11987/commits/8e2393309b4e94ced0ff91d08336b596a140d980 implements the actual`mediaId` passing 
- https://github.com/wordpress-mobile/WordPress-Android/pull/11987/commits/17f82174acf01c855b5666bddfd470b974908c98 implements the `NotificationIntentLoader` interface so error notifications can trigger the error handling screen when tapped

#### CASE A:

1. run the app
2. tap on the FAB to open the "Add new" bottom sheet
3. tap on `Story` , the capture mode should appear
4. tap on the upload button to the left of the capture button
5. the media picker appears
6. tap on the W icon at the right side of the bottom bar
7. select one or more images or videos
8. observe these get added to the composer.

#### CASE B:
Test also the same on a private site and verify it works.

#### CASE C:
Test the same but instead of tapping the W icon, tap on the one on the left and make sure all the options (including Free Photo library) work correctly.

See video https://cloudup.com/cspOViokYB8

~Known issues (VIDEOS):~
~- remote videos can take time to load, and we're not showing a loading indicator. Should be good if you wait for a bitt.~
~- sometimes it gets stuck for some videos the first time, working on it~

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
```diff
@@ -221,4 +235,21 @@ class StoryComposerActivity : ComposeLoopFrameActivity(), SnackbarProvider, Medi
         addingMediaToEditorProgressDialog = progressDialogHelper
                 .updateProgressDialogState(this, addingMediaToEditorProgressDialog, uiState, uiHelpers)
     }
+
+    override fun getAuthHeaders(url: String): Map<String, String> {
+        return authenticationUtils.getAuthHeaders(url)
+    }
+
+    // NotificationIntentLoader
+    override fun loadIntentForErrorNotification(): Intent {
```
mzorz, line 244: Another seemingly unrelated change - this is so the stories library can properly load the intent for an error notification in case it needs to show one. For example, if you don't have the right auth headers for loading media from a private site - alternatively try adding remote media to the story, and then turning airplane mode on (haven't tried but I think it should work the same way). Then once the error notification is shown, tapping on it will trigger the action defined by such intent loaded here. Feel free to leave this part of the review to @jd-alexander @malinajirka  🙇

malinajirka: I tried to test this, but I failed. The error notifications never shows up for me.

WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoPickerActivity.java
```diff
@@ -322,20 +323,27 @@ public void doNext(Uri uri) {
         }
     }
 
-    private void doMediaIdSelected(long mediaId, @NonNull PhotoPickerMediaSource source) {
-        // if user chose a featured image, track image picked event
-        if (mBrowserType == MediaBrowserType.FEATURED_IMAGE_PICKER) {
+    private void doMediaIdsSelected(ArrayList<Long> mediaIds, @NonNull PhotoPickerMediaSource source) {
+        if (mBrowserType == MediaBrowserType.FEATURED_IMAGE_PICKER && mediaIds != null && mediaIds.size() == 1) {
```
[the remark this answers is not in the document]
mzorz: oh you're totally right. I think I was confused by the fact that I was trying to repurpose `doMediaIdSelected` (which is singular)  to `doMediasIdsSelected` (plural), and as such I was trying to encapsulate the previous conditions under which this was being executed as much as possible. And with that I missed the null case in the else branch 🤕 

Nice find!  Fixed in b7058f0e70 👍

malinajirka:
Thanks for the changes! LGTM, feel free to merge it when the CI finishes.

mzorz:
😄 just for completeness: thanks for your review and reporting the thumbnail bug @malinajirka ! 🙇 
 - the thumbnail bug has been tracked elsewhere, I confirm I was able to reproduce as well and we'll tackle it as part of first alpha 👍 
- addressed the other comment in https://github.com/wordpress-mobile/WordPress-Android/commit/b7058f0e70271f312d3709d26d2d472fd7c282fc

Thank you! 🙌

PR #12006, opened 2020-05-22
Prepublishing Nudges : Schedule - Secondary Action

### LINKED ISSUE #11954: Prepublishing Nudges : Schedule's secondary action of Publish triggers scheduling
### Expected behavior
When the user schedules a post they should see the `PrimaryAction` of `Schedule`. If they click the more button in the toolbar to display the `SecondaryAction` and they press `Publish Now` they should see `Publish Now` in the bottom sheet. 

### Actual behavior
Currently, when the `Publish Now` button is pressed and the bottom sheet is shown, `Schedule Now` is seen. 

1. Create a new post. 
2. Click Post Settings. 
3. Change the publish date to one in the future. 
4. Ensure that `Schedule` is being shown as the `PrimaryAction`. 
5. Click the more icon. 
6. Click `Publish Now`. 
7. Observe that `Schedule Now` is shown. 

<kbd></kbd>

Tested on Google Pixel 4 XL, Android 10, WPAndroid `https://github.com/wordpress-mobile/WordPress-Android/pull/11945` ()

jd-alexander:
Fixes #11954

## Findings
When a post is `SCHEDULED` it has a `SecondaryAction` of `Publish Now` in the Editor's toolbar. This secondary action needs to allow the user to transform the post from scheduled to published. 

## Solution
Pass a flag that signifies that the `SecondaryAction` was triggered. Utilize that to trigger a `UseCase` that overwrites the `PostStatus` and sets the `DateCreated` of the `PostModel` to the current date. 

### New Post
1. Create a new post. 
2. Change the `Publish Date` to one in the future using the `Post Settings`
3. Go back to the Editor. 
4. Click the more icon in the toolbar. 
5. Click Publish. 
6. Notice that the bottom sheet says `Publish Now` and not `Schedule Now.`

<kbd></kbd>

-------------------------- 

### Scheduled Post
1. Go to the Post List. 
2. Click on a Scheduled post. 
3. Click the more icon in the toolbar. 
4. Click Publish Now. 
5. Notice that the bottom sheet says `Update Now` and not `Schedule Now.`

<kbd></kbd>

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
@@ -50,16 +50,13 @@ class PrepublishingHomeViewModel @Inject constructor(
         val prepublishingHomeUiStateList = mutableListOf<PrepublishingHomeItemUiState>().apply {
             add(HeaderUiState(UiStringText(site.name), StringUtils.notNullStr(site.iconUrl)))
 
-            add(
-                    HomeUiState(
+            add(HomeUiState(
                             actionType = VISIBILITY,
                             actionResult = getPostVisibilityUseCase.getVisibility(editPostRepository).textRes,
                             onActionClicked = ::onActionClicked
-                    )
-            )
+            ))
 
-            add(
-                    HomeUiState(
+            add(HomeUiState(
                             actionType = PUBLISH,
                             actionResult = editPostRepository.getPost()?.let { postImmutableModel ->
                                 val label = postSettingsUtils.getPublishDateLabel(postImmutableModel)
@@ -70,8 +67,7 @@ class PrepublishingHomeViewModel @Inject constructor(
                                 }
                             },
                             onActionClicked = ::onActionClicked
-                    )
-            )
+            ))
 
             if (!editPostRepository.isPage) {
                 add(HomeUiState(
@@ -79,8 +75,7 @@ class PrepublishingHomeViewModel @Inject constructor(
                         actionResult = getPostTagsUseCase.getTags(editPostRepository)?.let { UiStringText(it) }
                                 ?: run { UiStringRes(R.string.prepublishing_nudges_home_tags_not_set) },
                         onActionClicked = ::onActionClicked
-                )
-                )
+                ))
             }
 
             add(getButtonUiStateUseCase.getUiState(editPostRepository, site) { publishPost ->
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
@@ -12,9 +12,9 @@ import org.wordpress.android.ui.posts.PrepublishingHomeItemUiState.ActionType.TA
 import org.wordpress.android.ui.posts.PrepublishingHomeItemUiState.ActionType.VISIBILITY
 import org.wordpress.android.ui.posts.PrepublishingHomeItemUiState.HeaderUiState
 import org.wordpress.android.ui.posts.PrepublishingHomeItemUiState.HomeUiState
-import org.wordpress.android.ui.posts.editor.SecondaryEditorAction
 import org.wordpress.android.ui.posts.prepublishing.home.usecases.GetButtonUiStateUseCase
 import org.wordpress.android.ui.posts.prepublishing.visibility.usecases.GetPostVisibilityUseCase
+import org.wordpress.android.ui.posts.prepublishing.visibility.usecases.PublishPostImmediatelyUseCase
 import org.wordpress.android.ui.utils.UiString.UiStringRes
 import org.wordpress.android.ui.utils.UiString.UiStringText
 import org.wordpress.android.util.StringUtils
@@ -27,6 +27,7 @@ class PrepublishingHomeViewModel @Inject constructor(
     private val getPostVisibilityUseCase: GetPostVisibilityUseCase,
     private val postSettingsUtils: PostSettingsUtils,
     private val getButtonUiStateUseCase: GetButtonUiStateUseCase,
+    private val publishPostImmediatelyUseCase: PublishPostImmediatelyUseCase,
     private val analyticsTrackerWrapper: AnalyticsTrackerWrapper
 ) : ViewModel() {
     private var isStarted = false
@@ -45,7 +46,9 @@ class PrepublishingHomeViewModel @Inject constructor(
         isStarted = true
 
         if (isPublishSecondaryEditorAction) {
-
+            publishPostImmediatelyUseCase.updatePostToPublishImmediately(editPostRepository) {
+                setupHomeUiState(editPostRepository, site)
+            }
         } else {
             setupHomeUiState(editPostRepository, site)
         }
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
@@ -14,7 +14,7 @@ import org.wordpress.android.ui.posts.PrepublishingHomeItemUiState.HeaderUiState
 import org.wordpress.android.ui.posts.PrepublishingHomeItemUiState.HomeUiState
 import org.wordpress.android.ui.posts.prepublishing.home.usecases.GetButtonUiStateUseCase
 import org.wordpress.android.ui.posts.prepublishing.visibility.usecases.GetPostVisibilityUseCase
-import org.wordpress.android.ui.posts.prepublishing.visibility.usecases.PublishPostImmediatelyUseCase
+import org.wordpress.android.ui.posts.prepublishing.home.usecases.PublishPostImmediatelyUseCase
 import org.wordpress.android.ui.utils.UiString.UiStringRes
 import org.wordpress.android.ui.utils.UiString.UiStringText
 import org.wordpress.android.util.StringUtils
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
@@ -39,18 +39,14 @@ class PrepublishingHomeViewModel @Inject constructor(
     private val _onSubmitButtonClicked = MutableLiveData<Event<PublishPost>>()
     val onSubmitButtonClicked: LiveData<Event<PublishPost>> = _onSubmitButtonClicked
 
-    fun start(
-        editPostRepository: EditPostRepository,
-        site: SiteModel) {
+    fun start(editPostRepository: EditPostRepository, site: SiteModel) {
         if (isStarted) return
         isStarted = true
 
         setupHomeUiState(editPostRepository, site)
     }
 
-    private fun setupHomeUiState(
-        editPostRepository: EditPostRepository,
-        site: SiteModel) {
+    private fun setupHomeUiState(editPostRepository: EditPostRepository, site: SiteModel) {
         val prepublishingHomeUiStateList = mutableListOf<PrepublishingHomeItemUiState>().apply {
             add(HeaderUiState(UiStringText(site.name), StringUtils.notNullStr(site.iconUrl)))
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
@@ -44,43 +44,40 @@ class PrepublishingHomeViewModel @Inject constructor(
     fun start(
         editPostRepository: EditPostRepository,
         site: SiteModel,
-        isPublishSecondaryEditorAction: Boolean,
-        isNewPost: Boolean
+        isPrimaryEditorAction: Boolean
     ) {
         if (isStarted) return
         isStarted = true
 
-        if (isPublishSecondaryEditorAction) {
-            publishPostImmediatelyUseCase.updatePostToPublishImmediately(editPostRepository, isNewPost) {
-                setupHomeUiState(editPostRepository, site)
-            }
-        } else {
-            setupHomeUiState(editPostRepository, site)
-        }
+        setupHomeUiState(editPostRepository, site)
     }
 
     private fun setupHomeUiState(editPostRepository: EditPostRepository, site: SiteModel) {
         val prepublishingHomeUiStateList = mutableListOf<PrepublishingHomeItemUiState>().apply {
             add(HeaderUiState(UiStringText(site.name), StringUtils.notNullStr(site.iconUrl)))
 
-            add(HomeUiState(
-                    actionType = VISIBILITY,
-                    actionResult = getPostVisibilityUseCase.getVisibility(editPostRepository).textRes,
-                    onActionClicked = ::onActionClicked
-            ))
+            add(
+                    HomeUiState(
+                            actionType = VISIBILITY,
+                            actionResult = getPostVisibilityUseCase.getVisibility(editPostRepository).textRes,
+                            onActionClicked = ::onActionClicked
+                    )
+            )
 
-            add(HomeUiState(
-                    actionType = PUBLISH,
-                    actionResult = editPostRepository.getPost()?.let { postImmutableModel ->
-                        val label = postSettingsUtils.getPublishDateLabel(postImmutableModel)
-                        if (label.isNotEmpty()) {
-                            UiStringText(label)
-                        } else {
-                            UiStringRes(R.string.immediately)
-                        }
-                    },
-                    onActionClicked = ::onActionClicked
-            ))
+            add(
+                    HomeUiState(
+                            actionType = PUBLISH,
+                            actionResult = editPostRepository.getPost()?.let { postImmutableModel ->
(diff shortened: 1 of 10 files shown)


WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
```diff
@@ -39,11 +41,22 @@ class PrepublishingHomeViewModel @Inject constructor(
...
+        editPostRepository: EditPostRepository,
+        site: SiteModel,
+        isPublishSecondaryEditorAction: Boolean,
+        isNewPost: Boolean
+    ) {
         if (isStarted) return
         isStarted = true
 
-        setupHomeUiState(editPostRepository, site)
+        if (isPublishSecondaryEditorAction) {
```
malinajirka, line 53: I'm not sure I follow this flow. SecondaryAction is an abstract term - how do we know that we need to "publishImmediately"? 

If we are just relying on the existing usage I'm not sure it's a good idea. All this code would ideally work (without any changes to the prepublishing nudges classes) even after someone decides to add "Save as draft, update, save, publish, schedule, ..." as a secondary action to the Editor. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/home/usecases/PublishPostImmediatelyUseCase.kt
```diff
@@ -0,0 +1,31 @@
...
+
+import org.wordpress.android.fluxc.model.PostModel
+import org.wordpress.android.fluxc.model.post.PostStatus
+import org.wordpress.android.ui.posts.EditPostRepository
+import org.wordpress.android.ui.posts.EditPostRepository.UpdatePostResult
+import org.wordpress.android.util.DateTimeUtilsWrapper
+import javax.inject.Inject
+
+class PublishPostImmediatelyUseCase @Inject constructor(private val dateTimeUtilsWrapper: DateTimeUtilsWrapper) {
+    fun updatePostToPublishImmediately(
```
malinajirka, line 11: Where was this logic located before? Ideally, we'd
1. re-use the original code and don't introduce a new use case 
2. remove the original code and use this class at the new location
3. refactor the original code so both the original and new location use this class

As we discussed, changing post statuses and date creation is very tricky and we need to be extremely careful not to introduce bugs.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
jd-alexander: > I'm not sure I follow this flow. SecondaryAction is an abstract term - how do we know that we need to "publishImmediately"?

hey @malinajirka Thanks for asking. This is done by [a flag passed to the bottom sheet](https://github.com/wordpress-mobile/WordPress-Android/pull/12006/files#diff-38b6aea9cfff5f3e0e4b7389b44de602R1736). So the flag is only set in the `PUBLISH_NOW` secondary action in the `EditPostActivity`. I wasn't sure if the logic would be clear enough.
Let me know what you think. 
Another approach would be to pass in the `SecondaryAction` enum and using a use case to determine this.

jd-alexander: > If we are just relying on the existing usage I'm not sure it's a good idea. All this code would ideally work (without any changes to the prepublishing nudges classes) even after someone decides to add "Save as draft, update, save, publish, schedule, ..." as a secondary action to the Editor. Wdyt?

If they add those actions it wouldn't be triggered. But indeed the solution wasn't clear enough to make that known so I am open to suggestions in addition to what was mentioned above. Thanks 😄

WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/home/usecases/PublishPostImmediatelyUseCase.kt
jd-alexander: Thanks @malinajirka 

> re-use the original code and don't introduce a new use case
remove the original code and use this class at the new location
refactor the original code so both the original and new location use this class

This logic wasn't located anywhere. It's newly introduced due to the functionality needed.

> As we discussed, changing post statuses and date creation is very tricky and we need to be extremely careful not to introduce bugs.

Thanks for the reminder. I don't want to introduce any bugs at all 😅  

Let me know what you think about this solution. The class was introduced because when the user has a post that was scheduled and `Schedule` is the `PrimaryAction` when they click that action in the `Editor` you will see the bottom sheet. Within the bottom sheet, the `Publish Date` will have the date that's set so that the post can be published in the future. 

If you dismiss the bottom sheet and go back to the Editor, and press the more icon, the `Publish Now` Secondary Action is there. When the user presses this, I am expecting that they would want to see that the post is going to be published immediately because currently it's scheduled. The way to make `Immediately` is to clear the date and set the status. Let me know if this makes sense. I took this approach because we want the `EditPostRepository` to be the source of truth and once the user presses `Publish Now` I think it's okay that we clear the `Scheduled` `PostStatus`.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
malinajirka: >If they add those actions it wouldn't be triggered. But indeed the solution wasn't clear enough to make that known so I am open to suggestions in addition to what was mentioned above. Thanks 😄

What do you mean it wouldn't get triggered? If I add "showPrepublishingNudgeBottomSheet(..)" [here](https://github.com/wordpress-mobile/WordPress-Android/blob/ac2b69651a99d04f1856e302f725e5abde659ffa/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L1241) or [here](https://github.com/wordpress-mobile/WordPress-Android/blob/ac2b69651a99d04f1856e302f725e5abde659ffa/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L1245), it'll get triggered right? I'll pass "isSecondaryAction" == true, since it's true that it's a secondary action. And the BottomSheet will perform "PublishNow". Even though the user pressed "Save/Submit.." and I correctly set "isSecondaryAction == true". Or Am I missing something?


If I'm missing something ignore the rest of this message.

I think we have two options here
1. Pass a specific action the user pressed - eg. using an enum.
2. Rename "isPublishSecondaryEditorAction" (even the flag) to "PublishNow".

The current solution is imo confusing. Because it seems like "publishSecondaryEditorAction" is always just "PublishNow" which is not true.

WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/home/usecases/PublishPostImmediatelyUseCase.kt
malinajirka: Thanks for the explanation! 

>This logic wasn't located anywhere. It's newly introduced due to the functionality needed.

Someone had to change the status and clear the date even before we introduced the prepublishing nudges dialog, right? Wasn't it perhaps [this](https://github.com/wordpress-mobile/WordPress-Android/blob/a86967533294b8bdedb2160eefeb7aec2b92e227/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L1802) place? 

It seems the "EditPostActivity.uploadPost(.)` method is never invoked with `publish == true` now. Could we perhaps clean up the code there which would result in removing [this](https://github.com/wordpress-mobile/WordPress-Android/blob/a86967533294b8bdedb2160eefeb7aec2b92e227/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L1802) and I believe we'd follow the `2.` step from my previous comment ("remove the original code and use this class at the new location
"). Wdyt?


Can you please also elaborate why we want to change the status to "Draft" in certain scenarios? The use case is called `PublishPostImmediately`, so changing the status to "Draft" seems quite confusing. Wdyt? I'd at least suggest adding a comment explaining why we need it (if we really need it).

jd-alexander: Just to point it out here I discovered this method in the `PostUtils` 
https://github.com/wordpress-mobile/WordPress-Android/blob/b3d296e30c4d2da0ae0faefe0ebf348c0581650b/WordPress/src/main/java/org/wordpress/android/ui/posts/PostUtils.java#L549

I am thinking to reuse this for consistency.

jd-alexander: > Can you please also elaborate why we want to change the status to "Draft" in certain scenarios? The use case is called PublishPostImmediately, so changing the status to "Draft" seems quite confusing. Wdyt? I'd at least suggest adding a comment explaining why we need it (if we really need it).

Thanks for this question. I added a comment to make it clearer. When a post is created it's `PostStatus` is Draft. When it is `Draft` then Publish Now is shown. If it's Published then `Update Now` is shown. 

If the user has a post as `Scheduled` and it was previously `Draft` I thought it would be a bad experience to show them `Update Now` when `Publish Now` is pressed so this was the safest/clearest way I thought of doing it without additional logic elsewhere.

jd-alexander: > It seems the EditPostActivity.uploadPost(.) method is never invoked with publish == true now. Could we perhaps clean up the code there which would result in removing this and I believe we'd follow the 2. step from my previous comment ("remove the original code and use this class at the new location"). Wdyt?

Sounds good. It is passed with `true` here though. It just seems a bit hidden since it's getting the value from the `onClicked` event.

https://github.com/wordpress-mobile/WordPress-Android/blob/7148518e873fb95a5a167840a42168387feef1f7/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L1751

jd-alexander: > Could we perhaps clean up the code there which would result in removing this and I believe we'd follow the 2. step from my previous comment ("remove the original code and use this class at the new location"). Wdyt?

Thanks dcca109 85e732c

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
jd-alexander: Resolved by clearing the status before opening the sheet.

jd-alexander:
Hey @malinajirka The `SAVE`, `PENDING_REVIEW` actions will be done in another PR. Thanks for the review. Let me know what you think about this solution.

WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/home/usecases/PublishPostImmediatelyUseCase.kt
malinajirka: >If the user has a post as Scheduled and it was previously Draft I thought it would be a bad experience to show them Update Now when Publish Now is pressed so this was the safest/clearest way I thought of doing it without additional logic elsewhere.

If I understand this correctly we are changing the status to "Draft" even though we want to publish the post just to show the correct title on the button? How would you feel about fixing the button label based on the `isNewPost` flag instead?

>Thanks dcca109 85e732c

Related note: PublishPostImmediatelyUseCase.updatePostToPublishImmediately has "publishPost" flag. Imagine you wouldn't want to look into the body of the class -> what do you think would happen if you called `updatePostToPublishImmediately` with `publishPost == false`? Tbh I'd expect the usecase wouldn't do anything at all in this scenario - it's called "updateToPublish" and I'm passing publish == false".

jd-alexander: > bh I'd expect the usecase wouldn't do anything at all in this scenario - it's called "updateToPublish" and I'm passing publish == false"

Exactly! This is spot on. The logic related to the button should can depend on the `isNewPost` 

> How would you feel about fixing the button label based on the isNewPost flag instead?

I agree with this.

jd-alexander:
Hi @malinajirka so I had a go with the `newPost` approach. It works. The only issue I realized during testing was because we aren't changing the status to get the `Publish Now` to show but we are doing it using the flag, so we have an inconsistency with the primary toolbar action as seen in the screenshot below, where it's showing `Update`. 

To reproduce. 
1. Schedule a post. 
2. Click Publish Now in the toolbar. 
3. See this the issue. 

<kbd></kbd>

What do you think about this? Is it something we can live with for now or would you rather the approach of changing the `PostStatus` to `Draft`. I know we should try to modify the `PostStatus` as little as possible but I think the approach I was suggesting could work cleaner than this one. To explain. 

When you first create a new post the status is `Draft`. When you schedule it, the `PostStatus` changes to `Scheduled` and when you click `Publish Now` setting it back to `Draft` doesn't seem like a bad thing to do. With that, we won't have any inconsistencies with the buttons. Honestly, we really should have had `undo` operations on posts. It would save us so much time 

😅 Anyhow, please let me know what you think about this. This current approach is fine to me and will make the solution simple enough and it works - for the most part 😉

WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/home/usecases/PublishPostImmediatelyUseCase.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.wordpress.android.fluxc.model.post.PostStatus
+import org.wordpress.android.fluxc.model.post.PostStatus.SCHEDULED
+import org.wordpress.android.ui.posts.EditPostRepository
+import org.wordpress.android.util.DateTimeUtilsWrapper
+import javax.inject.Inject
+
+class PublishPostImmediatelyUseCase @Inject constructor(private val dateTimeUtilsWrapper: DateTimeUtilsWrapper) {
+    fun updatePostToPublishImmediately(
+        editPostRepository: EditPostRepository,
+        publishPost: Boolean
```
malinajirka, line 13: It seems this parameter is not used anymore - to be more precise it's always "true". 

As discussed above, I think the parameter is confusing anyway `updatePostToPublishImmediately` with parameter `publishPost` -> it shouldn't do anything at all if the parameter is false imo.

jd-alexander: Yes, this is so true. Thanks for spotting. I moved it here when doing the refactoring but it's definitely not needed. 🙏

jd-alexander:
Thanks again for the review @malinajirka they have been very helpful! 

>  I'm a bit worried it'll cause troubles in scenarios such as this one:
> 
> Create a post
> Enter a title
> Publish the post
> Open the post after it gets published
> Edit the title
> Go to Post Settings and set publish date to future
> Click on Publish now in the overflow menu
> Press back -> Is the post in "Published" or "Draft" state?

Good observation. I reverted to the first version of the implementation I came up with to solve this problem especially based on this comment

> If you think changing the status to draft is better, please go for it. I'll make sure to test all the scenario I will be able to think of and we'll see if it causes any really issues.

😉 

So yes, with this solution I currently have this issue that doesn't occur because I am selecting the `PostStatus` based on the post being new or not. So in the scenario, you explained above the post is in the `Published` state since it's not new 😄 

Also, we don't have the issue of needing to determine which button to show etc. because we are modifying the `PostStatus` so that issue is omitted. 

We have spoken at length here about extracting the logic to a common component so it could be reshared. Upon reflection, I believe this shouldn't be done for this specific implementation because the logic within the use case is different from the `publishPost` logic within the `EditPostActivity`. If it weren't for the `newPost` behavior then we could share the logic as it would be the same but it isn't so I won't make any changes there. If necessary I could always put a comment in this new use case and tell where the inspiration is from, although I am not sure if that's even required. Please let me know what you think about this.

jd-alexander:
> I'm a bit confused. The scenario I was describing wasn't happening in the pre-force-push state 😕 .

Hmmmm. Could you expound a bit more on this? Thanks I am sorry if I made this confusing but before the force push we weren't switching the `PostStatus` between `Draft` & `Published` to determine the button state but in this current (reverted) version we are doing this which ensures the issue doesn't take place. 

> Note on "force push": I always felt like it's better to avoid using force-push for these types of scenarios where the code has already been reviewed. Every force-push is a risk of breaking the git history - eg. if another branch is based on top of the forced-pushed branch or when someone has the pre-force-push branch in their local repo and they push it into the remote

Okay this is a really good point. Thanks for pointing it out. I will always merge the changes in and resolve any conflicts when reverting from now on 🙏 

> . Moreover, the reviewer needs review all the changes again as it's a bit difficult to get oriented in what has and hasn't changed. Or is there an easy way how to see what has changed? (I tried comparing 325af92 with 5a1f25d but the result was still confusing)

Yes, this is a good point. Honestly, I made the decision on the basis that the new code was somewhat irrelevant since I wasn't going to use it but now thinking about the importance of historic preservation from a code standpoint it does indeed lessen the reviewer's mental load for sure. I am noting this. Thanks for the points. I appreciate them. 

> I've reviewed the changes and tested the behavior and it LGTM. Thanks for all the changes @jd-alexander ! ;)

Again, thanks for the review and the thoughts toward improving the review process. I am noting them for future PRs. Děkuju 🙏

PR #12008, opened 2020-05-22
Prepublishing Nudges : Refactor Visibility to match Post Status

### LINKED ISSUE #11889: Prepublishing Nudges : Refactor Post Visibility to match Post Status
Show the following options in the visibility - “Draft”, “Private”, “Published”, “Pending Review”, “Password protected”. Ideally, change the status in the PostModel after the user hits the publish now button so we don’t need to handle weird situations when the user leaves the bottom sheet without publishing the changes.

jd-alexander:
Fixes #11889

## Solution
Previously, we only had three `Visibility` states. Public, Private & Password Protected. However, during testing it was realized that there would be issues when trying to map each `Visibility` to the `PostStatus` and a post that is `Public` could mean many things so, for consistency and maintainability, the `PostStatus` behavior in Post Settings was replicated. 

## Testing
1. Create a new post. 
2. Click Publish. 
3. Click the Visibility. 
4. Test each of them to ensure changing them does the same in Post Settings. 

Before | After 
--------|-------

## Reviewing 

## Submitter Checklist


jd-alexander:
@ashiagr I am wondering if I should remove the `Pending Review` from the Status within in Prepublishing Nudges - Visibility since we can't switch to that `PostStatus` since the prepublishing nudges should only be visible for statuses where the post will be published - in the past, now, or future.

jd-alexander:
So for the `PENDING_REVIEW` it's option will be kept in the `Visibility` setting and `GetButtonUiStateUseCase` will be modified to support it.

ashiagr:
Thanks @jd-alexander ! Agree it will be easier to review in #12042.

One more question, I swear the last one for this PR 😅 :

If I select visibility status: **Draft** for a future publish date, should we show **Publish Now** or **Schedule Now** button? Post Settings displays **Publish** so it is consistent with that behaviour and not a fault of changes in this PR but it looks like a discrepancy, wdyt?

publish_draft_schedulesd

Since we don't display a confirmation dialog in prepublishing nudges so a draft that one wants to publish at a later date might get published immediately 🤔.

jd-alexander:
Thanks @ashiagr I will investigate this and get back to you 🙇

jd-alexander:
Hi @ashiagr I am going to create an issue to track this but in the meantime we can merge this in once you are good. Yes, this seems like a discrepancy that should be addressed.

ashiagr:
Thanks for creating the [issue](https://github.com/wordpress-mobile/WordPress-Android/issues/12086). This was the only thing I wanted to get addressed for this PR. Approving rest of the changes 👍.

PR #12013, opened 2020-05-25
Prepublishing Nudges : Disable Publish Date on Private Posts. 

### LINKED ISSUE #11889: Prepublishing Nudges : Refactor Post Visibility to match Post Status

jd-alexander:
Fixes #11889

## Solution
1. The Publish Date Label is bind to the `PostStatus` so that it knows it should be `Immediately` once the post is a `PRIVATE` one. 

2. When we are changing the `PostStatus` to `PRIVATE` we always overwrite the date to be sure that is is `Now` or the `CurrentDate` so the `Immediately` label is returned. 

3. The `UiState` powering the home menu actions were given an `enabled` flag so the enabled/disabled state could be reflecting in the color of the text and the `OnClickListener` enabled state could be toggled.

## Testing

1. Create a new post. 
2. Click Publish. 
3. Change the `Visibility` to `Private`. 
4. Observe that the publish date menu action is grey and it can't be clicked and it says `Immediately`. 

Screenshot | Functionality In Action 
--------|-------

------------

1. Create a new post. 
2. Click Publish.
3. Change the `Publish Date` to one in the future. 
4. Observe that the date is seen in the Publish Date label and it can be clicked.
3. Now change the `Visibility` to `Private`. 
4. Observe that the publish date menu action is grey and it can't be clicked and it says `Immediately`. 

Before | After 
--------|-------

## Reviewing 
2. Ensure that https://github.com/wordpress-mobile/WordPress-Android/pull/11889 is merged. 
5. Merge :) 

## Submitter Checklist


WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/visibility/usecases/UpdatePostStatusUseCase.kt
```diff
@@ -31,6 +32,12 @@ class UpdatePostStatusUseCase @Inject constructor() {
 
         editPostRepository.updateAsync({ postModel: PostModel ->
             postModel.setStatus(postStatus.toString())
+
+            // since PRIVATE posts always have to be immediate we set the current date when updating to be sure.
```
malinajirka, line 36: Do you please have a link to the location where this logic was located before?

>to be sure

Imho it's good practice to avoid adding code "to be sure". It either needs to be there or it doesn't. Including these types of "safety" branches makes the code very difficult to refactor in the future. It basically says that even the original author of the code wasn't sure if it was necessary - it's ten times more difficult to find out for the person who wants to re-use/refactor the code. Wdyt?

WordPress/src/test/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModelTest.kt
```diff
@@ -284,6 +284,39 @@ class PrepublishingHomeViewModelTest : BaseUnitTest() {
...
+
+        viewModel.start(editPostRepository, site)
+
+        val uiState = getHomeUiState(TAGS)
+
+        assertThat(uiState?.actionEnabled).isTrue()
+    }
+
+    @Test
+    fun `verify that VISIBILITY action is disabled if PostStatus is PRIVATE`() {
```
malinajirka, line 310: I think the name of this test is supposed say `action is enabled`

WordPress/src/test/java/org/wordpress/android/ui/posts/prepublishing/visibility/usecases/UpdatePostStatusUseCaseTest.kt
```diff
@@ -77,6 +82,18 @@ class UpdatePostStatusUseCaseTest : BaseUnitTest() {
         assertThat(editPostRepository.getPost()?.status).isEqualTo(expectedPostStatus.toString())
     }
 
+    @Test
+    fun `verify that when updatePostStatus is called with PRIVATE Visibility dateTimeUtilsWrapper is used`() {
```
malinajirka, line 86: We might want to consider replacing this test with 
`verify that when updatePostStatus is called with PRIVATE Visibility the creation date is set to now`. Mocking the `currentTimeInIso8601` and verifying it's set to the mocked value will also verify that the method was called.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewHolder.kt
```diff
@@ -35,8 +36,14 @@ sealed class PrepublishingHomeViewHolder(
             uiState.actionResult?.let { resultText ->
                 actionResult.text = uiHelpers.getTextOfUiString(itemView.context, resultText)
             }
-            actionLayout.setOnClickListener {
-                uiState.onActionClicked.invoke(uiState.actionType)
+
+            if (uiState.actionEnabled) {
```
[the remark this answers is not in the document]
jd-alexander: Yes, thanks for the reminder this is a really good point. I will implement both states so we can be sure that the recycled views behave as expected.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewHolder.kt
```diff
@@ -35,8 +36,14 @@ sealed class PrepublishingHomeViewHolder(
...
+                actionLayout.setOnClickListener {
+                    uiState.onActionClicked.invoke(uiState.actionType)
+                }
+            } else {
+                actionType.setTextColor(ContextCompat.getColor(itemView.context, R.color.gray_20))
```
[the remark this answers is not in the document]
jd-alexander: When you say passing even the disabled state. Oh you mean I should ensure that the colors for the text and the click being enabled or not are clearly defined right. I just realized ‘actionEnabled’ is still hiding what it actually does. Let me know if I’m my thoughts are correct.

WordPress/src/test/java/org/wordpress/android/ui/posts/prepublishing/visibility/usecases/UpdatePostStatusUseCaseTest.kt
jd-alexander: Exactly. Thanks. I woke up with this very thought 😅

WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/home/usecases/GetPublishDateLabelUseCase.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.wordpress.android.R
+import org.wordpress.android.fluxc.model.post.PostStatus.PRIVATE
+import org.wordpress.android.ui.posts.EditPostRepository
+import org.wordpress.android.ui.posts.PostSettingsUtils
+import org.wordpress.android.ui.utils.UiString
+import org.wordpress.android.ui.utils.UiString.UiStringRes
+import org.wordpress.android.ui.utils.UiString.UiStringText
+import javax.inject.Inject
+
+class GetPublishDateLabelUseCase @Inject constructor(private val postSettingsUtils: PostSettingsUtils) {
```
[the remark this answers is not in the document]
jd-alexander: Yes, this is what we should be doing. I am just realizing that the contents in this use case need to shared in the `PostSettingsUtils` so that the `PostSettings` screen can behave similarly to this bottom sheet. Thank you so much for pointing this out.

WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/visibility/usecases/UpdatePostStatusUseCase.kt
jd-alexander: > Do you please have a link to the location where this logic was located before?

So I came up with this logic on my own based on the discussion we had in the prepublishing nudges channel. This is a new approach we are utilizing to ensure that the `PRIVATE` posts can't be scheduled or to be more specific, `PRIVATE` posts won't give the impression that they are scheduled. 

> Imho it's good practice to avoid adding code "to be sure". It either needs to be there or it doesn't. Including these types of "safety" branches makes the code very difficult to refactor in the future. It basically says that even the original author of the code wasn't sure if it was necessary - it's ten times more difficult to find out for the person who wants to re-use/refactor the code. Wdyt?

This is a really good point. So answer your main concern this code definitely needs to be there. When I said just to be sure I meant that a post can be in the state where it's being published `Immediately` but it can also be in a state where it was being `SCHEDULED` and then the `PostStatus` is changed to `Private` so we should always clear a future date if a post is set to a `PostStatus` of `PRIVATE`. This logic will also need to be shared as well, so I am thinking of moving it somewhere else because the `PostSettings` should have this behavior.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewHolder.kt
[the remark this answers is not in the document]
malinajirka: Yeah, that's exactly what I meant. It's not necessary, it just felt more consistent, since we pass the strings for actionType and actionResult.

WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/visibility/usecases/UpdatePostStatusUseCase.kt
malinajirka: Thanks for the clarification! ;)

>This logic will also need to be shared as well, so I am thinking of moving it somewhere else because the PostSettings should have this behavior.

That'd be great! ;) 

Just a double-check, I don't recall how these creation dates work. Won't this code modify the date even when we edit a year old private post or is it a desired behavior?

jd-alexander: > That'd be great! ;)
Awesome man 😄 

>Just a double-check, I don't recall how these creation dates work. Won't this code modify the date even when we edit a year old private post or is it a desired behavior?

This a really good point. We should check in with the iOS @leandroalonso  about this as well. I just looked at their PR and it doesn't seem like they took this into consideration. I am trying to think of another approach that we could take. How about only setting it to `Immediately` if the date was `Scheduled` i.e the date is scheduled for the future? So if the user selects a date that's in the past or modifies a post in the past that date is still shown?

leandroalonso: > Won't this code modify the date even when we edit a year old private post or is it a desired behavior?

At least in iOS, yes, this will change the date. And I'd say no, it's not a desired behavior. 😬

jd-alexander: Awesome. Love how we addressed this team @leandroalonso @malinajirka 😄

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewHolder.kt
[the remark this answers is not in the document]
jd-alexander: Resolved here 14646e1

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewHolder.kt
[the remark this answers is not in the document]
jd-alexander: Resolved here 
14646e1

WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/home/usecases/GetPublishDateLabelUseCase.kt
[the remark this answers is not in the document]
jd-alexander: I deleted this class and left the logic within the `PostSettingsUtils` untouched because it worked well.

WordPress/src/test/java/org/wordpress/android/ui/posts/prepublishing/visibility/usecases/UpdatePostStatusUseCaseTest.kt
jd-alexander: Resolved this by creating tests for the new use case 937f9e6

WordPress/src/test/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModelTest.kt
jd-alexander: Resolved here 9f6cd9f

jd-alexander:
@malinajirka 

In this PR 

- I fixed the view holder issues. 
- I updated the publish date logic that was using a new use case unnecessarily. 
- Post Settings now responds to changing the Post Status to Private. 

Main things to test on the Prepublishing Nudges and Post Settings are as follows: 
( the test instruction refers to the bottom sheet so you can just use the same flow/instructions to test the Post Settings.) 

1. Create a new post. 
2.Click Publish. 
3. Change the Publish Date to one in the future. 
4. Go back to the bottom sheet. 
5. Change the Visibility to Private.
6. Ensure that the Publish Date says Immediately. 

-------------------

1. Create a new post. 
2.Click Publish. 
3. Change the Publish Date to one in the past. 
4. Go back to the bottom sheet. 
5. Change the Visibility to Private.
6. Ensure that the Publish Date has the date and it doesn't say Immediately. 

---------------------

1. Create a new post. 
2.Click Publish. 
3. Change the Publish Date to one in the future. 
4. Go back to the bottom sheet. 
5. Change the Visibility to Private.
6. Ensure that the Publish Date says Immediately. 
7. Try to Click the Publish Date and ensure it's disabled.

jd-alexander:
This is the only thing I did I wasn't so sure about. I tried disabling the Publish Date Text Views as seen on the left below but they did not look like how normal disabled controls usually look. (Please ignore the tags and visibility being disabled) but on the right hand side you see me using the text colors that I had to acquire using a color picker to define the values in the XML and that is what I am not sure about. Let me know your thoughts. 

`.setEnabled(boolean enabled)` | Using colors.  
--------|-------
  <kbd></kbd>      |      <kbd></kbd>

jd-alexander:
Ready for another review @malinajirka I used the colors from our palette. Thanks for the suggestion 👍 

Before | After
--------|-------
 <kbd></kbd>      |     <kbd></kbd>

PR #12020, opened 2020-05-25
Homepage settings in site settings

planarvoid:
This PR adds homepage settings to site settings. There are two states:
- Classic blog - where your homepage is your blog posts 
- Static homepage - where you can select a page that will be your homepage and your posts page

You cannot select the same page for both. Both can be not selected. In that case the API uses 0 to represent this field. 

The server remembers your previous selection of homepage/posts page even if you select "Classic blog" settings and save it. When you change it back to "Static homepage", it will be there.

To test:
1. Go to a site with Classic blog settings
2. go to Site Settings
3. Notice there is a Homepage settings item under the General items
4. Notice it has a summary "Classic blog"
5. Click on Homepage settings
6. A dialog is opened
7. The "Classic blog" radio button is selected
8. Click on "Static homepage"
9. Notice two Drop down items appear on the screen
10. Select a site for homepage/posts page
11. Click on "Accept" 
12. Notice that the summary now states "Static homepage"
13. Reopen the dialog
14. Notice that the radio item "Static homepage" is set and the drop down items show your previous selection

To test:
Use steps 1 - 10 from the previous test
11. Click on "Cancel
12. Notice that the summary has not changed
13. Reopen the dialog
14. Notice that the radio item "Classic blog" is still set

To test:
Use steps 1 - 8 from the first test
9. Select the same page for both "Homepage" and the "Posts page" 
10. Notice that there is now an error and the "Accept button is disabled"

To test:
Try combinations of the steps above, try slowing down your internet so you see that there is a progress bar and the "Accept" button is disabled when you're saving the result.

PR submission checklist:


planarvoid:
Thanks for the review @develric ! This is a great catch and it's actually something I forgot to implement. The homepage settings should be hidden on self-hosted sites. I've removed it and also made sure that the crash can't happen if the `showOnFront` parameter is missing in this commit 303bb8b

PR #12031, opened 2020-05-26
WP Stories integration: UploadService trigger

mzorz:
Builds on top of #11987

This PR implements the needed changes to listen to a StorySaveResult and then triggering the UploadService to upload the composed images.

For this, a new class is introduced, called `StoryMediaSaveUploadBridge` that is both a `LifecycleObserver` and a `EditorMediaListener`, which carries with this responsibility.

#### How it works
First, the StoryComposerActivity is now creating a backing Post, which will contain the story content in the format of a Gallery.

The post is instantiated and kept empty, given we won't populate its content until later when we know the media Ids for each of the saved story frames.

The StoryMediaSaveUploadBridge is registered as a LifeycleObserver of the Application class, and subscribes to EventBus `StorySaveResult` events in its `onCreate()` observer override.
When all of a Story's frames are successfully saved to disk (which is the responsibility of FrameSaveManager and FrameSaveService in the stories library), this class will then:
- add each of the saved images as Media items to the WP Media Store (which will then be uploaded to the site). When these are created, we obtain a local `mediaId` for each (this is simply the `id` in the MediaStore).
- the Post that was earlier created contains just a gallery shortcode for now (until we implement the actual Gutenberg story block) to demonstrate the functionality for alpha, and we use a placeholder with the localids,  for example
```
[gallery type="slideshow" ids="placeholderLocalId3,placeholderLocalId4,placeholderLocalId5"]
```
- the UploadService takes care of uploading the media. Once these have been uploaded, we obtain the remote MediaId (the actual `mediaId` field in FluxC's MediaStore) and the post content is updated by replacing the placeholders with each remoteId correspondingly, for example:
```
[gallery type="slideshow" ids="190,191,192"]
```
- once media is finished uploading, the UploadService also takes care of uploading the corresponding Post.

- in the case the user decides to abandon the Story creation screen without pubslishing, the backing Post will be discarded. Any frames saved to that point will remain as media in their device (not media in the WP libraries, just on their device as raw image / video files)

### Error handling
- if an error occurs while saving, the UploadService is never triggered. The Story won't be uploaded (and no new media will be created on the site's Media Library) until all of the frames are saved successfullly
- once the Post is ready to upload and it's enqueued to upoload, the UploadService takes care of everything else (retrying, offline support etc) as per with any other regular Post.

### Videos
Videos are not supported yet, given the Gallery only supports images. There's no code in this PR or in the stories library in general to prevent this, given it'd be just adding unnecessary work.  If you add a video to a Story, and it gets saved correctly, etc. it will be uploaded to the site and added to the Gallery, but it won't show in the resulting Post.

To test:
- create a story (only add text to your frames, not emoji for now)
- tap PUBLISH
- observe it gets saved, and then it gets uploaded
- once it's published you'll be able to see it on the web. Tap on the published snackbar "VIEW" button if you'd like.

PR submission checklist:


mzorz:
As discussed, plans are to add tests for this new class. Lifting the `[not ready for review]` flag here.

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
```diff
@@ -0,0 +1,170 @@
...
+
+    @Suppress("unused")
+    @OnLifecycleEvent(ON_DESTROY)
+    fun onDestroy(source: LifecycleOwner) {
+        // note: not sure whether this is ever going to get called if we attach it to the lifecycle of the Application
+        // class, but leaving it here prepared for the case when this class is attached to some other LifeCycleOwner
+        // other than the Application.
+        cancelAddMediaToEditorActions()
+        EventBus.getDefault().unregister(this)
+    }
```
jd-alexander, line 77: Could we use the `eventBusWrapper` here since we are going to be testing this class in the near future? WDYT?

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
```diff
@@ -0,0 +1,170 @@
...
+        // this is similar to addNewMediaItemsToEditorAsync in EditorMedia
+        launch {
+            addLocalMediaToPostUseCase.addNewMediaToEditorAsync(
+                    uriList,
+                    site,
+                    freshlyTaken = false, // we don't care about this
+                    editorMediaListener = this@StoryMediaSaveUploadBridge,
+                    doUploadAfterAdding = true
+            )
+            postUtils.preparePostForPublish(editPostRepository.getEditablePost()!!, site)
```
jd-alexander, line 97: We try to avoid using the `!!` operator. Can we change this logic a bit to handle this null possibility? Especially if we don't expect the `post` to be null here. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
```diff
@@ -0,0 +1,170 @@
...
+            savePostToDbUseCase.savePostToDb(WordPress.getContext(), editPostRepository, site)
+
+            if (networkUtils.isNetworkAvailable()) {
+                postUtils.trackSavePostAnalytics(
+                        editPostRepository.getPost(),
+                        site
+                )
+                uploadService.uploadPost(WordPress.getContext(), editPostRepository.id, true)
```
jd-alexander, line 105: Can we inject this context somehow so that we avoid using the static method call?

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
```diff
@@ -0,0 +1,170 @@
...
+                // SAVED_ONLINE
+            } else {
+                // SAVED_LOCALLY
+            }
```
jd-alexander, line 109: Let me know if we plan to implement this `else`. If not we can remove it 🙏

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
```diff
@@ -0,0 +1,170 @@
...
+    // endregion
+
+    private fun cancelAddMediaToEditorActions() {
+        job.cancel()
+    }
+
+    @Subscribe(sticky = true, threadMode = MAIN)
+    fun onEventMainThread(event: StorySaveResult) {
+        // only trigger the bridge preparation and the UploadService if the Story is now complete
+        // otherwise we can be receiving successful retry events for individual frames we shoulnd't care about just
```
jd-alexander, line 121:
```suggestion
        // otherwise we can be receiving successful retry events for individual frames we shouldn't care about just
```

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
```diff
@@ -0,0 +1,170 @@
...
+import org.wordpress.android.util.helpers.MediaFile
+import javax.inject.Inject
+import javax.inject.Named
+import kotlin.coroutines.CoroutineContext
+
+/*
+ * StoryMediaSaveUploadBridge listens for StorySaveResult events triggered from the StorySaveService, and
+ * then transforms its result data into something the UploadService can use to upload the Story frame media
+ * first, then obtain the media Ids and collect them, and finally create a Post with the Story block (
+ * (simple WP gallery for alpha) with he obtained media Ids.
```
jd-alexander, line 41:
```suggestion
 * (simple WP gallery for alpha) with the obtained media Ids.
```

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
```diff
@@ -0,0 +1,170 @@
...
+        } else {
+            // TODO WPSTORIES add TRACKS for ERROR
+            // AnalyticsTracker.track(Stat.MY_SITE_ICON_UPLOAD_UNSUCCESSFUL);
+        }
+    }
+
+    private fun isStorySuccessfullySavedAndComplete(event: StorySaveResult): Boolean {
+        return (event.isSuccess() &&
+                event.frameSaveResult.size == StoryRepository.getStoryAtIndex(event.storyIndex).frames.size)
+    }
```
jd-alexander, line 144: I was wondering if this logic could be encapsulated elsewhere like in a use case. It seems unrelated to the purpose of the bridge. Let me know, it seems simple so it could rename here.  😄

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
```diff
@@ -0,0 +1,170 @@
...
+                // lets add an EVENT for START UPLOADING MEDIA
+                // AnalyticsTracker.track(Stat.STORIES_BLA_BLA_ADDED_MEDIA_OR_SOMETHING);
+            }
+        } else {
+            // TODO WPSTORIES add TRACKS for ERROR
+            // AnalyticsTracker.track(Stat.MY_SITE_ICON_UPLOAD_UNSUCCESSFUL);
+        }
+    }
+
+    private fun isStorySuccessfullySavedAndComplete(event: StorySaveResult): Boolean {
```
jd-alexander, line 141: Nitpick : Wondering if this name can be simplified 🙏 
```suggestion
    private fun isStorySavingComplete(event: StorySaveResult): Boolean {
```

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
```diff
@@ -0,0 +1,170 @@
...
+    override fun appendMediaFiles(mediaFiles: Map<String, MediaFile>) {
+        // Create a gallery shortcode and placeholders for Media Ids
+        val idsString = mediaFiles.map {
+            WP_STORIES_POST_MEDIA_LOCAL_ID_PLACEHOLDER + it.value.id.toString()
+        }.joinToString(separator = ",")
+        editPostRepository.update { postModel: PostModel ->
+            postModel.setContent("[gallery type=\"slideshow\" ids=\"$idsString\"]")
+            true
+        }
+    }
```
jd-alexander, line 155: This logic isn't clear to eye at a glance so I would want us to encapsulate this in a use case and write some tests to verify it's behavior. If so we could create an issue to track this. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
```diff
@@ -0,0 +1,170 @@
...
+
+    override val coroutineContext: CoroutineContext
+        get() = mainDispatcher + job
+
+    @Inject lateinit var editPostRepository: EditPostRepository
+
+    @Suppress("unused")
+    @OnLifecycleEvent(ON_CREATE)
+    fun onCreate(source: LifecycleOwner) {
+        EventBus.getDefault().register(this)
```
jd-alexander, line 66: What do you think about replacing these `EventBus` references with the wrapper version? 😄

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
```diff
@@ -202,6 +236,22 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
...
+            post.setStatus(DRAFT.toString())
+            post.setPostFormat(POST_FORMAT_WP_STORY_KEY)
+            post
+        }
+        editPostRepository.savePostSnapshot()
+        // this is an artifact to be able to call savePostToDb()
+        // also, Story posts are always PUBLISHED
+        editPostRepository.getEditablePost()?.setStatus(PUBLISHED.toString())
+        site?.let {
+            savePostToDbUseCase.savePostToDb(WordPress.getContext(), editPostRepository, it)
```
jd-alexander, line 251: Same static context comment 🙏

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
mzorz: I think if the Post is `null` at this point we'd rather crash than fail silently. There's not much we can do without a backing Post, so in this case the !! operator has intended meaning. As you said, given especially we don't expect things to be `null` here. Do you have something different in mind? I'm interested in understanding what other approaches exist 🙏 let me know @jd-alexander

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
mzorz: 👍

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
mzorz: yes please see https://github.com/wordpress-mobile/WordPress-Android/pull/12074/files#diff-6b809e71f14c003d525006c098f98625

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
mzorz: Gotcha, thanks for this comment. Even when we're going to remove this part and change it for an actual Gutenberg block (Tug is working on that), it makes sense to encapsulate the "block creation" in a use case as suggested. Would this be okay to do on a separate PR after alpha1 when we focus on refactors / testability or do you think we should do it here? Let me know 🙏 🙇

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
jd-alexander: Thanks for this @mzorz I am learning as I go along about how to handle these null values gracefully. What I have seen in other parts of WP Android is the usage of the `require*` method. Example: 

```kotlin
reportTransactionState(false)
            if (isUpdated) {
                requireNotNull(post).let {
                    _postChanged.value = Event(it)
         } }
```

Wdyt? 😄

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
jd-alexander: Lovely! Thanks for that link 🙏

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
jd-alexander: > Would this be okay to do on a separate PR after alpha1 when we focus on refactors / testability or do you think we should do it here? Let me know 🙏 🙇

Honestly, we can do this after alpha when we focus on the reactors. Once we are tracking these things I am good 😄

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
mzorz: 👍 made the change in 86fb2ad01d - also adding the `removeStickyEvent()` method to the wrapper

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
mzorz: I totally forgot about that one - even when I recently made similar changes in lots of places in 11934 🤕 - addressed in 2d813494bb

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
mzorz: Been checking this, I think we need to make a context factory so it's injectable. I've seen sometimes we use `init()` to pass some variables to ViewModels and other injectable classes in WPAndroid, so went that way for now in 85bda8b237

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
mzorz: 👍 addressed in 414beab72f

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
mzorz: Tracked in the Stories repository project 👍

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
mzorz: did in https://github.com/wordpress-mobile/WordPress-Android/commit/86fb2ad01d5fba438d6c20979d5c6c03f36e2ceb

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
mzorz: did in https://github.com/wordpress-mobile/WordPress-Android/commit/86fb2ad01d5fba438d6c20979d5c6c03f36e2ceb

mzorz: this one addressed in fb65ea2d46

mzorz:
> Hi @mzorz great work on this 😄 I left some comments but overall the solution works really nicely. It was great seeing the story converted to a post!

Thanks a lot for your review! Should be ready for a second round @jd-alexander 🙇

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt
jd-alexander: Awesome. Thanks 😄

Issue #12039, opened 2020-05-27
Reader Improvements: Repository Main Issue


malinajirka:
Parent issue - #12026.

This is a Main Issue for the introduction of (reactive) repository layer. More details can be found on pbArwn-uz-p2. 
The main goal is to hide the implementation of Reader services and Reader actions from the VM layer. VMs will just get a lifedata from a repository, transform it into an UIState and pass the transformed livedata to the view layer.

**BackLog:**

**InProgress:**

**Done:**
- [x]  Fix "load more" todo ([link](https://github.com/wordpress-mobile/WordPress-Android/blob/3c39826a9c299e29737a06bb8a1aef8753245ce7/WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderDiscoverDataProvider.kt#L83))

- [x] Create dummy methods for "Get a list a pre-sorted array of interests" and "Save the selected interests" to unblock development of ["Select topics" screen](https://github.com/wordpress-mobile/WordPress-Android/issues/12027)
- [X] Create dummy methods for "Get a list of related interests" and "Cards based discovery feed (can return just LiveData<List<ReaderPost>> for now)" to unblock development of ["Discover" screen](https://github.com/wordpress-mobile/WordPress-Android/issues/12028)
- [x] *Choose any existing action from ReaderServices eg "Get an array of posts from a single interest" and try to implement the reactive repository approach we discussed
         - load data from the local db when someone starts observing the data
         - request data from the server when someone starts observing the data (if the data haven't been request in lets say last 10 minutes)
         - listen for event bus events which indicate that the data have been changed in the db + update the livedata
- [x] * Choose any existing action from ReaderActions eg. "Like a post"
                    - create a suspendable function which invokes ReaderActions.performLikeAction
                    - return result from the suspendable function when the listener provided to performLikeAction is invoked
- [x] Make sure the VM knows when a request is in progress ([link](https://github.com/wordpress-mobile/WordPress-Android/blob/3c39826a9c299e29737a06bb8a1aef8753245ce7/WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderDiscoverDataProvider.kt#L59))
- [x] Implement PTR functionality
- [x] Send all started/updated events through the communication channel

Open Questions: 
1. How do we let the UI know that the data are being loaded?
2. How do we propagate errors to the VM?
3. What will be the lifecycle of this component?

PR #12046, opened 2020-05-28
Add dummy methods

zwarm:
Part of #12039 

This PR adds two dummy methods to the ReaderTagRepository. 

Open Questions
1. Should a `List<ReaderTag>` be returned directly or will observers subscribe for changes. 
2. Should we using the `ReaderTagList` instead of `List<ReaderTag>`
3. Are there requirements for saving - even for a dummy - request params, etc

No tests needed, only contains stub methods.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderTagRepository.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderTagRepository.kt
@@ -1,3 +1,45 @@
 package org.wordpress.android.ui.reader.repository
 
-class ReaderTagRepository
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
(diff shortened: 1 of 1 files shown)


malinajirka:
Thanks @zwarm for working on this
>Should a List<ReaderTag> be returned directly or will observers subscribe for changes.

I think we shouldn't change the data under user's hands. So imo we should return it directly. (I don't think we'll cache these values as this is a one-time screen.)

>Should we using the ReaderTagList instead of List<ReaderTag>

I'm not 100% sure about this. But I'd probably use it for consistency.

>Are there requirements for saving - even for a dummy - request params, etc

I think the dummy methods should be as simple as possible. It doesn't really matter what data they return if they return something.

zwarm:
Thanks @malinajirka for the feedback. I've updated in 34b0db3 to directly return data as ReaderTagList. I'm removing the draft label.

zwarm:
> I'm sorry I missed this before. But it'd be nice to add a contructor with @Inject annotation ...
 Me too - wasn't thinking that far ahead yet, but great catch.

> We could also consider injecting the `Dispatchers.IO`, but that's a minor thing as this class is far from its final state. 
I suspect we'll be injecting quite a few things in its final state :)

Thanks for the 👍

PR #12047, opened 2020-05-28
Unified Login & Sign-Up: Login library theme and styles update

### LINKED ISSUE #11785: Update the UI of the login flow
The task is to update the UI of all the steps of the login/signup flow. This flow uses shared components so the screen changes will be overlapping
- The toolbar is changing, instead of a logo it now contains a title with the flow you're in ("Get Started" changes either into "Log In" or into "Sign Up"), the help is now an icon
- The action buttons are moving to the bottom of the screen
- The colors and styling are changing
- Go step by step and compare the screens to the desings to see all the changes
- If possible, the changes should be behind a feature flag

Continue with WordPress com

renanferrari:
Fixes #11785

Depends on #11944

This PR updates the Login Library's theme and various component styles, and also bumps its `minSdkVersion` to 21 and `targetSdkVersion` to 29, since those are already the versions used by both of its clients apps (WPAndroid and WCAndroid).

The theme updates make use of the new Material Design system for colors, typography and shapes, and is now better separated, which should make a lot easier to extend it and override specific theme attributes and styles from now on. It also includes a dark theme.

Both the theme and the styles were renamed to reflect best practices regarding theming on Android, so most of the changes on layout files are due to those changes.

The key changes are in the following files under the `res/values` folder of the login library:
- styles.xml
- themes.xml
- types.xml

Old login theme overloads were removed from the app, since the new theme was based on the look we currently want to achieve.

## To test

These changes are mostly on the code side rather than on the UI, but you can simply navigate through the login flow and make sure its appearance is mostly the same as it was here: #11944.

## Notes
- This PR does not include any changes do the login flow screens layouts. Those are being handled separatedly here: #12048
- Styles that were being used only by the prologue and the epilogue screen were removed from the library, since it doesn't include those screens.

---

PR submission checklist:


renanferrari:
>This is so much better than it was before! And it looks really good 👍 Thanks for the change. 

No problem, glad you liked it! 😊

>Do you know how it works with Woo? Do they have to change their styles?

Yes. They will have to change their theme and styles, but the changes being introduced here should actually make things easier for them, for two main reasons:
1. They would already have to change the styles used by their app anyway, as they will soon need to implement an unified flow based on the one we are introducing here (internal ref: p91TBi-2KJ-p2). This means that most of the changes they would have to make first, will already be implemented by the library itself (only needing a few tweaks).
2. They are already using a very similar theming scheme on their app and a lot of the changes here were actually based on what they currently have over there.

The way both apps were using the library before was like working _around_ the default library theme instead of working _with_ it. Hopefully this will change that.

That being said, I think it's probably reasonable for us to get another pair of eyes from Woo on both this PR and the next one (#12048) before we merge them. I'll make sure of that.

renanferrari:
Just adding a note here that we may need to override the login theme on the WP app and add some of our main theme custom attributes there, as to avoid crashes when accessing screens missing those attributes. Another option would be to change affected layouts to avoid using said attributes. I will make sure to expand on this and handle it before merging this PR.

renanferrari:
@AmandaRiu Thanks a lot for reviewing these, really appreciate it! 😊

PR #12061, opened 2020-06-01
App config annotation processing

planarvoid:
This PR adds Annotation processing to generate the Remote config defaults for the Firebase remote config. I've created two new modules for annotations and processor. Since this is the first time I've created a module, it would be great if you could let me know if they make sense (naming, location, purpose). 

The code generates the `RemoteConfigDefaults.kt` that is used to define default values of the experiment/feature. This object exposes a map of remote config defaults for the firebase.

To test:
- Add an Experiment/Feature based on the Examples
- Make the project
- Make sure the `RemoteConfigDefaults.kt` is there and contains your new experiment/feature with the default value.

PR submission checklist:


libs/WordPressProcessors/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+ext.kotlinVersion = '1.3.61'
+
+kapt {
+    generateStubs = true
+}
+
+dependencies {
+    implementation project(':libs:WordPressAnnotations')
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
+    implementation "com.google.auto.service:auto-service:1.0-rc6"
```
malinajirka, line 13: We might want to consider defining all the dependencies' versions in the main build.gradle - [here](https://github.com/wordpress-mobile/WordPress-Android/blob/4c855834299c176a2711e3af22efc1183c4252b9/build.gradle#L116). Wdyt?

planarvoid:
@malinajirka thanks for the check! I've moved the library versions to WPAndroid where it made sense

PR #12068, opened 2020-06-01
WP Stories integration - implement StoryNotificationTrackerProvider interface

mzorz:
This PR builds on top of https://github.com/wordpress-mobile/WordPress-Android/pull/12059

This PR implements the new `StoryNotificationTrackerProvider` interface, which allows the host app to be signaled of when a Notification is being produced by the StorySaveService in the Stories library. Then, it can properly channel the event through to its own `SystemNotificationsTracker`.

Also in this PR we're checking `ARG_NOTIFICATION_TYPE` for when an error notification is tapped and properly tracking it down in StoryComposerActivity (see onCreate() in that class).

To test:
1. inject an error when saving an image in PhotoEditor (for example introduce a line that throws an Exception in `saveImageFromPhotoEditorViewAsLoopFrameFile().`
2. run the WP app, create a new story with an image background frame and tap PUBLISH
3. observe the error notification is shown, and also the `trackShownNotification` method implemented in the WordPress class is executed. (placing a breakpoint there should be enough).
4. Finally, tap on the error notification and verify the `trackTappedNotification` method is executed as well.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/WordPress.java
```diff
@@ -995,4 +1005,32 @@ public void onTrimMemory(final int level) {
             }
         }
     }
+
+    private class StoryNotificationTrackerProvider implements NotificationTrackerProvider {
```
jd-alexander, line 1009: Hey @mzorz what do you think about putting this class into another file? Just checking because this class is getting a bit big. Let me know though because it's also a really small class 🙏

PR #12069, opened 2020-06-01
Add stub functions to support reader improvement dev

zwarm:
Part of #12039 

This PR adds new dummy methods to the ReaderTagRepository and the ReaderPostRepository.

Open Questions
1. Are the methods in the correct classes? 
2. Not sure if `recommendedInterests' should be a ReaderTagList or actually posts?
3.`getDiscoveryFeed()`returns a `LiveData<List<String>>` for now, but I wonder if this needs to be ReaderPostList or something entirely new. 

No tests needed, only stub methods.

PR submission checklist:


malinajirka:
>1. Are the methods in the correct classes?

LGTM

>2. Not sure if recommendedInterests' should be a ReaderTagList or actually posts? 

We'll display a card with a list of tags the user can click on -> ReaderTagList LGTM.

>3.getDiscoveryFeed()returns a LiveData<List>` for now, but I wonder if this needs to be ReaderPostList or something entirely new.

I think it'll be something entirely new in the future. There is a chance the "getDiscoveryFeed" will return a structure which will contain both "recommended interests" and "discover posts". However, I'd vote for keeping them separate for now and returning "ReaderPostList" from `getDiscoveryFeed`.

Issue #12027, opened 2020-05-26
Reader Improvements: Select Topics Main Issue


malinajirka:
Parent issue - https://github.com/wordpress-mobile/WordPress-Android/issues/12026.

This is a Main Issue for the "Select Topics" screen. Goal of this screen is to provide the user with a list of populars topics and let them choose a few. The topics they choose will be used to create a "tailored" content on the discover tab.

This screen should be shown only to users who don't follow any tags (new users or users who haven't used Reader much).

**BackLog:**

**InProgress:**

**Done:**
- [x] Create an empty fragment, VM and VMTest #12049
- [x] Show a list of "chips" with topics on the UI - get mocked list of topics from ReaderRepository #12060
- [x] Make the "chips" selectable and propagate selected items into the VM #12073
- [x] Add tracking - tbd
- [x] Introduce "Done" button - pass the selected topics to ReaderRepository (the method in ReaderRepository can be empty) #12073
- [x] Introduce "Loading popular topics"  screen/state #12113
- [x] Introduce "Loading popular topics" failed screen/state #12201
- [x] Navigate to Discover #12218
- [x] Fetch List of user's tags and show "Select Topics" screen only if the list is empty - get mocked list of tags from ReaderRepository #12255
- [x] Polish the UI  #12296

Open Questions:
- Will we show this screen in fullscreen or just as a content of the "discover" tab? -> Most likely as fullscreen

PR #12073, opened 2020-06-02
RI: Propagate selected "chip" state into the VM, introduce "Done" button

ashiagr:
Master issue: #12027

This PR includes following tasks:
- [ ] Make the "chips" selectable and propagate selected items into the VM
- [ ] Introduce "Done" button - navigate to Discover and pass the selected topics to ReaderRepository

### To test:
- Go to Reader tab
- Notice that Done button is initially hidden
- Wait for the interests "chips" to load
- Notice that Done button is disabled with text "Select a few to continue"
- Select few interest chips
- Notice that Done button is enabled with text "Done"
- De-select all interest chips
- Notice that Done button is disabled with text "Select a few to continue"
- Select an interest chip to enable Done button and click it
- Existing Reader is shown with the Discover tab selected (yet to replace Discover tab with the new blank Discover screen)

done

Note:

Need to refine
1. styling
2. transition to Discover screen
will do in upcoming PRs.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
```diff
@@ -1,57 +1,151 @@
...
+    private val _navigateToDiscover = MutableLiveData<Event<Unit>>()
+    val navigateToDiscover: LiveData<Event<Unit>> = _navigateToDiscover
+
+    private val selectedInterests = ReaderTagList()
+
     fun start() {
-        if (initialized) return
+        if (initialized) {
+            val uiState = uiState.value as UiState
+            updateUiState(uiState.copy(interestsUiState = getInterestsUiStateWithSelectedStates()))
```
ashiagr, line 37: It is currently not possible to get list of selected ids from `ChipGroup` included with `material lib 1.1.0 ` https://git.io/JfX4j. So I'm saving selected items inside `selectedInterests`. Also `ChipGroup` doesn't seem to restore selected states for all chips on orientation change (just restores first selected chip), so I'm updating it with the saved selected states inside `onStart`. Let me know if you have any suggestion.

I can revisit this logic when we upgrade to `material lib ver >= v1.2.0-alpha01 ` where `ChipGroup` includes a [method to get the selected ids](https://git.io/JfXBq).

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsFragment.kt
```diff
@@ -24,33 +28,78 @@ class ReaderInterestsFragment : Fragment(R.layout.reader_interests_fragment_layo
...
     }
+
+    private fun navigateToDiscover() {
+        val fragmentTransaction = parentFragmentManager.beginTransaction()
+        fragmentTransaction.setCustomAnimations(
+            R.anim.activity_slide_in_from_right, R.anim.activity_slide_out_to_left,
+            R.anim.activity_slide_in_from_left, R.anim.activity_slide_out_to_right
+        )
+        fragmentTransaction.replace(R.id.fragment_container, ReaderDiscoverFragment(), tag).commit()
+    }
```
ashiagr, line 104: Not sure which transition animation we should use here. I've temporarily used slide animation.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
```diff
@@ -1,57 +1,151 @@
...
 import javax.inject.Inject
 
 class ReaderInterestsViewModel @Inject constructor(
     private val readerTagRepository: ReaderTagRepository
 ) : ViewModel() {
     var initialized: Boolean = false
 
-    private val _uiState = MutableLiveData<UiState>()
+    private val _uiState: MutableLiveData<UiState> = MutableLiveData(
+        UiState(listOf(), ReaderTagList(), DoneButtonHiddenUiState)
```
ashiagr, line 25: Should I convert `UiState` to include `InitialUiState` and `ContentUiState`? Can do that in an upcoming PR when I add the loading state.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
malinajirka: I'm not sure I understand this part. Why do we need to update the uiState in `start`? AFAIU `start` is invoked after configuration change and `uiState: LiveData<UiState>` will retain its state, right?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
```diff
@@ -1,57 +1,151 @@
...
+            val enableDoneButton = selectedInterests.isEmpty() && isChecked
+            val disableDoneButton = selectedInterests.size == 1 && !isChecked
+
+            updateSelectedInterests(it.interests[index])
+
+            if (enableDoneButton || disableDoneButton) {
+                val uiState = uiState.value as UiState
+                updateUiState(
+                    uiState.copy(
+                        interestsUiState = getInterestsUiStateWithSelectedStates(),
```
malinajirka, line 72: AFAIU we are updating the uiState only when the state of the done buttons changes. It means that there are scenarios when the uiState doesn't correspond to the state of the UI (selected items). It feels like it goes against one of the main benefits of using UiState. Is this intentional? Can't we always keep it up-to-date? Am I missing something? Wdyt?

WordPress/src/test/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModelTest.kt
```diff
@@ -66,20 +74,134 @@ class ReaderInterestsViewModelTest {
...
+            // Given
+            val mockInterests = getMockInterests()
+            whenever(readerTagRepository.getInterests()).thenReturn(mockInterests)
+
+            // When
+            initViewModel()
+            // Select an interest
+            viewModel.onInterestAtIndexToggled(index = 0, isChecked = true)
+            // Then
+            assertThat(requireNotNull(viewModel.uiState.value).interestsUiState.filter { it.isChecked }.size)
```
malinajirka, line 146: I feel like these two assertions don't belong to this test. These assertions are already being tested in the test above -> each the test should ideally have a single purpose/focus/reasonToFail. Wdyt?

WordPress/src/test/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModelTest.kt
```diff
@@ -66,20 +74,134 @@ class ReaderInterestsViewModelTest {
...
+            // Given
+            val mockInterests = getMockInterests()
+            whenever(readerTagRepository.getInterests()).thenReturn(mockInterests)
+
+            // When
+            initViewModel()
+            viewModel.onInterestAtIndexToggled(index = 0, isChecked = true)
+
+            // Then
+            assertThat(requireNotNull(viewModel.uiState.value).interestsUiState.filter { it.isChecked }.size)
```
malinajirka, line 123: I feel this assertion doesn't belong to this test. Each test should ideally have a single purpose/focus/reasonToFail.  I'd consider creating a new test "Interest selected after onInterestAtIndexToggled invoked" or similar. Wdyt?

WordPress/src/test/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModelTest.kt
```diff
@@ -66,20 +74,134 @@ class ReaderInterestsViewModelTest {
...
+    @ExperimentalCoroutinesApi
+    @Test
+    fun `done button hidden on start switches to disabled state when tags received from repo`() =
+        coroutineScope.runBlockingTest {
+            // Given
+            val mockInterests = getMockInterests()
+            whenever(readerTagRepository.getInterests()).thenReturn(mockInterests)
+
+            // Pause dispatcher so we can verify done button initial state
+            coroutineScope.pauseDispatcher()
```
malinajirka, line 95: TIL ❤️

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsFragment.kt
```diff
@@ -24,33 +28,78 @@ class ReaderInterestsFragment : Fragment(R.layout.reader_interests_fragment_layo
...
+            text = getString(doneButtonUiState.titleRes)
+        }
+        uiHelpers.updateVisibility(done_button, doneButtonUiState.visible)
+    }
+
     private fun updateInterests(interestsUiState: List<InterestUiState>) {
         interestsUiState.forEachIndexed { index, interestTagUiState ->
-            val chip = layoutInflater.inflate(R.layout.reader_interest_filter_chip, interests_chip_group, false) as Chip
+            var chip = interests_chip_group.findViewWithTag<Chip>(interestTagUiState.title)
+            if (chip == null) {
```
malinajirka, line 73: We might want to consider moving the body of this `if` into a separate method, wdyt?

We could write something like 
`val chip = interests_chip_group.findViewWithTag<Chip>(interestTagUiState.title) ?: createChipView(...)`

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsFragment.kt
```diff
@@ -24,33 +28,78 @@ class ReaderInterestsFragment : Fragment(R.layout.reader_interests_fragment_layo
...
+                ) as Chip
+                with(chip) {
+                    layoutDirection = View.LAYOUT_DIRECTION_LOCALE
+                    tag = interestTagUiState.title
+                    setOnCheckedChangeListener { compoundButton, isChecked ->
+                        if (compoundButton.isPressed) {
+                            viewModel.onInterestAtIndexToggled(index, isChecked)
+                        }
+                    }
+                    interests_chip_group.addView(chip)
```
malinajirka, line 87: I'm not 100% sure, but could we simplify this by removing this line and setting `attachToRoot` parameter on `layoutInflater.inflate` to `true`?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsFragment.kt
```diff
@@ -24,33 +28,78 @@ class ReaderInterestsFragment : Fragment(R.layout.reader_interests_fragment_layo
...
-                }
+                isChecked = interestTagUiState.isChecked
             }
-
-            interests_chip_group.addView(chip)
         }
     }
+
+    private fun navigateToDiscover() {
+        val fragmentTransaction = parentFragmentManager.beginTransaction()
```
malinajirka, line 98: We might want to consider using the ActivityLauncher and letting the parent handle the navigation. It feels like it's not responsibility of this fragment to replace it with "ReaderFragment". Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
```diff
@@ -1,57 +1,151 @@
...
+
     private fun transformToInterestsUiState(interests: ReaderTagList) =
         interests.map { interest ->
-            InterestUiState(
-                interest.tagTitle
-            )
+            InterestUiState(interest.tagTitle)
+        }
+
+    private fun getInterestsUiStateWithSelectedStates(): List<InterestUiState> {
```
malinajirka, line 100: Do we needs this method? Do we even need the `selectedInterests` field? Can't we simply store the check/notChecked state inside the uiState and consider the uiState as the single source of truth?

ashiagr:
Hey 👋

Thanks for the review! Just want to quickly address these comments:

> I'm not sure I understand this part. Why do we need to update the uiState in start? AFAIU start is invoked after configuration change and uiState: LiveData<UiState> will retain its state, right?

> AFAIU we are updating the uiState only when the state of the done buttons changes. It means that there are scenarios when the uiState doesn't correspond to the state of the UI (selected items). It feels like it goes against one of the main benefits of using UiState. Is this intentional? Can't we always keep it up-to-date? Am I missing something? Wdyt?

> Do we needs this method? Do we even need the selectedInterests field? Can't we simply store the check/notChecked state inside the uiState and consider the uiState as the single source of truth?

As we use "Single UI State" principle, refreshing entire UI for every single "chip" click didn't seem quite right, especially when `ChipGroup` already highlights the chip when we click it and vice versa. So I just saved selected states into `selectedInterests` list and utilised it when we really want to update the UI in scenarios like **refreshing the Done button** state or when **UI is recreated**.

Also in the newer material lib `ChipGroup` includes a method to get all the selected ids -> so instead of storing each individual chip state in the VM, we might just pass these `selectedIds` to the VM on done button click. Since we don't have this method now, I'm using `selectedInterest` list so that it is easier to migrate later.

Let me know what you think of it. I'll update the logic accordingly.

malinajirka:
>As we use "Single UI State" principle, refreshing entire UI for every single "chip" click didn't seem quite right, especially when ChipGroup already highlights the chip when we click it and vice versa.

I'm a bit worried if someone else reads the code and doesn't have the context it might be quite difficult to use for them. For example they can rely on the "isChecked" state but it won't work in certain scenarios. 
I'd personally consider one of the two following options:
- Remove `isChecked` from the UIState to make sure we have single source of truth and the state is never de-synced
- Always update the UIState to make sure the `isChecked` is always synced with the UI (we can come up with some optimizations if it has any performance impact)


SideNote: 
>refreshing entire UI for every single "chip" click

Tbh I'm not sure - does it actually refresh the whole UI? I think if you set the same state on a view (eg. setChecked(true) on a view which was already checked) Android won't refresh it. Wdyt?

ashiagr:
Thanks, I'll go with the second option for better code readability:

> Always update the UIState to make sure the isChecked is always synced with the UI (we can come up with some optimizations if it has any performance impact)

> Tbh I'm not sure - does it actually refresh the whole UI? I think if you set the same state on a view (eg. setChecked(true) on a view which was already checked) Android won't refresh it. Wdyt?

Not sure, I can test how it works.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsFragment.kt
ashiagr: Refactored here: 36dab46

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
ashiagr: Refactored here: bfaaee9

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
ashiagr: Refactored here: bfaaee9

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
ashiagr: Refactored here: bfaaee9

WordPress/src/test/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModelTest.kt
ashiagr: Refactored here: a456645

WordPress/src/test/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModelTest.kt
ashiagr: Refactored here: a456645

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsFragment.kt
ashiagr: Tried setting true, inflator returns `ChipGroup` instead of `Chip` layout. Can we keep it this way?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsFragment.kt
ashiagr: Can I do it in an upcoming PR? I need to figure how to replace a fragment in the `WPMainNavigationView` and don't want to block this PR because of it.

malinajirka: We might want to consider replacing it in the ReaderFragment - I'm not sure. Either case, definitely a good idea to do this in a separate PR.

PR #12085, opened 2020-06-02
Show keyboard upon entry

### LINKED ISSUE #12084: Show keyboard on search domain screen
### Expected behavior
When landing on the domain search screen I expect the keyboard to be displayed

### Actual behavior
The keyboard is not displayed and the user has to tap on the textfield

1. On "My site" choose "switch site"
2. Choose "Create WordPress.com site 
3. Choose any kind of site
4. See the domain search screen where you have to tap on the textfield to show the keyboard

zwarm:
Fixes #12084 

This PR sets focus on the search input field and opens the keyboard upon entry to the search domain view. The `SiteCreationSearchInputUIState` now includes a showKeyboard field, so if for someone reason we don't want to show the keyboard on open we just need to set the field to false.

To test:
1. On "My site" choose "switch site"
2. Choose "Create WordPress.com site
3. Choose any kind of site
4. On entry to the domain search screen, the focus should be on the input text and the keyboard should be showing

NOTE: Existing behavior leaves keyboard open at all times incase the user wants to add or remove characters in the search field. If this changes, we can easily modify the code to close the keyboard as well.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/sitecreation/domains/SiteCreationDomainsViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/sitecreation/domains/SiteCreationDomainsViewModel.kt
@@ -193,7 +193,8 @@ class SiteCreationDomainsViewModel @Inject constructor(
                         searchInputUiState = createSearchInputUiState(
                                 showProgress = state is Loading,
                                 showClearButton = isNonEmptyUserQuery,
-                                showDivider = state.data.isNotEmpty()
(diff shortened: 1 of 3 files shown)


zwarm:
> The keyboard is re-shown after a configuration change. But I don't think it's an issue.

Yes, and I noticed it too and should have added that bit into the existing code note. My apologies.  _Supposedly_ in Android 11 there will be a way to detect if the keyboard is open. 🤞  Thanks for the review

PR #12089, opened 2020-06-03
Unified login tracking 1 - Screens

### LINKED ISSUE #11797: Unified Login & Sign-Up: Update tracking events

planarvoid:
Fixes #11797

This PR adds screen tracking based on the definition. If you agree, I'll write down the logic later. I've created one event `unified_login_step` which defines every single screen shown during the login/signup. This event has 3 properties. The first one is `source` (which wasn't defined in the requirements but I think it makes sense). It tracks the source of the login flow (whether it comes from a share action/Jetpack and so on). The second one is `flow` which defines whether the flow is login, signup, login with magic link or password and so on. All of these flows have the third property - `step`. Step is a single screen in the flow. 
There were a few steps/flows missing from the requirements, for example Social login. 
Let me know if these properties make sense!

I've created a singleton that behaves like a state machine and keeps the current flow. My thinking behind this is that some of the steps are shared between the flows and to know which flow we're in would require some heavy logic changes (pass the flow around between the fragments). My approach seems like much cleaner solution. Let me know what you think.

To test (the steps are not well defined because we need to track everything):
- Add filter for `unified_login_step` to `LogCat`
- Go through all the login/signup flows you can imagine
- Check that every screen view emits an event that makes sense
- Try a login through Jetpack 
- Check that the source is changed to `jetpack`
- Try a login through other sources (share/deeplink)
- Check that the source is correct

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
@@ -52,6 +52,7 @@ import org.wordpress.android.ui.LocaleAwareActivity;
 import org.wordpress.android.ui.RequestCodes;
 import org.wordpress.android.ui.accounts.HelpActivity.Origin;
 import org.wordpress.android.ui.accounts.SmartLockHelper.Callback;
+import org.wordpress.android.ui.accounts.UnifiedLoginTracker.Flow;
 import org.wordpress.android.ui.accounts.UnifiedLoginTracker.Source;
 import org.wordpress.android.ui.accounts.login.LoginPrologueFragment;
 import org.wordpress.android.ui.accounts.login.LoginPrologueListener;
@@ -95,6 +96,8 @@ public class LoginActivity extends LocaleAwareActivity implements ConnectionCall
     private static final String KEY_SMARTLOCK_HELPER_STATE = "KEY_SMARTLOCK_HELPER_STATE";
     private static final String KEY_SIGNUP_FROM_LOGIN_ENABLED = "KEY_SIGNUP_FROM_LOGIN_ENABLED";
     private static final String KEY_SITE_LOGIN_AVAILABLE_FROM_PROLOGUE = "KEY_SITE_LOGIN_AVAILABLE_FROM_PROLOGUE";
+    private static final String KEY_UNIFIED_TRACKER_SOURCE = "KEY_UNIFIED_TRACKER_SOURCE";
+    private static final String KEY_UNIFIED_TRACKER_FLOW = "KEY_UNIFIED_TRACKER_FLOW";
 
     private static final String FORGOT_PASSWORD_URL_SUFFIX = "wp-login.php?action=lostpassword";
 
@@ -184,6 +187,11 @@ public class LoginActivity extends LocaleAwareActivity implements ConnectionCall
 
             mIsSignupFromLoginEnabled = savedInstanceState.getBoolean(KEY_SIGNUP_FROM_LOGIN_ENABLED);
             mIsSiteLoginAvailableFromPrologue = savedInstanceState.getBoolean(KEY_SITE_LOGIN_AVAILABLE_FROM_PROLOGUE);
+            String source = savedInstanceState.getString(KEY_UNIFIED_TRACKER_SOURCE);
+            if (source != null) {
+                mUnifiedLoginTracker.setSource(source);
+            }
+            mUnifiedLoginTracker.setFlow(savedInstanceState.getString(KEY_UNIFIED_TRACKER_FLOW));
         }
     }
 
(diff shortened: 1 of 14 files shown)


WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
```diff
@@ -149,15 +152,24 @@ protected void onCreate(Bundle savedInstanceState) {
                     }
                     break;
                 case SELFHOSTED_ONLY:
+                    mUnifiedLoginTracker.setSource(Source.SELF_HOSTED);
```
renanferrari, line 155: I know we're reusing the names that are already being used by the application here, but I just wanted to point out that `SELF_HOSTED` is not an accurate name for this source, since it's possible to add non self-hosted sites from it as well. 

I've used "Add Site" before to describe this source, but I don't think it's a good name as well. Besides that, an argument could be made that changing this would cause even more confusion, so I'll leave it up to you to decide what's the better alternative here.

WordPress/src/main/java/org/wordpress/android/ui/accounts/UnifiedLoginTracker.kt
```diff
@@ -0,0 +1,64 @@
...
+import javax.inject.Singleton
+
+@Singleton
+class UnifiedLoginTracker
+@Inject constructor(private val analyticsTracker: AnalyticsTrackerWrapper) {
+    private var currentSource: Source = DEFAULT
+    private var currentFlow: Flow? = null
+    fun track(step: Step) {
+        track(step = step)
+    }
```
renanferrari, line 17: This puts the app in a infinite loop when called, eventually crashing it. Maybe we should just use the other `track` function and invert the parameters order?

WordPress/src/main/java/org/wordpress/android/ui/accounts/UnifiedLoginTracker.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    fun track(flow: Flow? = null, step: Step) {
+        currentFlow = flow
+        if (BuildConfig.UNIFIED_LOGIN_AVAILABLE) {
+            currentFlow?.let {
```
renanferrari, line 22: What do you think of logging an error here when `currentFlow` is null so we are able to know we're missing some flows? Or maybe `currentFlow` could have a default value, like "unknown"? I'll leave this up to you.

libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginEmailFragment.java
```diff
@@ -369,6 +369,7 @@ protected void next(String email) {
             startProgress();
             mRequestedEmail = email;
             mDispatcher.dispatch(AccountActionBuilder.newIsAvailableEmailAction(email));
+            mAnalyticsListener.trackEmailFilled();
```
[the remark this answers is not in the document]
planarvoid: You're right, this event feels out of place. I've only added it because it was in the specifications but it doesn't really make sense. I'll remove it and replace it with a generic interaction/click event in another PR.

WordPress/src/main/java/org/wordpress/android/ui/accounts/UnifiedLoginTracker.kt
planarvoid: ah damn, I've added the named nullable parameters as one of the steps to clean it up but I haven't noticed this obvious non-sense. Thanks for pointing it out! I'll just pass the `currentFlow` as a parameter instead.

WordPress/src/main/java/org/wordpress/android/ui/accounts/UnifiedLoginTracker.kt
planarvoid: I've originally had a crash here to find out when it's not working. I can still add a crash in debug and log an error. I'm afraid just logging an error will not give us much information. What do you think?

WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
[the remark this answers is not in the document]
planarvoid: What do you think about renaming it to `LOGIN_WITH_SITE`? It's very explicit but I think that's a good thing in this case.

WordPress/src/main/java/org/wordpress/android/ui/accounts/UnifiedLoginTracker.kt
renanferrari: I like this approach, thank you!

renanferrari: I really like this approach, thank you!

WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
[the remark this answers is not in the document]
renanferrari: So, this source is mainly used when the user is already logged in and taps the "+" button on the Site List screen. I'm afraid that if we call it `LOGIN_WITH_SITE` it could be confused with when the user taps the "Enter your site address" button in the Prologue screen.

I honestly feel like I'm being too picky with this, so feel free to leave it as `SELF_HOSTED` like it was before. Besides, I just realized that the button to add a site in the app is actually called "Add a self-hosted site", so it might make sense to leave it that way, even if it's not entirely accurate.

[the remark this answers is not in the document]
planarvoid: I've renamed it back to `SELF_HOSTED`

PR #12090, opened 2020-06-03
Missing indicators in pages list

planarvoid:
This PR fixes a missing functionality where there were not indicators for `Homepage` and `Posts Page` on the pages list. I'm adding these indicators into the field that previously only contained the date the page was created.

To test:
- Go to the pages list on a WP.com or Jetpack site
- Notice that there are Homepage/Posts Page indicators visible
- Try to set Homepage/Posts Page
- The indicators are updated

To test 2:
- The indicators are not visible on a self-hosted site without a Jetpack

homepage_in_page_list

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageListViewModel.kt
```diff
@@ -401,14 +401,20 @@ class PageListViewModel @Inject constructor(
...
     }
 
     private data class ItemUiStateData(
         val labels: List<UiString>,
         @ColorRes val labelsColor: Int?,
         val progressBarUiState: ProgressBarUiState,
         val showOverlay: Boolean,
-        val actions: Set<Action>
+        val actions: Set<Action>,
+            val subtitle: Int? = null
```
mzorz: nit: extra spaces here

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageListViewModel.kt
```diff
@@ -292,14 +292,14 @@ class PageListViewModel @Inject constructor(
                                 ScheduledPage(
```
mzorz, line 292: wdyt about adding named parameters here for the first 3 params? I know it's not necessary but given the subtitle optional param has been placed right below title, it now looks weird to have most of the parameters in named fashion with only the first few with a different notation. Also, comparing these two it looks a lot clearer to have names for everything given it's quite a list of parameters to look into :)

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageListViewModel.kt
```diff
@@ -316,14 +316,14 @@ class PageListViewModel @Inject constructor(
                     it.remoteId,
```
mzorz, line 316: same here if it makes sense to you

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageListViewModel.kt
```diff
@@ -338,14 +338,14 @@ class PageListViewModel @Inject constructor(
...
```
mzorz, line 338: and here

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageListViewModel.kt
planarvoid: good idea 👍 , I'll change it in all the places

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageListViewModel.kt
planarvoid: thanks, I've fixed that and one other lint error

WordPress/src/main/java/org/wordpress/android/ui/pages/PageItemViewHolder.kt
```diff
@@ -80,8 +81,19 @@ sealed class PageItemViewHolder(internal val parent: ViewGroup, @LayoutRes layou
...
                 val date = if (page.date == Date(0)) Date() else page.date
-                time.text = DateTimeUtils.javaDateToTimeSpan(date, parent.context)
+                val stringDate = DateTimeUtils.javaDateToTimeSpan(date, parent.context)
                         .capitalizeWithLocaleWithoutLint(parent.context.currentLocale)
+                val subtitle = page.subtitle
+                pageSubtitle.text = if (subtitle == null) {
+                    stringDate
+                } else {
+                    String.format(
+                            Locale.getDefault(),
```
mzorz, line 91: I've seen we have a very inconsistent use of `getLocale()` throughout the app.
Sometimes we use `Locale.US`, sometimes we sue `String.format()` without indicating the Locale, sometimes `Locale.getDefault()` as you use here, and sometimes we even have our own `LanguageUtils` function that does exactly that, called `getCurrentDeviceLanguage`.

This is just a comment as I was curious to see if there was a particular "way we do things"  in the codebase but it seems to be all evenly split. No action need be taken but thought I'd still make the observation here, in case this triggers any thoughts.

planarvoid: I'm honestly not sure why this method needs a Locale, the documentation says it's for localization. I think we should always use `getDefault` for user facing texts. We should use Locale.US if we want a consistent result (that's not translated) - this is for example for tracking or for API calls

PR #11457, opened 2020-03-18
Check user rights to create content to give access to main bottom sheet

develric:
Fixes #11388

When a user is an Author/Contributor of a site we already remove the access to pages in My Site screen (both Quick Action and List row item):

<p align="center">
  
</p>

This PR extends that check also to the FAB opening directly the Editor for post creation when a user is a User/Contributor (or not a Self-Hosted site Admin).

cc FYI @rachelmcr and  @designsimply.

## To test:
In the following the test must be repeated both with a WP.com and a Self-Hosted site.

- Add a user with Author/Contributor role to a site (you can do it for example from the web)
- Login in the app with that user and switch to the site
- Note that pages access are removed like the image above
- Tap on the Fab and check you get the editor opened for post editing
- Create the post and publish it; note you get it published in the posts list
- Change the User role to be an Editor/Administrator (you can do it for example from the web)
- Restart the app and check the desired site is selected
- Note you now have access to pages content
- Tap on the Fab and check you get the bottom sheet for post/page creation
- Test you can create both post and page and get it uploaded and reported in the post/page lists
- Clean the app cache and login again to show the FAB tooltip. Check that on a site with Editor/Administrator rights you get `Create a post or page` message where on a site with Author/Contributor rights you get `Create a post` message.
- Long press on the FAB to get a toast. Check that on a site with Editor/Administrator rights you get `Create a post or page` message where on a site with Author/Contributor rights you get `Create a post` message.
- Use Talkback to get the description of the FAB. Check that on a site with Editor/Administrator rights you get `Create a post or page` message where on a site with Author/Contributor rights you get `Create a post` message.

Hi @osullivanchris 👋, if you want/can have a look into it I think you can use the apk generated in this PR for checking.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #10720, opened 2019-10-31
IA Project: create post/page from My Site

develric:
Fixes  #10618 and supersedes PR #10617

As part of the Information Architecture project this PR adds the possibility to add a new post or page from the My Site screen.

This modifications are currently behind a feature flag together with the modifications for the Me Move. The common feature flag was renamed accordingly from ME_ACTIVITY_AVAILABLE to INFORMATION_ARCHITECTURE_AVAILABLE

We introduced visually two modifications highlighted below:

| 3 items bottom menus + fab | bottom sheet on fab tap |
|---|---|
| image | image|

## NOTES
1. The feature flag is active only in the zalpha flavor that reaches internal testers for now if we merge in develop (cc @designsimply )

## To test

1. Compile and run the zalpha flavor
2. Check that the Me and the new post menu items are not in the bottom nav bar
<p align="center">
  
</p>

3. Switch between My site/Reader/Notifications pages and check that the FAB is only available in the My site screen
4. Tap on each menu item in the bottom navbar to check that the correct screen is visualized
5. Select the FAB and check that a bottom sheet is opened. The Editor is opened when selecting the Create new Page/Post from the bottom sheet
6. Create a post/page and tap on the back arrow. You should be back on the main screen and notified by a snackbar that the post was saved on line.
7. Access to the Pages/Posts list and check the new created Page/Post was created and available in the correct list (Published/Draft).
8. Compile and run vanilla flavor
9. Check that you have all the 5 menu item in the Bottom nav bar and the FAB is not present
10. Smoke test to check that the menus activate the expected screens and the app works as usual. 

PS: i think we do not need to include notes in the RELEASE NOTES for the above since it is still behind a feature flag, but not fully sure about it. Let me know wdyt. 

PR submission checklist:

- [x] I have considered adding unit tests where possible.

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #12094, opened 2020-06-03
WPMainActivity: fix selected site being null when the ViewModel is started

mzorz:
This fixes a **potential** bug, that is not apparent in the app (it's not really causing a problem right now but is latent there).  Found this out while making changes to introduce WP Stories in https://github.com/wordpress-mobile/WordPress-Android/pull/11930.

In `develop`, `initViewModel()` is getting called _before_ `initSelectedSite()`, we should only call `initViewModel` _after_ initializing the selected site, given the VM initialization depends on the data brought by the selected site.

Looking a bit `initViewModel()` has been introduced in this commit
https://github.com/wordpress-mobile/WordPress-Android/commit/c2591b47163be37aac6e4a0224b9a78c9ac555c0
in this PR https://github.com/wordpress-mobile/WordPress-Android/pull/10720/
In that PR it was all good, as initViewModel wasn't depending on the selected site :thumbsup:

But then in this other commit https://github.com/wordpress-mobile/WordPress-Android/commit/5b3d219e1f7b81bfee0fb16906b2c8ca936c362c
belonging to this PR https://github.com/wordpress-mobile/WordPress-Android/pull/11457  a dependency on `hasFullAccessToContent()` is introduced, which in turn calls `getSelectedSite()`. Being that `mSelectedSite` is still `null` in `onCreate()` (where the mViewModel.start() call is made) then the `hasFullAccessToContent()` method returns false, even when the user is an admin for the site. (this can be tried out with the debugger).

The checks for user rights are made _after_  the view model is initialized.

Look [here](https://github.com/wordpress-mobile/WordPress-Android/commit/5b3d219e1f7b81bfee0fb16906b2c8ca936c362c#diff-1cf6428dcfe0d429dfdcf7398ef4c056R480-R483
):

```
     mViewModel.start(
                mSiteStore.hasSite() && mBottomNav.getCurrentSelectedPage() == PageType.MY_SITE,
                hasFullAccessToContent()
        );
```

Given the initializer is called in `onResume()`, moving `mViewModel.start()` right after that call in `onResume` seems to be a solution we could use (this is what this PR does).
Also there shouldn't be a risk of running this several times given the `isStarted` flag in `start()`will be `true` but, let me know.

Pinging the writer / reviewer of such PR given they'll be more familiar with that @planarvoid @develric . Let me know your thoughts!

To test:
- follow the same instructions as described in https://github.com/wordpress-mobile/WordPress-Android/pull/11457 and verify this doesn't break the functionality introduced there.

1. create a new user with a new blog (or if you have one, you can use it)
2. add it to be a contributor on another blog of yours
3. the user should have now at least 2 blogs, one of which they're an admin of, and the other one where they're just contributor
4. tapping on the `switch site` button should show these 2 sites
5. switching between them should hide / show the Pages option as desired.
6. Verify the Page option is shown / hidden by re-starting the app with the selected site being the one your user is an admin of (Page option should show)
7. then switch the site to the site where your user is a contributor only, restart the app and verify the app doesn't show the Page option after restarting the app.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
@@ -483,11 +483,6 @@ public class WPMainActivity extends LocaleAwareActivity implements
                 return null;
             });
         });
-
-        mViewModel.start(
-                mSiteStore.hasSite() && mBottomNav.getCurrentSelectedPage() == PageType.MY_SITE,
-                hasFullAccessToContent()
-        );
     }
 
     private @Nullable String getAuthToken() {
@@ -702,6 +697,11 @@ public class WPMainActivity extends LocaleAwareActivity implements
 
         // Load selected site
         initSelectedSite();
+
+        mViewModel.start(
+                mSiteStore.hasSite() && mBottomNav.getCurrentSelectedPage() == PageType.MY_SITE,
+                hasFullAccessToContent()
+        );
 
         // ensure the deep linking activity is enabled. We might be returning from the external-browser
         // viewing of a post


develric:
Hi there 🙂 ! I agree with @planarvoid that the FAB rights are better evaluated in the onFabClicked, we need anyway also to manage the state of the tooltip before a user interaction happens, showing a message related only to posts (and now stories! really excited to see this comes true @mzorz 😉) and not pages in cases when the user has no rights for pages.

IIRC I think I was passing the `hasFullAccessToContent` to the `mViewModel.start` basically to set a default status for the `setMainFabUiState(isFabVisible, hasFullAccessToContent)` (in fact the `hasFullAccessToContent(@Nullable SiteModel site)` function, as Mario already spotted, passing a null will default to the more restrictive rights case)...we then have already in the onResume the `mViewModel.onResume(hasFullAccessToContent());` to set the fabUIState accordingly. I agree though that the fact we pass a value to the start seems not a default but a current/correct state so maybe this can create confusion/doubts if not issues.

Didn't try in code, but wondering if we could do something like the following:
- keep the mViewModel.start in initViewModel but do not pass parameters so delay the fabUiState initialization
- in onResume do something like: 
```
mViewModel.onResume(mSiteStore.hasSite() && mBottomNav.getCurrentSelectedPage() == PageType.MY_SITE,
                hasFullAccessToContent());
```

- and in mViewModel.onResume use the setMainFabUiState so that we use a valid state and also make use of the `appPrefsWrapper.isMainFabTooltipDisabled()` shared prefs state.

What do you both think?

mzorz:
Thanks for your review and comments @develric @planarvoid ! 🙇 

>  (and now stories! really excited to see this comes true @mzorz 😉) 

Thank you! :D this PR was just an excuse so I could link it and make you look into the other PR lol (jkjk)

> I think we need to keep the call in the onFabClicked method because the site can change at any time (at least looking at the code it's changing constantly). What can happen is that you can get the rights to access the Pages in a OnSiteChanged update. I'm wondering if we can maybe remove the hasFullAccessToContent from the start method? Does it make any sense @develric @mzorz ?

and

>  I agree with @planarvoid that the FAB rights are better evaluated in the onFabClicked,


Exactly :) this is why I added that there in my other PR (for my own purposes) [here](https://github.com/wordpress-mobile/WordPress-Android/pull/11930/files#diff-49a69b655427953631c4a08a261abd22R98-R101)  :)

> Your change doesn't really change this behaviour. We would still call the setMainFabUiState twice but this time it will have the hasFullAccessToContent==true in both cases.

That's right - but it'll make things consistent (avoid evaluating a null case that shouldn't exist). All this change does is make sure to only make the call _after_  the site has been initialized. Also thinking of that I think I could have used `onStart()` instead of `onResume`, but still doesn't really matter much.

Re: this:
> I'm wondering if we can maybe remove the hasFullAccessToContent from the start method? Does it make any sense @develric @mzorz ?

This is related to this @develric says:

> we need anyway also to manage the state of the tooltip before a user interaction happens, showing a message related only to posts

so basically the tooltip depends on it, and the start method calls that so the tooltip is correctly set.

Now onto this!
> IIRC I think I was passing the hasFullAccessToContent to the mViewModel.start basically to set a default status for the setMainFabUiState(isFabVisible, hasFullAccessToContent) (in fact the hasFullAccessToContent(@Nullable SiteModel site) function, as Mario already spotted, passing a null will default to the more restrictive rights case)...we then have already in the onResume the mViewModel.onResume(hasFullAccessToContent()); to set the fabUIState accordingly. I agree though that the fact we pass a value to the start seems not a default but a current/correct state so maybe this can create confusion/doubts if not issues.

AH this clarifies things for me 😌 . Also, I completely missed the `onResume()` call to the ViewModel which already does everything this PR was trying to do anyway. That's nice, thanks for pointing that out. You know what? I think if this is intended behavior (by design) as you explained we could just as well add a comment in code explaining this? and that's it probably.

Let me know what you think about adding this (and removing the change in this PR):
```
// The first time this is called in onCreate we still haven't initialized mSelectedSite, which hasFullAccessToContent depends on, and as such the state will be initialized with the most restrictive rights case (that is, will assume hasFullAccessToContent is false). This is OK and will be frequently checked to normalize the UI state whenever mSelectedSite changes.
```

Also for a further iteration, we may as well make `mSelectedSite` a ViewModel attribute and have a LiveData around it, and cascade the changes as per usual in MVVM? Given this is all we want (there are a lot of things that change in the UI for this so, looks like `mSelectedSite` is a state variable that should belong to the ViewModel). Also, not intending to introduce such a change right now, but putting my thoughts out there just for discussing @planarvoid @develric.

Let me know your thoughts and I'll proceed to add the comment (and remove this change) in code if you agree for now 👍

planarvoid:
thanks for the detailed reply @mzorz, it makes sense 👍  

> Let me know what you think about adding this (and removing the change in this PR):

sure 👍 

>Also for a further iteration, we may as well make mSelectedSite a ViewModel attribute and have a LiveData around it, and cascade the changes as per usual in MVVM? Given this is all we want (there are a lot of things that change in the UI for this so, looks like mSelectedSite is a state variable that should belong to the ViewModel).

I think we should have something like a `SelectedSiteRepository` which would keep the currently selected site and expose a `LiveData` endpoint (or Flow or anything reactive). This could be reused all over the place and it would be updated on the eventbus event. This would enable us to stop passing around the SiteModel. This would be true in most cases, in some cases (like in stats) we sometimes want to show the data for other than selected site as well. For example when you open the screen from a widget for site A and you have selected site B - you want to keep the selection but show data for another site.

PR #12099, opened 2020-06-04
Update README.md

ajithrn:
added notes about avoiding update plugins while building app on Android Studio. Which might cause interrupting the build process due to plugin conflict. 

Fixes #
```
FAILURE: Build failed with an exception.

* Where:
Build file '[...]/WordPress-Android/build.gradle' line: 21

* What went wrong:
An exception occurred applying plugin request [id: 'com.gradle.build-scan', version: '2.0.2']
> Failed to apply plugin [id 'com.gradle.build-scan']
   > The build scan plugin is not compatible with this version of Gradle.
     Please see https://gradle.com/help/gradle-6-build-scan-plugin for more information.

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.
```

build-scan plugin is not compatible with Gradle > 6.0.1. This project require 5.4.1 so avoid update Gradle since it might cause plugin conflicts and interrupt build process in android studio

To test:
N/A

PR submission checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ it on [Google Play](https://play.google.com/store/apps/details?id=org.wordpress.
 Notes:
 
 * To use WordPress.com features (login to WordPress.com, access Reader and Stats, etc) you need a WordPress.com OAuth2 ID and secret. Please read the [OAuth2 Authentication](#oauth2-authentication) section.
+* While building the app in Android Studio ignore all the update plugin prompts (updating plugins may resulted in plugin conflicts and interrupt the build process).
 
 
 ## OAuth2 Authentication ##
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ it on [Google Play](https://play.google.com/store/apps/details?id=org.wordpress.
 Notes:
 
 * To use WordPress.com features (login to WordPress.com, access Reader and Stats, etc) you need a WordPress.com OAuth2 ID and secret. Please read the [OAuth2 Authentication](#oauth2-authentication) section.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,6 +20,7 @@ it on [Google Play](https://play.google.com/store/apps/details?id=org.wordpress.
 Notes:
 
 * To use WordPress.com features (login to WordPress.com, access Reader and Stats, etc) you need a WordPress.com OAuth2 ID and secret. Please read the [OAuth2 Authentication](#oauth2-authentication) section.
+* While building the app in Android Studio ignore all the update plugin prompts (updating plugins may resulted in plugin conflicts and interrupt the build process).
```
hypest, line 23: Thanks for the PR @ajithrn !

What do you think about narrowing the scope of this hint? Do you perhaps mean some particular plugin to avoid updating? Currently the hint is a bit too general.

hypest:
> Fixes #
> Avoid build errors due to plugin conflict in android studio

Do you perhaps recall a particular error you encountered? It would be useful to report in case others encounter the same issue. Thanks!

README.md
ajithrn: @hypest avoid updating Gradle since this project depends on 5.4.1 and all the version above 6.0 wont support gradle.build-scan hence the app wont compile.

hypest: Aha, I see. Yes, we try to avoid that update if the effort to make it work is not planned. Good find!

How about we put something like this to the readme then:
```
* While loading/building the app in Android Studio ignore the prompt to update the gradle plugin version as that will probably introduce build errors. On the other hand, feel free to update if you are planning to work on ensuring the compatibility of the newer version.
```

ajithrn: sounds great.

PR #12111, opened 2020-06-05
Re-enable editor UI tests

### LINKED ISSUE #11897: Testing: Fix and re-enable editor UI tests
The [Editor UI tests](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/androidTest/java/org/wordpress/android/e2e/EditorTests.java) were disabled in https://github.com/wordpress-mobile/WordPress-Android/pull/11530 due to failures discussed [here](https://github.com/wordpress-mobile/WordPress-Android/pull/11506#issuecomment-603829709):

> the problem seems to be that the tests expect Aztec (not Gutenberg) but for some reason sometimes Gutenberg is getting enabled.

We should fix those tests and enable both those Aztec editor tests and the [Block editor tests](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/androidTest/java/org/wordpress/android/e2e/BlockEditorTests.java).

mkevins:
Fixes #11897

### Description

This PR adds a `ViewAction` utility method to set explicitly the editor preference, and sets the specific editor for the Classic and Gutenberg tests.

Note: this PR does not yet address the startup time issue with the Gutenberg editor, <s>so I'm leaving this as draft for now. If a solution to that is not soon available, these changes should allow disabling of only Gutenberg tests, so the Aztec tests can still run.</s> It seems these issues may be resolved? I'll mark this for review, and revisit if there are still problems with the tests timing out.

**Update:**

I encountered some timeout issues with the Gutenberg editor test enabled. Additionally, there were problems with the `signUpWithMagicLink` test (mismatches on API endpoints reported from WireMock). I'm not sure why, but this goes away after disabling the Gutenberg editor test, so I suspect it is related to the editor flag. I've disabled the Gutenberg editor test and everything passes now. It may be worth fixing the Gutenberg editor test in a separate PR so that we don't block re-enabling the Aztec editor tests (since it seems that requires a bit more investigation).

To test: 

Editor UI tests should pass locally and on CI.

PR submission checklist:


mkevins:
Thanks for reviewing and testing Oguz!

PR #12113, opened 2020-06-05
RI:  Choose Interests - Introduce loading state

ashiagr:
Master Issue: #12027

This PR introduces loading state into the Choose Interests screen.

### To test

- Launch app and go to Reader tab
- Notice that progress bar is shown on a blank screen
- Wait for a while
- Notice that 
    - Progress bar is hidden
    - Done button with text "Select a few to continue" is shown in disabled state
    - Interest chips are displayed along with title, subtitle 

#### Note

This PR includes UI changes from PR: Choose Interests UI #12110. These are still being finalised (like whether Done button should be scrollable in the landscape mode), so please ignore them in this PR review.

### Reviewing

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
```diff
@@ -90,11 +88,34 @@ class ReaderInterestsViewModel @Inject constructor(
         _uiState.value = uiState
     }
 
-    data class UiState(
+    sealed class UiState(
         val interestsUiState: List<InterestUiState>,
         val interests: ReaderTagList,
```
ashiagr, line 93: `InterestsUiState`, `Interests` are only needed when data is loaded. Will refactor to not include them for other states in the next PR that handles failed state.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
```diff
@@ -90,11 +88,34 @@ class ReaderInterestsViewModel @Inject constructor(
...
     ) {
+        object InitialUiState : UiState(
+            interestsUiState = emptyList(),
+            interests = ReaderTagList(),
+            doneButtonUiState = DoneButtonHiddenUiState,
+            progressBarVisible = true
+        )
+
+        data class ContentLoadSuccessUiState(
+            val interestTagsUiState: List<InterestUiState>,
```
malinajirka, line 107: I think we can use the `override` keyword so we don't need to pass the values into the constructor. The current implementation creates two separate fields instead of using the field defined in the parent. Wdyt?

ashiagr: Using @override is a good idea, thanks!

Also I was thinking to remove interests related fields from the parent and include only in the `ContentLoadSuccessUIState` https://github.com/wordpress-mobile/WordPress-Android/pull/12113#discussion_r440124637

ashiagr:

> I'm not sure if you plan to update the UI later if so, please ignore this comment. I believe the loading state should look like this

I'll update it later as part of https://github.com/wordpress-mobile/WordPress-Android/pull/12110

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
malinajirka: 👍 sounds good, thanks!

PR #12114, opened 2020-06-05
Guard for nulls in processors and abort if necessary

mkevins:
Fixes: #12104

This PR guards for nulls in the processor Json handling. While it isn't clear why ids would not be present in parsed json from the block header, returning false will leave the content unchanged. Since no changes can be made to the parsed structure without matching the id, aborting should be the safest route.

I believe the reason why the id is missing is that it is not present in the templates, though I have not yet confirmed this.

### Steps to test

* Install and launch the application
* Tap on "Log In"
* Log in to the app using valid credentials
* Go to My site and tap on "Pages" icon
* Go to Drafts and tap on "CREATE A PAGE" button
* Select the Service template (observe template preview)
* Tap "Apply" option
* Insert Image block
* Add an image with "Choose from device"
* Leave the editor (tap back button) before the upload completes

**Expected:**

The app should not crash, and upload should complete successfully

* When upload has finished (observed in the notifications) reopen the page
* Switch to HTML view

**Expected:**

The url for the added image should be a remote URL (i.e. `https://` not `file://`)

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/mediauploadcompletionprocessors/CoverBlockProcessor.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/mediauploadcompletionprocessors/CoverBlockProcessor.java
@@ -1,5 +1,6 @@
 package org.wordpress.android.ui.posts.mediauploadcompletionprocessors;
 
+import com.google.gson.JsonElement;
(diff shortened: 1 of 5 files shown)


mkevins:
Thanks Chip for reviewing and testing so quickly!

PR #12116, opened 2020-06-05
Unified login failure tracking

### LINKED ISSUE #11797: Unified Login & Sign-Up: Update tracking events

planarvoid:
Fixes #11797

I was thinking about the failure tracking and in the end I ended up with one universal event that contains all the source/flow/step of the last emitted Step. It also has a field for a general error message. This way we don't have to create an enum with all the error types but we can still analyze the failures by the message if we want to. It is also very flexible (because I can reuse the existing error messages). I'm tracking the error every time it's shown to the user, not when a failure happens (because some are expected). One downside I can see is that the message will be localized but I'm wondering if that matters. What do you think?

I'll add unit tests once we use a feature flag (after we merge develop)

To test:
- Test various flows that fail (wrong password input, wrong email...)
- Notice that there is a new event `unified_login_failure` tracked
  - It has source, flow and step the same as the previous event
  - It has a failure message as an additional property

PR submission checklist:


libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/GoogleFragment.java
```diff
@@ -225,6 +225,7 @@ protected void finishFlow() {
     }
 
     protected void showError(String message) {
+        mAnalyticsListener.trackFailure(message);
         finishFlow();
         mGoogleListener.onGoogleSignupError(message);
     }
```
renanferrari, line 231: I was initially confused on why this wasn't being handled inside the `onGoogleSignupError` callback but after looking at `UnifiedLoginTracker.trackFailure()` I'm assuming it's to preserve the current flow and step. If that's the case, I think we need a way to communicate this behavior more clearly. `UnifiedLoginTracker` is currently not a part of the login library, so other library clients will only be able to look at `LoginAnalyticsListener.trackFailure()` to get more context on how this is supposed to work, which doesn't make it obvious. Let me know what you think.

renanferrari: Another thought I just had about this: do we really need to track the current flow and step along with failures and clicks? I'm not very familiar with Tracks, but wouldn't we be able to use it to check what was the last tracked flow/step before a certain failure event was tracked?

planarvoid: > I was initially confused on why this wasn't being handled inside the onGoogleSignupError callback but after looking at UnifiedLoginTracker.trackFailure() I'm assuming it's to preserve the current flow and step. If that's the case, I think we need a way to communicate this behavior more clearly. UnifiedLoginTracker is currently not a part of the login library, so other library clients will only be able to look at LoginAnalyticsListener.trackFailure() to get more context on how this is supposed to work, which doesn't make it obvious. Let me know what you think.
Hm, this is an interesting catch. Not having the `trackFailure` in the `onGoogleSignupError` is better and it should be there. However you're right that we might lose the flow/step by this change. I'll look into it. 
> Another thought I just had about this: do we really need to track the current flow and step along with failures and clicks? I'm not very familiar with Tracks, but wouldn't we be able to use it to check what was the last tracked flow/step before a certain failure event was tracked?
My thinking behind this is that it will be much easier to analyze if a certain step is failing if we have the data also in the failure event. If you don't, you have to use funnels to get that information. I've asked the data team about this and they agree that it is useful to keep the flow and the step on the failure events. We should just make sure we're tracking the correct information :-)

planarvoid:
Thanks for the review @renanferrari 

> I would suggest you to check with our data team before moving forward.

I've checked with them and they agree that this is probably the best approach at the moment (finding the right granularity is hard).

libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/GoogleFragment.java
planarvoid: I've moved the failure tracking to the `onGoogleSignupError`. I've tested it (slow data -> try google login -> turn on airplane mode -> failure + dialog) and it works correctly and keeps the last flow & the step. I think this approach is better 👍

PR #12117, opened 2020-06-05
Add click tracking to unified login and signup flow

### LINKED ISSUE #11797: Unified Login & Sign-Up: Update tracking events

planarvoid:
Fixes #11797

This PR adds tracking to all the clicks in the login flow. The events tracked are `unified_login_interaction` (do you think it would be better to use "click" instead?). The event contains the previous source/flow/step to know for sure what happened before in the flow. I tried to cover all the clicks by looking for all the onClickListeners but it's possible I missed some of the cases.

To test:
- Filter `unified_login_interaction` in Logcat
- Try various login flows
- Check that all the clicks are tracked and that the value tracked makes sense

PR submission checklist:


planarvoid:
thanks for the review @renanferrari 

> I did found a bug though: when we click the Continue with Google button and the Google dialog appears, the flow changes to google_login, but if we dismiss the dialog, the flow stays as google_login. This causes the click to the Continue or help button to be logged as in the google_login flow.

This is a great catch! I've implemented a solution that sets the flow & step in `onResume` in the Email and Prologue fragments. It doesn't seem to be issue anywhere else (as far as I can see). I think it's because the other fragments are not kept when you jump around). It seems to work fine now. Let me know what you think!

PR #12122, opened 2020-06-08
Prepublishing Nudges : Submit for review

### LINKED ISSUE #12042: Prepublishing Nudges : Add Submit For Review / PENDING_REVIEW to bottom sheet
### Current behavior
The app crashes when the app goes into the state of `Pending Review` or `Save`
### Expected behavior
The user should be able to select `Submit For Review` within the bottom sheet without issue. Currently, this causes an `Exception` because we didn't anticipate the behavior. 

The user should also be able to trigger the bottom sheet when they change the `PostStatus` to `PENDING_REVIEW` in the `Post Settings`

Tested on Google Pixel 4 XL, Android 10, WPAndroid https://github.com/wordpress-mobile/WordPress-Android/pull/12008

jd-alexander:
Fixes #12042 

## Solution
Ensure that `Submit For Review` & `Save` actions don't crash the Editor when the `PostStatus` reflects them. 

#### User can publish - Pending Review. 

1. Create a new post. 
2. Click Publish. 
3. Change Visibility to `Pending Review` 
4. Observe that `Save` is shown once you navigate back to the home of the bottom sheet. 

#### User cant publish - Pending Review or Private. ( Note : I haven't verified this one on device. Is this available in self hosted sites because I was trying to do it there and it wouldn't work. I might need to create another account tomorrow so I can lessen the permissions on it.) 

1. Create a new post. 
2. Click Publish. 
3. Change Visibility to `Pending Review` or `Private`
4. Observe that `Submit` is shown once you navigate back to the home of the bottom sheet. 

## Reviewing 

## Submitter Checklist


WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -1347,10 +1347,10 @@ private void performPrimaryAction() {
             case PUBLISH_NOW:
             case UPDATE:
             case SCHEDULE:
+            case SUBMIT_FOR_REVIEW:
```
malinajirka, line 1350: I might have asked this already, but I forgot the explanation, sorry.

Is it okay that we track `EDITOR_POST_PUBLISH_TAPPED` for all the actions (not just "publish")?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeItemUiState.kt
```diff
@@ -16,28 +16,40 @@ sealed class PrepublishingHomeItemUiState {
     data class HeaderUiState(val siteName: UiStringText, val siteIconUrl: String) :
             PrepublishingHomeItemUiState()
 
-    sealed class SubmitButtonUiState(
+    sealed class ButtonUiState(
         val buttonText: UiStringRes,
         val publishPost: PublishPost,
         open val onButtonClicked: (PublishPost) -> Unit
```
malinajirka, line 22: I believe we can move this away from the constructor - it can be a regular field (in ButtonUiState). We won't need to pass it as an argument in all the children. The `override` keyword will make sure to write the value in to the correct field. Wdyt? (similar [example](https://github.com/wordpress-mobile/WordPress-Android/blob/9bdfb35a3cd085544a21b8f05d9f35ea59518861/WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterListItem.kt#L17))

jd-alexander: Nice. Thanks for this 😉

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
jd-alexander: This is a good observation. Thanks!

jd-alexander: f4a0fcd

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeItemUiState.kt
jd-alexander: cdcc032

jd-alexander:
Thanks for the review @malinajirka I appreciate it

> You don't need a self-hosted site to test this. Just invite a .com user to one of your sites and give them just "contributor" permissions. (https://wordpress.com/support/user-roles/#adding-users-to-your-site)

Nice! This worked well!

> It seems this issue is still reproducible:
> 
> Create a new post as contributor
> Enter title
> Click on "Submit"
> Bottom sheet is opened
> Click on Submit Now
> Notice the post gets uploaded as draft - not as pending review

Thanks for spotting this! My solution was to check if the user has publishing privileges and then based on that I set the post to either `DRAFT` or `PENDING`. dd12405

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
malinajirka: Follow up question :D - is it okay that we don't track events for the remaining buttons?

jd-alexander:
This can't be merged as yet because I am having some Gutenberg issues after doing the merge of `develop` into the feature branch. Working on getting the feature branch to green first.

malinajirka:
Thanks @jd-alexander, gj! 

I've encountered one more bug - probably related to the recent changes.
1. Switch to a site in contributor role
2. Click on the create new post button
3. Click on Submit
4. Notice the "Publish date" is empty.

jd-alexander:
Oh yes. Thanks for spotting that bug. I found it yesterday and made an issue for it 😎

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
jd-alexander: Thanks. They should be tracked as well. 🙏

PR #12123, opened 2020-06-08
Show post status on post list search results

malinajirka:
Adds post status to search results on post list - based on [this discussion](https://github.com/wordpress-mobile/WordPress-Android/pull/11291#issuecomment-589876505).

To test:
1. Open Post List
2. Set author filter to "Me"
3. Verify the rows contain just date
4. Set author filter to "Everyone"
5. Verify the rows contain date and author name
6. Click on the search icon and type something
7. Notice the search results contain date, name and post status

* Perform the above steps in both compact and full view

| Before | After
|------|-------|
| Screenshot_1591609598 | Screenshot_1591609363 |

PR submission checklist:

@SylvesterWilmott Could you please take a quick look at this change. Thanks!

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PostListItemViewHolder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PostListItemViewHolder.kt
@@ -20,7 +20,6 @@ import androidx.recyclerview.widget.RecyclerView
 import org.wordpress.android.R
 import org.wordpress.android.ui.utils.UiHelpers
 import org.wordpress.android.ui.utils.UiString
-import org.wordpress.android.util.LocaleManagerWrapper
 import org.wordpress.android.util.expandTouchTargetArea
 import org.wordpress.android.util.getDrawableFromAttribute
 import org.wordpress.android.util.image.ImageManager
@@ -40,8 +39,7 @@ sealed class PostListItemViewHolder(
     @LayoutRes layout: Int,
     parent: ViewGroup,
     private val imageManager: ImageManager,
-    private val uiHelpers: UiHelpers,
-    private val localeManagerWrapper: LocaleManagerWrapper
+    private val uiHelpers: UiHelpers
 ) : RecyclerView.ViewHolder(LayoutInflater.from(parent.context).inflate(layout, parent, false)) {
     private val featuredImageView: ImageView = itemView.findViewById(R.id.image_featured)
     private val titleTextView: WPTextView = itemView.findViewById(R.id.title)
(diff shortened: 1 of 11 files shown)


WordPress/src/main/java/org/wordpress/android/ui/posts/PostListItemViewHolder.kt
```diff
@@ -132,6 +137,13 @@ sealed class PostListItemViewHolder(
         }
     }
 
+    private fun updatePostInfoLabel(view: TextView, uiStrings: List<UiString>?) {
+        val concatenatedText = uiStrings?.joinToString(separator = "  ·  ") {
```
malinajirka, line 141: I decided to move the concatenation from the PostListItemUiStateHelper to the view holder. This isn't ideal, but since we don't have access to the context in the PostListItemUiStateHelper we can't transform post status UiStringRes to the actual string.

WordPress/src/test/java/org/wordpress/android/viewmodel/posts/PostListItemUiStateHelperTest.kt
```diff
@@ -897,7 +903,23 @@ class PostListItemUiStateHelperTest {
         )
 
         // Assert
-        assertThat((state.data.dateAndAuthor as UiStringText).text).isEqualTo(FORMATTER_DATE)
+        assertThat(state.data.postInfo!!.size).isEqualTo(1)
+        assertThat(state.data.postInfo!![0]).isEqualTo(UiStringText(FORMATTER_DATE))
+    }
+
+    @Test
+    fun `post status is dispalyed when isSearch == true`() {
```
oguzkocer, line 911:
```suggestion
    fun `post status is displayed when isSearch == true`() {
```

WordPress/src/main/java/org/wordpress/android/viewmodel/posts/PostListItemUiStateHelper.kt
```diff
@@ -153,19 +158,35 @@ class PostListItemUiStateHelper @Inject constructor(
...
+        if (isSearch) {
+            val postStatusText = when (postStatus) {
+                UNKNOWN -> R.string.unknown
+                PUBLISHED -> R.string.post_status_post_published
+                DRAFT -> R.string.post_status_draft
+                PRIVATE -> R.string.post_status_post_private
+                PENDING -> R.string.post_status_pending_review
+                TRASHED -> R.string.post_status_post_trashed
+                SCHEDULED -> R.string.post_status_post_scheduled
             }
```
oguzkocer, line 186: This doesn't necessarily need to happen in this PR, but it might be useful to create a utility function for this to make it reusable. I guess for now, this is just something to keep in mind.

malinajirka: I agree with both statements;). I'll leave it for now and we can move it when we find the need to re-use it.

SylvesterWilmott:
>I'll leave that to the "pro"

Thanks @oguzkocer haha

The design looks good but I agree here. We should capitalise those labels and the timestamp (month names).

malinajirka:
@SylvesterWilmott @oguzkocer I've removed the `toLowerCase` call. It's ready for another round. Thanks!

PR #12133, opened 2020-06-09
Use view.postDelayed instead of creating a handler

planarvoid:
Calling `postDelayed` on `new Handler()` could potentially lead to a leak of the main activity (since the `Runnable` references it). Using the handler of the view instead should tie the handler lifecycle to the view and be stopped when the view is destroyed.

To test:
- Install leak canary
- Turn off internet while using the app
- Stress test the connection bar (click & rotate & go to background)
- Check that there are no leaks 

PR submission checklist:


malinajirka:
Thanks @planarvoid! I've added logs to the app and it's still leaking the activity. I believe we need to manually invoke v.removeCallbacks() on the view. Or we could create a new Holder as a field and cancel it in onDestroy/onStop. Wdyt?

planarvoid:
thanks for the review @malinajirka . I've changed the implementation to use a hold a Handler as a field and use it to trigger the runnable.

malinajirka (approved):
LGTM, thanks!

PR #12135, opened 2020-06-09
Leak in animation

planarvoid:
The new posts bar animation could live longer than the fragment which could lead to leaks. This PR adds a `clearAnimation` method called in `onPause`. We don't need to finish the animation if the user doesn't see it.

To test:
- Install leak canary
- Go to reader post list
- Check that there are no leaks

PR submission checklist:


planarvoid:
thanks for the check @malinajirka, I think it's correct to stop the animation when the screen goes to the background (even if it's still visible) but I'm also fine with the other option so I moved it to `onStop`.

PR #12149, opened 2020-06-10
Prepublishing Nudges : Clear private post

### LINKED ISSUE #12148: Prepublishing Nudges : Clear future date when Private Post
## Expected behavior
The `Publish Date` should be cleared when the user switches the visibility to Private.

## Actual behavior
If the user switches to other `PostStatuses` eg. `Pending` and then switches to `Private` the `Publish Date` is not cleared. This is because the logic that clears the date is looking for the `Scheduled` `PostStatus` to clear but it should also check to see if the date is in the future. 

1. Create a new post. 
2. Click Publish. 
3. Select Publish Date. 
4. Select a future date. 
5. Select Pending Review and then select Private.
6. Go back and observe the future date is still present.  

Tested on Google Pixel 4 XL, Android 10, WPAndroid `prepublishing_nudges` ()

jd-alexander:
Fixes #12148

## Solution
If the user switches to other `PostStatuses` eg. `Pending` and then switches to `Private` the `Publish Date` is not cleared if it's in the future. This is because the previous logic that clears the date was looking for the `Scheduled` `PostStatus` to clear but now it just checks to see if the date is in the future. 

## Testing
1. Create a new post. 
2. Click Publish. 
3. Select Publish Date. 
4. Select a future date. 
5. Select Pending Review and then select Private.
6. Go back and observe the future date is cleared.  

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/visibility/usecases/UpdatePostStatusUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/prepublishing/visibility/usecases/UpdatePostStatusUseCase.kt
@@ -5,10 +5,14 @@ import org.wordpress.android.fluxc.model.PostModel
 import org.wordpress.android.fluxc.model.post.PostStatus
 import org.wordpress.android.ui.posts.EditPostRepository
 import org.wordpress.android.ui.posts.EditPostRepository.UpdatePostResult
+import org.wordpress.android.ui.posts.PostUtilsWrapper
 import org.wordpress.android.util.DateTimeUtilsWrapper
 import javax.inject.Inject
 
-class UpdatePostStatusUseCase @Inject constructor(private val dateTimeUtilsWrapper: DateTimeUtilsWrapper) {
+class UpdatePostStatusUseCase @Inject constructor(
+    private val dateTimeUtilsWrapper: DateTimeUtilsWrapper,
+    private val postUtilsWrapper: PostUtilsWrapper
+) {
     fun updatePostStatus(
         postStatus: PostStatus,
         editPostRepository: EditPostRepository,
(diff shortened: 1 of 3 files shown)


jd-alexander:
>SideNote: I'm getting a bit lost in all the changes. Could you please remind me why we want to "reset" the date just for "Private" status and not for the others?

Hi @malinajirka I am back to this comment. I reset it due to the discussion we had about `PRIVATE` posts not being able to be scheduled on none of the platform. I am tracking this convo to double check if any other resets need to be done. Thanks. 

I was just doing some testing and I realized that say the `PostStatus` is `PENDING` and you click `Password Protected` in the bottom sheet should the `PostStatus` remain? I am asking this because in the current version of the bottom sheet that's what happens and it seems to be the same experience in the Post Settings. The difference is that the user won't know what the associated `PostStatus` was unless they look at the button(s) or go to Post Settings and see it under `Status & Visibility`. Let me know if I need to provide screenshots/gifs to explain this further thanks!

malinajirka:
>I was just doing some testing and I realized that say the PostStatus is PENDING and you click Password Protected in the bottom sheet should the PostStatus remain? I am asking this because in the current version of the bottom sheet that's what happens and it seems to be the same experience in the Post Settings. The difference is that the user won't know what the associated PostStatus was unless they look at the button(s) or go to Post Settings and see it under Status & Visibility. Let me know if I need to provide screenshots/gifs to explain this further thanks!

I can see the issue. I'm not sure what to do with it though - tbh I thought the idea was that password protected posts are equal to "Published with password". If that's not the case, we are mixing two completely unrelated things within one menu. Either case, this needs to be clarified before we merge the changes into develop. Publishing something the user didn't want to publish would be a disaster.

PR #12151, opened 2020-06-10
Preserve site multi-selections in site picker on configuration change.

### LINKED ISSUE #8350: Site Picker multi-selection not preserved on configuration change
### Expected behavior
When you select multiple sites in Site Picker, the selection should be retained on configuration change.

### Actual behavior
Selection is not retained on configuration change.

1. Use account with multiple sites and first alphabetical site hidden.
2. Go to My Site tab.
3. Tap Switch Site button in top card view.
4. Notice active site is selected.
5. Long-press next site in list (below divider if present).
6. Notice no sites are selected.
7. Notice top toolbar title shows 0 selected.
8. Tap any site.
9. Notice top toolbar title shows 1 selected.
10. Rotate the device.
11. Notice selection mode is disengaged and selected sites are not retained.

##### Tested on [device], Android [version]
Nexus 5X emulator, api 28

hafizrahman:
Fixes #8350.

## To test:

0. Use an account with at least 3 WordPress.com sites, start in portrait mode
1. select Switch Site
2. Make sure at least one WordPress.com site is hidden. To hide a site, long press on any site on the list until the toolbar changes. Select a site, then select "Hide" at the top. The site should be hidden from the list.
3. Now the actual testing can start. Long press on any  site in the list to open the multiple editing mode again. Doing this should make the hidden site appear.
4. Select a visible site. This should enable the "Hide" menu at the top.
5. Turn device to landscape mode. This should trigger layout change. Make sure that the selected site remains selected, and the "Hide" menu remains enabled.
6. Switch back to portrait mode.
7. De-select the visible site. Select a hidden site. This should enable the "Show" menu at the top.
8. Turn device to landscape mode. This should trigger layout change. The selected site should remain selected, and the "Show" menu should remain enabled.
9. Switch back to portrait mode.
10. De-select the hidden site. Select a visible site and a hidden site. This should enable both "Show" and "Hide" menu at the top.
11. Turn device to landscape mode. This should trigger layout change. The selected sites should remain selected, and the "Show" and "Hide" menu items should remain enabled.
12. Switch back to portrait mode.
13. Play around with using the Show/Hide/Select All/Deselect All menu options. The functionalities should remain working and unaffected by this PR.

## More info:
This change involves more moving parts than I initially thought, so I wrote down explanation of what the code change did [on this Gist](https://gist.github.com/hafizrahman/6c1e829a52d7714c540b379397e52d09).

## PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/SitePickerActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/SitePickerActivity.java
@@ -84,6 +84,12 @@ public class SitePickerActivity extends LocaleAwareActivity
     private static final String KEY_LAST_SEARCH = "last_search";
     private static final String KEY_REFRESHING = "refreshing_sites";
 
+    // Used for preserving selection states after configuration change.
+    private static final String KEY_SELECTED_POSITIONS = "selected_positions";
+    private static final String KEY_IS_IN_EDIT_MODE = "is_in_edit_mode";
+    private static final String KEY_IS_SHOW_MENU_ENABLED = "is_show_menu_enabled";
+    private static final String KEY_IS_HIDE_MENU_ENABLED = "is_hide_menu_enabled";
+
     private SitePickerAdapter mAdapter;
     private EmptyViewRecyclerView mRecycleView;
     private SwipeToRefreshHelper mSwipeToRefreshHelper;
@@ -97,6 +103,13 @@ public class SitePickerActivity extends LocaleAwareActivity
     private SitePickerMode mSitePickerMode;
     private Debouncer mDebouncer = new Debouncer();
     private SitePickerViewModel mViewModel;
+
+    private HashSet<Integer> mHashSelectedPositions = new HashSet<>();
+    private boolean mIsInEditMode;
+
+    private boolean mShowMenuEnabled = false;
+    private boolean mHideMenuEnabled = false;
+
 
     @Inject AccountStore mAccountStore;
     @Inject SiteStore mSiteStore;
@@ -170,6 +183,11 @@ public class SitePickerActivity extends LocaleAwareActivity
                 return null;
             }));
         }
+
+        // If the picker is already in editing mode from previous configuration, re-enable the editing mode.
+        if (mIsInEditMode) {
+            startEditingVisibility();
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


WordPress/src/main/java/org/wordpress/android/ui/main/SitePickerActivity.java
```diff
@@ -98,6 +104,13 @@
     private Debouncer mDebouncer = new Debouncer();
     private SitePickerViewModel mViewModel;
 
+    private HashSet<Integer> mHashSelectedPositions = new HashSet<>();
```
jd-alexander, line 107: Hi @hafizrahman Nitpick 😅  what do you think about renaming this to `mSelectedPositions` for consistency with all the other variables since they don't contain any identifier of their underlying type?

WordPress/src/main/java/org/wordpress/android/ui/main/SitePickerActivity.java
```diff
@@ -360,6 +384,11 @@ private void restoreSavedInstanceState(Bundle savedInstanceState) {
             isInSearchMode = savedInstanceState.getBoolean(KEY_IS_IN_SEARCH_MODE);
             lastSearch = savedInstanceState.getString(KEY_LAST_SEARCH);
             mSitePickerMode = (SitePickerMode) savedInstanceState.getSerializable(KEY_SITE_PICKER_MODE);
+
+            mHashSelectedPositions = (HashSet<Integer>) savedInstanceState.getSerializable(KEY_SELECTED_POSITIONS);
```
jd-alexander, line 388: Just a note here. We have an `unchecked cast` from `java.io.Serializable` to `java.util.HashSet<java.lang.Integer>` that could lead to a`ClassCastException` if at runtime the type cannot be found. In these situations, a `Parcelable` is normally used to avoid this but I think for this solution we can use this as-is.

WordPress/src/main/java/org/wordpress/android/ui/main/SitePickerAdapter.java
```diff
@@ -433,7 +436,14 @@ void setEnableEditMode(boolean enable) {
         }
 
         mIsMultiSelectEnabled = enable;
-        mSelectedPositions.clear();
+
+        // If some selectedPositions exist and are passed from the activity, use it.
+        // Otherwise, reset the selected positions.
```
jd-alexander, line 441: Just some suggestions to improve the comment a bit. Let me know what you think @hafizrahman 😄 
```suggestion
        // If the selectedPositions value that is passed from the activity isn't empty then we add that to the Adapter's mSelectedPositions otherwise we reset the selected positions .
        // Otherwise, reset the selected positions.
```

WordPress/src/main/java/org/wordpress/android/ui/main/SitePickerActivity.java
hafizrahman: That makes a lot of sense! I did not notice the no-type in variable name rule before, thanks.

WordPress/src/main/java/org/wordpress/android/ui/main/SitePickerAdapter.java
hafizrahman: Yup! Sounds so much better. Added in the next commit.

hafizrahman:
@jd-alexander Thank you for the review! I added a commit for the smaller changes. Still looking at the file conflict now.

hafizrahman:
I think I screwed up this PR because my local copy was old and in trying to update things, a lot of updated but unrelated files got included in the PR instead.

I'm closing this and will make a new PR starting off a recent update instead.

PR #12157, opened 2020-06-10
Feature Branch : Prepublishing Nudges

jd-alexander:
## Solution
All code in this feature branch has been reviewed and tested. The only code modifications that took place were : 

- Fixed lint errors 7206cc7 035165f 
- Resolved conflicts with `develop` during merges

## Reviewing 

## Submitter Checklist


malinajirka:
Just noting that the behavior mentioned [here](https://github.com/wordpress-mobile/WordPress-Android/pull/12149#issuecomment-643068593) is imo a blocker.

jd-alexander:
> Thanks @jd-alexander! I've smoked tested the app and I haven't encountered any crashes or obvious bugs. As I mentioned on Slack, I'd recommend merging these changes after the freeze on Monday to give us a bit more time to test the changes in develop.

Thanks for your review and thoughts @malinajirka 🙏 I will be going ahead with the merge to meet the milestone as planned. I am turning these bugs into issues and I will address them so they can be merged in. 

> Btw I noticed another a bit weird behavior. When I open a draft I created 2 days ago and I click on "Publish", the "Publish Date" row displays the old date. I think the users will expect to see the date which will be recorded not an irrelevant current state of the post, wdyt? (it's similar issue to the "Draft" state we discussed last week). Wdyt?

Thanks! Yes, this should be updated to the current time! I will be tracking this.

jd-alexander:
Thanks for the quick response and help @cameronvoell 😄

PR #12161, opened 2020-06-11
Add GIF support to Gutenberg editor

planarvoid:
This PR adds gif support to the gutenberg editor. The change is pretty simple and follows the same approach like adding images from an external source.

To test (does not work on Release since the feature flag is still turned off):
- Open a post in Gutenberg
- Add an image
- Notice that there is an "Add from Tenor" option
- Click on it
- Search for a GIF
- Add the GIF to the post
- Check that it's correctly uploaded and the post could be saved 
 
PR submission checklist:


malinajirka:
Thanks @planarvoid! The changes look good overall. I've encountered one issue though:

1. Add Image or Media&Text block
2. Click on "Add image"
3. Select Tenor
4. Enter "Dog"
5. Select multiple images
6. Click on "Add"
7. Notice only the first image gets inserted

I also noticed the icon in the menu is identical to "Choose from Free Photo Library". I assume this part is located in React. Perhaps, it's intentional 🤷

frosty:
Fwiw on iOS we only allow a single selection from the gif picker.

planarvoid:
I'm moving this back to draft since it will require a bit more work. We currently don't support a single item selection so we'd need to change the UI and the logic to support it.

khaykov:
Just a heads-up - I offered to help @develric (while @planarvoid is afk) with this adding single selection support to this PR, and here it is - #12244

planarvoid:
hey @malinajirka, this is ready for another check. I've also added a remote config field for tenor setup

planarvoid:
I've reviewed those changes (since they were made by @khaykov ) in another PR so I think this PR is ready to go @malinajirka

PR #12175, opened 2020-06-12
Prepublishing Nudges : bottom sheet shown above keyboard 

### LINKED ISSUE #12150: Prepublishing Nudges : bottom sheet showing above keyboard.
### Expected behavior
That the bottom sheet will not be shown above the height of the keyboard.

### Actual behavior
Sometimes in the Editor when the post is loading if you press Publish you will see the bottom sheet above the space.

1. Create new post. 
2. Click Publish quickly. 
 
<kbd></kbd>

jd-alexander:
Fixes #12150

To test:
1. Create a new post. 
2. Click Publish before the keyboard closes. 
3. Even after it gets collapsed the first time you can press publish again.
4. Ensure the bottom sheet never look like below. 

<kbd></kbd>
PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -1810,7 +1810,11 @@ private void showPrepublishingNudgeBottomSheet() {
         if (fragment == null) {
             PrepublishingBottomSheetFragment prepublishingFragment =
                     PrepublishingBottomSheetFragment.newInstance(getSite(), mIsPage);
-            prepublishingFragment.show(getSupportFragmentManager(), PrepublishingBottomSheetFragment.TAG);
+
+            long delayMs = 100;
+            new Handler().postDelayed(
```
malinajirka, line 1815: Even though it won't happen very often (probably never), I believe this can lead to a leak of the EditPostActivity. We might want to consider creating the handler as a field and canceling it in `onDestroy`. Wdyt?

jd-alexander: Yes, that makes sense! That's a good practice 🙏

jd-alexander: Resolved in 54f61ef

malinajirka:
Thanks @jd-alexander! I've been testing the app and I found two issues:

**Issue 1:**
```
2020-06-15 08:50:24.444 2162-2162/org.wordpress.android.beta E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.wordpress.android.beta, PID: 2162
    java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
        at androidx.fragment.app.FragmentManager.checkStateLoss(FragmentManager.java:1691)
        at androidx.fragment.app.FragmentManager.enqueueAction(FragmentManager.java:1731)
        at androidx.fragment.app.BackStackRecord.commitInternal(BackStackRecord.java:321)
        at androidx.fragment.app.BackStackRecord.commit(BackStackRecord.java:286)
        at androidx.fragment.app.DialogFragment.show(DialogFragment.java:181)
        at org.wordpress.android.ui.posts.EditPostActivity.lambda$showPrepublishingNudgeBottomSheet$19$EditPostActivity(EditPostActivity.java:1804)
        at org.wordpress.android.ui.posts.-$$Lambda$EditPostActivity$HZE7canxsiG-3a0qTcwIcL_YKXc.run(Unknown Source:4)
        at android.os.Handler.handleCallback(Handler.java:790)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:164)
        at android.app.ActivityThread.main(ActivityThread.java:6494)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)
```

Steps to reproduce
- Optional: Increase delay to 2000ms
1. Open Editor
2. Type Something
3. Click on Publish and click on "Home" button right after that
4. Notice the app crashes

**Issue 2:**
- Optional: Increase delay to 2000ms
1. Open Editor
2. Type something
3. Click on publish 3 times as fast as you can
4. Notice that "show" is invoked 3 times (the first two invocation can leak the activity, because no-one ever cancels them)

I have been thinking about it and I think we'll need to use a single Handler and single runnable (not re-create a new one whenever we invoke showPrepublishingNudgeBottomSheet()). Looking at the first crash, I think we'll want to cancel the handler in `onStop` instead of `onDestroy` (sorry for my previous misleading comment). We will also probably want to cancel it in `showPrepublishingNudgeBottomSheet` to make sure we don't leak the activity when the user presses the button multiple times (another option instead of cancelation is to check if the handler is already running and ignore the second press of the user).

SideNote: I'm bumping the milestone to 15.2 as this issue doesn't seems vital. If you feel like we should merge it into 15.1, we can merge a fix into the next version.

jd-alexander:
Thanks for the review @malinajirka I tried to reproduce those crashes but I wasn't able to do so.  

> I have been thinking about it and I think we'll need to use a single Handler and single runnable (not re-create a new one whenever we invoke showPrepublishingNudgeBottomSheet()). Looking at the first crash, I think we'll want to cancel the handler in onStop instead of onDestroy (sorry for my previous misleading comment). We will also probably want to cancel it in showPrepublishingNudgeBottomSheet to make sure we don't leak the activity when the user presses the button multiple times (another option instead of cancelation is to check if the handler is already running and ignore the second press of the user).

Thanks for this! I have logic that ensures the bottom sheet isn't shown multiple times so increasing the delay causing multiple fragments to be added. 

Canceling the runnable in the appropriate lifecycle methods is always a good solution to prevent leaks. I will implement this. 

> SideNote: I'm bumping the milestone to 15.2 as this issue doesn't seems vital. If you feel like we should merge it into 15.1, we can merge a fix into the next version.

Appreciate it 🙏

PR #12177, opened 2020-06-12
Epa refactoring on upload error

malinajirka:
Refactors "OnUploadError" method in the EditPostActivity. There shouldn't be any changes in the behavior.

To test:
1. Enable "Collect information" in app privacy settings
2. Open Gutenberg
3. Turn on airplane mode
4. Add an image block and try to upload an image
5. Notice the track event in LogCat - something similar to 
`Tracked: editor_upload_media_failed, Properties: {"ext":"jpg","age_ms":685,"bytes":19951,"mime":"image\/jpeg","megapixels":0,"error_type":"CONNECTION_ERROR"}`

- Perform the same test for VideoBlock
- Test both image and video in Aztec

PR submission checklist:


libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/AztecEditorFragment.java
```diff
@@ -1396,30 +1396,22 @@ public void onMediaUploadProgress(final String localMediaId, final float progres
     }
 
     @Override
-    public void onMediaUploadFailed(final String localMediaId, final EditorFragmentAbstract.MediaType
-            mediaType, final String errorMessage) {
+    public void onMediaUploadFailed(final String localMediaId) {
```
malinajirka, line 1399: I've removed two parameters here
1. ErrorMessage - this parameter wasn't being used at all
2. MediaType - this parameter was used in the switch. But if you check the switch, it executes the same code no matter the mediaType. The only case when the mediaType has some effect on the code path is when it's null. However, [this](https://github.com/wordpress-mobile/WordPress-Android/blob/ee93098546f355636c6151b39c61ccd3de6a8836/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragmentAbstract.java#L165) method used to get the MediaType never returns null. Therefore, I believe it's safe to remove it altogether.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -1501,28 +1501,7 @@ private void onUploadSuccess(MediaModel media) {
         }
     }
 
-    private void onUploadError(MediaModel media, MediaError error) {
-        String localMediaId = String.valueOf(media.getId());
-
-        Map<String, Object> properties = null;
```
malinajirka, line 1507: This code was doing multiple things 
1. Track failed upload
2. Construct error message
3. Determine MimeType/MediaType

However, as mentioned below - error and mediaType was never being used. So the only remaining thing is the tracking which was moved to MediaEditor.

PR #12181, opened 2020-06-12
Docs: Add issue triage doc

rachelmcr:
This PR adds documentation for what issue triage is, why we do it, and how it's done (labels and prioritization).

This corresponds with the iOS doc added in https://github.com/wordpress-mobile/WordPress-iOS/pull/13925 and documents the agreed-on priority matrix (internal ref: paaHJt-UE-p2). As the processes are updated/clarified we can update the doc in both repos.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -18,6 +18,7 @@
 ## Others
 
 - [Frequently Asked Questions](faq.md)
+- [Issue Triage](issue-triage.md)
 
 ## Further Reading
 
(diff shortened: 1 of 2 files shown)


rachelmcr:
@designsimply This should be pretty straightforward to review since it matches the iOS docs. Since I'll be AFK for the next week, feel free to merge (or make any needed edits and then merge) while I'm out.

PR #12185, opened 2020-06-12
Issue/10726 site pages author picker

### LINKED ISSUE #10726: Post List: cannot see author picker for pages
### Expected behavior

I expected to be able to filter pages by me or everyone.

### Actual behavior

There's a space for an author filter toggle at the left side of the pages filter lost, but the option to toggle between my own pages and everyone's pages is missing.

1. Switch to a site with published pages from more than one author.
1. Go to My Site > Site Pages.
1. Look for the author picker at top left.

##### Tested on Pixel 3, Android 10, WPAndroid 13.5-rc-2

Screenshot_20191101-135948 Screenshot_20191101-135840
<sup>Site Pages author toggle cannot be seen while logged in as design5279 admin role (left) and then testuser0313 author role (right).</sup>

Screenshot_20191101-140006 Screenshot_20191101-140002
<sup>Compare to Blog Posts where the author toggle can be seen while logged in as design5279.</sup>

zwarm:
Fixes #10726 

This is a master-feature PR. The code has already been reviewed. However, it'd be great if someone with more experience with this feature take a look.

Goal of this series of PRs was to add an author filter to the pages view so that it matches functionality with the posts feature. The author name will be shown on the pages list when the author selector is set to Everyone. Search results are not filtered; however author name is shown.

There is one specific item that I would appreciate feedback on:
    - The Posts feature saves the author filter state in app prefs. I implemented Pages to use the same appPrefs property, so the filter selection is maintained across pages to posts. I was 50-50 on setting up a new property and I am open to suggestions.
    

**To test:**
- Launch WPAndroid
- Select a site not owned by you, but one in which you are an author
- Navigate to Pages
- Toggle the author selector to Me
- Ensure the author name is not shown
- Toggle the author selector to Everyone
- Ensure the author name is shown
- Tap the search icon
- Search for a page
- Ensure the author name is shown in the search results

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/pages/PageItemViewHolder.kt
```diff
@@ -176,6 +163,42 @@ sealed class PageItemViewHolder(internal val parent: ViewGroup, @LayoutRes layou
                 }
             }
         }
+
+        @ExperimentalStdlibApi
```
malinajirka, line 167: Just a note: This logic would ideally be located in the VM (and related classes). However, this refactoring is above the scope of this PR.

WordPress/src/main/java/org/wordpress/android/ui/pages/PagesFragment.kt
```diff
@@ -264,6 +278,26 @@ class PagesFragment : Fragment() {
             savedInstanceState.getSerializable(WordPress.SITE) as SiteModel
         }
 
+        viewModel.authorUIState.observe(activity, Observer { state ->
+            state?.let {
+                val authorSelectionVisibility = if (state.isAuthorFilterVisible) View.VISIBLE else View.GONE
```
malinajirka, line 283: We could use `UiHelpers.updateVisibility()` here. But it's not a big deal.

WordPress/src/main/java/org/wordpress/android/ui/pages/PagesFragment.kt
```diff
@@ -264,6 +278,26 @@ class PagesFragment : Fragment() {
...
+                pages_author_selection.visibility = authorSelectionVisibility
+                pages_tab_layout_fading_edge.visibility = authorSelectionVisibility
+
+                val tabLayoutPaddingStart =
```
malinajirka, line 287: We might want to consider moving this logic to the VM and utilizing UIDimen. Wdyt?

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageListViewModel.kt
```diff
@@ -124,6 +128,7 @@ class PageListViewModel @Inject constructor(
 
             pagesViewModel.pages.observeForever(pagesObserver)
             pagesViewModel.invalidateUploadStatus.observeForever(uploadStatusObserver)
+            pagesViewModel.authorSelectionUpdated.observeForever(authorSelectionChangedObserver)
```
malinajirka, line 131: I guess this approach is fine, since it's already used in this class. However, in general I think we should consider `observeForever` as a last resort and "code smell".

WordPress/src/main/res/values/strings.xml
```diff
@@ -2708,8 +2713,6 @@
 
     <!-- Posts -->
     <string name="post_list_author">Post author</string>
-    <string name="post_list_author_me">Me</string>
```
malinajirka, line 2711: When we remove/rename string we also remove all the translations. Even though the key of the string might not be ideal, I'd consider keeping `post_list_author` and `post_list_author_me` and introducing
<string name="page_list_author_me" translatable="false">@string/post_list_author_me</string>
...

Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/pages/PageItemViewHolder.kt
```diff
@@ -176,6 +163,42 @@ sealed class PageItemViewHolder(internal val parent: ViewGroup, @LayoutRes layou
...
+                        parent.context.getString(R.string.pages_item_subtitle),
+                        stringDate,
+                        parent.context.getString(subtitle))
+            } else {
+                String.format(
+                        Locale.getDefault(),
+                        parent.context.getString(R.string.pages_item_subtitle_date_author),
+                        stringDate,
+                        author,
+                        parent.context.getString(subtitle!!))
```
malinajirka, line 199: Just checking: Is the `!!` intentional?

WordPress/src/main/java/org/wordpress/android/ui/pages/PageItemViewHolder.kt
zwarm: Agreed - I went back and forth on it myself. Good next hackweek project to clean this viewholder up.

WordPress/src/main/java/org/wordpress/android/viewmodel/pages/PageListViewModel.kt
zwarm: Agreed, kept with the model used in the class for this issue - managing the observe and then remove. But yes, observerForever is always active and probably not needed for any of these here.

WordPress/src/main/java/org/wordpress/android/ui/pages/PageItemViewHolder.kt
zwarm: Yes, it was intentional because subtitle is nullable. I'll wrap it in a let instead and commit.

WordPress/src/main/java/org/wordpress/android/ui/pages/PagesFragment.kt
zwarm: TIL - thanks. Changed in the next commit.

WordPress/src/main/java/org/wordpress/android/ui/pages/PagesFragment.kt
zwarm: That's a really good idea, but for time constraint, going to leave it where it is for now. Adding to my  back pocket task list.

WordPress/src/main/res/values/strings.xml
zwarm: I did not know that about strings. I went ahead and added back the Me & Everyone strings. Thanks for the tip and for teaching me.

zwarm: In commit 879ff76

WordPress/src/main/java/org/wordpress/android/ui/pages/PageItemViewHolder.kt
malinajirka: You didn't have to change it. I honestly just asked if it was a leftover or an intentional usage. Either case the let looks good, thanks!

WordPress/src/main/res/values/strings.xml
zwarm: aha - I can't leave those in there because they are unused and tests will fail. Going to pull them out and remove from all translation string files. Removed in commit 6d428d4

malinajirka: I'm not sure what has happened with my comment but I'm pretty sure I posted
```and introducing <string name="page_list_author_me" translatable="false">@string/post_list_author_me</string>"```

This would ensure that we re-use the existing translations for "me" and "everyone".

zwarm: I took them out. Lesson learned.

malinajirka:
LGTM, thanks! Feel free to merge it when the CI finishes.

PR #12191, opened 2020-06-13
Display content information metrics

hypest:
Expands #12186 to include the WPAndroid UI to get and show the content information metrics.

Gutenberg-mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/2383

Follows this design, shared by @iamthomasbishop over Slack:

Screen Shot 2020-06-12 at 11 51 51 AM

Changes introduced:
1. Adds a "Content info" menu time in the 3-dot menu while editing a post/page. The menu item triggers a call to get the latest content, including the content information.
2. Upon receiving the latest content and info, a popup dialog displays the information

To test:
1. Run the app and open an existing post, preferably a long one
2. While the spinning progress bar is still on, tap on the "Content info" menu item inside the 3-dot menu
3. Notice the "Editor is still loading" toast being shown
4. Wait for the post to finish loading
5. Tap on the "Content info" menu item again
6. Notice the popup dialog with the block/word/character counts displayed

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,7 @@
+15.2
+-----
+* [**] Block editor: Display content metrics information (blocks, words, characters count).
+
 15.1
 -----
 * Fixes issue on Notifications tab when two screens were drawn on top of each other
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
@@ -16,7 +16,6 @@ import android.view.DragEvent;
 import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
-import android.view.MenuItem.OnMenuItemClickListener;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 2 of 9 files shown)


hypest:
(replying to [this comment](https://github.com/wordpress-mobile/gutenberg-mobile/pull/2380#issuecomment-643528033))
> Is it possible to only add the plural ("s") when the given type's value > 1? In other words "1 block" vs. "2 blocks", etc.

👋 @iamthomasbishop , to answer the same for Android, it won't be very easy because we don't currently support plural variants on GlotPress. We usually workaround that by having two separate strings defined, one for singular and one for plural. Thing is, in the sentence at hand, there are three numbers that can individually be singular or plural and even though we could concatenate them as three separate sections, that's not recommended as we can't be sure if it will be correct for all locales. It's better to send the full sentence to the translators, but that will mean we'll need some complex logic to define and use those strings. Not sure it's worth it.

Instead, and again talking about the WPAndroid case where I've implemented the popup dialog design, what do you think about using plurals only but formatting the whole string like this:
```
Blocks: number
Words: number
Characters: number
```

I think that would read OK for numbers being `1` too.

PR #12196, opened 2020-06-13
Prepublishing Nudges : Persist post changes that occur in the bottom sheet 

### LINKED ISSUE #12195: Prepublishing Nudges : Persist the prepublishing bottom sheet post in the post list
### Expected behavior
That the post should be persisted so when orientation changes or if the app gets killed the post's state will be preserved. 

### Actual behavior
The post is not persisted so the changes are not reflected in the Post List's prepublishing bottom sheet when an orientation change takes place.

2. Click Publish.
3. Make a change to the Publish Date. 
4. Rotate the screen and navigate back. 
5. Notice that the Publish Date is not the same as the one that was set. 

jd-alexander:
Fixes #12195
This is a partial solution to get this conversation started. More details below. 

## Partial Solution
Persist the post so the content that the user changes get persisted. This ensures that configuration changes do not cause the user to lose the changes that they make. 

Previously, we agreed that the post didn't have to get updated because the user would lose their changes to the post if they decided not to publish that specific post. But based on this specific bug that occurred it's best if the `PostModel` that gets modified is persisted so it can always be a source of truth in case there's any state loss.

## Concerns

1. So this solution just persists the post to the database but it gets overwritten when you do a refresh is done from the server. ( Making an assumption here.)Therefore, the user won't have these changes if they come off this screen. 
2. Since the `Post List` wasn't created for this kind of post-editing, I went searching for other components that had a close enough flow. I found this use case `SavePostToDbUseCase` and it seems to save the post with the appropriate business logic to signify the changes made. Nonetheless, it's reliance on `context` doesn't sit well with me especially since I just wrote lint rules to avoid this 😅  but it seems like the better full proof solution. 
3. I notice that because the `PostModel` is being modified and the list is being invalidated, certain post actions are disappearing as a result. 

<kbd></kbd>

4. Also in the above screenshot, I am realizing that the `Private` post doesn't show `Immediately` when the post is not a local draft, so even though it's showing the current date it doesn't say `Immediately`. 

## Testing
1. Go to the Draft Tab on the Post List. 
2. Click Publish on a Post. 
3. Change any of the options and don't navigate from the screen when it's changed.
4. Rotate and ensure that when you navigate back you see the data/selection. 
5. If you close the bottom sheet the Post in the Post List should be updated to reflect the new state. 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
@@ -691,7 +691,7 @@ public class EditPostActivity extends LocaleAwareActivity implements
             return null;
         }));
         mEditPostRepository.getPostChanged().observe(this, postEvent -> postEvent.applyIfNotHandled(post -> {
-            mViewModel.savePostToDb(this, mEditPostRepository, mSite);
+            mViewModel.savePostToDb(mEditPostRepository, mSite);
             return null;
         }));
     }
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PostListMainViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PostListMainViewModel.kt
@@ -1,5 +1,6 @@
 package org.wordpress.android.ui.posts
 
+import android.content.Context
 import android.content.Intent
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleOwner
@@ -20,7 +21,6 @@ import org.wordpress.android.analytics.AnalyticsTracker.Stat.POST_LIST_SEARCH_AC
 import org.wordpress.android.analytics.AnalyticsTracker.Stat.POST_LIST_TAB_CHANGED
 import org.wordpress.android.fluxc.Dispatcher
 import org.wordpress.android.fluxc.generated.ListActionBuilder
-import org.wordpress.android.fluxc.generated.PostActionBuilder
 import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
 import org.wordpress.android.fluxc.model.PostModel
 import org.wordpress.android.fluxc.model.SiteModel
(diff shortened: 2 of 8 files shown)


malinajirka:
>So this solution just persists the post to the database but it gets overwritten when you do a refresh is done from the server. ( Making an assumption here.)Therefore, the user won't have these changes if they come off this screen.

We can (should) mark the post as "isLocallChanged = true". The local changes will never get overwritten by the server version unless the user explicitly confirms it. This logic is located [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/75050e8e588c95ea5ba836194ece8aa0011b8472/fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java#L784).

>I found this use case SavePostToDbUseCase and it seems to save the post with the appropriate business logic to signify the changes made.

Yeah, this class seems like exactly what we need :). EditPostActivity refactoring is already paying off :P.

>Nonetheless, it's reliance on context doesn't sit well with me especially since I just wrote lint rules to avoid this 😅 but it seems like the better full proof solution.

It seems the class could get away with the application context, which can be automatically injected into its constructor. I'm not sure about this, wdyt?

>I notice that because the PostModel is being modified and the list is being invalidated, certain post actions are disappearing as a result.

If I understand this correctly, it's a good behavior, right? When we change the post status we also need to change the actions as they are status specific. Wdyt?
If you think we should change them when the user dismisses the dialog we could theoretically save the post in "onDismiss/onButtonPressed". The user would lose their changes if the app crashed, but it shouldn't be a big deal since the amount of changes they can make in the bottom sheet is very limited.

>Also in the above screenshot, I am realizing that the Private post doesn't show Immediately when the post is not a local draft, so even though it's showing the current date it doesn't say Immediately.

Ohh, I noticed this last week and I forgot to report it :(. Yeah, I think it's a bug - even though it's minor.



I think you are heading in a great direction! Great job on finding the relevant code which could be reused!

jd-alexander:
Thanks @malinajirka I will be working on finalizing this shortly!

jd-alexander:
Let me know what you think about this. I haven’t tested all scenarios due to webinar prep. But updated the PR to get feedback.

malinajirka:
Thanks @jd-alexander! It looks good overall. I'd just suggest getting rid of the context in the VM. We can inject application context directly into the `SavePostToDbUseCase`. Wdyt?

jd-alexander:
> Thanks @jd-alexander! It looks good overall. I'd just suggest getting rid of the context in the VM. We can inject application context directly into the SavePostToDbUseCase. Wdyt?

Oh nice! Just realized that can be done.

malinajirka:
LGTM, thanks!

PR #12198, opened 2020-06-13
Prepublishing : Old draft state 

### LINKED ISSUE #12197: Prepublishing Nudges : Old draft state for date
### Expected behavior
Opening an old draft and clicking on "Publish" should show the "Publish Date" as the current date.

### Actual behavior
Opening an old draft and clicking on "Publish" shows the "Publish Date" row with the old date.

2. Click Publish on a post created a day or so ago ( either from the Post List or Editor)
3. Notice the older date being shown. 

jd-alexander:
Fixes #12197

Shared to get thoughts. 

## Findings
When a draft is opened in the Post List or Editor and the date is in the past it gets cleared.
## Solution
`PublishPostImmediatelyUseCase` was used to add this functionality to the Editor & Post List.

## Testing
1. Go to the Post List. 
2. Press Publish on an old draft. 
3. Notice it says Immediately instead of the old date. 
[Video Example](https://drive.google.com/file/d/1n7GQCYQqMP5-DwtXmV8vO_ev4RyLLDkn/view?usp=sharing)
--------

## Testing
1. Go to the Post List. 
2. Edit an old draft.
2. Press Publish (if you change the status to Draft, the outcome below works). 
3. Notice it says Immediately instead of the old date. 

[Video Example](https://drive.google.com/file/d/12XZorKwwQSvrar6RL-Qig4yurZa0A-fG/view?usp=sharing)

So the above happens because when a post is not new it's deemed as `Published` so this logic needs to be improved.

## Reviewing 

## Submitter Checklist


malinajirka:
Thanks for working on this @jd-alexander!

I haven't reviewed the code line by line, but the overall approach seems ok to me. AFAIU the issue you are dealing with is that when the user changes a status of a published post to draft it resets the original publish date, right? Tbh I don't consider that to be an issue - the user changed the status, so reseting the publish date seems fine.

jd-alexander:
> I haven't reviewed the code line by line, but the overall approach seems ok to me. AFAIU the issue you are dealing with is that when the user changes a status of a published post to draft it resets the original publish date, right? Tbh I don't consider that to be an issue - the user changed the status, so reseting the publish date seems fine.

Thanks for this review! I will do another run through and then make this ready for review!

jd-alexander:
Let me know what you think about this. I haven’t tested all scenarios due to webinar prep. But updated the PR to get feedback.

WordPress/src/main/java/org/wordpress/android/ui/posts/PostListMainViewModel.kt
```diff
@@ -432,12 +435,20 @@ class PostListMainViewModel @Inject constructor(
         )
     }
 
-    private fun showPrepublishingBottomSheet(post: PostModel) {
+    fun showPrepublishingBottomSheet(post: PostModel) {
```
malinajirka, line 438: We might want to consider renaming this method to something else - it doesn't just "show bottom sheet" anymore. Wdyt?

AFAIK this method is used only when the user clicks on "Publish" button right? Shouldn't we always update the date to the current date?

WordPress/src/test/java/org/wordpress/android/ui/posts/PostListMainViewModelTest.kt
```diff
@@ -186,4 +195,50 @@ class PostListMainViewModelTest : BaseUnitTest() {
...
+        // act
+        viewModel.start(site, PostListRemotePreviewState.NONE, currentBottomSheetPostId, editPostRepository)
+        viewModel.showPrepublishingBottomSheet(mock())
+
+        // assert
+        verify(publishPostImmediatelyUseCase, times(1)).updatePostToPublishImmediately(any(), any())
+    }
+
+    @Test
+    fun `if PostStatus is DRAFT and Publish Date is is in the future then don't call publish immediately use case`() {
```
malinajirka, line 216: Is there a scenario in which the user clicks on "Publish" but the date is in the future? Do we really want to show the future date but publish the post immediately?

WordPress/src/main/java/org/wordpress/android/ui/posts/PostListMainViewModel.kt
jd-alexander: Good find. can this approach work? b51fd05

WordPress/src/test/java/org/wordpress/android/ui/posts/PostListMainViewModelTest.kt
jd-alexander: No, we don't! b51fd05

WordPress/src/main/java/org/wordpress/android/ui/posts/PostListMainViewModel.kt
malinajirka: Yeah, I think it should work. Thanks!

malinajirka: Hmm so I was wrong. I think the `postStatus == DRAFT` condition needs to stay there. We show the publish button even on the published tab when the post has some local changes.

Btw I also noticed that when I change the post status in the post settings the change doesn't get stored. I think I already reported this, but mentioning it again just to be sure we didn't miss it.

post-status-lost

malinajirka:
Thanks @jd-alexander! I've found one last issue, but it looks good otherwise.

WordPress/src/main/java/org/wordpress/android/ui/posts/PostListMainViewModel.kt
jd-alexander: Good find. I am wondering if this change is related to this project. I am going to investigate this because at first glance it seems unrelated.

malinajirka: Definitely worth checking 👍.

malinajirka: I can confirm it's not related to these changes and is actually expected behavior - https://github.com/wordpress-mobile/WordPress-Android/blob/978f3ab9d27f3dde01f1488fc7a8c7688e5f1416/WordPress/src/main/java/org/wordpress/android/ui/uploads/PostUploadHandler.java#L630

jd-alexander: Awesome. Thanks for confirming 🙏

jd-alexander:
Thanks Jirka. I’m having some Gutenberg issues so I will resolve that and then resolve the conflicts and merge 🙏🏾

jd-alexander:
@mzorz the only code change since Jirka's review is bc88b9a 4288508 where I fixed some issues that were causing the tests to fail due to related changes being merged in. So once these look good and the functionality works, we can merge. Thanks 🙏

mzorz:
Just tested the latest, found some weird stuff for me on a Pixel 3 emulator with the jalapeno build.

#### Aztec:
Didn't work, see screenshot
steps to reproduce:
1. find an old draft that is written on Aztec.
2. follow steps for opening an old draft and tap PUBLISH.
3. observe the error.

Screen Shot 2020-06-29 at 17 55 33


#### Posts list:
I verified it works up to the steps provided, but I noticed something  - if you go to the Posts list and tap on “Publish” for an old draft, but then just tap back (and I noticed also, I have to tap back **twice** to dismiss the bottom sheet, observed as well above in the editor), and then go to the Posts list again: the Post will show itself as "updating" and now the last edit time will be "now". But I didn't do anything to it, it should have remained with the old date.
The same happens when opening the bottom sheet from the editor: if you dismiss it, it still produced a change in the Post and now reads "now".

Done it again and now it says "Local changes" (there should not be any change to my post, right?)

See here:
botttomsheetupdatetlist

#### Gutenberg
Didn't work for me, see gif here:

botttomsheetupdatet



Can you take a look and see if you see the same issues there @jd-alexander ?

jd-alexander:
Hey @mzorz thanks for testing. 

> I verified it works up to the steps provided, but I noticed something - if you go to the Posts list and tap on “Publish” for an old draft, but then just tap back (and I noticed also, I have to tap back twice to dismiss the bottom sheet, observed as well above in the editor), and then go to the Posts list again: the Post will show itself as "updating" and now the last edit time will be "now". But I didn't do anything to it, it should have remained with the old date.

Hey, @mzorz this is the expected behavior because what the logic is doing is that it clears the old date and sets the current date. 

> Done it again and now it says "Local changes" (there should not be any change to my post, right?)

Yes, it should show this because we are changing the date to "now" / Immediately, so it now has changes. 

> Didn't work for me, see gif here:

Oh for here, it just shows the current date instead of `Immediately` due to the status being `Published`. @malinajirka do you think that's something we should address for this PR as well since it's related to the behavior we are trying to implement here?

mzorz:
> Hey, @mzorz this is the expected behavior because what the logic is doing is that it clears the old date and sets the current date.

Just to double check: I meant when tapping on Publish, then the bottom sheet is presented. If at that point I choose not to go ahead and as such tap back, I'm surprised the Post still was updated. I wouldn't expect the post to be updated if I declined / didn't confirm any intent. Is this still expected behavior?

I understand what you say that the logic clears the old date and sets the current date and therefore that's what I can see in the tests, but that's exactly what I'd argue / don't fully understand: why do we update it when it's not what the user wants? note it also changes the order in the list (most recent first). Let me know your thoughts.

> Oh for here, it just shows the current date instead of Immediately due to the status being Published

Note that the Post I'm showing in the example is in the Drafts list, so not sure why / how come the status is published (it shouldn't be right? I'm tapping on an item in the Drafts list after all)

malinajirka:
Thanks for such a thorough tests @mzorz ! ;)

>Just to double check: I meant when tapping on Publish, then the bottom sheet is presented. If at that point I choose not to go ahead and as such tap back, I'm surprised the Post still was updated. I wouldn't expect the post to be updated if I declined / didn't confirm any intent. Is this still expected behavior?

I agree that this is arguable. However, I'm not sure I agree with this part -> `didn't confirm any intent` -> you already clicked on "Publish". We require the confirmation so we don't publish something the user didn't want to publish. AFAIU we still don't publish the post, right? We just change its state so it's prepared to be published. If you for example set 5 tags in the bottom sheet and then dismiss it, would you expect to lose your changes?

>Note that the Post I'm showing in the example is in the Drafts list, so not sure why / how come the status is published

This really depends on whether the post is published or not. It can have "publish" status but not be published yet - in other words it can contain local changes which haven't been confirmed yet, therefore the post hasn't been published yet.
If it's a post displayed in "draft" tab and it doesn't contain any changes, then you found a bug :).

mzorz:
> I agree that this is arguable. However, I'm not sure I agree with this part -> didn't confirm any intent -> you already clicked on "Publish". We require the confirmation so we don't publish something the user didn't want to publish. AFAIU we still don't publish the post, right? We just change its state so it's prepared to be published. If you for example set 5 tags in the bottom sheet and then dismiss it, would you expect to lose your changes?

I agree with the tags being saved (and as such, the Post being updated) case 👍 ,  that's a good case where we want to be "clever" and do things for users (users _did_ something, and we want to hold their backs and save what they did - it matches the expectation they'd have that their actions are preserved).

I don't agree with the date being changed if the user did not do anything else than tapping on `Publish` the first time before giving consent / confirmation, as this even breaks away from pre-existing behavior 🤷 - we should still go an extra mile and leave things "untouched" when the user does nothing else but taps `back` when in the pre-publishing nudge. As a user, I think I would very surprised and lose confidence (i.e. several thoughts would cross my mind like "hey what just happened? what else has been modified in this post? why? I didn't do anything!" etc). I think it's a sensible matter.

I think we could potentially have a variable that "remembers" the actual date before changing it, and just re-set it if they just tap back and nothing else is touched?
I mean, if you definitely need the date to be set to "now" so to be able show the "Immediately" label (which is what I believe this PR is trying to address), then if that prepares the state of things well for how the pre-publishing bottom sheet handles things, you could do some of the following:

For example:
1. before this [line here ](https://github.com/wordpress-mobile/WordPress-Android/pull/12198/files#diff-38b6aea9cfff5f3e0e4b7389b44de602R1438) we could remember the actual date and pass it as an argument to the bottom sheet fragment where you create it (in the runnable instantiation),
2. when the user touches anything on the bottom sheet UI except closing or tapping back we could flag the UI dirty (just keep a `var dirty: Boolean` sort of thing)
3. and then in `onBackClicked`  and/ or `onCloseClicked`  if the dirty flag is not set, then set the post's date to what it was back.

You could do things more elaborately and use something fancier as an interface that lets you rollback a set of variables etc but I think this alone would still go a long way already in covering this concern.

Let me know your thoughts @malinajirka @jd-alexander 🙇

malinajirka:
👍 I think you are right. It'd feel more user friendly if we reverted the date back to the previous value. Since the app can crash while the bottom sheet is opened I'd consider faking the "immediate" label instead of actually changing the date in the post field. I'd change the actual field of the post only when the user manually changes the date or when they click on the submit button in the bottom sheet. I'm not sure how complicated would that be, wdyt @jd-alexander ?

mzorz:
>  I'd consider faking the "immediate" label instead of actually changing the date in the post field. I'd change the actual field of the post only when the user manually changes the date or when they click on the submit button in the bottom sheet

That sounds good 👍

malinajirka:
Btw the last gif Mario shared in https://github.com/wordpress-mobile/WordPress-Android/pull/12198#issuecomment-651368269 really looks like a bug to me. Notice, how the primary button in the editor changes from "Publish" to "Update" right after Mario clicked on it. Wdyt @jd-alexander ?

jd-alexander:
Great conversation @mzorz & @malinajirka 
Thank you so much for all these comments. They have truly helped me to understand this issue in a deeper way and I totally agree with the solutions that are floating around here to create the optimum experience for our users. Let me address some specific action items. 

>  Since the app can crash while the bottom sheet is opened I'd consider faking the "immediate" label instead of actually changing the date in the post field. I'd change the actual field of the post only when the user manually changes the date or when they click on the submit button in the bottom sheet. I'm not sure how complicated would that be, wdyt @jd-alexander ?

1. Thanks @malinajirka this is actually a really good idea. It shouldn't be too difficult. All I would need to do is check if a `DRAFT` has an older date and show `Immediately`. This logic is located in the `PostSettingsUtil` so it would be added there so that it can be used by the prepublishing bottom sheet and the `Post Settings`. 

2. Then for the publishing part of things, I checked the logic in the `preparePostForPublish` function of the `PostUtils` and it does indeed clear the old date and set the current date, so we are good on that side :) 

> Btw the last gif Mario shared in #12198 (comment) really looks like a bug to me. Notice, how the primary button in the editor changes from "Publish" to "Update" right after Mario clicked on it. Wdyt @jd-alexander ?

Yes, this looks like a bug. Probably the `Published` status was added mistakenly. I haven't checked the code but thinking about the changes I made, the issue could be that this draft being opened after being created gets marked as a post that isn't new so the status gets set to `Published` for some reason. I will check this out. Thanks.

jd-alexander:
I am going to be resolving this in another PR since this was targeting a release branch and pretty much all the commits here won't be needed.

Hey @mzorz I wanted your feedback on this! So I did the fix we had spoken about for this issue where we are setting the Publish Date text as `Immediately` instead of changing the actual date of the `PostModel` The only issue I realized was that the Calendar would still have the old date instead of the new one and only if the user presses the `Immediately` button would there be a change to the current date. Is this something we can live with for this fix? 

 
Before | After 
--------|-------
 <kbd></kbd> | <kbd></kbd>

mzorz:
> Is this something we can live with for this fix?

For sure! Seems like a pretty minor thing 👍

jd-alexander:
Thanks @mzorz ! closing this and will link the other PR with this fix here.

PR #12201, opened 2020-06-15
RI: Choose interests - introduce failed state

ashiagr:
Master Issue: #12027

This PR introduces failed state into the Choose Interests screen.
Currently just reused site creation fullscreen error layout to display error message, retry button. 

### To test
- Enable airplane mode on the device
- Launch app and go to Reader tab
- Notice that no network error message is shown along with a Retry button
- Disable airplane mode on the device
- Click Retry
- Notice that progress bar is shown and interests chips are displayed after a while

#### Note
- This PR also refactors code to remove interests related fields from the parent `UiState` and include them only in the `ContentLoadSuccessUiState`.
Ref: https://github.com/wordpress-mobile/WordPress-Android/pull/12113#discussion_r440149546

- As the tags repository currently does not return any errors, so just handling network error.  

- I can see a "No Connection" snackbar when there's no connection. I'm not sure if we'd want to reuse it instead of full screen error layout and how we can retry n/w request in this case (considering we'll also handle API errors later).

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
```diff
@@ -37,15 +41,20 @@ class ReaderInterestsViewModel @Inject constructor(
...
-        viewModelScope.launch {
-            val tagList = readerTagRepository.getInterests()
-            val currentUiState = uiState.value as UiState
-            updateUiState(
-                ContentLoadSuccessUiState(
-                    interestTagsUiState = transformToInterestsUiState(tagList),
-                    interestTags = tagList,
-                    doneBtnUiState = currentUiState.getDoneButtonState()
+        updateUiState(LoadingUiState)
+        if (networkUtils.isNetworkAvailable()) {
```
malinajirka, line 45: I believe the VM doesn't/shouldn't even know where is the data retrieved from. Moreover, if we have some data in the cache I assume, we'll want to show them instead of showing a fullscreen error. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
```diff
@@ -89,49 +102,73 @@ class ReaderInterestsViewModel @Inject constructor(
...
-            doneButtonUiState = doneBtnUiState,
             progressBarVisible = false,
             titleVisible = true,
-            subtitleVisible = true
+            subtitleVisible = true,
+            fullscreenErrorLayoutVisible = false
         )
 
-        private val checkedInterestsUiState = interestsUiState.filter { it.isChecked }
+        sealed class ContentLoadFailedUiState constructor(
```
malinajirka, line 126: I'd consider renaming this to something like "FullScreenErrorUiState". The fact that the content load fails doesn't necessarily mean we'll show a progress so the name might be a bit misleading in some scenarios. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
ashiagr: You're right. Removed n/w condition here: 8a83960.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
ashiagr: Renamed here 1097a5a to keep them shorter and consistent with names used in Discover page.

ashiagr:
Thanks for the review, ready for another look !

PR #12218, opened 2020-06-18
RI: Choose Interests - Navigate to discover

ashiagr:
Master Issue:  #12027

This PR attempts to show and close "Choose Interests" screen from the `ReaderFragment`. 

I couldn't find a scenario in the app where we navigate from 
_bottom nav tab fragment_ => _full screen fragment_ that remains above the bottom nav bar. 

Let me know if there is one.

Tried to use `WPMainNavigationView` to add `ReaderInterestsFragment` to backstack for Reader tab and pop it when Done button is clicked but finally used below approach as it appeared to be simpler:

   - Start with a hidden `AppBar` inside the `ReaderFragment` without any tabs 
   - Add `ReaderInterestsFragment` as a child fragment.
   - On Done button click inside the `ReaderInterestsFragment`, use the parent fragment's viewmodel (`ReadeViewModel`) to trigger these actions:
        - Remove `ReaderInterestsFragment` child fragment
        - Fetch reader viewpager tabs 
        - Unhide `AppBar` and start displaying the reader tabs

### To test

1. Launch the app
2. Go to Reader tab
3. Notice that  "Choose Interests" screen is shown
4. Choose some interests and press the Done button
5. Notice that  "Choose Interests" screen is closed and Discover tab is shown

----

6. Close the app and Repeat Steps 1 - 3
7. Notice that  "Choose Interests" screen is shown
8. Click device back button
9. App goes to background (as we were still on the `ReaderFragment` attached to the bottom nav bar)

#### Notes

 - The [condition to show the screen](https://github.com/wordpress-mobile/WordPress-Android/blob/9b86e7dba03bb333bfb48b8c2fafa9788902f6d5/WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt#L46)  is not final. It will depend on this separate task:
      - [ ] Fetch List of user's tags and show "Select Topics" screen only if the list is empty - get mocked list of tags from ReaderRepository

As we're still finalising the approach, pending tasks for this PR
- [X] Test app for config changes, bottom bar tab toggles etc 
- [ ] Add transition animations 

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderFragment.kt
@@ -125,7 +125,8 @@ class ReaderFragment : Fragment(R.layout.reader_fragment_layout) {
                         updateTabs(it)
                     }
                 }
-                uiHelpers.updateVisibility(app_bar, uiState.appBarVisible)
+                app_bar.setExpanded(uiState.appBarExpanded)
+                uiHelpers.updateVisibility(tab_layout, uiState.tabLayoutVisible)
                 searchMenuItem?.isVisible = uiState.searchIconVisible
             }
         })
@@ -138,7 +139,7 @@ class ReaderFragment : Fragment(R.layout.reader_fragment_layout) {
 
         viewModel.selectTab.observe(viewLifecycleOwner, Observer { selectTabAction ->
             selectTabAction.getContentIfNotHandled()?.let { tabPosition ->
-                view_pager.currentItem = tabPosition
+                view_pager.setCurrentItem(tabPosition, false)
             }
         })
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderFragment.kt
@@ -153,6 +153,12 @@ class ReaderFragment : Fragment(R.layout.reader_fragment_layout) {
             }
         })
 
+        viewModel.closeReaderInterests.observe(viewLifecycleOwner, Observer { event ->
+            event?.getContentIfNotHandled()?.let {
+                ReaderActivityLauncher.closeReaderInterests(this)
+            }
+        })
+
         newsCardViewModel.openUrlEvent.observe(viewLifecycleOwner, Observer {
             it?.getContentIfNotHandled()?.let { url ->
                 val activity: Activity? = activity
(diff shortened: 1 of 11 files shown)


malinajirka:
>I couldn't find a scenario in the app where we navigate from
bottom nav tab fragment => full screen fragment that remains above the bottom nav bar.

I assume you meant "full screen fragment that remains above the bottom **APP** bar", right?

The solution you are proposing LGTM. I think it'd be nice if the app bar animated back from the top, when we show the tabs screen. I have seen that in some other app - not sure which.

I haven't done a full review, since you mentioned it's a WIP. Let me know when it's ready for review. Thanks @ashiagr !

malinajirka:
Btw don't forget that we are merging this code into develop. I haven't seen anything problematic, just mentioning it as I almost forgot about it yesterday:D.

ashiagr:
> I assume you meant "full screen fragment that remains above the bottom APP bar", right?

Yes, `APP` bar - that's the right word.

> The solution you are proposing LGTM.

🙌

> I think it'd be nice if the app bar animated back from the top, when we show the tabs screen. I have seen that in some other app - not sure which.

I'll try to fine-tune necessary animations.

ashiagr:
Noticed an edge case in Pixel Android 10 emulator related to app not going to background on back press. As discussed on slack, added a [note in project board](https://github.com/orgs/wordpress-mobile/projects/63#card-40569141) with steps to reproduce.

Also will add tests in the next task when we fetch user tags so that test names are more meaningful. 

@malinajirka can you take another look?

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderFragment.kt
```diff
@@ -168,4 +193,26 @@ class ReaderFragment : Fragment(R.layout.reader_fragment_layout) {
             }
         }
     }
+
+    private fun showReaderInterests() {
+        val readerInterestsFragment = childFragmentManager.findFragmentByTag(ReaderInterestsFragment.TAG)
+        if (readerInterestsFragment == null) {
+            childFragmentManager.beginTransaction()
+                .replace(
```
malinajirka, line 201: Just out of curiosity (I have never been sure which methods should be used): Any reason you used "replace" instead of "add" and "commitNow" instead of "commit"?

ashiagr: `commitNow` [disallows adding to back stack](https://github.com/aosp-mirror/platform_frameworks_support/blob/2167a9174649d66111f2be5a0b1fc96b15cc48aa/fragment/java/android/support/v4/app/BackStackRecord.java#L671), which is what we want in this case. 

Also it

> [Commits this transaction synchronously.](https://developer.android.com/reference/android/app/FragmentTransaction.html#commitNow()) Any added fragments will be initialized and brought completely to the lifecycle state of their host and any removed fragments will be torn down accordingly before this call returns.

This article [the many flavors of commit ](https://medium.com/@bherbst/the-many-flavors-of-commit-186608a015b1) explains it well
> If you need synchronicity and you are not adding your transaction to the back stack, use commitNow(). 

I have no preference of using `replace` instead of `add`. Found it used in several parts of the existing code and so went ahead with it.

ashiagr:
Thanks for the review! 

> I've noticed the position of the loading indicator on the discover tab seems off. Do you think it could be somehow related to these changes?

I'll check this out in the next PR.

PR #12225, opened 2020-06-18
Prepublishing Nudges : Fixed issue where rotating device would cause new tag to not persist

### LINKED ISSUE #12224: Prepublishing Nudges: tags added to the tag editor after the device is rotated are not saved
Steps to reproduce:

1. Open any post for editing.
1. Add some tags.
1. Tap Publish.
1. Tap Tags.
1. Rotate the device.
1. Observe that previously entered tags remain visible.
1. Add a new tag to the list (after the device was rotated).
1. Tap the left arrow to go back or X to close the tag editor.

**Result:** tags remain in the tag editor after rotation but any tags added to the tag editor after the device is rotated are not saved. Tags added before rotating are saved.

<p style="clear:both;"><sup>Tested with WPAndroid 15.1-rc-1 on Pixel 3 Android 10.</sup></p>

jd-alexander:
Fixes #12224

## Solution

The `EditPostRepository` needs to be set before`isStarted` is called because when rotation takes place the `ViewModel` instance is retained thus it's already started. Nonetheless, the `EditPostRepository` reference needs to be set to the `ViewModel`

## Testing

1. Open any post for editing.
2. Add some tags.
3. Rotate the device.
4. Observe that previously entered tags remain visible.
5. Add a new tag to the list (after the device was rotated).
6. Tap the left arrow to go back or X to close the tag editor.
7. Observe that all the tags that have been added exist. 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingTagsViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingTagsViewModel.kt
@@ -39,10 +39,11 @@ class PrepublishingTagsViewModel @Inject constructor(
     val toolbarTitleUiState: LiveData<UiString> = _toolbarTitleUiState
 
     fun start(editPostRepository: EditPostRepository) {
+        this.editPostRepository = editPostRepository
+
         if (isStarted) return
         isStarted = true
 
-        this.editPostRepository = editPostRepository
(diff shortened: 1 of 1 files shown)


malinajirka:
Thanks for the review! Btw it seems the milestone is set to 15.2, but the target branch is 15.1. Should we update the milestone for consistency?

PR #12247, opened 2020-06-22
Rename get started flow

planarvoid:
Based on the discussion this PR renames the `get_started` flow to `wordpress_com`

PR submission checklist:


planarvoid:
thanks for the check @renanferrari , it makes sense to include this change as well. I wasn't sure if it was agreed upon or not. The changes are implemented in 3627da8

PR #12254, opened 2020-06-23
Change source for wordpress_com login

planarvoid:
This PR adds a different source for the login from the Me fragment. As far as I could see the `WPCOM_LOGIN_ONLY` wasn't previously used but it should've been used for this case anyway. The only difference in functionality (if I see correctly) is slightly different flow when trying to log in with site address. Does that look correct @renanferrari ?

To test:
- Log in with a self-hosted site
- Go to Me fragment
- Click on Log in to WordPress.com
- Check that the flow starts with the new source

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/accounts/UnifiedLoginTracker.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/accounts/UnifiedLoginTracker.kt
@@ -63,6 +63,7 @@ class UnifiedLoginTracker
         DEEPLINK("deeplink"),
         REAUTHENTICATION("reauthentication"),
         SELF_HOSTED("self_hosted"),
+        ADD_WORDPRESS_COM_ACCOUNT("add_wordpress_com_account"),
         DEFAULT("default")
     }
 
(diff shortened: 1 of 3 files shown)


WordPress/src/main/java/org/wordpress/android/ui/accounts/UnifiedLoginTracker.kt
```diff
@@ -63,6 +63,7 @@ class UnifiedLoginTracker
         DEEPLINK("deeplink"),
         REAUTHENTICATION("reauthentication"),
         SELF_HOSTED("self_hosted"),
+        ADD_WORDPRESS_COM_ACCOUNT("add_wordpress_com_account"),
```
planarvoid, line 66: We can decide to rename this if it doesn't make sense

planarvoid:
I'm not sure `onboarding` is the best name @renanferrari . We should investigate if there isn't any other flow that uses the default value (if it's only for onboarding, we should rename also the login step and I'd do it in another PR).

renanferrari:
While reviewing #12116 and #12117 I noticed we somehow missed a bug that should have been handled here. I opened a new issue for it here: #12268.

Issue #12028, opened 2020-05-26
Reader Improvements: New Discover Main Issue


malinajirka:
Parent issue - #12026.

This is a Main Issue for the "New Discover" screen. This screen will show a list of posts based on the topics the user follows. We'll show two types of cards on this screen in this phase of the Reader Improvements project - regular post and recommended topics. We plan to add more types of cards in the future.

**BackLog:**

- [ ] Consider enabling Like action on bookmark posts ([link](https://github.com/wordpress-mobile/WordPress-Android/blob/3c1cf2bb6473069f0a7a728de7689c5f2ee5bbe7/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostUiStateBuilder.kt#L221)) (low priority)
- [ ] Introduce loading shimmer effect (low priority)
- [ ] Share action in the legacy fragment isn't using the `ReaderPostCardActionsHandler` and is not handling ActivityNotFoundException (low priority)

**InProgress:**
- [ ] Placeholder

**Done:**
- [x] Finalise expandable tags view  (@ashiagr ) #12615, #12479
- [x] Refactor like use case so that is can be used in the legacy fragment via the Like Action (low priority) @zwarm 
- [x] Fix "Follow" action (DiscoverDataProvider should automatically update the data and update the state of the "follow" icon) (high priority) (@ashiagr ) #12667
- [x] Implement LoadMore (paging) - against mocked ReaderRepository (@malinajirka)
- [x] Introduce Recommended topics card (@jd-alexander)
- [x] Remove NewsCard (low priority) (@malinajirka)
- [x] Verify "Turn on/off notifs" action works (DiscoverDataProvider should automatically update the data and update the state of the "notifs" icon) (high priority) (@ashiagr )
- [x] Remove this out-dated todo ([link](https://github.com/wordpress-mobile/WordPress-Android/blob/1768f99bab1e18847d2830d1cb9b4fdb27b63e72/WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderPostBookmarkUseCase.kt#L82))

- [x] Fix "Like" action (DiscoverDataProvider should automatically update the data and update the state of the "like" icon) (high priority) (@zwarm)
- [x] Make sure QuickStart works - especially when the user opens reader on "select interest" screen (high priority) (@malinajirka )
- [x] Introduce "Welcome to Reader. Discover millions of blogs" card (medium priority) (@ashiagr ) #12692
[...]

PR #12256, opened 2020-06-23
Issue/12028 discover view holder

malinajirka:
Parent issue #12028 

This PR initialises views on the reader post card based on ReaderPostUiState. 

I plan to replace [this method](https://github.com/wordpress-mobile/WordPress-Android/blob/588b9582c3ad022b91fd1be63111ac306d76c4ef/WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderPostAdapter.java#L442) and start using the adapter there. But we are not there yet. I decided to create this PR to limit the size of this PR. The code isn't being used in production yet. However, you can test that at least the dummy/mocked discover post looks +- okay.

To test:
1. Make sure the reader improvements feature flag is on
2. Clear app data
3. Open reader
4. Select some interests and click on submit
5. Notice the discover post is being rendered correctly

- we'll run thorough tests when we start using this code for all the posts in the Reader. So don't worry about testing to much now.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderPostViewHolder.kt
```diff
@@ -0,0 +1,110 @@
...
+        // Action buttons section
+        initActionButton(uiState.postId, uiState.blogId, uiState.likeAction, count_likes)
+        initActionButton(uiState.postId, uiState.blogId, uiState.reblogAction, reblog)
+        initActionButton(uiState.postId, uiState.blogId, uiState.commentsAction, count_comments)
+        initActionButton(uiState.postId, uiState.blogId, uiState.bookmarkAction, bookmark)
+
+        state.onItemRendered.invoke(uiState.postId, uiState.blogId)
+    }
+
+    private fun initFeaturedImage(state: ReaderPostUiState) {
```
ashiagr, line 59: Should we rename these `init` functions to `update`, we're not initialising views from VH init, just updating them from VH bind, wdyt?

PR #12269, opened 2020-06-24
Issue/12250 change site title from my site screen

### LINKED ISSUE #12250: [My Site] Make site title tappable
To make it easier for users to change the site title, let's make the site title in the site detail header view tappable.

Screenshot_20200622-212110_WordPress

When tapped, if the user is an admin of the site we'll present a popup menu with one option:

* Change site title

On tapping the "Change site title" item:

* If the user is an admin of the site, we'll navigate the user to the Site Settings screen and highlight the Site title row using the same indicator that we use for Quick Start.
* If they're not an admin, we'll show a snackbar stating: "The site title can only be changed by a user with the administrator role."

khaykov:
Fixes #12250

This PR adds the ability to change a site title by tapping on it in the My Site screen.

I repurposed `ProfileInputDialogFragment` which had generic dialog code, suitable for like this.

You can't change site title on:
- Self Hosted sites or jetpack sites not connected to wpcom account.
- If you don't have admin privileges.

If the site title is empty site URL will be displayed instead. This is a preexisting logic.

@osullivanchris Original issue mentioned displaying a popup menu with single "Change Site Title" option, but I'm not sure if we need to add this extra step. What do you think?

Normal flow:
[Image from Gyazo](https://gyazo.com/710a251f910ac8dc9bdbfb3a72a6715d)

When the title is empty:
[Image from Gyazo](https://gyazo.com/8587efc69af7cc1573d6df7a0ba5cbb3)

To test:
- While logged in as admin to wpcom or connected JP site tap on site title on My Site screen.
- You should see the title changer dialog.
- Updated the title and confirm that it is reflected on My Site screen and in Site Settings.
----
- Try to do the same thing without an internet connection or as a non-admin user, and confirm that you can't change the title.

One preexisting setting related issue is if you try to update settings while having a very poor internet connection and leave the app before settings are propagated to endpoint they will not be saved. 

PR submission checklist:


osullivanchris:
Per discussions on slack

- We'll keep @khaykov 's approach of using the input dialog to change site name
- We'll add some hint text to give context
- For a user who does not have permissions: rather than showing a SnackBar error, we'll show them the same dialog. The field will be disabled, and the hint text will give an explanation that they cannot change the name (but they will with this paradigm at least understand what is going on and how it would work with permissions)
- We'll make minor tweaks to get a few extra pixels touch target out of the url, without making a noticeable difference to the layout

WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
```diff
@@ -481,6 +488,43 @@ class MySiteFragment : Fragment(),
         }
     }
 
+    private fun showTitleChangerDialog() {
+        // if site is self hosted we don't do anything on the tap
+        if (!SiteUtils.isAccessedViaWPComRest(selectedSite)) {
+            return
+        }
```
renanferrari, line 495: Is there anything that would prevent us from making this check outside the click listener? That way we could disable the click altogether instead of just not doing anything — and the user wouldn't see the tap feedback.

khaykov: That's a good idea! We can't do it when setting up click listeners since site object might not be there yet, so I'm doing it in onResume, and it works ok :)

khaykov:
Huh, so weird, I thought I posted a comment after changes, but I guess not? Anyway.

@renanferrari I implemented changes discussed with @osullivanchris, and ready for another round :)

[Image from Gyazo](https://gyazo.com/3204ad357bd38199951ce09e54aac2d7)

oguzkocer:

osullivanchris:
Hey @khaykov just realised I never saw this one after AFK. I reviewed the build from #807445 while I was looking at that one, and its working really well, nice job. 

I think the hint text is good. And the error state when I don't have permission to change site was really clear to me. 

Thanks!

PR #12275, opened 2020-06-25
Added a special focus point position for VIEW_SITE task

### LINKED ISSUE #12257: [Quick Start] View Site indicator hidden by FAB
### Expected behavior

When following the View Site Quick Start tour, the blue dot prompting the user to tap the View Site row should always be visible.

### Actual behavior

The View Site indicator is hidden behind the Floating Action Button.

The simplest thing to start with here may be to move the dot closer to the ‘View Site’ label rather than the edge of the row.

Alternatively, we added extra padding at the very bottom of the view on iOS so that the user can scroll to the end of a list and see ‘below’ the FAB. Could be a route to consider. 

* Create a site and follow the View Site Quick Start tour

khaykov:
Fixes #12257

This PR moves a position of a focus point for the VIEW_SITE task so it would not be covered by FAB.
Screenshot_1593054088

To test:
- Create new site and start Quick Start.
- Navigate to `Customize Your Site` tasks, and select `View Your Site`.
- Find a focus point on new View Site row on MySite screen, and make sure it's not obstructed by FAB.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
@@ -1213,16 +1213,24 @@ class MySiteFragment : Fragment(),
         val focusPointSize = resources.getDimensionPixelOffset(R.dimen.quick_start_focus_point_size)
         val horizontalOffset: Int
         val verticalOffset: Int
-        if (isTargetingBottomNavBar(activeTutorialPrompt!!.task)) {
-            horizontalOffset = quickStartTarget.width / 2 - focusPointSize + resources
-                    .getDimensionPixelOffset(R.dimen.quick_start_focus_point_bottom_nav_offset)
-            verticalOffset = 0
-        } else if (activeTutorialPrompt!!.task == UPLOAD_SITE_ICON) {
-            horizontalOffset = focusPointSize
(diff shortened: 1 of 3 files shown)


osullivanchris:
@khaykov hey, in #12257 @frosty also captured

> Alternatively, we added extra padding at the very bottom of the view on iOS so that the user can scroll to the end of a list and see ‘below’ the FAB. Could be a route to consider.

also discussed on p2-pbArwn-Gt

Would just like to know your opinion there. When we display FABs, it does seems strange that the last item in a list will always be obscured by the FAB above it on the Z axis. Is there a general way that Android solves this? My suggestion here had been to add extra blank space to the bottom of the list so that you could scroll _beyond_ the list and access the entire last row if necessary. But it doesn't seem a very elegant solution.

padding

renanferrari:
>My suggestion here had been to add extra blank space to the bottom of the list so that you could scroll beyond the list and access the entire last row if necessary. But it doesn't seem a very elegant solution.

@osullivanchris AFAIK that's actually considered a good practice on list screens where the FAB doesn't hide on scroll.

khaykov:
@renanferrari I fixed the lint issue and also added bottom padding to the screen :)

Screenshot_1593128835

osullivanchris:
> I fixed the lint issue and also added bottom padding to the screen :)

Sweet, looks good!

PR #12281, opened 2020-06-25
Check for remote field in features

planarvoid:
It was previously possible to use the annotation for a remote feature but to miss to add the `remoteField` field parameter. This PR adds a generated class that checks whether all the features that are annotated with the `Feature` annotation have non-null remote field. 

To test:
- Create a feature annotated with `Feature` without the `remoteField`
- Run the app
- Notice that it crashes
- Add the `remoteField` 
- Notice that it doesn't crash any more

PR submission checklist:


malinajirka:
It's ready to be merged when 15.2 is freezed.

PR #12295, opened 2020-06-26
Issue/12028 use new viewholder in reader adapter

malinajirka:
Parent issue #12028 
Not Ready for merge - I want to make sure that we merge this into 15.3 so we can have two more weeks in develop + two more weeks in code freeze in case we encounter some bugs.

Merge instructions
1. Review this PR
2. Make sure https://github.com/wordpress-mobile/WordPress-Android/pull/12283/files#diff-347e23014dc81f0d20ce42d4792e0015 is merged
3. Make sure 15.2 has already been freezed
4. Update target branch to develop
5. Merge this PR

This PR starts using new ReaderPostUiStateBuilder and new ReaderPostViewHolder to render regular posts in Reader (on all screens - following, discover, liked, tag detail, blog detail etc).

I've removed the previous ReaderPostViewHolder which contained both the business logic and view layer logic. Now business logic is located in ReaderPostUiStateBuilder. ReaderPostViewHolder is a dump viewHolder which just updates the views based on the uiState (with exception of video thumbnail logic, which is still located there as the refactoring isn't as simple).

This is the last big PR in this series. I'll just fix bugs we will find and add bunch of unit tests for the the builder. 

Again,  I think this might be easier to review by commits. Let me know if I can help in any way with the review - we could for example jump on a call and do the review in pair.

To test:
Ideally compare current develop with this branch side by side - test all the actions, different types of posts (with video, featured image, just text, blog detail, tag detail, ... anything else you can think of).

Known issues:
- Like button animation is missing - I'll keep the TODO and implement the animation if we have time.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderPostAdapter.java
```diff
@@ -445,168 +328,99 @@ private void renderPost(final int position, final ReaderPostViewHolder holder) {
...
-                holder.mTxtText.setVisibility(View.GONE);
+        Function3<Long, Long, Boolean, Unit> onBookmarkClicked = (postId, blogId, aBoolean) -> {
+            toggleBookmark(post.blogId, post.postId);
+            notifyItemChanged(position);
+            return Unit.INSTANCE;
+        };
+        Function3<Long, Long, Boolean, Unit> onLikeClicked = (postId, blogId, aBoolean) -> {
+            toggleLike(ctx, post, position, holder);
+            return Unit.INSTANCE;
+        };
```
ashiagr, line 341: Not sure what's the recommended practice to create click listener objects - is it fine to create new click listeners objects inside `onBindViewHolder`?  

I've seen them to be created inside VH init at places in the existing code for e.g. [here](https://github.com/wordpress-mobile/WordPress-Android/blob/2974e0111e0704d0feedece7a68472f74f429216/WordPress/src/main/java/org/wordpress/android/ui/sitecreation/domains/SiteCreationDomainViewHolder.kt#L42).

Considering they're light objects, I think we might want to keep them here for better code readability. Lmkwyt.

malinajirka: Great question! In general, it's better to create them in the init. However, I need access to the post, vh and context in these legacy onclick listeners. So I decided to keep them within the method - in case we notice any performance impact, which I doubt, we'll need to do a bit more refactoring.

malinajirka:
>Also found some unused code in the ReaderPostAdapter. I'm not reporting them here as I believe they will be removed in future PRs as refactoring is still in progress.

Good point, I'll make sure to remove all the unused code in the upcoming PRs.

>Just noticed one crash:

Good catch! Fixed in 00667c5ed78bd3a81582f4316cf0aec0b5ba283f

PR #12306, opened 2020-06-28
WordPress Stories : Story title bottom sheet integration part 1

### LINKED ISSUE #12305: WordPress Stories : Implement bottom sheet for story title UI
story

- [x] Add the Story Title UI to the bottom sheet
- [x] Integration bottom sheet with Story Composer

jd-alexander:
Fixes #12305
## Solution
To integrate the existing prepublishing bottom sheet with the Story Composer. Currently, this solution hardcodes the story title UI within the bottom sheet for pages and post just so this PR can be tested. In future PRs, the entry point will be from the Story Composer. 

## Testing
1. Create a new post. 
2. Click publish. 
3. See that the bottom sheet UI asks for a story title and you are able to enter one. 

Before | After 
--------|-------
 <kbd></kbd>       |       <kbd></kbd>

## Concerns
1. Currently, even though I am requesting the story title's edit text focus and showing the keyboard it still doesn't show up initially. 
2. The story title UI might need some tweaking but keeping it as is for now, and will improve on it in future PRs. 

## Reviewing 

## Submitter Checklist


WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeAdapter.kt
```diff
@@ -18,6 +20,8 @@ import javax.inject.Inject
 private const val headerViewType: Int = 1
 private const val homeItemViewType: Int = 2
 private const val submitButtonViewType: Int = 3
+private const val storyTitleViewType: Int = 4
```
mzorz, line 23: Nitpick: I wonder why we don't use enums and use them as Ints so it's easier to derive a `viewType` for the viewHolder (and the code looks cleaner and there wouldn't be ambiguity / overlap chance). Feel free to leave as is if that's the way things are being handled for other viewtypes in other places though, but thought I'd comment on my thoughts.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
```diff
@@ -91,7 +96,8 @@ class PrepublishingHomeViewModel @Inject constructor(
                                 ?: run { UiStringRes(R.string.prepublishing_nudges_home_tags_not_set) },
                         actionClickable = true,
                         onActionClicked = ::onActionClicked
-                ))
+                )
```
mzorz, line 99: small nitpick this should be moved to the right

mzorz: I mean - indented

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeAdapter.kt
jd-alexander: Thanks! This is a good idea that's worth exploring! I didn't change this due to it being the approach being used so I will keep this in mind for refactoring 🙏

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
jd-alexander: Thanks! @mzorz is there any way that the checkstyle tooling can be improved to catch things like these? It's sometimes hard to spot them easily.

mzorz: I'm sure there is, but it's a nitpick anyway. Given we have other priorities right now, we can leave as is 👍

PR #12307, opened 2020-06-28
WordPress Stories : Story publish bottom sheet integration part 2

### LINKED ISSUE #12305: WordPress Stories : Implement bottom sheet for story title UI

jd-alexander:
Fixes #12305

## Solution
To perform this integration with the Story Composer, these are the main things that were done. 

- The bottom sheet was supplied with a boolean param that determines if the `Story Title UI` should be shown. 
- The `StoryComposerActivity ` implemented the `EditPostActivityHook` that's necessary for the functionality of the bottom sheet to update the post with the changes that are taking place on the visibility, publish date and tags. 
- `PrepublishingEventProvider` a component that allows the host app to respond to the next button event, that allows the bottom sheet to be shown before the publish operation is completed. 
- `EditPostRespository` now responds to all changes made in the bottom sheet and persists those changes to the database so if there is a configuration change the user doesn't lose the changes they made. 
- Updating the story title causes the current title to be updated. 

## Testing
1. Create a new story post. 
2. Click Publish. 
3. See the bottom sheet like below. 
4. Add a story title.

Before | After 
--------|-------
<kbd></kbd>   |     <kbd></kbd>
  
## Concerns
1. Currently, the bottom sheet isn't above the keyboard due to the Story Composer being a full-screen activity. A custom solution will be created to solve this issue.
<kbd></kbd>

2. The button in the bottom sheet says "Update Now" instead of "Publish Now". This is because the `PostStatus` is `Published`. Based on current design, once the post is new it should say "Draft" so that the current button is "Publish Now". The current post publishing flow in the `EditPostActivity` sets the `PostStatus` to "Published" when necessary before the post is uploaded, so this will be investigated to ensure that it's done appropriately either by that flow or by our own logic, where once the post is set to be uploaded we can then mark it as being "Published."

<kbd></kbd>

3. The bottom sheet has other operations i.e it allows the user to change the status to others that wouldn't publish the post now but schedule it, mark as review, etc. We will need to investigate this and ensure that the user is able to save a draft story post, or schedule one. 

Visibility Options | Submit Button Changed 
--------|-------

4. This PR didn't implement any dismissal logic that would hide buttons etc. so that will be done in another PR. 

5. The flow for setting the current title to the StoryRepository needs verification and testing. Also should the Story Title be set as the post title? 

## Reviewing 

## Submitter Checklist


WordPress/src/main/java/org/wordpress/android/ui/posts/PostsListActivity.kt
```diff
@@ -335,7 +335,11 @@ class PostsListActivity : LocaleAwareActivity(),
             event.applyIfNotHandled {
                 val fragment = supportFragmentManager.findFragmentByTag(PrepublishingBottomSheetFragment.TAG)
                 if (fragment == null) {
-                    val prepublishingFragment = newInstance(site, editPostRepository.isPage)
+                    val prepublishingFragment = newInstance(
+                            site = site,
+                            isPage = editPostRepository.isPage,
+                            isStoryPost = false
+                    )
```
mzorz, line 342: can we consist these 2 flags into an enum that holds `Post, Page, Story post`  type of values? So we wouldn't be passing around 2 variables when we could pass one, and moreover we would make sure to not allow any ambiguities or impossible states such as `isPage` and `isStoryPost` being both true (which cannot be)

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
```diff
@@ -109,6 +127,19 @@ class PrepublishingHomeViewModel @Inject constructor(
         _uiState.postValue(prepublishingHomeUiStateList)
     }
 
+    private fun onStoryTitleChanged(storyTitle: String) {
+        updateStoryTitleJob?.cancel()
+        updateStoryTitleJob = launch(bgDispatcher) {
+            delay(THROTTLE_DELAY)
```
mzorz, line 133: could you explain why this is needed? Are we using this delay elsewhere? thanks

mzorz:
Another thing: we should add Tracks here as well. Please refer to https://github.com/wordpress-mobile/WordPress-Android/pull/12074 to see how / which events we're adding. I assume you'd have similar events in `EditPostActvity` and related classes for publishing nudges / publishing itself. Feel free to plan for adding these changes (re: tracking)  on an isolated PR @jd-alexander 👍

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
jd-alexander: Oh, so this was used in the `PrepublishingTagsViewModel`( [link to file](https://github.com/wordpress-mobile/WordPress-Android/blob/db1d3611f9b745fc0532f3990daf7be6bf153a5a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingTagsViewModel.kt)) so that there's a delay since every single character change received would cause a save operation so I just used the same approach here. Let me know what you think.

jd-alexander:
> Awesome to see this being put together @jd-alexander ! A few comments:

Awesome! Thanks! 

> Regarding this and the UPDATE NOW labelling on the bottom sheet, it is important that we know one intrinsic feature of Stories is that these will be published by default. I guess we can handle this as a draft internally, as long as we always set it to PUBLISH when the bottom sheet is ready for users to give their consent.

Yes! I will check out the logic and yes, once the user presses publish now then I would change the status to `Published`.

WordPress/src/main/java/org/wordpress/android/ui/posts/PostsListActivity.kt
jd-alexander: This is a really important improvement. I will track this and get it done when I get some free time 🙏 https://github.com/wordpress-mobile/WordPress-Android/issues/12316

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewModel.kt
mzorz: That explains it very well, could we add a comment to code explaining this? 🙏

jd-alexander: Resolved here 6427330

jd-alexander:
Hey @mzorz I tracked all the issues here and resolved a comment about adding a comment about the delay 😄 so this should be ready again.

PR #12320, opened 2020-06-30
Show popup more menu part 1

malinajirka:
Parent issue #12028 

This PR 
1. Introduces `ReaderPostMoreButtonUiStateBuilder` - the logic is copied from [here](https://github.com/wordpress-mobile/WordPress-Android/blob/4c66010ebc5f966b117e304cf7c533c233329cee/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java#L2598) and the view attributes are copied from [here](https://github.com/wordpress-mobile/WordPress-Android/blob/0a6af83a5fd234207c944d5e539f0a0348d8d199/WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderMenuAdapter.java#L72).
2. Instead of passing a listener for each button on the card item, we now pass a single onButtonClicked listener with ReaderPostCardActionType which is used to determine which button was clicked.

To test:
I'm targeting a working branch since I'll keep improving/refactoring the same part of the code in the upcoming PRs. In order to increase efficiency I think it'd be best to test it in the last PR of this series so we don't keep testing the same behavior repeatedly. 

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderCardUiState.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderCardUiState.kt
@@ -3,6 +3,9 @@ package org.wordpress.android.ui.reader.discover
 import android.text.Spanned
 import android.view.View
 import androidx.annotation.AttrRes
+import androidx.annotation.DrawableRes
+import org.wordpress.android.ui.reader.discover.ReaderPostCardAction.PrimaryAction
+import org.wordpress.android.ui.reader.discover.ReaderPostCardAction.SecondaryAction
 import org.wordpress.android.ui.reader.models.ReaderImageList
 import org.wordpress.android.ui.utils.UiDimen
 import org.wordpress.android.ui.utils.UiString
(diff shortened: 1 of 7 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderCardUiState.kt
```diff
@@ -56,14 +60,42 @@ sealed class ReaderCardUiState {
...
+        @AttrRes val labelColor: Int,
+        @DrawableRes val iconRes: Int,
+        @AttrRes val iconColor: Int = labelColor,
+        override val isSelected: Boolean = false,
+        override val type: ReaderPostCardActionType,
+        override val onClicked: (Long, Long, ReaderPostCardActionType) -> Unit
+    ) : ReaderPostCardAction()
+}
+
+enum class ReaderPostCardActionType {
```
malinajirka, line 91: I'm not sure about this whole `ReaderPostCardAction` structure. I definitely plan to improve it a bit more. For example `ReaderPostCardActionType` contains both Primary buttons and the buttons hidden under the "more/..." menu, which might be a bit confusing. Any suggestions are welcome :).

ashiagr: Agree, it looks bit odd to have `ReaderPostCardActionType` with both primary and secondary actions without any distinction. Can't it be converted to a nested sealed class and accessed like `ReaderPostCardActionType.Primary.Bookmark`?

malinajirka: 👍 Good idea, I'll do that in a new PR. ;)

PR #12322, opened 2020-06-30
Implement unsupported fallback tracks

marecar3:
Implements changes from: https://github.com/WordPress/gutenberg/pull/23450
`gutenberg-mobile` ref: https://github.com/wordpress-mobile/gutenberg-mobile/pull/2449

We are adding:
- `gutenberg_unsupported_block_webview_shown`
  - params: `"block": "{block_internal_name}"`
- `gutenberg_unsupported_block_webview_closed` 
  - params: `"action": "save" / "dismiss"` accordingly.

cc @etoledom @guarani 

To test:

- Run Metro server on https://github.com/wordpress-mobile/gutenberg-mobile/pull/2449  to activate the webview option on unsupported blocks ensuring that the required JS code is running.
- Open the unsupported block web view
  - Check that `gutenberg_unsupported_block_webview_shown` is triggered.
- Close the web view by dismissing and by saving
  - Check that `gutenberg_unsupported_block_webview_closed` is triggered with the corresponding property.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/posts/editor/EditorTracker.kt
```diff
@@ -104,8 +109,17 @@ class EditorTracker @Inject constructor(
...
+                Stat.EDITOR_GUTENBERG_UNSUPPORTED_BLOCK_WEBVIEW_SHOWN
         }
 
-        analyticsTrackerWrapper.track(currentStat, mapOf("editor" to editorName))
+        if (properties.isEmpty()) {
+            analyticsTrackerWrapper.track(currentStat, mapOf("editor" to editorName))
+        } else {
+            properties.put("editor", editorName)
+            analyticsTrackerWrapper.track(currentStat, properties)
+        }
```
mchowning, line 123: Not a big deal, but instead of having this if-else block, could we just do the following regardless of whether the property map is empty?

```
properties["editor"] = editorName // or properties.put("editor", editorName)
analyticsTrackerWrapper.track(currentStat, properties)
```

We could take this even further and avoid mutating the map by accepting a non-mutable map as the function parameter (instead of the mutable map the function currently requires) and creating a new map with the proper "editor" value:

```
val updatedProperties = properties + ("editor" to editorName)
analyticsTrackerWrapper.track(currentState, updatedProperties)
```

marecar3: You are Kotlin ninja, text for the tips! 👍

marecar3: Fixed.

PR #12323, opened 2020-06-30
Show popup more menu part 3 - fix button clicks

malinajirka:
Parent issue #12028 

Merge instructions
1. Review this PR
2. Make sure https://github.com/wordpress-mobile/WordPress-Android/pull/12321
3. Update the target branch to develop
4. Remove "Not ready for merge" label
5. Merge this PR

This is a last PR in this mini-series. This PR fixes clicks on items in the popup/more/... menu on post card items. I've also removed some unused code and added missing @nonNull annotation to the ReaderPostListFragment.

To test:
Bookmark
1. Click on "Bookmark" icon on one of the items in the Following tab
2. Notice the icon gets filled and a snackbar message shown
3. Click on "ViewAll" action in the snack bar -> notice the post you bookmarked is in the "Saved" list
4. Switch back to the Following tab
5. Click on "Unbookmark" on the post
6. Notice the icon changes back from filled state to outlined state.

Reblog
1. Click on "reblog" icon on one of the items in the following tab
2. Notice site picker is shown

Comment
1. Click on "comment" icon on one of the items in the following tab
2. Notice post comments are shown

Like
1. Click on "Like" icon on one of the items in the following tab
2. The icon gets filled and the count number gets increased
3. Click on the icon again and notice the icon changes back from filled state to outline state and the count decreases

More - Following
1. Click on the More button on one of the items in the following tab
2. Notice the "Following" action is green
3. Click on "Following"
4. Click on the More button again
5. Notice the action is now blue and says "Follow"
6. Click on "Follow"
7. Click on the More button again
8. Notice the "Following" action is green again

More - Turn on site Notifications
1. Click on the More button on one of the items in the following tab
2. Notice "Turn on site notifications" is grey
3. Click on it
4. Click on the More button again
5. Notice the action now says "Turn off site notifications" and is green
6. Click on it
7. Click on the More button again
8. Notice the action now says "Turn on site notifications" and is grey again

More - Share
1. Click on the More button on one of the items in the following tab
2. Click on "Share"
3. Notice OS default share picker is shown

More - Visi Site
1. Click on the More button on one of the items in the following tab
2. Click on "Visit site"
3. Notice a browser with the post is shown

More - Block this site
1. Click on the More button on one of the items in the following tab
2. Click on "Block this site"
3. Notice the site is removed from the feed and a snackbar notification is shown 

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
@@ -98,6 +98,7 @@ import org.wordpress.android.ui.reader.adapters.ReaderSearchSuggestionAdapter;
 import org.wordpress.android.ui.reader.adapters.ReaderSearchSuggestionRecyclerAdapter;
 import org.wordpress.android.ui.reader.adapters.ReaderSiteSearchAdapter;
 import org.wordpress.android.ui.reader.adapters.ReaderSiteSearchAdapter.SiteSearchAdapterListener;
+import org.wordpress.android.ui.reader.discover.ReaderPostCardActionType;
 import org.wordpress.android.ui.reader.reblog.NoSite;
 import org.wordpress.android.ui.reader.reblog.PostEditor;
 import org.wordpress.android.ui.reader.reblog.SitePicker;
@@ -148,8 +149,8 @@ import static org.wordpress.android.fluxc.generated.AccountActionBuilder.newUpda
 
 public class ReaderPostListFragment extends Fragment
         implements ReaderInterfaces.OnPostSelectedListener,
-        ReaderInterfaces.OnPostPopupListener,
         ReaderInterfaces.OnFollowListener,
+        ReaderInterfaces.OnPostListItemButtonListener,
         WPMainActivity.OnActivityBackPressedListener,
         WPMainActivity.OnScrollToTopListener,
         ReblogActionListener {
@@ -711,7 +712,7 @@ public class ReaderPostListFragment extends Fragment
 
     @Override
     @SuppressWarnings("unchecked")
-    public void onAttach(Context context) {
+    public void onAttach(@NonNull Context context) {
         super.onAttach(context);
 
         // detect the bottom nav controller when this fragment is hosted in the main activity - this is used to
@@ -831,7 +832,7 @@ public class ReaderPostListFragment extends Fragment
     }
 
     @Override
-    public void onSaveInstanceState(Bundle outState) {
+    public void onSaveInstanceState(@NonNull Bundle outState) {
         AppLog.d(T.READER, "reader post list > saving instance state");
 
         if (mCurrentTag != null) {
@@ -1395,11 +1396,6 @@ public class ReaderPostListFragment extends Fragment
         populateSearchSuggestionRecyclerAdapter(null); // always passing null as there's no need to filter
     }
 
-    private void resetSearchSuggestions() {
-        resetSearchSuggestionAdapter();
-        resetSearchSuggestionRecyclerAdapter();
-    }
-
     /*
      * create and assign the suggestion adapter for the search view
      */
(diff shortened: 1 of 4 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -2580,88 +2572,45 @@ private void trackTagLoaded(ReaderTag tag) {
...
-//                        AnalyticsTracker.track(Stat.READER_ARTICLE_VISITED);
-//                        ReaderActivityLauncher.openPost(view.getContext(), post);
-//                        break;
-//                }
-//            }
-//        });
-//        listPopup.show();
+    public void onButtonClicked(ReaderPost post, ReaderPostCardActionType actionType) {
+        switch (actionType) {
+            case FOLLOW:
```
malinajirka, line 2578: These actions are mostly copied from the commented out code (OnItemClickListener) above.

ashiagr:
Nice work @malinajirka ! I haven't tested it completely. Sharing observations so far:

- [x] Bookmark

While I toggled bookmark action, noticed discrepancies in the image title visibility. It might not be related to bookmark action but definitely something we should look into:

title_on_image

After I click bookmark and then "View All" in the snackbar, saved tab is shown by restarting the activity. If RI FF flag is turned on, it leads to interest picker display. Can we just switch to saved tab instead of restarting the activity? 

bookmark_activity_recreated

- [x] Reblog
- [x] Comment
- [x] Like
- [ ] More - Following
- [ ] More - Turn on site Notifications
- [x] More - Share
- [x] More - Visit Site
- [x] More - Block this site

malinajirka:
>While I toggled bookmark action, noticed discrepancies in the image title visibility. It might not be related to bookmark action but definitely something we should look into:

Nice find. I must admit I have no idea why it's happening. The photo title seems to be unrelated to the post - it's a recycling issue. However, I've used a debugger and verified 5 times, that the `view.visibility = if (visible) View.VISIBLE else View.GONE` is evaluated as `false`, but the view is visible anyway. I've fixed it in https://github.com/wordpress-mobile/WordPress-Android/pull/12323/commits/ad27f1362b35d7cc8f087c72113ddffec1d3577c, but if you have any idea why it's happening let me know.
I've also suppressed the `notifyItemChanged` animation when the "bookmark" button is pressed in https://github.com/wordpress-mobile/WordPress-Android/pull/12323/commits/4669f0e42076855074af0c3ef59753470c4c3ff7. (I don't like this manual re-render approach but it's used in this adapter, so I kept it for consistency. Better solution would be to correctly implement diff util and suppress the animation there.)


>After I click bookmark and then "View All" in the snackbar, saved tab is shown by restarting the activity. If RI FF flag is turned on, it leads to rechecking of followed tags etc. Can't we just switch to saved tab instead of restarting the activity? I understand app will check for followed tags from local database once the implementation is final, so we might not see any visible loading screen. LMKWYT.

We could add `FLAG_ACTIVITY_SINGLE_TOP` flag to the intent and handle the navigation in `onNewIntent`. I've created a note on the project board about this issue. I'd suggest we get back to it if it's causing any troubles when we have the final solution. Wdyt?

ashiagr:
Thanks for taking a look! 

**image title visibility**
>  if you have any idea why it's happening let me know.

Found an [Android bug](https://issuetracker.google.com/issues/73186245) for constraint layout `Group` component (used as `frame_photo`). I think we can eliminate this `Group` element - it was added as a replacement for the wrapper layout while transitioning to flat UI hierarchy. As the issue exists in develop as well, let's do it in a separate GH issue. You can assign it to me.

**saved tab is shown by restarting the activity**
> We could add FLAG_ACTIVITY_SINGLE_TOP flag to the intent and handle the navigation in onNewIntent. I've created a note on the project board about this issue. I'd suggest we get back to it if it's causing any troubles when we have the final solution. Wdyt?

Agree 👍. Let's get back to it later.

PR #12330, opened 2020-06-30
Stories/bottom sheet publish flow

jd-alexander:
Fixes #12330 

##  Solution
The initial `PostStatus` is set to `DRAFT`. Once the user has pressed the submit button we then prepare the post for publishing using shared logic from the publish flow of other parts of the app. 

## Testing
1. Create a new post. 
2. Click Publish. 
3. Release that the button says `Publish Now` instead of `Update Now`

Before | After 
--------|-------
 <kbd></kbd>  | <kbd></kbd> 

## Reviewing 
Only 1 reviewer is needed but anyone can review. 

## Submitter Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.
- [x] I have considered adding accessibility improvements for my changes.
- [x] If it's feasible, I have added unit tests.

PR #12333, opened 2020-06-30
Issue 12330/bottom sheet publish flow

jd-alexander:
Fixes #12330 

##  Solution
The initial `PostStatus` is set to `DRAFT`. Once the user has pressed the submit button we then prepare the post for publishing using shared logic from the publish flow of other parts of the app. 

## Testing
1. Create a new post. 
2. Click Publish. 
3. Release that the button says `Publish Now` instead of `Update Now`

Before | After 
--------|-------
 <kbd></kbd>  | <kbd></kbd> 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
@@ -44,7 +44,6 @@ import org.wordpress.android.ui.posts.EditPostRepository
 import org.wordpress.android.ui.posts.EditPostSettingsFragment.EditPostActivityHook
 import org.wordpress.android.ui.posts.PostEditorAnalyticsSession
 import org.wordpress.android.ui.posts.PostEditorAnalyticsSession.Outcome.CANCEL
-import org.wordpress.android.ui.posts.PostUtilsWrapper
 import org.wordpress.android.ui.posts.PrepublishingBottomSheetFragment
 import org.wordpress.android.ui.posts.ProgressDialogHelper
 import org.wordpress.android.ui.posts.ProgressDialogUiState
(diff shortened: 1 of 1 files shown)


WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
```diff
@@ -417,6 +416,16 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
...
+    private fun prepareAndProcessStoryPostForPublishing() {
+        editPostRepository.update { postModel ->
+            postUtilsWrapper.preparePostForPublish(
+                    postModel,
+                    requireNotNull(site)
+            )
+        }
         processStorySaving()
     }
 }
```
jd-alexander, line 431: Upon doing some more investigation I am realizing that this logic is not required because we do setup the post for publishing. @mzorz in the `StoryMediaSaveUploadBridge` you do this here 

https://github.com/wordpress-mobile/WordPress-Android/blob/bbc3c278385c049ebb6b49216fc2d40574d6143c/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryMediaSaveUploadBridge.kt#L98

So yes, the `Published` `PostStatus` gets added there and that's actually the right place to do it, so I am thankful that I spotted it :)

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
```diff
@@ -289,14 +288,12 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
     private fun saveInitialPost() {
         editPostRepository.set {
             val post: PostModel = postStore.instantiatePostModel(site, false, null, null)
-            post.setStatus(DRAFT.toString())
+            post.setStatus(PostStatus.DRAFT.toString())
```
mzorz, line 291: This was 🤯 - thanks for spotting that bad import! 🍻 🙇

PR #12357, opened 2020-07-07
Refactor reblog action on ReaderPostListFragment and ReaderDiscoverFragment

malinajirka:
Parent issue #12028 

Merge instructions:
1. Review this PR
2. Review https://github.com/wordpress-mobile/WordPress-Android/pull/12358
3. Make sure https://github.com/wordpress-mobile/WordPress-Android/pull/12335 is merged
4. Update target branch to `develop`
3. Merge this PR
4. Merge https://github.com/wordpress-mobile/WordPress-Android/pull/12358 PR

This issue implements "reblog" action in the new ReaderDiscoverFragment. In order to re-use most of the existing code for the action used in ReaderPostListFragment, the feature was refactored.

Snackbar messages are handled in https://github.com/wordpress-mobile/WordPress-Android/pull/12358.

To test:

**No Sites**
1. Disable RI feature flag if you enabled it before
2. Create a new WPCom account (without sites)
3. Open Reader
4. Click on "Reblog" action
5. Notice "No sites" screen is shown

**Single Site**

1. Create/Attach a site on/to your account
2. Open Reader
3. Click on "Reblog" action
4. Notice an editor with the correct content is shown

**Multiple Sites**
1. Create/Attach another site on/to your account
2. Open Reader
3. Click on "Reblog" action
4. Notice a site picker is shown -> pick a site
5. Notice an editor with the correct content is shown

PR submission checklist:


WordPress/src/test/java/org/wordpress/android/ui/reader/reblog/ReblogUseCaseTest.kt
```diff
@@ -0,0 +1,121 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.Mock
+import org.mockito.junit.MockitoJUnitRunner
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.store.SiteStore
+import org.wordpress.android.models.ReaderPost
+
+@RunWith(MockitoJUnitRunner::class)
+class ReblogUseCaseTest {
```
malinajirka, line 17: Tests were copied from ReaderPostListViewModelTest.kt

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -430,9 +430,32 @@ public void onCreate(Bundle savedInstanceState) {
             initSubFilterViewModel();
         }
 
-        handleReblogStateChanges();
+        mViewModel.getNavigationEvents().observe(getViewLifecycleOwner(),
```
malinajirka, line 433: This method is almost identical to `handleReblogStateChanges` which is removed a few lines below.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
```diff
@@ -117,6 +130,20 @@ class ReaderDiscoverViewModel @Inject constructor(
...
 
+    fun onReblogSiteSelected(siteLocalId: Int) {
+        // TODO malinjir almost identical to ReaderPostCardActionsHandler.handleReblogClicked.
+        //  Consider refactoring when ReaderPostCardActionType is transformed into a sealed class.
+        val state = reblogUseCase.onReblogSiteSelected(siteLocalId, pendingReblogPost)
+        val navigationTarget = reblogUseCase.convertReblogStateToNavigationEvent(state)
+        if (navigationTarget != null) {
+            _navigationEvents.postValue(Event(navigationTarget))
+        } else {
+            // TODO malinjir show toast R.string.reader_reblog_error
```
malinajirka, line 141: This is handled in https://github.com/wordpress-mobile/WordPress-Android/pull/12358

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
```diff
@@ -78,8 +80,15 @@ class ReaderPostCardActionsHandler @Inject constructor(
...
 
-    private fun handleReblogClicked(postId: Long, blogId: Long) {
-        AppLog.d(AppLog.T.READER, "Reblog not implemented")
+    private fun handleReblogClicked(post: ReaderPost) {
+        val state = reblogUseCase.onReblogButtonClicked(post)
+        val navigationTarget = reblogUseCase.convertReblogStateToNavigationEvent(state)
+        if (navigationTarget != null) {
+            _navigationEvents.postValue(Event(navigationTarget))
+        } else {
+            // TODO malinjir show toast R.string.reader_reblog_error
```
malinajirka, line 89: Handled in https://github.com/wordpress-mobile/WordPress-Android/pull/12358

WordPress/src/main/java/org/wordpress/android/ui/reader/reblog/ReblogUseCase.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.wordpress.android.ui.reader.discover.ReaderNavigationEvents.ShowNoSitesToReblog
+import org.wordpress.android.ui.reader.discover.ReaderNavigationEvents.ShowSitePickerForResult
+import org.wordpress.android.util.AppLog
+import org.wordpress.android.util.AppLog.T
+import org.wordpress.android.util.BuildConfig
+import javax.inject.Inject
+
+@Reusable
+class ReblogUseCase @Inject constructor(private val siteStore: SiteStore) {
+    fun onReblogButtonClicked(post: ReaderPost): ReblogState {
```
malinajirka, line 20: Content of these methods was moved from ReaderPostListViewModel.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -430,9 +430,32 @@ public void onCreate(Bundle savedInstanceState) {
...
+                    } else if (navTarget instanceof ShowNoSitesToReblog) {
+                        ReaderActivityLauncher.showNoSiteToReblog(getActivity());
+                    } else {
+                        throw new IllegalStateException("Action not supported in ReaderPostListFragment");
+                    }
+                    return Unit.INSTANCE;
+                }));
 
-        mViewModel.start(mReaderViewModel);
+                mViewModel.start(mReaderViewModel);
```
ashiagr, line 458:
```suggestion
     mViewModel.start(mReaderViewModel);
```

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverFragment.kt
```diff
@@ -58,9 +66,22 @@ class ReaderDiscoverFragment : Fragment(R.layout.reader_discover_fragment_layout
                     is SharePost -> ReaderActivityLauncher.sharePost(context, post)
                     is OpenPost -> ReaderActivityLauncher.openPost(context, post)
                     is ShowReaderComments -> ReaderActivityLauncher.showReaderComments(context, blogId, postId)
+                    ShowNoSitesToReblog -> ReaderActivityLauncher.showNoSiteToReblog(activity)
```
ashiagr, line 69:
```suggestion
                    is ShowNoSitesToReblog -> ReaderActivityLauncher.showNoSiteToReblog(activity)
```

malinajirka: I'll need to think about this. The `is` is "missing" here, because `ShowNoSitesToReblog` is an object. On one hand I think it's nice to see it's an object even in the `when`. On the other hand I agree it looks a bit weird that the `is` isn't there.

develric:
Hi all, as agreed with @malinajirka tested this PR since participated a bit to the reblog dev. Didn't go into a full code review since @ashiagr already did it but scanned the code and LGTM (also like the distinction between NavigationEvent and ReblogState 😊 ).

The reblog functionality worked well for me in the 3 use cases, no deviation found from expected 👍 ! I'm only sharing couple of things I thought of/found even if I don't think they are directly related to this PR but maybe better to share:

1. I suspect the states of the action buttons at the bottom of the post cards are still temporary in this PR, but just to be sure we grasped it, i wonder if we are going to manage the not available functionality as dimmed or removing the button. Here below for example on the left the previous approach (removing the commment) and on the right the current (maybe temporary) one (dimming comment)

image

Also if we go into the post details view it seems we are currently keeping the "remove what's not available" approach.

image

2. I think we already addressed it in the past but at least for me it seems when having only self-hosted site (it worked well with wp.com users) when I tap on SAVED > GO TO FOLLOWING

image

I get the following error

```
java.lang.IllegalArgumentException: Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkParameterIsNotNull, parameter tag
        at org.wordpress.android.ui.reader.viewmodels.ReaderPostListViewModel.onEmptyStateButtonTapped(Unknown Source:2)
        at org.wordpress.android.ui.reader.ReaderPostListFragment.setCurrentTagFromEmptyViewButton(ReaderPostListFragment.java:1852)
        at org.wordpress.android.ui.reader.ReaderPostListFragment.access$2400(ReaderPostListFragment.java:150)
        at org.wordpress.android.ui.reader.ReaderPostListFragment$11.onClick(ReaderPostListFragment.java:1754)
```

malinajirka:
Thanks so much @develric!

>I suspect the states of the action buttons at the bottom of the post cards are still temporary in this PR,

Not really, these are already merged into production.

>if we are going to manage the not available functionality as dimmed or removing the button

We are going to use the "dimm" approach. Can you forsee any issues with it?

>Also if we go into the post details view it seems we are currently keeping the "remove what's not available" approach.

We haven't gotten to redesigning the post detail yet. It's planned for Reader Improvements Phase 3.

>I think we already addressed it in the past but at least for me it seems when having only self-hosted site (it worked well with wp.com users) when I tap on SAVED > GO TO FOLLOWING

@zwarm Could you please look into this issue? It seems "Go to following" button on the empty "Saved" tab doesn't work for pure self-hosted users.

develric:
Hi Jirka 😊

> We are going to use the "dimm" approach. Can you forsee any issues with it?

I don't think this can create any issue AFAIU, I was mainly thinking to the difference in approach (dimm vs gone) between the postcard and the post detail but as you explained below

> We haven't gotten to redesigning the post detail yet. It's planned for Reader Improvements Phase 3.

so if not already addressed, maybe something to think of while going into phase 3 if the observation I did makes sense (and it is not granted 😜 )

PR #12393, opened 2020-07-08
Issue/12263 updated post creation qs task

### LINKED ISSUE #12263: Guide users to F.A.B. to create posts during Quick Start
After introducing the FAB, we want to emphasise it as a shortcut to creating posts.

Rather than guiding users to navigate to blog posts and then create a post, lets point them to the F.A.B. and sheet to create their post. Images below

Note: we are taking a slightly different approach with posts and pages so that users have the opportunity to learn creating both ways. See #12262 

blog posts via fab

khaykov:
Fixes #12263

This PR updates the flow for Publish Post Quick Start Task to accommodate IA changes.

Quick Start is relatively flexible in ways it could be used, but it also means it's hard to have a lot of universal functionality for it, and there is a lot of special cases. It can totally use refactoring to accommodate potential new use cases, but it's outside of the scope of this project.

[Image from Gyazo](https://gyazo.com/c05103e93ef248aa04f8c58f028e02e4)

To test:
- Create a new site, and start a Quick Start.
- Navigate to Grow Your Audience category, and select `Publish a post ` task.
- Make sure content of Snackbar is updated, and the focus point is displayed above FAB.
- Rotate the device and make sure the focus point is still visible.
- Tap on the FAB.
- Confirm that bottom dialog is shown and focus point is displayed on `Blog post` row.
- Rotate the device and make sure the focus point is still visible.
- Tap on `Blog post` row, which will take you to the Editor.
- Navigate back from Editor and confirm that the task was completed.

PR submission checklist:


osullivanchris (changes requested):
Tested and everything is working well, I think the new flow makes sense. 

One issue, same as in #12290 - it didn't automatically push me on to the next task.

khaykov:
@osullivanchris I wasn't able to reproduce this. Can you confirm that the task was not completed before you started it? If it was completed you will not see the prompt to move to the next one.
[Image from Gyazo](https://gyazo.com/f79e036717c84a705cc529ef25b710bc)

WordPress/src/main/java/org/wordpress/android/ui/main/MainBottomSheetFragment.kt
```diff
@@ -57,10 +72,71 @@ class MainBottomSheetFragment : BottomSheetDialogFragment() {
...
+        quickStartEvent = event
+
+        if (quickStartEvent?.task == QuickStartTask.PUBLISH_POST) {
+            showQuickStartFocusPoint()
+        }
+    }
+
+    private fun showQuickStartFocusPoint() {
+        // we are waiting for RecyclerView to populate itself with views and then grab the one we need one when it's ready
+        content_recycler_view?.viewTreeObserver?.addOnGlobalLayoutListener(object : OnGlobalLayoutListener {
```
planarvoid, line 103: I think the correct solution with the MVVM would be to include the information whether the quick start is showing into the `MainActionListItem` object. This way you'd only have to call something like `viewModel. showQuickStartFocusPoint` here, this would update the live data with the main actions and you can observe the results and show the point there. The view holder in the recycler view should the quick start point if the `MainActionListItem` object has the flag set. Does it make sense? If you do it this way, you don't need to use the `addOnGlobalLayoutListener ` or find the view holder, all will be automatic.

planarvoid (changes requested):
It works well, good job 👍 . I think we should do some architectural changes to keep the code extensible and correctly follow MVVM

osullivanchris:
> @osullivanchris I wasn't able to reproduce this. Can you confirm that the task was not completed before you started it? If it was completed you will not see the prompt to move to the next one.

Oh yeah that makes sense @khaykov . It was previously completed. Makes sense that it only automatically progresses for new tasks. I didn't realise we had that logic. All good.

WordPress/src/main/java/org/wordpress/android/ui/main/MainBottomSheetFragment.kt
khaykov: I absolutely agree that the correct implementation should utilize MVVM. I learned so much since the time Quick Start was implemented, so I would rewrite it all in a heartbeat (current implementation is not cool) :) 

While putting the logic into view holder, MainActionListItem, and ViewModel is the right way to go, I'm not inclined to do it with the current implementation of Quick Start, since it was built to be as non-intrusive and as decoupled as possible. I would like to keep the logic consistent between different QS tours, so in case we would like to scrap, replace, or modify it, it would be pretty straightforward (so we won't have some parts baked into ViewModel and other manipulating views from above).

oguzkocer:

khaykov:
@planarvoid Hope, I addressed your concerns in the comment :)

@osullivanchris If the feature looks good, do you mind approving this PR?

osullivanchris (approved):

WordPress/src/main/java/org/wordpress/android/ui/main/MainBottomSheetFragment.kt
planarvoid: First of all let me say I trust your judgement so feel free to merge this PR 👍 
However, consider a few things:
- I think it's OK to introduce a new, better way of implementing QuickStart that's not consistent and use this method in all the places where we use view models
- This is a new part of code, with view models and Kotlin. Working around the MVVM flow tells the future devs it's ok to do this when it's not. 
- The GlobalLayoutListener approach is a hack around the RecyclerView
- QuickStart in the viewmodel could be completely extracted. What you'd need is a boolean flag on the model itself. You could have an injected class that tells you `shouldShowQuickStartOnBottomSheet` to populate the flag. You already handle the clicks. 
- The future devs will be facing the same dilemma - clean and inconsistent or hacky and consistent

A I said, feel free to merge this PR. Maybe create a task to implement QS in a ViewModel?

planarvoid (approved):

WordPress/src/main/java/org/wordpress/android/ui/main/MainBottomSheetFragment.kt
khaykov: @planarvoid You are totally right. Especially given all the talk about tech debt :) Not the best idea to go around proper implementation, even if it's justified in one way or another. I'll integrate it into an existing flow, and submit a commit shortly.

khaykov:
@planarvoid Updated the code to follow MVVM and added some tests 👍 Ready for another round :)

planarvoid:
Thanks for that @khaykov, I really appreciate it! 
I've had a super busy day today and I wasn't able to review this PR. I can do it on Monday. Is it OK to move the milestone to 15.5?

khaykov:
@planarvoid Sure thing! Thank you 🙇  Moved the milestone 👍

WordPress/src/main/java/org/wordpress/android/ui/main/MainBottomSheetFragment.kt
```diff
@@ -45,6 +46,15 @@ class MainBottomSheetFragment : BottomSheetDialogFragment() {
             (dialog?.content_recycler_view?.adapter as? AddContentAdapter)?.update(it ?: listOf())
         })
 
+        viewModel.showQuickStarInBottomSheet.observe(this, Observer { showQuickStartFocusPoint ->
```
planarvoid, line 49: I'd actually merge this with the `mainActions` live data and add a `boolean` flag to the `MainActionListItem.CreateAction`. This way you could remove this piece of code and you wouldn't have to use the payload to update the recycler view.

WordPress/src/main/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModel.kt
```diff
@@ -42,6 +42,9 @@ class WPMainActivityViewModel @Inject constructor(
     private val _createAction = SingleLiveEvent<ActionType>()
     val createAction: LiveData<ActionType> = _createAction
 
+    private val _showQuickStarInBottomSheet = MutableLiveData<Boolean>()
+    val showQuickStarInBottomSheet: LiveData<Boolean> = _showQuickStarInBottomSheet
+
```
planarvoid, line 47: Instead of exposing this live data, you could merge both this and the main actions into one view state model for the whole bottom sheet. It could look like this: 
```
private val _showQuickStarInBottomSheet = MutableLiveData<Boolean>()

    private val _mainActions = MutableLiveData<List<MainActionListItem>>()
    val mainActions: LiveData<List<MainActionListItem>> = merge(_mainActions, _showQuickStarInBottomSheet) { mainActions, showQuickStart ->
        if (showQuickStart != null && mainActions != null) {
            mainActions.map { if (it is CreateAction && it.actionType == CREATE_NEW_POST) it.copy(showQuickStart = showQuickStart) else it }
        } else {
            mainActions
        }
    }
```
There are a few advantages of this principle. One is that you can see the amount of LiveData already existing in this class. This helps us keep everything related in one place. It's easy to check the state of the whole component if everything is in one place. It's also very easy to test the whole state. You can test both that an action is present and that it's showing the quick start. 

This example requires something like this in the `CreateAction`
```
data class CreateAction(
        override val actionType: ActionType,
        @DrawableRes val iconRes: Int,
        @StringRes val labelRes: Int,
        val onClickAction: ((actionType: ActionType) -> Unit)?,
        val showQuickStart: Boolean = false
    ) : MainActionListItem()
```

WordPress/src/test/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModelTest.kt
```diff
@@ -164,6 +168,44 @@ class WPMainActivityViewModelTest {
         assertThat(viewModel.isBottomSheetShowing.value!!.peekContent()).isEqualTo(true)
     }
 
+    @Test
```
planarvoid, line 171: ❤️

planarvoid (commented):
Thanks for the changes! It's looking much better. I have one additional improvement suggestion 👍 . Let me know what you think

khaykov:
@planarvoid Thank you so much for the suggestions! I was contemplating integrating QS with `mainActions`, but my initial approach was a bit clunky, so I went the other way. Using `MediatorLiveData`, as you suggested, is a great way to simplify it 👍

planarvoid (approved):
thanks for the changes! It looks really good and works well 👍 . I think this is a great approach for most of the screens

Tiay8885 (commented):
`issue/12263-updated-post-creation-qs-task `

PR #12413, opened 2020-07-10
Remove Group constraint layout element

### LINKED ISSUE #12353: Remove Group constraint layout component
### Expected behavior
When I toggle "Bookmark" on Reader post list item, photo title's visibility doesn't change.

### Actual behavior
When I toggle "Bookmark" on Reader post list item, photo title sometimes disappears/reappears unexpectedly - happens just sometimes and just on some posts.

Reported [here](https://github.com/wordpress-mobile/WordPress-Android/pull/12323#issuecomment-652817423)

1. Open reader
2. Scroll to a post with "Photo title"
3. Click on "Bookmark" button
4. Notice the title disappears/reappears - if it doesn't happen scroll to a different post

This issue is caused by an Android bug. Constraint layout "Group" element doesn't play well with visibility changes to individual elements in the group. More info [here](https://github.com/wordpress-mobile/WordPress-Android/pull/12323#issuecomment-654647911)

ashiagr:
Fixes #12353

To test:

- Toggle "Bookmark" on random Reader post list items.
- Notice that photo title visibility doesn't change and whenever visible it actually belongs to that post.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderPostAdapter.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderPostAdapter.java
@@ -392,7 +392,7 @@ public class ReaderPostAdapter extends RecyclerView.Adapter<RecyclerView.ViewHol
         };
 
         Function2<Long, Long, Unit> onVideoOverlayClicked = (postId, blogId) -> {
-            ReaderActivityLauncher.showReaderBlogPreview(ctx, post);
+            ReaderActivityLauncher.showReaderVideoViewer(ctx, post.getFeaturedVideo());
             return Unit.INSTANCE;
         };
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderPostViewHolder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderPostViewHolder.kt
@@ -115,7 +115,7 @@ class ReaderPostViewHolder(
     }
 
     private fun updateDiscoverSection(state: ReaderPostUiState) {
-        uiHelpers.updateVisibility(image_discover_avatar, state.discoverSection != null)
+        uiHelpers.updateVisibility(image_discover_avatar, state.discoverSection?.discoverAvatarUrl != null)
         uiHelpers.updateVisibility(layout_discover, state.discoverSection != null)
         uiHelpers.setTextOrHide(text_discover, state.discoverSection?.discoverText)
         if (state.discoverSection?.discoverAvatarUrl == null) {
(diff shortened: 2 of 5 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderPostViewHolder.kt
```diff
@@ -65,7 +65,6 @@ class ReaderPostViewHolder(
         updateFeaturedImage(state)
         uiHelpers.updateVisibility(image_video_overlay, state.videoOverlayVisibility)
         uiHelpers.setTextOrHide(text_photo_title, state.photoTitle)
-        uiHelpers.updateVisibility(frame_photo, state.photoFrameVisibility)
```
malinajirka, line 68: Shouldn't we just replace `frame_photo` with `image_featured` on this line instead of removing it?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderPostViewHolder.kt
```diff
@@ -116,6 +115,7 @@ class ReaderPostViewHolder(
     }
 
     private fun updateDiscoverSection(state: ReaderPostUiState) {
+        uiHelpers.updateVisibility(image_discover_avatar, state.discoverSection != null)
```
malinajirka, line 118: Shouldn't the condition be `state.discoverSection?.discoverAvatarUrl != null`, wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderPostViewHolder.kt
ashiagr: Yep but it should also be visible when featured video is available - it is noticeable after removing the "Group" element. Updated here: 38bd162

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderPostViewHolder.kt
ashiagr: Updated here 38ec932

ashiagr:
Noticed that video click listener was not added and video click action was not correct - it was opening the blog instead of playing the video - sorry that I could not catch it in my review earlier. I addressed in this PR - 66eefc8.

WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderPostAdapter.java
```diff
@@ -392,7 +392,7 @@ private void renderPost(final int position, final ReaderPostViewHolder holder) {
         };
 
         Function2<Long, Long, Unit> onVideoOverlayClicked = (postId, blogId) -> {
```
ashiagr, line 394: We're not using (postId, blogId) - can we remove these?

malinajirka: We are not using them in the adapter because the renderPost method has access to a ReaderPost reference. However, I believe the VM won't have access to such reference so we'll need them (eg [here](https://github.com/wordpress-mobile/WordPress-Android/blob/9e20bd906574062a829071b0a15f5843c9292b36/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt#L130)). I'll clean-up all these listeners after the actions are actually implemented.

PR #12415, opened 2020-07-10
Disable primary buttons on empty fields

### LINKED ISSUE #12243: Unified Login & Sign-Up: Primary buttons are enabled before form fields are filled

### Expected behavior
The primary button used on the login flow should only be enabled when the relevant screen form fields have been filled.

### Actual behavior
On most screens, the primary button is enabled before any of the form fields have been filled. 

Note: we should probably add a list of the screens where this behavior is desired (cc @mattmiklic).

planarvoid:
Fixes #12243 

This PR disables the primary buttons when the input fields are empty on following screens:
- Login with email - when the email input is empty
- Password with input - when the user decides to input password instead of using magic links
- Username and password input - when the user is logging in with a site address

One more screen that was already implementing this behaviour is the screen where the user inputs their site address to login. Is there any other screen you think we should fix?

To test:
- Start with a cleaned app
- Click on "Continue with Wordpress" and continue to the next screen
- Notice that the "Continue button is disabled
- Input anything into the "Email" input
- Notice that the "Continue button is now enabled
- Try writing and deleting the email, rotating the screen
- Use a valid email and continue to the next screen
- On the magic link screen decide to login with password instead
- Notice that the "Continue" button is now disabled
- Type a password
- Notice that it's now enabled

To test:
- Start with a cleaned app
- Click on "Enter your site address"
- Input a valid self-hosted site address and continue to the next screen
- Notice that the "Continue" button is now disabled
- Type in a username and a password
- Notice that the continue button is only enabled when both username and password are set

PR submission checklist:

### DIFF
--- a/libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginEmailPasswordFragment.java
+++ b/libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginEmailPasswordFragment.java
@@ -109,6 +109,7 @@ public class LoginEmailPasswordFragment extends LoginBaseFormFragment<LoginListe
     @Override
     public void onResume() {
         super.onResume();
+        updatePrimaryButtonEnabledStatus();
 
         // connect to the Service. We'll receive updates via EventBus.
         mServiceEventConnection = new AutoForeground.ServiceEventConnection(getContext(),
@@ -125,6 +126,15 @@ public class LoginEmailPasswordFragment extends LoginBaseFormFragment<LoginListe
 
         // disconnect from the Service
         mServiceEventConnection.disconnect(getContext(), this);
+    }
+
+    private void updatePrimaryButtonEnabledStatus() {
+        View view = getView();
+        if (view != null) {
+            Button primaryButton = (Button) view.findViewById(R.id.primary_button);
+            String currentPassword = mPasswordInput.getEditText().getText().toString();
+            primaryButton.setEnabled(!currentPassword.trim().isEmpty());
+        }
     }
 
     @Override
@@ -266,6 +276,7 @@ public class LoginEmailPasswordFragment extends LoginBaseFormFragment<LoginListe
         mPasswordInput.setError(null);
 
         LoginWpcomService.clearLoginServiceState();
+        updatePrimaryButtonEnabledStatus();
     }
 
     private void showPasswordError() {
--- a/libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginEmailPasswordFragment.java
+++ b/libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginEmailPasswordFragment.java
@@ -129,11 +129,8 @@ public class LoginEmailPasswordFragment extends LoginBaseFormFragment<LoginListe
     }
 
     private void updatePrimaryButtonEnabledStatus() {
-        View view = getView();
-        if (view != null) {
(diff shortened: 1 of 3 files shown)


libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginEmailPasswordFragment.java
```diff
@@ -127,6 +128,14 @@ public void onPause() {
         mServiceEventConnection.disconnect(getContext(), this);
     }
 
+    private void updatePrimaryButtonEnabledStatus() {
+        View view = getView();
+        if (view != null) {
```
renanferrari, line 133: Why exactly do we need to check this here?

libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginUsernamePasswordFragment.java
```diff
@@ -261,6 +261,20 @@ public void onSaveInstanceState(Bundle outState) {
...
     }
 
+    @Override public void onResume() {
+        super.onResume();
+        updatePrimaryButtonEnabledStatus();
+    }
+
+    private void updatePrimaryButtonEnabledStatus() {
+        View view = getView();
+        if (view != null) {
```
renanferrari, line 271: Same thing here.

libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginEmailPasswordFragment.java
planarvoid: I've added this check because Android Studio complains that the view might be null with a warning. I'm fine with removing it if you think it doesn't make sense 👍

renanferrari: I understand how the check is necessary for `LoginEmailFragment` since we're using the `view` to find a button:

```java
private void updateContinueButtonEnabledStatus() {
    View view = getView();
    if (view != null) {
        Button continueButton = (Button) view.findViewById(R.id.login_continue_button); // <-- We use the view here
        String currentEmail = mEmailInput.getEditText().getText().toString();
        continueButton.setEnabled(!currentEmail.trim().isEmpty());
    }
}
```

But for `LoginEmailPasswordFragment` or `LoginUsernamePasswordFragment` it looks like we are not using the `view` to do anything directly:

```java
private void updatePrimaryButtonEnabledStatus() {
    View view = getView();
    if (view != null) {
        // We don't use the view anywhere?
        String currentPassword = mPasswordInput.getEditText().getText().toString();
        getPrimaryButton().setEnabled(!currentPassword.trim().isEmpty());
    }
}
```

Are we keeping the check there just to make sure the Fragment's main view was already inflated or am I missing something?

planarvoid: ah you're right, I think I used to use the view and removed it later. I'll fix this!

libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginUsernamePasswordFragment.java
planarvoid: Removed in 824abb8

planarvoid:
thanks for the review @renanferrari . I've removed the non-null checks and this should be ready for another pass 👍

PR #12419, opened 2020-07-10
Avoid NPE when webeditor is null

mchowning:
~When testing unsupported block editing for the gutengerg 1.32 release, I observed that on my pressable site the webEditor was null and this was crashing the app. This addresses that by reversing the `.equals()` check to avoid the NPE.~

UPDATE: This PR was originally fixing a NPE, but it grew to address a newly discovered bug on self-hosted sites, ~and then eventually to just disable the feature for the time being to allow for more testing~.

To test:
Load the gutenberg editor on a site where the webeditor value is null (a Pressable non-jetpack connected site is what I used), and confirm that the app does not crash.

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,7 +2,7 @@
 -----
 * [*] Improve wording for confirmation dialogs when trashing posts
 * [***] Block Editor: Adds Copy, Cut, Paste, and Duplicate functionality to blocks
-* [***] Block Editor: Users can now individually edit unsupported blocks found in posts or pages. Not available on Jetpack connected self-hosted sites.
+* [***] Block Editor: Users can now individually edit unsupported blocks found in posts or pages. Not available on selfhosted sites or sites defaulted to classic editor.
  
 15.2
 -----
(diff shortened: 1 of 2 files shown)


mchowning:
Tested and I am seeing that 
- I **can** edit unsupported posts with wpcom sites, including atomic. 
- I **cannot** edit unsupported blocks on self-hosted sites regardless of whether they are connected to jetpack. 

I believe this is the expected behavior at this time, so I think this PR is good to merge.

mchowning:
We realized that it is possible for wp.com sites to have the block editor disabled as well, and we have not thoroughly tested that scenario, so a214faf disables the editing unsupported blocks feature since we need to merge the release. We can always update the beta app before it's released if we decide to turn it back on later after some more testing.

guarani:
Tested build 60897 on Galaxy S10 with Android 10.

- [x] [TC001: User can edit unsupported blocks on regular WP.com sites](https://github.com/wordpress-mobile/test-cases/blob/master/test-cases/gutenberg/unsupported-block-editing.md#tc001)
- [x] [TC002: Ensure discarded edits are not persisted](https://github.com/wordpress-mobile/test-cases/blob/master/test-cases/gutenberg/unsupported-block-editing.md#tc002)
- [x] [TC003: User can edit unsupported blocks on regular WP.com Business (Atomic) sites](https://github.com/wordpress-mobile/test-cases/blob/master/test-cases/gutenberg/unsupported-block-editing.md#tc003)

Issue #12316, opened 2020-06-29
WordPress Stories : Refactor post type boolean flags to enum


jd-alexander:
Create an enum that we can utilize to determine the type of post that's triggering the bottom sheet as described [here](https://github.com/wordpress-mobile/WordPress-Android/pull/12307#discussion_r446934266)

https://github.com/wordpress-mobile/WordPress-Android/blob/e30d89d991d5be36078a47af7b56910efee2593d/WordPress/src/main/java/org/wordpress/android/ui/posts/PostsListActivity.kt#L341

> can we consist these 2 flags into an enum that holds Post, Page, Story post type of values? So we wouldn't be passing around 2 variables when we could pass one, and moreover we would make sure to not allow any ambiguities or impossible states such as isPage and isStoryPost being both true (which cannot be)

Also do refactoring of view types within the `PrepublishingViewHolder`https://github.com/wordpress-mobile/WordPress-Android/pull/12306#discussion_r446943631

SiobhyB:
Stories have been deprecated as of 24.1 (released to users yesterday), so I'll go ahead to close this issue.

PR #12442, opened 2020-07-14
issue-12442/ added events for bottom sheet.

### LINKED ISSUE #12315: WordPress Stories : Add tracks for Prepublishing Integration
As mentioned [here](https://github.com/wordpress-mobile/WordPress-Android/pull/12307) tracks will need to added so that we know how many people are publishing stories.

jd-alexander:
Fixes #12315

## Solution
Added analytics events to track when the bottom sheet is opened and when the publish button is clicked. 

## Testing
1. Create a new story post. 
2. Open the bottom sheet. 
3. Ensure that the `STORY_POST_BOTTOM_SHEET_OPENED` is triggered. 
4. Click Publish. 
5. Ensure that the  `STORY_POST_PUBLISH_TAPPED ` is triggered.
------------
1. Create a new story post. 
2. Open the bottom sheet. 
3. Ensure that the `STORY_POST_BOTTOM_SHEET_OPENED` is triggered. 
4. Dismiss the bottom sheet. 
5. Ensure that the  `PREPUBLISHING_BOTTOM_SHEET_DISMISSED ` is triggered.

## Reviewing 

## Submitter Checklist


aforcier:
This is working well!

> should we make events more broad for the bottom sheet being opened (regardless of where it is being opened / closed / etc?), as opposed to having them specialized to the Story composer and as such only triggering them when the bottom sheet is opened in the Story composer?

I wondered about this too, it seems a bit off to have `story_post_bottom_sheet_opened` but then `prepublishing_bottom_sheet_dismissed`. I think a good approach might be to use broad naming everywhere (`prepublishing_bottom_sheet_opened`, `prepublishing_bottom_sheet_dismissed`).

That brings me to:

> should we add more information about the Post on each event? Seems not necessary if we can create a funnel with the information being gathered already on #12074, but wanted to double check with you

I think it might end up being useful to be extra informative here and attach a bit of extra information here - probably whether it's a page and whether it's a story post since those are the two params given to the bottom sheet fragment when it's initialized.

I could also see instead of the story post flag, having a `context` param that specifies what editor the bottom sheet is attached to instead (`gutenberg` or `wp_stories_creator`, though I suppose Aztec is in the mix as well 🤔).

I'm not sure how any of this aligns with how WP Mobile is thinking about analytics for the bottom sheet in general though, @jd-alexander? I would maybe defer to that team and just go with `prepublishing_bottom_sheet_opened` and no params in this PR.

jd-alexander:
Thanks for all the feedback @aforcier 😄 

> I wondered about this too, it seems a bit off to have story_post_bottom_sheet_opened but then prepublishing_bottom_sheet_dismissed. I think a good approach might be to use broad naming everywhere (prepublishing_bottom_sheet_opened, prepublishing_bottom_sheet_dismissed).

This is a really good point. I will make the relevant change so the naming is consistent. 

> I think it might end up being useful to be extra informative here and attach a bit of extra information here - probably whether it's a page and whether it's a story post since those are the two params given to the bottom sheet fragment when it's initialized.

Thanks! This makes sense! I can track this as another issue and refactor how we go about handling all the events for the Prepublishing Nudges after chatting with the team. @mzorz gave some good feedback about how we are handling the type of `PostModel` the bottom sheet is working with by [using an enum](https://github.com/wordpress-mobile/WordPress-Android/issues/12316) that represents a `Page`, `Post` or `Story`. So with something like this implemented, an analytics helper could then attach the respective properties based on the post type.  A question though. How is historical data impacted by changes like this? Is it a case where since we are just adding properties to the events then we would be good? 

> I could also see instead of the story post flag, having a context param that specifies what editor the bottom sheet is attached to instead (gutenberg or wp_stories_creator, though I suppose Aztec is in the mix as well 🤔).

Thanks for sharing this! This is what I wanted to know if this kind of data should be added to each event or if other events could be used in a funnel to get this data, so it's good to get some clarification on this. 

> I'm not sure how any of this aligns with how WP Mobile is thinking about analytics for the bottom sheet in general though, @jd-alexander? I would maybe defer to that team and just go with prepublishing_bottom_sheet_opened and no params in this PR.

I could do a check with my team about this because indeed there has to be consistency in how we are approaching these events on both platforms so that the funnels and metrics we are able to get from each platform are similar.

jd-alexander:
Update so I spoke to the team and garnered some feedback I will share below : 

> I could also see instead of the story post flag, having a context param that specifies what editor the bottom sheet is attached to instead (gutenberg or wp_stories_creator, though I suppose Aztec is in the mix as well 🤔)

So for this comment, when we were tracking the publish action we weren't too concerned about the editor that opened the bottom sheet because we were simply tracking other events that were more centered around the goals of the project such as when the sheet is opened when published is clicked, tags are added etc. But in this case, if it will be helpful it wouldn't hurt to include them in this PR. I would have to do some refactoring as mentioned for the param that tells the type of post. I would prefer to wait a bit on that part and create an issue for doing this so when I get to that enum/ post type refactor I would create a PR on top of that to add the data.

I will be renaming the event to `prepublishing_bottom_sheet_dismissed` for consistency and then we can merge this PR.

Again anymore more thoughts on this matter are greatly appreciated as I am still learning how to create events with substantial data. cc @mzorz

jd-alexander:
So for this change eeb8104 I ended up removing that event because I realized it doesn't need to be tracked because when the submit button is clicked this is when the post upload is triggered which will result in the post published event being emitted. Let me know if that was a good call 🙏

jd-alexander:
Fixing the conflicts now!

mzorz:
> So for this change eeb8104 I ended up removing that event because I realized it doesn't need to be tracked because when the submit button is clicked this is when the post upload is triggered which will result in the post published event being emitted. Let me know if that was a good call 🙏

What happens if you tap on PUBLISH and then don't have connectivity at all: will the post published event be triggered? (disclaimer: I haven't tried, hence I'm asking) I think the tap on publish event was good to get tracked anyway, as it proves a 100% intent from the user (they undoubtedly tapped on the PUBLISH button, as opposed to us inferring it from a different event). Let me know your thoughts

aforcier:
> What happens if you tap on PUBLISH and then don't have connectivity at all: will the post published event be triggered? (disclaimer: I haven't tried, hence I'm asking) I think the tap on publish event was good to get tracked anyway, as it proves a 100% intent from the user (they undoubtedly tapped on the PUBLISH button, as opposed to us inferring it from a different event). Let me know your thoughts

Agreed, I think it's worth tracking this, even if it covers similar ground now. Another example might be, when we eventually have story drafts, the user would presumably be able to publish directly from the post list - and this might conceivably be something we'd want to distinguish in tracks events.

jd-alexander:
Okay I agree as well about tracking the submit action irregardless of the tracking that's taking place in other parts of the flow . Thanks for the thoughts. I will revert that change. 🙏🏾

jd-alexander:
@mzorz this is ready for another round :)

PR #12453, opened 2020-07-15
Prepublishing Nudges / Post Settings - Older drafts says Immediately instead of Publish On

### LINKED ISSUE #12197: Prepublishing Nudges : Old draft state for date

jd-alexander:
Fixes #12197

## Findings
This previous PR https://github.com/wordpress-mobile/WordPress-Android/pull/12198 contains the direction for this current implementation where we are showing `Immediately` instead of the older date on the `PostModel`.

## Solution
This solution simply shows `Immediately` instead of the `Publish On`.
## Testing

1. Go to the Post List. 
2. Click publish on a old draft. 
3. Ensure the bottom sheet's Publish Date says Immediately instead of the date. 

<kbd></kbd>

-----------
1. Go to the Post List. 
2. Open an old draft in the Editor. 
3. Ensure the bottom sheet's Publish Date says Immediately instead of the date. 

<kbd></kbd>

----------
1. Go to the Post List. 
2. Open an old draft in the Editor. 
3. Click on the Post Settings. 
4. See Immediately under the Publish Date. 

<kbd></kbd>

## Concerns 
I was wondering if the `PostSettingsUtilsTest` would have to be refactored for this change. The tests still passed and covered it but a second opinion would be appreciated. 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PostSettingsUtils.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PostSettingsUtils.kt
@@ -40,7 +40,7 @@ class PostSettingsUtils
             } else if (postUtilsWrapper.isPublishDateInTheFuture(postModel.dateCreated)) {
                 labelToUse = resourceProvider.getString(R.string.schedule_for, formattedDate)
             } else {
-                labelToUse = resourceProvider.getString(R.string.publish_on, formattedDate)
+                labelToUse = resourceProvider.getString(R.string.immediately)
             }
         } else if (postUtilsWrapper.shouldPublishImmediatelyOptionBeAvailable(status)) {
             labelToUse = resourceProvider.getString(R.string.immediately)


jd-alexander:
@malinajirka added you as a reviewer here because we spent a good amount of time with PostSettings, so I know you will have more context about these changes, hence my question, in the PR description about needed a test case to cover this change. Let me know what you think 😄

malinajirka:
Thanks @jd-alexander! 

It seems this PR reveals another bug. IMO the user should be able to back-date a publish date. 

So we have two separate issues in `develop`
1. When I click on an old draft the publish date says "Publish on 1st of June" but the action button says "Publish Now". The question is whether the post should be published on 1st of June and the button should say "Publish".  Or if the post should be published now and the action button should say "Publish Now".

2. When I select an old date, I should be able to publish the post with the selected date.

Regarding the change in this PR - the user is still allowed to pick an old date, but the label keeps saying "Immediately". This doesn't feel right to me. Wdyt? (+ the issue with back-dating a post) 

I think label and behavior of the bottom sheet action button should correspond to the action user invoked to open the bottom sheet (assuming the user doesn't do any changes within the bottom sheet). So when I click on "Publish", I'd expect to see "Publish" action in the bottom sheet. When I click on "Publish Now", I'd expect to see "Publish Now" action in the bottom sheet. When the user changes something within the bottom sheet, the button should reflect that change (I believe this part almost works). 


>I was wondering if the PostSettingsUtilsTest would have to be refactored for this change. The tests still passed and covered it but a second opinion would be appreciated.

I like to add a unit test for each bug I fix. I think it's a good practice to prevent the bug from happening again in the future.

jd-alexander:
Closing this and will convert this into an issue so another contributor can continue.

PR #12454, opened 2020-07-15
Issue/12039 reader like action

zwarm:
Part of #12039 

This PR is 
- implements the Post "like" action.
- Refactors the ReaderRepository events to their own file

Background: `ReaderPostActions.performLikeAction` is a static method that runs an async operation to submit the "like" request to the server. The method does not wait for the request to return, but instead returns a true/false to the user immediately. The method constructs two listeners (error/success) for callbacks on the rest request. There is no action if the rest call is successful, but in the case of a failure the local DB changes are reverted.

With this implementation I worked within the confines of the existing `ReaderPostActions.performLikeAction` method and created a new EventBus event `OnReaderRepositoryEvent` which will get posted once the rest callbacks (success or error) have been completed or there are no changes to be made. The addition of these events will not effect the existing functionality and will only be listened for in `PostLikeActionUseCase`.  

The `PostListActionUseCase` transforms the `OnReaderRepositoryEvent` to a coroutine so the rest callbacks can be handled outside of the static method. This gives us flexibility in the way we handle and/or communicate back to the caller. 
 
I am hopeful that we can use this pattern for the other Reader Actions implementations. 

In terms of the Like action itself, we should consider what the caller needs to be aware of. Do they need to receive Unchanged, Success, and Failure states? Should the new methods return a `LiveData<Event<Boolean>>` or possibly a `LiveData<Event<ReaderPost>>`?  
It was decided that no response is returned to the caller of the Like function and optimistic success will granted. Failures will be returned on the communicationChannel. In either case, the repository will determine if posts need to be reloaded and the observers to LiveData will be notified.

I would love to hear your thoughts. Also, I am not tied to any of the chosen event or use case names, so feel free to offer suggestions. There are plenty of todos that will also give some clues to what I am thinking for a future iteration.

~~FYI: Sorry about the number of file changed, this PR had to be based off of PR #12270 which hasn't been merged into develop yet.~~


WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderPostActions.java
```diff
@@ -91,6 +96,9 @@ public static boolean performLikeAction(final ReaderPost post,
             @Override
             public void onResponse(JSONObject jsonObject) {
                 AppLog.d(T.READER, String.format("post %s succeeded", actionName));
+                final OnPostLikeSuccess onPostLikeSuccess =
+                        new OnReaderRepositoryEvent.OnPostLikeEnded.OnPostLikeSuccess(post.postId, post.blogId);
+                EventBus.getDefault().post(onPostLikeSuccess);
```
zwarm, line 101: New post on eventbus

WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderPostActions.java
```diff
@@ -106,13 +114,15 @@ public void onErrorResponse(VolleyError volleyError) {
                 AppLog.e(T.READER, volleyError);
                 ReaderPostTable.setLikesForPost(post, post.numLikes, post.isLikedByCurrentUser);
                 ReaderLikeTable.setCurrentUserLikesPost(post, post.isLikedByCurrentUser, wpComUserId);
+                EventBus.getDefault().post(OnPostLikeFailure.INSTANCE);
```
zwarm, line 117: New post on eventbus

WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderPostActions.java
```diff
@@ -67,6 +71,7 @@ public static boolean performLikeAction(final ReaderPost post,
         boolean isCurrentlyLiked = ReaderPostTable.isPostLikedByCurrentUser(post);
         if (isCurrentlyLiked == isAskingToLike) {
             AppLog.w(T.READER, "post like unchanged");
+            EventBus.getDefault().post(OnPostLikeUnChanged.INSTANCE);
```
zwarm, line 74: New post on eventbus

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderPostRepository.kt
```diff
@@ -3,451 +3,182 @@ package org.wordpress.android.ui.reader.repository
...
+
+    fun getTag(): ReaderTag {
+        return readerTag
+    }
 
-    suspend fun getDiscoveryFeed(): LiveData<ReaderPostList> =
-            withContext(bgDispatcher) {
-                delay(TimeUnit.SECONDS.toMillis(5))
-                getMockDiscoverFeed()
+    fun performLikeAction(post: ReaderPost, isAskingToLike: Boolean, wpComUserId: Long) {
```
zwarm, line 75: My gut says that there is an opportunity to move this into an "PostLikeManager" of sorts. Keep the repository lighter and the logic SRPish.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderPostRepository.kt
```diff
@@ -3,451 +3,182 @@ package org.wordpress.android.ui.reader.repository
...
+                    readerUpdatePostsEndedHandler,
+                    postLikeActionUseCase
+
+            )
+        }
+    }
+}
+
+// todo: annmarie once it's decided to keep this, then move out of repository file
+sealed class OnReaderRepositoryEvent {
```
zwarm, line 177: New Event declarations. This can be built out to support other reader actions (bookmark), etc.

WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderPostActions.java
malinajirka: I'd consider including blog+post id even for `OnPostLikeUnChanged` and `OnPostLikeFailure` so we can easily handle multiple calls in parallel. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeActionUseCase.kt
```diff
@@ -0,0 +1,54 @@
...
+import javax.inject.Inject
+import javax.inject.Named
+import kotlin.coroutines.resume
+import kotlin.coroutines.suspendCoroutine
+
+class PostLikeActionUseCase @Inject constructor(
+    @Named(IO_THREAD) private val ioDispatcher: CoroutineDispatcher,
+    private val eventBusWrapper: EventBusWrapper
+) : ReaderRepositoryDispatchingUseCase(ioDispatcher) {
+    private var continuation: kotlin.coroutines.Continuation<OnReaderRepositoryEvent>? = null
```
malinajirka, line 21: We might want to consider keeping a list/array/map of continuations so we can support multiple parallel calls, wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderPostRepository.kt
malinajirka: I agree that we might want to consider extracting this logic outside of this class. Having said that, I'd personally leave it here for now and extract it when we have all the actions implemented and we see what could be extracted. Wdyt?

malinajirka:
Great job @zwarm! I've reviewed the code and it all LGTM. The only thing I'm not sure about is if we will need to propagate to the UI that the like state has changed in the db right after it happens -> when the user clicks on the "Like" button I think we'll want to change the state on the Ui to "Liked". It seems the current implementation waits for the request to finish. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderPostActions.java
zwarm: 100% agree - and there is a todo in here somewhere with that exact thought. 👍

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderPostRepository.kt
zwarm: > leave it here for now
👍 

Yes, I can live with the todo's as placeholder reminders during development.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeActionUseCase.kt
zwarm: That's a great idea ... was only thinking single with this draft, but that makes a lot of sense.

zwarm:
> The only thing I'm not sure about is if we will need to propagate to the UI that the like state has changed in the db right after it happens -> when the user clicks on the "Like" button I think we'll want to change the state on the Ui to "Liked". 

@malinajirka This is definitely one of my open questions.

> Update: Another approach would be to show some kind of loading/inProgress state. Wdyt?

Is this an area where we'd be better off with a direct response from the method or a reactive stream? So a direct return of LikeState or  LiveData<Event<LikeState>> with postId, blogId, etc? I'm leaning towards the latter because we can continue to update as things occur - we could possibly include the updated ReaderPost on success. Thoughts?

malinajirka:
>Is this an area where we'd be better off with a direct response from the method or a reactive stream? So a direct return of LikeState or LiveData<Event> with postId, blogId, etc? I'm leaning towards the latter because we can continue to update as things occur - we could possibly include the updated ReaderPost on success. Thoughts?

My first though is that it'd just complicate things. I can think of two relatively easy options
1. Ignore that the current implementation uses optimistic update and just show progress state and wait for the result from the server.
2. Emit an event whenever someone changes something in the database -> listen for it in the repo layer and reload the data from the db on change.

Either case we can discuss all this on Monday. I'm sure there are more options I haven't considered.

zwarm:
> 2\. Emit an event whenever someone changes something in the database -> listen for it in the repo layer and reload the data from the db on change.

Ha - that's the same conclusion I came to yesterday when thinking about the other parts of the app updating the tables. I agree, we have plenty to discuss next week. :)

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeActionUseCase.kt
zwarm: In commit aec3678

WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderPostActions.java
zwarm: in commit aec3678

zwarm:
> > 2. Emit an event whenever someone changes something in the database -> listen for it in the repo layer and reload the data from the db on change.
> 
> Ha - that's the same conclusion I came to yesterday when thinking about the other parts of the app updating the tables. I agree, we have plenty to discuss next week. :)

Logic for ReaderPostTable modifications will be in a separate PR.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderDiscoverRepository.kt
```diff
@@ -63,6 +70,28 @@ class ReaderDiscoverRepository constructor(
...
+        postLikeActionUseCase.stop()
+    }
+
+    fun performLikeAction(post: ReaderPost, isAskingToLike: Boolean, wpComUserId: Long) {
+        launch {
+            val event: ReaderRepositoryEvent
+            try {
+                event = postLikeActionUseCase.perform(post, isAskingToLike, wpComUserId)
+            } catch (e: IllegalStateException) {
+                return@launch
```
malinajirka, line 82: We might want to at least log an exception here. Empty catch statements are considered a bad practice. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderPostRepository.kt
```diff
@@ -29,16 +35,18 @@ class ReaderPostRepository(
...
     override val coroutineContext: CoroutineContext
         get() = bgDispatcher
 
     private var isStarted = false
+    private var isDirty = false
 
-    private val _postsForTag = ReactiveMutableLiveData<ReaderPostList>(
-            onActive = { onActivePostsForTag() }, onInactive = { onInactivePostsForTag() })
-    val postsForTag: LiveData<ReaderPostList> = _postsForTag
+    private val _mutablePosts = ReactiveMutableLiveData<ReaderPostList>(
```
malinajirka, line 47: In general, we don't use the `mutable` prefix in the codebase. The underscore is what indicates that the livedata are mutable. I'd consider removing the `mutable` part. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderPostRepository.kt
```diff
@@ -66,6 +76,30 @@ class ReaderPostRepository(
...
     }
 
+    // todo: annmarie - Possibly implement a "LikeManager" that will encapsulate all the "UseCase".
+    fun performLikeAction(post: ReaderPost, isAskingToLike: Boolean, wpComUserId: Long) {
+        launch {
+            val event: ReaderRepositoryEvent
+            try {
+                event = postLikeActionUseCase.perform(post, isAskingToLike, wpComUserId)
+            } catch (e: IllegalStateException) {
+                return@launch
```
malinajirka, line 86: Same as above.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderRepositoryEvents.kt
```diff
@@ -0,0 +1,31 @@
+package org.wordpress.android.ui.reader.repository
```
malinajirka, line 1: Have you considered using data classes? For example `PostLikeSuccess, PostLikeFailure ...` could be data classes, right?

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeActionUseCase.kt
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    override fun stop() {
+        super.stop()
+        eventBusWrapper.unregister(this)
+        continuations.run { clear() }
+    }
+
+    data class PostLikeRequest(val postId: Long, val blogId: Long, val isAskingToLike: Boolean, val wpComUserId: Long) {
+        override fun equals(other: Any?): Boolean {
```
malinajirka, line 62: Could you please elaborate why we need to implement equals/hashcode for a data class? It seems the custom implementation doesn't add any extra checks comparing to what the compiler would generate for the data class.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeActionUseCase.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    suspend fun perform(
+        post: ReaderPost,
+        isAskingToLike: Boolean,
+        wpComUserId: Long
+    ): ReaderRepositoryEvent {
+        val request = PostLikeRequest(post.postId, post.blogId, isAskingToLike, wpComUserId)
+
+        if (continuations[request] != null) {
+            throw IllegalStateException("Perform Like action has already been sent for this request")
```
malinajirka, line 36: I might have been the one who proposed throwing an exception here. However, I'm a bit worried it might bite us. Imagine the user keeps clicking on "like->unlike -> like -> unlike". The app might crash with this exception, right?

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeActionUseCase.kt
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        return suspendCoroutine { cont ->
```
malinajirka, line 39: I'm wondering if we need to use `suspendCancellableCoroutine` here. We are calling `super.stop()`, which invokes `parentJob.cancel()`. But I'm not sure it'll have any effect when the suspendCoroutine isn't cancelable. Tbh I'm not too familiar with these concepts. Let me know what you think.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderDiscoverRepository.kt
zwarm: Sure, no problem, I'll log an event.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderPostRepository.kt
zwarm: No biggie, I'll remove it.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderPostRepository.kt
zwarm: I'll add the log.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeActionUseCase.kt
zwarm: > Imagine the user keeps clicking on "like->unlike -> like -> unlike". 

@malinajirka I tested that scenario and it works as I expected. It stops a duplicate request from happening and did not crash. Now if that is not the required behavior, we can turn the map into a queue and process each request in order, so that we can match each request with the proper continuation.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeActionUseCase.kt
zwarm: @malinajirka  I was experimenting with time, x number of minutes from last request, etc. This can actually go now. I'll kill it

zwarm: Remove din 4eb707e

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeActionUseCase.kt
malinajirka: Ohh I understand it now. We are catching the exception in the layer above. In that case the issue is somewhere else. Is there any reason why we are throwing an exception for expected(valid) user flow? Moreover, the current design forces you to know the implementation of PostLikeActionUseCase when you want to use it which might not be ideal, wdyt?
I'd consider logging a debug message here and returning from the method instead of throwing an exception, wdyt?

zwarm: @malinajirka Yeah, I just poked around in stats and they seem do similar, but I am going to get rid of the exception, because pressing the button a bazillion times doesn't mean it's wrong. I wonder if what we need is a small delay before kicking off the request to the repo  - since the UI is optimistic? If we are sending these requests async, can we really guarantee the order of execution and even if that, we can't guarantee the order of return on the bus either. I'll need to change a timestamp to the request object to make it unique (mapping request/response)   wdyt?

malinajirka: >can we really guarantee the order of execution

Nope, we can't easily guarantee that. Tbh I wouldn't worry about it. 

What we could theoretically do is show a snackbar when the user initiates opposite (like vs unlike) action while the previous action is still ongoing and ignore the user request. But I'm not sure it's worth it. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderRepositoryEvents.kt
zwarm: Ohh I thought I did make them data classes. In commit 46d91a6

zwarm: I take that back, I originally had them as data classes, then switched them to class so I can use the `PostLikeEnded` event in the callback without having to check for each individual class type. I'll noodle on this to see if I can handle a different way. I like the use of class name instead of a value in the payload. We'll see, but yes, it was intentional not to use a data class

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeActionUseCase.kt
zwarm: >But I'm not sure it's worth it.

I don't think it's worth it. Let's go with the original intent, but remove the exception. If you send a duplicate request when one is running, it gets tossed. thumbs up or thumbs down?

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeActionUseCase.kt
zwarm: Yeah, using a continuation is new to me as well. I believe using the `suspendCancellableCoroutine` gives us a chance to do something like `continuation.invokeOnCancellation { println("ended because of cancel") }`, but not sure if we need that makes a difference here. I am going to change it to `suspendCancellableCoroutine` and do some more research.

zwarm: Okay, switched it out in f0a984a

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderRepositoryEvents.kt
malinajirka: >so I can use the PostLikeEnded event in the callback without having to check for each individual class type

I'm not sure I understand this statement. However, as you said - "We'll see".

PR #12469, opened 2020-07-16
WP Stories: keep state

mzorz:
This PR provides support for Stories serialization / state keeping, particularly used when the StoryComposerActivity is sent to the background so we don't lose users work.

To test:
See steps in https://github.com/Automattic/portkey-android/pull/435

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
```diff
@@ -902,6 +902,9 @@ private void setSite(Intent data) {
     @Override
     public void onActivityResult(int requestCode, int resultCode, Intent data) {
         super.onActivityResult(requestCode, resultCode, data);
+        if (mSelectedSite == null) {
+            initSelectedSite();
+        }
```
jd-alexander, line 907: Hey @mzorz could you give me some more context why this change was necessary for this PR?

mzorz: oh sure thing! Forgot to mention this. When you set the `Don't keep activities` to ON in developer settings, you'll find that just sitting there on the main screen of WPAndroid, sending it to the background and then bringing it to the foreground will do strange things because the `mSelectedSite` is lost. There are safeguards for this same situation elsewhere (just do a search for `initSelectedSite()` and you'll see ;) ).
This is particularly needed if you want to create a story because our flow there is we first start the media picker for result (and at that point the main activity is gone!), as soon as you have chosen something, get back for onActivityResult in the main activity but the selected site is lost so you'll be sitting there 😅 .

PR #12470, opened 2020-07-16
Fix/code freeze localization

loremattei:
This PR updates the `merge_strings_xml.py` script to use double quotes in the XML declaration. 

We have had an issue where this script changed the quotes in the XML declaration from double to single and the next script in the flow changed them back to double quotes. This caused unnecessary commits where the only changes were the quotes and sometimes it caused failures. 

The root cause of the issue is that most XML libraries don't allow the user to select single or double quotes for the XML declaration: they just have a default. Since we use different languages in our flow, the XML libraries are different and they have different standards. 

Since most of our tooling in written in Ruby which defaults to double quotes and all the other `strings.xml` files use double quotes, I chose to fix the issue moving everything to double quotes. 

Unluckily, I couldn't find a clean way to fix it for the reasons above, so I ended up using what seems to be a very common solution which is manually add the header. 
Recent versions of Python allow a slightly better way for this as a `xml_declaration` parameter has been added to `ElementTree.tostring` to get the output without the library generated header, but since we don't use Python a lot and we don't require a specific minimum Python version in our setup, I opted for the slightly more hacky solution of manually replacing the first line of the library output which works with every Python version.

If you have any better solution or suggestion, I'm happy to update this PR with it :-). 
@Tug please let me know if you see any issue/possible side effect with this. 

To test:
- Run `./tools/merge_strings_xml.py` and verify that the header has double quotes. 

PR submission checklist:

### DIFF
--- a/tools/merge_strings_xml.py
+++ b/tools/merge_strings_xml.py
@@ -98,7 +98,10 @@ def merge_strings(main_xml, extra_sections):
         add_section(main_root, insertion_point_index, section_name, new_elements)
     # make sure xml file ends with a newline
     main_root.tail = '\n'
-    xml_output_tree.write(main_xml, encoding='utf-8', xml_declaration=True)
+    xml_string = ('<?xml version="1.0" encoding="UTF-8"?>' + '\n' + 
+        ET.tostring(main_root, encoding='utf8').split('\n',1)[1])
+    with open(main_xml, 'wb') as xml_file:
+        xml_file.write(xml_string)
 
 def main():
     merge_strings(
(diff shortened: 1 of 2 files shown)


tools/merge_strings_xml.py
```diff
@@ -98,7 +98,10 @@ def merge_strings(main_xml, extra_sections):
         add_section(main_root, insertion_point_index, section_name, new_elements)
     # make sure xml file ends with a newline
     main_root.tail = '\n'
-    xml_output_tree.write(main_xml, encoding='utf-8', xml_declaration=True)
+    xml_string = ('<?xml version="1.0" encoding="UTF-8"?>' + '\n' + 
+        ET.tostring(main_root, encoding='utf8').split('\n',1)[1])
+    with open(main_xml, 'wb') as xml_file:
```
Tug, line 103: Not sure about the `wb` file mode here, is it expected to write a simple utf8 string?

Tug: > When writing in binary mode, Python makes no changes to data as it is written to the file

Seems good to me :)

PR #12479, opened 2020-07-20
RI: Create expandable tags view

ashiagr:
Parent Issue: #12028

**Review Instructions**

1. Review PR #12615
2. Make sure PR #12615 is merged to develop
3. Merge develop to this PR
4. Review this PR

**Description**

This PR creates custom expandable tags view for the discover post card.

Custom expandable tags view extends `ChipGroup` widget and adds two types of chips:
- Tag chip (corresponding to each tag)
- Overflow Indicator chip (added to the end of the `ChipGroup` as last child)

Overflow indicator chip is shown by hiding (`View.GONE`) all tag chips outside the visible area of the `ChipGroup`. 

**ChipGroup states**
Collapsed (single line) | Expanded (full height)
--- | ---
collapsed | Screenshot 2020-08-04 at 5 57 50 PM

**Edge conditions**
- [Overflow indicator chip is hidden](https://github.com/wordpress-mobile/WordPress-Android/blob/c546b78ff7979456060ed6c2c7b3940c31d0e799/WordPress/src/main/java/org/wordpress/android/ui/reader/views/ReaderExpandableTagsView.kt#L128-L129) if there are no tag chips outside the visible area of the `ChipGroup`. 

- In the collapsed state, after hiding tag chips outside the visible area of the `ChipGroup`, there might be a case where the overflow indicator chip is still not shown. [Last visible tag chip is hidden](https://github.com/wordpress-mobile/WordPress-Android/blob/c546b78ff7979456060ed6c2c7b3940c31d0e799/WordPress/src/main/java/org/wordpress/android/ui/reader/views/ReaderExpandableTagsView.kt#L121-L125) in this case so that the overflow indicator chip gets shown.
chips

- Chip max width is set to a fraction (.75) of `ChipGroup` width. This is done to fit the overflow indicator chip in the collapsed state (single line) if the first chip title is very large.

- [Condition to find if tag chip is outside bounds](https://github.com/wordpress-mobile/WordPress-Android/blob/a5515b02e0225d5eb6cf2d7273ac3d59e2681765/WordPress/src/main/java/org/wordpress/android/ui/reader/views/ReaderExpandableTagsView.kt#L110-L118) is based on RTL and LTR layout direction of the `ChipGroup`.

**To test**

Note: Actual post tags retrieved in PR #12615 and set on expandable tag view in 63c9d8d

Prerequisite: RI FF is enabled

- Launch app (in portrait mode) and go to Reader tab
- Notice that 
    - tags are visible on the post card (only on discover tab)
    - correct tags (corresponding to post) are displayed
    - overflow chip is shown if tags cannot fit in the collapsed state
    - overflow chip displays correct no. of chips that are hidden
- Click overflow chip
- Notice that layout expands with a hide all option
- Click hide all option
- Notice that layout collapses, displaying an overflow indicator
- Switch to landscape mode
- Notice that expandable view adjusts to new width
- Click a tag 
- Notice that topic details page is opened for selected tag

**Note**
~~Used [requestLayout()](https://github.com/wordpress-mobile/WordPress-Android/blob/c546b78ff7979456060ed6c2c7b3940c31d0e799/WordPress/src/main/java/org/wordpress/android/ui/reader/views/ReaderExpandableTagsView.kt#L102) to get chip coordinates inside `isChipWithinBounds` else those are returned zero. I tried to override `FlowLayout.onMeasure` (`FlowLayout` is `ChipGroup` parent) but it can only be called from within the same library group (`com.google.android.material`). Let me know if you have suggestions to improve.~~

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostUiStateBuilder.kt
```diff
@@ -122,6 +125,10 @@ class ReaderPostUiStateBuilder @Inject constructor(
             post.takeIf { post.cardType == VIDEO }
                     ?.let { post.featuredVideo }
 
+    private fun buildExpandedTagsViewVisibility(post: ReaderPost) = post.isDiscoverPost
+
+    private fun buildTags(post: ReaderPost) = ReaderTagList() // TODO: ashiagr - get tags from post
```
ashiagr, line 130: Topics on the card are currently set to empty tags list and needs to set from the post.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostUiStateBuilder.kt
```diff
@@ -122,6 +125,10 @@ class ReaderPostUiStateBuilder @Inject constructor(
             post.takeIf { post.cardType == VIDEO }
                     ?.let { post.featuredVideo }
 
+    private fun buildExpandedTagsViewVisibility(post: ReaderPost) = post.isDiscoverPost
```
ashiagr, line 128: I think I'll have to add a check for RI FF here so that tags do not display on develop. Will verify and add.

malinajirka: Yep, I think you are correct.

Btw the new discover tab doesn't work with "real discover posts" -> "isDiscoverPost" will return false for them. "Real Discover post" is a post from "discover.wordpress.com" and the new feed will contain posts from various blogs.

WordPress/src/main/java/org/wordpress/android/ui/reader/views/ReaderExpandableTagsView.kt
```diff
@@ -0,0 +1,154 @@
...
+import com.google.android.material.chip.ChipGroup
+import org.wordpress.android.R
+import org.wordpress.android.WordPress
+import org.wordpress.android.models.ReaderTag
+import org.wordpress.android.models.ReaderTagList
+import org.wordpress.android.ui.utils.UiHelpers
+import org.wordpress.android.util.DisplayUtils
+import javax.inject.Inject
+
+class ReaderExpandableTagsView @JvmOverloads constructor(
```
malinajirka, line 18: Great job on this class!!! I can't think of a better approach than using `addOnPreDrawListener`. So I'd go for it. 

The only thing I'd consider is to move the logic into a testable helper class - some things could be perhaps moved even into a uiState, but I'm not 100% it's necessary/desired.

malinajirka:
I haven't done a full review, but I like the approach overall ;)! GJ

WordPress/src/main/java/org/wordpress/android/ui/reader/views/ReaderExpandableTagsView.kt
ashiagr: > move the logic into a testable helper class
Agree. I was thinking about how to make it testable. VM wasn't a right choice. I'll consider using a helper class.

ashiagr: Adde todo note [here](https://github.com/orgs/wordpress-mobile/projects/63#card-43363060)

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostUiStateBuilder.kt
ashiagr: Done here: 63c9d8d

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostUiStateBuilder.kt
ashiagr: Updated here: 25fa018

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostUiStateBuilder.kt
```diff
@@ -43,11 +44,14 @@ class ReaderPostUiStateBuilder @Inject constructor(
...
     private val readerUtilsWrapper: ReaderUtilsWrapper,
-    private val readerPostMoreButtonUiStateBuilder: ReaderPostMoreButtonUiStateBuilder
+    private val readerPostMoreButtonUiStateBuilder: ReaderPostMoreButtonUiStateBuilder,
+    private val readerPostTagsUiStateBuilder: ReaderPostTagsUiStateBuilder,
+    private val appPrefsWrapper: AppPrefsWrapper
 ) {
     // TODO malinjir move this to a bg thread
     fun mapPostToUiState(
         post: ReaderPost,
+        isDiscover: Boolean = false, // set to true for new discover tab
```
zwarm, line 54: @ashiagr - Does the future of `ReaderPostUiStateBuilder` make it possible to use on other tabs? Following, Saved, Likes, Automatic, etc. If so, I wonder if there will be any future logic directly related to what tab is shown and if so, perhaps implementing something other than a boolean may be an option. Feel free to ignore my suggestion, as boolean serves it purpose perfectly for our focus.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
```diff
@@ -168,10 +168,10 @@ class ReaderInterestsViewModel @Inject constructor(
 
     private fun transformToInterestsUiState(interests: ReaderTagList) =
```
zwarm, line 169: Perhaps change the reference/methods names so that it no longer reflects the deleted `InterestUiState` class? Not needed, but may be easier for future us to understand later. wdyt?

ashiagr: I have to rename these variables and classes to avoid tag/topic/interest confusion. Will do in a separate PR. Added note here: https://github.com/orgs/wordpress-mobile/projects/63#card-43573240

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostUiStateBuilder.kt
ashiagr: We might want to convert it to tab name. We'll see how it turns out. Thanks for pointing it out :).

Issue #12411, opened 2020-07-10
Bottom sheet : story title cursor disappears after typing


jd-alexander:
## Expected behavior
The cursor should remain as you type. 

## Actual behavior
The cursor disappears once you start typing. 

### Steps to reproduce the behavior
1. Create a story post. 
2. Click Publish. 
3. Start typing the story title. 

Tested on Google Pixel 4 XL, Android 10, WPAndroid `wp stories alpha` ()

jd-alexander:
Resolved in https://github.com/wordpress-mobile/WordPress-Android/pull/12486

jd-alexander:
Resolved here https://github.com/wordpress-mobile/WordPress-Android/pull/12520

Issue #12334, opened 2020-06-30
WordPress Stories : Story Title UI & Keyboard Issues


jd-alexander:
- [x] Currently, the Story Title is not persisted as the Post Title
- [x] The story title is not being shown on the bottom sheet when the user has set it. 
- [x] The story title disappears when the user navigates to a different section of the bottom sheet
- [x] The story title should have a default value of `Untitled` once the user doesn't set a title.
- [x] The keyboard should be shown when the bottom sheet is presented.
- [x] Ensure that the keyboard gets dismissed appropriately when the bottom sheet is dismissed etc.

mzorz:
Just coming here to keep things documented - I'm documenting here some possibilities explored and the result of analyzing the code and related PRs.

As mentioned here https://github.com/wordpress-mobile/WordPress-Android/pull/12486

> We are currently using a style that makes the window full screen but that interferes with the keyboard's ability to adjust the view. i.e adjustResize and the fullScreen styles don't play well together.

Already discussed this elsewhere but wanted to leave my findings written here I've been trying a few things re: `adjustResize` and `windowFullScreen` being in combination as a possible cause of the problem, and was able to zero that out.

- For example, if we force the `EditPostActivity` (which uses `adjustResize` and is not `windowFullScreen`) to open the bottom sheet for the story format (which contains the story_title `EditText` component), you'll see we have the same problem about the keyboard overlapping the bottom sheet (see [this gist to try](https://gist.github.com/mzorz/aad7245e7b5a2bfbb7a99bedc2ee19c7), also I uploaded a video here https://cloudup.com/c9b6YwnYO7G).

- Just in case the EditPostActivity was introducing some more noise (maybe also because we have other objects such as EditText->Aztec that grab focus and show the keyboard here), I made yet another test [with this gist here](https://gist.github.com/mzorz/bfd6505d1f5a9aab1729d2d721ac700a), which just shows the prepublishing bottom sheet when you tap the FAB on WPMainActivity, also added `adjustResize` attribute to WPMainActivity in the manifest, and still can see the overlapping problem there.

[...]

PR #12486, opened 2020-07-21
Issue 12407/story title container focus

### LINKED ISSUE #12407: WordPress Stories : Tap on story title container for input focus
### Expected behavior
Expected to be able to tap anywhere in the story title area to gain focus on the input control.

### Actual behavior
Have to tap exactly on the ‘Give your story a title’ text to enter a title. 

1. Create a new story post. 
2. Click Publish. 
3. Click on an area outside the hint/ edit text control. 
4. See that the control does not start accepting the user input via the keyboard.

Tested on Google Pixel 4 XL, Android 10, WPAndroid `wp stories alpha` ()

jd-alexander:
Fixes #12407 #12411

## Solution
The main focus of this PR is to fix all the issues that are related to the keyboard and the story title's `EditText` focus. 

1. You can now click on anywhere within the story title's layout to trigger the keyboard and focus c03e2d5
2. The keyboard now opens up when the bottom sheet is opened. 

## Testing

1. Create a story post. 
2. Click Next. 
3. See the bottom sheet open with the keyboard active and the story title focused.

<kbd></kbd>

-------

1. Create a story post. 
2. Click Next. 
3. See the bottom sheet open with the keyboard active and the story title focused.
4. Close the keyboard. 
5. Click anywhere within the story title layout like below.

<kbd></kbd>

------- 
## Concerns

The issue I am having with this solution, which is probably unrelated to these changes but more related to this `StoryComposerActivity` is the way how we handle the window. We are currently using a style that makes the window full screen but that interferes with the keyboard's ability to adjust the view. i.e `adjustResize` and the `fullScreen` styles don't play well together. 

This is main part of the style that is doing this

```
        <item name="android:immersive">true</item>
        <item name="android:windowFullscreen">true</item>
```

So the issue that's taking place now is since we can't utilize the `android:windowSoftInputMode` everything is being done manually. We even have logic to [adjust the view when the keyboard is opened](https://github.com/wordpress-mobile/WordPress-Android/blob/feature/wp-stories/alpha2/WordPress/src/main/java/org/wordpress/android/util/KeyboardResizeViewUtil.java). 

The bug that's happening now is when we navigate to the tags option we are getting an [IllegalArgumentException](https://stackoverflow.com/questions/7100555/preventing-catching-illegalargumentexception-parameter-must-be-a-descendant-of/12888761#12888761).

<kbd></kbd>

```
 java.lang.IllegalArgumentException: parameter must be a descendant of this view
        at android.view.ViewGroup.offsetRectBetweenParentAndChild(ViewGroup.java:6143)
        at android.view.ViewGroup.offsetDescendantRectToMyCoords(ViewGroup.java:6072)
        at android.view.ViewRootImpl.scrollToRectOrFocus(ViewRootImpl.java:3920)
        at android.view.ViewRootImpl.draw(ViewRootImpl.java:3493)
        at android.view.ViewRootImpl.performDraw(ViewRootImpl.java:3418)
        at android.view.ViewRootImpl.performTraversals(ViewRootImpl.java:2755)
        at android.view.ViewRootImpl.doTraversal(ViewRootImpl.java:1721)
        at android.view.ViewRootImpl$TraversalRunnable.run(ViewRootImpl.java:7598)
```

### At first I thought the below points were causing the issue : 

So far based on my research it seems we there's some inconsistency taking place with the `EditText` on the `TagsFragment` and how it receives focus because sometimes the bug doesn't present itself, while other times it causes a crash. 

I have tried. 

* Clearing the focus of the `EditText`
* Clearing the full-screen mode window flag programmatically. 

These didn't work so looking to hear your thoughts.

### Approach after validating the above doesn't matter by commenting out that code

So it seems, the issue could be that we have the keyboard open while changing fragments, so that could be causing the exception to occur. Still investigating.

An idea I am now playing with to move this functionality forward is if we would focus the `EditText` on the `TagsFragment` but close the keyboard before doing the navigation. Only caveat of this would be the would have to click the keyboard open, tapping the edit text. 

## Reviewing 

## Submitter Checklist


WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeViewHolder.kt
```diff
@@ -81,8 +83,17 @@ sealed class PrepublishingHomeViewHolder(
...
+            }
+
+            storyTitleLayout.setOnClickListener {
+                focusStoryTitleEditTextAndShowKeyboard()
+            }
+
             storyTitle.postDelayed({
-                storyTitle.requestFocus()
+                focusStoryTitleEditTextAndShowKeyboard()
             }, STORY_TITLE_EDIT_TEXT_REQUEST_FOCUS_DELAY)
```
mzorz, line 97: There's now a chained delay of `STORY_TITLE_EDIT_TEXT_REQUEST_FOCUS_DELAY` here: 
1. to call focusStoryTitleEditTextAndShowKeyboard, at this point the focus and keyboard are requested
2. another one to requestFocus (that acttually has the effect of getting the keyboad up).

This is weird in reality because tapping on the NEXT button on the Story composer makes a sequence of weird movements instead of everything happening all at once (or at least closely together). Also, `STORY_TITLE_EDIT_TEXT_REQUEST_FOCUS_DELAY` is 1000 millis, probably 2 seconds is too long. 

Here's a video showing this https://cloudup.com/c-wQOeSvHE1

I'm going to check this in a deeper way and see what I can find

jd-alexander: Thanks! I didn't see that delay initially. I will check this out as well to see.

jd-alexander:
We are closing this PR, since we are going to be taking another approach at addressing this issue. A linked issue with more details will follow shortly!

jd-alexander:
A continuation of these efforts will be linked to this issue https://github.com/wordpress-mobile/WordPress-Android/issues/12334

jd-alexander:
Associated PR here https://github.com/wordpress-mobile/WordPress-Android/issues/12334

PR #12490, opened 2020-07-21
RI: Fetch "Interests" tags from ReaderTagRepository

ashiagr:
This PR 
- Updates `ReaderTagRepository` to return Interest API results to the Interests VM utilising `ReaderRepositoryCommunication` and `FetchInterestTagsUseCase`
- Handles error

~~Looking for an early feedback before moving ahead, so marking it as draft.~~

### To Test

Prerequisite - RI FF flag s enabled

- Launch and go to Reader tab
- Interests from server are shown 

----

- Enable Airplane mode
- Launch and go to Reader tab
- Error layout is displayed with Retry option
- Disable Airplane mode
- Click Retry
- Error layout is hidden and interests are shown

To test remote request failure error, comment out `SuccessWithData` response and return `RemoteRequestFailure` [here](https://github.com/wordpress-mobile/WordPress-Android/blob/f16419db09700f8e9a8985c3944c3a0a662107d7/WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/FetchInterestTagsUseCase.kt#L52).

----

- While the interests are shown, go to My Site -> Me -> App Settings and change interface language (e.g. Spanish)
- Return to Reader tab
- Notice that interests are shown in the selected language (returned from the server).

### ToDo
- Clicking "Comics" interest is not highlighting it 🤷‍♀️, although it enables Done button. Looking into it. 
UPDATE: There are duplicate tags with "Comics" title in the feed. There shouldn't be such a case in the final feed. I'll put a "unique" check to fix it. - 2843b0f
- ~~I'm looking for a better solution for language change support. I've removed `isStarted` check so that we can re-request interests from the server for the changed language but it now happens every time we toggle bottom tabs or rotate screen. I'm thinking to compare last and new language in the interests VM for this something like conditional re-request of tags [here](https://github.com/wordpress-mobile/WordPress-Android/blob/f16419db09700f8e9a8985c3944c3a0a662107d7/WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt#L76).~~ - 7ac80c2
- Do we want to show "Contact Support" on remote request failure? We haven't yet decided on the error layout design.
- Tests for `FetchInterestTagsUseCase` not yet added. Will include along with other use cases to save interests and get user tags.

---

device-2020-07-21-190416

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/FetchInterestTagsUseCase.kt
```diff
@@ -0,0 +1,54 @@
...
+            throw IllegalStateException("Fetch already in progress.")
+        }
+        return suspendCoroutine { cont ->
+            continuation = cont
+            eventBusWrapper.register(this)
+
+            ReaderUpdateServiceStarter.startService(
+                contextProvider.getContext(),
+                EnumSet.of(INTEREST_TAGS)
+            )
```
ashiagr, line 36: I'm registering to event bus inside the use case, I've not seen this practice elsewhere in the code. Let me know if you see any downside.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderRepositoryCommunication.kt
```diff
@@ -2,6 +2,7 @@ package org.wordpress.android.ui.reader.repository
 
 sealed class ReaderRepositoryCommunication {
     object Success : ReaderRepositoryCommunication()
+    data class SuccessWithData<out T>(val data: T) : ReaderRepositoryCommunication()
```
ashiagr, line 5: Passing interests directly to the VM via `SuccessWithData`.  We might want to keep a single `Success` state with an optional data param. Lmkwyt.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderEvents.java
```diff
@@ -43,16 +43,22 @@ public String getTagName() {
     public static class RecommendedBlogsChanged {
     }
 
-    public static class InterestTagsFetched {
+    public static class InterestTagsFetchEnded {
         private final ReaderTagList mInterestTags;
+        private final boolean mDidSucceed;
```
ashiagr, line 48: Added boolean to detect if request succeeded similar to `SearchPostsEnded`

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/FetchInterestTagsUseCase.kt
zwarm: @ashiagr - There is similar logic in #12454, but I used init & stop to set up the bus to support multiple requests at the same time. Not sure if you'll need to do that here, but wanted to mention it. Also, in the same PR, I have a `ReaderRepositoryDispatchingUseCase` that I use for all the repo usecases to set up coroutine context.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderRepositoryCommunication.kt
zwarm: @ashiagr - I like this idea for those non-reactive repository situations and I can go with either option. In one situation we check for the class and the other a quick optional check.

ashiagr: Looks like we'll have to keep just one Success state to avoid `else` condition in `when` clause as seen [here](https://github.com/wordpress-mobile/WordPress-Android/pull/12490/commits/baab29d148ea34a32d9390cdc915c19aa48347d2#diff-8f7fbeb1ff107a11e53a17c2518cd0cfR68-R70).

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/FetchInterestTagsUseCase.kt
ashiagr: @malinajirka what do you think? Should we use `init` & `stop` pattern in this case? I'll make changes accordingly.

ashiagr:
@zwarm , @malinajirka - ready for another look 🙇‍♀️.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/FetchInterestTagsUseCase.kt
malinajirka: Since this useCase doesn't support parallel request by its nature, I think the current implementation looks good. I don't think we need to change anything.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/FetchInterestTagsUseCase.kt
```diff
@@ -0,0 +1,60 @@
...
+        }
+    }
+
+    @Subscribe(threadMode = ThreadMode.MAIN)
```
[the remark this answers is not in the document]
ashiagr: Switched to BACKGROUND thread in ad7c6b0.

PR #12493, opened 2020-07-21
Issue 12492/story composer viewmodel part 1

### LINKED ISSUE #12492: WordPress Stories : Create StoryComposerViewModel for StoryComposerActivity
reference https://github.com/Automattic/portkey-android/issues/438
### Expected behavior
StoryComposerActivity should have an associated ViewModel with it's business logic. 

### Actual behavior
StoryComposerActivity has the business logic within it so it needs to be extracted to a ViewModel.

jd-alexander:
Fixes #12492

## Solution

This is the first set of changes to the `StoryComposerActivity` that primarily involves moving the business logic that was within the `StoryComposerActivity`'s `onCreate()` to the `StoryComposerViewModel` 's `start()` method. 

- a `StoryComposerViewModel` has been created to host the business logic. 
- `saveIntialPost` has been converted to a use case and moved to the `StoryComposerViewModel`
- `SaveInitialPostUseCase` now has unit tests to verify it's behavior. These tests are located in `SaveInitialPostUseCaseTest `.
- the logic related to the `savedInstanceState` being null or not was moved into the `ViewModel`. All other operations where data was being fetched from the intent or bundle were left in the `StoryComposerActivity` and then the necessary values were passed to the `start()` method of the `ViewModel`. 
- The `AnalyticsUtils` usage was added to the `AnalyticsUtilsWrapper` so that the component could be injected into the `ViewModel` during test.
- Saving state to the bundle is now handled in the `ViewModel` since, it's business logic that determines what should be saved 40d7d20

## Testing

1. Create a new story post. 
2. Add slides to it. 
3. Publish the story. 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/stories/SaveInitialPostUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stories/SaveInitialPostUseCase.kt
@@ -5,6 +5,7 @@ import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.post.PostStatus
 import org.wordpress.android.fluxc.store.PostStore
 import org.wordpress.android.ui.posts.EditPostRepository
+import org.wordpress.android.ui.posts.EditPostRepository.UpdatePostResult
 import org.wordpress.android.ui.posts.SavePostToDbUseCase
 import javax.inject.Inject
 
@@ -20,9 +21,15 @@ class SaveInitialPostUseCase @Inject constructor(
         }
         editPostRepository.savePostSnapshot()
         // this is an artifact to be able to call savePostToDb()
-        editPostRepository.getEditablePost()?.setPostFormat(StoryComposerActivity.POST_FORMAT_WP_STORY_KEY)
-        site?.let {
-            savePostToDbUseCase.savePostToDb(editPostRepository, it)
-        }
+        editPostRepository.updateAsync({ postModel ->
+            postModel.setPostFormat(StoryComposerActivity.POST_FORMAT_WP_STORY_KEY)
+            true
+        }, { _, result ->
+            if (result == UpdatePostResult.Updated) {
+                site?.let {
+                    savePostToDbUseCase.savePostToDb(editPostRepository, it)
+                }
+            }
+        })
     }
 }
--- a/WordPress/src/main/java/org/wordpress/android/ui/stories/SaveInitialPostUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stories/SaveInitialPostUseCase.kt
@@ -12,7 +12,7 @@ class SaveInitialPostUseCase @Inject constructor(
     val postStore: PostStore,
     val savePostToDbUseCase: SavePostToDbUseCase
 ) {
-     fun saveInitialPost(editPostRepository: EditPostRepository, site: SiteModel?) {
+    fun saveInitialPost(editPostRepository: EditPostRepository, site: SiteModel?) {
         editPostRepository.set {
             val post: PostModel = postStore.instantiatePostModel(site, false, null, null)
             post.setStatus(PostStatus.DRAFT.toString())
--- a/WordPress/src/main/java/org/wordpress/android/ui/stories/SaveInitialPostUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stories/SaveInitialPostUseCase.kt
@@ -1,0 +1,28 @@
+package org.wordpress.android.ui.stories
+
+import org.wordpress.android.fluxc.model.PostModel
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.model.post.PostStatus
+import org.wordpress.android.fluxc.store.PostStore
+import org.wordpress.android.ui.posts.EditPostRepository
+import org.wordpress.android.ui.posts.SavePostToDbUseCase
+import javax.inject.Inject
+
+class SaveInitialPostUseCase @Inject constructor(
+    val postStore: PostStore,
+    val savePostToDbUseCase: SavePostToDbUseCase
+) {
+     fun saveInitialPost(editPostRepository: EditPostRepository, site: SiteModel?) {
+        editPostRepository.set {
+            val post: PostModel = postStore.instantiatePostModel(site, false, null, null)
+            post.setStatus(PostStatus.DRAFT.toString())
+            post
+        }
+        editPostRepository.savePostSnapshot()
+        // this is an artifact to be able to call savePostToDb()
+        editPostRepository.getEditablePost()?.setPostFormat(StoryComposerActivity.POST_FORMAT_WP_STORY_KEY)
+        site?.let {
+            savePostToDbUseCase.savePostToDb(editPostRepository, it)
+        }
+    }
+}
--- a/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
@@ -1,8 +1,14 @@
 package org.wordpress.android.ui.stories
 
+import android.net.Uri
 import android.os.Bundle
+import android.webkit.URLUtil
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.LifecycleRegistry
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.Observer
 import androidx.lifecycle.ViewModel
 import org.wordpress.android.WordPress
 import org.wordpress.android.fluxc.Dispatcher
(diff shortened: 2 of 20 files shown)


WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
```diff
@@ -0,0 +1,72 @@
...
+    private val analyticsUtilsWrapper: AnalyticsUtilsWrapper
+) :
+        ViewModel() {
+    private lateinit var editPostRepository: EditPostRepository
+    private lateinit var site: SiteModel
+    private lateinit var postEditorAnalyticsSession: PostEditorAnalyticsSession
+
+    fun start(
+        editPostRepository: EditPostRepository,
+        postId: LocalId,
```
jd-alexander, line 27: For the `postId` I am using the `LocalId` wrapper class so that it's easier to identify the id as being a local one versus remote.

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
```diff
@@ -0,0 +1,72 @@
...
+            )
+        } else {
+            editPostRepository.loadPostByLocalPostId(postId.value)
+        }
+
+        setupPostEditorAnalyticsSession(postEditorAnalyticsSession)
+
+        notificationType?.let {
+            systemNotificationsTracker.trackTappedNotification(it)
+        }
```
jd-alexander, line 52: I decided to put this in the `ViewModel` because it is business logic. We are saying that if the `notificationType` isn't null then we should track it since once it's not null it means we are resuming the `StoryComposerActivity` from a notification.

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
```diff
@@ -0,0 +1,97 @@
...
+
+    fun onStoryDiscarded() {
+        // delete empty post from database
+        dispatcher.dispatch(PostActionBuilder.newRemovePostAction(editPostRepository.getEditablePost()))
+        postEditorAnalyticsSession.setOutcome(CANCEL)
+    }
+
+    override fun onCleared() {
+        super.onCleared()
+        postEditorAnalyticsSession.end()
```
jd-alexander, line 95: I moved this from the `onDestroy` of the `StoryComposerActivity` I am just putting this comment here to verify that this will behave correctly with the `ViewModel`s lifecycle. I will investigate to verify.

jd-alexander: I checked this out and using `onCleared()` is fine because the lifecycle of the `ViewModel` is scoped to the component that it gets attached to. In this case, the `onCleared()` is directly associated with the activity's `onDestroy()`.

jd-alexander:
@malinajirka you don't have to do a full in-depth review of this and the other two PRs I am tagging you in 😄 I just want your thoughts on this refactor process that I am doing. I am basically extracting the logic in the `StoryComposerActivity` to `ViewModel` and `usecases`. The main questions I had in mind while doing this refactor are: 

1. Should an `intent` be passed to the `ViewModel`? This is being done because an `AnalyticsUtil` method that does tracking needs it. 

2. Also, we are utilizing the `EditorMedia` class from the `EditPostActivity`. I left it in the activity since it feels like a component that acts like a `ViewModel` since it has `uiState` that can be observed and such forth. Eg, I have logic that looks like this still in the `StoryComposerActivity`

```
    private fun handleMediaPickerIntentData(data: Intent) {
        // TODO move this to EditorMedia
        val ids = ListUtils.fromLongArray(
                data.getLongArrayExtra(
                        MediaBrowserActivity.RESULT_IDS
                )
        )
        if (ids == null || ids.size == 0) {
            return
        }

        editorMedia.addExistingMediaToEditorAsync(WP_MEDIA_LIBRARY, ids)
    }
```
So I wasn't sure if I should pass those ids to the `ViewModel` and then let it trigger a `LiveData` to update `editorMedia`. 

Let me know what you think 🙏

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
malinajirka: Just a minor note: This change will change the behavior a bit. When you for example rotate your device "onDestroy" would have been called and session end tracked. But now, it gets tracked only when you leave the screen. I guess that's what we want anyway, but I thought I'd better mention it.

WordPress/src/main/java/org/wordpress/android/ui/stories/SaveInitialPostUseCase.kt
```diff
@@ -0,0 +1,28 @@
...
+) {
+    fun saveInitialPost(editPostRepository: EditPostRepository, site: SiteModel?) {
+        editPostRepository.set {
+            val post: PostModel = postStore.instantiatePostModel(site, false, null, null)
+            post.setStatus(PostStatus.DRAFT.toString())
+            post
+        }
+        editPostRepository.savePostSnapshot()
+        // this is an artifact to be able to call savePostToDb()
+        editPostRepository.getEditablePost()?.setPostFormat(StoryComposerActivity.POST_FORMAT_WP_STORY_KEY)
```
malinajirka, line 23: In general, it's preferred to work only with immutable posts and use `update/updateAsync` methods for modifications. Let me know if it'd work for you.

malinajirka:
Really a great job @jd-alexander! I've scanned through the changes and I'm really glad to see such improvements!

>Should an intent be passed to the ViewModel? This is being done because an AnalyticsUtil method that does tracking needs it.

This is actually quite a complicated topic. My personal opinion is that ideally it shouldn't. However, if you are still able to write unit tests for the methods, I wouldn't worry about it too much (I'm not sure if we can mock/create intents in a unit test environment). 
If Intents can't be used in unit tests, I'd emit an Event from the VM and do the actually tracking in the fragment (similar to how we show snackbars, do navigation etc).


>Also, we are utilizing the EditorMedia class from the EditPostActivity. I left it in the activity since it feels like a component that acts like a ViewModel since it has uiState that can be observed and such forth. Eg, I have logic that looks like this still in the StoryComposerActivity

Great question! I would personally vote against using EditorMedia within this activity/vm. EditorMedia was created as a VM for the EditPostActivity (it isn't actual VM since we didn't get that far with the refactoring). Since we usually don't re-use VMs on other screens I wouldn't reuse the EditorMedia either. What we can do instead is inject `addExistingMediaToPostUseCase` into our story vm and use it. UseCases are re-usable components which can be used throughout the app. If you realize you need more than just that, I think we might want to extract it from the EditorMedia into a new class and use that class in both activities. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
mzorz: Good notes there!  just leaving another note here as well 😅 : given the StoryComposerActivity is fixed to portrait for product design reasons, rotation in particular wouldn't have affected it  - of course the Activity would still be re-created on other configuration changes still 👍 so, good call moving it to the ViewModel's `onClear` so the session is kept until the user actually goes away from the Activity

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
mzorz: 👍

mzorz:
> If Intents can't be used in unit tests, I'd emit an Event from the VM and do the actually tracking in the fragment (similar to how we show snackbars, do navigation etc).

Ah nice! I can see how that would work - thanks for the suggestion 🙇 

> I would personally vote against using EditorMedia within this activity/vm. EditorMedia was created as a VM for the EditPostActivity (it isn't actual VM since we didn't get that far with the refactoring). Since we usually don't re-use VMs on other screens I wouldn't reuse the EditorMedia either.

I think I understand the reasoning (we don't reuse VMs; we reuse UseCases), but I was wondering if EditorMedia can be repurposed given it has the same style and sequencing of operation against the same kind of objects. EditorMedia is sort of a container for business logic that operates on Media/Post objects from mandates coming from the UI, but the rules of operation actually come from how a Post / Media is structured rather than by the way the user will operate on the screen. We're going to probably end up having duplicated code given the business logic for operating on these data types from an editor is very similar if not equal (i..e create a Post, add media to it - in fact this can be seen in the current implementation). So, wondering if this could be seen as a particular case where the screen / UI may change (EditPostActivity, StoryComposerActivity) but the VM is essentially the same. There's some precedent in this logic with shared [ViewModel instances between fragments](https://developer.android.com/topic/libraries/architecture/viewmodel#sharing), but even here the ViewModelStoreOwner would be different in each case which is perfectly fine. As for testability, if we see the logic departs from one another at some point then it of course will make sense to write that logic in a different VM and add tests for the whatever new cases appear, but I'm having trouble seeing that departure of functionality right now; does that make sense?

Looking forward to learning what your thoughts are @malinajirka @jd-alexander 🙏

WordPress/src/main/java/org/wordpress/android/ui/stories/SaveInitialPostUseCase.kt
jd-alexander: Sure that works! I refactored this. Updated here 9bed509

jd-alexander:
> Really a great job @jd-alexander! I've scanned through the changes and I'm really glad to see such improvements!

Awesome! Thanks @malinajirka  🙇 

> This is actually quite a complicated topic. My personal opinion is that ideally it shouldn't. However, if you are still able to write unit tests for the methods, I wouldn't worry about it too much (I'm not sure if we can mock/create intents in a unit test environment).
If Intents can't be used in unit tests, I'd emit an Event from the VM and do the actually tracking in the fragment (similar to how we show snackbars, do navigation etc).

Nice! This is a good idea! Thanks! 

> Great question! I would personally vote against using EditorMedia within this activity/vm. EditorMedia was created as a VM for the EditPostActivity (it isn't actual VM since we didn't get that far with the refactoring). Since we usually don't re-use VMs on other screens I wouldn't reuse the EditorMedia either. What we can do instead is inject addExistingMediaToPostUseCase into our story vm and use it. UseCases are re-usable components which can be used throughout the app. If you realize you need more than just that, I think we might want to extract it from the EditorMedia into a new class and use that class in both activities. Wdyt?

I totally understand this! I think we can track this for now because the `EditorMedia` is properly tested and works well with our current implementation. I would refactor this now but there are other areas that could be helped with testing so I would want to focus there a bit more. Wdyt? 

> Looking forward to learning what your thoughts are @malinajirka @jd-alexander 🙏

Thanks for these thoughts @mzorz in principle I know we could refactor the logic within `EditorMedia` or the respective use cases to support the `StoryComposerActivity` / `StoryComposerViewModel` but in the interest of time, as mentioned above we might want to tackle this later. I am happy we are having this discussion, so we have a good baseline moving forward as to what components are actually reusable and which components are designed for single use.

malinajirka:
Thank you for your comments ;)! 

This is a though one:P. If we were using all the methods on the EditorMedia on both screens I guess it'd make a lot of sense to consider reusing the whole "vm" (I still wouldn't be 100% sure about it). However, we are using just a couple of methods but the Story classes still have access to the rest even though they are completely unrelated. Moreover, I feel like the EditorMedia contains mostly UI logic and the actual logic is encapsulated in the UseCases. For example the `addExistingMediaToEditorAsync` doesn't do anything else but starts a useCase on a bg thread.

The fact that two screens need to show a progress indicator when a media is being uploaded/added or that they need to show a snackbar message on error isn't imo good enough reason to reuse the vm - it's a very similar to making a decision when/where to use an inheritance. 

Another thing is that EditorMedia isn't a cleanly written class. It's a class that was created during the EditPostActivity refactoring and we had to fit into the timeline. Re-using a class with a tech-debt on a new screen technically just adds more tech debt into the project.

>There's some precedent in this logic with shared ViewModel instances between fragments

I think that's a bit different use case. When you for example have a site creation wizard. Each step can have it's own VM but their parent can have a VM as well and it's completely fine to share this parent VM between "steps VMs". 

>if we see the logic departs from one another at some point then it of course will make sense to write that logic in a different VM

Imagine someone in the future decides that they want to modify something just on one of the screens. Is can be as simple change as showing a different snackbar message on error. I'm concerned that they won't have time to rewrite all this into a different VM. What they'll most probably do is they'll add "if(EditPostActivity) "do this" else "do that". As this will keep happening we might end up with another unmaintainable class no-one understands.

So imo if we think some of the code should be re-used because it contains logic which shouldn't be duplicated. I think the best approach would be to extract that logic into another class and re-use that class.

This is of course all just my opinion + only theory. Let me know what you think :).

mzorz:
Thanks a lot for these insights to your point of view @malinajirka , it's really enjoyable to have this conversation in lieu of better code 🙌 

You made all great points there - I went ahead and wrote some follow up PRs to extract a `StoryEditorMedia` class first in https://github.com/wordpress-mobile/WordPress-Android/pull/12516, and then another PR that makes that class an actual ViewModel in https://github.com/wordpress-mobile/WordPress-Android/pull/12518 🙇

PR #12496, opened 2020-07-21
Replace temporary GravatarUtils from the Login library with the updated one from the Utils library

renanferrari:
Depends on #12494

During the UL&S project, we introduced a temporary copy of the `GravatarUtils` class from the Utils library to the Login library, in order to quickly overcome some limitations with some of its methods.

This class has now been updated on the Utils library and this PR replaces the temporary one with it.

To test:
- Navigate through both the login and signup flows and make sure the account's Gravatar loads correctly on all the screens where it is used (e.g. Magic Link screen, Password screen, etc). 
- Make sure it also loads correctly on both Login and Signup Epilogue screens.

Notes:
- Use a device with API 21 and 28 to test.
- Avoid using an account with no sites to test, as it might make the Post-Signup Interstitial to be shown instead of the Epilogue.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginHeaderViewHolder.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginHeaderViewHolder.java
@@ -11,11 +11,11 @@ import androidx.recyclerview.widget.RecyclerView;
 import org.wordpress.android.R;
 import org.wordpress.android.fluxc.model.AccountModel;
 import org.wordpress.android.fluxc.model.SiteModel;
-import org.wordpress.android.login.util.AvatarHelper.FakeGravatarUtils;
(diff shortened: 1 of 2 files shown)


renanferrari:
Hey @khaykov, thanks a lot for reviewing this! 🙇‍♂️

>I got the following exception in logcat

After some investigation into this, I found that this problem was already present before this PR (and before the UL&S even). I haven't looked into the cause yet — the exception mentions the event name doesn't match the regex, but at first sight I couldn't figure out why. I'll create a new issue for this and handle it separately.

>sign-up magic link email progress dialog is not dismissing automatically for me.

I couldn't reproduce this problem on my end. I've tested with a Pixel 3 XL and a Motorola One (both on Android 10) and an emulator with Android 5.0. Do you mind creating an issue for this one with some more details?

PR #12501, opened 2020-07-22
Issue 12492/story composer viewmodel part2

### LINKED ISSUE #12492: WordPress Stories : Create StoryComposerViewModel for StoryComposerActivity

jd-alexander:
Fixes #12492 

## Solution
This PR is a continuation of https://github.com/wordpress-mobile/WordPress-Android/pull/12493 where we are moving the logic from the `StoryComposerActivity` to the `StoryComposerViewModel`.

* We have observed the `LiveData` objects in the `ViewModel` for changes so that we can update the respective components in the activity. 
* `appendMediaFiles()` logic was moved to the `ViewModel` and a `LiveData` object triggers adding the media uris to the story. 
* `onStorySaveButtonPressed()` no longer opens the bottom sheet directly. passes that responsibility to the `ViewModel` where the `ViewModel` triggers a `LiveData` that the activity is observing so that it can open the bottom sheet.
* `onSubmitButtonClicked` is now routed through the `ViewModel` and the `setUntitledStoryTitleIfTitleEmptyUseCase` performs it's task before updating a `LiveData` that triggers to the story saving. 
* `setUntitledStoryTitleIfTitleEmpty()` was moved to a use case and tests were written for it. 

## Testing
1. Create a story. 
2. Add some emoji & text to the frames. 
3. Publish the story. 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
@@ -162,8 +162,7 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
         }
 
         editorMedia.start(requireNotNull(site), this, STORY_EDITOR)
-
-        startObserving()
+        setupEditorMediaObserver()
     }
 
     override fun onLoadFromIntent(intent: Intent) {
@@ -244,7 +243,7 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
         return true
     }
 
-    fun handleMediaPickerIntentData(data: Intent) {
+    private fun handleMediaPickerIntentData(data: Intent) {
         // TODO move this to EditorMedia
         val ids = ListUtils.fromLongArray(
                 data.getLongArrayExtra(
@@ -258,7 +257,7 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
         editorMedia.addExistingMediaToEditorAsync(WP_MEDIA_LIBRARY, ids)
     }
 
-    private fun startObserving() {
+    private fun setupEditorMediaObserver() {
         editorMedia.uiState.observe(this,
                 Observer { uiState: AddMediaToPostUiState? ->
                     if (uiState != null) {
@@ -291,20 +290,6 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
                     contentIfNotHandled?.show(this)
                 }
         )
-    }
-
-    private fun saveInitialPost() {
-        editPostRepository.set {
-            val post: PostModel = postStore.instantiatePostModel(site, false, null, null)
-            post.setStatus(PostStatus.DRAFT.toString())
-            post
-        }
-        editPostRepository.savePostSnapshot()
-        // this is an artifact to be able to call savePostToDb()
-        editPostRepository.getEditablePost()?.setPostFormat(POST_FORMAT_WP_STORY_KEY)
-        site?.let {
-            savePostToDbUseCase.savePostToDb(editPostRepository, it)
-        }
     }
 
     // EditorMediaListener
--- a/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
@@ -166,12 +166,14 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
 
         viewModel.openPrepublishingBottomSheet.observe(this, Observer { event ->
             event.applyIfNotHandled {
+                analyticsTrackerWrapper.track(PREPUBLISHING_BOTTOM_SHEET_OPENED)
                 openPrepublishingBottomSheet()
             }
         })
 
-        viewModel.saveStory.observe(this, Observer { event ->
+        viewModel.submitButtonClicked.observe(this, Observer { event ->
             event.applyIfNotHandled {
+                analyticsTrackerWrapper.track(Stat.STORY_POST_PUBLISH_TAPPED)
                 processStorySaving()
             }
         })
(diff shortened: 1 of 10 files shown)


WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
```diff
@@ -406,20 +384,11 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
...
-        showPrepublishingBottomSheet()
-    }
-
-    private fun setUntitledStoryTitleIfTitleEmpty() {
-        if (editPostRepository.title.isEmpty()) {
-            val untitledStoryTitle = resources.getString(R.string.untitled)
-            storyRepositoryWrapper.setCurrentStoryTitle(untitledStoryTitle)
-            updateStoryPostTitleUseCase.updateStoryTitle(untitledStoryTitle, editPostRepository)
-        }
+        viewModel.openPrepublishingBottomSheet()
```
malinajirka, line 387: We might want to consider renaming the method to "viewModel.onStorySaveButtonPressed" and moving both "analyticsTrackerWrapper.track(PREPUBLISHING_BOTTOM_SHEET_OPENED)" and "openPrepublishingBottomSheet()" there. The view layer ideally shouldn't know/care what is going to happen on "buttonPress". Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
```diff
@@ -162,9 +153,27 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
         }
 
         editorMedia.start(requireNotNull(site), this, STORY_EDITOR)
+        setupEditorMediaObserver()
+        setupViewModelObservers()
+    }
+
+    private fun setupViewModelObservers() {
+        viewModel.mediaFilesUris.observe(this, Observer { uriList ->
+            addFramesToStoryFromMediaUriList(uriList)
```
malinajirka, line 162: This might be planned for future refactoring, so feel free to ignore this comment.... We might want to consider moving both these methods into the VM. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
```diff
@@ -14,20 +22,41 @@ import org.wordpress.android.ui.notifications.SystemNotificationsTracker
...
     private val systemNotificationsTracker: SystemNotificationsTracker,
     private val saveInitialPostUseCase: SaveInitialPostUseCase,
+    private val savePostToDbUseCase: SavePostToDbUseCase,
+    private val setUntitledStoryTitleIfTitleEmptyUseCase: SetUntitledStoryTitleIfTitleEmptyUseCase,
     private val analyticsUtilsWrapper: AnalyticsUtilsWrapper,
     private val dispatcher: Dispatcher
-) :
-        ViewModel() {
+) : ViewModel(), LifecycleOwner {
+    private val lifecycleRegistry = LifecycleRegistry(this)
```
malinajirka, line 40: Could you please elaborate why does the VM need to be LifeCycleOwner?

malinajirka, line 40: Could you please elaborate why does the VM need to be LifeCycleOwner?

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
```diff
@@ -87,8 +116,38 @@ class StoryComposerViewModel @Inject constructor(
         postEditorAnalyticsSession.setOutcome(CANCEL)
     }
 
+    private fun updateStoryPostWithChanges() {
+        editPostRepository.postChanged.observe(this, Observer {
```
jd-alexander, line 120: Hi @malinajirka so I needed to make this `ViewModel` a `LifecycleOwner` because we need to observe changes in the `EditPostRepository` and do the relevant updates. I utilized the approach that was present in the `PostListMainViewModel` I remember you saying that a lot of approaches taken there could be improved so let me know if there's a better way to do this. Thanks 🙏

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
jd-alexander: Good question. More details here https://github.com/wordpress-mobile/WordPress-Android/pull/12501#discussion_r459675301

jd-alexander: Good question. More details here #12501 (comment)

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
jd-alexander: Thanks for this comment. Yes, it would be good to move these methods into a VM. They are located within the `ComposeLoopFrameActivity` from the portkey project so when refactoring is being done there I assume that would happen then.

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
malinajirka: Ohh I see. Thanks for the explanation. Even though I'm not convinced observing within VM is a good practice I can't think of an easy enough solution right now. So I'd keep it as is.

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
jd-alexander: Resolved here https://github.com/wordpress-mobile/WordPress-Android/pull/12501/commits/c61b51fd52bc98dc2e35569afdd91e9e711f57d0

PR #12502, opened 2020-07-22
Issue/12039 reader table events

### LINKED ISSUE #12039: Reader Improvements: Repository Main Issue

zwarm:
Addresses part of  #12039 

This PR introduces EventBus events into `ReaderPostTable.java` and `ReaderDatabase.java` to notify the Reader repositories that the underlying `tbl_posts` has been modified (inserts, updates, deletes). If the posts stream has active observers, the repository will refresh itself, otherwise it sets a flag and will await active observers before refreshing.

This should not affect any existing workflows.

~~This PR is branched off of and builds upon #12454. Once #12454 is merged, then I will update this PR with the latest.~~


WordPress/src/main/java/org/wordpress/android/datasets/ReaderPostTable.java
```diff
@@ -526,10 +532,15 @@ public static int deletePostsWithTag(final ReaderTag tag) {
         }
 
         String[] args = {tag.getTagSlug(), Integer.toString(tag.tagType.toInt())};
-        return ReaderDatabase.getWritableDb().delete(
+        int rowsDeleted = ReaderDatabase.getWritableDb().delete(
```
zwarm, line 535: This is an instance in the method was directly returning the int value. I changed to keep the value local and will only post the new event is rows have actually been deleted.

WordPress/src/main/java/org/wordpress/android/datasets/ReaderPostTable.java
```diff
@@ -538,20 +549,30 @@ public static int removeTagsFromPost(long blogId, long postId, final ReaderTagTy
         }
 
         String[] args = {Integer.toString(tagType.toInt()), Long.toString(blogId), Long.toString(postId)};
-        return ReaderDatabase.getWritableDb().delete(
+        int rowsDeleted = ReaderDatabase.getWritableDb().delete(
```
zwarm, line 552: Another instance of the method directly returning the int value. I changed to keep the value local and will only post the new event is rows have actually been deleted.

WordPress/src/main/java/org/wordpress/android/datasets/ReaderPostTable.java
```diff
@@ -538,20 +549,30 @@ public static int removeTagsFromPost(long blogId, long postId, final ReaderTagTy
...
+        if (rowsDeleted > 0) {
+            EventBus.getDefault().post(ReaderPostTableActionEnded.INSTANCE);
+        }
+        return rowsDeleted;
     }
 
     public static int deletePostsInBlog(long blogId) {
         String[] args = {Long.toString(blogId)};
-        return ReaderDatabase.getWritableDb().delete("tbl_posts", "blog_id = ?", args);
+        int rowsDeleted = ReaderDatabase.getWritableDb().delete("tbl_posts", "blog_id = ?", args);
```
zwarm, line 565: Another instance of the method directly returning the int value. I changed to keep the value local and will only post the new event is rows have actually been deleted.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderDiscoverRepository.kt
```diff
@@ -5,39 +5,53 @@ import androidx.lifecycle.MutableLiveData
...
     private val fetchPostsForTagUseCase: FetchPostsForTagUseCase,
-    private val readerUpdatePostsEndedHandler: ReaderUpdatePostsEndedHandler
+    private val readerUpdatePostsEndedHandler: ReaderUpdatePostsEndedHandler,
+    private val postLikeActionUseCase: PostLikeActionUseCase
 ) : CoroutineScope {
     override val coroutineContext: CoroutineContext
         get() = bgDispatcher
 
     private var isStarted = false
+    private val isDirty = AtomicBoolean()
```
zwarm, line 54: AtomicBoolean may be overkill, but considering there may be multiple event posts felt it was safer. Open to thoughts.

Issue #12319, opened 2020-06-30
Reader Improvements: Introduce new Discover endpoint


malinajirka:
Parent issue - https://github.com/wordpress-mobile/WordPress-Android/issues/12026.

Goal of this task is to implement the new Discover endpoint in the Reader. This endpoint will be used to retrieve data for the discover tab in the Reader. 

**BackLog:**
- [x] Placeholder 

**InProgress:**
- [x] Placeholder

**Done:**
- [x] Refactor ReaderUpdatePostsEndedHandler -> FetchDiscoverCardsEnded event
- [x] Init the endpoint
- [x] Pass user's followed tags to the /cards endpoint
- [x] Init missing models
- [x] Parse the data
- [x] Cache the data 
- [x] Load cached data in offline
- [x] Return the data back to the UI
- [x] Save and Load PageHandle
- [x] Clear jsonCards database on "Fetch_first_page" request

PR #12507, opened 2020-07-23
Add support for parsing data from the new discover endpoint

malinajirka:
Parent issue #12319 

This PR adds parsing logic for the new discover endpoint

To test:
I've tested the app by faking responses in Charles and manually starting the service from the discover fragment. However, since this code isn't being used yet I don't think it's necessary for the reviewer to actually test the app. We'll have opportunity to test it before it's used in production.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/models/ReaderPost.java
+++ b/WordPress/src/main/java/org/wordpress/android/models/ReaderPost.java
@@ -83,13 +83,13 @@ public class ReaderPost {
 
         ReaderPost post = new ReaderPost();
 
-        post.postId = json.optLong("ID");
-        post.blogId = json.optLong("site_ID");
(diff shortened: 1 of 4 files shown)


malinajirka:
I realized an issue in the "simplifyJson" approach I took. We need to explicitly copy the parts we need from the json instead of removing the parts that we don't need. Because if we add support for more types of cards on the backend in the future the legacy clients might be storing tons of unnecessary data.

malinajirka:
This is ready for review again :).

malinajirka:
>How are you handling Json errors or is that not necc. here?

Good question, but I'm not sure how to answer :D. I personally thing that if the API send us a broken json, we should just crash so we are aware of the issue asap. Do you think it'd be better to play it safe and manually catch json exceptions? cc @zwarm

PR #12508, opened 2020-07-23
Add fun to support refresh requests

zwarm:
Part of #12039 

This PR exposes a function for PTR support. `refreshPosts()` will kickoff the fetch from the server and results will be returned on the `posts` stream. 

To test:
The functionality isn't available in the fragment/viewmodel yet, so I rigged up a SwipeRefreshLayout and tested that the call made it to the repo and data was returned. 

PR submission checklist:


zwarm:
> It looks good for current discover feed.  I think its time we should discuss this pending conversation about getting posts for multiple tags

@ashiagr  Thanks, this PR is for current only. A future PR will take into account multiple tags. THanks for calling it out.

PR #12514, opened 2020-07-24
Issue/12324 storage limit with stock media

### LINKED ISSUE #12324: Media: Adding images from Free Photo Library ignores site storage limit
### Expected behavior

I expect that if my WordPress.com site has reached the site storage limit, I won't be able to upload any more media to the site.

### Actual behavior

If I add images from the Free Photo Library (either in Media or in the editor), it uploads images to my site even if my site has reached the site storage limit.

0. On a WordPress.com Simple site, fill the media library up to the site's storage limit.
1. Go to My Site > Media in the app. Notice that it says "100% of [storage limit] used" at the bottom of the screen.
2. Tap the + icon to add media to the media library.
3. Select "Choose from Free Photo Library" and select an image to add.
4. Notice that the image is uploaded even though the site's media library is full.
5. Start a new blog post.
6. Add an image block to the post.
7. Tap on the image block, select "Choose from Free Photo Library," and select an image to add to the post.
8. Notice that the image is uploaded even though the site's media library is full.

You can also repeat the steps, uploading an image from the device instead of from the Free Photo Library, to confirm you get an error in that scenario.

develric:
Fixes #12324

I went back and forth on how to fix this (creating and removing branches in FluxC and here) also due to how to include in the fix not only the media library but also the Editors side (Aztec/GB); finally I tried to stay simple and (I hope) consistent enough. 

- Mapped EXCEEDS_SITE_SPACE_QUOTA_LIMIT MediaErrorType to user message.
- Avoided showing Retry action in snackbar with quota exceeded error.
- Avoided to attempt an upload from Stock Media Picker when quota has been exceeded. This applies both to Media Library and Editors since they both use the StockMediaPicker

Note that I think this solution will probably evolve along the picker consolidation project.

### WP.com with free plan
- Select a site with free plan with storage limit exceeded (I can provide a test site with storage limit exceeded if needed)
- Go to Media Library and check quota info is available at the bottom and it says 100% (**NOTE:** I was thinking we should possibly saturate the percentage to not report > 100%; even if we risk to not report to user the accurate real condition; wdyt?)

<p align="center">
  
</p>

- try to upload from device/camera and check you get a snackbar as below

<p align="center">
  
</p>

- try to upload from Pexels and check you get the following Toast message

<p align="center">
  
</p>

- Create a post from editor(s) adding from Pexels and check you get the same toast as above in StockMediaPicker
- try to free space and check that now you can upload from anywhere and successfully create posts/pages

### WP.com site with business plan && JP site 
- Go to Media Library and check no quota info is available at the bottom

<p align="center">
  
</p>

- try to upload from device/camera/Pexels and check you are able to upload image/video
- Create a post from editor(s) and check you are able to add and publish video/image from every source

### Pure self-hosted site (no access to Free Library menu)
- Check that you have no access to Free Photo Library at all

### Smoke test picker and adding image/video in editor(s)
- Smoke test picker and adding image/video in editor(s)

PR submission checklist:


planarvoid:
> I was thinking we should possibly saturate the percentage to not report > 100%; even if we risk to not report to user the accurate real condition; wdyt?

I think this is a good idea 👍 

Screenshot 2020-07-27 at 10 26 35

I'd remove the "Cannot upload" message from this snackbar. I think it's enough to say:
"1 file not uploaded. Storage quota exceeded". Also I'd maybe change the "Cannot upload" message to something else on the other dialog as well. Maybe "Cannot upload file because storage quota was exceeded". Feels a bit more human :-)

libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/FormatUtils.java
```diff
@@ -55,8 +55,25 @@ public static final String formatFileSize(long size, final String[] unitStrings)
...
     public static final String formatPercentage(double value) {
+        return formatPercentageLimit100(value, false);
+    }
+
+    /*
+     * returns the passed double percentage (0 to 1) formatted as an human readable percentage. Ex: 0.251 returns 25.1%
+     * if limit100 is true, it limits the percentage to 100%
+     */
+    public static final String formatPercentageLimit100(double value, boolean limit100) {
+        double limit = 1.0001;
```
planarvoid, line 66: I'm wondering why the limit isn't just "1" :-)

develric: I think I was having in mind that since we are setting fraction digits to be 1 the decimal number that matters is not the 3rd but the 4th, so well...maybe no strong reasons actually but hope should not hurt 🙃

planarvoid: I think using "1" would be better but this also works 👍

PR #12516, opened 2020-07-24
WP Stories: story composer ViewModel part4

mzorz:
Builds on top of #12504 

Comes from #12493 (see [comments](https://github.com/wordpress-mobile/WordPress-Android/pull/12493#issuecomment-663472463))

### Things done in this PR
- created a new StoryEditorMedia class as a starting point for a proper VM
- moved out a an enum and a listener we needed on both EditorMedia and StoryEditorMedia: `EditorMediaListener` interface and `AddExistingMediaSource` enum

removed:
- `refreshBlogMedia`, didn't use it
- `retryFailedMediaAsync`: didn't use it but maybe in the future
+ `purgeMediaToPostAssociationsIfNotInPostAnymoreAsync`: I think we'll need this one as soon as we allow editing (very soon, before launch). Took it away for now to leave things in a clean state, will revisit soon
- `reattachUploadingMediaForAztec`, we don't have Aztec.
- `definitelyDeleteBackspaceDeletedMediaItemsAsync` we don't need this either
- `updateDeletedMediaItemIds`: removed
+ `onMediaDeleted`: we'll want to use this when editing posts as well. Deleting for now.
- `onMediaUploadError`: didn't use it, we don't show upload errors on the UI.
- `updateMediaUploadStateBlocking` (we don't upload with the UI on the front)
- `cancelMediaUploadAsync` (we don't upload with the UI on the front)
- `addGifMediaToPostAsync` (we don't use gifs for now)
- `addFreshlyTakenVideoToEditor` (we add video differently because the StoryComposer has its own camera capture mode)
- `addExistingMediaToEditorAsync` variants we didn't use (removed)

#### Some notes / questions
- `AddMediaToStoryPostUiState`. This class is a copy & paste of `AddMediaToPostUiState` in `EditorMedia`. Wonder how to make this better; I didn't feel it was OK to reuse the same class. Probably not a big deal, but copy/paste leaves me with a bitter taste 🤷‍♂️ 

To test:
- Make sure EditPostActivity media handling works well (add media from different sources, reattach, cancel, etc)
- Make sure StoryComposerActivity works well (add media from different sources, save, publish). 

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/EditorMediaListener.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/EditorMediaListener.kt
@@ -1,0 +1,13 @@
+package org.wordpress.android.ui.posts.editor.media
+
+import org.wordpress.android.fluxc.model.PostImmutableModel
+import org.wordpress.android.ui.posts.EditPostActivity.OnPostUpdatedFromUIListener
+import org.wordpress.android.util.helpers.MediaFile
+
+interface EditorMediaListener {
+    fun appendMediaFiles(mediaFiles: Map<String, MediaFile>)
+    fun syncPostObjectWithUiAndSaveIt(listener: OnPostUpdatedFromUIListener? = null)
+    fun advertiseImageOptimization(listener: () -> Unit)
+    fun getImmutablePost(): PostImmutableModel
+}
+
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/EditorMediaListener.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/EditorMediaListener.kt
@@ -10,4 +10,3 @@ interface EditorMediaListener {
     fun advertiseImageOptimization(listener: () -> Unit)
     fun getImmutablePost(): PostImmutableModel
 }
-
--- a/WordPress/src/main/java/org/wordpress/android/WordPress.java
+++ b/WordPress/src/main/java/org/wordpress/android/WordPress.java
@@ -90,7 +90,7 @@ import org.wordpress.android.ui.posts.editor.ImageEditorTracker;
 import org.wordpress.android.ui.prefs.AppPrefs;
 import org.wordpress.android.ui.reader.tracker.ReaderTracker;
 import org.wordpress.android.ui.stats.refresh.lists.widget.WidgetUpdater.StatsWidgetUpdaters;
-import org.wordpress.android.ui.stories.StoryMediaSaveUploadBridge;
+import org.wordpress.android.ui.stories.media.StoryMediaSaveUploadBridge;
 import org.wordpress.android.ui.uploads.UploadService;
 import org.wordpress.android.ui.uploads.UploadStarter;
 import org.wordpress.android.util.AppLog;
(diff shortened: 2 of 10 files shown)


WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/EditorMedia.kt
```diff
@@ -144,7 +126,7 @@ class EditorMedia @Inject constructor(
                     site,
                     freshlyTaken,
                     editorMediaListener,
-                    (editorType == POST_EDITOR) // also start upload if this is a Posts editor
```
jd-alexander, line 147: @mzorz this was removed because it's always true right?

jd-alexander: Okay, I am seeing where we used to have `STORY_EDITOR` for stories. Will review this in more detail.

jd-alexander: is it a case where this is not necessary anymore since we aren't sharing the `EditorMedia` with the `StoryComposerActivity` anymore?

mzorz: Right, there's no longer a disntinction in EditorMedia for Stories so, it's always `true` there 👍

jd-alexander: Okay awesome!

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/EditorMediaListener.kt
```diff
@@ -0,0 +1,12 @@
+package org.wordpress.android.ui.posts.editor.media
+
+import org.wordpress.android.fluxc.model.PostImmutableModel
+import org.wordpress.android.ui.posts.EditPostActivity.OnPostUpdatedFromUIListener
+import org.wordpress.android.util.helpers.MediaFile
+
+interface EditorMediaListener {
```
malinajirka, line 7: No need to change anything. I'm just sharing my thoughts about this listener.

We utilized this listener in the EditPostActivity even though it was never a perfect solution. It can very easily lead to a huge memory leak when the VM keeps reference to the listener (activity) after the activity is destroyed. Ideally, we'd be utilizing livedata or some other form of communication between the view and the vm layers.

mzorz:
> I truly appreciate that you worked on these improvements @mzorz! Thank you 🙇

Thank you! As said, you made a really good job explaining your point of view on the [other PR](https://github.com/wordpress-mobile/WordPress-Android/pull/12493#issuecomment-663472463). With that in mind, it was easy to follow up and know what to do :D 

> I personally wouldn't worry about it. I don't consider every copy-paste to be a "bad practice". I like to think about it deeply when I need to copy paste something more than two times - that's usually a bad sign.

Yes,  makes sense. Same thing happened to me re: thinking of it deeply and then decided to have it; it just made sense to have its own separate class even if it did / looked the same, given their tight relationship with their purpose (to contribute to the containing VM's mission) begs for this distinction to be made, especially when the actual effort these 2 PRs are trying to tackle is that of writing code that is consistent with the VM approach. 👍 

Thank you for your reviews @jd-alexander @malinajirka ! 🙇

Issue #10012, opened 2019-06-08
Account Settings Shows Unchanged Email Address as Empty


theck13:
### Expected
Saving an unchanged email address shows the previous value in the ***Email Address*** field.

Alternatively, there would be validation on the dialog input field to enable the ***OK*** button only after the value entered is different than the value when the dialog was created.

### Observed
Saving an unchanged email address shows an empty value in the ***Email Address*** field.

### Reproduced
1. Go to ***Me*** tab.
2. Tap ***Account Settings*** option.
3. Notice email address shown in ***Email Address*** summary.
4. Tap ***Email Address*** option.
5. Tap ***OK*** button.
6. Notice email address hidden in ***Email Address*** summary.
7. Tap back arrow.
8. Tap ***Account Settings*** option.
9. Notice email address shown in ***Email Address*** summary.

##### Tested
Google Pixel 3 on Android 9.0 with WordPress alpha-172

hafizrahman:
In looking at this and trying to replicate the issue, the issue still persists but I'm seeing a new behavior:

1. Go to Me tab.
2. Tap Account Settings option.
3. Notice email address shown in Email Address summary.
4. Tap Email Address option.
5. Tap OK button without changing any email.
6. For a second or so, the `Click the verification link in the email sent to %1$s to confirm your new address` snackbar appears, then disappears quickly
7. Notice email address hidden in Email Address summary.
8. Tap back arrow.
9. Tap Account Settings option.
10. Email address is now still hidden,
11. Tap back arrow twice to the "My Site" screen,
12. Go back to Me > Account Settings
13. The hidden email is now shown again.

hafizrahman:
I am unable to replicate this on a local build, though. After step 4, it gives me the error message:

`Your client is not permitted to set the user's email`

and then none of the other steps happen.

This is on version 15.0-rc-1.

hypest:
> Your client is not permitted to set the user's email

Ah, looks like this is one of the privileged fields that only the official app can change. Sorry about that @hafizrahman :(

jkmassel:
This appears to have been resolved.

PR #12517, opened 2020-07-24
Issue/12319 new discover endpoint db table

malinajirka:
Parent issue #10012 

This is a follow up for https://github.com/wordpress-mobile/WordPress-Android/pull/12507

Merge instructions:
1. Review this PR
2. Make sure https://github.com/wordpress-mobile/WordPress-Android/pull/12507 is merged
3. Update target branch to `issue/12319-new-discover-endpoint-main-issue `
4. Remove "Not Ready for Merge" tag
5. Merge this PR

This PR creates a new table which will be used for storing simplified responses from the /read/cards/ endpoint. The data will be displayed on the "discover" tab in Reader. 

The endpoint returns different types of cards. Currently, only two types are supported - `posts` and `interests_you_might_like`. However, more types will be added in the future. 

We considered creating a table for each type but it felt like an overkill. We decided to store just a simplified version of the server response into the database. Since we already have an existing table for post objects these are stored into tbl_posts. The simplified json response contains just postId, siteId and pseudoId. All the other fields of a post can be loaded from the tbl_posts table.

To test:
Make sure the app doesn't crash or log any migration errors when the reader improvements feature flag is turned off. Rest of this functionality will be tested when we start propagating the data to the UI.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/services/discover/ReaderDiscoverLogic.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/services/discover/ReaderDiscoverLogic.kt
@@ -6,7 +6,10 @@ import com.wordpress.rest.RestRequest.Listener
 import org.json.JSONArray
 import org.json.JSONObject
 import org.wordpress.android.WordPress
+import org.wordpress.android.datasets.ReaderDiscoverCardsTable
+import org.wordpress.android.datasets.ReaderPostTable
 import org.wordpress.android.models.ReaderPost
+import org.wordpress.android.models.ReaderPostList
 import org.wordpress.android.models.ReaderTag
 import org.wordpress.android.models.ReaderTagList
 import org.wordpress.android.models.ReaderTagType.DEFAULT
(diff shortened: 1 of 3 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/services/discover/ReaderDiscoverLogic.kt
```diff
@@ -127,6 +130,12 @@ class ReaderDiscoverLogic constructor(private val completionListener: ServiceCom
         return cards
     }
 
+    private fun insertPostsIntoDb(posts: List<ReaderPost>) {
+        val postList = ReaderPostList()
+        postList.addAll(posts)
+        ReaderPostTable.addOrUpdatePosts(null, postList)
```
malinajirka, line 136: I'm wondering if we want to insert the posts under a fake tag so we can easily delete them later. Wdyt?

zwarm: @malinajirka - why a fake tag? Couldn't they use the "Discover" readerTag ... that's what they are ? No one but this endpoint should be adding `discover` readerTag to the table anymore.

malinajirka: >No one but this endpoint should be adding discover readerTag to the table anymore.

We'll need to think about this too. Discover tag was always used for posts from "discover.wordpress.com". If it's a server side tag as well we might not be able to just change it. My bet is that it's still used in Calypso's Reader.

PR #12518, opened 2020-07-24
WP Stories: story composer ViewModel part5

mzorz:
Builds on top of #12516 

This PR converts StoryEditorMedia into a StoryEditorMediaViewModel. 

#### Description
In reality there's not much needed to be done given the logic already was working and tested, but making this a ViewModel makes it comply with general architecture patterns we're implementing throughout WPAndroid, and prepares it for further functionality addition.

#### Notes
- the StoryViewModel now knows about StoryEditorMediaViewModel through the EditorMediaListener interface, wonder if this is a concern

To test:
- test the Stories functionality works (add local, external media, publish etc)

Just 1 reviewer needed for approval, but happy to learn from any comments! 

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
@@ -60,7 +60,6 @@ import org.wordpress.android.util.analytics.AnalyticsUtilsWrapper
 import org.wordpress.android.util.helpers.MediaFile
 import org.wordpress.android.viewmodel.Event
 import org.wordpress.android.widgets.WPSnackbar
-import java.lang.ref.WeakReference
 import java.util.Objects
 import javax.inject.Inject
 
@@ -155,7 +154,7 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
             )
         }
 
-        storyEditorMediaViewModel.start(requireNotNull(site), WeakReference(this))
+        storyEditorMediaViewModel.start(requireNotNull(site), this)
         setupStoryEditorMediaObserver()
         setupViewModelObservers()
     }
--- a/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
@@ -26,7 +26,6 @@ import org.wordpress.android.WordPress
 import org.wordpress.android.analytics.AnalyticsTracker.Stat
 import org.wordpress.android.analytics.AnalyticsTracker.Stat.PREPUBLISHING_BOTTOM_SHEET_OPENED
 import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
-import org.wordpress.android.fluxc.model.PostImmutableModel
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.store.PostStore
 import org.wordpress.android.push.NotificationType
@@ -38,7 +37,6 @@ import org.wordpress.android.ui.media.MediaBrowserActivity
 import org.wordpress.android.ui.media.MediaBrowserType
 import org.wordpress.android.ui.pages.SnackbarMessageHolder
 import org.wordpress.android.ui.photopicker.PhotoPickerActivity
-import org.wordpress.android.ui.posts.EditPostActivity.OnPostUpdatedFromUIListener
 import org.wordpress.android.ui.posts.EditPostRepository
 import org.wordpress.android.ui.posts.EditPostSettingsFragment.EditPostActivityHook
 import org.wordpress.android.ui.posts.PostEditorAnalyticsSession
@@ -47,7 +45,6 @@ import org.wordpress.android.ui.posts.ProgressDialogHelper
 import org.wordpress.android.ui.posts.ProgressDialogUiState
 import org.wordpress.android.ui.posts.PublishPost
 import org.wordpress.android.ui.posts.editor.media.AddExistingMediaSource.WP_MEDIA_LIBRARY
-import org.wordpress.android.ui.posts.editor.media.EditorMediaListener
 import org.wordpress.android.ui.posts.prepublishing.PrepublishingBottomSheetListener
 import org.wordpress.android.ui.stories.media.StoryEditorMediaViewModel
 import org.wordpress.android.ui.stories.media.StoryEditorMediaViewModel.AddMediaToStoryPostUiState
@@ -57,16 +54,13 @@ import org.wordpress.android.util.ListUtils
 import org.wordpress.android.util.WPMediaUtils
 import org.wordpress.android.util.analytics.AnalyticsTrackerWrapper
 import org.wordpress.android.util.analytics.AnalyticsUtilsWrapper
-import org.wordpress.android.util.helpers.MediaFile
 import org.wordpress.android.viewmodel.Event
 import org.wordpress.android.widgets.WPSnackbar
-import java.util.Objects
 import javax.inject.Inject
 
 class StoryComposerActivity : ComposeLoopFrameActivity(),
         SnackbarProvider,
         MediaPickerProvider,
-        EditorMediaListener,
         AuthenticationHeadersProvider,
         NotificationIntentLoader,
         MetadataProvider,
@@ -154,7 +148,7 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
             )
         }
 
-        storyEditorMediaViewModel.start(requireNotNull(site), this)
+        storyEditorMediaViewModel.start(requireNotNull(site), viewModel)
         setupStoryEditorMediaObserver()
         setupViewModelObservers()
     }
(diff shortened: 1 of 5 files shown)


WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
```diff
@@ -151,7 +148,7 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
             )
         }
 
-        storyEditorMedia.start(requireNotNull(site), this)
+        storyEditorMediaViewModel.start(requireNotNull(site), viewModel)
```
jd-alexander, line 151: I know @malinajirka will have some thoughts about this 😅 What I normally do in a case like this, is let the `StoryComposerActivity` implement this listener and then I would pass the events/data back to the `ViewModel` and then it would know what to do with them at that point.

Again, this is a personal opinion of mine. You can wait for @malinajirka response so we can decide a course of action 😄

jd-alexander: So my idea is that all the functions from the listener that are implemented in the `ViewModel` could still remain and we would remove the overrides and let the activity call those functions from its overrides of the listener.

mzorz: Sounds good, I knew this was going to be controversial :D - so basically if I'm getting you right we'd:
- keep the StoryComposerActivity as an implementer of EditorMediaListener
- remove the EditorMediaListener overrides from the ViewModel
- Add a wrapper method to access editPostRepository through the ViewModel (instead of directly from the Activity) in the getImmutablePost override

Then the problem of having a listener which is the Activity and now is no longer the ViewModel, is that we probably need to be careful with the reference to it that is kept in the ViewModel (as the Activity would also be the EditorMediaListener). On a configuration change, we need to get rid of the listener reference (which is the destroyed Activity) even if the ViewModel instance will remain the same. Is this correct? Wdyt?

Thanks for your comments 🙏

malinajirka: >Then the problem of having a listener which is the Activity and now is no longer the ViewModel, is that we probably need to be careful with the reference to it that is kept in the ViewModel (as the Activity would also be the EditorMediaListener). On a configuration change, we need to get rid of the listener reference (which is the destroyed Activity) even if the ViewModel instance will remain the same. Is this correct? Wdyt?

💯 agree! That's the main reason we never turned the EditorMedia into a real VM in the EditPostActivity.

Since both `StoryComposerViewModel` and `StoryEditorMediaViewModel` share the same scope, I believe it's fine to have a reference from one to the other. I'm not a fan of having `observe` methods within the VM. But since this listener doesn't use LiveData, I think it's ok to implement the interface within the VM. Tbh I would even say it's preferred exactly because of the possible memory leaks Mario mentioned.

Another approach is something I mentioned on the previous PR -> get rid of the listener and use LiveData instead. We'd need to think about where to put it if we did that. But that's another story :). 

Wdyt? 

@jd-alexander Can you foresee any issues with the current approach?

WordPress/src/main/java/org/wordpress/android/ui/stories/media/StoryEditorMediaViewModel.kt
```diff
@@ -18,22 +19,22 @@ import org.wordpress.android.ui.posts.editor.media.AddExistingMediaSource
...
 import kotlin.coroutines.CoroutineContext
 
-class StoryEditorMedia @Inject constructor(
+class StoryEditorMediaViewModel @Inject constructor(
     private val mediaUtilsWrapper: MediaUtilsWrapper,
     private val addLocalMediaToPostUseCase: AddLocalMediaToPostUseCase,
     private val addExistingMediaToPostUseCase: AddExistingMediaToPostUseCase,
     @Named(UI_THREAD) private val mainDispatcher: CoroutineDispatcher
-) : CoroutineScope {
+) : ViewModel(), CoroutineScope {
```
malinajirka, line 37: We might want to consider extending the `ScopedViewModel` to save a few lines of code. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
```diff
@@ -52,6 +55,9 @@ class StoryComposerViewModel @Inject constructor(
     private val _submitButtonClicked = MutableLiveData<Event<Unit>>()
     val submitButtonClicked: LiveData<Event<Unit>> = _submitButtonClicked
 
+    private val _shouldShowOptimizeImagesAdvertising = MutableLiveData<Event<() -> Unit>>()
```
malinajirka, line 58: We might want to consider renaming this to something like `advertizeImageOptimizationEvent` or `showOptimizeImagesDialog` or similar. `Should show` evokes "boolean" result to me, wdyt?

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
mzorz: > Since both StoryComposerViewModel and StoryEditorMediaViewModel share the same scope, I believe it's fine to have a reference from one to the other. I'm not a fan of having observe methods within the VM. But since this listener doesn't use LiveData, I think it's ok to implement the interface within the VM. Tbh I would even say it's preferred exactly because of the possible memory leaks Mario mentioned.

Excellent, also thanks for explaining how this was considered for EditPostActivity as well. 🙌🏻

> Another approach is something I mentioned on the previous PR -> get rid of the listener and use LiveData instead. We'd need to think about where to put it if we did that. But that's another story :).

This is interesting, I can see how things would be decoupled here and conform to the lifecycle states. May be worth a follow up PR once we think where to put what 👍

WordPress/src/main/java/org/wordpress/android/ui/stories/media/StoryEditorMediaViewModel.kt
mzorz: ah, missed using that one 👍 addressed in da0e12ddc6

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
mzorz: 👍 addressed in 2b583f9ee3

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
jd-alexander: Hi everyone! I can't foresee any issues with the current approach so let's keep it. Thanks for expounding on this point. I appreciate the discussion!

jd-alexander: > Since both StoryComposerViewModel and StoryEditorMediaViewModel share the same scope, I believe it's fine to have a reference from one to the other. I'm not a fan of having observe methods within the VM. But since this listener doesn't use LiveData, I think it's ok to implement the interface within the VM. Tbh I would even say it's preferred exactly because of the possible memory leaks Mario mentioned.

Thanks for this. I want you to expound a bit on how memory leaks work based on the solution I described. The `ViewModel` would not contain a reference to the `Activity`, it would just have functions that get called by the methods of the listener. Nonetheless, I am totally fine with just sharing the same scope and keeping this current solution 😄 I just want some clarification on the ideas I expressed as it's something I normally do and I want to follow best practices 🙏

mzorz: > Thanks for this. I want you to expound a bit on how memory leaks work based on the solution I described. The ViewModel would not contain a reference to the Activity, it would just have functions that get called by the methods of the listener.

sure thing! so, earlier in one of the above comments you wrote this:

> [...] and let the activity call those functions from its overrides of the listener.

So from this I understood the suggestion included having the Activity implement the EditorMediaListener interface, and call the ViewModel's functions from its overrides, right? But here's the thing:

>  The ViewModel would not contain a reference to the Activity, it would just have functions that get called by the methods of the listener.

that's what it would seem like in the surface, but the bummer here is that several of the use cases used in the ViewModel need an EditorMediaListener to work (for example `addExistingMediaToPostUseCase`, `addLocalMediaToPostUseCase`, etc. and it gets passed down to other use cases these are composed of as well) - sometimes the listener even being called later async, and as such if the EditorMediaListener is indeed an Activity instance, then we would indeed have a reference to the Activity in the ViewModel, moreover this Activity instance may potentially be destroyed by the time the listener overrides get called by the different usecases in the VM, which we didn't want in the first place :).

As per my understanding, a refactor could be done along the lines of what you and Jirka suggested by using LiveData for communication, but would demand an effort beyond the scope of the PR (although it leaves me curious as well I admit).

Did you have one of those cases in mind, or even something different? Let me know if that clarifies things a bit, as I'm also paying attention to not miss any useful ideas you all may have 🙇 

> I am totally fine with just sharing the same scope and keeping this current solution 😄 I just want some clarification on the ideas I expressed as it's something I normally do and I want to follow best practices 

Awesome! Let me know if this works then - or should we explore further 🙏

malinajirka: >This is interesting, I can see how things would be decoupled here and conform to the lifecycle states. May be worth a follow up PR once we think where to put what 👍

I personally kind of like the vm2vm communication through a listener. The part that'd be worth improving is EditPostActivity, but changing it into a state that we wouldn't need that listener(activity reference) might be quite complicated.

jd-alexander: Oh, awesome. I totally understand this now @mzorz @malinajirka Thanks for the lessons 🙇

PR #12520, opened 2020-07-24
Issue 12334/story title layout refactor

### LINKED ISSUE #12334: WordPress Stories : Story Title UI & Keyboard Issues

jd-alexander:
Fixes #12334 

## Solution
In this PR I am removing the `Story Title` component from the `RecyclerView` because it's causing focus issues.
A separate view has been creating to host this functionality and it's being added directly about the `RecyclerView` to mimic the effect of having a header view. 

Along with this functionality, you are also able to tap anywhere in the story title layout to trigger the focus of the `EditText` and open the keyboard.

## Testing

1. Create a story post. 
2. Click Next. 
3. See the bottom sheet open with the keyboard active and the story title focused.

<kbd></kbd>

-------

1. Create a story post. 
2. Click Next. 
3. See the bottom sheet open with the keyboard active and the story title focused.
4. Close the keyboard. 
5. Click anywhere within the story title layout like below.

<kbd></kbd>

-----
1. Create a story post. 
2. Click Next. 
3. See the bottom sheet open with the keyboard active and the story title focused.
4. Click the Tags action. 
5. See that the keyboard is opened. 
6. Navigate back and see that the keyboard is opened for the story title. 

 <kbd></kbd>

-----
1. Create a story post. 
2. Click Next. 
3. See the bottom sheet open with the keyboard active and the story title focused.
4. Click the Publish Date action. 
5. See that the keyboard is closed before showing the Publish Date UI. 

<kbd></kbd>

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeAdapter.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeAdapter.kt
@@ -11,16 +11,15 @@ import org.wordpress.android.ui.posts.PrepublishingHomeItemUiState.HomeUiState
 import org.wordpress.android.ui.posts.PrepublishingHomeItemUiState.StoryTitleUiState
 import org.wordpress.android.ui.posts.PrepublishingHomeViewHolder.PrepublishingHeaderListItemViewHolder
 import org.wordpress.android.ui.posts.PrepublishingHomeViewHolder.PrepublishingHomeListItemViewHolder
-import org.wordpress.android.ui.posts.PrepublishingHomeViewHolder.PrepublishingStoryTitleItemViewHolder
 import org.wordpress.android.ui.posts.PrepublishingHomeViewHolder.PrepublishingSubmitButtonViewHolder
 import org.wordpress.android.ui.utils.UiHelpers
 import org.wordpress.android.util.image.ImageManager
+import java.lang.IllegalStateException
 import javax.inject.Inject
 
 private const val headerViewType: Int = 1
 private const val homeItemViewType: Int = 2
 private const val submitButtonViewType: Int = 3
-private const val storyTitleViewType: Int = 4
 
 class PrepublishingHomeAdapter(context: Context) : RecyclerView.Adapter<PrepublishingHomeViewHolder>() {
     private var items: List<PrepublishingHomeItemUiState> = listOf()
@@ -33,7 +32,6 @@ class PrepublishingHomeAdapter(context: Context) : RecyclerView.Adapter<Prepubli
 
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): PrepublishingHomeViewHolder {
         return when (viewType) {
-            storyTitleViewType -> PrepublishingStoryTitleItemViewHolder(parent, uiHelpers, imageManager)
             headerViewType -> PrepublishingHeaderListItemViewHolder(parent, uiHelpers, imageManager)
             homeItemViewType -> PrepublishingHomeListItemViewHolder(parent, uiHelpers)
             submitButtonViewType -> PrepublishingSubmitButtonViewHolder(parent, uiHelpers)
@@ -54,10 +52,11 @@ class PrepublishingHomeAdapter(context: Context) : RecyclerView.Adapter<Prepubli
 
     override fun getItemViewType(position: Int): Int {
         return when (items[position]) {
-            is StoryTitleUiState -> storyTitleViewType
             is HeaderUiState -> headerViewType
             is HomeUiState -> homeItemViewType
             is ButtonUiState -> submitButtonViewType
+            is StoryTitleUiState ->
+                throw IllegalStateException("StoryTitleUiState is not supported by the PrepublishingHomeAdapter")
         }
     }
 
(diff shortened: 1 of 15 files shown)


WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeAdapter.kt
```diff
@@ -54,10 +52,11 @@ class PrepublishingHomeAdapter(context: Context) : RecyclerView.Adapter<Prepubli
 
     override fun getItemViewType(position: Int): Int {
         return when (items[position]) {
-            is StoryTitleUiState -> storyTitleViewType
             is HeaderUiState -> headerViewType
             is HomeUiState -> homeItemViewType
             is ButtonUiState -> submitButtonViewType
+            is StoryTitleUiState ->
+                throw IllegalStateException("StoryTitleUiState is not supported by the PrepublishingHomeAdapter")
```
mzorz, line 59: oh nice to indicate support like this from the get go 👍

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryTitleHeaderView.kt
```diff
@@ -0,0 +1,56 @@
...
+        story_title.focusAndShowKeyboard()
+
+        story_title_content.setOnClickListener {
+            story_title.focusAndShowKeyboard()
+        }
+
+        story_title.addTextChangedListener(object : TextWatcher {
+            override fun beforeTextChanged(s: CharSequence?, start: Int, count: Int, after: Int) {}
+            override fun onTextChanged(s: CharSequence?, start: Int, before: Int, count: Int) {}
+
```
mzorz, line 50: super minor np we could remove this blank line

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryTitleHeaderView.kt
```diff
@@ -0,0 +1,56 @@
...
+            override fun afterTextChanged(view: Editable?) {
+                uiState.onStoryTitleChanged.invoke(view.toString())
```
mzorz, line 52: what if the view is null, woudn't the `toString()` throw an NPE here?

WordPress/src/main/java/org/wordpress/android/util/ViewUtils.kt
```diff
@@ -35,3 +38,43 @@ fun View.expandTouchTargetArea(@DimenRes dimenRes: Int, heightOnly: Boolean = fa
         parent.touchDelegate = TouchDelegate(touchTargetRect, this)
     }
 }
+
+/**
+ * Ensures that the keyboard is only shown when the view has received focused.
+ *
+ * https://developer.squareup.com/blog/showing-the-android-keyboard-reliably/
+ */
+fun View.focusAndShowKeyboard() {
```
mzorz, line 47: 👍 already reviewed this class when checking the other PR, agreed this is useful 🙇

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryTitleHeaderView.kt
jd-alexander: Thanks for spotting 🙏

jd-alexander: Resolved here 949dfb6

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryTitleHeaderView.kt
jd-alexander: Yes, it would! Resolved here 2032884

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingViewModel.kt
```diff
@@ -64,6 +67,7 @@ class PrepublishingViewModel @Inject constructor(private val dispatcher: Dispatc
     }
 
     private fun navigateToScreen(prepublishingScreen: PrepublishingScreen) {
+        _dismissKeyboard.postValue(Event(Unit))
```
mzorz, line 70:
```suggestion
        // if destintaion is HOME or TAGS and the current is HOME OR TAGS just don't dismiss the keyboard to avoid
        // re-layouts
        if (prepublishingScreen != TAGS && prepublishingScreen != HOME) {
            _dismissKeyboard.postValue(Event(Unit))
        }
```

I just tried this small tweak here to avoid the flickering that is observed when the keyboard is  up and we navigate to the Tags section: given we know both the HOME and the TAGS screens have an EditText, we can ask to send the dismissKeyboard signal only when we're not either in one of these nor navigating towards one of these. At this point in code we only know where we want to navigate to, but it's ok since landing on any of these two we'll want the keyboard to stay up if it was already up ;) (i.e. don't dismiss it).
For the case where this is not a story and hence there's no EditText in the HOME screen, we're ok too, because there wouldn't have been a keyboard up anyway.
Let me know your thoughts 👍

jd-alexander: Thanks for this @mzorz 
I had done this but I discovered this bug on my Google Pixel 4 XL, where sometimes the `KeyboardResizeUtil` doesn't get triggered.
Check this video out and let me know your thoughts. https://drive.google.com/file/d/1tb9lc0Yi_uzVKv9_XRdwoALFGnnW2xH9/view?usp=sharing

mzorz: Oh I see! ok let me check a bit and get back here

PR #12523, opened 2020-07-25
WP Stories: introduce feature flag

mzorz:
This PR introduces a feature flag so WP Stories is only present on dev and internal builds (wasabi, jalapeno, alpha) and hidden from release (vanilla) builds. The flag is `true` for any build that is not the vanilla flavor.

To test:
1. Verify that the create Story option appears in the main screen FAB in the following builds:
    * `JalapenoDebug` build
    * `WasabiDebug` build
    * `ZalphaRelease` build
Also test that if you have full access to the site (you're an Admin) you'll be presented 3 options (Create Post, Page, Story).
If not (i.e. you're an Editor), you'll be presented with "Post, Story" options only.
The FAB tooltip is updated as well, so on a clean install it should show "Create a post, page or story" for admins, and "Create a post or story" for Editors.

2. Verify that Story option does *not* exist in a 
    * `VanillaRelease` build
   In this one, the FAB functionality should be as follows:
       *  if you have full access, the FAB should present 2 options (create Post, Page)
       * if you're only an editor (not admin), the FAB should directly create a new post and present the default selected editor (Aztec/Gutenberg)

The FAB tooltip in this case should remain as usual, so on a clean install it should show "Create a post or page" for admins, and "Create a post" for Editors (contributors).

Unit tests have been updated to reflect this as well (will have to be updated once more when the feature flag is removed).

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/util/config/WPStoriesFeatureConfig.kt
+++ b/WordPress/src/main/java/org/wordpress/android/util/config/WPStoriesFeatureConfig.kt
@@ -1,0 +1,7 @@
+package org.wordpress.android.util.config
+
+import org.wordpress.android.BuildConfig
+import javax.inject.Inject
+
+class WPStoriesFeatureConfig
+@Inject constructor(appConfig: AppConfig) : FeatureConfig(appConfig, BuildConfig.WP_STORIES_AVAILABLE)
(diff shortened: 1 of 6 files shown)


WordPress/src/main/java/org/wordpress/android/util/config/WPStoriesFeatureConfig.kt
```diff
@@ -0,0 +1,7 @@
+package org.wordpress.android.util.config
+
+import org.wordpress.android.BuildConfig
+import javax.inject.Inject
+
+class WPStoriesFeatureConfig
+@Inject constructor(appConfig: AppConfig) : FeatureConfig(appConfig, BuildConfig.WP_STORIES_AVAILABLE)
```
planarvoid, line 7: ❤️

mzorz:
Thank you for your review @planarvoid 🙇 🙌 

> There are 2 more reviewers assigned to this PR so I'm not merging it but feel free to merge it

Pinged @aforcier and @jd-alexander so they're aware, merging this one 👍 

Thanks again!

PR #12526, opened 2020-07-27
RI: Save Interests (logged in user)

ashiagr:
Parent Issue: #12027

This PR 
- Saves interests (locally + remotely) for a logged in user
- Handles errors while saving

#### To Test

Prerequisite - RI FF flag is enabled

**On Success**

- Launch and go to Reader tab
- Interests from server are shown
- Select few interests 
- Notice that "Done" button is enabled
- Click "Done" button
- Notice that progress bar is shown and Done button is disabled
- Wait for few seconds
- Notice that Interests Picker is dismissed
- Click "Filter" bar in the Following tab
- Notice that bottom sheet is shown
- Click "Tags" section in the bottom sheet
- Notice that selected interests are shown in the "Tags" section 

| On Success |
|---|
save_interests

-----

**On Error**

- Launch and go to Reader tab
- Interests from server are shown
- Select few interests, notice that "Done" button is enabled
- Disable airplane mode
- Click "Done" button
- Notice that progress bar is shown and "Done" button is disabled
- Wait for few seconds
- Notice that request fails, "Done" button is enabled and network error is shown in a snackbar

To test remote request failure error, comment out `Success` response and return `RemoteRequestFailure` [here](https://github.com/wordpress-mobile/WordPress-Android/blob/9ed4e5e45c82095ee68c1b827f1c93d60cf1aa61/WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/FollowInterestTagsUseCase.kt#L43-L47).

| On Error |
|---|
| error |

#### Note

- Saving interests for logged out/ self-hosted site will be handled in a separate PR, coming soon. 
- Get user tags (used to determine when to show "Interests Picker") accesses dummy method in the tags repository. So even when Interests are saved, "Interests Picker" will be displayed again the next time. 
- Unit tests for use case will be covered in upcoming PRs.
- Error message needs copy review, will mention in the final design review.

PR submission checklist:


WordPress/src/main/res/values/strings.xml
```diff
@@ -1307,7 +1307,7 @@
     <string name="reader_filter_empty_tags_action">Add a tag</string>
     <string name="reader_filter_self_hosted_empty_sites_tags_action">Log in to WordPress.com</string>
     <string name="reader_dot_separator" translatable="false">\u0020\u2022\u0020</string>
-    <string name="reader_error_generic_title">There was a problem handling the request. Please try again later.</string>
+    <string name="reader_error_request_failed_title">There was a problem handling the request. Please try again later.</string>
```
malinajirka, line 1310: This change will remove all the translation in GlotPress. We might want to consider creating a new untranslatable key and referencing the old key. Wdyt?

ashiagr: Old key `reader_error_generic_title` was added only on the Interests Picker screen which is still not a live feature. I just renamed it. Wondering if we still need to fix it.

ashiagr:
@malinajirka, as discussed on Slack, I've merged `addTag` and `addTags`. Ready for another look, thanks 🙏.

WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderTagActions.java
```diff
@@ -67,59 +74,79 @@ public void onErrorResponse(VolleyError volleyError) {
...
-                ReaderTagList tags = parseFollowedTags(jsonObject);
-                ReaderTagTable.replaceFollowedTags(tags);
+        com.wordpress.rest.RestRequest.Listener listener = jsonObject -> {
+            AppLog.i(T.READER, "add tag succeeded");
+            // the response will contain the list of the user's followed tags
+            ReaderTagList followedTags = parseFollowedTags(jsonObject);
+            ReaderTagTable.replaceFollowedTags(followedTags);
+            if (actionListener != null) {
                 ReaderActions.callActionListener(actionListener, true);
             }
```
ashiagr, line 113: I'd want to remove this action listener used to communicate result when following a single tag (on Topic Details and ReaderSubsActivity). I can do it when I refactor "Topic Details" to support logged out scenario. Or let me know if I should do it in this PR.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/FollowInterestTagsUseCase.kt
```diff
@@ -0,0 +1,54 @@
...
+import org.wordpress.android.ui.reader.repository.ReaderRepositoryCommunication.Error.RemoteRequestFailure
+import org.wordpress.android.ui.reader.repository.ReaderRepositoryCommunication.Success
+import org.wordpress.android.util.EventBusWrapper
+import org.wordpress.android.util.NetworkUtilsWrapper
+import javax.inject.Inject
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.resume
+import kotlin.coroutines.suspendCoroutine
+
+class FollowInterestTagsUseCase @Inject constructor(
```
ashiagr, line 19: We might want to rename it as `FollowTagsUseCase` so that we can reuse it when following a single tag as well. Do we want to structure it more like `PostLikeActionUseCase` ?

WordPress/src/main/res/values/strings.xml
malinajirka: If it was in develop before the code freeze the strings were already uploaded to GlotPress and translated to at least the main languages. Having said that I guess it's not a big deal. We can just rename it.

ashiagr:
Thanks for the review, @malinajirka !

> One thing I'm not sure about is showing the "progress" indicator over the content. I'd at least consider putting a small circle background below/behind it (similar to how pull to refresh looks). Wdyt?

Sure, I'll add it before the final design review.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt
```diff
@@ -101,10 +109,39 @@ class ReaderInterestsViewModel @Inject constructor(
...
+                            Event(SnackbarMessageHolder(R.string.reader_error_request_failed_title))
+                        )
+                    }
+                    updateUiState(
+                        contentUiState.copy(
+                            progressBarVisible = false,
+                            doneButtonUiState = DoneButtonEnabledUiState(titleRes = R.string.reader_btn_done)
+                        )
+                    )
+                } else -> { // Do Nothing
```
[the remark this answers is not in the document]
ashiagr: Removed here 19734be

malinajirka:
Thanks for the change! ;) We can merge it as soon as the CI finishes.

PR #12530, opened 2020-07-27
RI: Save Interests (logged out user/ self-hosted site) 

ashiagr:
Parent Issue: #12027

This PR saves interests locally for logged out user (self-hosted site).

#### To Test

Prerequisite - RI FF flag is enabled

**Selected interests saved locally for logged out user**

0. Fresh install the app or clear app data + enable RI FF flag 
1. Login using a **self hosted site**
2. Go to Reader tab
3. Interests are shown
4. Select few interests
5. Click "Done" button
6. Wait for few seconds
7. Notice that Interests Picker is dismissed
8. Click "Filter" bar  -> "Tags" section in the Following tab
9. Notice that selected interests are shown in the "Tags" section

**Network error not shown on saving interests for logged out user** 

10. Repeat steps 0-3
11. Enable airplane mode
12. Repeat steps 4-9
13. Notice that network error is not shown as tags are saved locally

**Locally saved tags replaced by logged-in user's followed tags** 

1. Login using a **wp.com account**
2. Repeat steps 2-8 above
3. Notice that wp.com account's followed tags are shown in the "Tags" section

#### Notes
- Get user tags (used to determine when to show "Interests Picker") accesses dummy method in the tags repository. So even when Interests are saved, "Interests Picker" will be displayed again the next time. - Fixed in #12548

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderTagActions.java
```diff
@@ -102,51 +106,63 @@ public static boolean addTags(@NotNull final List<ReaderTag> tags,
...
-            // the response will contain the list of the user's followed tags
-            ReaderTagList followedTags = parseFollowedTags(jsonObject);
-            ReaderTagTable.replaceFollowedTags(followedTags);
-            if (actionListener != null) {
-                ReaderActions.callActionListener(actionListener, true);
+        if (!isLoggedIn && AppPrefs.isReaderImprovementsPhase2Enabled()) {
+            // TODO: Revisit logic when ReaderTagRepository's getUserTags is implemented
+            if (!AppPrefs.getReaderRecommendedTagsDeletedForLoggedOutUser()) {
+                // Delete any previously saved recommended followed tags
+                ReaderTagTable.setRecommendedTags(new ReaderTagList());
```
malinajirka, line 113: I believe the "replaceFollowedTags" method removes all the tags from the DB and inserts the provided tags. We might want to consider removing this line. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/services/update/ReaderUpdateLogic.java
```diff
@@ -188,17 +188,23 @@ public void run() {
...
+                        // now replace local topics with the server topics
+                        ReaderTagTable.replaceTags(serverTopics);
+                    }
                     // broadcast the fact that there are changes
                     EventBus.getDefault().post(new ReaderEvents.FollowedTagsChanged(true));
                 }
 
                 // save changes to recommended topics
-                if (mAccountStore.hasAccessToken()) {
+                if (mAccountStore.hasAccessToken() && !AppPrefs.isReaderImprovementsPhase2Enabled()) {
```
malinajirka, line 207: I thought "recommended" tags are used only for a logged out user. However, it seems this if is executed only for a logged in user. Where do we use recommended tags?

WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderTagActions.java
```diff
@@ -102,51 +106,63 @@ public static boolean addTags(@NotNull final List<ReaderTag> tags,
...
-        com.wordpress.rest.RestRequest.Listener listener = jsonObject -> {
-            AppLog.i(T.READER, "add tag succeeded");
-            // the response will contain the list of the user's followed tags
-            ReaderTagList followedTags = parseFollowedTags(jsonObject);
-            ReaderTagTable.replaceFollowedTags(followedTags);
-            if (actionListener != null) {
-                ReaderActions.callActionListener(actionListener, true);
+        if (!isLoggedIn && AppPrefs.isReaderImprovementsPhase2Enabled()) {
+            // TODO: Revisit logic when ReaderTagRepository's getUserTags is implemented
+            if (!AppPrefs.getReaderRecommendedTagsDeletedForLoggedOutUser()) {
```
malinajirka, line 111: I'm still a bit confused if we actually need to delete the "recommended" tags. However, I'd consider adding a comment either here or above the method declartion - explaining the history. Moreover, I'd consider explaining the history even on the PR so it's easier to find the answer why this code was ever added. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderTagActions.java
```diff
@@ -102,51 +106,63 @@ public static boolean addTags(@NotNull final List<ReaderTag> tags,
...
         }
-
-        com.wordpress.rest.RestRequest.Listener listener = jsonObject -> {
-            AppLog.i(T.READER, "add tag succeeded");
-            // the response will contain the list of the user's followed tags
-            ReaderTagList followedTags = parseFollowedTags(jsonObject);
-            ReaderTagTable.replaceFollowedTags(followedTags);
-            if (actionListener != null) {
-                ReaderActions.callActionListener(actionListener, true);
+        if (!isLoggedIn && AppPrefs.isReaderImprovementsPhase2Enabled()) {
```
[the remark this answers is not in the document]
ashiagr: I refactored this part and moved deletion logic to `ReaderUpdateLogic` [here](https://github.com/wordpress-mobile/WordPress-Android/blob/8bf3a53f33cc68a730e585b38d2a324012d3b5ad/WordPress/src/main/java/org/wordpress/android/ui/reader/services/update/ReaderUpdateLogic.java#L193) in PR #12548. If you don't mind, can I create separate methods in that PR?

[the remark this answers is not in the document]
malinajirka: Sure, no problem ;). Thanks

WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderTagActions.java
malinajirka: What I meant was that we could just call `ReaderTagTable.replaceFollowedTags(newTags);` -> which removes the previous and inserts the new -> so we technically didn't need to call `ReaderTagTable.setRecommendedTags(new ReaderTagList());`. However, since you replaced it and we are removing also the posts for these tags, it's all good now;)! Thanks!

WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderTagActions.java
malinajirka: So I think I have a better understanding of the code now. I'm still wondering if we could remove this whole if statement and always invoke `ReaderTagTable.replaceFollowedTags(newTags);`. This wouldn't delete the posts related to tags which are not in the database anymore - however, these would get deleted automatically [here](https://github.com/wordpress-mobile/WordPress-Android/blob/ef7bfa7c8ca6362c3890f4242b897ed29997cd57/WordPress/src/main/java/org/wordpress/android/datasets/ReaderPostTable.java#L208).
I might be missing something very important, so please don't assume it'll work :D. I just thing it's worth asking/exploring as it'd simplify the code. Let me know what you think :). I'll be happy to help exploring in case you can't foresee any issues from the top of your head.

ashiagr: > always invoke ReaderTagTable.replaceFollowedTags(newTags);

This is what [replaceFollowedTags fn](https://github.com/wordpress-mobile/WordPress-Android/blob/87c1f4bf0a1297fdcf412042fd6181beb9cafc6b/WordPress/src/main/java/org/wordpress/android/datasets/ReaderTagTable.java#L97-L99) does
```
// first delete all existing followed tags, then insert the passed ones
String[] args = {Integer.toString(ReaderTagType.FOLLOWED.toInt())};
db.execSQL("DELETE FROM tbl_tags WHERE tag_type=?", args);
addOrUpdateTags(tags);
```

Will it not delete all locally saved followed tags for a logged out user? We're supposed to preserve them.

malinajirka: Ohh, I see ... We want to delete the tags only the first time (that's the reason we can use "replaceFollowedTags" in this scenario). However, we don't want to delete it after. Thanks for clarification!

PR #12532, opened 2020-07-27
Issue 12334/story title layout refactor keyboard fix

mzorz:
Builds on top of https://github.com/wordpress-mobile/WordPress-Android/pull/12520

This PR tries to alleviate the flicker effect observed in #12520 when switching to/from the Tags fragment / Home fragment when in a Story Post.

I'be been able to observe that the fragments need to be re-laid out in order to make space for them while the keyboard is shown or hidden, which is what triggers the `KeyboardResizeViewUtil`'s `ViewTreeObserver.OnGlobalLayoutListener()`

What this PR does is then:
- if we're showing the prepublishing bottom sheet for a Story, we avoid hiding / showing the keyboard while navigating from Home and Tags fragment (as both these fragments in the bottom sheet have EditText components to be used). This way, the lower section of the screen seems to remain the same (it is actually being repainted though, but it gives less of a flicker effect given the size and positioning of the keyboard window is the same)
- as a result of now not hiding / showing the keyboard, the `KeyboardResizeViewUtil`'s `ViewTreeObserver.OnGlobalLayoutListener()` is not triggered when navigating between these fragments and as such, we need to request a layout on the window ourselves, which we are doing in `onResume()` in the respective fragments.

This works well although it arguably adds some complexity (while the PR is short still), so open to thoughts. For users this should be better than seeing the keyboard disappearing/reappearing at once while the flickering for the fadeout/fadein of the home/tags fragments take place, all at the same time.

To test:
1. create a story
2. tap on PUBLISH
3. observe the story title field gets focused, keyboard is shown
4. tap on Tags
5. observe the keyboard remains occupying the same space on the screen (some parameters for the keyboard change as the IME type is requested, for example title caps, etc), also observe the tags fragment gets drawn correctly and you can scroll thorugh the tags list, drag the bottom sheet down, etc
6. tap back arrow on the bottom sheet
7. observe the keyboard remains occupying the same space
8. switch back and forth several times to make sure the fragments get redrawn correctly every time.

See video here https://cloudup.com/c-AZvLXg13Y

**Note**: if we agree on this approach, I'll add some unit tests for the new flags that have been added to ViewModels. Leaving this as a draft PR for now.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeFragment.kt
@@ -65,7 +65,7 @@ class PrepublishingHomeFragment : Fragment() {
             "arguments can't be null."
         }
         if (isStoryPost) {
-            requireActivity().getWindow().getDecorView().requestLayout()
+            requireActivity().window.decorView.requestLayout()
         }
         super.onResume()
     }
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeFragment.kt
@@ -60,6 +60,16 @@ class PrepublishingHomeFragment : Fragment() {
         initViewModel()
     }
 
+    override fun onResume() {
+        val isStoryPost = checkNotNull(arguments?.getBoolean(IS_STORY_POST)) {
+            "arguments can't be null."
+        }
+        if (isStoryPost) {
+            requireActivity().getWindow().getDecorView().requestLayout()
+        }
+        super.onResume()
+    }
+
     private fun initViewModel() {
         viewModel = ViewModelProviders.of(this, viewModelFactory)
                 .get(PrepublishingHomeViewModel::class.java)
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingTagsViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingTagsViewModel.kt
@@ -23,6 +23,7 @@ class PrepublishingTagsViewModel @Inject constructor(
     @Named(BG_THREAD) private val bgDispatcher: CoroutineDispatcher
 ) : ScopedViewModel(bgDispatcher) {
     private var isStarted = false
+    private var dontCloseKeyboard = false
     private lateinit var editPostRepository: EditPostRepository
     private var updateTagsJob: Job? = null
 
@@ -38,8 +39,9 @@ class PrepublishingTagsViewModel @Inject constructor(
     private val _toolbarTitleUiState = MutableLiveData<UiString>()
     val toolbarTitleUiState: LiveData<UiString> = _toolbarTitleUiState
 
-    fun start(editPostRepository: EditPostRepository) {
+    fun start(editPostRepository: EditPostRepository, dontCloseKeyboard: Boolean = false) {
         this.editPostRepository = editPostRepository
+        this.dontCloseKeyboard = dontCloseKeyboard
 
         if (isStarted) return
         isStarted = true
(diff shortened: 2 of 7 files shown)


WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingTagsViewModel.kt
```diff
@@ -38,8 +39,9 @@ class PrepublishingTagsViewModel @Inject constructor(
     private val _toolbarTitleUiState = MutableLiveData<UiString>()
     val toolbarTitleUiState: LiveData<UiString> = _toolbarTitleUiState
 
-    fun start(editPostRepository: EditPostRepository) {
+    fun start(editPostRepository: EditPostRepository, dontCloseKeyboard: Boolean = false) {
```
jd-alexander, line 42: Hey @mzorz I was wondering if it would be better to call this `closeKeyboard` and then flip the value before it's used in `onBackButtonClicked()`? Let me know

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingViewModel.kt
```diff
@@ -67,7 +68,9 @@ class PrepublishingViewModel @Inject constructor(private val dispatcher: Dispatc
     }
 
     private fun navigateToScreen(prepublishingScreen: PrepublishingScreen) {
-        _dismissKeyboard.postValue(Event(Unit))
+        if (prepublishingScreen != TAGS && prepublishingScreen != HOME) {
```
jd-alexander, line 71: Do we want to put a comment here as to why we don't want to do dismissal for these two screens? Also would this be clearer if we just use `prepublishingScreen != PUBLISH` ? LMK 🙏

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingHomeFragment.kt
```diff
@@ -60,6 +60,16 @@ class PrepublishingHomeFragment : Fragment() {
         initViewModel()
     }
 
+    override fun onResume() {
+        val isStoryPost = checkNotNull(arguments?.getBoolean(IS_STORY_POST)) {
+            "arguments can't be null."
+        }
+        if (isStoryPost) {
+            requireActivity().getWindow().getDecorView().requestLayout()
```
jd-alexander, line 68: Nitpick 😅 
```suggestion
            requireActivity().window.decorView.requestLayout()
```

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingTagsViewModel.kt
mzorz: nice, changed the name and meaning in c24f425087

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingViewModel.kt
mzorz: I understand the simplification that `prepublishingScreen != PUBLISH` would bring, however I think it's better to be explicit in the case we in the future add other screens that don't need the dismiss event to be sent. The problem here is more about navigating from/ to Home / Tags which have EditText. Probably we should flag them in some other meaningful way (as it looks arbitrary otherwise), but it starts making the workaround more convoluted.
As I keep thinking about it, I realize a comment in this case would suffice for the most part, so I'll add the suggested optimization and the comment 👍

mzorz: Addressed in fd0019fb4f 👍

mzorz:
Ready for another round @jd-alexander 🙏  - I also added some tests in 98d4979250 😄  let me know 🙇

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingViewModel.kt
jd-alexander: Awesome. Thanks @mzorz for this explanation! The comment makes everything clear now and thanks for implementing my solution. You really didn't have to because your explanation makes sense😅 but I truly appreciate it 😄

PR #12534, opened 2020-07-27
Issue 12524/post list fab menu

### LINKED ISSUE #12524: WordPress Stories :  Add a menu to the Post List FAB button for creating a story post & blog post
Portkey reference issue https://github.com/Automattic/portkey-android/issues/404

### Expected behavior

Story Post & Blog Post should be available in a bottom sheet menu on the Post List page. 

### Actual behavior
Currently, the FAB button just creates a new post. 

1. Open the app. 
2. Click the Post List. 
3. Click the FAB button. 

jd-alexander:
Fixes #12524
PR description and testing details coming shortly.

## Solution
This solution adds the `Story Post` and `Blog Post` options to the Post List - FAB create button. This functionality even though being merged into `develop` will not be available for release builds.

## Testing
1. Click on the Post List. 
2. Click on the FAB button. 
3. You should now see a bottom sheet with a list of options. 

<kbd></kbd>

--------
1. Create a new install of the app.

1. Click on the Post List. 
2. See the "Create a post or story" tooltip for the first launch.

<kbd></kbd>

## Unit Tests
Unit tests will be created in another PR following this one. 

## Reviewing 

## Submitter Checklist


WordPress/src/main/java/org/wordpress/android/ui/posts/PostListCreateMenuFragment.kt
```diff
@@ -0,0 +1,72 @@
...
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View? {
+        return inflater.inflate(R.layout.add_content_bottom_sheet, container)
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        val recyclerView = view.findViewById<RecyclerView>(R.id.content_recycler_view)
```
mzorz, line 39: question, is there any reason why we don't use kotlinx synthetic extension ( `import kotlinx.android.synthetic....` )  and write this more clearly as in `content_recycler_view.layoutManager = ...` and such? let me know 🙏

jd-alexander: Nope, I do use them. This was an oversight because I got this logic from the other implementation. Will resolve. 🙏

jd-alexander: Resolved a344477

WordPress/src/main/java/org/wordpress/android/ui/posts/PostListAction.kt
```diff
@@ -48,6 +50,9 @@ fun handlePostListAction(
         is PostListAction.NewPost -> {
             ActivityLauncher.addNewPostForResult(activity, action.site, action.isPromo, POST_FROM_POSTS_LIST)
         }
+        is PostListAction.NewStoryPost -> {
+            ActivityLauncher.addNewStoryForResult(activity, action.site, STORY_FROM_POSTS_LIST)
```
mzorz, line 54: I think we'll need to show the media picker here instead, something like this:

```
            ActivityLauncher.showPhotoPickerForResult(
                    activity,
                    WP_STORIES_MEDIA_PICKER,
                    action.site,
                    null // this is not required, only used for featured image in normal Posts
            )
```
similar to what method `handleNewStoryAction()` does in WPMainActivity. Also, we'll need to process `onActivityResult()` to get the user's selection and such.

jd-alexander: Hi @mzorz I resolved this in the follow-up comments below. Let me know your thoughts.

WordPress/src/main/java/org/wordpress/android/ui/stories/StoriesMediaPickerResultHandler.kt
```diff
@@ -0,0 +1,71 @@
...
+import org.wordpress.android.util.AppLog
+import org.wordpress.android.util.AppLog.T.UTILS
+
+class StoriesMediaPickerResultHandler {
+    companion object {
+        /* return true if MediaPickerResult was handled */
+        fun handleMediaPickerResultForStories(
+            data: Intent,
+            activity: Activity?,
+            selectedSite: SiteModel?
```
jd-alexander, line 20: Since these are nullable should I do a check to ensure if either or null we don't continue here? The original logic did not check for this so I proceeded as is but I am now just checking with you to get a clearer understanding of how you view this. LMK 🚢

mzorz: Good question - I think it's fine given there are other checks elsewhere that would prevent this from happening here. We could requireNotNull() but given we're not handling errors at this point I think it's ok to let them be handled elsewhere 👍

jd-alexander: Okay thanks much!

PR #12535, opened 2020-07-27
Expose Bookmark Reader Actions

zwarm:
Part of #12039 

This PR continues with exposing `ReaderActions` methods via the repository. Included in this PR are the `addToBookmarked` and `removeFromBookmarked` reader action methods.  

To test:
This was tested locally by forcing requests via the repository. It will be further tested as we move forward with the RI project.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostBookmarkActionUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostBookmarkActionUseCase.kt
@@ -1,0 +1,23 @@
+package org.wordpress.android.ui.reader.repository.usecases
+
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.withContext
+import org.wordpress.android.models.ReaderPost
+import org.wordpress.android.modules.IO_THREAD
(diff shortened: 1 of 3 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostBookmarkActionUseCase.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.withContext
+import org.wordpress.android.models.ReaderPost
+import org.wordpress.android.modules.IO_THREAD
+import org.wordpress.android.ui.reader.actions.ReaderPostActionsWrapper
+import javax.inject.Inject
+import javax.inject.Named
+
+class PostBookmarkActionUseCase @Inject constructor(
```
malinajirka, line 11: We have `ReaderPostBookmarkUseCase` and this name seems quite similar. How would you feel about renaming it to something like "UpdateBookmarkStateUseCase"?

zwarm:
Decided  that we are not moving forward with adding this functionality to the repository. Closing the PR .

PR #12540, opened 2020-07-28
Adds unit tests for GetDiscoverCardsUseCase

malinajirka:
Parent issue #12319 

This PR adds unit tests for GetDiscoverCardsUseCase. I also fixed an issue when a post is not found in local db - we just log a message and continue.

To test:
No need to test anything.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCase.kt
@@ -48,7 +48,7 @@ class GetDiscoverCardsUseCase @Inject constructor(
                                 if (post != null) {
                                     cards.add(ReaderPostCard(post))
                                 } else {
-                                    appLogWrapper.d(READER, "Post from /cards json not found in local db")
+                                    appLogWrapper.d(READER, "Post from /cards json not found in ReaderDatabase")
                                     continue@forLoop
                                 }
                             }
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCase.kt
@@ -3,25 +3,30 @@ package org.wordpress.android.ui.reader.repository.usecases
 import dagger.Reusable
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.withContext
-import org.wordpress.android.datasets.ReaderDiscoverCardsTable
-import org.wordpress.android.datasets.ReaderPostTable
+import org.wordpress.android.datasets.ReaderDiscoverCardsTableWrapper
+import org.wordpress.android.datasets.wrappers.ReaderPostTableWrapper
+import org.wordpress.android.fluxc.utils.AppLogWrapper
 import org.wordpress.android.models.discover.ReaderDiscoverCard
 import org.wordpress.android.models.discover.ReaderDiscoverCard.InterestsYouMayLikeCard
 import org.wordpress.android.models.discover.ReaderDiscoverCard.ReaderPostCard
 import org.wordpress.android.models.discover.ReaderDiscoverCards
 import org.wordpress.android.modules.IO_THREAD
 import org.wordpress.android.ui.reader.ReaderConstants
+import org.wordpress.android.util.AppLog.T.READER
 import javax.inject.Inject
 import javax.inject.Named
 
 @Reusable
 class GetDiscoverCardsUseCase @Inject constructor(
     private val parseDiscoverCardsJsonUseCase: ParseDiscoverCardsJsonUseCase,
+    private val readerDiscoverCardsTableWrapper: ReaderDiscoverCardsTableWrapper,
+    private val readerPostTableWrapper: ReaderPostTableWrapper,
+    private val appLogWrapper: AppLogWrapper,
     @Named(IO_THREAD) private val ioDispatcher: CoroutineDispatcher
 ) {
     suspend fun get(): ReaderDiscoverCards =
             withContext(ioDispatcher) {
-                val cardJsonList = ReaderDiscoverCardsTable.loadDiscoverCardsJsons()
+                val cardJsonList = readerDiscoverCardsTableWrapper.loadDiscoverCardsJsons()
                 val cards: ArrayList<ReaderDiscoverCard> = arrayListOf()
 
                 if (cardJsonList.isNotEmpty()) {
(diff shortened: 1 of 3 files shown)


WordPress/src/test/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCaseTest.kt
```diff
@@ -0,0 +1,102 @@
...
+
+@InternalCoroutinesApi
+@RunWith(MockitoJUnitRunner::class)
+class GetDiscoverCardsUseCaseTest {
+    private lateinit var useCase: GetDiscoverCardsUseCase
+    private val readerDiscoverCardsTableWrapper: ReaderDiscoverCardsTableWrapper = mock()
+    private val parseDiscoverCardsJsonUseCase: ParseDiscoverCardsJsonUseCase = mock()
+    private val mockedJsonArray: JSONArray = mock()
+    private val mockedJsonObject1: JSONObject = mock()
+    private val mockedJsonObject2: JSONObject = mock()
```
jd-alexander, line 33: Just a suggestion here, I understood the tests clearly with these values but I think if we gave these objects some more meaningful names they would probably be easier to read. Something like `InterestsCardJson` and `ReaderPostCardJson`. Let me know what you think, you really don't have to make this change because we are mocking JSON objects so the connotations here are fine 🙏

WordPress/src/test/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCaseTest.kt
```diff
@@ -0,0 +1,102 @@
...
+        whenever(parseDiscoverCardsJsonUseCase.parseSimplifiedPostCard(anyOrNull())).thenReturn(Pair(101, 102))
+        whenever(readerPostTableWrapper.getBlogPost(anyLong(), anyLong(), anyBoolean())).thenReturn(mock())
+        whenever(mockedJsonObject1.getString(ReaderConstants.JSON_CARD_TYPE))
+                .thenReturn(ReaderConstants.JSON_CARD_POST)
+        whenever(mockedJsonObject2.getString(ReaderConstants.JSON_CARD_TYPE))
+                .thenReturn(ReaderConstants.JSON_CARD_INTERESTS_YOU_MAY_LIKE)
+    }
+
+    @Test
+    fun `interest you might like card is transformed into InterestsYouMayLikeCard object`() = test {
```
jd-alexander, line 62: could we add the word json after card so it's clear just by the method name that we are doing deserialization?LMK 🙇

WordPress/src/test/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCaseTest.kt
```diff
@@ -0,0 +1,102 @@
...
+        whenever(mockedJsonObject1.getString(ReaderConstants.JSON_CARD_TYPE))
+                .thenReturn(ReaderConstants.JSON_CARD_INTERESTS_YOU_MAY_LIKE)
+        // Act
+        val result = useCase.get()
+        // Assert
+        assertThat(result.cards[0]).isInstanceOf(InterestsYouMayLikeCard::class.java)
+    }
+
+    @Test
+    fun `post card is transformed into ReaderPostCard object`() = test {
```
jd-alexander, line 73: Same as above 😄

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCase.kt
```diff
@@ -3,33 +3,38 @@ package org.wordpress.android.ui.reader.repository.usecases
...
+    private val appLogWrapper: AppLogWrapper,
     @Named(IO_THREAD) private val ioDispatcher: CoroutineDispatcher
 ) {
     suspend fun get(): ReaderDiscoverCards =
             withContext(ioDispatcher) {
-                val cardJsonList = ReaderDiscoverCardsTable.loadDiscoverCardsJsons()
+                val cardJsonList = readerDiscoverCardsTableWrapper.loadDiscoverCardsJsons()
                 val cards: ArrayList<ReaderDiscoverCard> = arrayListOf()
 
                 if (cardJsonList.isNotEmpty()) {
```
jd-alexander, line 32: Quick question, I normally test for logic like what happens if the `cardJsonList` is empty but in this case, nothing happens so since it's obvious is it not worth testing?

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCase.kt
```diff
@@ -39,8 +44,13 @@ class GetDiscoverCardsUseCase @Inject constructor(
                             ReaderConstants.JSON_CARD_POST -> {
                                 // TODO we might want to load the data in batch
                                 val (blogId, postId) = parseDiscoverCardsJsonUseCase.parseSimplifiedPostCard(cardJson)
-                                val post = ReaderPostTable.getBlogPost(blogId, postId, false)
-                                cards.add(ReaderPostCard(post))
+                                val post = readerPostTableWrapper.getBlogPost(blogId, postId, false)
+                                if (post != null) {
+                                    cards.add(ReaderPostCard(post))
+                                } else {
+                                    appLogWrapper.d(READER, "Post from /cards json not found in local db")
```
jd-alexander, line 51: Since we have multiple databases in the app how about saying: 
```suggestion
                                    appLogWrapper.d(READER, "Post from /cards json not found in ReaderDatabase")
```

WordPress/src/test/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCaseTest.kt
malinajirka: 👍 Agree - fixed in 66f7a9314c5d6d620387ca60ff34466f69fb0f42

WordPress/src/test/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCaseTest.kt
malinajirka: Good idea - fixed in 66f7a9314c5d6d620387ca60ff34466f69fb0f42

WordPress/src/test/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCaseTest.kt
malinajirka: Fixed in 66f7a9314c5d6d620387ca60ff34466f69fb0f42

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCase.kt
malinajirka: Great catch, definitely worth adding a test just to be sure - 66f7a9314c5d6d620387ca60ff34466f69fb0f42

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCase.kt
malinajirka: Fixed in 66f7a9314c5d6d620387ca60ff34466f69fb0f42

malinajirka:
@jd-alexander Thank you for the detailed review. All good tips which improved the code! It's ready for another round.

PR #12543, opened 2020-07-28
Issue 12524/post list create menu tests

### LINKED ISSUE #12524: WordPress Stories :  Add a menu to the Post List FAB button for creating a story post & blog post

jd-alexander:
Fixes #12524 

## Solution

Created unit tests for #12534 

## Testing

Run the unit tests.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PostListMainViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PostListMainViewModel.kt
@@ -50,6 +50,7 @@ import org.wordpress.android.util.SiteUtils
 import org.wordpress.android.util.ToastUtils.Duration
 import org.wordpress.android.util.analytics.AnalyticsTrackerWrapper
 import org.wordpress.android.util.analytics.AnalyticsUtils
+import org.wordpress.android.util.config.WPStoriesFeatureConfig
 import org.wordpress.android.viewmodel.Event
 import org.wordpress.android.viewmodel.SingleLiveEvent
 import org.wordpress.android.viewmodel.helpers.DialogHolder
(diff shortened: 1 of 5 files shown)


jd-alexander:
> Code looks good! :D

Awesome! Thanks @mzorz 

> One note though, I've seen all tests are assuming Stories to be available, but we should account for this by using WPStoriesFeatureConfig and arranging it properly as for example in the tests added here. Can we make the changes and add some tests for the cases when the FeatureConfig.isEnabled() is true/false? 🙏 let me know

Thanks for sharing this 🙏 I didn't add any of these in my tests because due to the nature of how this bottom sheet is integrated with the Post List the `WPStoriesFeatureConfig.isEnabled()` is being used there to show this bottom sheet only when Stories is available. Otherwise it just creates a new post. So yes, all of these tests and functionality assume that Stories is available hence the need to show a bottom sheet with both the option of creating a blog post and a story post. For more details you can check the logic I put here
https://github.com/wordpress-mobile/WordPress-Android/blob/468c9d4ca9b5806fca1b60b88ce8a911d0750479/WordPress/src/main/java/org/wordpress/android/ui/posts/PostsListActivity.kt#L234
 and let me know your thoughts on this if there's something I missed in any way. Again, thanks for a thoughtful review.

jd-alexander:
Thanks for expounding on the testing needs @mzorz I refactored the code a bit and implemented the tests in 
03420a0 so you can review the whole diff changes again to verify my approach. 🙇

PR #12547, opened 2020-07-29
Issue/base appbar styling

khaykov:
This PR sets the base for future AppBar redesign work. It's focused on modifying global theme values and making sure the My Site, Reader, and Notification screens look and behave as expected (excluding nested screens for now).

The big item in this PR is logic related to making AppBars "Lif on Scroll" working with Fragments located in ViewPager. To correctly change the AppBar elevation it needs to know which view is being scrolled. By default, AppBar looks for the first scrollable view inside parent CoordinatorLayout, alternatively, we can explicitly tell AppBar ID of the scrollable view it should observe. Neither options work too well with multiple nested fragments inside ViewPager that all use the same layout with the same ID for scrollable view. For this reason, we need to tell the app

My approach is to set a unique ID to a scrollable view and send it to the parent fragment or activity after the fragment in ViewPager becomes active (when it's selected and you can interact with it). Changing ID of the view after we get a reference to it in Fragment should not be an issue, according to my research and testing. This also works with View Binding of Kotling Extensions - on the first time you get a view by ID it's stored in cache map with the original ID as a key, and on subsequent requests, the cached reference is used - you can confirm this by decompiling the kotlin bytecode.
It tried to extract the code and make it as easy as possible to implement, It works, but I'm open to any suggestions 🙏 

Note about Notification List - because of the `CollapsingToolbarLayout` with specific scroll flags we can set the `liftOnScroll` to `false` and `AppBar` will figure out the elevation based on `CollapsingToolbarLayout` behavior.

Another big change is in the behavior of reader AppBar - now it scrolls away to give you more space for content:

On the visual side, I had no redlines, so I eyeballed the look. @osullivanchris please let me know if anything looks out of place :)

appbar

To test:
- Check My Site, Reader and Notifications screens and make sure they work as expected.
- Make sure "lift on scroll" works when you scroll the content. In Light mode you will see shadow below the AppBar, find Dark Mode AppBars elevation will change it will become lighter/flush on scroll.
- Make sure gravater at My Site screen changes size with scroll.
- Make sure divider between tabs and content disappears on scroll.

PR submission checklist:


khaykov:
Thanks for the feedback, @osullivanchris ! I hope I addressed most of it :)


> Title style should be Headline4 34px per mocks - can you check if that is the case? Looks slightly small to me but might just be the device you screenshotted from

It is 👍 I'm using Headline6 for the collapsed title and Headline4 for expanded.

> Change from My site to My Site - looks strange with the large header not in title case

Done 👍 

> If possible, increase avatar from 24px -> 32px. Found when I was working on this it looks better, gives it more structure. With all the whitespace there, 24 looks a bit lost

It's actually 32dp. I upped it to 40dp (next available size).

> If possible I think tabs on white need a divider line beneath them on initial load - not sure if part of this task.

Yes! This one slipped through the cracks when I was splitting code for the PR. I added it back. I also thought the tabs in Dark Mode need that one too? 

> Filter bar in the reader reader - similarly as discussed put a div below rather than shadow on initial load

> Filter bar stays sticky on scroll in reader - is this intentional?

Filter bar is not a part of AppBar - it's a part of the Following posts list, so it could be tricky to manipulate it based on the scroll. It has 2dp elevation, and I can swap it for a divider, but the elevation will not change on scroll. 

> Padding for the large titles - headings in the mocks I have 16px in from the screen edge. In the build it looks like 24px or more. Doesn't line up nicely with the other icons etc. Looks correct once scrolled. You can see its moving up and to the left on that transition, if I'm right it should just be up, keeping same padding. But maybe its something to do with the component and could be intentional. Let me know.

Default padding for expanded title is larger. I set it manually to 16dp 👍 

> Search icon in reader tab, cog icon in notifications tab are different colours. Should be secondary color. Think reader -> search icon is correct

We are using medium emphasis for all non-toolbar icons. I can change it, but if so we might also consider changing the filter icon color? What do you think?

osullivanchris:
> It's actually 32dp. I upped it to 40dp (next available size).

Great I'll take a look at that

> Yes! This one slipped through the cracks when I was splitting code for the PR. I added it back. I also thought the tabs in Dark Mode need that one too?

Yeah I think it will help in both!

> Filter bar is not a part of AppBar - it's a part of the Following posts list, so it could be tricky to manipulate it based on the scroll. It has 2dp elevation, and I can swap it for a divider, but the elevation will not change on scroll.

Think we might need to discuss the filter bar more, but not necessarily part of this PR

> We are using medium emphasis for all non-toolbar icons. I can change it, but if so we might also consider changing the filter icon color? What do you think?

Ah I think I made a mistake actually! I saw the grey cog in your screenshots and thought I was looking at the notifications app bar. So I thought one app bar had a medium emphasis icon and the other a high emphasis. So the original problem doesn't exist :) We could consider changing filter icon color but not an issue right now. Filter bar seems to be causing problems in general and might need some small rethinking later


Everything looking good for now but haven't been able to log into the build to see the changes on device yet.

osullivanchris:
A few more bits of feedback with the build in my hand

- My site title still incorrect. I'm sure you'll get to it 😄 
- 40px user icon looks better. But looks a little tight to the top. Ever so slightly. I don't want to add too much customisation though. Do you agree? 
- Similarly, would it be possible to scale it down to the original 32px on scroll? It now looks too big next to the collapsed header. Know this is asking a lot...but think it could look really good.
- Divider line below tabs is definitely helping. However...again might be a pain - I think we should hide the divider line on scroll and just have the shadow. Doesn't look right to me to have line and shadow
- The 16px left padding looks much better than 24 👍 
- On reader, scroll so header is hidden. Header has no background at any time. Scroll back, it appears with a grey background, which then disappears again when you get to the top. Thinking we should always use the grey background, like in notifications, even if it doesnt get shadow right now due to filter bar. Slightly tricky one. Open to your suggestion here.

khaykov:
@osullivanchris 

> My site title still incorrect. I'm sure you'll get to it 😄

Not sure what was that, but it should be correct :)

> 40px user icon looks better. But looks a little tight to the top. Ever so slightly. I don't want to add too much customisation though. Do you agree?
> Similarly, would it be possible to scale it down to the original 32px on scroll? It now looks too big next to the collapsed header. Know this is asking a lot...but think it could look really good.

Done 👍 It's 32dp in collapsed state and 40dp in expanded. Let me know what you think :)

[Image from Gyazo](https://gyazo.com/6346dadd1977bf21ffbd0291cfd1d456)

> Divider line below tabs is definitely helping. However...again might be a pain - I think we should hide the divider line on scroll and just have the shadow. Doesn't look right to me to have line and shadow

Fixed 👍
[Image from Gyazo](https://gyazo.com/369fb717e30d39c80a9f744114d1c8bc)

> On reader, scroll so header is hidden. Header has no background at any time. Scroll back, it appears with a grey background, which then disappears again when you get to the top. Thinking we should always use the grey background, like in notifications, even if it doesnt get shadow right now due to filter bar. Slightly tricky one. Open to your suggestion here.

This works as intended - when you scroll up in initial state when appbar is expanded, the content does not go below the appbar, you actually push the appbar above the screen. When you scroll back, appbar appears, and since the content of reader is below it you see it as elevated. When you get to the top, the content of the reader gets out from below the appbar, and appbar elevation return to 0. Hope this makes sense :)

osullivanchris:
> Done 👍 It's 32dp in collapsed state and 40dp in expanded. Let me know what you think :)

This looks great in the gif above!

> Fixed (divider/shadow)

Nice! Good job figuring that one out, glad its possible.

> Hope this makes sense :)

Yeah I think so!

khaykov:
Thanks for taking a look @planarvoid !

I fixed the toolbar reappearing on configuration change 👍 

Regarding:

> I know it's intentional but I find it weird that the toolbar is only hidden on scroll in the Reader and nowhere else. I also don't see a reason to make it smaller if it disappears on scroll 🤷

I'll let @osullivanchris answer this :D

khaykov:
@planarvoid if there are no issues with PR, do you mind if we merge it? It's a feature branch, and we are planning to do a final review/tweaking with design team, before shipping it. This PR blocks the rest of the work, so I'm would really love to get it merged :)

osullivanchris:
> I know it's intentional but I find it weird that the toolbar is only hidden on scroll in the Reader and nowhere else. I also don't see a reason to make it smaller if it disappears on scroll 🤷

fair question! We're hiding it in the reader to allow the user to focus more on the content especially with large amounts of text. We have quite a lot of navigation taking up a large part of the screen.

I still think the size reduction is a really nice animation so there's no harm in keeping it here, consistent with the other two tabs. And as the team work on reader they can re-evaluate what they want to hide / show on scroll.

PR #12549, opened 2020-07-29
Memory leak fix in photo picker adapter

planarvoid:
While investigating the PhotoPickerAdapter for the media picker consolidation project, I've noticed that there was a non-static async task loading the device files. This task was leaking context. I'm extracting and fixing this in this PR. This is a first step of a possible refactoring where we can move the loading to a viewmodel.
There should be no functionality changes in this PR. One thing is that I moved the check whether the loaded list is the same as the previous data out of the async task to the callback. I think this is not a big deal but I'll change this in the next step when 
To fix the context leak I've replaced AsyncTask with coroutines

To test:
- Go to Gutenberg editor
- Add image/video block
- Pick "Choose from device"
- See that the media picker works as expected

PR submission checklist:


planarvoid:
@develric thanks for the review! I've managed to fix the lint issue so I think this PR should be ready to merge now (once you check the changes from AsyncTask to coroutines)

PR #12555, opened 2020-07-29
Improve site icon edit message for self hosted

develric:
Partially covers #9179

Currently we need some improvements [at least on the API side](https://github.com/wordpress-mobile/WordPress-Android/issues/9179#issuecomment-664489409) to allow the site icon to be set for pure self-hosted sites.

Meanwhile seems reasonable to improve the user message when attempting to change site icon for self-hosted.

**NOTE:** while testing this PR I noticed that the set of a site icon from the WordPress Media Library is not working. Checked this issue is in develop from at least [this commit](dfac5d1710c5d85b875a3f0084778225b82ac779) (not saying that one introduced it, it's just how back I went). It's working fine in current beta 15.4-rc-1. I will address this in a separate PR.

### Pure self-hosted
- setup a pure self-hosted site (like using https://jurassic.ninja/) and add it the the app by address
- Check that you get the below message and you can close it without getting any icon update

<p align="center">
  
</p>

### WP.COM/JP/ATOMIC with admin account
- Setup one of the above site type
- Login into the app with an admin account role (or a role above contributor)
- Try to Add/Edit the site icon and check you get the following

| Screenshot_1596041288 | Screenshot_1596041770 |
|---|---|

- Check you are able to Add/Edit the site icon as usual

### WP.COM/JP/ATOMIC with contributor account
- Same step as before but you should get the following message(s)

| Screenshot_1596048222 | Screenshot_1596048274 |
|---|---|

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #12556, opened 2020-07-29
Activity Result not set when changing site icon


develric:
While testing #12555 I found an issue with setting the site icon of a site from the WP Media Library.

### Expected behavior
When setting a site icon in My Site, selecting an image from the WP Media Library I would expect the site icon to be changed accordingly

### Actual behavior
When setting a site icon in My Site, selecting an image from the WP Media Library the site icon is not changed and nothing happens.

### Steps to reproduce the behavior
- Go in My Site for a WP.com/JP site
- Tap on the site icon and acknowledge the dialog that appears
- Go to WP Media Library from the screen that appears
- Select an image and note that the site icon is not updated

mzorz:
Closing via #12558

PR #12558, opened 2020-07-29
Changing the doMediaIdsSelected function logic to allow returning correct activity result.

develric:
Fixes #12555 

The linked issue is referred specifically to setting a site icon, but I believe this can have a broader impact when picker is used with `mBrowserType` not being a `WP_STORIES_MEDIA_PICKER` nor a `FEATURED_IMAGE_PICKER`. 

Since I'm not familiar with WP Stories code I'm assuming things here so this PR is more on the side of making us aware of the issue than pretending this is a full fix of all possible use cases 😊 

PS: issue seems present in current develop, current beta seems ok so we should be fine fixing it before next code freeze AFAIU

### To test

For the site icon part you can follow the linked #12556 issue steps but as said this is more a trigger so we can check also other picker cases.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoPickerActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoPickerActivity.java
@@ -328,23 +328,25 @@ public class PhotoPickerActivity extends LocaleAwareActivity
 
     private void doMediaIdsSelected(ArrayList<Long> mediaIds, @NonNull PhotoPickerMediaSource source) {
         if (mediaIds != null && mediaIds.size() > 0) {
-            if (mBrowserType == MediaBrowserType.FEATURED_IMAGE_PICKER) {
-                // if user chose a featured image, track image picked event
-                mFeaturedImageHelper.trackFeaturedImageEvent(
-                        FeaturedImageHelper.TrackableEvent.IMAGE_PICKED,
-                        mLocalPostId
-                );
-
+            if (mBrowserType == MediaBrowserType.WP_STORIES_MEDIA_PICKER) {
+                // TODO WPSTORIES add TRACKS (see how it's tracked below? maybe do along the same lines)
                 Intent data = new Intent()
-                        .putExtra(EXTRA_MEDIA_ID, mediaIds.get(0))
+                        .putExtra(MediaBrowserActivity.RESULT_IDS, ListUtils.toLongArray(mediaIds))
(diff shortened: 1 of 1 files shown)


WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoPickerActivity.java
```diff
@@ -328,23 +328,25 @@ public void doNext(Uri uri) {
...
     private void doMediaIdsSelected(ArrayList<Long> mediaIds, @NonNull PhotoPickerMediaSource source) {
         if (mediaIds != null && mediaIds.size() > 0) {
-            if (mBrowserType == MediaBrowserType.FEATURED_IMAGE_PICKER) {
-                // if user chose a featured image, track image picked event
-                mFeaturedImageHelper.trackFeaturedImageEvent(
-                        FeaturedImageHelper.TrackableEvent.IMAGE_PICKED,
-                        mLocalPostId
-                );
-
+            if (mBrowserType == MediaBrowserType.WP_STORIES_MEDIA_PICKER) {
```
mzorz, line 331: is there a reason why the order of the checks was flipped? Seems to have the same results so I'm curious

WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoPickerActivity.java
```diff
@@ -328,23 +328,25 @@ public void doNext(Uri uri) {
...
-                        .putExtra(EXTRA_MEDIA_ID, mediaIds.get(0))
+                        .putExtra(MediaBrowserActivity.RESULT_IDS, ListUtils.toLongArray(mediaIds))
+                        .putExtra(ARG_BROWSER_TYPE, mBrowserType)
                         .putExtra(EXTRA_MEDIA_SOURCE, source.name());
                 setResult(RESULT_OK, data);
                 finish();
             } else {
-                // TODO WPSTORIES add TRACKS (see how it's tracked above? maybe do along the same lines)
+                // if user chose a featured image, track image picked event
+                if (mBrowserType == MediaBrowserType.FEATURED_IMAGE_PICKER) {
```
mzorz, line 341: Oh I can see this section here is enclosed more tightly to only call the `trackFeaturedImageEvent()` method when the browserType is `FEATURED_IMAGE_PICKER` and then leave all of the rest (`else` clause) to always set a result, understood the difference now, forget the other comment 👍

PR #12562, opened 2020-07-30
RI: Drop unused recommended tags table

ashiagr:
This PR drops `tbl_tags_recommended` and removes dead code that used this table.

`tbl_tags_recommended` is not used for both logged in and logged out user. This was noticed in https://github.com/wordpress-mobile/WordPress-Android/pull/12530#discussion_r462791640.

**Case 1** Logged-out user
Recommended tags are parsed  [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/reader/services/update/ReaderUpdateLogic.java#L161) and inserted into `tbl_tags` as "Followed" tags [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/reader/services/update/ReaderUpdateLogic.java#L194-L195) for a logged out user. 
Recommended tags are not accessed from `ReaderTagTable.getRecommendedTags` but directly accessed from `tbl_tags` [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderTagAdapter.java#L162) and [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubFilterViewModel.kt#L117).

**Case 2** Logged-in user
Recommended tags are unnecessarily refreshed in the `tbl_tags_recommended` table for a logged in user [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/reader/services/update/ReaderUpdateLogic.java#L200-L209). These tags are needed only for logged out users, and as verified in **Case 1**, they don't use the table. So this is dead code and removed in 2d2a41b.

#### To test:

Login using wp.com account as well as self-hosted site. 
Verify that tags are displayed properly under Reader => Following tab => Filter bar => Tags section in the bottomsheet.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/models/ReaderTagType.java
```diff
@@ -4,24 +4,20 @@
...
     INTERESTS;
 
     private static final int INT_DEFAULT = 0;
     private static final int INT_FOLLOWED = 1;
-    private static final int INT_RECOMMENDED = 2;
-    private static final int INT_CUSTOM_LIST = 3;
-    private static final int INT_SEARCH = 4;
-    private static final int INT_BOOKMARKED = 5;
-    private static final int INT_INTERESTS = 6;
+    private static final int INT_CUSTOM_LIST = 2;
```
malinajirka, line 13: ⚠️ I believe we can't just change the values unless we do a full db migration. Unless I'm wrong, these values are used in the ReaderPostTable. I'd suggest simply leaving a comment - something like "// 2 - we used to use 2 for recommended tag - don't reuse it". Wdyt?

ashiagr: 🤦‍♀️, thanks for checking this. I'll shortly push a commit with the fix.

malinajirka: No worries ;). Thanks for the fix.

PR #12563, opened 2020-07-30
Issue/12319 clear cards cache when necessary

malinajirka:
Parent issue #12319 

Merge instructions
1. Review this PR
2. Check if https://github.com/wordpress-mobile/WordPress-Android/pull/12572 has been merged. If it was, update the target branch to develop.
5. Merge this PR

This PR takes care of clearing cached posts fetched from the /cards endpoint after we force fetch the first page. I don't like this solution, but I couldn't come up with anything better. We use a similar approach for "Bookmark/SavedForLater" posts. 

The solution introduces a new local only ReaderTag. All the posts fetched from the /cards endpoint are stored under this tag so it's easy to filter them and purge the cache after we fetch a new data.

To test:
1. Clear app data
1. Turn on the RI Feature Flag
2. Go to "Discover" tab and notice there are posts displayed
3. Go to following tab just to make sure you fetch some other posts
4. Open Stetho and check the first ten posts (ordered by rowId) - these are the posts for the discover tab
5. Kill the app and start it again
6. Open Discover
7. Open Stetho and notice the first ten posts are not there and they are now on the last 10 ids (they were removed and re-insterted)

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/datasets/ReaderDiscoverCardsTable.kt
+++ b/WordPress/src/main/java/org/wordpress/android/datasets/ReaderDiscoverCardsTable.kt
@@ -21,6 +21,10 @@ object ReaderDiscoverCardsTable {
         db.execSQL("DROP TABLE IF EXISTS tbl_discover_cards")
     }
 
+    fun reset() {
+        reset(getWritableDb())
+    }
+
     fun reset(db: SQLiteDatabase) {
         AppLog.i(AppLog.T.READER, "resetting ReaderDiscoverCardsTable")
         dropTables(db)
--- a/WordPress/src/main/java/org/wordpress/android/models/ReaderTag.java
+++ b/WordPress/src/main/java/org/wordpress/android/models/ReaderTag.java
@@ -228,6 +228,13 @@ public class ReaderTag implements Serializable, FilterCriteria {
         return true;
     }
 
+    /**
+     * Local only tag used to identify post cards for the "discover" tab in the Reader.
+     */
+    public static ReaderTag createDiscoverPostCardsTag() {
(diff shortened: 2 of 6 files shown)


malinajirka:
Thanks so much for the review @jd-alexander!

>I would approve the PR because I tested it using Stetho and got the expected results 😄 I didn't quite get this part of your comment they are now on the last 10 ids (they were removed and re-insterted)

yes sorry, I agree it wasn't clear enough.

The database after 
- step `3` should have 10 posts with ids 1 - 10
- step `4` should have +- 30 posts ( ids 1-10 are discover posts, ids 11-30 are following posts) [there might be less or more following posts it doesn't really matter)
- step `7` should have +- 30 posts, however ids 1-10 shouldn't be in the database at all (these posts were removed) and they should be re-inserted with ids `31-40`

Let me know if it's clearer now.

zwarm:
> * step `3` should have 10 posts with row ids 1 - 10 (discover posts)

@malinajirka - I was jumping in to lend a hand while @jd-alexander laptop was acting up, but my row numbers are not matching up to your steps. After step 3 I have more than 10 rows in the db (+/- 30), but maybe that's because the ReaderPostService is started earlier on in the app. Either way, I do see the discover posts and they do get removed from the db when following the remaining test instructions, so if you concur that the behavior is correct we can merge this PR. 

@jd-alexander - I hope you don't mind that I jumped in & if so, tell me to stay out of your way :)

malinajirka:
@zwarm Thanks for taking over this PR!!

>Either way, I do see the discover posts and they do get removed from the db when following the remaining test instructions

As you mentioned, this is the important part:). I think we can merge the PR!
(it's possible that the app displays a different tab (eg. following) when you log in -> that might be the reason why you see more posts in the db.

PR #12482, opened 2020-07-20
[Modal Layout Picker] Create Modal window for Modal Layout Picker

antonis:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2419

**Related PRs:**
- `gutenberg-mobile`: https://github.com/wordpress-mobile/gutenberg-mobile/pull/2494
- `gutenberg`: https://github.com/WordPress/gutenberg/pull/24061

## Description
Adds feature flags and a container view for the Modal Layout Picker

### Classic Editor and Gutenberg Editor with Modal Layout Picker Enabled.
- Create a new page from My Site
    - Create a new page by clicking on the Floating Toolbar Button on the My Site Page then select "Site Page"
    - Expect to see the modal layout picker container
    - Select "Create Blank Page"
    - Expect to see the page editor
- Create a new page from Pages.
    - Create a new page by clicking on the fab icon on the Pages screen.
    - Expect to see the modal layout picker container
    - Select "Create Blank Page"
    - Expect to see the page editor
- Create a new page from Pages when the site has no pages.
    - Create a new page by clicking on the 'Create a page' button.
    - Expect to see the modal layout picker container
    - Select "Create Blank Page"
    - Expect to see the page editor

### On the Modal Layout Picker
- Selecting the back icon at the top left should close the modal without creating a page.
- Swiping down around the "Grip" in the top middle of the modal should close the modal without creating a page.
- Scrolling up in the stubbed rows should collapse the header
- Toggle Dark mode and check the style seems consistent with other views

## Screenshots:
| Container | Action |
|---|---|
|device-2020-07-22-115223|device-2020-07-22-114828|
|---|---|
| Dark Mode | Collapsed |
|---|---|
|device-2020-07-22-115242|device-2020-07-22-115310|

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #2452, opened 2015-03-23
Crash report 3.8: NPE in StringUtils.getMd5IntHash


maxme:
```
java.lang.RuntimeException: Unable to start activity ComponentInfo{org.wordpress.android/org.wordpress.android.ui.reader.ReaderPostListActivity}: java.lang.NullPointerException
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2121)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2146)
       at android.app.ActivityThread.access$700(ActivityThread.java:140)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1238)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loop(Looper.java:137)
       at android.app.ActivityThread.main(ActivityThread.java:4944)
       at java.lang.reflect.Method.invokeNative(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:511)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1038)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:805)
       at dalvik.system.NativeStart.main(NativeStart.java)
Caused by: java.lang.NullPointerException
       at org.wordpress.android.util.StringUtils.getMd5IntHash(StringUtils.java:80)
       at org.wordpress.android.util.StringUtils.getMd5Hash(StringUtils.java:90)
       at org.wordpress.android.util.GravatarUtils.blavatarFromUrl(GravatarUtils.java:79)
       at org.wordpress.android.util.GravatarUtils.blavatarFromUrl(GravatarUtils.java:76)
       at org.wordpress.android.ui.reader.views.ReaderBlogInfoView.showBlogInfo(ReaderBlogInfoView.java:118)
       at org.wordpress.android.ui.reader.views.ReaderBlogInfoView.loadFeedInfo(ReaderBlogInfoView.java:56)
       at org.wordpress.android.ui.reader.ReaderPostListFragment.loadBlogOrFeedInfo(ReaderPostListFragment.java:1274)
       at org.wordpress.android.ui.reader.ReaderPostListFragment.onActivityCreated(ReaderPostListFragment.java:444)
       at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:848)
[...]

PR #2446, opened 2015-03-20
fix #2439: check for null cursor

maxme:

PR #12564, opened 2020-07-30
[Modal Layout Picker] Create Layout Picker sections and thumbnail UI using static data

antonis:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2436

**Related PRs:**
- **Depends on**: https://github.com/wordpress-mobile/WordPress-Android/pull/12482
- Followed by: https://github.com/wordpress-mobile/WordPress-Android/pull/12573 (and other [Modal Layout Picker tasks](https://github.com/wordpress-mobile/gutenberg-mobile/issues?q=is%3Aissue+is%3Aopen+%5BModal+Layout+Picker%5D))

## Description

This PR implements the presentation of static layout data as part of the effort to rework the Starter Page Template Picker for new page.
In detail it implements the following:
+ Layouts grouped into sections based on their category and horizontally scrollable within their category rows. 
+ Preview image on each layout cell
+ Selecting a Layout:
    + Updates the border-color
    + Displays the checkmark
    + Changes the bottom display to have buttons for "Preview" and "Create Page"
    + Deselects any currently selected cells
+ Deselecting a Layout:
    + Reverts the border-color.
    + Hides the checkmark.
    + Revert the bottom Buttons to be "Create Blank Page"
+ Handles Dark Mode for the new UI items. 

## To test:

**Note**: For demo/testing purposes, the categories repeat the same layout 30 times because the static data only has one layout per category. 

### General Navigation
1. Category rows should be horizontally scrollable within their sections with a preview image
    - Pick a row and scroll horizontally. 
    - Expect the category to scroll to the end and back again
1. Scrolling down and back up on the page should retain the scroll position of the category row
    - Pick a row and scroll horizontally.
    - Scroll vertically.
    - Scroll back to your reference row and expect it to be in the same location
1. Scrolling down and back up on the page should retain the selected layout
    - Select a layout
    - Scroll vertically and/or horizontally.
    - Scroll back to your reference layout and expect it to be selected still
### Preview Images
1. Layouts that haven't loaded yet should display a preview placeholder.
1. Images should be cached for use offline
    - Open the Layout picker and allow some images to download
    - Leave the layout picker
    - Either go offline or monitor the network calls
    - Open the layout picker
    - Expect the images for the layout picker to be pulled from the cache
### Selecting a Layout
1. Selecting a layout should highlight the cell and add a checkmark
    - Open the layout picker and select a cell
1. Only one layout should be selected at any given time
    - Open the layout picker and select a cell
    - Select a different cell (either in the same row or a different row)
    - Expect the old cell to deSelect and the new one to be selected
1. Selecting a Layout should show the buttons "Preview" and "Create Page"
    - Open the layout picker and select a cell
    - *Note*: The "Preview" and "Create Page" actions will be addressed by [#2452](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2452) and [#2446](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2446)
1. Deselecting a layout should revert the bottom bar to "Create Blank Page"
    - Open the layout picker and select a cell
    - Deselect the cell by tapping on it again
    - Expect the bottom buttons to revert to "Create Blank Page"

## Screenshots:
| Top scroll | Middle scroll - selection |
|---|---|
|device-2020-08-07-134403|device-2020-08-07-134423|

| Top scroll (Dark Mode) | Middle scroll - selection (Dark Mode) |
|---|---|
|device-2020-08-07-134521|device-2020-08-07-134541|

|Landscape - selection |
|---|
|device-2020-08-07-134455|

| Image placeholder | Action |
|---|---|
|device-2020-08-07-134051|w|

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mlp/GutenbergPageLayouts.kt
```diff
@@ -0,0 +1,155 @@
+package org.wordpress.android.ui.mlp
+
+//
+// TODO: THIS SHOULD BE MOVED IN RN BRIDGE?
```
chipsnyder: I put it in the bridge for the iOS side that way ideally we can find a way to share these static layouts in the future.

antonis:
Hello @iamthomasbishop 👋 ,
I have drafted the layout picker sections using static data.
You can test this Pull Request by downloading the [APK here](https://64137-9306568-gh.circle-artifacts.com/0/Artifacts/WordPress-pr-12564-build-64137.apk).
Thank you in advance for your feedback 🙇

chipsnyder (commented):
Hey @antonis, great work here too I only found a couple of small issues while testing.

## Selection is maintained after closing the view
It seems like the selection is preserved if I leave and come back. 
1. Open the layout picker
1. Select a layout
1. Select the up button
1. Come back
Notice: The layout is still selected.

RetainedSelectedRowOnBack 2020-08-03 16_39_44

## Scroll position going down and back up
1. Scroll horizontally in a category (such as the first row)
1. Scroll to the bottom
1. Come back to the top
Notice the first row is reset to the default scroll position

Looking at the Google Play store if I do the same thing they keep track of the scroll position. I've noticed other apps like Netflixs will do this as well. What are your thoughts on cloning that behavior?

antonis:
Hey @chipsnyder 👋 ,
thank you for the feedback. I updated the PR tackling both issues.

WordPress/src/main/java/org/wordpress/android/ui/mlp/GutenbergPageLayouts.kt
```diff
@@ -0,0 +1,164 @@
...
+    val description: String,
+    val emoji: String
+)
+
+/**
+ * Static page layout data factory
+ * NOTE: This implementation will be deprecated and the layouts for self hosted sites will be provided by a new
+ * public API endpoint (/common-block-layouts)
+ */
+object GutenbergPageLayoutFactory {
```
antonis, line 30: @chipsnyder My assumption is that this layout factory implementation will be used only temporarily for testing this PR and the new endpoint @enejb is creating will replace it.
Thus there is no need to move this in RN Bridge at this point.

WordPress/src/main/java/org/wordpress/android/ui/mlp/GutenbergPageLayouts.kt
```diff
@@ -0,0 +1,164 @@
+package org.wordpress.android.ui.mlp
+
+data class GutenbergPageLayouts(
+    val layouts: List<GutenbergLayout>,
+    val categories: List<GutenbergLayoutCategory>
+) {
+    fun layouts(categorySlug: String) =
+            layouts.filter { l -> l.categories.any { c -> c.slug == categorySlug } }
```
malinajirka, line 8: We usually don't have any logic in data classes, but I guess this logic is simple enough. However, the name of the method seems a bit confusing - it doesn't contain any verb and it's not clear what it does (my first guess would be that it returns `layouts` variable without any changes). How would you feel about calling it something like "getFilteredLayouts"?

WordPress/src/main/java/org/wordpress/android/ui/mlp/LayoutsAdapter.kt
```diff
@@ -0,0 +1,83 @@
...
+    ) {
+        val layout = layouts[position]
+        imageManager.load(holder.preview, ImageType.THEME, layout.preview, FIT_CENTER)
+        holder.selected.setVisible(layout.selected)
+        holder.preview.contentDescription = if (layout.selected) context.getString(
+                R.string.mlp_layout_selected,
+                layout.title
+        ) else layout.title
+        holder.container.setOnClickListener {
+            selectionListener.layoutTapped(layout)
```
malinajirka: I'm not sure if I've seen this usage/approach anywhere else in the codebase. What we usually do is that we add a onClick lambda to the uistate - [example](https://github.com/wordpress-mobile/WordPress-Android/blob/4585d19679b9f651d39dfd3372de544a093b221c/WordPress/src/main/java/org/wordpress/android/ui/sitecreation/segments/SiteCreationSegmentViewHolder.kt#L70). How would you feel about something like that?

WordPress/src/main/java/org/wordpress/android/ui/mlp/LayoutsAdapter.kt
```diff
@@ -0,0 +1,83 @@
...
+        }
+        selectionListener.selectedItemData.observe(selectionListener.lifecycleOwner, Observer {
```
malinajirka: Imo viewHolder shouldn't be responsible for listening to state changes - ideally, it'd just receive data and render them. When something needs to be changed the vm updates the data - passes them to the adapter - the viewholder re-renders them. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/mlp/LayoutsAdapter.kt
```diff
@@ -0,0 +1,83 @@
...
+        val v = LayoutInflater.from(parent.context)
+                .inflate(R.layout.modal_layout_picker_layouts_card, parent, false)
+        return ViewHolder(v)
+    }
+
+    override fun getItemCount(): Int {
+        return layouts.size
+    }
+
+    override fun onBindViewHolder(
```
malinajirka: We might want to consider extracting the viewHolder into a separate class to follow the SRP. Something like [this](https://github.com/wordpress-mobile/WordPress-Android/blob/b71ea8fe1ba4f4f3666881bfd6ad50066b9ef92f/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderPostViewHolder.kt#L31) just way simpler. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/mlp/ModalLayoutPickerFragment.kt
```diff
@@ -121,6 +135,15 @@ class ModalLayoutPickerFragment : BottomSheetDialogFragment() {
                     }
                 }
         )
+
+        viewModel.selectedLayoutSlug.observe(this,
```
malinajirka: NitPick: Ideally the VM would tell the view layer what to do -> so instead of passing a selection and deciding what to do with it here. We could pass a buttonsUiState with (blankPageVisible = false,... ). Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/mlp/ModalLayoutPickerListDiffCallback.kt
```diff
@@ -9,10 +15,24 @@ class ModalLayoutPickerListDiffCallback(
...
 ) : Callback() {
+    object Payload
+
     override fun areItemsTheSame(oldItemPosition: Int, newItemPosition: Int): Boolean {
         val newItem = newList[newItemPosition]
         val oldItem = oldList[oldItemPosition]
-        return oldItem == newItem
+        if (newItem.type == oldItem.type) {
+            return when (newItem.type) {
+                TITLE -> (newItem as? Title)?.visible == (oldItem as? Title)?.visible
```
malinajirka: I'm not sure I understand this condition for `title`. If one title is invisible and the other is visible that means the list item can't be recycled? Could you please clarify it a bit, thanks!

WordPress/src/main/java/org/wordpress/android/ui/mlp/ModalLayoutPickerListDiffCallback.kt
```diff
@@ -9,10 +15,24 @@ class ModalLayoutPickerListDiffCallback(
...
+                SUBTITLE -> true
+                CATEGORIES -> true // for now
+                LAYOUTS -> {
+                    val newLayoutCategory = (newItem as? LayoutCategory)
+                    val oldLayoutCategory = (oldItem as? LayoutCategory)
+                    return newLayoutCategory?.description == oldLayoutCategory?.description &&
```
malinajirka: Similar to titles - if two categories have different descriptions or sizes they can't be recycled?

WordPress/src/main/res/drawable/ic_layout_selection_overlay.xml
```diff
@@ -0,0 +1,11 @@
+<vector android:autoMirrored="true" android:height="230dp"
+    android:viewportHeight="230" android:viewportWidth="148"
+    android:width="148dp" xmlns:android="http://schemas.android.com/apk/res/android">
+    <path android:fillColor="#00000000"
+        android:pathData="M4,1L144,1A3,3 0,0 1,147 4L147,226A3,3 0,0 1,144 229L4,229A3,3 0,0 1,1 226L1,4A3,3 0,0 1,4 1z"
+        android:strokeColor="#C9356E" android:strokeWidth="2"/>
```
malinajirka: We usually include the width of the icon in its name eg "ic_trash_white_24dp.xml". We might want to consider following this convention for consistency.

We also try to use only colors from the colors.xml - ideally re-use those which are already defined there under the "Semantic" label. Eg for this color we'd use `@color/accent_50`. 

Wdyt?

WordPress/src/main/res/values/dimens.xml
```diff
@@ -540,12 +540,20 @@
...
     <dimen name="mlp_bottom_shadow_height">5dp</dimen>
     <dimen name="mlp_bottom_shadow_height_overlap">-5dp</dimen>
     <dimen name="mlp_bottom_button_vertical_margin">16dp</dimen>
     <dimen name="mlp_bottom_button_horizontal_margin">20dp</dimen>
+    <dimen name="mlp_bottom_button_horizontal_central_margin">16dp</dimen>
     <dimen name="mlp_categories_row_height">52dp</dimen>
-
+    <dimen name="mlp_layout_card_height">204dp</dimen>
+    <dimen name="mlp_layout_card_width">136dp</dimen>
+    <dimen name="mlp_layout_card_margin_start">20dp</dimen>
```
malinajirka, line 552: We might want to consider referencing the existing margins. Also margin "29dp" seems a bit odd, is that intentional?

malinajirka (commented):
Really great job @antonis! 

I've tested the app and it works great. The only thing I noticed is that we lose horizontal scroll position on an orientation change (I'm not sure it's worth fixing though). 

I've left some in-code comments. Most of them are very minor. The only one I'd consider a bit more important is the one regarding `LayoutSelectionListener` - listener implementing lifecycleOwner and livedata seems very strange to me. Let me know what you think ;). Thanks

WordPress/src/main/res/values/dimens.xml
antonis: I agree with you @malinajirka. I'm changing this to reference existing margins (`@dimen/margin_medium...large etc`) where possible.
I got the 29dp margin from Figma, but you are right. I made a mistake since I included the separator line. The correct value is 28dp and is not needed at all ;)
|margin with the seperator line|margin without the separator|
|---|---|
|Screenshot 2020-08-27 at 3 18 45 PM|Screenshot 2020-08-27 at 3 38 30 PM|

Fixed with bfe6a0104be54e94a4b0c8e591d733ec6ea56266

WordPress/src/main/java/org/wordpress/android/ui/mlp/GutenbergPageLayouts.kt
antonis: I agree with you. Fixed with a31fc15ff9cd4399abe47715dcec84af30a1d617

WordPress/src/main/res/drawable/ic_layout_selection_overlay.xml
antonis: Fixed with 605aa7aa4c65902d1a250d63af40042ff8e8a27b

iamthomasbishop:
@antonis Apologies on the delays for design review, and I hope this is the right place to drop review notes -- if not, let me know and I will drop them in the appropriate place 😄 

I've taken the latest build linked above for a spin, and jotted down some notes and feedback. It's feeling pretty close already and I'll give it another run tomorrow, but here are my notes:

* I noticed when a row of thumbnails rests below the surface containing the category chips/buttons, I'm unable to interact with or even scroll the categories container -- instead, the thumbnails row is scrolled
* Would it be possible to style the sheet in its full-height state with rounded corners, and to hold the same scrim/backdrop style that it has when it's not fully docked? [Here's a visual for example](https://cloudup.com/cfQBWkVGFXi).
* Can we by any chance replicate the same multi-select capabilities that we're rolling with on iOS?
* I'm seeing the large title disappear every once in a while, especially when interacting with the category chips.
* The thumbnail rows have a bit too much padding on the bottom, but it sounds like you've sorted that out above.
* Would it be possible to apply a smooth transition to the style changes between an un-selected and selected thumbnail image? I believe Material's recommendation for such a transition is an `easeOut` transition with a `100ms` duration, but if we're already using transitions elsewhere natively, it might make more sense to use something that exists.
* Can we use the standard elevation shadow on the app bar, even when the category bar is "fixed"
* Can we use Noto Serif (bold) for headings?
* What font-size is the compact app bar using (when it is "fixed" on scroll)? It feels slightly large, but it might just be my eyes playing tricks, or because I'm used to seeing titles left-aligned on the 72dp keyline. (note: I believe 20sp is the standard size)

Thanks for all of the great work on this thus far! 👏

malinajirka:
@antonis We might want to consider fixing the design-review issues in a separate PR(s). I'm a bit worried it might get out of hands.

This PR is too big from the beginning - we usually create multiple PRs and we merge them into a working branch. When all the PRs are reviewed by developer(s) and ready to go - we ask for a design review. It's way easier to review for developers - each PR has a single focus. And when the whole feature is working the designer can jump in. Another benefit is that we don't have developer and design comments within a single PR as it's a bit difficult to get oriented in for all parties.

antonis:
> @antonis We might want to consider fixing the design-review issues in a separate PR(s). I'm a bit worried it might get out of hands.

I agree @malinajirka . I try to cover the technical fixes on this and get it merged. I can follow up on the design fixes separately and fix them in another PR.

iamthomasbishop:
@antonis @malinajirka Totally fine to break this work up as necessary. I was under the impression that this was awaiting design review, so I wanted to share my notes and feedback based on the current state -- to be clear, I don't expect all of the above things I mentioned to be addressed immediately, or even as part of this specific PR. Please feel free to ping me when ready for another round of design review.

WordPress/src/main/java/org/wordpress/android/ui/mlp/LayoutsAdapter.kt
antonis: I agree with you @malinajirka . Fixed with 5bf4fb52f048808ba7a73fe794037f055d1c5e90

WordPress/src/main/java/org/wordpress/android/ui/mlp/LayoutsAdapter.kt
antonis: Fixed with 8815f40d45fd18ae3d7165c2242b5ddcbd06240f

WordPress/src/main/java/org/wordpress/android/ui/mlp/LayoutsAdapter.kt
antonis: I agree. Fixed with 05962fa0f4e6eff4c9a13e5372348929e172f06f

WordPress/src/main/java/org/wordpress/android/ui/mlp/ModalLayoutPickerListDiffCallback.kt
antonis: You are right. This check is not needed. Fixed with baee77e672b474f115007ee6ae593172d4065660

WordPress/src/main/java/org/wordpress/android/ui/mlp/ModalLayoutPickerListDiffCallback.kt
antonis: As above fixed with baee77e672b474f115007ee6ae593172d4065660

WordPress/src/main/java/org/wordpress/android/ui/mlp/ModalLayoutPickerFragment.kt
antonis: That is a great improvement suggestion. Fixed with dee9adae5844b2988b7ce9ee0fdfccd509138d01

malinajirka (approved):
Great job! Thanks for all the changes ;)! I've re-reviewed the code and re-tested the changes and all seems to work as expected. The design issues will be fixed in upcoming PRs - I'm merging this PR.

Issue #12561, opened 2020-07-30
Site Pages and Posts Search Clear Content is misbehaving on Configuration Change


ParaskP7:
### Expected behavior
When searching for `Site Pages` or `Blog Posts` using a specific search query (say `tags`) everything works as expected. All pages or posts that match that search query are shown, as expected. Then, when clicking the `Clear Button (X)`, the search query and results gets cleared. However, if the screen gets rotated first (a configuration change) and then the `Clear Button (X)` gets clicked then, the search query gets cleared but not the search results. Afterwards, rotating back to the original orientation, the state of the screen gets rurther messed up.

### Actual behavior
Clicking on the `Clear Button (X)` during a configuration change (like a rotation) should clear the search query and the displayed results. Also, the overall screen state should not become inconsistent with what the user expects when rotating the screen.

### Steps to reproduce the behavior
1. Open `My Site`
1. Choose a `site`
1. Navigate to `Site Pages` or `Blog Posts`
1. Search for something that will produce a set of results, for example search for `tags`
1. Click the `Close Button (X)` on the top right
1. Observe that the search query and results gets cleared
1. Search again for `tags`
1. This time rotate the screen before clicking on the `Close Button (X)
1. Observer that the search query and results are visible (as expected)
1. Now click the `Close Button (X)` once more
1. Observer that the search query gets cleared but not the search results
1. Rotate the screen back to the initial orientation
1. Observer that the search query reappears and clicking the `Close Button (X)` produces the same set of problems

##### Tested on [device], Android [version], WPAndroid [version]
Tested on Pixel 2 XL, Android 10, WPAndroid 15.4-rc-1

ParaskP7:
The above might also relate to:
- Set Parent Search should clear contents on multiple configuration changes #11316 and
- Pages Search should clear contents on multiple configuration changes #11317

designsimply:
Thank you for the report @ParaskP7! 

I tested using WPAndroid 15.4-rc-1 on Pixel 3 Android 10 and verified that if I search the Posts List then rotate the device and tap "X" that it clears the form field but the results below it are still visible when they should not be. ([55s](https://cloudup.com/cLTjlNbYDoF))

designsimply:
Tangentially related: https://github.com/wordpress-mobile/WordPress-Android/issues/11913.

PR #12565, opened 2020-07-30
Enable Search for Self-Hosted Sites

### LINKED ISSUE #12058: Add the ability to search the post list for self-hosted site
### Expected behavior

I expected that any kind of site I added to the app would give me the option to search the post or page list.

### Actual behavior

Only WordPress.com/Jetpack sites have the option for this feature. 

1. Go to My Site and select a WordPress.com site.
2. Go to Blog Post List
3. You'll see the search icon in the top menu.
4. Go back to My Site and select a self-hosted site. 
5. Go to Blog Post List
6. The icons doesn't display and there is no way to search through posts.

##### Tested on Samsun Tab A, Android 9, WPAndroid 14.9-rc-1

It does look like this was implemented purposefully in #10023, but it was requested by a user with a self-hosted site in 3002439-zen and kind of caught me off guard since I didn't realize this behaviour would be different. I wasn't able to find another report of this anywhere that wasn't related to the old PR.

ParaskP7:
Fixes #12058 
Prerequisite [WordPress-FluxC-Android #1646](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1646)

Description
-----
This feature enables search for self-hosted sites, something that was previously only available for WordPress.com/Jetpack sites. See issue above for more info.

In general the solution is split into two parts:
1. Make the `search icon` available for when the user navigates to a self-hosted site, just as it is available for non self-hosted sites.
2. Add the user's `search query` to the XML RPC descriptor that is responsible to later make an API request for the self-hosted site. Note that with the `develop` version of this repo the `search query` argument is not available since a new version of **Android FluxC** is required (see `To test` section below).

So, the above work is closely related to the [[Feature] Add Search Filter for XML RPC Requests #1646](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1646) PR, which needs to be merge first.

Blog Posts for Self-Hosted Site (Before)
-----

Blog Posts for Self-Hosted Site (After)
-----

Test:
-----
1. Build the `WordPress` app with a `fluxcVersion` of `e32d20633e68ca1fba1726b75463aa7219ca88f0` (update that within the project's parent `build.gradle` file)
1. Sign in to the app and add a self-hosted site
1. Navigate to `Blog Posts`
1. Verify that the `Search Icon` is available on the top right
1. Click on the `Search Icon` and verify the search functionality works as expected and in general works as it used to work for non self-hosted site
1. The above means that search is enabled for self-hosted sites

While testing note that search in general, for both for self-hosted and non self-hosted sites has a configuration change related bug, see #12561. However, it is out of the scope of this task to fix this bug, but it will be considered as part of another task, later on in the future.

PR submission checklist:
-----

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 15.5
 -----
- 
+* [***] Enables search on post list for self-hosted sites.
+
 15.4
 -----
 * [***] The login and signup flows were unified and have a new design.
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostSettingsFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostSettingsFragment.java
@@ -744,17 +744,28 @@ public class EditPostSettingsFragment extends Fragment {
 
     private void updatePassword(String password) {
         EditPostRepository editPostRepository = getEditPostRepository();
-        if (editPostRepository != null) {
-            editPostRepository.updateAsync(postModel -> {
-                postModel.setPassword(password);
-                return true;
-            }, (postModel, result) -> {
-                if (result == UpdatePostResult.Updated.INSTANCE) {
-                    mPasswordTextView.setText(password);
-                }
-                return null;
-            });
-        }
+        if (editPostRepository == null) return;
+
+        String trimmedPassword = password.trim();
+        Boolean isNewPasswordBlank = trimmedPassword.isEmpty();
+        String previousPassword = editPostRepository.getPassword();
+        Boolean isPreviousPasswordBlank = previousPassword.trim().isEmpty();
+
+        // Nothing to save
+        if (isNewPasswordBlank && isPreviousPasswordBlank) return;
+
+        // Save untrimmed password if not blank, else save empty string
+        String newPassword = isNewPasswordBlank ? trimmedPassword : password;
+
+        editPostRepository.updateAsync(postModel -> {
+            postModel.setPassword(newPassword);
+            return true;
+        }, (postModel, result) -> {
+            if (result == UpdatePostResult.Updated.INSTANCE) {
+                mPasswordTextView.setText(newPassword);
+            }
+            return null;
+        });
     }
 
     private void updateCategories(List<Long> categoryList) {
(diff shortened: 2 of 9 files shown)


ParaskP7:
I am not sure about the two items in the `PR Submission Checklist` so please guide me to be able to ✅ them as well:
- I have considered adding accessibility improvements for my changes.
- I have considered if this change warrants user-facing release notes and have added them to RELEASE-NOTES.txt if necessary.

Also, another ✅ that I would have liked to check are adding UI tests where possible, much I am not sure what is the process for adding those.

ParaskP7:
`lint` and `test` are failing on for **CircleCI** due to compilation errors. This is expected, since this PR is based on a future version of `FluxC`. As such, only when a commit with the newest version of `FluxC` will be added to this PR (or the `develop` gets it), will **CircleCI** pass these checks.

PS: Locally, both `lint` and `test` are running fine.

ParaskP7:
> Thanks for the changes @ParaskP7! The apps works as expected, great job!
> 
> I have a couple of questions/suggestions.
> 
> 1. The step 3 is mentioning "Site Pages". I'm wondering if we need to test site pages, since we modified just Post List's VMs. Wdyt?
> 2. I agree that a SiteModel can theoretically have "UNKNOWN" origin - at least the "enum" supports it. However, afaik we don't support that type anywhere else in the app. The app needs to communicate with either WPComRest or XML-RPC. If it doesn't know the type it can't communicate with the site at all. That's the main reason why we never found a need to introduce "isAccessedViaXmlRpc" -> `!isAccessedViaWPComRest == isAccessedViaXmlRpc` (btw there is a typo in the code "View" vs "Via"). I'm wondering if we really want to introduce it now. Wdyt?
Many thanks for reviewing my PR @malinajirka and verifying its functionality! 🌟

1. You are right, this change is only about `Blog Posts`, I wonder why I added a step to verify `Site Pages` as well. I am going to update the description and remove it. It is totally misleading. I am guessing since I was constantly testing both to verify they work the same, I posted the same testing instructions. 😶
2. I wan't aware about the `UNKNOWN` indeed, thanks for clarifying that for me. Also, thanks for noting the `View` typo as well, will update it with a commit. So based on what you say, and if I understand correctly, then overall the `isSearchAvailable` LiveData value is not needed anymore and can be removed from the `PostListMainViewModel`. Then, every site with have search available by default. Is that right?

malinajirka:
> You are right, this change is only about Blog Posts, I wonder why I added a step to verify Site Pages as well. I am going to update the description and remove it. It is totally misleading. I am guessing since I was constantly testing both to verify they work the same, I posted the same testing instructions. 😶

No worries ;)!

> I wan't aware about the UNKNOWN indeed, thanks for clarifying that for me. Also, thanks for noting the View typo as well, will update it with a commit. So based on what you say, and if I understand correctly, then overall the isSearchAvailable LiveData value is not needed anymore and can be removed from the PostListMainViewModel. Then, every site with have search available by default. Is that right?

Yep, sounds good to me ;).

ParaskP7:
@malinajirka all your suggestions have been now added in this PR. Please let me know if everything looks as expected!

PR #12573, opened 2020-07-31
[Modal Layout Picker] Create Layout Picker category bar UI using static data

antonis:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2437

**Related PRs:**
- **Depends on**: https://github.com/wordpress-mobile/WordPress-Android/pull/12564

## Description

This PR implements the category bar in the header of the [Starter Page Template Picker](https://github.com/wordpress-mobile/gutenberg-mobile/issues?q=is%3Aissue+is%3Aopen+%5BModal+Layout+Picker%5D) for new pages.

## To test:

**Note**: For demo/testing purposes, the categories repeat the same layout 30 times because the static data only has one layout per category. 

### Select a Category
1. Select a category
    - **Expect** to see the sections condensed to be only the selected category filter
1. Deselect the category
    - **Expect** to see the other categories return

### Select Multiple-Categories 
1. Select a category
    - **Expect** to see the sections condensed to be only the selected category filter
1. Select a different category
    - **Expect** to see the new section added to the list
1. Deselect a category
    - **Expect** the list to update again to be only the selected categories
1. Deselect the rest of the categories
    - **Expect** the all of the categories to return

### Selecting a Layout then filter on the same category
1. Select a layout in a category such as Blog
1. Select the category in the filter bar that matches the category of the selected layout
    - **Expect** to see the section updated to only show the selected category, and for the selected layout to still be selected.
1. Deselect the category in the filter
    - **Expect** to see the other categories return, and for the selected layout to still be selected.

### Selecting a Layout then filter on the same category deselect layout
1. Select a layout in a category such as Blog
1. Select the category in the filter bar that matches the category of the selected layout
    - **Expect** to see the section updated to only show the selected category, and for the selected layout to still be selected.
1. Deselect the layout
    - **Expect** the bottom bar to revert to "Create a Blank Page"
1. Deselect the category in the filter
    - **Expect** to see the other categories return, and for no layout to be selected

### Selecting a Layout then filter on a different category
1. Select a layout in a category such as Blog
1. Select a category in the filter bar that does not match the category of the selected layout
    - **Expect** to see the section updated to only show the selected category, and for the bottom bar to remain as "Preview" and "Create Page"
1. Deselect the category in the filter
    - **Expect** to see the other categories return, and for the selected layout to still be selected.

### Selecting a Layout then filter on a different category select a different layout
1. Select a layout in a category such as Blog
1. Select a category in the filter bar that does not match the category of the selected layout
    - **Expect** to see the section updated to only show the selected category, and for the bottom bar to remain as "Preview" and "Create Page"
1. Select a different layout
1. Deselect the category in the filter
    - **Expect** to see the other categories return, and for the selected layout from step 3 to still be selected.

## Screenshots:
| No category | Selected category |
|---|---|
|device-2020-08-03-134049|device-2020-08-03-134102|

| No category dark | Selected category dark |
|---|---|
|device-2020-08-03-134736|device-2020-08-03-134716|

| Selection | Selection dark |
|---|---|
|||

PR submission checklist:


antonis:
Hello @iamthomasbishop 👋 ,
I have drafted the layout categories selection using static data. There are still some rough edges that I need to take care of (e.g. landscape mode and some UI bugs) but at this point this should behave very close to the final MLP.
You can check the screenshots attached to the PR or download the [APK here](https://64358-9306568-gh.circle-artifacts.com/0/Artifacts/WordPress-pr-12573-build-64358.apk).

I have a question regarding the expected behavior when the user selected layout is filtered out (*Selection dark* gif above). Should we just hide it or also clear the user selection?

Thank you in advance for your feedback 🙇 

cc @chipsnyder

chipsnyder:
> I have a question regarding the expected behavior when the user selected layout is filtered out (Selection dark gif above). Should we just hide it or also clear the user selection?

IMO if the selection is filtered out then we should treat it as being deselected.

iamthomasbishop:
> I have a question regarding the expected behavior when the user selected layout is filtered out (Selection dark gif above). Should we just hide it or also clear the user selection?

I have a counter to chipsnyder's opinion 😃 I would actually keep the selection, because:

* If I've tapped a layout and move about, then brought back into view via the filters, its state is stored. In most cases, I want to keep it selected as I explore.
* If I want to make a new selection, all I have to do is tap a different one.

Some of the early designs tried a couple of treatments/call-outs for the "selected" one, but in the end decided to keep it super simple and understood the trade-off would be exactly this situation (when filters toggle the row a selected item is in).

WordPress/src/main/java/org/wordpress/android/ui/mlp/CategoryViewHolder.kt
```diff
@@ -0,0 +1,57 @@
...
+        category.text = uiState.title
+        emoji.text = uiState.emoji
+        container.contentDescription = if (uiState.selected) parent.context.getString(
+                R.string.mlp_layout_selected,
+                uiState.title
+        ) else uiState.title
+        container.setOnClickListener {
+            uiState.onItemTapped.invoke()
+        }
+        container.backgroundTintList
```
ashiagr: This line doesn't seem to be doing anything, right? Can we remove it?

WordPress/src/main/java/org/wordpress/android/ui/mlp/CategoryViewHolder.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    private fun setSelectedStateUI(selected: Boolean) {
+        check.setVisible(selected)
+        emoji.setVisible(!selected)
+        container.backgroundTintList = parent.context.getColorStateListFromAttribute(
+                if (selected) attr.categoriesButtonBackgroundSelected else attr.categoriesButtonBackground
+        )
+        category.setTextColor(
+                parent.context.getColorFromAttribute(
+                        if (selected) attr.categoriesButtonTextSelected else attr.categoriesButtonText
```
ashiagr: Can we pass selected state values from the `ViewModal` itself instead of adding this logic in the `ViewHolder`?

Also  instead of creating our own layouts to filter content, we can use pre-built [material filter-chips](https://material.io/components/chips#filter-chips). These can be added in a [ChipGroup](https://developer.android.com/reference/com/google/android/material/chip/ChipGroup) and put into a single line using `singleLine`. This is just a suggestion, feel free to ignore it 🙂.

WordPress/src/main/res/values/styles.xml
```diff
@@ -1202,4 +1211,27 @@
...
+    <style name="ModalLayoutPickerCategoryTitle">
+        <item name="android:layout_width">wrap_content</item>
+        <item name="android:layout_height">wrap_content</item>
+        <item name="android:layout_gravity">center</item>
+        <item name="android:ellipsize">end</item>
+        <item name="android:maxLines">1</item>
+        <item name="android:textSize">@dimen/text_sz_medium</item>
+        <item name="android:layout_marginEnd">@dimen/mlp_category_horizontal_margin</item>
+        <item name="android:textAppearance">?attr/textAppearanceBody2</item>
+    </style>
```
ashiagr: I think we can extract some of the common properties to a parent style, lmkwyt :).

WordPress/src/main/res/drawable/modal_layout_picker_category_background.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<shape
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
+
+    <solid
+        android:color="#ededed">
```
ashiagr: We should avoid using hex colors directly in the xml, wdyt?

WordPress/src/main/res/values-night/styles.xml
```diff
@@ -68,6 +68,15 @@
...
         <item name="actionMenuTextColor">@color/primary_disabled_selector</item>
+
+        <!-- Modal Layout Picker -->
+        <item name="categoriesButtonText">@color/white</item>
+        <item name="categoriesButtonBackground">@color/mlp_categories_button_background_dark</item>
+        <item name="categoriesButtonTextSelected">@color/black</item>
+        <item name="categoriesButtonBackgroundSelected">@color/mlp_categories_button_background_selected_dark</item>
+        <item name="categoriesButtonCheckIconColor">@color/black</item>
+        <item name="mlpDividerColor">@color/mlp_divider_dark</item>
+        <item name="categoriesBackground">@color/mlp_categories_background_dark</item>
```
ashiagr: Instead of using separate colors in `values-night/styles.xml` and `values/styles.xml ` we can use color attributes from the Material Design theme.

e.g.
For `categoriesButtonText` which is
`<item name="categoriesButtonText">@color/white</item>` in dark mode 
and
` <item name="categoriesButtonText">@color/black</item>` in light mode

we can use `colorOnSurface` which gives the same output and we need not add additional style attributes specially for Modal Layout Picker. Lmkwyt.

I myself get confused when to use new style or use these existing color attributes 🙂 but in the end find that we can eliminate most of such styles.

WordPress/src/main/java/org/wordpress/android/ui/mlp/ModalLayoutPickerFragment.kt
```diff
@@ -75,19 +80,30 @@ class ModalLayoutPickerFragment : BottomSheetDialogFragment() {
...
     }
 
+    /**
+     * Sets the header and title visibility
+     * @param visible if true the title is shown and the header is hidden
+     */
+    private fun setTitleVisibility(visible: Boolean) {
+        if (visible == (title.visibility == View.VISIBLE)) return // No change
+        title.visibility = if (visible) View.VISIBLE else View.INVISIBLE
+        header.visibility = if (!visible) View.VISIBLE else View.INVISIBLE
```
ashiagr: Can we utilise `UiHelpers.updateVisibility` here?
Also `DisplayUtils.isLandscape(context)` on line 87 seems like a small business logic. Can we use `DisplayUtilsWrappers` and get title ui state from the VM?

WordPress/src/main/java/org/wordpress/android/viewmodel/mlp/ModalLayoutPickerViewModel.kt
```diff
@@ -46,62 +48,74 @@ class ModalLayoutPickerViewModel @Inject constructor(
...
-    private val _listItems = MutableLiveData<List<ModalLayoutPickerListItem>>()
-    val listItems: LiveData<List<ModalLayoutPickerListItem>> = _listItems
+    private val _layoutCategories = MutableLiveData<List<LayoutCategoryUiState>>()
+    val layoutCategories: LiveData<List<LayoutCategoryUiState>> = _layoutCategories
+
+    /**
+     * Tracks the categories
+     */
+    private val _categories = MutableLiveData<List<CategoryListItemUiState>>()
+    val categories: LiveData<List<CategoryListItemUiState>> = _categories
```
ashiagr: Wondering if we can use a single UiState object that represents the state of the screen 🤔.  

Internal ref: paqN3M-1W

[Example](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt#L196)

WordPress/src/main/res/layout/modal_layout_picker_fragment.xml
```diff
@@ -1,38 +1,104 @@
...
-        android:layout_height="wrap_content" />
+        android:layout_height="wrap_content"
+        android:background="?attr/categoriesBackground"
+        android:fitsSystemWindows="false">
 
-    <include
-        layout="@layout/modal_layout_picker_header"
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content" />
+        <com.google.android.material.appbar.CollapsingToolbarLayout
```
ashiagr, line 17: Thank you for using `CollapsingToolbarLayout` 💙.

ashiagr (commented):
I tested described flows and everything came across very well. Nice job @antonis !
Just left few code-related comments/ suggestions, lmkwyt. 

Also I haven't reviewed test cases yet, will review after this first pass.

WordPress/src/main/java/org/wordpress/android/ui/mlp/CategoryViewHolder.kt
antonis: You are right :|
Removed with b3dd6f962cc4f41173fd276d1c55decb76befe62

WordPress/src/main/java/org/wordpress/android/ui/mlp/CategoryViewHolder.kt
antonis: That's a good suggestion. I've moved the state values to UiState with 52cbeff78a157f014c00b607b00139e8e8950642.

Using the [material filter-chips](https://material.io/components/chips#filter-chips) looks like a good improvement. I suggest to leave this to be revisited along with [design feedback](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2563). Wdyt?

WordPress/src/main/res/values/styles.xml
antonis: Fixed with ad3622c01fc7d40d9b9027bcafbef44f0e44f583

WordPress/src/main/res/drawable/modal_layout_picker_category_background.xml
antonis: You are right. This shouldn't be here. Fixed with 60c6a5c9b4eaa8936b0dd8a90c8c88c656af8899

WordPress/src/main/java/org/wordpress/android/ui/mlp/ModalLayoutPickerFragment.kt
antonis: I agree. I've added a new method in `UIHelpers` to switch between `View.VISIBLE/INVISIBLE` with ce8416047d7801abdb39c3b7725b73c48632ca7b. I hope that the naming of the new method is not confusing. I reversed the logic to be somewhat more profound. Setting to `View.GONE` causes a flickering combined with the scrolling layout.
I'll update on your second point.

WordPress/src/main/res/layout/modal_layout_picker_fragment.xml
antonis: Thank you [for suggesting this](https://github.com/wordpress-mobile/WordPress-Android/pull/12482#issuecomment-669743793) :) I wish I had persisted more on using the `CollapsingToolbar` in my initial implementation

WordPress/src/main/java/org/wordpress/android/ui/mlp/ModalLayoutPickerFragment.kt
antonis: >Also DisplayUtils.isLandscape(context) on line 87 seems like a small business logic. Can we use DisplayUtilsWrappers and get title ui state from the VM?

Fixed with 28127ff24f45191828d7e41036043aff778674a7

WordPress/src/main/res/values-night/styles.xml
antonis: Fixed with 90dc9c398364d3e5b9b1c83c77a2e3c0fdb2a895

WordPress/src/main/java/org/wordpress/android/ui/mlp/CategoryViewHolder.kt
ashiagr: > I suggest to leave this to be revisited along with design feedback. Wdyt?

Sounds good 👍

WordPress/src/main/res/values-night/styles.xml
ashiagr: Thank you! Wondering if we can eliminate remaining ones as well 🤔. These use different shades of grey:

```
categoriesBackground
categoriesButtonBackground
categoriesButtonBackgroundSelected
mlpDividerColor
```

Is it possible that we combine emphasis or alpha values with color attributes like `colorSurface` to get that shade? We can address it in design review. Feel free to ignore it here.

WordPress/src/main/java/org/wordpress/android/viewmodel/mlp/ModalLayoutPickerViewModel.kt
antonis: I agree with you Ashita. That will make things a lot clearer. I tackled this with 5897942a501cf0793ce0395afc1c073fe3371a31
Note that I plan to enhance and use the loading/error states in the next PR that will implement the service calls.

antonis:
> Can we add a method `onAppBarOffsetChanged` in the VM, pass `verticalOffset`, `scrollThreshold` as params and call `setHeaderTitleVisibility` from inside that method based on `verticalOffset < scrollThreshold `condition?

Yes we can. Thank you for suggesting this :)
I tackled this with 429444d3775bafac43b303093ff469f0df7a1a1f

WordPress/src/main/java/org/wordpress/android/ui/mlp/ModalLayoutPickerFragment.kt
ashiagr: Since `DisplayUtils.isLandscape` used a `context` we decided to keep it in the VM and pass the value to VM using `VM.start()` in 429444d

WordPress/src/main/java/org/wordpress/android/viewmodel/mlp/ModalLayoutPickerViewModel.kt
ashiagr: Fixed in 5897942

ashiagr (approved):
Changes look good to me. Also I've reviewed the tests. 
Thank you for all the hard work here @antonis ! 

I haven't merged this PR yet because of the "Needs Design Review" label. Feel free to merge if it is not a blocker.

PR #12579, opened 2020-08-01
Issue/topics you may like

jd-alexander:
Parent Issue: #12028

## Solution

This PR adds the "topics you may like" section to the `Discover` fragment of the Reader Improvements Phase II. 

## Testing
1. Set the `isReaderImprovementsPhase2Enabled` to true. 
2. Go to the Reader tab. 
3. Click the Discover tab. 
4. You will see the topics you may like at the top of the list. 
5. You can click on one of them to get more details in the `ReaderPostListFragment`. 

## Unit Tests
These will be attached as another PR.

Screenshot | Demo 
--------|-------
 <kbd></kbd>   |       <kbd></kbd>

## Reviewing 

## Submitter Checklist


WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
```diff
@@ -121,6 +133,11 @@ class ReaderDiscoverViewModel @Inject constructor(
         }
     }
 
+    private fun onReaderTagClicked(tag: String) {
+        val readerTag = readerUtilsWrapper.getTagFromTagName(tag, INTERESTS)
```
jd-alexander, line 137: I am just putting this comment here to type out the use of the `INTERESTS` here to ensure that I selected the right type.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostUiStateBuilder.kt
```diff
@@ -98,6 +104,14 @@ class ReaderPostUiStateBuilder @Inject constructor(
         )
     }
 
+    fun mapTagListToReaderInterestUiState(interests: ReaderTagList, onClicked: ((String) -> Unit)) =
+            ReaderInterestCardUiState(interests.take(READER_INTERESTS_LIST_SIZE).map { interest ->
+                ReaderInterestUiState(
+                        interest.tagTitle,
+                        interests.indexOf(interest) != READER_INTERESTS_LAST_INDEX, onClicked
+                )
+            })
```
jd-alexander, line 113: So for this approach here, the idea is that we need to have 5 options for the topics the user may like so I simply took the first five topics and I utilize the current index to determine if the `uiState` should have a divider since we don't want the last item of the interest card list to show a divider. Let me know if this approach is fine or if it can be simplified or improved in anyway. I have the last index hardcoded and I was wondering if I should create a variable to store the result of `interest.take(5)` and then utilize that to determine the index. So yes, any feedback or improvement here is welcome. 

Also let me know if this is the right place for this type of business logic. I was wondering if this should be from the data layer. I put it here because this is somewhat view logic, so let me know that as well.

WordPress/src/test/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCaseTest.kt
```diff
@@ -56,6 +60,11 @@ class GetDiscoverCardsUseCaseTest {
                 .thenReturn(ReaderConstants.JSON_CARD_POST)
         whenever(mockedInterestsCardJson.getString(ReaderConstants.JSON_CARD_TYPE))
                 .thenReturn(ReaderConstants.JSON_CARD_INTERESTS_YOU_MAY_LIKE)
+
```
jd-alexander, line 63: This was put here so that the tests would pass in this PR. Since this use case won't need the `ReaderTagRepository` when the API endpoint for the interests you may like is working, I will remove this logic at the time of its implementation.

WordPress/src/main/res/layout/reader_interest_card.xml
```diff
@@ -0,0 +1,40 @@
...
+
+    <androidx.constraintlayout.widget.ConstraintLayout
+        android:id="@+id/root_layout"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content">
+
+        <androidx.recyclerview.widget.RecyclerView
+            android:id="@+id/interests_list"
+            android:layout_marginTop="@dimen/margin_extra_large"
+            android:layout_width="match_parent"
```
ashiagr, line 17: Np: We should try to use `0dp` instead of `match_parent` for widgets contained in a `ConstraintLayout`. From [official doc](https://developer.android.com/reference/android/support/constraint/ConstraintLayout.html#RelativePositioning)

> Important: MATCH_PARENT is not recommended for widgets contained in a ConstraintLayout.  Similar behavior can be defined by using MATCH_CONSTRAINT with the corresponding left/right or top/bottom constraints being set to "parent".

WordPress/src/main/res/layout/reader_interest_card.xml
```diff
@@ -0,0 +1,40 @@
...
+
+        <org.wordpress.android.widgets.WPTextView
+            android:id="@+id/interests_you_may_like_header"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:fontFamily="serif"
+            android:layout_margin="@dimen/margin_extra_large"
+            android:text="@string/reader_discover_interests_header"
+            android:textAppearance="?attr/textAppearanceHeadline6"
+            android:textColor="?attr/colorOnSurface"
```
ashiagr, line 32: These style attributes can be set using reader styles xml. Same can be done for styling `reader_interest_item`.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostUiStateBuilder.kt
ashiagr: Few observations:

1.  If there are less than 5 elements - let's say total of 3, `interests.indexOf(interest) != READER_INTERESTS_LAST_INDEX`  will be true for the last element at index 2 and divider will be displayed for the last element. wdyt?
2. `onClicked` can be moved to a new line (each param on separate line).
3.  I don't think it is necessary to put visibility logic inside the VM. We can create a `buildDividerVisibility` fn similar to the `buildPhotoFrameVisibility` inside `ReaderPostUiState` for consistency.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderCardUiState.kt
```diff
@@ -62,6 +62,14 @@ sealed class ReaderCardUiState {
             val onDiscoverClicked: ((Long, Long) -> Unit)
         )
     }
+
+    data class ReaderInterestCardUiState(val interest: List<ReaderInterestUiState>) : ReaderCardUiState() {
+        data class ReaderInterestUiState(
```
ashiagr, line 67: Np: `interest` can be pluralised here.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
ashiagr: `INTERESTS` tag type is used just to fetch the interests (shown on the interest picker) from the server. They're later saved as `FOLLOWED` tags into the `tbl_tags`. `getTagFromTagName` first looks for a tag in the table, it  won't find any tag with the `INTERESTS` tag type:

https://github.com/wordpress-mobile/WordPress-Android/blob/a3988b1b990862b2e7ee16df6737c449a43349f1/WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderUtils.java#L216-L217

I think right tag type will be 'FOLLOWED' here. 

Having said that, I'll recheck if I can eliminate `INTERESTS` tag type.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderInterestCardViewHolder.kt
```diff
@@ -0,0 +1,27 @@
...
+    parentView: ViewGroup
+) : ReaderViewHolder(parentView, R.layout.reader_interest_card) {
+    override fun onBind(uiState: ReaderCardUiState) {
+        uiState as ReaderInterestCardUiState
+
+        if (interests_list.adapter == null) {
+            interests_list.layoutManager = LinearLayoutManager(interests_list.context, RecyclerView.VERTICAL, false)
+            val readerInterestAdapter = ReaderInterestAdapter(uiHelpers)
+            interests_list.adapter = readerInterestAdapter
+        }
```
ashiagr, line 24: Can this initialisation logic can be moved to VH's `init` block?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
malinajirka: I believe we'll be using "DEFAULT" type for these. We don't expect anything special from them - we just need to open tag detail, right?

WordPress/src/main/res/layout/reader_interest_card.xml
jd-alexander: Oh yes, thanks for this!

WordPress/src/main/res/layout/reader_interest_card.xml
jd-alexander: oh, nice! I never knew that existed. Will address this 🙏

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
ashiagr: Looks like we cannot get a relative endpoint url for DEFAULT type

https://github.com/wordpress-mobile/WordPress-Android/blob/9d3d5af87f60e7008350e308be2eaed863c9a82b/WordPress/src/main/java/org/wordpress/android/ui/reader/services/post/ReaderPostLogic.java#L283-L306

WordPress/src/main/res/layout/reader_interest_card.xml
jd-alexander: Resolved here 0d668b4

WordPress/src/main/res/layout/reader_interest_card.xml
jd-alexander: Resolved here 0d668b4

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderCardUiState.kt
jd-alexander: Thanks :) 290fd3c

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderInterestCardViewHolder.kt
jd-alexander: Yes, of course 😄 67bb984

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
jd-alexander: Hi friends :) Using `DEFAULT` causes the functionality not to work. As seen below
<kbd></kbd>

interests_not_visible

jd-alexander:
This is how the current implementation with real data looks:
 Before | After 
--------|-------
 <kbd></kbd>        |  <kbd></kbd>

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostUiStateBuilder.kt
jd-alexander: Great observations :) 6409d7f

jd-alexander:
Awesome! Thanks @ashiagr 💯 I kept it as a draft as I will be pushing a set of tests for this PR. I will link here once done!

jd-alexander:
I know this was already approved @ashiagr but I want another review because I modified code :)

zwarm:
> I know this was already approved @ashiagr but I want another review because I modified code :)

I'm here to review, but the last commit points to a param change? What am I missing? Can @jd-alexander let me know what changed?

jd-alexander:
@zwarm hey Annmarie! I had merged the changes from `develop` and then I started seeing this error. I am going to investigate it a bit more tomorrow morning to see what's causing it because I thought it was the param position but on further investigation, I now realize that isn't the case 🙏

jd-alexander:
Hey @zwarm it's okay to merge this PR. I spoke with Ashita and the increase in the param was due to the `isDiscover` parameter that was added to `mapPostToUiState()`

jd-alexander:
I am merging `develop` into this PR to resolve the test failure being caused due to a JDK update https://github.com/wordpress-mobile/WordPress-Android/pull/12703

jd-alexander:
This is ready for another review & merge @zwarm @ashiagr Thanks for all the help so far 😄

ashiagr (commented):
Hey Joel! 
Regarding `ReaderTagType` in `ReaderDiscoverViewModel.onReaderTagClicked`, noticed that with the INTERESTS type - tag selected from this section cannot be followed/ unfollowed on the Topic details screen because it tries to look for FOLLOWED tag type in the `tbl_tags`. I believe we should use FOLLOWED tag type when we pass this tag to the Topic details screen. wdyt?

jd-alexander:
Interesting. Thanks @ashiagr I am going to look into this!

ashiagr (approved):
All looks good to me 👍. Thank you so much for this PR 🙌 !!!

PR #12580, opened 2020-08-03
Add view model to photo picker

planarvoid:
This PR prepares the Photo picker for the media picker consolidation. I've transformed the fragment to use a ViewModel and a UI state object with the list of media items. This moves the screen towards a unidirectional data flow and makes it less error prone are more readable. There should be no functionality changes.

To test:
- Go to a Gutenberg post
- Add an image block
- Click on "Add image"
- Select "Choose from device"
- The photo picker should behave the same way as before
- The screen keeps position on orientation
- The animations still work correctly
- Adding image to a block works correctly

To test 2:
- Repeat steps from the previous test with "Gallery" block

To test 3:
- Repeat steps with a "Video" block

PR submission checklist:


planarvoid:
> I see a difference with previous behaviour though: when in single selection mode (like adding a video/image) if you select a thumbnail you are no more able to unselect that thumbnail tapping again into it.

> It seems a reasonable behaviour tbh since you can unselect using the back arrow, but since it's different from before I wanted to check with you if it was made on purpose or not 🙇‍♂️.

You're absolutely right! I saw this functionality but it didn't occur me it could be unexpected and different from develop. I've fixed that in https://github.com/wordpress-mobile/WordPress-Android/pull/12580/commits/50fb5122a461217e630a7b61906049c361762be8

planarvoid:
hey @develric , thanks for the check! This is a great catch 👍 . I've been returning the selected URIs in the image order, not in the selection order. This is fixed now. I'll add tests for this functionality in the next PR.

PR #12584, opened 2020-08-03
Remove animateLayoutChanges from discover fragment

malinajirka:
Parent issue #12028

Fixes the following exception which sometime occurs when you switch tabs in reader:
```
java.lang.IllegalStateException: Page(s) contain a ViewGroup with a LayoutTransition (or animateLayoutChanges="true"), which interferes with the scrolling animation. Make sure to call getLayoutTransition().setAnimateParentHierarchy(false) on all ViewGroups with a LayoutTr...
```
It was caused by "animateLayoutChanges" flag on the discover fragment container. This flag isn't necessary for this screen as the RecyclerView will take care of layout animation changes on the items. We could potentially animate from progress shown state to progress hidden state, but we don't do that anywhere else anyway.

To test:
Smoke test discover screen with RI FF flag on

PR submission checklist:

### DIFF
--- a/WordPress/src/main/res/layout/reader_discover_fragment_layout.xml
+++ b/WordPress/src/main/res/layout/reader_discover_fragment_layout.xml
@@ -3,8 +3,7 @@
     xmlns:app="http://schemas.android.com/apk/res-auto"
     android:id="@+id/constraint_layout"
     android:layout_width="match_parent"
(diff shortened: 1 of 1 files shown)


zwarm (approved):
LGTM 👍 
Tested on an emulator and physical device. Did not encounter any Illegalstateexceptions

PR #12593, opened 2020-08-04
Issue/12319 refactor cards event handling

malinajirka:
Parent issue #12319 

Note: "Not ready for merge" - https://github.com/wordpress-mobile/WordPress-Android/pull/12563 needs to be merged first

I did some changes to the discover repo. Let me know what you think about them. I don't have a strong opinion about any of them I just did what felt right at the moment.

- Removed/Inlined `ReaderUpdatePostsEndedHandler.kt`  - I first tried to refactor it so we could listen for both PostUpdated events and DiscoverCardsUpdated events. However, the result felt over-engineered. I tried to put the `Subscribe` methods to the data provider and it felt simpler. Since we decided to keep the data provider single focused (provides up-to-date data) I feel like keeping the subscribe methods there is fine. It's a class which won't continue growing. Wdyt?
- Unsubscribed the ReaderDiscoverDataProvider from the `PostUpdated` events. Unless I'm missing something they are no longer necessary since we listen for `FetchDiscoverCardsEnded` and `ReaderPostTableActionEnded` wdyt?
- Removed the factory method from the discover data provider. It will always use "DISCOVER_POST_CARDS" tag so we don't need to provide it within the constructor -> we can let the dagger initialize the data provider.
- Fixed `shouldAutoUpdateTagUseCase` for the "DISCOVER_POST_CARDS" tag (https://github.com/wordpress-mobile/WordPress-Android/commit/931bb57ffad878ad2b950834907b47c521172c08)

To test:
1. Enable RI FF 
2. Open Reader
3. Select interests and click on "Done"
4. Notice the data load correctly on the discover tab

PR submission checklist:

- [] I have considered adding unit tests where possible.


zwarm:
>  Since we decided to keep the data provider single focused (provides up-to-date data) I feel like keeping the subscribe methods there is fine. It's a class which won't continue growing. Wdyt?
> * Unsubscribed the ReaderDiscoverDataProvider from the `PostUpdated` events. Unless I'm missing something they are no longer necessary since we listen for `FetchDiscoverCardsEnded` and `ReaderPostTableActionEnded` wdyt?
> * Removed the factory method from the discover data provider. It will always use "DISCOVER_POST_CARDS" tag so we don't need to provide it within the constructor -> we can let the dagger initialize the data provider.
> * Fixed `shouldAutoUpdateTagUseCase` for the "DISCOVER_POST_CARDS" tag ([931bb57](https://github.com/wordpress-mobile/WordPress-Android/commit/931bb57ffad878ad2b950834907b47c521172c08))

zwarm:
> Since we decided to keep the data provider single focused (provides up-to-date data) I feel like keeping the subscribe methods there is fine. It's a class which won't continue growing. Wdyt?
Agree, now that we have settled on the providers having a single responsibility, it more streamlined to have the subscribe methods directly in the class itself.

zwarm:
> Removed the factory method from the discover data provider. It will always use "DISCOVER_POST_CARDS" tag so we don't need to provide it within the constructor -> we can let the dagger initialize the data provider.

Yes, SRP again. 👍

PR #12595, opened 2020-08-04
Enable @mentions feature in vanilla flavors for version 15.4

maxme:
The `GUTENBERG_MENTIONS` feature flag is always true, but we'll keep the flag in case we want to revert it quickly.

Note: the feature was enabled and tested in alpha releases. This doesn't require a new beta release.


maxme:
@mchowning I update the release notes for 15.5.

Issue #12318, opened 2020-06-30
Reader Improvements: Topic Detail Main Issue


malinajirka:
Parent issue - https://github.com/wordpress-mobile/WordPress-Android/issues/12026.

This is a Main Issue for the "Topic detail" screen. Goal of this task is to update the design of the screen's header and add "Follow" button so the users can easily start following the topic.

**BackLog:**

**InProgress:**

**Done:**
- [x] Add a header with topic title to the screen #12355
- [x] Add follow/unfollow button to the header and implement the functionality #12355, #12434, #12435
- [x] Add tracking - tbd
- [x] Polish the UI #12478

Open Questions:
- Will we show this screen in fullscreen or just as a content of the "discover" tab? -> Most likely as fullscreen

PR #12600, opened 2020-08-05
RI: Enable Follow button for logged out user(self-hosted site)

ashiagr:
Parent Issue: #12318

This PR 
- Enables "Follow" button for logged out user(self-hosted site).
- Displays "Interests Picker" on switch from "Followed" to "Discover" tab if no followed tags found for user.

#### To test

Prerequisites - Reader Improvements feature flag is enabled

 **01. Tag/ topic is followed/ unfollowed**

1. Login using a self hosted site
2. Go to Reader tab
3. Go to Topic details screen 
(via Followed tab -> Filter bar -> Select tag (from bottom sheet) -> Select post -> Click a tag on post)
4. Notice that Follow button is shown
5. Follow/ Unfollow topic
6. Notice that selected topic gets followed/ unfollowed properly and corresponding tag is shown/ hidden respectively under 
Followed tab -> Filter bar -> Bottom sheet Tags section

 **02. Interest picker shown on switch to discover tab if no followed tags found for user**
1. Login using a self hosted site or wp.com account with followed tags
2. Go to Reader => Followed tab
3. Unfollow all followed tags
[**Logged in user**
Followed tab -> Click Filter bar -> Gear icon
**Logged out user**
Followed tab -> Click Filter bar -> Select tag (from bottom sheet) -> Select post for tag -> Click selected tag on post -> Unfollow]
4. Switch from Followed tab to Discover tab
5. Notice that Interests Picker is shown

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
@@ -16,6 +16,7 @@ import org.wordpress.android.modules.BG_THREAD
 import org.wordpress.android.modules.UI_THREAD
 import org.wordpress.android.ui.prefs.AppPrefsWrapper
 import org.wordpress.android.ui.reader.ReaderEvents
+import org.wordpress.android.ui.reader.repository.usecases.tags.GetFollowedTagsUseCase
 import org.wordpress.android.ui.reader.tracker.ReaderTracker
 import org.wordpress.android.ui.reader.tracker.ReaderTrackerType.MAIN_READER
 import org.wordpress.android.ui.reader.usecases.LoadReaderTabsUseCase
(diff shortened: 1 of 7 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
```diff
@@ -116,9 +118,24 @@ class ReaderViewModel @Inject constructor(
...
+        if (selectedTag?.isDiscover == true &&
+                appPrefsWrapper.isReaderImprovementsPhase2Enabled()) {
+            launch {
+                val userTags = getFollowedTagsUseCase.get()
+                if (userTags.isEmpty()) {
+                    isReaderInterestsShown = true
+                    initialized = false
+                    _uiState.value = InitialUiState
+                    _showReaderInterests.value = Event(Unit)
+                }
```
malinajirka, line 132: This comment is not related to this PR: Flags like "initialized" (with the current usage) and "isReaderInterestsShown" might not be maintainable in the long run. Imagine we keep adding such flags - the class would become complex spaghetti code (especially since some combinations of these flags might not be valid -> eg "initialized = false, isReaderInterestsShown = true").
I'm wondering if we could simply turn these flags into UiStates. Technically this screen always goes through these states:  [Display interests screen] -> loadTabs -> Initialize tabs -> show content. 

Any thoughts? If you think it's worth considering, could you please create a note on the project board and move it to the last position in the "Android Backlog" column. Thanks!

ashiagr: Good advice, thanks! Note added [here](https://github.com/orgs/wordpress-mobile/projects/63#card-43190584).

Issue #12599, opened 2020-08-04
List of Users Available for Mentions Does Not Refresh


mchowning:
### Describe the bug
On Android, the list of users available for use in an @-mention is not refreshed after it is initially loaded. In other words, the initial list of users appears to never get updated after it is initially downloaded.

### To Reproduce
In the following steps, the same bug can be seen by either performing an @-mention when typing a comment or adding an @-mention inside the gutenberg editor. For convenience, the steps below only specify adding a mention within the gutenberg editor.

1. Using WPAndroid, switch to a test site and add a mention within the gutenberg editor
2. **Observe that there are N mention candidates**
3. On the web, add a new user as an Editor to the test site
4. Kill WPAndroid, restart it, and add a mention within the gutenberg editor
5. **Observe that there are still N mention candidates (i.e., the newly added user _is not_ listed)**
6. Delete all app data (or uninstall and reinstall the app)
7. Open the app and add a mention within the gutenberg editor
8. **Observe that there are now N+1 mention candidates (i.e., the newly added user _is_ listed)**
9. On the web, remove the user you added to the test site in step 3
10. Kill WPAndroid, restart it, and add a mention with the gutenberg editor
11. **Observe that there are N+1 mention candidates (i.e., the newly removed user _is still_ listed)**
12. Delete all app data (or uninstall and reinstall the app) 
13. Open the app and add a mention within the gutenberg editor
14. **Observe that there are N mention candidates (i.e., the newly removed user is no longer listed)**

### Expected behavior
Inserting a mention should display a list with the **current** list of users for the site. At worst, killing the app and restarting should refresh the list of users. It should _not_ be necessary to wipe the app's data to get an updated list of users.

## Smartphone
 - Device: Pixel 1
 - OS: Android X
 - Version: WPAndroid 0ebce45f8b (HEAD of `develop` on August 4, 2020)

## Additional context
This appears to be caused by [this code](https://github.com/wordpress-mobile/WordPress-Android/blob/8bbb383475060dffe313582a52a8e41a0e5a35dc/WordPress/src/main/java/org/wordpress/android/ui/suggestion/util/SuggestionUtils.java#L40-L44), which pulls the list of mentions from the database and then only makes an API call for the list of users if the database table is empty.

guarani:
👋 @mchowning with https://github.com/wordpress-mobile/WordPress-Android/pull/12607 merged, can this be closed?

mchowning:
[...]

PR #12607, opened 2020-08-05
Make mentions api call when user inserts a mention

mchowning:
This is a fix to address https://github.com/wordpress-mobile/WordPress-Android/issues/12599 by fetching the latest users when the mentions UI is displayed as long as we haven't received the users in the last 60 seconds. This throttle is being used to prevent too many requests from being made since on large sites the response size is not insignificant I tested a site with >200 users and found the response was about a third of a megabyte.

**Note that this fix is targeting the release branch. Please be extra thorough in testing this.** 🙇 

I also made the same changes for the retrieval of tags since that is done in the same place and at the same time as suggestions. I actually think we may not be using the tags at all in the app, so this may present an opportunity for cleanup in the future, but I didn't want to include that here since we're targeting a release branch.

#### Ensure that https://github.com/wordpress-mobile/WordPress-Android/issues/12599 is fixed
1. Switch to a test site and add a mention within the Gutenberg editor
2. Observe that there are N mention candidates
3. On the web, add a new user as an Editor to the test site
4. Kill WPAndroid, restart it, and add a mention within the Gutenberg editor
5. **Observe that there are now (possibly after a short delay so the API call has time to complete) N+1 mention candidates**, i.e., the newly added user _is_ listed. Before this fix, the newly listed user would not be listed until the app's data was cleared.
6. On the web, remove the user you added to the test site in step 3
7. Add a new mention with the Gutenberg editor
8. **Observe that there are now (possibly after a short delay so the API call has time to complete) N mention candidates (i.e., the newly removed user _is not_ listed)**

#### Check the throttle
1. Kill and start the app
2. Open the mentions UI
3. **Observe the suggestions and tags call is made (either via proxy or the apps logs)**
4. Exit the mentions UI and open the mentions UI again in under 60 seconds
5. **Observe the suggestions and tags call is NOT made (a message that the call is being skipped will be included in the logcat logs)**
6. Wait until 60 seconds from the original response has passed
7. Open the mentions UI
8. **Observe the suggestions and tags call is made (either via proxy or the logcat logs)**

#### Make sure the throttle handles failures appropriately
1. Kill and start the app
2. Open the mentions UI and ensure that the suggestions call that is made fails (i.e. rewrite the response to be a 400)
3. Exit the mentions UI and open the mentions UI again in _under_ 60 seconds
4. **Observe the suggestions and tags call is made (either via proxy or the logcat logs)** despite the fact that a call was already attempted within the last 60 seconds.

PR submission checklist:


loremattei:
I moved this to 15.6 after having checked with Matt on chat.

guarani:
Hey @mchowning, I see this is still a draft PR — is this awaiting a review from me?

mchowning:
Thanks for checking @guarani ! No review needed yet, but I'll probably mark it ready for review today or tomorrow.

WordPress/src/main/java/org/wordpress/android/ui/suggestion/service/SuggestionService.java
```diff
@@ -47,27 +50,43 @@ public IBinder onBind(Intent intent) {
...
         };
-        RestRequest.ErrorListener errorListener = new RestRequest.ErrorListener() {
-            @Override
-            public void onErrorResponse(VolleyError volleyError) {
-                AppLog.e(AppLog.T.SUGGESTION, volleyError);
-                removeSiteIdFromSuggestionRequestsAndStopServiceIfNecessary(siteId);
-            }
+        RestRequest.ErrorListener errorListener = volleyError -> {
+            AppLog.e(AppLog.T.SUGGESTION, volleyError);
+            removeSiteIdFromSuggestionRequestsAndStopServiceIfNecessary(siteId);
```
mchowning, line 89: There is no change to the logic in these listeners, this is just switching to using the lambda syntax.

WordPress/src/main/java/org/wordpress/android/ui/suggestion/service/SuggestionService.java
```diff
@@ -47,27 +50,43 @@ public IBinder onBind(Intent intent) {
...
         mCurrentlyRequestingSuggestionsSiteIds.add(siteId);
-        RestRequest.Listener listener = new RestRequest.Listener() {
-            @Override
-            public void onResponse(JSONObject jsonObject) {
-                handleSuggestionsUpdatedResponse(siteId, jsonObject);
-                removeSiteIdFromSuggestionRequestsAndStopServiceIfNecessary(siteId);
-            }
+        RestRequest.Listener listener = jsonObject -> {
+            handleSuggestionsUpdatedResponse(siteId, jsonObject);
+            SUGGESTION_REQUEST_THROTTLER.onResponseReceived(siteId);
```
mchowning, line 84: This line is the only change to the logic in the suggestion listeners. The rest of the diff is just the result of switching to using the lambda syntax.

WordPress/src/main/java/org/wordpress/android/ui/suggestion/service/SuggestionService.java
```diff
@@ -103,24 +122,16 @@ private void removeSiteIdFromSuggestionRequestsAndStopServiceIfNecessary(long si
...
         mCurrentlyRequestingTagsSiteIds.add(siteId);
-        RestRequest.Listener listener = new RestRequest.Listener() {
-            @Override
-            public void onResponse(JSONObject jsonObject) {
-                handleTagsUpdatedResponse(siteId, jsonObject);
-                removeSiteIdFromTagRequestsAndStopServiceIfNecessary(siteId);
-            }
+        RestRequest.Listener listener = jsonObject -> {
+            handleTagsUpdatedResponse(siteId, jsonObject);
+            TAG_REQUEST_THROTTLER.onResponseReceived(siteId);
```
mchowning, line 129: This line is the only change to the logic in the tags listeners. The rest of the diff is just the result of switching to using the lambda syntax.

mchowning:
This is ready for review @guarani and @cameronvoell !  🙇

hypest:
Removed myself from the reviewers as I will most probably not have the time today to jump in but, @cameronvoell let me know if after reviewing you feel you'd need me to have a look for any reason. Thanks!

mchowning:
> I moved this to 15.6 after having checked with Matt on chat.

I wasn't clear that although this PR shouldn't hold up the release branch it would was still going to target the 15.5 release. I've gone ahead and moved this back to the 15.5 milestone. Sorry for the confusion @loremattei ! 🙇

mchowning:
@guarani noted an issue where mentions did not always load the first time in comments. We confirmed that this is not an issue introduced in this PR (I reproduced it all the way back in WPAndroid 14.0), and have filed an issue for it: https://github.com/wordpress-mobile/WordPress-Android/issues/12682.

PR #12609, opened 2020-08-06
Fix load more functionality which expected only Post items in the list

malinajirka:
Parent issue #12028

Follow up for https://github.com/wordpress-mobile/WordPress-Android/pull/12603

I realized an issue with the previous implementation. We will be displaying different types of cards in the discover tab - post cards, interests you might like, blogs you might like etc. The load more detection logic was expecting only post cards - so if the `size - 2` card wasn't a post card, the load more request wouldn't have been emitted. This PR fixes this issue.

To test:
1. Clear app data
2. Enable RI FF
3. Open Discover tab
4. Scroll down and notice more and more items are being loaded

PR submission checklist:


zwarm (approved):
> The load more detection logic was expecting only post cards
Nice switch out using UIState instead of post/blog id. More posts are loaded as expected. 
👍

PR #12620, opened 2020-08-07
Media upload error codes wpcom jp in media lib

develric:
This PR is part of a series related to #12325 

The changes here should address the Media Library with WP.com/JP scenario. The equivalent scenarios for GB Editor will be covered in subsequent PRs.

This has a companion FluxC PR [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1657) 

Here below the error messages you should get from endpoint API for `rest_upload_limited_space`, `rest_upload_file_too_big` and `rest_upload_user_quota_exceeded`

```
rest_upload_limited_space|Not enough space to upload. 3,145 KB needed. Back

rest_upload_file_too_big|This file is too big. Files must be less than 1500000 KB in size. Back

rest_upload_user_quota_exceeded|You have used your space quota. Please delete files before uploading. Back
```

Note that we changed the FluxC to get the user message part of the string (that after the `|` sign) to be localized from the API (adding the `locale` query param that was not provided before).

One thing we should consider is maybe to filter out the ` Back` ending part, it does not seem to be a message to the user and not sure how it is used.

### To test:
The following apply both to WP.com and JP sites.

- Open the Media Library screen and try to upload a file
- Force trigger the 3 errors as described above (I can provide access to my sandbox or you can use Charles proxy rewrite feature or AS debugger for this)
- Check you get the API user message in the snackbar
- Check you get the message localized if you change app language

UPDATE: while further checking this I see the the message is cut in the snackbar since too long, maybe something we can discuss and address in another PR.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #12325, opened 2020-06-30
Handle new error codes from the media upload endpoint


maxme:
### Expected behavior

In case of an error during media upload, we should show the exact error to help the user take a decision.

The list of errors are defined [here in Calypso](https://github.com/Automattic/wp-calypso/blob/ec15ddaba3c45905cf71f3ca09624694e8632a8d/client/state/media/reducer.js#L72-L88). Most interesting are:

- rest_upload_limited_space
- rest_upload_file_too_big
- rest_upload_user_quota_exceeded

Note that these error codes are actually stored in the message field. Here is an sample response we get from the https://public-api.wordpress.com/rest/v1.1/sites/{SITE_ID}/media/new/ API endpoint in case of an upload on a quota limited site:

```
{
	"error": "upload_error",
	"message": "rest_upload_user_quota_exceeded|You have used your space quota. Please delete files before uploading. Back"
}
```

- On the network side (FluxC): It would be a good idea to split the error message (something linke `.split("|")`), store the error code in the FluxC event that is sent back to the app.
- On the UI, we should handle that error and display the message correctly instead of showing a generic message. This must be done:
  - In the media library.
  - In the editor.

### Actual behavior

We show a generic error in the editor and in the media library:

Media Library|Editor
-|-
|

More details on paaHJt-17e-p2

designsimply:
When an image upload fails for an image larger than 2650 x 2650 pixels on a Bluehost Jetpack site, there is a very generic error message both in the Media and the Editor—would it be possible to improve this error message in the app? 

> Media upload failed.
> CURL error 28: Operation timed out after 30001 mil…

<img src="https://user-images.githubusercontent.com/1119271/113953940-27b68280-97d6-11eb-8bd5-01dbd3995965.png" width="270" alt="Screenshot_20210407-191738"> <img src="https://user-images.githubusercontent.com/1119271/113953944-284f1900-97d6-11eb-9d82-c6c28b29ea7c.png" width="270" alt="Screenshot_20210407-191617">
<sup>Tested with WPAndroid 17.1-rc-1 on Pixel 3 Android 11.</sup>

The following error is shown on the Bluehost WP Admin Media screen if I try uploading the same image there, and it's a much better error message:

> Post-processing of the image failed likely because the server is busy or does not have enough resources. Uploading a smaller image may help. Suggested maximum size is 2500 pixels.

[...]

PR #12625, opened 2020-08-07
Media upload error codes self-hosted in media lib

develric:
This PR is part of a series related to #12325

The changes here should address the Media Library with self-hosted scenario. The equivalent scenarios for GB Editor will be covered in subsequent PRs.

This has a companion FluxC PR [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1658)

The Fluxc modifications map the 401 and 500 upload media errors that can be found [here](https://github.com/WordPress/WordPress/blob/649ab58d96491549f8b617d2702bb437488bb16d/wp-includes/class-wp-xmlrpc-server.php#L6298)

NOTE: I was trying to use a similar approach as we did in #12620 ; that is leveraging the user error message that the API is actually able to provide and display them to the user. Unfortunately AFAIU in the case of XML-RPC API we do not send or use a `locale`parameter that the client app can send to the server to declare it's idiom. The only thing I was able to get is a localized message changing the language server side, but that is not a per user triggered localization so we cannot use this approach. I reverted to using general string resources that only map the category of the error (401 and 500).

### To test
Access to a pure self-hosted site in the app.

- Open the Media Library screen and try to upload a file
- Force trigger the errors as per the link above (I can provide access to a test site I have setup for that or you can use Charles proxy rewrite feature or AS debugger for this)
- Check you get the relevant string resources message in the snackbar

NOTE: if you are using the Charles proxy approach, the message received from the server should have a 200 http status and the following xml content:

#### 401 case
```
<?xml version="1.0" encoding="UTF-8"?>
<methodResponse>
  <fault>
    <value>
      <struct>
        <member>
          <name>faultCode</name>
          <value><int>401</int></value>
        </member>
        <member>
          <name>faultString</name>
          <value><string>Sorry, you are not allowed to upload files.</string></value>
        </member>
      </struct>
    </value>
  </fault>
</methodResponse>
```

#### 500 case
```
<?xml version="1.0" encoding="UTF-8"?>
<methodResponse>
  <fault>
    <value>
      <struct>
        <member>
          <name>faultCode</name>
          <value><int>500</int></value>
        </member>
        <member>
          <name>faultString</name>
          <value><string>Could not write file.</string></value>
        </member>
      </struct>
    </value>
  </fault>
</methodResponse>
```

PR submission checklist:


WordPress/src/main/res/values/strings.xml
```diff
@@ -1594,6 +1594,8 @@
     <string name="error_media_canceled">Uploading media were canceled</string>
     <string name="error_media_quota_exceeded">Storage quota exceeded</string>
     <string name="error_media_quota_exceeded_toast">Cannot upload file.\nStorage quota was exceeded.</string>
+    <string name="error_media_xmlrpc_not_allowed">Operation not allowed</string>
```
planarvoid, line 1600: Could we maybe make this a bit more explicit? Maybe something like "Operation not allowed by XMLRPC"?

planarvoid (commented):
I've tested both the cases with Charles rewrite and they show the correct error 👍 . I have one suggestion on the FluxC PR. Let me know what you think.

develric:
Moving the milestone to 15.7 since makes sense to take care of [this](https://github.com/wordpress-mobile/WordPress-Android/pull/12672#issuecomment-674990529) design improvements before to ship these chained PRs.

develric:
Hi @planarvoid , just noting that I'm addressing the copy change to `error_media_xmlrpc_not_allowed` string in #12672 since all discussion about copy change happened in that PR.

planarvoid (approved):
Looks good 👍  thanks!

PR #12655, opened 2020-08-11
Update reader site header UI

zwarm:
This PR updates the Reader Site Header UI according to the design concepts provided.  It addresses this [task](https://github.com/orgs/wordpress-mobile/projects/63#card-42500990)

There were a couple of gotchas that were uncovered while addressing this PR.
The `ReaderSiteHeaderView` has support for viewing by feed or blog. In both cases, ReaderBlog data is gathered from the ReaderBlogTable, but feed/blog requests use different endpoints to fetch the data from the server. The icon image url is not available in the `/read/feed/{feedId}` endpoint, so I had to add an extra server call to `/read/sites/{blogId}` to get the url when the request is for a feed. 

In addition, when working within the context of a feed, the blogId is set to the feedId value. This means that I had to first query the post table for the "blogId" for the feed. This was necessary to make the correct call to the server for the image. 

I have added comments in the code to highlight why changes were made. Please let me know if they are not clear. 

Light | Dark
--------|------- 
<kbd></kbd> | <kbd></kbd>

The new following functionality for non logged in users is behind a FF for RI improvements, so the logged out functionality would look like the following if this PR was released without a FF.
<kbd></kbd>

To test
1. Go to reader
2. Tap on a site name in a post card
3. See the new site header

PR submission checklist:


WordPress/src/main/res/layout/reader_site_header_view.xml
```diff
@@ -1,113 +1,94 @@
...
+            style="@style/ReaderTextView.Site.Header.Body"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:ellipsize="end"
+            android:gravity="start"
+            android:maxLines="3"
+            android:paddingEnd="@dimen/margin_extra_large"
+            android:paddingStart="@dimen/margin_extra_large"
+            android:paddingTop="@dimen/margin_medium"
+            android:paddingBottom="@dimen/margin_extra_large"
```
ashiagr, line 87: Can these be moved to `ReaderTextView.Site.Header.Body` style?

zwarm: > Can these be moved to `ReaderTextView.Site.Header.Body` style?

Sure can, see commit 6338b46

mbshakti:
Looks fantastic, thanks for working!

PR #5963, opened 2017-05-29
7.5 - Fix SecurityException in Me Fragment

daniloercoli:
`Me` Fragment: Fix a `SecurityException` by downloading the "external" picture immediately, if it's not available in the media store, before starting manipulating it with UCrop.

Fixes #5818 (I hope this is the final fix for this issue).

To test:
- Open the `Me` Fragment
- Tap on the picture
- Select Gallery
- Get a picture from an external store (Google Drive)
- Make sure the picture is correctly accessed by UCrop

PR #12661, opened 2020-08-12
Make local copy of media when using 3rd party URIs

mzorz:
THIS IS WIP.

Fixes https://github.com/Automattic/stories-android/issues/493, https://github.com/Automattic/stories-android/issues/420

Related to the issue described and fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/5963 for other parts of the app.

### Description of the issue

When a 3rd party app media URI is fetched, it lasts for about as much as the context where the content provider was queried exists.
In the Stories flow, when tapping on the FAB to create a new Story and then tap on  "add a story", we first show the media picker (that is, in the context of the main screen). From the app's photo picker you can go to the native picker, and from there you can open the list of apps that can expose filtered media (such as Google Photos app). Once you make your pick, a set of temporal URIs are provided, and these are valid for as long as the context of the original request lasts.
Hence, we could see strange behavior such as picking several pictures and only some of them added, or getting to add them all as slides (pick 5 images and get 5 slides added), and show as thumbnails OK given these were populated as soon as the Story composer is created, but then would show black boxes when switching story slides given those provided URIs were no longer valid.

If, for example, you'd start a Story and then would stay there (keeping the Activity as a context), then opening the media picker from the Story composer and picking Google Photos pictures from there would in turn work alright, given these URIs were still valid for the context.

### Solution
Then long term solution is to fetch and keep a copy of the media, as we do elsewhere.

The problem of our use case is that we _start our flow from the main screen_  and then pass the URis to another context.
We need to save those URI's image data and then pass the local copy URIs to the composer instead, all of this also taking into account it could be a lengthy operation so, we'll need to show a progress dialog and make sure the context doesn't go away until we're done.

This PR introduces a `FetchAndOptimizeLocalMediaIfNeededUseCase` and made `StoriesMediaPickerResultHandler` a CoroutineScope so URIs picked in the mediapicker from 3rd party apps are first copied locally and then fed into the Stories composer.

### To test:
You should have Google Photos installed and logged in, with a few pictures taken on another device (so we make sure these pics are not local to the device).

1. open the WP app
2. tap on the FAB, select Add story post
3. when the media picker shows up, tap on the left bottom button and then choose "Choose from device"
4. the native media picker appears. Tap on the hamburger icon on the top-left of the screen. At the bottom of the list, a few apps should appear. Look for Google Photos.
5. pick a few images
6. observe these get added to your new Story, and a dialog appears for a while if processing needs be done first.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
@@ -120,7 +120,7 @@ import org.wordpress.android.ui.quickstart.QuickStartMySitePrompts
 import org.wordpress.android.ui.quickstart.QuickStartMySitePrompts.Companion.getPromptDetailsForTask
 import org.wordpress.android.ui.quickstart.QuickStartMySitePrompts.Companion.isTargetingBottomNavBar
 import org.wordpress.android.ui.quickstart.QuickStartNoticeDetails
-import org.wordpress.android.ui.stories.StoriesMediaPickerResultHandler.Companion.handleMediaPickerResultForStories
+import org.wordpress.android.ui.stories.StoriesMediaPickerResultHandler
 import org.wordpress.android.ui.stories.StoriesTrackerHelper
 import org.wordpress.android.ui.stories.StoryComposerActivity
 import org.wordpress.android.ui.themes.ThemeBrowserActivity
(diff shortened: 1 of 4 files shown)


mzorz:
Closing in favor of https://github.com/wordpress-mobile/WordPress-Android/pull/12670

PR #12665, opened 2020-08-13
Copy photo picker to media picker

### LINKED ISSUE #12639: Create a MediaPickerActivity and Fragment behind a flag
Copy & paste the PhotoPickerActivity and the PhotoPickerFragment into new MediaPickerActivity and MediaPickerFragment (and view model). These classes should be available behind a flag that's turned off for now. This allows us to develop the new picker and run both versions and compare the functionality.

planarvoid:
Fixes #12639 
We'll be developing the new media picker side by side with the current photo picker. That's why we've decided to copy the current files and deprecate the old ones. This PR does two things:
- copies all the classes related to the Photo picker to the `mediapicker` subpackage and renames them to Media picker
- marks all the Photo picker classes as deprecated with a message to also change the `MediaPicker` version of the file if someone updates the `PhotoPicker`

These new files are not used anywhere yet, when we start developing a feature, we should go step by step and replace the photo picker activity by a conditionally started media picker activity (based on the flag value).

The later we merge this PR, the higher chance is that some functionality from the photo picker will be missed in the media picker. On the other hand this is not a big deal because we'll be testing the media picker a lot anyway before it's merged. 

To test:
- Nothing to test here, maybe try to launch the app but the new files are not used yet

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/photopicker/mediapicker/DeviceListBuilder.kt
```diff
@@ -0,0 +1,81 @@
+package org.wordpress.android.ui.photopicker.mediapicker
```
develric, line 1: Wonder if we should place the mediapicker package at the same level as the photopicker (that is directly inside ui) so (if I got intention right) we can remove it completely once we switch (or do we need to keep the photopicker package? I see we have the `PermissionsHandler.kt` but maybe makes sense to move it as well? 🤷‍♂️).

WordPress/src/main/java/org/wordpress/android/ui/photopicker/mediapicker/MediaPickerFragment.kt
```diff
@@ -0,0 +1,456 @@
...
+import android.widget.PopupMenu
+import androidx.appcompat.app.AppCompatActivity
+import androidx.fragment.app.Fragment
+import androidx.lifecycle.Observer
+import androidx.lifecycle.ViewModelProvider
+import androidx.lifecycle.ViewModelProviders
+import androidx.recyclerview.widget.GridLayoutManager
+import kotlinx.android.synthetic.main.photo_picker_fragment.*
+import kotlinx.android.synthetic.main.photo_picker_fragment.view.*
+import kotlinx.android.synthetic.main.stats_list_fragment.*
```
develric: NP: AS says this import is unused. Wonder why the linter didn't say anything, maybe because the class is not still used 🤔

develric (commented):
Hi @planarvoid , when we first talked about this approach we were a bit afraid of keeping track of eventual modifications to original files but I like the solution you used with deprecation 👍!
Also thanks for using very modular commits, it was much easier for me to follow the changes progressive logic 🙇‍♂️!

I left a couple of minor comments/np for you to evaluate but otherwise LGTM 👍 (also smoke tested the Media Library and around)! Let me know 🙇‍♂️!

WordPress/src/main/java/org/wordpress/android/ui/photopicker/mediapicker/DeviceListBuilder.kt
planarvoid: hmm, that's a good point. On one hand having both in the same parent package makes it clear they are related to each other. On the other hand you're right that it would make it easier to remove the old code once this work is done. What do you think if I moved the new classes in another PR since there are already some other PRs that depend on this package location? (we can move the `PermissionsHandler` once we remove the photo picker 🤷 )

planarvoid:
thanks for the check @develric ! I've replied on the comments, let me know what you think

WordPress/src/main/java/org/wordpress/android/ui/photopicker/mediapicker/DeviceListBuilder.kt
develric: SGTM 👍

develric (approved):
Hi @planarvoid , thanks for the follow up; I agree we can move the package in subsequent PR. LGTM :shipit:

PR #12400, opened 2020-07-09
Issue/12028 save for later

malinajirka:
Parent issue #12028

This tasks extracts logic related to "save for later/bookmark" action on reader post list items. ReaderPostBookmarkUseCase contains all the related logic. It propagates events to the view layer using LiveData.

Merge Instructions
1. Review this PR
2. Make sure https://github.com/wordpress-mobile/WordPress-Android/pull/12358 is merged
3. Update target branch to develop
4. Remove "Not Ready for Merge" label
5. Merge this PR

To test:

Dialog
1. Clear app data
2. Open Reader - Following tab
3. Click on Bookmark action on one of the items
4. Notice a dialog is shown
5. Click on OK and go to Saved tab
6. Notice the post is saved there

Snackbar
1. Open Reader - Following tab
1. Click on Bookmark action on another item
2. Notice Snackbar is shown
3. Click on "View All" action in the snackbar
4. Notice the "Saved" tab is shown and the post is there

Offline
1. Open Reader - Following tab
1. Click on Bookmark action on yet another item
2. Notice snackbar is shown
3. Wait for cca 10s
4. Turn on airplane mode
5. Go to Saved tab
6. Open the post and notice the content (even images) are loaded correctly even though you are offline

Unbookmark
1. Click on unbookmark action on one of the saved items
2. Notice the post gets unbookmarked and is removed from the saved tab (it might still be there but in a collapsed mode with "Undo" action)

The new Discover tab can be potentially tested as well but it's a bit complicated. We'll test all the actions before it's released. I personally wouldn't worry about it now.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #12667, opened 2020-08-13
RI: Discover - fix follow action

ashiagr:
Parent Issue: #12028

### Description

~~Follow Site (from more menu) logic in `ReaderPostListFragment` is extracted to `ReaderPostFollowUseCase` and similar to `ReaderPostBookmarkUseCase` in [Issue/12028 save for later #12400 ](https://github.com/wordpress-mobile/WordPress-Android/pull/12400), different `LiveData` instances are used to communicate data like snackbar message or refreshing post on failure in the legacy post adapter. `ReaderPostFollowUseCase` is used on discover tab VM as well as `ReaderPostListFragment`'s VM.~~

### Suggestions
1. We might want to use single `Livedata` (of generic type) in the usecase, observing it in the VM and updating respective `LiveData` instances in the VM based on result type.

2. In the future, we might also want to consider using [Flow](https://codelabs.developers.google.com/codelabs/advanced-kotlin-coroutines/#13) in the usecases when it is released as stable version.

    > Using Flow, it's natural to collect data in the ViewModel, Repository, or other data layers when needed.

    > Since Flow is not tied to the UI, you don't need a UI observer to collect a flow. This is a big difference from LiveData which always requires a UI-observer to run. It is not a good idea to try to observe a LiveData in your ViewModel because it doesn't have an appropriate observation lifecycle.

UPDATE
- As per last discussion, used `Flow` in the useCase - b331349
- Flow values are collected and appropriate livedata are updated in the handler - e74bb0a
- Also utilised notifications useCase for enabling notification on follow. - ab3956d
- Utilised isExternalFeed method to display 
    - notifications menu in the more popup - 80c4fd6
    - enable notifications snackbar - 8a48dd7

### To test

**Following -> Follow**
1. Open Reader - Following tab
2. Click "Following" in the more menu on a followed site post
3. Notice  in the more menu: "Following" (in green) changes to "Follow" (in blue) 

**Follow -> Following + Snackbar to enable Notifications**
1. Continue from above
2. Click "Follow" in the more menu
3. Notice Snackbar is shown
4. Click on "Enable" action in the snackbar
5. Notice  in the more menu 
    5a - Follow (in blue) changes to Following (in green)
    5b - "Turn off site notifications" is shown (in green)

Repeat above steps for an external feed like https://timesofindia.indiatimes.com/rssfeedstopstories.cms. Notice that Enable notifications snackbar is not shown.

**Network Failure**
1. Continue from above
2. Enable Airplane mode
3. Click "Following" in the more menu
4. Notice that a snackbar is shown with an error message for no network

**Request Failure**
1. Make changes in the code [here](https://github.com/wordpress-mobile/WordPress-Android/pull/12667/files#diff-9758e44dc0bc3288df32504d25282546R48) to return `RequestFailed` state
2. Reinstall app -> Open Reader -> Following tab
3. Click "Following" in the more menu on a followed site post
4. Notice in the more menu: "Following" (in green) changes to "Follow" (in blue) 
5. Wait for few secs
6. Notice in the more menu: 
    6a - Follow (in blue) changes to Following (in green)
    6b - Snackbar is shown with an error message

Above tests should also work for discover tab (with RI FF enabled).

### Notes
~~1. Targeting feature branch `issue/12028-reader-post-action-usecases` instead of `develop`  till we reach an agreement on final design for usecases.~~
~~2. Above tests might not work as expected if the site to be followed has a wrong `blogId` set in the `tbl_blog_info` table. This is due to an existing bug where `blogId` is wrongly set to `feedId` for a blog followed as a feed. `ReaderBlog` doesn't parse `blog_ID` key [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/models/ReaderBlog.java#L66-L67). This issue is being fixed separately.~~. Fixed in #12694
3. Have not entirely removed [these methods](https://github.com/wordpress-mobile/WordPress-Android/blob/21d5e25380bdbf67f7e65ee6a38950ce44c322e2/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java#L2619-L2656) extracted to `ReaderPostFollowUseCase` as they're also used on Site Details -> `ReaderSiteHeaderView` and `ReaderPostDetailFragment` -> `ReaderPostDetailHeaderView`, `ReaderSimplePostView`.
~~4. Extracted logic needs further refactoring.~~
5. Will add tests once these changes are approved (Added "Not Ready for Merge" label till then). 

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/pages/SnackbarMessageHolder.kt
```diff
@@ -2,9 +2,12 @@ package org.wordpress.android.ui.pages
 
 import androidx.annotation.StringRes
 
+const val INVALID_MESSAGE_RES = -1
```
malinajirka: I'm wondering if we could use UIString instead of using an invalid id, wdyt?

ashiagr: Converting to UIString looks like a great idea! I'll create a separate PR as it will impact several classes.
I'll convert both messageRes and buttonTitleRes to UIString.

ashiagr: Converted here: #12740

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -477,6 +478,8 @@ public void onCreate(Bundle savedInstanceState) {
                 })
         );
 
+        mViewModel.getRefreshPost().observe(getViewLifecycleOwner(), this::setFollowStatusForBlog);
```
malinajirka: Hmm, I'm wondering if this isn't a bit misleading. We are observing "refreshPost" event but we are calling "setFollowStatusForBlog" -> we technically shouldn't know that the VM wanted us to update the follow status. Moreover, if I as a future developer see `refreshPost` livedata in the VM and I use it I expect that the posts will get refreshed. However, it won't be the case. 
Would it work if we invoked `getPostAdapter().refresh()` instead of `setFollowStatusForBlog(..)`? If not, I'd consider renaming this LiveData to "updateFollowStatus" (the livedata in the handler could be called "followStatusUpdated"), wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderSiteFollowUseCase.kt
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        return suspendCoroutine { cont ->
+            continuation = cont
+            readerBlogActionsWrapper.followBlogForPost(post, isAskingToFollow, actionListener)
+        }
+    }
+
+    sealed class FollowSiteState {
+        data class ReaderPostData(
```
malinajirka: We might want to consider renaming this to something like "PostFollowStatusChanged". Since "ReaderPostData" technically isn't a state, yet it extends `FollowSiteState`. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
```diff
@@ -194,6 +235,35 @@ class ReaderPostCardActionsHandler @Inject constructor(
         _navigationEvents.postValue(Event(ShowReaderComments(blogId, postId)))
     }
 
+    private fun prepareEnableNotificationSnackbarAction(blogName: String?, blogId: Long): () -> Unit {
+        return {
```
malinajirka, line 290: Feel free to ignore this: This will be a bit difficult to test. I'd consider moving the creation of `notificationMessage` into a different class. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
```diff
@@ -194,6 +235,35 @@ class ReaderPostCardActionsHandler @Inject constructor(
...
+            _snackbarEvents.postValue(
+                    Event(
+                            SnackbarMessageHolder(
+                                    UiStringText(notificationMessage),
+                                    UiStringRes(R.string.reader_followed_blog_notifications_action),
+                                    buttonAction = {
+                                        analyticsTrackerWrapper
+                                                .track(FOLLOWED_BLOG_NOTIFICATIONS_READER_ENABLED, blogId)
+                                        siteNotificationsUseCase.updateSubscription(blogId, NEW)
+                                        siteNotificationsUseCase.updateBlogInDb(blogId, false)
```
malinajirka: Since we are accessing the db here, I'd consider launching a coroutine on a bg thread here. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostMoreButtonUiStateBuilder.kt
```diff
@@ -39,8 +41,8 @@ class ReaderPostMoreButtonUiStateBuilder @Inject constructor(
                     )
             )
 
-            // When blogId and feedId are not equal, post is not a feed so show notifications option.
-            if (post.blogId != post.feedId) {
+            // When post not from external feed so show notifications option.
+            if (!readerUtilsWrapper.isExternalFeed(post.blogId, post.feedId)) {
```
malinajirka, line 60: Nice catch :)

malinajirka (commented):
Thanks @ashiagr! I've tested the changes and found one super minor issue - we can either ignore it or create a low priority ticket for groundskeeping (develop behaves the same way).

1. Follow a post
2. Enable notificaitons
3. Unfollow the post
4. Follow the post
5. Notice the snackbar with "enable" is shown even though the notifs are already enabled
- the question is whether we shouldn't show the snackbar in this scenario or if we should disable notifications when you unfollow a site.

I have left some in-code comments. None of them is important - let me know what you think.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
ashiagr: Agree, done in b66b46d

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
ashiagr: Extracted to different class here d3795d8.

ashiagr:

Regarding 

> I've tested the changes and found one super minor issue - we can either ignore it or create a low priority ticket for groundskeeping (develop behaves the same way).

Decided to fix it in this PR itself - afe628e

> the question is whether we shouldn't show the snackbar in this scenario or if we should disable notifications when you unfollow a site.

I think it would make more sense to disable site notifications once we unfollow it.

Tests will be added in a separate PR. Added it to backlog here: #12821.

Ready for another round 🙇‍♀️!

malinajirka:
Thanks for the changes @ashiagr! It seems to work as expected (from the user point of view). However, I can see an error in LogCat
1. Click on More -> follow blog
2. Click on More -> enable notifs
3. Click on More -> unfollow blog
4. See error in logcat (it doesn't occur every time, but most of the times)




```
2020-09-01 11:21:31.189 7441-7562/org.wordpress.android.prealpha E/SQLiteDatabase: Error inserting EMAIL_POSTS_FREQUENCY=null SHOULD_EMAIL_POSTS=false FEED_ID=108454787 SUBSCRIPTION_ID=609519261 BLOG_NAME=Site Title SHOULD_NOTIFY_POSTS=false BLOG_ID=182078938 URL=http://gfsdgw343ye4atgaeye5yh.car.blog SHOULD_EMAIL_COMMENTS=false
    android.database.sqlite.SQLiteConstraintException: UNIQUE constraint failed: SubscriptionModel.FEED_ID, SubscriptionModel.URL (code 2067)
        at android.database.sqlite.SQLiteConnection.nativeExecuteForLastInsertedRowId(Native Method)
        at android.database.sqlite.SQLiteConnection.executeForLastInsertedRowId(SQLiteConnection.java:783)
        at android.database.sqlite.SQLiteSession.executeForLastInsertedRowId(SQLiteSession.java:788)
        at android.database.sqlite.SQLiteStatement.executeInsert(SQLiteStatement.java:86)
        at android.database.sqlite.SQLiteDatabase.insertWithOnConflict(SQLiteDatabase.java:1548)
        at android.database.sqlite.SQLiteDatabase.insert(SQLiteDatabase.java:1417)
        at com.yarolegovich.wellsql.InsertQuery.execute(InsertQuery.java:61)
        at org.wordpress.android.fluxc.persistence.AccountSqlUtils.updateSubscriptions(AccountSqlUtils.java:147)
        at org.wordpress.android.fluxc.store.AccountStore.updateSubscriptions(AccountStore.java:1307)
        at org.wordpress.android.fluxc.store.AccountStore.onAccountAction(AccountStore.java:912)
        at org.wordpress.android.fluxc.store.AccountStore.onAction(AccountStore.java:818)
        at java.lang.reflect.Method.invoke(Native Method)
        at org.greenrobot.eventbus.AsyncPoster.run(AsyncPoster.java:46)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
        at java.lang.Thread.run(Thread.java:764)
```

I'm not 100% sure it's related to this PR, but seems like it. Wdyt?

ashiagr:
>  I can see an error in LogCat
```
2020-09-01 11:21:31.189 7441-7562/org.wordpress.android.prealpha E/SQLiteDatabase: Error inserting EMAIL_POSTS_FREQUENCY=null SHOULD_EMAIL_POSTS=false FEED_ID=108454787 SUBSCRIPTION_ID=609519261 BLOG_NAME=Site Title SHOULD_NOTIFY_POSTS=false BLOG_ID=182078938 URL=http://gfsdgw343ye4atgaeye5yh.car.blog SHOULD_EMAIL_COMMENTS=false
    android.database.sqlite.SQLiteConstraintException: UNIQUE constraint failed: SubscriptionModel.FEED_ID, SubscriptionModel.URL (code 2067)
  
```

I tested it occurs on develop as well (with RI FF disabled). Should we create a separate issue and address it there?

malinajirka (approved):
👍 Yes, create an issue and assign it to Groundskeeping triage please. Thanks 🙇

designsimply:
This is awesome. Thank you for working on this!

PR #12671, opened 2020-08-13
Add like use case

zwarm:
Parent Issue #12028 

This PR adds support the like action into `ReaderPostCardActionsHandler`. The `PostLikeUseCase` performs a post like or unlike based on the current state of the post. It returns results as a `ReaderRepositoryCommunication`, so that Network and Failures can be handled properly, in this case it shows the message on the snackbar.  There is no direct update to the discover data in this use case because that data is handled directly by observing the discover feed stream.

Notes: 
- I'm having a bit of trouble mocking the suspendCancellableCoroutine, so I may do that in a separate task. Yes, doing this in a separate task.
- Looking for a second set of eyes/thumbs up on the approach before removing draft/not ready for merge status.

To test:
1. Launch the app
2. Turn on the FF 
3. Relaunch the app
4. Navigate to the discover tab
5. Click "like" on a post
6. The Like icon will change colors

Try the above on both a post that is liked & unliked on the discover tab.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
```diff
@@ -102,8 +110,20 @@ class ReaderPostCardActionsHandler @Inject constructor(
         AppLog.d(AppLog.T.READER, "Block site not implemented")
     }
 
-    private fun handleLikeClicked(postId: Long, blogId: Long) {
-        AppLog.d(AppLog.T.READER, "Like not implemented")
+    private fun handleLikeClicked(post: ReaderPost) {
```
zwarm, line 161: Not sure if this all should be happening at this location.

malinajirka: I think we can either handle it here or return SnackbarMessageHolder instead of ReaderRepositoryCommunication. Imo both approaches are fine.

zwarm: I am partial to leaving it as is, because if we want to notify the UI that it's started we can. Plus it feels kind of odd that the use case knows that we want to set a Snackbar message a.k.a. a UI element. But then I guess you can say the same thing about the action handler, so what do I know 🤷 .

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeUseCase.kt
```diff
@@ -3,39 +3,55 @@ package org.wordpress.android.ui.reader.repository.usecases
...
+        val wpComUserId = accountStore.account.userId
         val request = PostLikeRequest(post.postId, post.blogId, isAskingToLike, wpComUserId)
 
+        if (!networkUtilsWrapper.isNetworkAvailable()) {
+            return NetworkUnavailable
+        }
+
         if (continuations[request] != null) {
-            return PostLikeEnded.PostLikeUnChanged(
+            return Failure(PostLikeUnChanged(
```
malinajirka: AFAIU `continuations[request] != null` means that the request is already in progress right? I feel like it's a bit weird to "fail" in this scenario. Wdyt?

P.S. Btw any reason why we differentiate between Failure and Error?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
malinajirka: >Plus it feels kind of odd that the use case knows that we want to set a Snackbar message a.k.a. a UI element.

yep, I kind of agree.

>But then I guess you can say the same thing about the action handler, so what do I know 🤷 .

The handler directly handles ui interactions so it feels ok to me that it works with UI elements.

malinajirka (commented):
Thanks @zwarm! I've reviewed the code and tested the changes and all looks great. I've left one minor comment, let me know what you think.

P.S. Also, let me know what troubles you had with the unit tests. We are testing suspendable functions in other parts of the app and I don't recall having any issues with them.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeUseCase.kt
malinajirka: Also the structure of `ReaderRepositoryEvent` and `ReaderRepositoryCommunication` is a bit confusing. We can easily create invalid states such as `Failure(PostLikeSuccess)`. Wdyt?

zwarm: @malinajirka 
> AFAIU `continuations[request] != null` means that the request is already in progress right? I feel like it's a bit weird to "fail" in this scenario. Wdyt?

Yeah, semantics ... the request didn't complete successfully as requested because it was already running, it's a duplicate request that results in unchanged. Is this really a success state? You can argue this both ways - success or error. At the end of the day, it's not an error that needs to be handled by the ui. That's why I used failure to indicate that the request didn't succeed, but I will remove Failure and use success instead.

> Also the structure of ReaderRepositoryEvent and ReaderRepositoryCommunication is a bit confusing. We can easily create invalid states such as Failure(PostLikeSuccess). Wdyt?

`ReaderRepositoryEvent` communicate the responses from the eventBus ... this will get removed when I get rid of failure

zwarm: All updated in commit b89165a. Thanks again

malinajirka: >You can argue this both ways - success or error.

Yep, I agree. I decided to go with a new "AlreadyRunning" state in the BlockSiteUseCase. However, it's a minor thing and your solution is imo fine as well. Thanks for the changes.

malinajirka (approved):
Thanks @zwarm! Feel free to merge it when lint and tests are passing on CI.

PR #12672, opened 2020-08-13
Media upload error codes in editor

develric:
This PR is part of a series related to #12325 and should improve the error codes presented to the user in the GB Editor when media upload errors occur.

I finally decided to leverage the `EditorFragmentListener` interface we already use adding the `onFailedToInsertMediaTapped` method.

From UI side I tried to stay near to what we have in iOS where we show this dialog when tapping on a failed to upload media

<p align="center">
  
</p>

And added the error message to the dialog we already have in GB

| Overlay on upload fail | On tap Dialog Before | On tap Dialog After |
|---|---|---|
| image | image | image | 

Maybe we can arrange the message in a different way like we could just use the Title field of the dialog, populating it with the error message if we have one or using the generic `Retry uploading?` string if we do not have a valid one. But I would need your thoughts on the above @osullivanchris, thanks 🙇‍♂️.

### To test
This must be tested both with WP.com/JP and oure self-hosted sites. You should get the same behavior/error messages that you get in #12620 and #12625 but using the editor as per below steps (notice that you should be able to trigger the upload errors as per steps in #12620 and #12625)

- Create a new post in GB 
- add an image/video/gallery (in this last case with multiple images)
- try to upload image(s)/video
- trigger the error so that you get a failed upload media
- tap on the failed to upload media and notice the dialog with the error message (note the error message is the same you should get when uploading the media from the Media Library)
- smoke test the dialog action buttons
- Close the editor, come again to the Draft post and tap on the failed to upload media. Notice you still get the same error message and you are able to use the dialog action buttons.

PR submission checklist:


libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragmentAbstract.java
```diff
@@ -205,6 +205,7 @@ public static MediaType getEditorMimeType(MediaFile mediaFile) {
         void getMention(Consumer<String> onResult);
         void onGutenbergEditorSetStarterPageTemplatesTooltipShown(boolean tooltipShown);
         boolean onGutenbergEditorRequestStarterPageTemplatesTooltipShown();
+        String onFailedToInsertMediaTapped(int mediaId);
```
planarvoid: I think I'd rename this to something like `getErrorMessageFromMedia` which would be more explicit.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -3133,4 +3133,15 @@ private void updateAddingMediaToEditorProgressDialogState(ProgressDialogUiState
         mAddingMediaToEditorProgressDialog = mProgressDialogHelper
                 .updateProgressDialogState(this, mAddingMediaToEditorProgressDialog, uiState, mUiHelpers);
     }
+
+    @Override
+    public String onFailedToInsertMediaTapped(int mediaId) {
+        MediaModel media = mMediaStore.getMediaWithLocalId(mediaId);
+
+        if (media != null) {
+            UploadUtils.getErrorMessageFromMedia(this, media);
```
planarvoid: I think you're missing a "return" here :-)

develric: 🤦‍♂️...I just hope it's the today super hot temperature and not a permanent thing 🙃! I modified it locally but missed commit/push...thanks for catching 🙇‍♂️! changed here fe9518f07c15bcacf34e7d5f68d9113b41c79f66

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragmentAbstract.java
develric: Agree, changed here 9f55e472b2c206adf718755d15498326f58d7c2e

develric:
Hi @planarvoid thanks for the review! I followed up to your comments 🙇‍♂️!

Also noticed the following: the `EditorFragmentListener` interface that we modified is implemented also in the org.wordpress.android.editor.example.EditorExampleActivity

AS is not complaining about the missing method that we added, so I guess that is excluded from build or it's pointing to a different version of the editor lib maybe. Anyway since I'm not familiar with this part of the code base I was wondering if we should take care of merging back this modifications from this PR to some other lib repo (like it happens for the utils lib for example). Do you know about this or have someone we could ping to clarify? (maybe @maxme ?)

planarvoid:
hey @develric , it seems like it's only part of the editor project (as is the interface). I think it will start failing in the editor once you open a PR with your changes there. My guess would be it's in a module that's not part of the WPAndroid project

planarvoid (approved):
Thanks for the changes! It works well now 👍  Feel free to merge this after you sort out the example activity (maybe it's ok to fix it in the PR to the editor?

osullivanchris:
> Maybe we can arrange the message in a different way like we could just use the Title field of the dialog, populating it with the error message if we have one or using the generic Retry uploading? string if we do not have a valid one.

this seems like a fair approach. 

Writing some thoughts...

I'm not sure that we can say with certainty that the title field alone will be the correct format for all messages. An example of a format which could be flexible is 

Heading: Short description of error
Body text: Any further error details, and information about resolution.

Three actions also feels a little confusing to me. I'd probably just expect to see 'Retry', and 'Cancel' (cancel being non-destructive). If I want to remove the image, we already display that trash icon on top of it, so there's another way to achieve 'remove'. If the images were uploaded as a group, I'd expect 'Retry' to retry all of them automatically. 

Here's an example then based on all that
Heading: Couldn't upload file
Body: It looks like you don't have permission to upload files. Make sure you have admin rights.
Actions: Try again  |  Cancel

In general we could probably do with an editorial review here. But it might wait until we have specific messages to work with. Also I hope that 'Operation not allowed' is an example and not one of our 'specific' messages as that sounds quite generic and vague to me!

develric:
Moving the milestone to 15.7 since makes sense to take care of [this](https://github.com/wordpress-mobile/WordPress-Android/pull/12672#issuecomment-674990529) design improvements before to ship these chained PRs.

develric:
Hey @osullivanchris , thanks for the follow up above 🙇‍♂️ ! Overall I agree with what you are suggesting, sharing few notes from my side to give some context from previous PRs:

We have 2 cases to cover:
1. [WP.com/JP](https://github.com/wordpress-mobile/WordPress-Android/pull/12620): we are leveraging the error messages that are sent by the API itself since we can get them already localized based on the client app user locale. Messages are like `You have used your space quota. Please delete files before uploading` and I think we should be fine with these messages since I guess they've been already editorial reviewed server side.
2. [Self-hosted sites](https://github.com/wordpress-mobile/WordPress-Android/pull/12625): in this case AFAIU we cannot get localized messages from the API based on the client app user locale (and we risk to receive messages localized based on the server side locale configuration independently from the app locale). This is something I think we should address server side and I have a note to look a bit more into it during next hack week (at least to try to involve the right people to get their advice on it). Based on the above, we cannot directly use the messages we receive from the self-hosted API and we tried to identify a message for each of the 2 categories of error we can get (some more details [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1658#discussion_r469789511)):
        - XML_RPC_OPERATION_NOT_ALLOWED (401 type): current message shown `Operation not allowed`
        - XML_RPC_UPLOAD_ERROR (500 type): current message shown `Internal server error occurred`

I think we could:

- Ask editorial for the self-hosted messages `Operation not allowed` and `Internal server error occurred`
- Use your proposed dialog arrangement
> Heading: Couldn't upload file
Body:  error message from API (WP.com/JP) OR from editorial (the 2 messages for self-hosted above)

- Use the actions you are proposing

> Actions: Try again | Cancel

Cancel will just dismiss the dialog, try again will only attempt upload for the current tapped media (at least at first check I could not see a reliable/easy way to understand if a media was part of a group, but in this way we lose the RETRY ALL action type; this seems the same behavior we have in GB for iOS tbh, are we ok with this, or do we prefer to keep the RETRY ALL action? maybe with a different naming 🤔  )

Let me know wdyt, thanks 🙇‍♂️

maxme:
> Also noticed the following: the EditorFragmentListener interface that we modified is implemented also in the org.wordpress.android.editor.example.EditorExampleActivity

This example activity is from `libs/editor/example/src/main/java/org/wordpress/android/editor/example/EditorExampleActivity.java` and it's not used anymore. It seems there are a bunch of missing method implementation already.

Ideally we will drop the `editor` library someday and move the `EditorFragment`s to the main app.

> Self-hosted sites: in this case AFAIU we cannot get localized messages from the API based on the client app user locale (and we risk to receive messages localized based on the server side locale configuration independently from the app locale). [...]  Based on the above, we cannot directly use the messages we receive from the self-hosted API and we tried to identify a message for each of the 2 categories of error we can get (some more details here):

This is correct, we should not uses these messages directly in the app. Extreme, but possible example: if a user is hosted on a German server, but have an English audience and English editors, they might get errors in German.

maxme:
> RETRY ALL action type; this seems the same behavior we have in GB for iOS tbh, are we ok with this, or do we prefer to keep the RETRY ALL action? maybe with a different naming 🤔 )

I'd keep the RETRY ALL option. If there is a network hiccup, and I try to upload 20 pics, I don't want to click 20 times on the picture and then on RETRY. I assume this will RETRY ALL that failed.

osullivanchris:
@develric thanks for the explanation, I'm getting a clearer picture :) 

In my words

- The first category of error messages are already meaningful and editorial reviewed, all good here
- The second category of messages are generic, and not meaningful. But it is because of some language and localisation issues that we'll resolve at a later time. However, in the meantime, we can ask editorial to review strings like `Operation not allowed` and `Internal server error occurred` which sound a bit robot-like at the moment!

Then, discussing the format. Glad you agree with the proposed arrangement. But I don't know how many messages this covers and how many different cases. So if we are applying this format broadly I can't say I have thought through every instance!

Re; retry all, I agree with you and Maxime, actually I was proposing the same thing here

> If the images were uploaded as a group, I'd expect 'Retry' to retry all of them automatically.

If its appearing in a blocking dialog like I'm seeing above, I think 'Retry' could mean retry all images. All sometimes being 1! Actually, if I upload 5 images and I see `Retry` and `Retry all` I have no idea what the distinction is, or what `Retry` would do?

develric:
Going ahead asking for copy review. Tying to give some context (let me know in case it's not enough or clear 🙇‍♂️ ).

For self-hosted sites (in the scenario of media file upload failure), we cannot rely on the error messages from self-hosted API since they are not localized based on the app client user locale. So we identified 2 categories of errors for which we need to define a good user message (note we can only discriminate if an error is a 401 or 500 error so we need an umbrella user message to cover each of those 2 categories only):

---

### 401 error code from server is related to the following error cases:

- 'Sorry, you are not allowed to upload files.' 
- 'Sorry, you have used your space allocation of %s. Please delete some files to upload more files.' 
- 'Sorry, you are not allowed to edit this post.' 

**user message we are currently using and need copy review:** `Operation not allowed`

---

### 500 error code from server is related to the following error cases:

- IXR_Error( 500, $upload_err ) -> an upload error of some type
- 'Could not write file %1$s (%2$s).' 

**user message we are currently using and need copy review:** `Internal server error occurred`

Thanks 🙇‍♂️

clucasrowlands:
Thanks for the Editorial ping.

> However, in the meantime, we can ask editorial to review strings like Operation not allowed and Internal server error occurred which sound a bit robot-like at the moment!

What about "Your action is not allowed" or "Action not allowed"? If those alternatives don't work, sticking with "operation" works, though I agree it does sound very robot-like to a general user.

"Internal server error occurred" sounds fine.

develric:
Trying to summarize some conversation me and @osullivanchris had over slack and that were put into actions in this PR:

1. Changed copy for self-hosted upload error from `Operation not allowed` to `Your action is not allowed`. We kept the `Internal server error occurred` error message (thanks @clucasrowlands for the copy review 🙇‍♂️ )
2. For the GB upload error dialog we did the following:
    - Heading: Couldn't upload file
    - Body: error message from API (WP.com/JP) OR from string resources (the 2 messages for self-hosted above)
    - Actions:
        - `Retry`: retries all or one uploads depending on number that failed
        - `Remove upload` : removes the currently tapped failed upload
        - And no Cancel / Dismiss action....but, let the user achieve this by tapping outside the dialog

Here is a couple of examples:

| image | image |
|---|---|

Chris let me know if I missed anything above; I included the sample screenshots above to try to ease your check, but if you want/need to test the APK you can download it from this PR (to create the error conditions is not super straightforward but ping me if you want to give it a try, thanks 🙇‍♂️). 

Hey @planarvoid , if Chris is fine with the current status this one should be ready for a last pass, thanks 🙇‍♂️ !

osullivanchris (approved):
> Chris let me know if I missed anything above; I included the sample screenshots above to try to ease your check, but if you want/need to test the APK you can download it from this PR (to create the error conditions is not super straightforward but ping me if you want to give it a try, thanks 🙇‍♂️).

Hey thanks for the great summary of what was discussed. I think its a good solution and happy to approve based on the screenshot, nice work!

planarvoid (approved):
Works well 👍  good job @develric 
I have two little things that I think could be safely ignored so feel free to merge it anyway.

- Upload an unsupported mime type (`webp`)
- The upload fails
- Click on an image
- The message doesn't start with a capital letter (this probably depends on the server but I think we can capitalize the first letter anyway)

Screenshot 2020-09-01 at 16 56 06

- Click on `Retry` 
- It fails again
- The message is now different 

Screenshot 2020-09-01 at 16 58 28

develric:
Hey @planarvoid thanks for the review 🙇‍♂️ !

For what I see the different messages can happen based on timing and how fast you Retry upload in between failures. In some conditions you can get a GENERIC_ERROR (resulting in the `We couldn't complete this action.` message) for example when an `UPDATE_MEDIA` action is triggered in FluxC and the upload fails from being in the uploading state [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/2054d62e1b5cf4660fa40e3fb05498f3a8c76ac6/fluxc/src/main/java/org/wordpress/android/fluxc/store/UploadStore.java#L355). In my understanding we should not do much for this maybe.

For the message without capitalized first letter: I was also thinking it was from API, but actually there are a whole category of messages in the FluxC `BaseUploadRequestBody.hasRequiredData` that are not capitalized 🙃  ! This said it probably makes sense to capitalize the message at presentation level when displaying the dialog. Tried to stay simple and changed it here 3293de835c60e7a3f94520e8af94fd7a78d8d783

Let me know wdyt 🙇‍♂️

UPDATE: just realized that the strings in FluxC `BaseUploadRequestBody.hasRequiredData` seems not to get localized in any way. Opened a ticket for that here wordpress-mobile/WordPress-FluxC-Android#1678.

planarvoid (approved):
Thanks for the changes, it works well 👍

PR #12673, opened 2020-08-13
Fix translatable attributes

loremattei:
This PR brings back the `translatable=false` attribute to some strings imported from libraries. They add the attribute in the original source code, but it got lost during the import. 

I'll address the bug in the importer script in another PR.

PR submission checklist:


mzorz:
I'll just wait for the formality of getting all green before merging and then will merge 👍

PR #12675, opened 2020-08-14
AppBar refresh - Stats, Pages, Posts, Media Browser and Domain Registration

khaykov:
This PR updates AppBar design and adds lift on scroll functionality to the following screens (and related nested screens):

- Stats
- Pages
- Posts
- Media Browser
- Domain Registration (only scrollable screen is contact details form).

What is lift on scroll? When you scroll content below the AppBar its elevation changes, and as a result shadow appears in light mode, and in Dark Mode color of AppBar gets lighter.

@osullivanchris I would like to get your feedback on abovementioned screens, and mostly on Stats :)

[Image from Gyazo](https://gyazo.com/98af9d5d00364aca9eae816c0b8097ad)

To test:

For all screens:

- Make sure nothing looks out of place (including ActionMode) :)
- Make sure the toolbar state is restored on configuration change.

Test 1:

- Navigate to the Stats screen
- Make sure that lift on scroll works with all the tabs and switching between tabs preserves lift on scroll state.
- Make sure that lift on scroll works in Insights Management, View All, and Details screens.

Test 2:

- Navigate to Pages screen.
- Make sure that lift on scroll works with all the tabs and switching between tabs preserves lift on scroll state.
- Make sure that lift on scroll works with Search.

Test 3:

- Navigate to Posts screen.
- Make sure that lift on scroll works with all the tabs and switching between tabs preserves lift on scroll state.
- Make sure that lift on scroll works with Search.

Test 4:

- Navigate to the Media Browser screen.
- Make sure that lift on scroll works with all the tabs and switching between tabs preserves lift on scroll state.

Test 5:

- Using a site with unredeemed domain credit navigate to Domain Registration flow.
- Domain suggestions screen has an inner scroll view which should not trigger lift on scroll.
- Select a domain name from suggestion and proceed to contact info form.
- Make sure lift on scroll works as expected.

PR submission checklist:


osullivanchris (changes requested):
Generally looks great. Great attention to detail with tab switching, search and everything working so well. I have just spotted a couple of things I'm unsure of on these screens.

> Make sure that lift on scroll works with Search.

It works. But on initial load, maybe we need a divider like we've done above tabs? Feels a little strange having a search bar that's so flat to the view, with no border or distinction. WDYT?

> and mostly on Stats :)

There's one thing here. It feels a little bit wrong to me to have the shadow on the tabs, but then the other bar fixed beneath the tabs. I think its because it means that the tabs have depth, but the bar doesn't. But how its working, the bar is at on the same plane as the tabs, but above the content scrolling beneath it.

I think it would make more sense if the tabs did not have a shadow, in this instance, but the bar beneath had the shadow, per option A below. However, I understand this could be tricky, and too specific to a certain use case. So alternatively I think the logic would also be fine if the bar was not sticky/fixed, and scrolled beneath the tabs, per optionB.

Actually I think optionA is similar to what we're doing in reader with the filterbar there.

stats shadow


One last small one. On Stats, for example, we are scrolling the main header out of view. On other views, we are keeping it fixed. Is that just existing behaviour, or a decision we've made?

khaykov:
@osullivanchris 

> One last small one. On Stats, for example, we are scrolling the main header out of view. On other views, we are keeping it fixed. Is that just existing behaviour, or a decision we've made?

It's an existing behavior :)

> It works. But on initial load, maybe we need a divider like we've done above tabs? Feels a little strange having a search bar that's so flat to the view, with no border or distinction. WDYT?

Hm, I'm not a fan of adding and removing the divider. I checked some google apps, and they don't stick to any specific pattern. Maybe something we can do with the search bar style? Eg:
[Image from Gyazo](https://gyazo.com/1e7a902ba438eee6863a7a590ad13402)

> There's one thing here. It feels a little bit wrong to me to have the shadow on the tabs, but then the other bar fixed beneath the tabs. I think its because it means that the tabs have depth, but the bar doesn't. But how its working, the bar is at on the same plane as the tabs, but above the content scrolling beneath it.

> Actually I think optionA is similar to what we're doing in reader with the filterbar there.

The bar is actually below the tabs, it's on the same plane with the list, and behaves (and looks) in the same way as the filter bar in reader.

Saying all this, I was able to elevate it and hopefully make it work like the Option A:

[Image from Gyazo](https://gyazo.com/6db4373cd7def45e1e9c923ed47a6ccb)

[Image from Gyazo](https://gyazo.com/a3a163b028007ac47000df9b91ca4c2e)

[Image from Gyazo](https://gyazo.com/1016476f31c3abfd04c9557dff3a22a2)

[Image from Gyazo](https://gyazo.com/4e1554f066c967ad648fc12a791f0a51)

Let me know what you think!

osullivanchris:
> The bar is actually below the tabs, it's on the same plane with the list, and behaves (and looks) in the same way as the filter bar in reader.

Oh I see that now, had to look closely I guess I hadn't noticed because the tabs aren't sticky. To me it feels a little funky when there's two levels of depth. Elevating the lower bar, and having the shadow beneath both objects (with divider in the middle) is right to my eye.

> Saying all this, I was able to elevate it and hopefully make it work like the Option A:

So yeah per the above sentence, I think this is working! There's a few GIFs there though and I'm not 100% sure what is what. But https://gyazo.com/a3a163b028007ac47000df9b91ca4c2e looks per option A and I think its working well

I think we're probably at a stage now where it would be good to get the eyes of a few other designers and devs on it. Others might have strong opinions about particular places in the app and ways we should use components. Might be good to have a design review session and get opinions on these details before we get too deep and have a polished, final thing that then needs to be changed if people disagree with us 😄

oguzkocer:
Hi there! `15.5` is being frozen, so this PR is being bumped to `15.6`. If these changes _have to_ be shipped with `15.5`, please target `release/15.6` and cc me in the PR so I can cut a new beta release. Also if you can keep the release dates in mind and bump the PRs & issues you are working on before the code freeze date, we'd really appreciate it.

---

As mentioned in https://github.com/wordpress-mobile/WordPress-Android/pull/12743#issuecomment-679037820, a milestone is not necessary for PRs against feature branches.

WordPress/src/main/java/org/wordpress/android/ui/domains/DomainRegistrationDetailsFragment.kt
```diff
@@ -519,4 +520,12 @@ class DomainRegistrationDetailsFragment : Fragment() {
             AndroidSupportInjection.inject(this)
         }
     }
+
+    override fun onResume() {
+        super.onResume()
+        if (activity is ScrollableViewInitializedListener) {
+            (activity as ScrollableViewInitializedListener)
```
planarvoid: you could do something like `(activity as? ScrollableViewInitializedListener)?.onScrollableViewInitialized` which is a bit shorter :-)

WordPress/src/main/java/org/wordpress/android/ui/pages/PagesFragment.kt
```diff
@@ -136,8 +139,10 @@ class PagesFragment : Fragment() {
     override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
         if (requestCode == RequestCodes.EDIT_POST && resultCode == Activity.RESULT_OK && data != null) {
             if (EditPostActivity.checkToRestart(data)) {
-                ActivityLauncher.editPageForResult(data, this@PagesFragment, viewModel.site,
-                        data.getIntExtra(EditPostActivity.EXTRA_POST_LOCAL_ID, 0), false)
+                ActivityLauncher.editPageForResult(
+                        data, this@PagesFragment, viewModel.site,
+                        data.getIntExtra(EditPostActivity.EXTRA_POST_LOCAL_ID, 0), false
```
planarvoid: I think all the params should be on separate lines. I'm surprised lint doesn't complain.

WordPress/src/main/java/org/wordpress/android/widgets/AppBarFadingEdgeView.kt
```diff
@@ -0,0 +1,74 @@
...
+                intArrayOf(
+                        ContextCompat.getColor(context, color.transparent),
+                        context.getColorFromAttribute(attr.wpColorAppBar)
+                )
+        )
+
+        val fadingAgeStateListDrawable = StateListDrawable()
+        fadingAgeStateListDrawable.addState(
+                intArrayOf(
+                        -com.google.android.material.R.attr.state_lifted,
```
planarvoid, line 66: I'm wondering how this works. Are you intentionally trying to use negative IntResource? or would you expect the value here?

planarvoid (commented):
hey @khaykov , it works really well! I've left a few minor comments + I've tested all the flows and I think I've found one bug in the last scenario: 

Test 5:
> Select a domain name from suggestion and proceed to contact info form.
> Make sure lift on scroll works as expected.

For me the list is lifted even without scrolling when I enter the screen and I think that's not expected, right? 

lift-visible-on-domain-registration

WordPress/src/main/java/org/wordpress/android/widgets/AppBarFadingEdgeView.kt
khaykov: Oh man, this is a weird and interesting decision that Android team made. Android is using resource ID integer to indicate the state of the view and since there are are no opposite states (eg. there is no state_disable, or state_not_lifted) they decide to use a negative resource ID integer to indicate opposite state. This kind of makes sense, but I wonder if there are better ways to do it 🤷 

Regardless, it's mentioned in [StateSet](https://developer.android.com/reference/android/util/StateSet) documentation, so I hope I'm not imagining things :)

WordPress/src/main/java/org/wordpress/android/ui/domains/DomainRegistrationDetailsFragment.kt
khaykov: Nice! Thanks for the tip :)

khaykov:
@planarvoid Thank you so much for the review! Hope I addressed all the comments :)

WordPress/src/main/java/org/wordpress/android/widgets/AppBarFadingEdgeView.kt
planarvoid: Looking into this makes my head spin. It seems your implementation is correct 👍  but man this is not how you design a standard interface... 😄

khaykov:
@osullivanchris @planarvoid if the PR looks OK, could you please approve? :)

planarvoid (approved):
thanks for the fixes! It looks good 👍  Feel free to merge this PR after @osullivanchris approves it

osullivanchris (approved):

osullivanchris:
Sorry just saw this and approved now. I don't think there were any outstanding questions from my side!

PR #12678, opened 2020-08-14
Create media picker launcher

planarvoid:
Create a media picker launcher that is used to trigger either a PhotoPickerActivity or MediaPickerActivity based on the flag. This should be replacing all the places where we start a PhotoPickerActivity. As part of this PR I've extracted the constants into a reused class called `MediaPickerConstants` so minimize a chance of breaking the contract when switching the flags.

To test:
- Turn off the flag
- Make sure the photo picker works (Image/Video/Gallery blocks, Stories) 
- Turn on the flag
- Make sure the photo picker works (Image/Video/Gallery blocks, Stories)

PR submission checklist:


develric:
Hi @planarvoid , code looks good overall 👍 ; I'm still completing the testing but I wanted to anticipate this issue I found (currently got it with FF disabled in WasabiDebug):

- go to Me screen
- tap on the avatar to change/set it
- check you get the following exception

```
2020-08-25 14:25:23.647 15591-15591/org.wordpress.android.beta E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.wordpress.android.beta, PID: 15591
    kotlin.TypeCastException: null cannot be cast to non-null type org.wordpress.android.fluxc.model.SiteModel
        at org.wordpress.android.ui.photopicker.PhotoPickerFragment.onViewCreated(PhotoPickerFragment.kt:107)
```

UPDATE: just noting that the same is true with active FF for MediaPickerFragment line 105

develric:
Just noting 2 things:
- I didn't find in my further testing any issues a part the one reported above
- Was trying to test the modification in `SignupEpilogueFragment` but since there is this [open PR fix ](https://github.com/wordpress-mobile/WordPress-Android/pull/12783)not already merged, I'm noting it here to test that scenario (i think we can also test that in subsequent PRs if that one is not merged and we are ready to go ahead, since we are at the beginning of the sprint)

renanferrari:
FYI: The crash reported by @develric above seems to be present since version 15.5, which is already in staged rollout. We should probably work on a hotfix for it. Let me know if I can help in any way.

planarvoid:
hey @develric, thanks for the review! as discussed, the crash was already fixed and merged in another PR so this should be ready for another check/merge. Thanks for the offer @renanferrari 👍

develric:
Hi @planarvoid I'm doing a test round here and after the merge from develop I could now check the `SignupEpilogueFragment` scenario; in this scenario at least in my setup I'm getting an issue with setting the avatar image:

- Go in the signup flow
- in the SignupEpilogueFragment try to set an avatar image
- you can select the image from the picker but the image is not set in the placeholder

in logcat I do not get exceptions.

Let me know if you can reproduce or if you want me to make some additional checks on that, thanks 🙇‍♂️

develric:
hi @planarvoid , leaving a quick note: I think we are missing the safe cast [here](https://github.com/wordpress-mobile/WordPress-Android/blob/2d7aaa2295ff2683fd80c5b0afd7af6652757f51/WordPress/src/main/java/org/wordpress/android/ui/photopicker/mediapicker/MediaPickerFragment.kt#L105) as well :)

planarvoid:
hey @develric , thanks for the review. Both of the items should be fixed now 👍

develric (approved):
Hi @planarvoid and thanks for the follow up. I confirm both issues are fixed 👍 ! :shipit:

PR #12693, opened 2020-08-17
Extract source interface for media picker

### LINKED ISSUE #12638: Extract an interface from the DeviceListBuilder to enable various sources
The PhotoPicker uses only local files right now. We should create a new `Source` interface that would support various sources of data (for example the WordPress media and the free photo library). The new interface should also support search and pagination. A good choice would be using a repository pattern.

planarvoid:
Fixes #12638 

This PR replaces the original DeviceListBuilder that was used to load media from the device with an interface and a class MediaLoader. This interface and the loader should support all kind of sources in the future. The loader supports pagination, refresh of data and error responses (and could be extended more). The functionality should be unchanged right now. 

The pagination is not used at the moment because the device list doesn't need to be paginated. We might add it as a feature in the future (because loading a lot of items might not be very efficient).

To test:
- test that the following functionality works as expected
- GB - Image block - Add image from device
- GB - Video block - Add video from device
- GB - Gallery block - Add images from device
- Add story -> Choose images and videos

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/photopicker/mediapicker/MediaItem.kt
```diff
@@ -5,5 +5,6 @@ import org.wordpress.android.util.UriWrapper
 data class MediaItem(
     val id: Long = 0,
     val uri: UriWrapper? = null,
+    val name: String? = null,
```
develric, line 8: Do we need this `name` field? I guess it could be for a future PR/usage?

develric (commented):
Hi @planarvoid , love usage of flow and channels, good job 👍 ! I followed the test steps and didn't find issues. I only left a comment that is more out of curiosity than anything, but feel free to ignore 😊  !

I'm not approving/merging this one so we can follow up to the landing #12678 first, but let me know if you think different 🙇‍♂️ .

WordPress/src/main/java/org/wordpress/android/ui/photopicker/mediapicker/MediaItem.kt
planarvoid: we don't need it, I was thinking ahead but right now it's not required. Once we implement search, we'll need it though. I'll check if I can remove it.

edit: let's keep it for now. The search already works around this field. Once we do search, we can check whether this is the best way to use it.

develric (approved):
Thanks for the clarification 🙇‍♂️ ! LGTM 👍 !

PR #12694, opened 2020-08-17
Fix/reader blog table

zwarm:
This PR fixes an issue with identifying a feed before a post makes it to the `ReaderSiteHeaderView` code.

Modifications include

- A new static method was added to `ReaderUtils` to determine if this is a feed and uses this value to make decisions in the header code. 
- When parsing data for blogtable, make sure to pull blogId in from response
- When requesting to follow a site from the header, don't send down the path to display enable notifications when this is a feed.
- Updated logic in `ReaderActivityLauncher` to check if this is a feed before adding feedId or blogId to arguments 
- Removed old code associated with previous [Update Reader Header Site UI ](https://github.com/wordpress-mobile/WordPress-Android/pull/12655)
- Removes animation from Following button

**Open question**: Should `ReaderUtils.isFeed` be renamed to `isExternalFeed`?

To test:
Test A: validate follow and enable subscription works for a site
- Log into the app
- Search for a site
- Tap the post header line to bring up the post card
- Tap the follow button in the header & quickly tap on the ENABLE notifications text in the snackbar
- Leave the app open because you'll need it again
- Validate that you are subscribed to the site you selected (Use your Wordpress.com site or rest client to run `/read/following/mine?meta=site,feed`)
- Tap the unfollow button for the same site
- Validate that the subscription for the site is gone
- Log out of the app

Test B: Validate that you are not prompted to enable notifications on tap of follow from header
- Log into the app
- Go to reader -> gear -> 
- Enter an RSS feed to follow: https://timesofindia.indiatimes.com/rssfeedstopstories.cms & tap the follow icon
- Back out to the reader and notice the feed shows up in following
- Tap on the header line of the post card
- The new Site header is shown as "Following"
- Tap the "following" button to unfollow
- The following button turns pink
- Tap to Follow 
- Note that you are not prompted to enable subscriptions - no snackbar is visible

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/datasets/ReaderBlogTable.java
```diff
@@ -382,25 +382,6 @@ public static boolean isTimeToUpdateBlogInfo(ReaderBlog blogInfo) {
         return (minutes >= ReaderConstants.READER_AUTO_UPDATE_DELAY_MINUTES);
     }
 
-    /**
```
zwarm, line 385: This was code was added in the first PR and can safely be removed because it is not used.

WordPress/src/main/java/org/wordpress/android/datasets/ReaderPostTable.java
```diff
@@ -1157,18 +1157,4 @@ private static void updateIsBookmarkedField(final ReaderPostList posts) {
...
         return getBlogIdPostIdsWithTagType(ReaderTagType.BOOKMARKED, 99999);
     }
-
-    /**
-     * getBlogIdForFeed is a helper method that will query the post table for a post with the requested
-     * feedId. It will return a limit of 1
-     * @param feedId
-     * @return The associated blogId
-     */
-    public static Long getBlogIdForFeed(long feedId) {
```
zwarm, line 1167: This was code was added in the first PR and can safely be removed because it is not used.

WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderBlogActions.java
```diff
@@ -505,87 +505,4 @@ public void onErrorResponse(VolleyError volleyError) {
...
-    /**
-     * Updates the imageUrl for a feed. Since the /read/feed/{feedId} endpoint does not have this data,
-     * the request is made to the /read/sites/{blogId} endpoint.
-     * @param blogId
-     * @param blogUrl
-     * @param feedId
-     * @param infoListener
-     */
-    public static void updateImageForFeedByBlog(long blogId,
-                                      final String blogUrl,
```
zwarm, line 518: This was code was added in the first PR and can safely be removed because it is not used.

WordPress/src/main/java/org/wordpress/android/ui/reader/views/ReaderSiteHeaderView.java
```diff
@@ -237,18 +193,18 @@ private void toggleFollowStatus(final View followButton) {
...
-        if (mFeedId != 0) {
+        if (mIsFeed) {
             isAskingToFollow = !ReaderBlogTable.isFollowedFeed(mFeedId);
         } else {
             isAskingToFollow = !ReaderBlogTable.isFollowedBlog(mBlogId);
         }
 
         if (mFollowListener != null) {
-            if (isAskingToFollow) {
+            if (isAskingToFollow && !mIsFeed) {
```
malinajirka, line 203: Could you please elaborate on this condition - it seems a bit suspicious -> the user clicks on `follow` and we perform `unfollow` action when it's a feed?

zwarm: Ash, I see the confusion. The listener handlers don't do any server logic, the followed one will dispatches a newFetchSubscriptionsAction and show a snackbar (which we don't want to show in the case of feed). The isFollowing just dispatches a newFetchSubscriptionsAction. I adjusted so it goes down the same path, but will not show a snackbar if blog is not > 0. Easier to understand. Adjusted in e59a23b

malinajirka:
Thanks @zwarm! It looks good overall. However, I encountered one issue.

>Test B: Validate that you are not prompted to enable notifications on tap of follow from header

I can see the snackbar in this scenario

zwarm:
> I can see the snackbar in this scenario

Ah, yes, that blogId > 0 had to be added into ReaderPostListFragment as well . That's been committed. The check in the header code took care of both places, had to readd to the other handler in commit 18f72ed

malinajirka (approved):
Works as described. Thanks! ;)

PR #12695, opened 2020-08-17
Switch translation reviewers builds to release

loremattei:
This PR switches to release builds for the "Translation Reviewers" build task. 

cc @akirk 

To test:
- Make sure CI is green. 

PR submission checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -354,13 +354,14 @@ jobs:
             VERSION_NAME="${APP_VERSION_PREFIX}-build-${TODAY_DATE}-${CIRCLE_BUILD_NUM}"
             echo "export VERSION_NAME=$VERSION_NAME" >> $BASH_ENV
 
+            cp ~/.android/debug_a8c.keystore ~/.android/debug.keystore
             bundle exec fastlane build_for_translation_review custom_version:"$VERSION_NAME"
       - android/save-gradle-cache
       - run:
           name: Prepare APK
           command: |
             mkdir -p Artifacts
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -354,13 +354,14 @@ jobs:
             VERSION_NAME="${APP_VERSION_PREFIX}-build-${TODAY_DATE}-${CIRCLE_BUILD_NUM}"
             echo "export VERSION_NAME=$VERSION_NAME" >> $BASH_ENV
 
+            cp ~/.android/debug_a8c.keystore ~/.android/debug.keystore
```
mokagio, line 357: > When running or debugging your project from the IDE, Android Studio automatically signs your app with a debug certificate generated by the Android SDK tools. The first time you run or debug your project in Android Studio, the IDE automatically creates the debug keystore and certificate in $HOME/.android/debug.keystore, and sets the keystore and key passwords.
>
> Because the debug certificate is created by the build tools and is insecure by design, most app stores (including the Google Play Store) do not accept apps signed with a debug certificate for publishing.
From the [Google docs](https://developer.android.com/studio/publish/app-signing#debugmode).

I think here it's okay to use the debug keystore because the translation build is only meant to be used internally and is distributed outside of the Play Store.

loremattei: Yup! Also, the Jalapeno build we use here is allowed to access the APIs when the apk is signed with the shared debug keystore only. The debug builds are already configured to use it, but the release ones expect other keys, so we have to use a trick here. This is the main reason for this change.

mokagio: Thanks for the extra context.

PR #12697, opened 2020-08-17
WP Stories: remove `post_format` handling

mzorz:
Removes Post format as a way to tell if a Post contains a Story.

Builds on top of #12733

cc @aforcier 

ToDo:
- [x] remove `isThisAnAlphaTestSite` checks once we have a way to tell which sites support the jetpack Stories block so we don't need to check for stories alpha sites anymore (should be taken care of by #12729)
- [x] remove `buildWPGalleryInPost`  code path in `SaveStoryGutenbergBlockUseCase` and related classes so we don't have the plain Gallery as a fallback (no need to anymore?) (should be taken care of by #12729)
- [x] make sure whether we still need / don't need `replaceMediaLocalIdWithRemoteMediaIdInPost` in `MediaUploadReadyListener`  (I think we still need it because in Stories we first create a Post, then add local media with local Ids, then update the Post, so we may still need to keep this). (should be taken care of by https://github.com/wordpress-mobile/WordPress-Android/pull/12734)
- [x] implement checks to offer / hide create a new Story post depending on block availability https://github.com/Automattic/stories-android/issues/417#issuecomment-669641911 (should be taken care of by #12728)

To test:
1. start a story
2. add a couple of new images
3. publish
4. verify it got published as a Story block.

PR submission checklist:


mzorz:
> make sure whether we still need / don't need replaceMediaLocalIdWithRemoteMediaIdInPost in MediaUploadReadyListener (I think we still need it because in Stories we first create a Post, then add local media with local Ids, then update the Post, so we may still need to keep this).

Note:  did a cleanup step left from #12729, that removed a now unused code path since the removal of the Gallery fallback mechanism for sites that don't support Stories. See bd663facd7

aforcier (approved):
Works great! :shipit:

PR #12712, opened 2020-08-18
Replace bottom bar with action mode icon in media picker

planarvoid:
This PR introduces UI changes to the consolidated media picker. This includes 2 main things:
- The "Edit" action for images is now moved to the action mode toolbar. It's visible if only images are selected
- The Bottom bar is gone (so I've removed all the logic around it and most of the logic around icon clicks)

To test:
- Go to Gutenberg editor
- Add an image block
- Click on "Add media"
- Click on "Choose from device"
- The media picker is opened
- Notice there is no bottom bar
- Select an item
- Notice that there is an edit button and a confirm button in the action bar toolbar
- Test both work as expected

To test 2:
- Go to Gutenberg editor
- Add an video block
- Click on "Add media"
- Click on "Choose from device"
- The media picker is opened
- Notice there is no bottom bar
- Select an item
- Notice that there is a confirm button in the action bar toolbar and the edit button is not visible
- Test the confirm button works as expected

To test:
- Go to Gutenberg editor
- Add an gallery block
- Click on "Add media"
- Click on "Choose from device"
- The media picker is opened
- Notice there is no bottom bar
- Select a few items
- Notice that there is an edit button and a confirm button in the action bar toolbar
- Test both work as expected

PR submission checklist:


WordPress/src/main/res/layout/media_picker_fragment.xml
```diff
@@ -0,0 +1,52 @@
...
+    android:background="@android:color/white">
+
+    <com.google.android.material.floatingactionbutton.FloatingActionButton
+        android:id="@+id/wp_stories_take_picture"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="end|bottom"
+        android:visibility="gone"
+        android:backgroundTint="@color/pink_50"
+        android:tint="@color/white"
```
develric: Comparing this with the `photo_picker_fragment.xml` it seems we have the following added:

```
    android:background="@android:color/white">                  <------

    <com.google.android.material.floatingactionbutton.FloatingActionButton
        android:id="@+id/wp_stories_take_picture"
        android:layout_width="wrap_content"
        android:layout_height="wrap_content"
        android:layout_gravity="end|bottom"
        android:visibility="gone"
        android:backgroundTint="@color/pink_50"                     <------
        android:tint="@color/white"                     <------
```

not sure if it was removed from the `deprecated` class instead 🤔

WordPress/src/main/java/org/wordpress/android/ui/photopicker/mediapicker/MediaPickerViewModel.kt
```diff
@@ -335,119 +275,25 @@ class MediaPickerViewModel @Inject constructor(
...
+    private fun clickIcon(icon: MediaPickerIcon) {
+        if (icon == WP_STORIES_CAPTURE) {
             if (!permissionsHandler.hasPermissionsToAccessPhotos()) {
                 _onPermissionsRequested.value = Event(PermissionsRequested.CAMERA)
                 lastTappedIcon = icon
                 return
             }
-        }
-        when (icon) {
-            ANDROID_CAPTURE_PHOTO -> trackSelectedOtherSourceEvents(
```
develric, line 349: Probably I'm just missing something here, but shouldn't we track these events somewhere else? (sorry in case it's a stupid point 🙃).

develric (commented):
Hi @planarvoid , checked this one and is looking good. I left a couple of comments for your feedback; let me know, thanks 😊

WordPress/src/main/java/org/wordpress/android/ui/photopicker/mediapicker/MediaPickerViewModel.kt
planarvoid: no worries. These icons were removed from the UI and the enum so this when no longer makes sense 👍

WordPress/src/main/res/layout/media_picker_fragment.xml
planarvoid: good point! I've implemented the changes from photo_picker_fragment here e48ff142f29e84adb0a2f63d1a782fd379ff8371

planarvoid:
thanks for the review @develric ! I've implemented your suggestion & answered your question

develric (approved):
Thanks for the follow up @planarvoid , LGTM 👍 !

PR #12713, opened 2020-08-18
WP Stories: build the VideoPress URL if mimeType is `video/videopress`

### LINKED ISSUE #12711: WP Stories: videos URL not welformed
### Expected behavior
Creating a story on a whitelisted site should produce a jetpack Story GB block with well-formed URLS for both pics and videos

### Actual behavior
Video URls are not right, see for example
```
<!-- wp:jetpack/story {"mediaFiles":[{"alt":"","caption":"","id":252,"link":"https://dekervit.wpsandbox.me/wp-1597759601712-mp4/","mime":"video/mp4","type":"video","url":"https://dekervit.wpsandbox.me/wp-1597759601712-mp4/"},{"alt":"","caption":"","id":251,"link":"https://dekervit.wpsandbox.me/wp-content/uploads/2020/08/wp-1597759601752.jpg","mime":"image/jpeg","type":"image","url":"https://dekervit.wpsandbox.me/wp-content/uploads/2020/08/wp-1597759601752.jpg"}]} -->
<div class="wp-block-jetpack-story wp-story"></div>
<!-- /wp:jetpack/story -->
```

The URL here reads `"url":"https://dekervit.wpsandbox.me/wp-1597759601712-mp4/"`

1. switch site to a whitelisted story site create a story
2. cretae a new story
3. add a video slide
4. tap NEXT and then add some title to the post
5. upload
6. verify the url is wrong and the video can't be played

mzorz:
Fixes #12711

The problem described in the aforementioned issue was due to having the target site enable VideoPress, but we had no 
support for VideoPress as of yet.

This PR fixes that by simply building up the VideoPress URL correctly for the case when the video's `mime` type is `video/videopress`.

**Note:** this other branch [here  ](https://github.com/wordpress-mobile/WordPress-Android/compare/issue/12711-fix-video-press-url?expand=1) explores what I had in mind, but it turns out the SiteModel being passed around is not needed (the value for `mIsVideoPressSupported`  is false for our whitelisted test site anyway), and we'd also need to do more work than what can be seen on that branch to actually support the original file URL in FluxC. I think it may not be a great idea to do all of that for VideoPress URLs given the content subdomain seems to have been fixed for a long time and has been kept as is.
If / when the case happens that different sites may have different URLs for VideoPress videos, we can go ahead and implement support for this on FLuxC to correctly populate a new field with this information.
Added a `TODO` comment there to remind our future selves when/if this is needed.

To test:
1. switch site to a whitelisted story site create a story
2. cretae a new story
3. add a video slide
4. tap NEXT and then add some title to the post
5. upload
6. verify the url is OK and the video plays correctly

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCase.kt
@@ -4,6 +4,7 @@ import com.google.gson.Gson
 import org.wordpress.android.fluxc.model.PostModel
 import org.wordpress.android.ui.posts.EditPostRepository
 import org.wordpress.android.ui.posts.PostUtils
+import org.wordpress.android.util.ShortcodeUtils
 import org.wordpress.android.util.StringUtils
 import org.wordpress.android.util.helpers.MediaFile
 import javax.inject.Inject
@@ -85,12 +86,22 @@ class SaveStoryGutenbergBlockUseCase @Inject constructor() {
         }
 
         // now replace the same in the mediaFileObjects, obtain the URLs and replace
+        val mediaUrl = getMediaFileUrl(mediaFile)
         storyBlockData?.mediaFiles?.filter { it.id == localMediaId }?.get(0)?.apply {
             id = mediaFile.mediaId.toInt()
-            link = mediaFile.fileURL
-            url = mediaFile.fileURL
+            link = mediaUrl
+            url = mediaUrl
         }
         post.setContent(createGBStoryBlockStringFromJson(requireNotNull(storyBlockData)))
+    }
+
+    private fun getMediaFileUrl(mediaFile: MediaFile): String {
+        if (VIDEOPRESS_MIME_TYPE.equals(mediaFile.mimeType)) {
+            return VIDEOPRESS_BASE_URL + ShortcodeUtils.getVideoPressIdFromShortCode(mediaFile.videoPressShortCode) +
+                    "/" + mediaFile.fileName
+        } else {
+            return StringUtils.notNullStr(mediaFile.fileURL)
+        }
     }
 
     private fun createGBStoryBlockStringFromJson(storyBlock: StoryBlockData): String {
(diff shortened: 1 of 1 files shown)


aforcier:
I was curious what happens when uploading a video from Gutenberg, it looks like the video block knows how to interpret that kind of URL:

```
<!-- wp:video {"autoplay":false,"guid":"NhaXbXse","id":305,"loop":false,"muted":false,"playsInline":false,"src":"https://site.me/wp_story1597806821800_0-mp4/"} -->
<figure class="wp-block-video wp-block-embed is-type-video is-provider-videopress"><div class="wp-block-embed__wrapper">
https://videopress.com/v/NhaXbXse?preloadContent=metadata
</div></figure>
<!-- /wp:video -->
```

So perhaps the ideal fix is changing the story block to support videopress URLs like that, and including the videopress guid on our end? cc @Tug

I'll proceed with reviewing this anyway, but the above seems like a better long-term fix.

aforcier (approved):
Working well @mzorz - :shipit:!

aforcier:
Just noting that my comment above came up in conversation with Tug and he mentioned it might be possible to have the story block rely on the video id alone, and remove the need for us to assemble the URL 'by hand' on mobile. He'll look into it but it's agreed that it's not very high priority and we're okay with the solution in this PR for now.

mzorz:
Agreed, the best path forward is having the block make the right interpretation - we'll dig into that when we get to https://github.com/Automattic/stories-android/issues/504 (just realized it was being tracked as a note in https://github.com/Automattic/stories-android/issues/249 but it wasn't being tracked as a separate issue that I could find); I had a few notes on the side for when the time comes, but given this came up here as well leaving them here for completeness:

see https://github.com/wordpress-mobile/gutenberg-mobile/pull/854#issuecomment-490440013

> To do this on Mobile, we might have to import the JetPack block, right?

> I think we just need to read the VideoPress hash, query the native side (there is an API call that from a VP hash does return video info) and replace it with the video URL when the call returns.

Related https://github.com/Automattic/jetpack/issues/12358

aforcier:
Quick follow-up, I believe with https://github.com/Automattic/jetpack/pull/16663/commits/d81748c419239fcba7ccfed981c44e6fff527a82 the story block now just needs the media library ID to render the video (and will resolve the appropriate URL on its own). We were already sending that, so I think this PR can be reverted and we can avoid relying on a hardcoded videopress URL. Seems to be working from my own tests with a build before this PR. @Tug  might have some more specifics.

PR #12718, opened 2020-08-19
Convert media picker activity to kotlin

planarvoid:
This PR converts MediaPickerActivity to Kotlin in order to make future changes easier. There are no functionality changes

To test:
- Go to Media picker (add a new story)
- Check that it starts correctly
- Select items
- Make sure the items are selected correctly when you quit the activity

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/photopicker/mediapicker/MediaPickerActivity.kt
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    override fun onMediaChosen(uriList: List<Uri>) {
+        if (uriList.size > 0) {
+            doMediaUrisSelected(uriList, APP_PICKER)
+        }
+    }
+
+    override fun onIconClicked(icon: MediaPickerIcon, allowMultipleSelection: Boolean) {
+        when (icon) {
```
develric, line 365: maybe we could stay with an `if` here? really small np, so feel free to ignore 😊

develric (approved):
Hi @planarvoid , checked this one and it LGTM 👍 ! I just left a super minor np 😊 !

WordPress/src/main/java/org/wordpress/android/ui/photopicker/mediapicker/MediaPickerActivity.kt
planarvoid: my thinking was that if we add more actions, the `when` will let us know we missed it 🤷  but I'm fine with both

planarvoid:
thanks for the review @develric ! let me know what you think about my answer. I'm cool with both approaches 👍

WordPress/src/main/java/org/wordpress/android/ui/photopicker/mediapicker/MediaPickerActivity.kt
develric: Thanks for sharing the reasoning behind. I'm fine with it 🙇‍♂️

develric:
> thanks for the review @develric ! let me know what you think about my answer. I'm cool with both approaches 👍

Your reasoning makes sense to me, thanks for sharing 🙇‍♂️ ! :shipit:

Issue #12601, opened 2020-08-05
Create WPAndroid screenshots on CI


loremattei:
We have a Fastlane setup to create enhanced screenshots for WPAndroid. It runs a dedicated set of UI tests which use mocked API calls and takes basic screenshots. Then, the screenshots are processed to add promo text, composites, etc. 

Currently:
- The screenshot generation lane has to be run on a local machine and takes quite a bit of time. 
- There are some small issues with some languages that need to be fixed. 

It would be nice to make it run on CI, triggered by a manual event (adding a label or opening a PR with a define name pattern or an API call, etc.). The final screenshots should be available for download, ready to be uploaded with Fastlane to the Play Store. 

In order to make this work, the following issues need to be fixed: 
- Screenshot for `sr` doesn't get generated.
- Screenshot generation for `ar` crashes in stats section due to a mocked data parsing issue.
- `id` locale screenshots are not generated properly. `in` seems to work. It seems to be one of the usual issues due to Android having changed some locale shortcodes in the past.
- Promo screenshots should be generated under the folder and with the layout that Fastlane expects to upload them, and the original simple screenshots shouldn't be there at the end of the job.

Stretch goal: upload them to Play Store from CI, after user review.

AliSoftware:
@loremattei 

> Screenshot generation for ar crashes in stats section due to a mocked data parsing issue.

Is this still true? I've run the lane locally and see all screenshots for `ar` properly generated on my Machine in `fastlane/metadata/android` (and also in `fastlane/promo_screenshots` after running the `create_promo_screenshots` lane)

> Promo screenshots should be generated under the folder/layout that Fastlane expects to upload them, and the original simple screenshots shouldn't be there at the end of the job.

Currently, raw screenshots are in `fastlane/metadata/android` and promo screenshots are in `fastlane/promo_screenshots`. I can change the output from `fastlane/promo_screenshots` to `fastlane/layout` if you prefer, but I'm not sure what you expect in that sentence about the raw screenshots:
 - should I delete `fastlane/metadata/android/**/images` even tho they are not in the same folder as the `promo_screenshots` and don't risk to cause any confusion?
 - What's the point of removing them is the screenshots are taken by the CI anyway?
 - Or is this point obsolete and was written at an earlier point where both the promo screenshots and the raw screenshots were all in the same folder?

loremattei:
> Is this still true? I've run the lane locally and see all screenshots for ar properly generated on my Machine

[...]

PR #12723, opened 2020-08-19
[Platform] Fix some PlayStore screenshots automation issues

AliSoftware:
Fix parts of #12601

## Scope

This PR addresses the following issues about screenshots:

* Screenshots for `sr` locale were not generated. That's because they were explicitly disabled in the `SUPPORTED_LOCALES` constant
* Screenshots for `id` locale were not generated properly: the screenshots were in English instead of Indonesian. This is because the locale short code recently changed from `id` to `in`

## Main changes

* `sr` was re-enabled using `promo_config: {}` instead of `promo_config: false`
* Indonesian locale was fixed by using `in` locale instead of `id` for `google_play` key, which is used both to tell the emulator which locale code to use and the name of the final `fastlane/metadata/android/<locale>/images` folder where the screenshots are made

## Fix for obsolete languages / folder renaming

The `SnapshotFastfile` also had a `fix_obsolete_locales_for_screenshots(locales: [{source_id:'in', target_id:'id'}], output_directory: "metadata/android")` method
   - This method aimed to move the `android/metadata/in` folder to `android/metadata/id`
   - This got introduced back in Nov 2018 by @loremattei (in c325379cc0b30fb14b0032d9d265f2014dce9d48), at a time when it seems `in` was considered the obsolete folder and `id` the new one.
  - I wonder if back in 2018 the situation was reversed and the locale went from `in` (pre-2018) to `id` (2018) back to `in` again (now)? – which would explain why that fix was in this direction back then, and we need to revert it today – but would love a confirmation to validate my understanding of that history.
   - In any way, since `in` is nowadays the new code (and `id` the old one), and that in practice GooglePlay seems to [support both](https://support.google.com/googleplay/android-developer/table/4419860?hl=en) `in` and `id`, the fix seems obsolete now.

This also mean that I had to move the metadata files (`full_description.txt`, `short_description.txt` etc) from `fastlane/metadata/id` to `fastlane/metadata/in` as `in` is the new folder name serving as reference for Indonesian metadata, and where the images will end up anyway

## Symlink

`WordPress/sec/main/res/values-in` is already a symlink which points to `values-id`, which allows us to support both locale codes old and new
_Note: When creating a new resource for the Indonesian language, Android Studio suggests `values-in` as the folder name (not `values-id`), confirming that it seems to be the new code to use_

I think it's still valuable to keep the symlink (as opposed to only keeping `values-in` and getting rid of `values-id`), in order to include both locale codes for different versions of the Java API depending on the devices the app runs on, since we seem to still support API 21+.

### Using region-specific locale

While `in` is the new locale code for the Indonesian language, the region-specific code `in-ID` could also be used. Multiple answers on Google suggested to use that code to replace of `id`. But in our case since we target Indonesian-the-language without any region-specific translations, `in` makes more sense.

Besides, using `google_play: 'in-ID'` in the `Fastfile` would cause more issues, because:
 * The `google_play` key is used both to tell which locale to run the emulator in, and to upload the screenshots to the PlayStore. But the PlayStore seems to only support `in` or `id`, not `in-ID` according to [this page](https://support.google.com/googleplay/android-developer/table/4419860?hl=en)
 * We could consider having different keys in `SUPPORTED_LOCALES` for `google_play` vs `emulator` – the latter defaulting to the value of `google_play` if not defined.
 * This would require to adapt some folder names to match the whole logic too

In practice, using `in` makes more sense in our context and makes for a simpler implementation anyway

## To test:

- [x] Run `bundle exec fastlane android screenshots` (alternatively, run `bundle exec fastlane android screenshots locale:sr` then `bundle exec fastlane android screenshots locale:in`)
- [x] Go take a coffee (full test took 109 minutes on my machine!)
- [x] Check that the `sr` screenshots got generated
- [x] Check that the `in` screenshots are properly translated in Indonesian.

---

PR submission checklist:


loremattei (approved):
Thank you @AliSoftware!

AliSoftware:
Thx 👍 Just consequently undrafed #12737

PR #12725, opened 2020-08-19
Issue/12028 tracking events

zwarm:
Parent Issue #12028 

This PR adds tracking for the following (matching [iOS]( https://github.com/wordpress-mobile/WordPress-iOS/pull/14619)):

**Interests**: select_interests_shown & select_interests_picked
**Discover**: reader_discover_shown(updated in https://github.com/wordpress-mobile/WordPress-Android/pull/12818), reader_discover_paginated, reader_chips_more_toggled, reader_discover_topic_tapped
**All reader**: reader_site_shown, reader_post_card_tapped, reader_pull_to_refresh, post_care_more_tapped

The following events are already tracked in the reader + new discover, so testing is not needed
reader_site_followed, reader_site_unfollowed, item_shared_reader, reader_article_visited

The following events are already tracked in the reader, but not in new discover. They are not implemented in this PR because the use cases have not been implemented yet. There is a separate task for thse
followed_blog_notifications_reader_menu_off, followed_blog_notifications_reader_menu_on, reader_blog_blocked

**To test:**
**selectInterestsShown**

- Tap reader
- Remove all tags you're following (Following -> Cog -> tap on the x next to each tag)
- Tap Discover
- The select interests should appear. 
- Check the log for: WordPress-STATS: 🔵 Tracked: select_interests_shown

**selectInterestsPicked**

- Do the steps from the above
- Select one or more topics
- Tap Done
- Check the log for: WordPress-STATS: 🔵 Tracked: select_interests_picked

**readerDiscoverShown** (updated in https://github.com/wordpress-mobile/WordPress-Android/pull/12818)

- Tap Discover tab
- Check the log for: WordPress-STATS: 🔵 Tracked: reader_discover_shown  

**readerSiteShown**

- Tap any blog title to display its posts
- Check the log for: WordPress-STATS: 🔵 Tracked: reader_site_shown

**readerDiscoverPaginated**

- Tap Discover
- Scroll down to the end of the list until new posts appear
- Check the log for: WordPress-STATS: 🔵 Tracked: reader_discover_paginated

**readerChipsMoreToggled**

- Tap reader
- Remove all tags you're following (Following -> Cog -> remove)
- Tap Discover
- Select interests (about 6 or 7). 
- Tap done
- Scroll to a card with a chip set that shows the `+ Chip`
- Tap on the '+ Chip' to expand the tags list
- Check the log for: WordPress-STATS: 🔵 Tracked: reader_chips_more_toggled 
- Tap the Hide All chip
- Check the log for: WordPress-STATS: 🔵 Tracked: reader_chips_more_toggled 

**readerCardTapped**

- Tap any Reader card
- Check the log for: WordPress-STATS: 🔵 Tracked:reader_post_card_tapped

**readerPullToRefresh**
- In any Reader tab perform a pull to refresh
- Check the log for: WordPress-STATS: 🔵 Tracked:eader_pull_to_refresh

**postCardMoreTapped**

- Tap the more menu button in any reader card
- Check the log for: WordPress-STATS: 🔵 Tracked:post_card_more_tapped 

**readerDiscoverTopicTapped** 
1. Tap Discover
2. Scroll until you see a list of "Topics you might like"
3. Tap any topic
4. Check the log for: WordPress-STATS: 🔵 Tracked:reader_discover_topic_tapped

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -985,6 +985,8 @@ private void updatePosts(boolean forced) {
             }
             // make sure swipe-to-refresh progress shows since this is a manual refresh
             mRecyclerView.setRefreshing(true);
+            // Track the manual PTR
+            AnalyticsTracker.track(Stat.READER_PULL_TO_REFRESH);
```
malinajirka: I'm wondering if this method (the else branch) is invoked only when the user pulls-to-refresh -> since we check if it's forced and if we should `REQUEST_REFRESH` or `REQUEST_NEWER` I'd say it's used even in other cases. We might want to consider putting it [here](https://github.com/wordpress-mobile/WordPress-Android/blob/9e585f3cdf8d8ab0fc8964b5e188ea967fba4671/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java#L1017) into `if(forced)track`. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderPostViewHolder.kt
```diff
@@ -175,6 +183,7 @@ class ReaderPostViewHolder(
 
     private fun onMoreClicked(uiState: ReaderPostUiState, actions: List<SecondaryAction>, v: View) {
         // TODO malinjir the popup menu was reused from the legacy implementation. It needs to be refactored.
+        AnalyticsTracker.track(Stat.POST_CARD_MORE_TAPPED)
```
malinajirka, line 188: It seems we have this tracking line at two locations - here and in ReaderDiscoverViewModel. onMoreButtonClicked(..). Is it intentional?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
```diff
@@ -100,10 +100,12 @@ class ReaderPostCardActionsHandler @Inject constructor(
     }
 
     private fun handleFollowClicked(post: ReaderPost) {
+        // todo: add tracking dependent upon implementation (tracked in ReaderBlogActions)
```
malinajirka: I like to add an identifier to todos related to a certain project/person -> so we can easily check that we haven't missed anything before we release the project. I for example use "TODO malinjir" but it can be "TODO ReaderImprovements" or anything like that. Since these todos are in one of the new classes it still would be pretty easy to find, but fulltext search might be a safer option. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderPostViewHolder.kt
```diff
@@ -58,6 +62,7 @@ class ReaderPostViewHolder(
         )
         uiState.postHeaderClickData?.onPostHeaderViewClicked?.let {
             layout_post_header.setOnClickListener {
+                AnalyticsTracker.track(READER_SITE_SHOWN)
```
malinajirka: We do most of the tracking within the VM is there any reason why we want to track these two cases here (view layer)? Tbh I wouldn't mind if everything would be in the view layer but it feels a bit inconsistent now. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
```diff
@@ -174,6 +184,10 @@ class ReaderViewModel @Inject constructor(
         uiState.value?.let {
             val currentUiState = it as ContentUiState
             val position = currentUiState.readerTagList.indexOfTagName(tag.tagSlug)
+            if (AppPrefs.isReaderImprovementsPhase2Enabled() &&
+                    currentUiState.readerTagList[position].isDiscover) {
+                analyticsTrackerWrapper.track(READER_DISCOVER_SHOWN)
```
malinajirka: I haven't run the app but I think we might be tracking this event twice here and in `onTagChanged `. I believe onTagChanged is invoked even after the user clicks on the "Discover sites" button on the empty views. Or could you please elaborate why we need to track the event here?

malinajirka (commented):
Thanks @zwarm! I have reviewed the code and left some comments. Let me know what you think ;)


P.S. Is it ok to release these changes before we enable RI FF or do we want to start tracking these events when we release the project?

P.S.2: The tests are failing.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
zwarm: @malinajirka  - I did consider that, but I believe this is the only location in which a true user initiated PTR action happens. The 'force' can be something other than user initiated and that's why I shied away from wrapping it in a force check.  But of course, I can be reading this incorrectly. I'll take some time this morning to test other scenarios and then comment again.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
zwarm: @malinajirka - no problem, cut & paste scenario. I usually do leave my name. Good catch.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderPostViewHolder.kt
zwarm: @malinajirka - I dropped it here because it covers all of reader, not just a particular view model, so we don't inadvertently miss a track.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderPostViewHolder.kt
zwarm: @malinajirka - actually you are correct, it's need to be collected for all of reader. If we stick with ReaderPostViewHolder then it will cover all of reader instead of having to make sure it's in every other place that we listen for the event. thoughts?

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderViewModel.kt
zwarm: Nope, it only needs to go in onTagChanged. In 54b1cb9

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
zwarm: In commit cab4cb2

zwarm:
> P.S. Is it ok to release these changes before we enable RI FF or do we want to start tracking these events when we release the project?

I reviewed the iOS PR and looks like it went straight to develop and I didn't see any wrapping in FF. Let me check in the RI room.

 
> P.S.2: The tests are failing.
I'll take a look - probably needs a develop merge.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderPostViewHolder.kt
malinajirka: Yep I think that's correct.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
malinajirka: AFAICT the only place where we invoke `loadData(true)` is [here](https://github.com/wordpress-mobile/WordPress-Android/blob/ca78990f3b6ebe893de634a6e3af3f81497f1a7e/WordPress/src/main/java/org/wordpress/android/ui/FilteredRecyclerView.java#L214) -> and that's actually a PTR helper.

>The 'force' can be something other than user initiated

Could you please share an example? Thanks

zwarm: @malinajirka  - you are correct, I retested some other scenarios and it looks like forced == manual only. I pushed the new commit at the same time you updated your comment. In 0332235 :)

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderPostViewHolder.kt
zwarm: Removed & in commit 5c65bda

zwarm:
> 1. `selectInterestsShown` should be tracked whenever I open the screen. Eg when I go to Reader -> My Site -> Reader it should be tracked twice.

Working on it - gets tricky. May need a second set of eyes

> 2. `selectInterestsPicked` ios tracks this event only when the request succeeds. We are still having a discussion what is the correct behavior.

Sounds like this was resolved to tracking user actions - Android collects on button tap. No changes needed.

> 3. `readerDiscoverShown` the event is tracked even after a configuration change and I believe it shouldn't.

Another tricky one. May need to track whether tracking was tracked . Another place where I may need a second set of eyes.

> 4. `reader_site_shown` - should be tracking the number of times the site detail screen is opened -> no matter from which part of the app you get there. So I'd consider moving the tracking to the corresponding fragment/activity/vm.
- Not exactly sure where or what version of ReaderPostListFragment this actually is or is this another fragment all together.  Is this an instance of `ReaderPostDetailFragment` or the view that shows on the tap of the Header/Title on a readerpost (not the siteheaderview)?

> 5. `readerCardTapped` we are thinking about renaming this to `readerPostCardTapped`. Since the current name is a bit misleading.
- The name has been changed to readerPostCardTapped in commit d7a5107

malinajirka:
Let's discuss it on the call ;)

zwarm:
Follow-up
> 1. `selectInterestsShown` should be tracked whenever I open the screen. Eg when I go to Reader -> My Site -> Reader it should be tracked twice.
 
Leaving as is, so we know that the screen has actually been presented. It will be tracked 1x per shown - will not track on config changes or bottom nav taps. Staying consistent within the app. 

> 3. `readerDiscoverShown` the event is tracked even after a configuration change and I believe it shouldn't.

Moved collection point to `ReaderDiscoverViewModel.start`

> 4. `reader_site_shown` - should be tracking the number of times the site detail screen is opened -> no matter from which part of the app you get there. So I'd consider moving the tracking to the corresponding fragment/activity/vm.

Android was already using `READER_BLOG_PREVIEWED`, so iOS changed to match us & Calpyso

malinajirka (approved):
Thanks @zwarm! I've reviewed the changes and it works as described. I'm just still not sure about the "reader_discover_shown". The VM can live for days and it records the event just once even though the user accesses that screen multiple times. I'm starting to think that it might be more useful to track the event in "onResume" and ignore that we'll double track it on a configuration change or when you open eg. a post detail and you go back. Either case, we can merge this PR now and decide later.

PR #9353, opened 2019-03-01
Fix loading when site model is incomplete

planarvoid:
Fixes #9352
This PR introduces a `SiteModelProvider`, which provides an up-to-date version of the SiteModel and triggers a data refresh, when the SiteModel is updated. I've replaced the `SiteModel` that was passed around as a parameter all around the use cases with the provider. The only place where we're still retrieving the `SiteModel` is inside of the `StatsFragment` where we initialize the `SiteModelProvider` in the `start` method of the `StatsViewModel`.

To test:
* Create a self-hosted site (jurassic.ninja)
* Add the self-hosted site in the app
* Go to stats
* Install Jetpack and connect it to your account
* The stats open with the correct data (empty) after the process is finished

Issue #9352, opened 2019-03-01
Stats not loaded right after connecting Jetpack 


planarvoid:
If you try to connect a self-hosted site to Jetpack to open stats, the loading fails the first time (you have to leave the screen and go back). The reason is that after the Jetpack install is finished the SiteModel is not yet updated with the remote `siteId`. We're triggering a `fetchSite` operation and it's up to race whether it can make it or not. 

Steps to reproduce:
* Create a self-hosted site (jurassic.ninja)
* Add the self-hosted site in the app
* Go to stats
* Install Jetpack and connect it to your account
* The stats open with en error "Data not loaded" and a retry button

PR #12728, opened 2020-08-20
Stories: Only show stories for sites that support them

aforcier:
Fixes https://github.com/Automattic/stories-android/issues/417.

In addition to the check for the stories feature flag being enabled, this PR now checks that the site is able to support the stories feature. This means either:

1. A WordPress.com simple site, or
2. A Jetpack-connected site with version 8.8 or higher

8.8 is the first Jetpack version to include the story block. The block is flagged as a beta block, which requires some extra steps to actually enable it on the site. When we're ready to release the stories feature, we'll want to update the minimum version set in this PR - this would be the sequence of events:

1. The story block in Jetpack is switched from beta to production
2. A new Jetpack version is released with the change (probably Jetpack 9.0)
3. We update the minimum version set in this PR to 9.0
4. The next WPAndroid release after step 3 is the first one we could use for a public beta

(This is being tracked on the stories project board.)

I also fixed the description toasts shown when long-pressing the FABs in the main activity and the post list. In the main activity we weren't accounting for stories, and in the post list no toast was being shown at all when stories were enabled.

Also, the FAB tooltip in the post list was always mentioning stories whether the feature was available or not - I fixed this as well.

Only one reviewer is really needed, ccing you @develric in case I missed anything important since I had to restructure `WPMainActivityViewModel` a bit.

### To test:
There's no need to test actually pushing a story to the site in this PR - a follow-up PR will remove the gallery fallback logic and call for this kind of test. Any site that isn't Tug's test site will just upload a gallery still as of this PR.

Instead, check that, for each site type:
- The options presented in the main activity and post list bottom sheets make sense
- The toasts shown when long pressing the FABs are consistent with the FAB options
- The tooltips shown over the FABs in both screens are consistent with the FAB options

These are the site type cases.

1. Simple WP.com site -> Story option shown ✅
2. Jetpack-connected site, version 8.8+ -> Story option shown ✅
3. Jetpack-connected site, version < 8.8 -> Story option not shown 🙅‍♂️
4. Self-hosted site -> Story option not shown 🙅‍♂️
5. (Optional) Atomic on WordPress.com site -> Story option shown ✅ (this is essentially the same as case 2, Atomic sites for this purpose are Jetpack-connected sites kept on the latest stable Jetpack version)

There's another degree of freedom in the main activity, which is whether the user has the ability to edit pages on the site. If not, there's no bottom sheet in the main activity if stories are disabled or unsupported.

(This is what the unit tests are mostly checking.)

Also check that the option is never shown if the feature flag is disabled, by e.g. [setting this to false](https://github.com/wordpress-mobile/WordPress-Android/blob/a3a756330e9cbf71dc75048563ed958a02bffda4/WordPress/build.gradle#L77).

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
+++ b/WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
@@ -27,6 +27,8 @@ public class SiteUtils {
     public static final String GB_EDITOR_NAME = "gutenberg";
     public static final String AZTEC_EDITOR_NAME = "aztec";
     public static final String WP_STORIES_CREATOR_NAME = "wp_stories_creator";
+    // TODO Update to the first version with the story block as a production and not a beta block
+    public static final String WP_STORIES_JETPACK_VERSION = "8.8";
     private static final int GB_ROLLOUT_PERCENTAGE_PHASE_1 = 100;
     private static final int GB_ROLLOUT_PERCENTAGE_PHASE_2 = 100;
 
@@ -285,6 +287,10 @@ public class SiteUtils {
             }
         }
         return false;
+    }
+
+    public static boolean supportsStoriesFeature(SiteModel site) {
+        return site != null && (site.isWPCom() || checkMinimalJetpackVersion(site, WP_STORIES_JETPACK_VERSION));
     }
 
     public static boolean isNonAtomicBusinessPlanSite(@Nullable SiteModel site) {
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
@@ -484,14 +484,14 @@ public class WPMainActivity extends LocaleAwareActivity implements
                     getMySiteFragment().requestNextStepOfActiveQuickStartTask(false);
                 }
             }
-            mViewModel.onFabClicked(hasFullAccessToContent(), shouldShowPublishPostQuickStartTask);
+            mViewModel.onFabClicked(mSelectedSite, shouldShowPublishPostQuickStartTask);
         });
 
         mFloatingActionButton.setOnLongClickListener(v -> {
             if (v.isHapticFeedbackEnabled()) {
                 v.performHapticFeedback(HapticFeedbackConstants.LONG_PRESS);
             }
-            mViewModel.onFabLongPressed(hasFullAccessToContent());
+            mViewModel.onFabLongPressed(mSelectedSite);
 
             int messageId = hasFullAccessToContent()
                     ? R.string.create_post_page_fab_tooltip
@@ -504,7 +504,7 @@ public class WPMainActivity extends LocaleAwareActivity implements
         ViewUtilsKt.redirectContextClickToLongPressListener(mFloatingActionButton);
 
         mFabTooltip.setOnClickListener(v -> {
-            mViewModel.onTooltipTapped(hasFullAccessToContent());
+            mViewModel.onTooltipTapped(mSelectedSite);
         });
 
         mViewModel.isBottomSheetShowing().observe(this, event -> {
@@ -560,8 +560,7 @@ public class WPMainActivity extends LocaleAwareActivity implements
 
         mViewModel.start(
                 mSiteStore.hasSite() && mBottomNav.getCurrentSelectedPage() == PageType.MY_SITE,
-                hasFullAccessToContent()
-        );
+                mSelectedSite);
 
         mMLPViewModel.init(DisplayUtils.isLandscape(this));
     }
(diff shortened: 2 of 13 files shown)


WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
```diff
@@ -287,6 +289,10 @@ public static boolean checkMinimalJetpackVersion(SiteModel site, String limitVer
         return false;
     }
 
+    public static boolean supportsStoriesFeature(SiteModel site) {
+        return site != null && (site.isWPCom() || checkMinimalJetpackVersion(site, WP_STORIES_JETPACK_VERSION));
```
mzorz, line 293: TIL about `checkMinimalJetpackVersion()` and `Version()` implementation

mzorz:
**tl;dr** I think we can keep things as they are, but I had to explore things deeply in order to understand whether the changes here (or even the code as it was in some cases) are in the right direction in terms of code design. I have thoughts about it but aligning to these thoughts most probably means a bigger refactor that exceeds the purpose of this PR. Sharing my thoughts below, feel free to save for later.

------ 

After a couple of iterations approaching changes to these classes involved in this PR I felt I had a better understanding of how things work here, so I was wondering whether passing a SiteModel to a ViewModel was technically advised, since when the site changes we need to refresh it somehow and make sure the UI changes are triggered, propagating / representing the changes to the ViewModel on the View.
The trigger for these thoughts was that I couldn't find examples in the codebase that were using a SiteModel being passed to a ViewModel, but found plenty of examples of SiteStore being injected in ViewModels, and then obtaining the SiteModel from the Store via its localId. This is probably a better way in which the VM knows about the Repository, and the Repository is responsible for entity models (i.e. in this case a SiteModel).

Some examples:
- DomainRegistrationDetailsViewModel: injects SiteStore, gets the actual model from re-obtaining it from the store, via a passed SiteModel's id in start()
- StorePostViewModel -> same
- HomePageSettingsViewModel -> same
- StatsSiteSelectionViewModel -> same
- ReaderViewModel -> injects AccountStore
etc.

Then I realized we're never keeping the selected site on `WPMainActivityViewModel`, but on `WPMainActivity` (which is weird, ideally the selectedSite would be kept in a Repository and the ViewModel would be able to interact with it as per MVVM architecture). So this is beyond the scope of this PR but it feels what we should really be doing here is probably modifying things to keep the state in the VM, instead of waiting for the user to click on the FAB to update the state  and then reflect it on the screen.

[StatsViewAllViewModel](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/StatsViewAllViewModel.kt) has a good example of this in which it gets a [StatsSiteProvider](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/utils/StatsSiteProvider.kt) injected in its constructor parameters, which in turn encapsulates the interaction in a Repository (in particular the very simple SelectedSiteStorage), and the Provider gets the updates from registering to FluxC `OnSiteChanged` events itself.
See https://github.com/wordpress-mobile/WordPress-Android/pull/9353 and the related issue description (https://github.com/wordpress-mobile/WordPress-Android/issues/9352) for a bit more reasoning on why passing the SiteModel may be not ideal - it was probably identified as an edge case but it makes things more robust.
(another ViewModel using that is `InsightsManagementViewModel` for example).

If we went down that route, I think not only we would get rid of passing SiteModel, but also would benefit of not having to pass booleans by encapsulating the selectedSite capabilities check logic in the ViewModel, and just rely on the VM and the Activity doing their business as usual given by triggering refreshes though LiveData observers. 

Open to discussing thoughts and learning from your points of view, but to make it clear I wouldn't block this PR with this discussion 😄

aforcier:
> So this is beyond the scope of this PR but it feels what we should really be doing here is probably modifying things to keep the state in the VM, instead of waiting for the user to click on the FAB to update the state and then reflect it on the screen.

I agree completely, I wasn't a big fan of my solution but it felt like an okay interim solution vs adding another boolean to pass to every method. In most other places I'd have gone for injecting the `SiteStore` as well, but as you outlined the site would have to be updated whenever there's a change, and the selected site really should live in the ViewModel, requiring a bigger refactor and a lot more care since this is a central part of the app and selectedSite logic has given us plenty of trouble in the past. I think that's definitely worth doing but out of scope for this PR.

aforcier:
> 3. Jetpack-connected site, version < 8.8 -> Story option not shown 🙅‍♂️ couldn't test this one but did step by step debugging and observed the checks for minimum version run correctly

👍 Just for future use, I forgot to mention that an easy way to test earlier Jetpack versions is to install the Jetpack beta plugin, that allows you to switch to earlier releases.

mzorz:
> I agree completely, I wasn't a big fan of my solution but it felt like an okay interim solution vs adding another boolean to pass to every method. In most other places I'd have gone for injecting the SiteStore as well, but as you outlined the site would have to be updated whenever there's a change, and the selected site really should live in the ViewModel, requiring a bigger refactor and a lot more care since this is a central part of the app and selectedSite logic has given us plenty of trouble in the past. I think that's definitely worth doing but out of scope for this PR.

That makes sense - let's merge this then, since it works as expected 😎

PR #12731, opened 2020-08-20
Revert "WP Stories: build the VideoPress URL if mimeType is `video/videopress`"

mzorz:
Reverts wordpress-mobile/WordPress-Android#12713

As per https://github.com/wordpress-mobile/WordPress-Android/pull/12713#issuecomment-676858706, and confirmed by @Tug elsewhere, we can revert this change now as URLs are being obtained on the web side from the media id 👍


aforcier (approved):
Double-checked that it's working, all good 👍
:shipit:

PR #12740, opened 2020-08-21
Convert SnackbarMessageHolder message and button title to UIString 

ashiagr:
This PR converts `SnackbarMessageHolder`'s params for message and button title to `UIString`. Earlier it accepted only string res id, with this change it now accepts both string res id as well as text.   

**To test**

That the project compiles and tests run successfully using the command
`./gradlew :WordPress:testVanillaDebugUnitTest`

**Open Question**

Should we change param type from `String` to `CharSequence` [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/utils/UiString.kt#L10) so that we can send html text with markup? I'll need it in the FollowUseCase [here](https://github.com/wordpress-mobile/WordPress-Android/blob/ea7b0115e2d36f498a3004be7e2f65126f8688ee/WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderPostFollowUseCase.kt#L101-L108).

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
@@ -679,10 +679,12 @@ public class EditPostActivity extends LocaleAwareActivity implements
         });
         mEditorMedia.getSnackBarMessage().observe(this, event -> {
             SnackbarMessageHolder messageHolder = event.getContentIfNotHandled();
-            if (messageHolder != null && messageHolder.getMessage() instanceof UiStringRes) {
-                UiStringRes message = (UiStringRes) messageHolder.getMessage();
-                WPSnackbar
-                        .make(findViewById(R.id.editor_activity), message.getStringRes(), Snackbar.LENGTH_SHORT)
+            if (messageHolder != null) {
+                WPSnackbar.make(
+                                findViewById(R.id.editor_activity),
+                                mUiHelpers.getTextOfUiString(this, messageHolder.getMessage()),
+                                Snackbar.LENGTH_SHORT
+                        )
                         .show();
             }
         });
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
@@ -161,6 +161,7 @@ import org.wordpress.android.ui.uploads.UploadUtilsWrapper;
 import org.wordpress.android.ui.uploads.VideoOptimizer;
 import org.wordpress.android.ui.utils.AuthenticationUtils;
 import org.wordpress.android.ui.utils.UiHelpers;
+import org.wordpress.android.ui.utils.UiString.UiStringRes;
 import org.wordpress.android.util.ActivityUtils;
 import org.wordpress.android.util.AniUtils;
 import org.wordpress.android.util.AppLog;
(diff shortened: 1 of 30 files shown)


WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -678,9 +679,10 @@ private void startObserving() {
         });
         mEditorMedia.getSnackBarMessage().observe(this, event -> {
             SnackbarMessageHolder messageHolder = event.getContentIfNotHandled();
-            if (messageHolder != null) {
+            if (messageHolder != null && messageHolder.getMessage() instanceof UiStringRes) {
```
malinajirka, line 682: Is there any reason why we can accept only UiStringRes and not UiStringText? Have we considered using `mUiHelpers.getTextOfUiString(this, messageHolder.getMessage())`. ?

WordPress/src/main/java/org/wordpress/android/ui/posts/PostsListActivity.kt
```diff
@@ -408,12 +408,12 @@ class PostsListActivity : LocaleAwareActivity(),
                     SnackbarItem(
                             SnackbarItem.Info(
                                 view = parent,
-                                textRes = UiStringRes(holder.messageRes),
+                                textRes = holder.message as UiStringRes,
```
malinajirka, line 411: I believe this manual cast is unnecessary -> moreover, it can result in an exception. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/PostsListActivity.kt
```diff
@@ -408,12 +408,12 @@ class PostsListActivity : LocaleAwareActivity(),
...
                                 duration = Snackbar.LENGTH_LONG
                             ),
-                            holder.buttonTitleRes?.let {
+                            holder.buttonTitle?.let {
                                 SnackbarItem.Action(
-                                    textRes = UiStringRes(holder.buttonTitleRes),
+                                    textRes = holder.buttonTitle as UiStringRes,
```
malinajirka, line 416: I believe this manual cast is unnecessary -> moreover, it can result in an exception. Wdyt?

malinajirka (commented):
Thanks so much @ashiagr! The changes look great - I found just two potential issues. Let me know what you think ;-)!

WordPress/src/main/java/org/wordpress/android/ui/posts/PostsListActivity.kt
ashiagr: Removed here a5e22f7

WordPress/src/main/java/org/wordpress/android/ui/posts/PostsListActivity.kt
ashiagr: Removed here a5e22f7

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
ashiagr: Fixed in 2dcad11

ashiagr:
Great finds, thank you! I've fixed both issues. 

I added a question in the description. Anything on that?

malinajirka (approved):
Thanks @ashiagr! It all LGTM now :).

Regarding the question - changing it sounds like a good idea. Feel free to change it either in this PR or in the usecase PR. Whichever you prefer. Also feel free to merge this PR ;).

Note: We might want to merge it after the freeze. I'm not sure.

ashiagr:
@malinajirka, I converted `uiString.message` and `uiHelpers.getTextOfUiString` from `String` to `CharSequence` here: https://github.com/wordpress-mobile/WordPress-Android/commit/f6620cff58ac273d731911e68115c03b75e82c35
Had to convert it back to `String` at few places. Let me know if we should merge these changes.

malinajirka:
I believe we can merge it. We can think about other possible solutions if we want to use html within the UiString/SnackBarMessageHolder. Thanks for working on this 🙇

PR #12748, opened 2020-08-21
Discover tab - error and progress handling

malinajirka:
Parent issue #12028

Implements error and progress states on the new Discover tab.

Progress
- if the content is empty, shows fullscreen progress screen
- if the content is not empty and is fetching the first page, shows the Pull-to-refresh progress
- if the content is not empty and is fetching more items, shows progress indicator at the bottom

Error
- if the full screen progress is displayed and an error occurs, show fullscreen error with retry
- if the content is displayed and an error occurs show an error snackbar message

To test:
1. Enable RI FF
2. Clear app data - you might want to slow down your connection on an emulator
3. Open Discover and notice "fullscreen progress is shown
4. Wait for the content to load
5. Pull to refresh and notice "PTR" progress indicator is shown
6. Make sure it disappears when the request finishes
6. Scroll to the bottom and notice "load more progress indicator" is shown
7. Make sure it disappears when the request finishes
8. Scroll to the bottom again and verify the progress indicator still works as expected

-----------------------
1. Enable RI FF
2. Clear app data - you might want to slow down your connection on an emulator
3. Open Discover, wait until the data for discover start loading (not the data for interest screen) and turn on airplane mode
4. Notice "fullscreen discover error is shown"
5. Turn off airplane mode and click on retry
6. Data are loaded
7. Pull to refresh and turn on airplane mode
8. Notice a snackbar an error message is shown and the PTR progress indicator disappears
9. Scroll to the bottom and notice an error message is shown and the load more progress indicator disappears

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
```diff
@@ -80,40 +86,80 @@ class ReaderDiscoverViewModel @Inject constructor(
...
-                                    onVideoOverlayClicked = this::onVideoOverlayClicked,
-                                    onPostHeaderViewClicked = this::onPostHeaderClicked,
-                                    onTagItemClicked = this::onTagItemClicked,
-                                    postListType = TAG_FOLLOWED
-                            )
-                            is InterestsYouMayLikeCard -> {
-                                postUiStateBuilder.mapTagListToReaderInterestUiState(
-                                        it.interests,
-                                        this::onReaderTagClicked
+            if (posts != null && posts.cards.isNotEmpty()) {
```
malinajirka, line 97: I've added this `isNotEmpty` check since we don't want to show an empty "ContentUiState" when there are no data in the local db yet.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderRepositoryEvents.kt
```diff
@@ -32,3 +34,12 @@ sealed class ReaderRepositoryCommunication {
         return "${this.javaClass.simpleName})"
     }
 }
+
+sealed class ReaderDiscoverCommunication {
```
malinajirka, line 36: I'm not sure if we want to introduce a new class for discover. I didn't want to block this PR on this issue. `ReaderRepositoryCommunication` instances don't know to what action they are related. Only the `Failure` has some information about the action, but it has a different issue (we can create invalid states eg `Failure(PostLikeEnded)`). I decided to create a copy for now and discuss this more later. Wdyt?

WordPress/src/main/res/layout/reader_fullscreen_error_with_retry.xml
```diff
@@ -0,0 +1,39 @@
...
+        android:layout_marginEnd="@dimen/margin_extra_extra_medium_large"
+        android:layout_marginStart="@dimen/margin_extra_extra_medium_large"
+        android:layout_marginTop="@dimen/margin_extra_large"
+        android:text="@string/retry"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@id/error_title" />
+
+    <org.wordpress.android.widgets.WPTextView
```
zwarm: The new stuff we've been doing has used `com.google.android.material.textview.MaterialTextView` instead of WPTextView + we've been moving colors and appearance to `styles`.  Not 100% necc, but may be a good idea to be consistent. Thoughts?

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderRepositoryEvents.kt
zwarm: @malinajirka - this is an interesting thought. imho, I like the idea of discover tasks because it's clear, SRP and allows us to react start to finish for a task that we care about. `Failure` has been removed, so that's no longer an option for the very reason you state. Let's leave it for this PR and then chat later as you suggested.

zwarm (approved):
@malinajirka Nice work. All tests pass as stated. It took some fancy clicking to get airplane mode on even with slower connection, but I did see the full screen error page. I also saw the message across the bottom showing no connection. 👍 

I left one comment on xml, your choice to change it or leave as is. If you decide to leave it, then you are all set to go ahead and merge. :)

WordPress/src/main/res/layout/reader_fullscreen_error_with_retry.xml
malinajirka: >The new stuff we've been doing has used com.google.android.material.textview.MaterialTextView instead of WPTextView 

Good catch! 

> we've been moving colors and appearance to styles. Not 100% necc, but may be a good idea to be consistent. Thoughts?

:P Tbh I'm not a fan of using styles unless the styles are used at more than +- 3 places. Imo it just makes the code less readable. I'm not saying styles shouldn't be used at all, but with our approach to design - each designer has a bit different style and the sections of the app aren't consistent, I don't see many benefits.
Having said that I exported the styles in reader_fullscreen_error_with_retry to reader_styles.xml for consistency. We can discuss this further on slack - perhaps there are some advantages I'm not seeing.

Both fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/12748/commits/695a3daf95d6319f16a7b0a61b93f14150e54eda

malinajirka:
Thanks for the review @zwarm 🙇 ! It's ready for another round :).

WordPress/src/main/res/layout/reader_fullscreen_error_with_retry.xml
zwarm: I see your point about more than 3 or more places. Sounds like we may need a discussion on what consistency means - at the very least we will be consistent within the reader section. Thanks.

PR #12777, opened 2020-08-24
RI: Implement "Turn on/off notifs" action on Reader discover tab

ashiagr:
Parent Issue: #12028
 
Adds "Turn on/ off site notifications" action to the items on the discover tab in Reader.

Logic is extracted from `ReaderPostListFragment` to `ReaderSiteNotificationsUseCase` with below fixes:
a) notification status is updated in the `tbl_blog_info` table only on successful subscription update.
b) network error snackbar is displayed if no network is available.

~~**Open Question**~~
~~- Discover feed is not yet updated on updates to `tbl_blog_info`. Should we post `ReaderBlogInfoTableActionEnded.INSTANCE` similar to `ReaderPostTableActionEnded` event?~~ - More menu can be refreshed with the changes in PR #12781 so no need to update the feed.

**To test**

I'd suggest just review the code and run unit tests for now. Turn on/off notifs depend on whether site is followed (it adds an entry to `tbl_blog_info` which is required to update notification status) and FollowUseCase is not merged yet.

We can test it once again when FollowUseCase is finalised.

Notes:

1. Haven't yet used this usecase in the legacy adapter as we update notification status after subscription is updated in this new usecase as compared to optimistic update in the legacy adapter. Plus legacy adapter is buggy - its doesn't care to revert status if request fails. Let me know if I should switch to new usecase there.

2. In case we want to do both optimistic update + update on request completion, we might want to reconsider this design.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderSiteNotificationsUseCase.kt
```diff
@@ -0,0 +1,105 @@
...
+        if (event.isError) {
+            continuation?.resume(false)
+            AppLog.e(
+                    API,
+                    ReaderSiteNotificationsUseCase::class.java.simpleName + ".onSubscriptionUpdated: " +
+                            event.error.type + " - " + event.error.message
+            )
+        } else {
+            continuation?.resume(true)
+            dispatcher.dispatch(AccountActionBuilder.newFetchSubscriptionsAction())
```
malinajirka: NitPick: We might want to consider moving this call to `toggleNotification(..)` (line 56) so the whole flow is within that method. Wdyt?

malinajirka (approved):
Thanks @ashiagr! LGTM.

Regarding your notes:
1. If we can, I'd like to re-use the same use-case even when we slightly change the behavior. Having said that, I wouldn't worry about it too much unless it's a straightforward task.
2. I think this solution is fine.

WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderSiteNotificationsUseCase.kt
ashiagr: Moved here d7403d4, test updated in 44da620

ashiagr:
Thanks for the review!

Regarding
> If we can, I'd like to re-use the same use-case even when we slightly change the behavior. 

Re-used usecase in legacy adapter here: c824466

Also moved `SnackbarMessageHolder` UI element from the useCase to the handler + used sealed class to return status to the handler similar to block blog useCase. (0e209b0, tests updated in 2097fbf)

WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderSiteNotificationsUseCase.kt
```diff
@@ -0,0 +1,113 @@
...
+            updateSubscription(blogId)
+        }
+
+        if (succeeded) {
+            updateBlogInDb(blogId)
+            dispatcher.dispatch(AccountActionBuilder.newFetchSubscriptionsAction())
+        } else {
+            return RequestFailed
+        }
+        return null
```
malinajirka: We might want to consider creating states we don't need to re-act to instead of returning null. When we return null for these states, the UseCase technically plays part in how and to what we'll react and we are trying to remove this "responsibility" from the UseCase. Moreover, I believe avoiding nulls wherever possible is always a win. Wdyt?

malinajirka (approved):
Great job @ashiagr! I've left one nitpick comment (feel free to ignore it) - it looks great overall! Thanks for all the changes ;)!

WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderSiteNotificationsUseCase.kt
ashiagr: Removed null -> 4a6aee5, added test case -> 18ec384

PR #12788, opened 2020-08-25
AppBar refresh - Me, Editor, Site Creation, Media Picker and Site Preview

khaykov:
This PR updates AppBar design and adds lift on scroll functionality to the following screens (and related nested screens)

- Me screen and all the nested screens (excluding Zendesk).
- Editor screens - Post Settings and Revisions.
- Site Creation flow (only the first step)
- Media Picker.
- Site Preview.

@osullivanchris Unfortunately I was not able to get a lift on scroll working in editor :( Any ideas on what we can do there instead?

To test:

For all screens:

- Make sure nothing looks out of place (including ActionMode) :)
- Make sure the toolbar state is restored on configuration change.

Test 1:

- Navigate to the Me screen and confirm that lift on scroll works (you might need to check the screen in landscape orientation).
- Check My Profile,  Account Settings, App Settings, and Help screens and confirm that lift on scroll works.

Test 2:

- Navigate to the Editor screen and check Post Settings and History + History Details screens to confirm that lift on scroll works.

Test 3:

- Start Site Creation flow and confirm that lift on scroll works on the first step.

Test 4:

- Open media picker (from the editor or by tapping on Blavatar on My Site screen) and confirm that lift on scroll works.

Test 4:

- Open Site Preview and confirm that lift on scroll works.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/AndroidManifest.xml
+++ b/WordPress/src/main/AndroidManifest.xml
@@ -211,7 +211,7 @@
         <!-- Posts activities -->
         <activity
             android:name=".ui.posts.EditPostActivity"
-            android:theme="@style/WordPress.Editor.NoActionBar"
+            android:theme="@style/WordPress.NoActionBar"
             android:windowSoftInputMode="stateHidden|adjustResize">
             <meta-data
                 android:name="android.support.PARENT_ACTIVITY"
@@ -266,7 +266,7 @@
         <activity
             android:name=".ui.history.HistoryDetailActivity"
             android:label="@string/history_detail_title"
-            android:theme="@style/WordPress.Editor.NoActionBar" />
+            android:theme="@style/WordPress.NoActionBar" />
 
         <!-- plans -->
         <activity
--- a/WordPress/src/main/java/org/wordpress/android/ui/FullScreenDialogFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/FullScreenDialogFragment.java
@@ -30,6 +30,8 @@ import androidx.fragment.app.FragmentActivity;
 import androidx.fragment.app.FragmentManager;
 import androidx.fragment.app.FragmentTransaction;
 
+import com.google.android.material.appbar.AppBarLayout;
+
 import org.wordpress.android.R;
 import org.wordpress.android.util.ColorUtils;
 import org.wordpress.android.util.ContextExtensionsKt;
@@ -50,11 +52,13 @@ public class FullScreenDialogFragment extends DialogFragment {
     private String mTitle;
     private Toolbar mToolbar;
     private boolean mHideActivityBar;
+    private boolean mIsLiftOnScroll;
     private int mToolbarTheme;
     private int mToolbarColor;
 
     private static final String ARG_ACTION = "ARG_ACTION";
-    private static final String ARG_HIDE_ACTIVITY_BAR = "ARG_HIDE_ACTIVITY_BAR";
+    private static final String ARG_HIDE_ACTIVITY_BAR = "ARG_IS_LIFT_ON_SCROLL";
+    private static final String ARG_IS_LIFT_ON_SCROLL = "ARG_LIFT_ON_SCROLL";
     private static final String ARG_SUBTITLE = "ARG_SUBTITLE";
     private static final String ARG_TITLE = "ARG_TITLE";
     private static final String ARG_TOOLBAR_THEME = "ARG_TOOLBAR_THEME";
@@ -94,6 +98,7 @@ public class FullScreenDialogFragment extends DialogFragment {
         dialog.setOnConfirmListener(builder.mOnConfirmListener);
         dialog.setOnDismissListener(builder.mOnDismissListener);
         dialog.setHideActivityBar(builder.mHideActivityBar);
+        dialog.setLiftOnScroll(builder.mIsLiftOnScroll);
         return dialog;
     }
 
@@ -105,6 +110,7 @@ public class FullScreenDialogFragment extends DialogFragment {
         bundle.putInt(ARG_TOOLBAR_THEME, builder.mToolbarTheme);
         bundle.putInt(ARG_TOOLBAR_COLOR, builder.mToolbarColor);
         bundle.putBoolean(ARG_HIDE_ACTIVITY_BAR, builder.mHideActivityBar);
+        bundle.putBoolean(ARG_IS_LIFT_ON_SCROLL, builder.mIsLiftOnScroll);
         return bundle;
     }
 
@@ -259,6 +265,7 @@ public class FullScreenDialogFragment extends DialogFragment {
         mToolbarTheme = bundle.getInt(ARG_TOOLBAR_THEME);
         mToolbarColor = bundle.getInt(ARG_TOOLBAR_COLOR);
         mHideActivityBar = bundle.getBoolean(ARG_HIDE_ACTIVITY_BAR);
+        mIsLiftOnScroll = bundle.getBoolean(ARG_IS_LIFT_ON_SCROLL);
     }
 
     /**
@@ -284,6 +291,12 @@ public class FullScreenDialogFragment extends DialogFragment {
 
         if (mToolbarColor > 0) {
             mToolbar.setBackgroundColor(getResources().getColor(mToolbarColor));
+        }
+
+        AppBarLayout appBarLayout = view.findViewById(R.id.appbar_main);
+        appBarLayout.setLiftOnScroll(mIsLiftOnScroll);
+        if (!mIsLiftOnScroll) {
+            appBarLayout.setLifted(false);
         }
 
         if (!mAction.isEmpty()) {
@@ -341,6 +354,15 @@ public class FullScreenDialogFragment extends DialogFragment {
      */
     public void setHideActivityBar(boolean hide) {
         this.mHideActivityBar = hide;
+    }
+
+    /**
+     * Set flag to enable or disable AppBar's lift on scroll.
+     *
+     * @param isLiftOnScroll boolean to toggle lift on scroll
+     */
+    public void setLiftOnScroll(boolean isLiftOnScroll) {
+        this.mIsLiftOnScroll = isLiftOnScroll;
     }
 
     /**
@@ -429,6 +451,7 @@ public class FullScreenDialogFragment extends DialogFragment {
         String mSubtitle = "";
         String mTitle = "";
         boolean mHideActivityBar = false;
+        boolean mIsLiftOnScroll = true;
         int mToolbarTheme = 0;
         int mToolbarColor = 0;
 
@@ -580,6 +603,17 @@ public class FullScreenDialogFragment extends DialogFragment {
          */
         public Builder setOnDismissListener(@Nullable OnDismissListener listener) {
             this.mOnDismissListener = listener;
+            return this;
+        }
+
+        /**
+         * Set flag to enable or disable AppBar's lift on scroll.
+         *
+         * @param isLifOnScroll boolean to toggle lift on scroll
+         * @return {@link Builder} object to allow for chaining of calls to set methods
+         */
+        public Builder setIsLifOnScroll(Boolean isLifOnScroll) {
+            this.mIsLiftOnScroll = isLifOnScroll;
             return this;
         }
     }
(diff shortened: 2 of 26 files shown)


osullivanchris (changes requested):
Nice progress again! A few things I spotted

- On `blog posts` and `media` screens, can we have a divider on the tabs on default load the same as we did on Reader? I think this can be a generic thing that we do with top tabs, as the structure of the tabs is lost without some separating element on initial load
- When I am on `Media` screen, I don't see a shadow on scroll. When I'm on `Blog Posts`, sometimes there is a shadow as I scroll and sometimes there isn't. 
- On dark mode, in the `Blog Posts` screen, we have that fading out effect for the top tabs. However, this has a color which is defined by the elevated state. On initial load, if you have not scrolled vertically, but scroll the tabs horizontally, you can see a grey color there which looks wrong. Tricky to get this one right. We'd either need to have a different fade out color, and display the right one at the right time, which might not be foolproof. Or we could drop the fade out effect for something simpler. 
- Site Preview looks so much better with a white header which feels neutral. I wonder if we could go one step further here and hide the header on scroll, which would give you an even better idea of how your site actually looks?

> @osullivanchris Unfortunately I was not able to get a lift on scroll working in editor :( Any ideas on what we can do there instead?

I think we need the separation. So we could
- default to treating the header as elevated. So it always has shadow in light mode, always lighter color in dark mode. *or*
- add a divider as we have done when we're using top tabs. But I think the divider might not look as nice on scroll as the shadow. 

What do you think out of those two options @khaykov ? Any opinions @iamthomasbishop ? (let me know if you need context)

Also when the editor is loading, there is a grey background, which then changes to white. Perhaps the loading state should also be white cc @iamthomasbishop

khaykov:
@osullivanchris 

> On blog posts and media screens, can we have a divider on the tabs on default load the same as we did on Reader? I think this can be a generic thing that we do with top tabs, as the structure of the tabs is lost without some separating element on initial load
> When I am on Media screen, I don't see a shadow on scroll. When I'm on Blog Posts, sometimes there is a shadow as I scroll and sometimes there isn't.
> On dark mode, in the Blog Posts screen, we have that fading out effect for the top tabs. However, this has a color which is defined by the elevated state. On initial load, if you have not scrolled vertically, but scroll the tabs horizontally, you can see a grey color there which looks wrong. Tricky to get this one right. We'd either need to have a different fade out color, and display the right one at the right time, which might not be foolproof. Or we could drop the fade out effect for something simpler.

This PR does not touch the screens you mentioned, you can check changes to them in #12675 :)

> Site Preview looks so much better with a white header which feels neutral. I wonder if we could go one step further here and hide the header on scroll, which would give you an even better idea of how your site actually looks?

Sure thing 👍 Looks like this:
[Image from Gyazo](https://gyazo.com/85e56a20a5882ca2a0f1da2b9861a27f)

[Image from Gyazo](https://gyazo.com/47dcaa2877216df755722544cf1d65dd)

> I think we need the separation. So we could
>
> - default to treating the header as elevated. So it always has shadow in light mode, always lighter color in dark mode. or
> - add a divider as we have done when we're using top tabs. But I think the divider might not look as nice on scroll as the shadow.
> What do you think out of those two options @khaykov ?

I like the first options. Should we also make always elevated on related screens (post settings and history)?

osullivanchris:
> This PR does not touch the screens you mentioned, you can check changes to them in #12675 :)

oh my bad!

> Sure thing 👍 Looks like this:

This looks good! Is there other places in the app where we show a preview of the site like this that we could do the same? I'm thinking there's a way to do this from reader at least. cc @mbshakti 

> I like the first options. Should we also make always elevated on related screens (post settings and history)?

Yeah I agree, always elevated seems safe. Hmm on the other related screens...no I don't think its necessary to change. I'd rather we just have one exception for the editor, which might adapt later as the editor evolves. If something changes there later, we don't need to think about post settings and history, as they'll just have the default behaviour from the rest of the app

iamthomasbishop:
> Any opinions @iamthomasbishop ?

@osullivanchris @khaykov Thanks for the ping! I'm not caught up on all of the context, but I think (hope?) enough to provide some coherent feedback 😆 

I was taking the latest build for a spin and naturally the first place I looked was the editor 😄 It feels _way_ better with a white app bar, but across the board we definitely need some separation on the app bar while in "resting" state. My suggestion would be to apply a 1dp gray divider on the "resting" state and then on scroll to apply the standard material elevation shadow. This is the approach that has felt most natural on other Android apps that I've seen.

## Editor/Preview

The above approach would also help with another concern I have, which is that the preview app bar doesn't have enough separation to distinguish between UI and the contents of the site/page itself. It jumped out at me right away because the demo site I was using has a white background and I think we might want to add a divider there just to make the distinction slightly more obvious. Here's a screenshot:

<a href="https://user-images.githubusercontent.com/1200432/91355552-ed9ef000-e7b3-11ea-981b-5fcd64205ac7.png"></a>

If it's not a pain, can I make a special request? 😀 Can we apply the same app bar behavior in the editor that is used on the preview screen in this build? This has always been my wish for the editor, but it hasn't been high enough on the priority list to be tackled yet.

A couple of other minor issues I noticed in the editor/preview:

* I'm not seeing the primary action (usually `Publish`, `Update`, etc) in the app bar -- any chance we can add that back? 
* On preview, it seems like the footer toolbar is collapsed by default -- shouldn't that be expanded by default and hidden on scroll, same as the top app bar?

Side note: there is another discussion going on about removing the site title from the app bar, so that will be cleaned up a bunch.

## General feedback

More generally, I know it's on the radar already, but changing the status bar to the same color as the app bar will be super helpful 👌

khaykov:
@osullivanchris 

> This looks good! Is there other places in the app where we show a preview of the site like this that we could do the same? I'm thinking there's a way to do this from reader at least.

All the places where we show preview will utilize this screen and this behavior :)

> Yeah I agree, always elevated seems safe. Hmm on the other related screens...no I don't think its necessary to change. I'd rather we just have one exception for the editor, which might adapt later as the editor evolves. If something changes there later, we don't need to think about post settings and history, as they'll just have the default behaviour from the rest of the app

Go it 👍 

@iamthomasbishop 

> The above approach would also help with another concern I have, which is that the preview app bar doesn't have enough separation to distinguish between UI and the contents of the site/page itself. It jumped out at me right away because the demo site I was using has a white background and I think we might want to add a divider there just to make the distinction slightly more obvious. 

I will direct this to @osullivanchris :)

> My suggestion would be to apply a 1dp gray divider on the "resting" state and then on scroll to apply the standard material elevation shadow. This is the approach that has felt most natural on other Android apps that I've seen.

I will look into it a bit more, but for now, the problem is that since the editor is a pretty unique view, we can't easily detect its exact scrolling state. So we can either have a shadow or a divider :)

> I'm not seeing the primary action (usually Publish, Update, etc) in the app bar -- any chance we can add that back?

Good catch! They are still there, but their text color is white :)

> On preview, it seems like the footer toolbar is collapsed by default -- shouldn't that be expanded by default and hidden on scroll, same as the top app bar?

This should not happen!

iamthomasbishop:
> since the editor is a pretty unique view, we can't easily detect its exact scrolling state. So we can either have a shadow or a divider

Ahh, okay, I should have read the back scroll more thoroughly 🤦‍♂️ Either one is fine by me, as long as there is visual separation -- although my "ideal" remains the divider <-> shadow approach, if we can figure out a way to do that (and of course if @osullivanchris agrees, bc it's his call in the end) 😄 

> Good catch! They are still there, but their text color is white 

Ahhhh, sure enough! I haven't looked around for other app bars that might have the same issue, but I'm sure you've already got it covered 😄 

> This should not happen!

Here's a screen recording of the behavior I'm seeing:

<a href="https://user-images.githubusercontent.com/1200432/91463058-8be09380-e850-11ea-97a0-de0c2f51ac26.gif"></a>

osullivanchris:
Lots to catch up on here. Thanks for all the feedback and discussion. Some of the questions @iamthomasbishop raised have already been answered/discussed. So I'm just going to answer the open questions. Anything else I've read and agree with what's said above!

**App Bar Behaviour in the editor**
- First choice is the divider on default load, shadow on scroll
- If @khaykov can't find a way to make it work...hmm I guess I'd probably lean for shadow rather than divider. Its a little more in keeping with what we're doing elsewhere

**Previews**
- Good spot on the missing CTA
- Great that previews are the same thing used everywhere 👌 
- A good spot by Thomas that white on white looks wrong. I hadn't run into a site with a white header yet. Think we'll need a divider on load on this one, switching to scroll, per other discussions.

**Status bar color -> white**

Yes! Totally agree. Its taken care of in another PR :) 

**Suggestion to use divider everywhere** 

So far, my opinion had been to leave it flush almost everywhere by default, but make exceptions as necessary. Mostly this is because the work was lead by the main tabs with the large titles. There it wouldn't make sense to show a divider. And the spacious headings with no separation look really nice. So I was considering the divider as the exception rather than the rule.

However, its cropping up more and more. We've already introduced this behaviour on pages with tabs, where we show a divider on default load, and change to shadow on scroll. We're now doing it on editor and preview. So I don't know if this tips the balance towards making it the default. @khaykov  do you have a preference here? We could pause on the decision for now. Review towards the end. Some pages looked fine with no divider by default.

iamthomasbishop:
> So far, my opinion had been to leave it flush almost everywhere by default, but make exceptions as necessary. Mostly this is because the work was lead by the main tabs with the large titles. There it wouldn't make sense to show a divider. And the spacious headings with no separation look really nice. So I was considering the divider as the exception rather than the rule.

That make sense @osullivanchris -- I think what I was referring to is mainly relevant for the "compact" app bars, meaning the ones that aren't using large titles (it def makes sense to go borderless in the cases of large titles).

khaykov:
@osullivanchris I fixed the preview behavior and added a separator below AppBar. Dealing with WebView was harder then I expected, and I was not able to solve one small issue - the toolbar is not returning to flush state if you reach the top of the screen with a fling. If you scroll all the way up, like on this gif it works. I assume there is a way to make it work, but I already spend to much time on previews :)

[Image from Gyazo](https://gyazo.com/f452e32f88bc0e6b6561c4fae553cc58)

I also added a separator to the editor and update appbar menu text color. 

[Image from Gyazo](https://gyazo.com/ccaea633fc65fd7e3b080389149dfc89)

[Image from Gyazo](https://gyazo.com/dfd08fb53cf2cf5a8f769c7d89744e45)


Regarding this:

> We've already introduced this behaviour on pages with tabs, where we show a divider on default load, and change to shadow on scroll. We're now doing it on editor and preview. So I don't know if this tips the balance towards making it the default. @khaykov do you have a preference here? We could pause on the decision for now. Review towards the end. Some pages looked fine with no divider by default.

I don't really have a strong opinion on this, but it can be changed relatively easily at any point.

osullivanchris:
> I don't really have a strong opinion on this, but it can be changed relatively easily at any point.

Cool. I think we've done a lot here for one round. Let's revisit that default when we look at the build as whole when these issues are all merged.

iamthomasbishop:
> I also added a separator to the editor and update appbar menu text color.

@khaykov that looks solid. 

One tiny thing, just a suggestion: I think we can exclude the app bar's bottom divider on dark theme, because the diff in color between the app bar and canvas should provide sufficient separation. // cc @osullivanchris agree or disagree? 😀

osullivanchris:
> One tiny thing, just a suggestion: I think we can exclude the app bar's bottom divider on dark theme, because the diff in color between the app bar and canvas should provide sufficient separation. // cc @osullivanchris agree or disagree? 😀

I agree! Buuut the app bar is not meant to be a different color to the canvas haha. Maybe it is just for the editor.

iamthomasbishop:
> Buuut the app bar is not meant to be a different color to the canvas haha. Maybe it is just for the editor.

Ahh, okay then -- if the app bar ends up being the same color as the canvas, then divider it is! 😆

khaykov:
@osullivanchris @iamthomasbishop regarding this:

> Buuut the app bar is not meant to be a different color to the canvas haha. Maybe it is just for the editor.

I asked around, and @mchowning pointed me to the place where we set the editor's background-color. Looks like it's currently not possible to change it from the client, and it's also shared with iOS. 

Regardless, thank you for all the design reviews! I can safely this PR for code review 👍

planarvoid (approved):
Looks good and works well so 👍  The app is looking amazing with the new toolbars! And the solution for the editor is good as well

khaykov:
Thank you so much for the review, @planarvoid!

PR #12791, opened 2020-08-25
Fix undo block site action

### LINKED ISSUE #12790: BlockSite undo action doesn't work
### Expected behavior
When I click on the undo action in the snackbar I expect all posts that have been deleted to re-appear.

### Actual behavior
When I click on the undo action in the snackbar nothing happens.

1. Open Reader
2. Click on the "more" button on one of the items
3. Click on "Block Site"
4. Notice posts from that blog disappear
5. Click on "undo" in the snackbar
6. Notice the posts don't reappear unless you pull down to refresh

malinajirka:
Fixes #12790 

This PR fixes two issues both related to the "undo block site" action. 

1. When a user clicks on “block site” we perform three actions 
 - load posts which are about to be deleted and we store them into BlockedBlogResult
 - delete all posts with the provided blogId in our local db
 - we send a request to the server to block that site
2. When the user clicks on undo
- we re-add the posts retained in BlockedBlogResult
- we send a request to the server to unblock the site

This sounds ok, but there were two bugs in the code
1. When we loaded the posts into BlockedBlogResult we used -`ReaderPostTable.getPostsInBlog` which returns only posts with `tag_name=""`. But the delete method removes all posts no matter the "tag_name". So when there weren't any posts with `tag_name=""`, the BlockedBlogResult was empty and the undo action didn't restore any posts.
2. Our post table can have multiple instances of the same post (one per tag). But the delete method removes all posts no matter their tag. The undo action inserts all posts from `BlockedBlogResult` under an empty tag. So if the user is lets say on "Nature" tag screen -> the posts disappear when they click on "block site" but they never reappear as they are re-inserted with `tag_name=""`.

I decided to store Map<Pair<reader_tag, reader_type>, ReaderPostList> into the BlockedBlogResult, so we actually restore all the deleted rows.

To test:

1. Open Reader
2. Click on the "more" button on one of the items
3. Click on "Block Site"
4. Notice posts from that blog disappear
5. Click on "undo" in the snackbar
6. Notice the posts re-appear

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderBlogActions.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderBlogActions.java
@@ -1,6 +1,7 @@
 package org.wordpress.android.ui.reader.actions;
 
 import android.text.TextUtils;
+import android.util.Pair;
 
 import com.android.volley.Request;
 import com.android.volley.Response;
@@ -13,9 +14,12 @@ import org.wordpress.android.WordPress;
 import org.wordpress.android.analytics.AnalyticsTracker;
 import org.wordpress.android.datasets.ReaderBlogTable;
 import org.wordpress.android.datasets.ReaderPostTable;
+import org.wordpress.android.datasets.ReaderTagTable;
 import org.wordpress.android.models.ReaderBlog;
 import org.wordpress.android.models.ReaderPost;
 import org.wordpress.android.models.ReaderPostList;
+import org.wordpress.android.models.ReaderTag;
+import org.wordpress.android.models.ReaderTagType;
 import org.wordpress.android.ui.reader.actions.ReaderActions.ActionListener;
 import org.wordpress.android.ui.reader.actions.ReaderActions.UpdateBlogInfoListener;
 import org.wordpress.android.util.AppLog;
@@ -25,11 +29,13 @@ import org.wordpress.android.util.VolleyUtils;
 import org.wordpress.android.util.analytics.AnalyticsUtils;
 
 import java.net.HttpURLConnection;
+import java.util.Map;
 
 public class ReaderBlogActions {
     public static class BlockedBlogResult {
         public long blogId;
-        public ReaderPostList deletedPosts;
+        // Key: Pair<ReaderTagSlug, ReaderTagType>, Value: ReaderPostList
+        public Map<Pair<String, ReaderTagType>, ReaderPostList> deleteRows;
         public boolean wasFollowing;
     }
 
@@ -443,7 +449,7 @@ public class ReaderBlogActions {
     public static BlockedBlogResult blockBlogFromReader(final long blogId, final ActionListener actionListener) {
         final BlockedBlogResult blockResult = new BlockedBlogResult();
         blockResult.blogId = blogId;
-        blockResult.deletedPosts = ReaderPostTable.getPostsInBlog(blogId, 0, false);
+        blockResult.deleteRows = ReaderPostTable.getTagPostMap(blogId);
         blockResult.wasFollowing = ReaderBlogTable.isFollowedBlog(blogId);
 
         ReaderPostTable.deletePostsInBlog(blogId);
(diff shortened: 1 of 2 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderBlogActions.java
```diff
@@ -443,7 +449,7 @@ public void onErrorResponse(VolleyError volleyError) {
     public static BlockedBlogResult blockBlogFromReader(final long blogId, final ActionListener actionListener) {
         final BlockedBlogResult blockResult = new BlockedBlogResult();
         blockResult.blogId = blogId;
-        blockResult.deletedPosts = ReaderPostTable.getPostsInBlog(blogId, 0, false);
+        blockResult.deleteRows = ReaderPostTable.getTagPostMap(blogId);
```
ashiagr, line 452: Can we rename it as `deletedRows`?

WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderBlogActions.java
```diff
@@ -478,9 +484,7 @@ public static void undoBlockBlogFromReader(final BlockedBlogResult blockResult)
         if (blockResult == null) {
             return;
         }
-        if (blockResult.deletedPosts != null) {
-            ReaderPostTable.addOrUpdatePosts(null, blockResult.deletedPosts);
-        }
+       undoBlockBlogLocal(blockResult);
```
ashiagr, line 487: It seems we're doing an optimistic undo action here but not reverting to original state if the undo request fails. So the app will display these posts even when the blog is blocked on the server.

Do you consider it as an issue?

ashiagr (commented):
Thanks for these fixes @malinajirka! I tested the flow and it works as described. Just left two comments, lmkwyt.

WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderBlogActions.java
malinajirka: The `undoBlockBlogFromReader` isn't invoked just when the user clicks on the "Undo" button but also if the user clicks on "block site" button and the request fails. Since we don't know the state on the server, there isn't much we can do imo. We assume that if the second request fails (undo), the first request (block site) probably failed as well.

WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderBlogActions.java
malinajirka: Ohh, that's a typo. Fixed in 6b504dc27eb7af9b545088102debef0261299b56

WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderBlogActions.java
ashiagr: AFAIU we call `undoBlockBlogFromReader` only when the user clicks on the "Undo" button. This method includes server call to undo block. 
When the user clicks on the "block site" button and the request fails, we just call `undoBlockBlogLocal` that only reverts updates to local db.
I meant to say that if the "undo" request fails inside `undoBlockBlogFromReader` we revert to original state. Here we know that undo action has failed on the server.

Having said that, I'm fine with this assumption as they happen in quick succession:
> We assume that if the second request fails (undo), the first request (block site) probably failed as well.

malinajirka: Ohh, you are right. We can handle it. However, I think it's probably not worth the effort due to the assumption mentioned above. Moreover, the worst case scenario is that the user will see some posts which are blocked until they refresh the page. Wdyt?

Either case, good catch :)

ashiagr: Let's go with this assumption 👍

ashiagr (approved):

PR #12792, opened 2020-08-25
Issue/12028 fix block site action

malinajirka:
Parent issue #12028

Merge instructions:
1. Make sure https://github.com/wordpress-mobile/WordPress-Android/pull/12791#issuecomment-679961254 is merged
2. Remove "Not Ready for Merge" label
3. Merge this PR

This PR refactors the BlockSite action into a useCase and re-uses that usecase in the ReaderPostListFragment and the "new" discover tab VM.

Since the useCase needs to return multiple results (in progress states), I decided to implement the use case using Kotlin Flows. The "Undo" action was causing a lot of design issues, since it's technically another action within a single usecase. I discussed the design with Vojta and he suggested to extract both SnackBarHolders (including the Undo action) and refresh posts events from the UseCase and keep it single focused - block site action. 
I decided to go with this solution after some consideration as it makes a lot of sense. The usecase is very focused now and it isn't tied to the view layer at all. The `ReaderPostCardActionsHandler` is responsible for transforming the usecase states into corresponding Ui Actions - show snackbar, refresh post.
Let me know what you think about it :). I'll add tests when we decide this is the solution we want to go with.

To test - Enable RI FF
1. Open Reader
2. Open Discover tab
3. Click on the more button on one of the items
4. Click on "Block Site"
5. Notice all posts from this site disappear
6. Notice Snackbar message with "Undo" action is shown
7. Click on "Undo" and notice the posts re-appear
--------------------
1. Open Reader
2. Open Discover tab
3. Click on the more button on one of the items
4. Turn on airplane mode
5. Click on "Block Site"
6. Notice an error snackbar appears

Perform the same steps on the Following Tab.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderBlogActions.java
```diff
@@ -442,19 +442,22 @@ public void onErrorResponse(VolleyError volleyError) {
         WordPress.sRequestQueue.add(request);
     }
 
-    /*
-     * block a blog - result includes the list of posts that were deleted by the block so they
-     * can be restored if the user undoes the block
-     */
-    public static BlockedBlogResult blockBlogFromReader(final long blogId, final ActionListener actionListener) {
```
malinajirka, line 449: I've split the `blockBlogFromReader` method into two methods `blockBlogFromReaderLocal` and `blockBlogFromReaderRemote `.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
```diff
@@ -52,7 +62,11 @@ class ReaderPostCardActionsHandler @Inject constructor(
     val snackbarEvents: LiveData<Event<SnackbarMessageHolder>> = _snackbarEvents
 
     private val _preloadPostEvents = MediatorLiveData<Event<PreLoadPostContent>>()
-    val preloadPostEvents = _preloadPostEvents
+    val preloadPostEvents: LiveData<Event<PreLoadPostContent>> = _preloadPostEvents
+
+    // Used only in legacy ReaderPostListFragment. The discover tab observes reactive ReaderDiscoverDataProvider.
+    private val _refreshPosts = MediatorLiveData<Event<Unit>>()
```
malinajirka, line 77: For future reference: `refreshPost` livedata are used only in the legacy ReaderPostListFragment. The new "discover" tab is observing all changes using the "DiscoverDataProvider" so it refreshes the post automatically.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
```diff
@@ -121,16 +135,43 @@ class ReaderPostCardActionsHandler @Inject constructor(
...
+        defaultScope.launch {
+            blockBlogUseCase.blockBlog(blogId).collect {
+                when (it) {
+                    is SiteBlockedInLocalDb -> {
+                        _refreshPosts.postValue(Event(Unit))
+                        _snackbarEvents.postValue(
+                                Event(
+                                        SnackbarMessageHolder(string.reader_toast_blog_blocked, string.undo, {
+                                            defaultScope.launch {
+                                                ReaderBlogActions.undoBlockBlogFromReader(it.blockedBlogData)
```
malinajirka: We could move the "undo" action into it's own usecase, but it felt like an overkill since it's a fire&forget action.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
```diff
@@ -121,16 +135,43 @@ class ReaderPostCardActionsHandler @Inject constructor(
         _navigationEvents.postValue(Event(OpenPost(post)))
     }
 
-    private fun handleBlockSiteClicked(postId: Long, blogId: Long) {
-        AppLog.d(AppLog.T.READER, "Block site not implemented")
+    private fun handleBlockSiteClicked(blogId: Long) {
```
malinajirka, line 166: As mentioned in the description `ReaderPostCardActionsHandler` is responsible for transforming the state into a UI action.

ashiagr (commented):
Great job @malinajirka! I liked: 
1. No use of UI elements like `SnackbarMessageHolder` in the useCase.
2. Ability to send multiple updates from the useCase without the use of `LiveData` that need to be explicitly observed from the VM (thanks to `Flows`). Plus you did not use any of the experimental `Flow` APIs.
3. Ability to refresh the post on both legacy and new discover adapter.

One thing I'd like to discuss is about the sealed class use in the useCase. 

Do you think we should have separate sealed classes for each useCase that would like to use similar architecture? There might be common types like `NoNetwork` or `RemoteError` that we could extract in a common class. Lmkwyt.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
ashiagr: > it felt like an overkill 

I just hope it doesn't become a wrong precedent in the future.

malinajirka:
> I liked:
> 1. No use of UI elements like SnackbarMessageHolder in the useCase.
> 2. Ability to send multiple updates from the useCase without the use of LiveData that need to be explicitly observed from the VM (thanks to Flows). Plus you did not use any of the experimental Flow APIs.
> 3. Ability to refresh the post on both legacy and new discover adapter.

Awesome! Thanks for the clear summary of the benefits;)!

> One thing I'd like to discuss is about the sealed class use in the useCase.

I pinged Annmarie about a similar thing last week (`ReaderRepositoryCommunication`). 

I completely agree that we might want to re-use some of the types. At the same time, I'd like to avoid forcing a use-case to follow only the existing types (eg. `SiteBlockedInLocalDb` is quite useful but very specific to this use case). 

I'd consider creating a "consolidation" task on the project board and get back to it when we have all the use cases. It'll be easier to extract the common logic and we'll be able to make a better decision when we have all the info. Wdyt @ashiagr ?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
malinajirka: If you prefer, I can wrap it into a use case. It won't hurt I guess. Wdyt?

ashiagr:
> I'd consider creating a "consolidation" task on the project board and get back to it when we have all the use cases. It'll be easier to extract the common logic and we'll be able to make a better decision when we have all the info. Wdyt @ashiagr ?

Works for me 👍

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
ashiagr: Yep lets wrap it in usecase

malinajirka:
Thanks @ashiagr! I've merged develop into the branch and wrapped the undo action into a usecase. It's ready for another round.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
malinajirka: Fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/12792/commits/be6113a938999241157e5ccd54667fbbc65a7c42

ashiagr (approved):
Thanks! All looks good to me. We can add tests in another PR. Approving this one 🚀.

PR #12793, opened 2020-08-25
Move media picker files into ui package

planarvoid:
As discussed, this PR moves the Media picker classes from `ui.photopicker.mediapicker` package to `ui.mediapicker`. The reason is to make it easier to remove the deprecated classes once the work on the project is done.

To test:
- there is nothing to test here, this was an automatic conversion

PR submission checklist:


WordPress/src/main/AndroidManifest.xml
```diff
@@ -577,7 +577,7 @@
             android:theme="@style/WordPress.NoActionBar" />
 
         <activity
-            android:name=".ui.photopicker.mediapicker.MediaPickerActivity"
+            android:name="org.wordpress.android.ui.mediapicker.MediaPickerActivity"
```
develric: I would consider to remove the base package `org.wordpress.android` for consistency with the other entries. wdyt?

develric (commented):
Hi @planarvoid , smoke tested this (and seems ok to me) and left a minor comment. Let me know 🙇‍♂️

WordPress/src/main/AndroidManifest.xml
planarvoid: sure, this was added automatically by AS

planarvoid:
thanks for the review @develric , I've updated the package name in AndroidManifest

develric (approved):
Thank for the follow up @planarvoid ; LGTM 👍 !

PR #12799, opened 2020-08-26
AppBar tweaks

khaykov:
This PR adds a variety of AppBar related visual changes:
 - Updated design of Site Picker on My Site screen.
 - The border around gravatar in My Site title and Pages/Posts author filter selector.
 - Updated scrolling behavior of Reader and Notifications tabs.
 - Blavatar background in Site Picker and My Site screen.

If you are interested, more details are available in the comments of this PR.

To test:
- Confirm that the site picker on My Site screen looks and behaves as expected.
- Set a site title to something long, and make sure that when it wraps to second line text size is reduced.
- Confirm that the scrolling with AppBar in Reader and Notifications behave in the same way.
- Make sure that the default site icon (globe on transparent BG) is visible over gray BG on My Site and Site Picker screen.

PR submission checklist:


osullivanchris (changes requested):
> For now, I only fixed the reader scroll, to make it consistent with notifications. 

I actually think we wanted to do the opposite change sorry! If I'm remembering the discussion correctly, we had feedback that too much of the view is taken up by navigation on the notifications screen. So we should make notifications similar to reader, hiding as much nav as possible on scroll. In the GIF it looks like you've done it the other way and made reader match notifications. Sorry for any confusion! 

> I think we can also squeeze a site picker redesign here if it's ready.

sounds good. I think the only question was how big a task this was for you? @frosty had asked on another post. If its feasible to build it I can do up a spec for this today.

> Any else I should add?

Not sure right now. There's a few things spread across other GH issues. I guess if they're captured somewhere else that's fine for now. @0nko had an idea about making the avatar even bigger on initial load. We could explore that for one thing.

khaykov:
@osullivanchris Thank you so much for comments and specs for site picker 🙇 

> I actually think we wanted to do the opposite change sorry! If I'm remembering the discussion correctly, we had feedback that too much of the view is taken up by navigation on the notifications screen. So we should make notifications similar to reader, hiding as much nav as possible on scroll. In the GIF it looks like you've done it the other way and made reader match notifications. Sorry for any confusion!

So, I think I fixed the issue with toolbar - when scrolling down we will show a collapsed version of the toolbar until you reach the top. This way parallax will not mess up the scrolling. Like this:
[Image from Gyazo](https://gyazo.com/25993643eb4e815d9e86a4f7340ec8cc)

I added the sane behavior to the Notifications screen.

I also updated the site picker! I wasn't sure what is the relation between title and subtitle, so I eyeballed them to get something that resembles your mocks. I had trouble aligning the 12dp icon next to the site URL and making it look good, so I changed its size to match the text (14dp) - if it's critical I can try and make it work with 12dp size.

[Image from Gyazo](https://gyazo.com/99629afd7b27fba67e27dde5e8c21998)

Also, I realized that the default site icon is whit, and it's invisible on white BG, so I added a bg color behind the site icon. Let me know what you think :)

In addition, as @iamthomasbishop mentioned this a couple of times, I updated the status bar color:

[Image from Gyazo](https://gyazo.com/f781a8d78134e83f5fdc490f6ef244bb)

osullivanchris:
> So, I think I fixed the issue with toolbar - when scrolling down we will show a collapsed version of the toolbar until you reach the top. This way parallax will not mess up the scrolling. Like this:

not seeing reader this way in the build just yet. But notifications behaving this way and I think its nice. Do you prefer it? Its nice to have so much of the screen for content. 

> I also updated the site picker! I wasn't sure what is the relation between title and subtitle, so I eyeballed them to get something that resembles your mocks. 

Do you mean the type sizes or just the space between them? Looking at the screenshot of the build, if its just the space I would reduce just a couple pixels if possible.

>  I had trouble aligning the 12dp icon next to the site URL and making it look good, so I changed its size to match the text (14dp) - if it's critical I can try and make it work with 12dp size.

I was also just trying to match the text so 14 is good! 

> Also, I realized that the default site icon is whit, and it's invisible on white BG, so I added a bg color behind the site icon. Let me know what you think :)

yeah that makes sense. @frosty mentioned today we still have a white border around the icon. If the icon was white its border might be unclear now. Could we change the border to 1px of the divider grey that we use and see how that looks? Should change to appropriate grey for dark mode too.

> In addition, as @iamthomasbishop mentioned this a couple of times, I updated the status bar color:

oh yeah looks much better, thanks! 👌

khaykov:
@osullivanchris I added border to the blavatar and removed extra padding between title and subtitle 👍 

[Image from Gyazo](https://gyazo.com/a80c43b6803948d42a61d5df2030f4f6)

osullivanchris:
> @osullivanchris I added border to the blavatar

Ah! I was actually talking about the border around the user avatar. That's what I was talking about on Slack, so we probably were talking about two different things! I think the same thing applies to both though. Could we add the light grey border around the user avatar? Also can you tell me what grey value you're using?

>  and removed extra padding between title and subtitle

cool how much padding was in there before? 

**Chevron icon color**

one other note here. In light mode, the chevron icon looks like its the same color as the other icons (list and quick actions). I'm happy with that. However, in dark mode, it looks like its absolute white. It should also have the same color as the other icons in dark mode I think.

khaykov:
@osullivanchris 

> Ah! I was actually talking about the border around the user avatar. That's what I was talking about on Slack, so we probably were talking about two different things! I think the same thing applies to both though. Could we add the light grey border around the user avatar? Also can you tell me what grey value you're using?

Gotcha! I added a border around avatars (this also affected avatar at Post list). We are using On Surface (white/black) color with Disabled emphasis (38%).

[Image from Gyazo](https://gyazo.com/36cb23137743badea8ab5b1af199decd)

[Image from Gyazo](https://gyazo.com/46df0f48f1a0afe824367520173d1f5e)

[Image from Gyazo](https://gyazo.com/0ad7d20ad770e1c15bcf00aeaff3f4fb)

> cool how much padding was in there before?

I had 4dp there (+ whatever there is from line-height).

> one other note here. In light mode, the chevron icon looks like its the same color as the other icons (list and quick actions). I'm happy with that. However, in dark mode, it looks like its absolute white. It should also have the same color as the other icons in dark mode I think.

Fixed it 👍

osullivanchris:
> Gotcha! I added a border around avatars (this also affected avatar at Post list). We are using On Surface (white/black) color with Disabled emphasis (38%).

cool looks good! Honestly I prefer without, but I think its necessary for the case of a black or white image.

> I had 4dp there (+ whatever there is from line-height).

So I was thinking in between would be perfect, about 2dp. Just to keep you busy 😜 But when I had a good look at the build I noticed the text wasn't exactly vertically centered either. Is it possible to tweak to achieve whats shown below? I'm not exactly sure of how you're defining it with margins and paddings so you dont need to follow the way I've defined it here, but if its possible to achieve the end result? (basically title moves up 4dp, url moves up 2dp). 

We also would need to make sure this doesn't break it when we see the two lines wrapped version

On that note have you checked how it works with long URL?

header tweaks
 

> Fixed it 👍

looks good!

khaykov:
@osullivanchris So this turned out to be a pretty complicated view arrangement, considering it has to change when there are more than 2 lines of text in the title.

I made a custom view, and I think It works ok. There are two ways to center things, as you can see from these screenshots:

We split the space into 2 containers and center text vertically inside each container. I think this resembles what you have in "After" image.
[Image from Gyazo](https://gyazo.com/9f2c127453484a8328d6ce6118e068a4)

Another way is to make both title and subtitle huddle together around the center like this:
[Image from Gyazo](https://gyazo.com/7dd87edb6a1f936368f0b6dbe58bab13)

Let me which one you prefer. And the long URL looks like this:
[Image from Gyazo](https://gyazo.com/c134837d6fc448fc1bdd86c263d0cc76)

osullivanchris:
> @osullivanchris So this turned out to be a pretty complicated view arrangement, considering it has to change when there are more than 2 lines of text in the title.

Sorry to make things complicated!! 😅 

> I made a custom view, and I think It works ok. There are two ways to center things, as you can see from these screenshots:

Before you made it custom, was it just a list item? I am just wondering how much cost I am creating here for the value of tweaking something 1-2dp. Perhaps we're losing out on some default behaviours and scalability by making a custom component.

Do you feel its an ok thing to do in this instance?


> There are two ways to center things

If we are going ahead with it, I actually think the second option (title and subheading huddling together) is matching the mock nearly exactly, so I would go with that one. Does it wrap nicely to achieve the two-line state? Or is it a totally different layout? If its a different layout entirely I'm just wondering how that works...do you need to run some calculation or something before you display the name?

khaykov:
@osullivanchris 
> Before you made it custom, was it just a list item? I am just wondering how much cost I am creating here for the value of tweaking something 1-2dp. Perhaps we're losing out on some default behaviours and scalability by making a custom component.

> If we are going ahead with it, I actually think the second option (title and subheading huddling together) is matching the mock nearly exactly, so I would go with that one. Does it wrap nicely to achieve the two-line state? Or is it a totally different layout? If its a different layout entirely I'm just wondering how that works...do you need to run some calculation or something before you display the name?

So, it was just 2 text views before. It's pretty easy to split the space equally and manipulate title and subtitle inside their containers (eg. huddle them together), the problem is that we want to change the text size when it wraps, and for this, we need to have a special logic that watches the number of lines in the text, updates the size accordingly and manipulates layout in a way that removes the equal sizing from the title and subtitle containers (and restores it if the title is changed back to short one). I don't like custom views, but we don't lose any text-specific functionality, and I think it's worth to have the most prominent part of the app looking as good as possible :)


It also wraps nicely:

[Image from Gyazo](https://gyazo.com/24bfac4984e640a44f6066ee1ad190ff)

osullivanchris:
Ok cool. Thanks for all the explanation @khaykov and glad you feel the customisation is justified! I forgot in my previous comment that we're changing type-size, so that alone is creating something complex.

I'm happy to go ahead with this if you are. Thanks!

WordPress/src/main/res/layout/base_site_listitem.xml
```diff
@@ -6,6 +6,7 @@
     android:gravity="center_vertical">
 
     <ImageView
+        style="@style/MySiteBlavatarImageView"
```
renanferrari, line 9: Apparently applying this style here makes the site list icon have a second background. It becomes more apparent when you look at the corners of the icon:
s1

Here's what it should look like:
s2
Maybe this can be fixed by loading the icon image using `BLAVATAR_ROUNDED_CORNERS` [here](https://github.com/wordpress-mobile/WordPress-Android/blob/eebe812bcaa52fdc45fff257f335c9bc9fdb9f32/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt#L991). That would also account for the difference in the background color of the default icon.

renanferrari (commented):
Thanks for the changes @khaykov, I really like how the new site picker layout is turning out. Code looks pretty good, but I did find some issues, mostly design-related:

1. There are some inconsistencies between the site icon from the new site picker and the existing one from the site list screen:
    - Background color is a bit lighter for the default icon in the site list screen. I think it would be good if we could use the same color resource for both. I also left a comment below regarding another issue, with a solution that I think would also address this.
    - There is no border on the site list icon. I noticed @osullivanchris mentioned that one of the reasons we're using a border there is to account for icons with white/black backgrounds, but it seems to me that those look better when the icon doesn't have a border. Here are some examples in [light mode](https://user-images.githubusercontent.com/830056/92062432-1eac9100-ed6f-11ea-98e3-259f092598a4.png) and [dark mode](https://user-images.githubusercontent.com/830056/92062427-1c4a3700-ed6f-11ea-823b-1c92028a894a.png). @osullivanchris Wouldn't it be fine to get rid of the border in the new site picker icon as well?
2. The custom `MySiteTitleAndSubtitleLabelView` seemed to work well, but then I found this weird bug where it keeps resizing (I think). I managed to capture it in video [here](https://cloudup.com/c0nfQQLl7E6). It only happens for a few sites, but I couldn't figure out why. Maybe you already gave this a try, but I wonder if it wouldn't be possible to use a Constraint Layout + Autosizing TextViews to achieve a similar effect?

khaykov:
Thanks for the review @renanferrari  ! The title glitch was from something we haven't considered - when changing site title size changes it's line numbers... Eg. if the title wraps, we change its size to smaller one, and it stops wrapping, so we change it's size to a bigger one.. and so on :) I originally tried to achieve this behavior with autosizing textview and constraint layout, but it didn't work well. But I tried using layout logic from custom view + autosizing textview, and it works!

I'll fix site icon after we hear back from @osullivanchris :)

osullivanchris:
Thanks for spotting some design issues @renanferrari !

On the border, there's tradeoffs whatever we do. It is actually interesting that there are examples where the icon looks good with no border. But that might be lucky rather than typical. [Made a final call](https://github.com/wordpress-mobile/WordPress-iOS/pull/14797#issuecomment-686381167) to use a subtle border on iOS.

However I'm mostly looking at the avatar. Actually initially I asked @khaykov to implement a border for the Me avatar in this PR. But we had a misunderstanding that it was about the site icon. In any case, my choice would probably be to use a subtle border for both.

Also I notice in https://user-images.githubusercontent.com/830056/92060577-8dd3b680-ed6a-11ea-9ef2-75af6323e11e.png - on default the site icon has no border radius. But on selection, it has a radius. Is that by design? We should probably use the same radius everywhere when we do.

renanferrari:
>The title glitch was from something we haven't considered - when changing site title size changes it's line numbers...

Oh, that makes sense! It seems to be a tricky one, so I'm glad I was able to catch it by chance.

>I originally tried to achieve this behavior with autosizing textview and constraint layout, but it didn't work well.

Well, I can't say I'm surprised by that 😆 Those two are usually troublesome when put together, but I thought it was worth to ask, just in case.

>On the border, there's tradeoffs whatever we do. It is actually interesting that there are examples where the icon looks good with no border. But that might be lucky rather than typical.

Thanks for clarifying, @osullivanchris. That makes sense for me. Out of curiosity, do you think it's fine to leave the icons from the site list without borders or is this something we should look into changing later on (in another PR)?

>Also I notice in https://user-images.githubusercontent.com/830056/92060577-8dd3b680-ed6a-11ea-9ef2-75af6323e11e.png - on default the site icon has no border radius. But on selection, it has a radius.

I'm not sure if I see what you're referring to. All icons seem to have a radius for me in that image 🤔

osullivanchris:
> Thanks for clarifying, @osullivanchris. That makes sense for me. Out of curiosity, do you think it's fine to leave the icons from the site list without borders or is this something we should look into changing later on (in another PR)?

I think its fine for now. Will discuss with Matt whether its worth changing or what we've learned here 😆 

> I'm not sure if I see what you're referring to. All icons seem to have a radius for me in that image 🤔

oh sorry. It was the [earlier image that didn't have radius](https://user-images.githubusercontent.com/830056/92062523-50255c80-ed6f-11ea-8877-8f07b4f893e6.png). But perhaps it was just related to the issue with backgrounds that you brought up

khaykov:
@osullivanchris @renanferrari I hope I got all the details correctly:
- Default icon color is same at My Site and Site List.
- I added the border to Site List, and changed its color to a lighter grey.
- I fixed the corner issue in Site List.

[Image from Gyazo](https://gyazo.com/c7df4199ca0ddd9f77624e3277c6bf32)

[Image from Gyazo](https://gyazo.com/0f4dcb6fe13b7ccf948d3d7f3b94f631)

osullivanchris:
Yeah I think this is all fine for me, thanks @khaykov !

WordPress/src/main/res/layout/my_site_title_subtitle_view.xml
```diff
@@ -9,6 +9,9 @@
         android:layout_width="match_parent"
         android:layout_height="0dp"
         android:gravity="bottom"
+        app:autoSizeMaxTextSize="@dimen/my_site_name_label_single_line_text_size"
+        app:autoSizeMinTextSize="@dimen/my_site_name_label_double_line_text_size"
+        app:autoSizeTextType="uniform"
```
renanferrari, line 14: Glad you were able to make this work! 👍

WordPress/src/main/res/layout/my_site_fragment.xml
```diff
@@ -50,104 +50,78 @@
...
-                    android:gravity="center_vertical"
-                    android:maxLines="1"
-                    android:text="@string/my_site_btn_switch_site" />
+                        android:layout_alignParentEnd="true"
+                        android:layout_centerVertical="true"
+                        android:layout_marginStart="@dimen/margin_extra_large"
+                        android:background="?attr/selectableItemBackgroundBorderless"
+                        android:contentDescription="@string/my_site_btn_switch_site"
+                        android:src="@drawable/ic_chevron_down_white_24dp"
+                        app:tint="?attr/wpColorOnSurfaceMedium" />
```
renanferrari: I didn't notice this the first time, but I believe we should give this button a larger touch area.

Giving it a padding of `12dp` and a margin of `4dp` should be enough. You'll probably need to remove the inner padding from the parent view and adjust the margin from the siblings, but I think it's worth it. Then you might also replace the background with a drawable [like this](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/res/drawable/ripple_oval.xml), to make the ripple bounded to its new size. 

Of course, consider all of that a merely suggestion (totally up to you!).

WordPress/src/main/res/values/styles.xml
```diff
@@ -502,6 +524,11 @@
         <item name="android:layout_marginStart">@dimen/content_margin</item>
     </style>
 
+    <style name="MySiteBlavatarImageView">
+        <!-- We need to specify both bg and bg ting for this to work in all cases -->
```
renanferrari: I believe this comment can be removed now?

```suggestion
    <style name="MySiteBlavatarImageView">
```

renanferrari (approved):
Great job on these changes! 

I left a few comments and suggestions below, but they are mostly nitpicks. Other than that, I also found [some issues on API 21](https://user-images.githubusercontent.com/830056/92257853-43f2e980-eeac-11ea-96f5-9a7c6f9c2381.png), but I don't believe those are critical, so feel free to tackle them in a separate PR and to merge this one after considering the comments below.

WordPress/src/main/res/layout/my_site_fragment.xml
khaykov: Thank you, that's a great suggestion!

WordPress/src/main/res/layout/my_site_title_subtitle_view.xml
khaykov: Totally missed this one🤦

khaykov:
@renanferrari Thank you for suggestion! Ready for another round :)

renanferrari (approved):
LGTM! :shipit:

PR #12811, opened 2020-08-27
Introduce media picker setup

planarvoid:
Previously we were using the browser type enum to determine the behaviour of the photo picker. I don't think this is a great approach because it requires a lot of knowledge from developers - you have to know the exact behaviour of the picker in each of the scenarios which is not documented anywhere. It also leads to a lot of random pieces logic all over the places. In this PR I'm introducing a more explicit `MediaPickerSetup` which has the fields that directly affect the behaviour of the media picker. This way you can check in the launcher that for example you're starting a photo picker from Gutenberg -> this means that only single selection is available, you can only select photos and you want to load photos from the device. This makes any changes to the behaviour very simple. 

As part of this PR I've move the intent building of the MediaPickerActivity inside its companion object.

To test:
- Turn the `CONSOLIDATED_MEDIA_PICKER` flag on
- Test the basic media flows - GB image, gallery, video, Stories, Avatar selection (you don't have to test everything, mostly just smoke test nothing is broken)

PR submission checklist:


develric (commented):
Thanks for this PR @planarvoid 🙇‍♂️ , I like and agree with the introduction of the MediaPickerSetup 👍 ; we are still passing around the browserType in the MediaPickerActivity Intent, not sure if you plan to remove it in a later PR (or maybe we cannot fully) but was wondering if we actually need it anymore in the .mediapicker package at least, wdyt?

While testing I got the following:

1. When changing avatar icon in the signup or standard scenario I got the following:

```
Caused by: kotlin.TypeCastException: null cannot be cast to non-null type org.wordpress.android.fluxc.model.SiteModel
        at org.wordpress.android.ui.mediapicker.MediaPickerActivity.onCreate(MediaPickerActivity.kt:112)
```

I think it's a known missing safe cast that's fixed in develop so we should be fine (let me know if you agree 🙇‍♂️  ; also I think we will fully test again the scenarios before removing the feature flag anyway)

2. Not sure it is related to this PR, but when opening the picker from a GB block (image etc...) for me the `x` icon to close the picker doesn't work (the picker stays open)

The others scenarios worked for me. Let me know 🙇‍♂️ .

planarvoid:
thanks for the review @develric 👍 
> we are still passing around the browserType in the MediaPickerActivity Intent, not sure if you plan to remove it

It's only used in the MediaPickerActivity and it felt like something we'll touch later. It's not used anywhere in the fragment and the view model. 

> When changing avatar icon in the signup or standard scenario I got the following: Caused by: kotlin.TypeCastException...

This should be fixed by develop merge to the branch. 

> Not sure it is related to this PR, but when opening the picker from a GB block (image etc...) for me the x icon to close the picker doesn't work (the picker stays open)

I've fixed this. It was caused by a wrong import for the `R.id.home` for the home icon during the Kotlin conversion. Correctly we should be importing `android.R.id.home`. After this fix this started working.

This should be ready for another round!

develric (approved):
Hi @planarvoid 👋

> It's only used in the MediaPickerActivity and it felt like something we'll touch later. It's not used anywhere in the fragment and the view model.

👍

> This should be fixed by develop merge to the branch.

Confirm it works well after merging from develop.

> I've fixed this. It was caused by a wrong import for the R.id.home for the home icon during the Kotlin conversion. Correctly we should be importing android.R.id.home. After this fix this started working.

It works thanks 🙇‍♂️

LGTM 👍 !

PR #12817, opened 2020-08-27
Add check for already running

zwarm:
Parent #12039 

This PR checks that the LoadMore fun does not kickoff duplicate requests by checking for a running process before starting another. 

I'm on the fence as to whether we want to wrap all "fetch" requests with a check, would love some input. Thanks.

PR submission checklist:


malinajirka (approved):
LGTM, thanks @zwarm! 

I don't think the other requests matter that much. This request is a bit specific, since it has a side-effect - we increment the "refresh" parameter. Moreover, the user can scroll to bottom -> top -> bottom -> ... and the vm will send the request everytime one of the last items in the list gets rendered.

PR #12818, opened 2020-08-28
Consolidate tracking onTabChanged in the main Reader view

malinajirka:
Parent issue #12028

I recommend reviewing this PR by commits.

Goal of this PR is to correctly track TabShown events in the main Reader view.
- Following tab - `reader_following_shown`
- Discover tab - `reader_discover_shown`
- Liked tab - `reader_liked_shown`
- Custom tab (available only on old .com accounts) - `reader_custom_tab_shown`
- Saved tab - `reader_saved_list_shown`
                  - `source` property `reader_filter` - tab in the main Reader view
                  - `source` property `post_list_saved_post_notice` - action in snackbar shown after the user bookmarks a post in the main Reader view
                  - `source` property `post_details_saved_post_notice` - action in snackbar shown after the user bookmarks a post in post detail view

The events are tracked when the user navigates to the tab
- When the user opens the main Reader view from the bottom navigation bar
- When the app comes from background and the main Reader view is shown
- When the user changes tab in the main Reader view

The events are NOT tracked when
- the device orientation changes
- user opens eg. post detail and they come back to the main Reader view

To test - enable RI FF
Verify that the events are tracked as per the description above.
+
1. Open Reader - discover tab
2. Kill the app
3. Open the app and notice only "reader_discover_shown" is tracked

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/tracker/ReaderTracker.kt
```diff
@@ -59,4 +76,52 @@ class ReaderTracker @Inject constructor(private val dateProvider: DateProvider)
     fun getAnalyticsData(): Map<String, Any> {
         return trackers.entries.associate { it.key.propertyName to it.value.accumulatedTime }
     }
+
+    fun trackReaderTabIfNecessary(readerTab: ReaderTab) {
+        if (appPrefsWrapper.isReaderImprovementsPhase2Enabled() && readerTab != appPrefsWrapper.getReaderActiveTab()) {
```
zwarm: @malinajirka - I know we can't fix the tracking issue introduced during the refactor, but doesn't using the FF prevent correct data from being collected after the PR is merged? Do we really need to gate this?

zwarm (approved):
Nice work @malinajirka !
I left one comment about tracking only when FF is on, so if this is not an issue please go ahead and merge.

I think using AppPrefs was the way to go with this one, we definitely needed a way to track what happened last to properly recover from background situations. I also like the addition of ReaderTracker; easy to follow, use, and test.

The testing went swimmingly. I see tracking for the tabs when tapped on, when selecting bottom nav tabs, on return from background. I did not see tracking events for orientation changes.

WordPress/src/main/java/org/wordpress/android/ui/reader/tracker/ReaderTracker.kt
malinajirka: You are absolutely right, this doesn't  need to (shouldn't) be behind a FF. Great catch ;). Fixed in ebd0bad832e0e7cc2e55ec076a28e054b7417fd5

malinajirka:
Thanks for the review @zwarm! I've removed the FF as you suggested ;). It's ready for another round.

PR #12820, opened 2020-08-28
Refactor request more check

zwarm:
Parent issue #12039 

This PR handles the instance in which the Discover service job does not get scheduled, which could lead to a forever progress spinner. It also introduces an extension function on ReaderDiscoverCommunication to determine if the task returned is a REQUEST_MORE. This is then used in the communication channel to flip the `isLoadMoreRequestInProgress` to false.

To Test
- Fire up the app and ensure the FF flag is flipped on
- Tap reader -> discover
- Scroll down to trigger the loading more request
- View the logcat for READER: reader discover load more cards task is finished

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderRepositoryEvents.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderRepositoryEvents.kt
@@ -1,6 +1,10 @@
 package org.wordpress.android.ui.reader.repository
 
+import org.wordpress.android.ui.reader.repository.ReaderDiscoverCommunication.Error
+import org.wordpress.android.ui.reader.repository.ReaderDiscoverCommunication.Started
+import org.wordpress.android.ui.reader.repository.ReaderDiscoverCommunication.Success
 import org.wordpress.android.ui.reader.services.discover.ReaderDiscoverLogic.DiscoverTasks
+import org.wordpress.android.ui.reader.services.discover.ReaderDiscoverLogic.DiscoverTasks.REQUEST_MORE
 
 sealed class ReaderRepositoryEvent {
     object ReaderPostTableActionEnded : ReaderRepositoryEvent()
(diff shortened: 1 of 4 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderRepositoryEvents.kt
```diff
@@ -39,5 +60,13 @@ sealed class ReaderDiscoverCommunication {
     sealed class Error(open val task: DiscoverTasks) : ReaderDiscoverCommunication() {
         data class NetworkUnavailable(override val task: DiscoverTasks) : Error(task)
         data class RemoteRequestFailure(override val task: DiscoverTasks) : Error(task)
+        data class ServiceNotStarted(override val task: DiscoverTasks) : Error(task)
     }
 }
+
+fun ReaderDiscoverCommunication.isRequestMore(): Boolean =
```
malinajirka: This is works as well. However, we might want to consider refactoring the ReaderDiscoverCommunication and use it.task == REQUEST_MORE. Wdyt?

```
sealed class ReaderDiscoverCommunication {
    abstract val task: DiscoverTasks
    data class Started(override val task: DiscoverTasks) : ReaderDiscoverCommunication()
    data class Success(override val task: DiscoverTasks) : ReaderDiscoverCommunication()
    sealed class Error : ReaderDiscoverCommunication() {
        data class NetworkUnavailable(override val task: DiscoverTasks) : Error()
        data class RemoteRequestFailure(override val task: DiscoverTasks) : Error()
    }
}
```

malinajirka (approved):
Thanks @zwarm! The changes look good and the app works as expected. I've left one minor suggestion but I'm approving this PR. Feel free to merge it.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/ReaderRepositoryEvents.kt
zwarm: @malinajirka - Yes!! I thought we weren't touching this until our refactoring chat, but I'll consider your comment the go-ahead, so I'm implementing now. :) Thanks. In commit 39a833a

malinajirka:
Awesome, thanks! Feel free to merge it when the CI passes.

PR #12812, opened 2020-08-27
[Platform] Generate PlayStore Screenshots on CI

AliSoftware:
Fixes main part of #12601 

⚠️  This PR must NOT be merged as is, as it contains patches that I had to apply in order to make it work on my fork.
⚠️  ~This PR is built on top of #12831 and should only be merged after that one.~ [EDIT: #12831 is now merged]

## Scope

This PR implements the main part of #12601, i.e. making the WPAndroid raw and promo screenshots run on CI and Firebase Test Lab.

### Content of this PR

* Update the UI Tests to use FTL's ScreenShotter library in addition of still supporting `screengrab`.
* Add a `generate_screenshots` parameter to the `.circleci/config.yml` to trigger the `Screenshots` workflow on demand
* This workflow then starts a `raw-screenshots` job on a Linux executor (needed to use `gcloud`, `gsutil` and FTL)
   * This job builds the project then push it to Firebase Test Lab to run the UI Tests on a matrix of devices and locales.
   * Those UI Tests generate the raw screenshots and store them in a Firebase Cloud Storage bucket
   * The job then downloads the generated raw screenshots from FCS, then persists the bare minimum (`Gemfile`, `Gemfile.lock` and `fastlane/` folder) into the CI workspace.
* The workflow then subsequently runs a `promo-screenshots` job on a Mac executor (macOS needed to use `rmagick` and `drawText`)
   * This job restores the workspace saved by `raw-screenshots`. This avoids doing a full git clone of the source code while all we need is `bundler`, `fastlane`, and the raw screenshots in `fastlane/screenshots/raw`, not the source code
   * Then it installs ImageMagick and rmagick
   * Finally it updates the promo strings from GlotPress, Create the Promo Screenshots from the raw screenshots and `playstoreres` resources, and ZIP the result

The PR also adds multiple `fastlane` lanes:

* One to trigger the pipeline on CI from the CLI, intended to be the one you call during release cut
* One to download the raw screenshots from FCS and the `gs://` bucket, and reorganise them in a decent folder hierarchy. This one is mainly used by CI itself, but could also be triggered manually for debugging
* Changes in the `create_promo_screenshots` to convert the raw screenshots from JPG (format used by FTL) to PNG (format expected by fastlane) and rename them appropriately. (That lane still continues to work with screengrab-generated raw screenshots too)

## ⚠️ PR Dependency

~This PR is build on top of #12831 (fix for the crash in `ar` locale and stubs). It should only be merged once #12831 is merged.~

[...]

PR #12831, opened 2020-08-29
[UITests] Fix crash with Arabic locale

AliSoftware:
Fixes parts of #12601

## Context

When running the UI Tests used to generate the screenshots (`WPScreenshotTest`), the tests crashes when using the `ar` locale (Arabic).

This is due to the mocks returned by WireMock, which generate an unparsable date, throwing an exception when decoding the JSON from those stubs 

### Root cause of the issue

We use WireMock's "templating" features to define our mocks, allowing us to inject dynamic dates. For that, we use [the {{now}} helper](http://wiremock.org/docs/response-templating/#date-and-time-helpers) in our mocks like [in this template mock here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/libs/mocks/WordPressMocks/src/main/assets/mocks/mappings/wpcom/stats/stats_top-posts.json#L9)

The issue with this is that WireMock uses the **current user locale** to render the date when transforming this `{{now}}` tag. This means that in any locale using Roman alphabet/dates/number system this may work, but for locales like arabic, the date once rendered by WireMock's `{{now}}` tag looks like some garbage text (character encoding also getting in the way of interpreting those valid arabic characters strangely), which then generates an `Unparsable date` exception while trying to decode the mocked JSON.

### Fixing the locale issue in mocks

This was a hard one to track. Ultimately, the solution boiled down to:

 * Registering a new `{{fnow}}` tag in `WireMockRule` that will do something similar to the `{{now}}` official tag, but use the `en_US_POSIX` locale instead
   * That work had already been started by @jkmassel a while ago; but the tag never got used in the stubs after all, and the parts of the implementation from back then that were still present in the codebase actually crashed WireMock with a `ClassCastException`.
   * I thus had to update the method signature in `UnlocalizedDateHelper` (implementation for our custom `{{fnow}}`  tag) to take an `Object` for the context instead of a `Date`. That's because `{{fnow}}` tag doesn't take a `Date` input but instead generates an output value independent of any input, so there's no reason we should expect the input context to be a `Date`, and in fact we won't even use the template's context as we won't need any input.
   * After that fix I was able to finally make a `{{fnow}}` tag in a JSON template trigger that `apply(…)` method… But then the method itself crashed; this was because we were trying to build a `Locale` using the code `"US"`, while that's not a valid code for a locale
    * Besides, as per ISO standards, the right locale to use when parsing "internet dates" (aka locale-agnostic dates) is **`en_US_POSIX`**, not `en_US` anyway.

Once I had the `{{fnow}}` registered and its implementation fixed, I was able to change all the instances of `{{now…` in the mock templates into `{{fnow…` instead, finally fixing the core of the stubs issue with non-roman (and arabic in particular) locales

_(Once again, the longest debugging sessions often lead to the smallest fixes… 😅)_

## Bonus: Fixing a small point forgotten in past locale-related PR

When I implemented the PR to fix the Indonesian locale code used for screenshots, I forgot to rename the `fastlane/playstoreres/id` folder to `in`. I realised that when testing other locales to validate the current PR, so I thought it would be worth including there too.

This change is not strictly mandatory as the proper resources will be downloaded in `fastlane/playstoreres/in` next time you run `download_promo_strings` anyway, but I thought it was cleaner to have the right folder name committed into git to avoid keeping any legacy naming around.

## To test:

* Launch an emulator in AndroidStudio, switch the locale to Arabic, then run the `WPScreenshotTest` in AndroidStudio
* Alternatively, edit the `WPScreenshotTest.java` file on line 42 to change `LOCALE_TEST_RULE = new WPLocaleTestRule()` to `LOCALE_TEST_RULE = new WPLocaleTestRule("ar", "en")`, then run the test
* Verify that the test doesn't crash anymore at the end and that the parsing of the dates from the stub don't fail with an exception


AliSoftware:
\cc @loremattei for review

Note: once/if that PR gets approved, I'll likely rebase #12812 on top of it once merged

AliSoftware:
Note: We need confirmation that the mocks are not a `git subtree` of `WordPressMocks` anymore and that those mocks are not shared with WPiOS anymore – contrary to what I found in [the e2e test README](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/androidTest/java/org/wordpress/android/e2e/README.md).

 * if they are still shared, that means that using `{{fnow}}` in those stubs will likely not work with WPiOS until we integrate a similar `fnow` extension to the standalone version of WireMock used by WPiOS, and that we'll then need to make a PR upstream in WordPressMocks repo to report those changes
 * if they are not shared with WPiOS anymore, as @jkmassel is thinking and has told me in Slack thread, then this won't be an issue. But then we might want to update the README.

[EDIT] Lorenzo confirmed to me in Slack that the mocks are not shared between iOS and Android anymore, so there's no risk of breaking WPiOS with those changes.

libs/mocks/WordPressMocks/src/main/assets/mocks/mappings/wpcom/stats/stats_top-posts.json
```diff
@@ -6,17 +6,17 @@
...
+                "{{fnow format='yyyy-MM-dd'}}": {
                     "postviews": [
                         {
                             "id": 0,
                             "href": "https://e2eflowtestingmobile.wordpress.com/",
                             "date": null,
                             "title": "Home page / Archives",
                             "type": "homepage",
-                            "views": 183,
+                            "views": 1832,
```
loremattei: Why this change?

AliSoftware: Whoops. This is a leftover from my debugging session, when I was trying to ensure that WireMock indeed used the `.json` file I thought it was using. Should be reverted indeed, good catch.

loremattei (approved):

PR #12835, opened 2020-08-31
PR to test #12681

mokagio:


WordPress/build.gradle
```diff
@@ -57,7 +57,7 @@ android {
         } else {
             versionName "alpha-237"
         }
-        versionCode 903
+        versionCode 902000
```
loremattei: This doesn't work here. We need a different versionCode for the Zalpha than the one used by the Beta (Vanilla). If we use the same code, it will fail when the second build is uploaded to the store.

PR #12838, opened 2020-08-31
Fix crash on reader discover tab

### LINKED ISSUE #12837: Reader: New Discover crashes on old Android SDK
### Expected behavior
The app doesn't crash after pull-to-refresh on discover tab on API 22.

### Actual behavior
The app crashes after the pull-to-refresh request completes on API 22

1. Enable RI FF
2. Open Discover tab in Reader
3. Pull to refresh
4. Notice the app crashes when the request completes

Android API 22

malinajirka:
Fixes https://github.com/wordpress-mobile/WordPress-Android/issues/12837

The issue was that the `companion` parameter was specified as non-null. However, on older API versions (pre-jobService) we don't use this attribute and we intentionally send `null` there.

To test:
1. Enable RI FF
2. Open discover tab
3. Pull to refresh
4. Notice the app doesn't crash and the content refreshes

PR submission checklist:


zwarm (approved):
Looks good.
Was able to generate the crash using current branch, then retested using this PR APK and all is good. 👍

PR #12855, opened 2020-09-01
Testing: Remove references to WordPressMocks repo

rachelmcr:
We are no longer trying to keep the [WordPressMocks](https://github.com/wordpress-mobile/WordPressMocks) repo updated with changes made on Android, since the API requests on the two platforms aren't identical and it's hard to keep the mocks in a state that works well on both platforms. This PR removes references to the WordPressMocks repo in the docs, to reflect that change.

Internal ref: paaHJt-PH-p2

PR submission checklist:


loremattei (approved):
Many thanks for this @rachelmcr!

PR #12856, opened 2020-09-01
RI: Get user tags only from db if size > 0

ashiagr:
This PR 

1. updates`ReaderTagRepository.getUserTags()` to return tags only from the db if count > 0 - d1e5582

    If there are already followed tags in the db, we can skip fetching user tags from the server (and avoid unnecessary full screen error message due to n/w error) as the new discover feed can be populated based on these tags.

2. deletes recommended tags added as followed tags for logged out users inside `GetFollowedTagsUseCase.get()` - fd6c575

    This is done so that they don't cause issues in calculating followed tags count during display of "Choose Interest" screen. 
    More context [here](https://github.com/wordpress-mobile/WordPress-Android/pull/12530#discussion_r462794836).

3.  fixes a loading issue in the discover feed - ef28a4e

#### To Test

Enable RI FF 

1. Launch app and login with account with followed tags
2. Enable airplane mode
3. Go to Reader -> Discover tab 
4. Notice that discover feed is shown and not blocked by any full screen error message

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/datasets/ReaderTagTable.java
```diff
@@ -363,4 +372,31 @@ private static int minutesSinceLastUpdate(ReaderTag tag) {
...
+        if (tagList == null || tagList.size() == 0) {
+            return;
+        }
+        deleteTags(tagList);
+
+        SQLiteDatabase db = ReaderDatabase.getWritableDb();
+        db.beginTransaction();
+        try {
+            for (ReaderTag tag : tagList) {
+                ReaderPostTable.deletePostsWithTag(tag);
```
malinajirka: Do we really need to delete the posts here? It feels a bit weird to access one table from another. I think "dangling" posts will get deleted automatically [here](https://github.com/wordpress-mobile/WordPress-Android/blob/f7760b32bd2cc702618858e1b941c2d92089505d/WordPress/src/main/java/org/wordpress/android/datasets/ReaderPostTable.java#L215). Wdyt?

WordPress/src/main/java/org/wordpress/android/datasets/ReaderTagTable.java
```diff
@@ -363,4 +372,31 @@ private static int minutesSinceLastUpdate(ReaderTag tag) {
         Date dtNow = new Date();
         return DateTimeUtils.minutesBetween(dtUpdated, dtNow);
     }
+
+    public static void deleteRecommendedTagsAddedAsFollowedTags() {
```
malinajirka: I'm wondering if we could delete the out-dated data in a migration script (when we get rid of the FF tomorrow). We could drop the whole tag table when it comes to that - it's just a cache anyway. This code will be here forever and every user will use it once at most. Wdyt? (I realize it technically wasn't introduced in this PR, but imo it's still worth considering)

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/tags/GetFollowedTagsUseCase.kt
```diff
@@ -3,16 +3,23 @@ package org.wordpress.android.ui.reader.repository.usecases.tags
...
 
 class GetFollowedTagsUseCase @Inject constructor(
     @Named(IO_THREAD) private val ioDispatcher: CoroutineDispatcher,
-    private val readerTagTableWrapper: ReaderTagTableWrapper
+    private val readerTagTableWrapper: ReaderTagTableWrapper,
+    private val accountStore: AccountStore,
+    private val appPrefsWrapper: AppPrefsWrapper
 ) {
     suspend fun get(): ReaderTagList = withContext(ioDispatcher) {
+        if (!accountStore.hasAccessToken() && appPrefsWrapper.isReaderImprovementsPhase2Enabled()) {
```
malinajirka: It feels a bit out of place here - we can discuss it further based on the above comments. Ignore this comment for now.

malinajirka (commented):
Thanks @ashiagr! I've done a first review pass. Let me know what you think.

WordPress/src/main/java/org/wordpress/android/datasets/ReaderTagTable.java
ashiagr: Yep, we can include it in migration script. Wasn't sure how to migrate with RI FF enabled.

WordPress/src/main/java/org/wordpress/android/datasets/ReaderTagTable.java
ashiagr: It was an existing method that I moved from `ReaderUpdateLogic` to the `ReaderTagTable`.  I suppose posts deletion can be handled at the line you linked.

WordPress/src/main/java/org/wordpress/android/datasets/ReaderTagTable.java
malinajirka: yeah, I agree that'd be quite tricky with a FF. Feel free to create the migration script ignoring the FF - we'll delete the tags for everyone no matter the FF. I'll create a PR for removing the FF tomorrow anyway. Thanks 🙇

WordPress/src/main/java/org/wordpress/android/datasets/ReaderTagTable.java
malinajirka: If we remove the tags in the migration script I believe we won't need this method, right?

WordPress/src/main/java/org/wordpress/android/datasets/ReaderTagTable.java
ashiagr: Deleted tags for everyone in f493d12. 
Removed now unused `deleteRecommendedTagsAddedAsFollowedTags()` in 8813d3e.

WordPress/src/main/java/org/wordpress/android/datasets/ReaderTagTable.java
ashiagr: Removed this method and used `deleteTags() `directly when local tags are synced with the server in in 5c6f8e3.

malinajirka (approved):
I've tested the migration with both a self-hosted and .com sites. Everything works as expected. Thanks for all the work Ashita;)!

PR #12862, opened 2020-09-02
Fix incorrect bundle location in ci attach_workspace command

cameronvoell:
> Originally when the “attach_workspace” work was put in place, the path was indeed used as the output folder for the JS bundle to be created and shared among the jobs in the CircleCI workflow. See [here the gradle code](https://github.com/wordpress-mobile/gutenberg-mobile/blob/fa6fcc0d6de8b14e506ab3b66a21d6ff834621aa/react-native-gutenberg-bridge/android/build.gradle#L165) that was setting that folder. But nowadays, the [output folder is different](https://github.com/WordPress/gutenberg/blob/33c83fa7bd917704cf23cfc7106598c11fd0c6e5/packages/react-native-bridge/android/build.gradle#L94). 

This PR is for correcting the output folder in CI to be consistent with our `buildAssetsFolder` defined in the [react-native-gutenberg-bridge npm package](https://github.com/WordPress/gutenberg/blob/33c83fa7bd917704cf23cfc7106598c11fd0c6e5/packages/react-native-bridge/android/build.gradle#L94). 

To test:

PR submission checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,9 +114,9 @@ jobs:
       - restore-gutenberg-bundle-cache
       - run:
           name: Ensure assets folder exists
-          command: mkdir -p libs/gutenberg-mobile/gutenberg/packages/react-native-bridge/android/src/main/assets
+          command: mkdir -p libs/gutenberg-mobile/gutenberg/packages/react-native-bridge/android/build/assets
       - attach_workspace:
-          at: libs/gutenberg-mobile/gutenberg/packages/react-native-bridge/android/src/main/assets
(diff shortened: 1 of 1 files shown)


cameronvoell:
@loremattei As I outlined in the description, this change looks like it is definitely correcting an error in assets folder location, as @hypest pointed out earlier (and outlined in the description here). It seems good that CI passed, but I'm not sure if I can create the conditions where you were seeing the fail earlier to verify that this will fix the missing bundle that was seen on CI elsewhere.

loremattei (approved):
Thank you for this fix @cameronvoell!

I verified that things look good now on CI.

:shipit:

PR #12867, opened 2020-09-02
Add more types to media picker

### LINKED ISSUE #12636: Enhance DeviceMediaListBuilder to support multiple file types
The DeviceListBuilder functionality is now very basic. As part of this ticket you should investigate how to load all the required files (Images, Video, Audio, Documents) and how to filter out unsupported MimeTypes (based on the MimeTypes class from FluxC). You should create a new parameter class for this filtering that should be used instead of MediaBrowserType.

planarvoid:
Fixes #12636 

This PR adds an option to pick Audio files and documents to the Media picker. The audio files are selected from the content provider. The documents are from the Downloads folder. The media picker should only show supported types. I've replaced the "Choose from device" option in the "Media Browser activity" with the media picker as part of this PR.

I've removed ID as the main item identifier in this PR and replaced it with URI. The reason is that the documents are not loaded through content provider and don't have IDs. 

To test:
- Enable the flag
- Go to My Site/Media
- Click on +/Choose from device
- Notice the media picker is shown
- Notice you see Audio files and documents in the list, ordered by the change date
- Notice that the selector sometimes overlaps the file name -> that will be fixed later
- Select a few files
- Confirm selection
- Files get uploaded

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/util/MediaUtilsWrapper.kt
```diff
@@ -47,4 +47,6 @@ class MediaUtilsWrapper @Inject constructor(private val appContext: Context) {
             EditorMediaUtils.getVideoThumbnail(appContext, videoPath, headers)
 
     fun isLocalFile(uploadState: String): Boolean = MediaUtils.isLocalFile(uploadState)
+
+    fun getExtensionForMimeType(mimeType: String?): String? = MediaUtils.getExtensionForMimeType(mimeType)
```
develric: Wonder if we should/could make the String return value not nullable maybe 🤔

develric (commented):
Great job @planarvoid 👍 !

Tested with API 21 and 29 and works well! Code looks also good, I left an inline minor comment in case it could make sense.

A couple of things I noticed are (and I think we can also address them eventually in other PR(s)):

- We already hide the edit icon from the appbar when we select videos in a group of multiple selected items, I would evaluate to do the same in general if not all selected items are images, wdyt?

- Sometimes I get the following: if I load a video from device, in the final upload phase a thumbnail is shown that is not taken from a preview of the video, but it seems taken from another of the images that are already loaded in the Media Lib; then after a short time from the end of upload the actual and correct video thumbnail is shown (I do not have steps to reproduce consistently but it happened a few times in my emulator). I don't think this is related to this PR tbh, so we can further investigate/address in another PR, but I preferred to mention so we do not overlook it (let me know if you want me to open a ticket for this 🙇‍♂️ ).

planarvoid:

> We already hide the edit icon from the appbar when we select videos in a group of multiple selected items, I would evaluate to do the same in general if not all selected items are images, wdyt?

Good point! I've fixed that in ec066e2 with some tests

> Sometimes I get the following: if I load a video from device, in the final upload phase a thumbnail is shown that is not taken from a preview of the video, but it seems taken from another of the images that are already loaded in the Media Lib; then after a short time from the end of upload the actual and correct video thumbnail is shown (I do not have steps to reproduce consistently but it happened a few times in my emulator). I don't think this is related to this PR tbh, so we can further investigate/address in another PR, but I preferred to mention so we do not overlook it (let me know if you want me to open a ticket for this 🙇‍♂️ ).

I think this happens in the Media browser activity (and the media grid adapter) because we don't correctly clear the previous item image in the recycler view. It's not related to this project though so maybe create an issue for it? Maybe we'll replace that screen with our new adapter. It doesn't make sense to have 2 screens that look the same and are in two different places (maybe we can just reuse the adapter).

develric (approved):
Thanks for the changes @planarvoid , looks good and works well 👍!

I opened a ticker #12867 for the issue we discussed about the wrong thumbnail on video upload that sometimes can be shown and I agree it is not related to this PR.

PR #12872, opened 2020-09-03
Refactor PostLikeUseCase

zwarm:
Parent issue #12028

This PR refactors the PostLikeUseCase
- to take advantage of flow
- integrated into the legacy ReaderPostListFragment. 

Implementation Details
- ReaderPostActions
    - Added two new Like action method, basically separates the single method into two. 
        `performLikeActionLocal`  and `performActionRemote`.   
        `ReaderDetailFragment` currently uses `performLikeAction()`, so the existing method calls the two new methods. We will address ReaderDetailFragment in a future phase of the RI project.
    - Removed ReaderEvents that were no longer needed
- ReaderPostActionsWrapper
    - Add wrappers around the two new ReaderPostActions methods listed above
- ReaderPostListViewModel
    - Add `onLikeButtonClicked()` to send the Like click through `ReaderPostActionHandler`
- ReaderPostAdapter
    - Removed toggleLike method
    - Use `mOnPostListItemButtonListener` for Like clicks
- Updated `ReaderPostCardsActionHandler`
    - Uses `PostLikeState`  instead of ReaderRepositoryCommunication
    - Include handling UpdateInLocalDB and posting to _refreshPosts
- PostLikeUseCase
    - Changed to use flows so we can handle localDb updates and refresh legacy reader tabs
    - Replaced ReaderRepositoryCommunication with PostLikeState 
    - Removed references to eventbus
- ReaderPostListFragment
    - Hooked up the Like case in `onButtonClicked`
- ReaderRepositoryEvent
    - Removed unused events
- ReaderLikeUseCaseTest
    - Fixed tests

**NOTE**: This PR does not include any PostLikeUseCase tests (I commented out existing ones), as they need to be rewritten to work with flow. The tests will follow behind shortly after I research how to test with flows.

To test:
1. Open the app with FF on
2. Navigate to the following tab
3. Like a post
4. The icon color will change and the count goes up
5. Unlike a post
6. The icon color will no longer be colored and the count goes down
7. Navigate to the Discover tab
8. Repeat steps 3-6

Regression test post detail
1. Open the app with FF on
2. Navigate to post detail
3. Like a post
4. The icon color will change and the count goes up
5. Unlike a post
6. The icon color will no longer be colored and the count goes down
PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeUseCase.kt
```diff
@@ -29,86 +24,101 @@ import kotlin.coroutines.Continuation
...
 
-    suspend fun perform(
-        post: ReaderPost,
-        isAskingToLike: Boolean
-    ): ReaderRepositoryCommunication {
-        return withContext(bgDispatcher) {
-            val wpComUserId = accountStore.account.userId
-            val request = PostLikeRequest(post.postId, post.blogId, isAskingToLike, wpComUserId)
+        if (!networkUtilsWrapper.isNetworkAvailable()) {
+            emit(Failed.NoNetwork(request))
```
malinajirka: Don't we need to use "return" or "else" branch here? Same goes for `continuations[request] != null`. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeUseCase.kt
```diff
@@ -29,86 +24,101 @@ import kotlin.coroutines.Continuation
...
     private val continuations:
-            MutableMap<PostLikeRequest, Continuation<ReaderRepositoryCommunication>?> = mutableMapOf()
+            MutableMap<PostLikeRequest, Continuation<PostLikeState>?> = mutableMapOf()
 
-    init {
-        eventBusWrapper.register(this)
-    }
+    suspend fun perform(post: ReaderPost, isAskingToLike: Boolean) = flow<PostLikeState> {
+        val wpComUserId = accountStore.account.userId
+        val request = PostLikeRequest(post.postId, post.blogId, isAskingToLike, wpComUserId)
```
malinajirka, line 38: We create a new flow for each caller of `perform(...)` -> do we need to send the PostListRequest back to the caller? The caller knows which parameters they provided to the use case. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeUseCase.kt
```diff
@@ -29,86 +24,101 @@ import kotlin.coroutines.Continuation
...
+            readerPostActionsWrapper.performLikeActionRemote(
+                    post,
+                    request.isAskingToLike,
+                    request.wpComUserId,
+                    actionListener
+            )
+        }
     }
 
     fun stop() {
```
malinajirka: It seems this method is not used. Moreover, I think it's a bit misleading - calling "clear" won't stop anything, it'll just "deadlock" a few coroutines as no-one ever resumes them. Wdyt?

malinajirka (commented):
Thanks for working on this @zwarm! I've found a couple of potential issues - let me know what you think.

Regarding the tests - you can just call `toList(mutableListOf())` instead of `collect`. You can check [these tests ](https://github.com/wordpress-mobile/WordPress-Android/blob/0d2b82a24d5942b820f4a208e6acd9f6423874e8/WordPress/src/test/java/org/wordpress/android/ui/reader/usecases/ReaderPostBookmarkUseCaseTest.kt#L27).

Btw lint and the tests are failing.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeUseCase.kt
zwarm: > It seems this method is not used. Moreover, I think it's a bit misleading - calling "clear" won't stop anything, it'll just "deadlock" a few coroutines as no-one ever resumes them. Wdyt?

Makes sense. Removed in commit ccfcc11

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeUseCase.kt
zwarm: > We create a new flow for each caller of perform(...) -> do we need to send the PostListRequest back to the caller? The caller knows which parameters they provided to the use case. Wdyt?

Good call. Gone in commit 20a66ac

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeUseCase.kt
zwarm: > Don't we need to use "return" or "else" branch here? Same goes for continuations[request] != null. Wdyt?

Added returns in commit 20a66ac

malinajirka (approved):
Perfect, thanks @zwarm! It works like a charm ;)

PR #12874, opened 2020-09-03
Fixing magiclink backstack issue when passwordless

khaykov:
@renanferrari As we discussed, I did quite a bit of research on the problem and tried a couple of different approaches (back stack listener, adding "artificial" back stack entries). I think that poping back stack when navigating to the "magic link sent" fragment is the least hacky, and relatively smooth for the user. For a brief moment between transitions, you can see the email input fragment appearing, but if you are not looking for it it's hard to tell.

It's a bit of a shame there is no "appropriate" way to do this, but maybe, as you mentioned, we will be able to do it with the new Navigation component at some point.

To test:
- Using the passwordless account, input email, continue and notice that magic link is getting automatically requested, and you are transferred to email check fragment. Press back/up button and confirm that you are back in the email input fragment.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
@@ -565,6 +565,10 @@ public class LoginActivity extends LocaleAwareActivity implements ConnectionCall
 
     @Override
     public void showMagicLinkSentScreen(String email, boolean allowPassword) {
+        if (!allowPassword) {
+            getSupportFragmentManager().popBackStackImmediate();
+        }
+
         LoginMagicLinkSentFragment loginMagicLinkSentFragment =
                 LoginMagicLinkSentFragment.newInstance(email, allowPassword);
         slideInFragment(loginMagicLinkSentFragment, true, LoginMagicLinkSentFragment.TAG);
--- a/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
@@ -564,10 +564,6 @@ public class LoginActivity extends LocaleAwareActivity implements ConnectionCall
 
     @Override
     public void showMagicLinkSentScreen(String email, boolean allowPassword) {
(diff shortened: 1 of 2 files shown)


WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
```diff
@@ -565,6 +564,10 @@ public void loginViaWpcomUsernameInstead() {
 
     @Override
     public void showMagicLinkSentScreen(String email, boolean allowPassword) {
+        if (!allowPassword) {
+            getSupportFragmentManager().popBackStackImmediate();
+        }
```
renanferrari: We should move this logic to `LoginMagicLinkRequestFragment` and replace `!allowPassword` with `mForceRequestAtStart`. This flag indicates if we're skipping the screen or not, so it's the best one to use in our case.

Then where we now have:

```java
if (mLoginListener != null) {
    mLoginListener.showMagicLinkSentScreen(mEmail, mAllowPassword);
}
```

 Should then be something like:

```java
if (mLoginListener != null) {
    if (mForceRequestAtStart) {
        getFragmentManager().popBackStackImmediate();
    }
    mLoginListener.showMagicLinkSentScreen(mEmail, mAllowPassword);
}
```

I'd probably also add a comment there, as it's not very obvious what's going on. Something like: 
`// We don't want the user to go back to this screen if they press back on the next one.`

renanferrari (commented):
Thanks a lot for the help with this, I really appreciate it! I tested the changes and `popBackStackImmediate()` really seems like the way to go. I didn't even notice the transition going back to the previous screen as you mentioned.

I did found an issue with this other scenario, though:
1. Log in using an email that has a password set.
1. On the Password screen, tap "Get a login link by email".
1. On the Magic Link Sent screen, press back.
1. Notice how we get back to the Magic Link Requested screen again, instead of getting back to the Password screen.

The fix should be pretty simple though, as I explain below.

khaykov:
@renanferrari Totally missed the case with regular magic link request for accounts with password :) Fixed it now!

renanferrari:
Thanks again for the help with this! LGTM :shipit:

PR #12878, opened 2020-09-03
Update media picker selector UI

### LINKED ISSUE #12654: Use the media picker component in Aztec
Use the new component in Aztec
- make sure everything works as expected
- make sure it's hidden behind a flag

planarvoid:
Fixes #12654 

This PR updates part of the new media picker UI - the selectors - based on the designs.
These are the new states: 
Single selection:
Screenshot 2020-09-03 at 15 23 42
Multi selection:
Screenshot 2020-09-03 at 15 23 16
 

To test:
- Go to Media library/+/Choose from device
- Try selecting a few things 
- Confirm the UI looks OK

To test 2:
- Go to GB editor/Image block/Add media/Choose from device
- Try selecting an item
- Confirm the UI looks OK 

PR submission checklist:


WordPress/src/main/res/values/dimens.xml
```diff
@@ -51,6 +51,7 @@
     <dimen name="media_settings_margin_tablet_large">96dp</dimen>
     <dimen name="media_settings_margin">@dimen/media_settings_margin_normal</dimen>
     <dimen name="media_selection_dot">24dp</dimen>
+    <dimen name="media_picker_selection_dot">20dp</dimen>
```
develric, line 54: Sorry if it's obvious; just to be sure it's not an overlook I was wondering why we reduced from 24dp to 20dp, is it by design?

WordPress/src/main/res/layout/media_picker_file_item.xml
```diff
@@ -60,22 +60,6 @@
...
     </RelativeLayout>
 
-    <com.google.android.material.textview.MaterialTextView
-        android:id="@+id/text_selection_count"
-        android:layout_width="@dimen/media_selection_dot"
-        android:layout_height="@dimen/media_selection_dot"
-        android:layout_marginBottom="@dimen/margin_medium"
-        android:layout_marginStart="@dimen/margin_medium"
-        android:background="@drawable/photo_picker_circle_selector"
-        android:elevation="4dp"
```
develric, line 70: Again just to be sure it was intentional, I see we removed the elevation in the `media_picker_selection_count.xml` (and eventually we keep the elevation it would be probably worth using a dimension for the 4dp; I'm sure you thought the same already right? 😄 )

develric (commented):
Hi @planarvoid , tested it and it works well 👍! I left a couple of comments inline with the code in case you think they can be worth. Also I just noticed there is a minor linter issue to fix. Let me know, thanks 🙇

WordPress/src/main/res/values/dimens.xml
planarvoid: yep, this is intentional from the design

WordPress/src/main/res/layout/media_picker_file_item.xml
planarvoid: I've removed the elevation because (and this is really hard to notice) it was adding this weird effect of a white shape in the middle of the selector. Is there any reason why we should keep it?

develric: > Is there any reason why we should keep it?

Just wanted to be sure it was intentional, I see the effect you mention especially with documents items where we have a light background. I'm fine with removing it, thanks for the clarification 🙇

develric (approved):
Thanks for the follow up comments @planarvoid, LGTM 👍

Issue #12854, opened 2020-09-01
Including a button to browse the files from the device in the Media Picker.


develric:
Even with the upcoming Media Picker Consolidation project we should still include a button to browse the files on the device.

We can place it in the right-hand position on the top appbar of the picker using the gridicons-image.svg icon for this since this matches with what can be found currently for selecting an image from the device.

designsimply:
Do you have a design file you could include with this issue?

Here are some current screenshots of the media picker in the Media and Editor screens taken with 15.6 beta:

Self-hosted Media|Self-hosted Editor|Jetpack Media|Jetpack Editor|Media > Choose from device|Editor > Choose from device
---|---|---|---|---|---
![Screenshot_20200903-130016](https://user-images.githubusercontent.com/1119271/92156461-3d01a380-ede6-11ea-8023-b7983fbd7450.png)|![Screenshot_20200903-130037](https://user-images.githubusercontent.com/1119271/92156458-3b37e000-ede6-11ea-85fb-8664ae79135e.png)|![Screenshot_20200903-125536](https://user-images.githubusercontent.com/1119271/92156463-3e32d080-ede6-11ea-8794-02ae44d588ed.png)|![Screenshot_20200903-125612](https://user-images.githubusercontent.com/1119271/92156462-3d9a3a00-ede6-11ea-8c9d-7b92b6cdf553.png)|![Screenshot_20200903-131225](https://user-images.githubusercontent.com/1119271/92157698-062c8d00-ede8-11ea-985e-912871f4461c.png)|![Screenshot_20200903-131332](https://user-images.githubusercontent.com/1119271/92157777-23615b80-ede8-11ea-9789-0eaee2d075a3.png)


<p style="clear:both;"><sup>Tested with WPAndroid 15.6-rc-2 on Pixel 3 Android 10 on sites running WordPress 5.5.1 Jetpack 8.9.</sup></p>

develric:
Hey @designsimply thanks for the above 🙇‍♂️; I started opening a PR #12884 related to this hopefully to ease some discussion about it. We are thinkering with @planarvoid already a bit and we will soon ping @SylvesterWilmott to have his input design wise (currently we have this comment thread that triggered the task pbArwn-nS-p2#comment-1556)

PR #12884, opened 2020-09-03
Browse for device items in media picker

develric:
Part of the fix for #12854 

This PR adds a menu in media picker that can be used to open the system browser for IMAGE and/or VIDEO. This tries to use an approach similar to what we were using before in the photo picker bottom bar.

NOTE: current PR takes care of this menu when adding media from GB blocks. A dedicated PR will address the behavior of that menu when using the picker in Media Library. Note also that the menu is actually presented to the user only when using the MEDIA & TEXT block that is the only case AFAIK that leave the user the possibility to add a VIDEO or and IMAGE. In all other cases VIDEO/IMAGE/GALLERY, the menu is not presented and the videos or images are directly filtered when browsing for items in device.

| image | image |
|---|---|

Some notes:
- The browse for items visibility/usage need to be reviewed when using  picker to browse for media in the Media Library. We will need to review logic based on the possibility to upload any file from there. This will be done in a different PR.
- ~~Not fully sure of the why but the safe cast (as?) for site model is not still present in develop for `MediaPickerActivity` but it's there in the above mentioned #12867 🤔 ; still all behind feature flag so no problems in production of course but this means that trying to set avatar icon will fail until merging #12867~~. #12867 has been merged into this one, so this point should be fine now.
- I'm reviewing tracking in MediaPickerViewModel.clickIcon in [another PR](https://github.com/wordpress-mobile/WordPress-Android/pull/12889)

## To test

Enable the media picker feature flag and run the app

### Site Icon

- Try to set the Site Icon and check the Picker is opened with Browse for items menu
- Check that selecting an image activate the action mode bar hiding the Browse for items menu
- Check that unselecting the image hides the action mode bar showing the Browse for items menu again
- Tap on the Browse for items menu and check all non IMAGES (or not supported images) files are grayed out 
- Check you can select an image and it is used as site icon

### GB post

- Create a new post with GB
- Place a VIDEO/IMAGE/GALLERY/MEDIA & TEXT blocks
- Try to set the above block content in the picker from the Browse for items menu
- Note that only for the `MEDIA & TEXT` block both IMAGE and VIDEO are presented not grayed in the system file browser
- Smoke test

## Story

- Create a new story
- Check you can set story slides from the Browse for items menu
- Smoke Test adding video/image from the Browse for items menu and from Media Picker

PR submission checklist:


planarvoid:
Hey @osullivanchris , I was wondering if we could get feedback on this menu from you :-). The way it currently behaves (that's the old behaviour) you had either images or videos showing in the list. If we were showing only one of the types, the action in the action bar took you directly to the system picker. If you had both types, we showed the dropdown menu with "Choose photo from device" and "Choose video from device". 
2 things have changed, now we have 4 types (images, videos, audios, documents available) and these 2 actions were replaced by one "Choose from device" in the media library. 
Because of this I'm suggesting we don't use a dropdown menu for the types and go directly to the system picker that filters out the types based on what we are showing in the media picker, for example:
- In gutenberg image block - the media picker shows you only images, the button in the toolbar will take you to the system picker with only images
- In stories - the media picker shows you images and videos, the button in the toolbar will take you to the system picker with images and videos
- In the Media library -> Choose from device option -> the media picker shows you all the types so the button in the toobar takes you to the system picker with the supported types
Does this sound ok to you? - [this is the issue for this change](https://github.com/wordpress-mobile/WordPress-Android/issues/12854#issuecomment-687092890)

Bear in mind that there will be cases where we'll show the popup menu. In stories we'll have the "Choose from stock photos" option in the menu.

osullivanchris:
@planarvoid hey! If I understand correctly, you're showing that the small menu is redundant in most cases because the user doesn't need to decide between photo and video at that point. Its either already decided because of the source, or they can choose from any type anyway.

So yeah, generally t makes sense to me to remove the menu and bring the user directly to the picker!

However there's a lot of different details in this PR and in #12854 though which I haven't fully digested. So I can't say that I've thought through the scenarios in detail or if there's some other open question here that I'm missing. So let me know if there's more detail you need.

develric:
Hi all 👋 ! I've updated this PR based on the above feedback; now we have as follows:

1. Merged from develop so now having also search icon in picker menu; following logic added:
    - When not yet having user storage permissions (like after a fresh install) both search and browse menu icons are not shown
    - When having permissions: both are shown but browse is hidden when search is expanded
<p align="center">
  
</p>

2. We are removing menus when user need to browse the device for media and the user is directly landed to the system file browser with the correct file type filtering (image only, video only, image and video only or all files with a supported WP MIME type); this applies both when coming from media library and from GB blocks
3. I'm not managing in this PR the Free Photo Library case for stories (documented in #12914; @osullivanchris: i will open a dedicated PR for that but currently I'm planning to use the popup menu approach using `Choose from device`and `Choose from Free Photo Library`entries for this specific case, let me know if you have different advice, thanks 🙇 )

Chris if it can help I can provide an APK in slack with the correct flag activated (I think it would be good so you can check what I described above 🙇 ).

PS: as a side note not related to this PR, while working on this I also opened this issue as reminder #12915; I will investigate it further in a dedicated PR.

osullivanchris:
Hey I'm not 100% sure where this PR begins and ends (as part of the general media consolidation project) but here's a few impressions I have from the build. Probably a mix of existing things, and possible some just not addressed yet:

- not asking photo/video seems intuitive to me
- using the media type in headers (choose photo / choose media) is working well
- It feels like search cannot be used in combination with other selection. For example, if I want to add pictures of fruit, and search 'banana' I can then select images with 'banana' in the title. But then, if I tap back, my selection is lost. There's no way to search and find things, and retain the selection, and combine it with other things outside the search selection. Is this a known limitation of what we're doing? Do we think its ok?
- In general, I found search wasn't particularly useful to me because I cannot see filenames, so I don't know what to search for anyway. If we can only search by file name, a different layout where I could see the names of the photos might be more useful if we do want people to search.
- I found two minor visual issues/bugs below. In one, you can see something strange is happening with the empty state text. In the other, before it loads the images from my device it is just blank as if there are no images available. We should show a spinner or an empty skeleton state.

Screenshot_20200909-130330
Screenshot_20200909-130500

I haven't been too close to this project so far. So I hope the feedback is relevant. If there are specific parts of the experience you want me to pay close attention to, please let me know! Thanks 😄

planarvoid:
hey @osullivanchris , thanks for the review! We'll definitely have a version once we're mostly done to polish the behaviour. I'll answer the stuff about search :-) 

> It feels like search cannot be used in combination with other selection. For example, if I want to add pictures of fruit, and search 'banana' I can then select images with 'banana' in the title. But then, if I tap back, my selection is lost. There's no way to search and find things, and retain the selection, and combine it with other things outside the search selection. Is this a known limitation of what we're doing? Do we think its ok?

We've discussed this before in DM. Once you select an item, you enter the "Selection" mode. Once you search, you enter the "Search" mode. You can't jump from the selection mode to the search mode. This is how the Google apps behave. Once option how to fix this would be to separate search and selection modes in the toolbar (have 2 toolbars or something like that).

> In general, I found search wasn't particularly useful to me because I cannot see filenames, so I don't know what to search for anyway. If we can only search by file name, a different layout where I could see the names of the photos might be more useful if we do want people to search.

We can search by anything we want. Filename was the thing that made sense once implementing this feature :-). Having another layout for filtered out items feels like something we could do.

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerViewModel.kt
```diff
@@ -431,11 +460,13 @@ class MediaPickerViewModel @Inject constructor(
         object Hidden : ActionModeUiModel()
     }
 
-    sealed class SearchUiModel {
-        object Collapsed : SearchUiModel()
-        data class Expanded(val filter: String) : SearchUiModel()
+    sealed class SearchUiModel(val isVisible: Boolean) {
+        class Collapsed(isVisible: Boolean) : SearchUiModel(isVisible)
+        class Expanded(isVisible: Boolean, val filter: String) : SearchUiModel(isVisible)
```
planarvoid: I think I'd probably introduce third child state here: Collapsed, Expanded and Hidden? Does it make sense to keep the information about search when it's not visible? I'd also keep data classes just to be sure (for the equality check in the `distinct`)

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerViewModel.kt
```diff
@@ -345,6 +356,23 @@ class MediaPickerViewModel @Inject constructor(
         }
     }
 
+    fun onBrowseForItems() {
+        if (mediaPickerSetup.isBrowser()) {
+            clickIcon(ANDROID_CHOOSE_FILE)
```
planarvoid: I think this approach could be simplified now that we have only 1 icon to do all the actions. We can have one enum value (`ANDROID_CHOOSE_FROM_DEVICE`) and we can build the allowed mimetypes from the media picker setup allowed types. We can pass this enum and the set of allowed types on icon click (this means this whole logic would disappear).

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerSetup.kt
```diff
@@ -27,6 +28,8 @@ data class MediaPickerSetup(
         intent.putExtra(KEY_CAMERA_ENABLED, cameraEnabled)
     }
 
+    fun isBrowser() = dataSource == DEVICE && allowedTypes.containsAll(MediaType.values().toList())
```
planarvoid: I'd rather not have this logic here. Otherwise it will start behaving like the browser type class :-).

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerActivity.kt
```diff
@@ -231,6 +237,22 @@ class MediaPickerActivity : LocaleAwareActivity(), MediaPickerListener {
...
+    private fun launchVideoLibrary(multiSelect: Boolean) {
+        WPMediaUtils.launchVideoLibrary(this, multiSelect)
+    }
+
+    private fun launchMediaLibrary(multiSelect: Boolean) {
+        WPMediaUtils.launchMediaLibrary(this, multiSelect)
+    }
+
+    private fun launchFileLibrary(multiSelect: Boolean) {
+        WPMediaUtils.launchFileLibrary(this, multiSelect)
```
planarvoid: we can create only one method in `WPMediaUtils` and pass the set of mime types as a parameter.

planarvoid (changes requested):
Thanks for the PR! It works really well! 👍  I have a couple of suggestions about possible improvements. It seems like we don't need to differentiate between various ways to choose file from the device. We can just extract this information from the allowed types and show whatever is allowed. Let me know what you thik

develric:
Thanks @osullivanchris 🙇 

For the bug you have noticed, I think it is same as this known issue (#12915) so hopefully we will cover it fixing that one in a dedicated PR

Re my third point above
>3. I'm not managing in this PR the Free Photo Library case for stories (documented in #12914; @osullivanchris: i will open a dedicated PR for that but currently I'm planning to use the popup menu approach using Choose from device and Choose from Free Photo Library entries for this specific case, let me know if you have different advice, thanks 🙇 )

we need your input on how to allow user choice between `choose from device´ and from `Free Photo Library` (this currently should happen only in stories AFAIU, but could apply to other cases in the future); current approach would be to come back to a popup menu (like I described in the issue #12914) but @planarvoid was suggesting maybe something like a Tab filtering could be better there (and I think I agree). Wdyt? Thanks 🙇 

Hi @planarvoid thanks for the review, from a first look I think I agree with your comments. I'm going to work on it and follow up 🙇

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerViewModel.kt
develric: Good point (especially the one related to data class and `distinct` usage 👍)...changed here 931218e1dd536536d50e0c1224cd1340b6213ac4

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerSetup.kt
develric: 😊...removed here 1de6b2a5a930a8e3d6578e4e2cfbe3c573bf0345

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerViewModel.kt
develric: I tried to refactor this a bit also promoting the `MediaPickerIcon` to a sealed class and introducing a few additional types to try to abstract and separate things a bit (MediaPickerIconType, ChooserContext and MediaPickerAction). Let me know if it's over-engineered (or simply wrong 🙃...); thanks 🙇

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerActivity.kt
develric: 👍...see [this comment above ](https://github.com/wordpress-mobile/WordPress-Android/pull/12884#discussion_r486983389)

develric:
Hi @planarvoid, thanks for your review 🙇 ; I tried to refactor things taking into account your comments, hope I didn't messed up 🙃; from testing/functionality point of view nothing should be changed respect your last review round. Let me know and thanks 🙇

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerFragment.kt
```diff
@@ -45,16 +50,80 @@ import org.wordpress.android.util.image.ImageManager
...
+        companion object {
+            @JvmStatic
+            fun fromNameString(iconTypeName: String): MediaPickerIconType {
+                return values().firstOrNull { it.name == iconTypeName }
+                        ?: throw IllegalArgumentException("MediaPickerIconType not found with name $iconTypeName")
+            }
+        }
+    }
+
+    enum class ChooserContext(
```
planarvoid, line 66: ❤️

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerFragment.kt
```diff
@@ -45,16 +50,80 @@ import org.wordpress.android.util.image.ImageManager
...
+        val mediaTypeFilter: String
+    ) {
+        PHOTO(ACTION_GET_CONTENT, RequestCodes.PICTURE_LIBRARY, UiStringRes(R.string.pick_photo), "image/*"),
+        VIDEO(ACTION_GET_CONTENT, RequestCodes.VIDEO_LIBRARY, UiStringRes(R.string.pick_video), "video/*"),
+        PHOTO_OR_VIDEO(ACTION_GET_CONTENT, RequestCodes.MEDIA_LIBRARY, UiStringRes(R.string.pick_media), "*/*"),
+        MEDIA_FILE(ACTION_OPEN_DOCUMENT, RequestCodes.FILE_LIBRARY, UiStringRes(R.string.pick_file), "*/*");
+    }
+
+    sealed class MediaPickerAction {
+        data class OpenSystemChooser(
```
planarvoid: I like this approach! Very low important - I'd consider using `Picker` instead of `Chooser` here for consistency - we have a Media picker and we'd open System picker here :-)

WordPress/src/test/java/org/wordpress/android/ui/mediapicker/MediaPickerViewModelTest.kt
```diff
@@ -365,6 +370,83 @@ class MediaPickerViewModelTest : BaseUnitTest() {
         assertSearchExpanded(query)
     }
 
+    @Test
+    fun `System picker opened for photo when allowed types is IMAGE only`() = test {
```
planarvoid: NIT: I'd not start the test name with a capital letter for consistency

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerViewModel.kt
```diff
@@ -323,15 +333,44 @@ class MediaPickerViewModel @Inject constructor(
...
+
+        _onIconClicked.postValue(Event(populateIconClickEvent(icon, mediaPickerSetup.canMultiselect)))
+    }
+
+    private fun populateIconClickEvent(icon: MediaPickerIcon, canMultiselect: Boolean): IconClickEvent {
+        val action: MediaPickerAction = when (icon) {
+            is ChooseFromAndroidDevice -> {
+                val allowedTypes = icon.allowedTypes
+                val (context, types) = when {
+                    listOf(IMAGE).containsAll(allowedTypes) -> {
```
planarvoid, line 355: this is not super readable but I don't know if there is anything better :/. Maybe `allowedTypes.all { it == IMAGE }`? It's not great either.

planarvoid (approved):
Thanks for the changes and for this PR! It looks good and works well 👍 . I'm approving it with a few non-important comments that you can easily ignore :-) Feel free to merge this PR

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerFragment.kt
develric: I like it 👍, changed here 4b05415acede9b7e3adf763a87d6b4e823f75315

WordPress/src/test/java/org/wordpress/android/ui/mediapicker/MediaPickerViewModelTest.kt
develric: makes sense, changed here a6426572fe5c81c197e02413dbd39a838ce84c98

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerViewModel.kt
develric: yep I was thinking to it also, but at the end could not find a clear winner; if it's fine with you I would keep this one 🙇

planarvoid (approved):
Thanks for the changes 👍

PR #12892, opened 2020-09-04
Hide fab for stories when item selected

### LINKED ISSUE #12644: Hide FAB in the stories mode when an item gets selected
The FAB triggers camera. It's currently only visible for the Stories mode. In this case we should only show it until the user selects an item. Once an item is selected, the FAB should be animated away.

develric:
Fixes #12644

This PR hides the FAB in the Media Picker for stories when an item get selected. No FAB is visible when the picker is not open in a story flow.

### To test

- Create a story
- Notice the Media Picker is opened and the FAB is visible
- Select one or more items and notice the FAB is hidden at the first selected item and stay hidden
- Unselect all items and notice the FAB is visible again
- Smoke test story slide creation
- Open the picker in other flows (GB blocks, Site/Avatar change) and check the FAB is always hidden independently from selected items.

PR submission checklist:


planarvoid (approved):
Thanks for the change, it's looking great! And it shows the beauty of a single state MVVM ❤️

PR #12895, opened 2020-09-05
Reader: Show preview images in lists for more posts

### LINKED ISSUE #12894: Reader: More posts should have preview images in the list view
### Expected behavior

I expect as many reader posts as possible to display a preview image in the feed and site view screens, even if they don't have a featured image set.

### Actual behavior

For the most part, posts containing images but without a featured image set are missing a preview image.

------

While working on supporting posts with a Story block in the reader, I noticed a few oddities with preview images in the Reader list views (both the feed view and the individual site view). This affects posts that contain images (either standalone or as part of some more complex block) but _don't have a featured image set_.

(This doesn't affect things like the Discover feed because those posts tend to have a featured image set.)

The problem is twofold:

#### 1. In the feed view, there is a discrepancy between posts having images between Android and iOS

On iOS, many posts will have a preview image in the feed view while those same posts don't have one on Android. For example:

reader-android-ios-diff

This is at least partly due to a bug with the regex patterns we use to parse image tags in the Android Reader.

#### 2. Some posts that have an image in the feed view are missing one in the site view

Even after fixing the issue mentioned in 1, there's still a discrepancy between those two screens. This also affects iOS [issue here]. The reason seems to be that the `/read/feed/` and `/read/sites/` endpoints return somewhat different content (at least for Jetpack sites), and it seems common for image tags to be missing a width attribute when fetching from the site endpoint - and the width attribute is the primary way both iOS and Android source preview images when no featured image is set.

Side note: I think an ideal fix to these issues would actually be server-side; the apps could pass a parameter to the Reader endpoints specifying a minimum image width, and the server would try to pick a suitable preview image URL for each post and include it in the endpoint response.

Follow sites with a good variety of posts. Posts with a story block, posts with image blocks (but no featured image set) primarily. I have a few I can share if asked.

For those sites, compare the image previews between iOS and Android, and between the feed view and the site view.

##### Tested on various devices, Android 10, WPAndroid `alpha-237`

**iOS issue:** https://github.com/wordpress-mobile/WordPress-iOS/issues/14849

aforcier:
Fixes #12894, allowing the reader to find suitable preview images for posts more often.

#### Feed list

reader-android-feed-beforeafter

#### Site list

reader-android-site-beforeafter

This PR has two parts:

### 1. Fixing an accidental regex breakage

A [huge checkstyle refactor from two years ago](https://github.com/wordpress-mobile/WordPress-Android/commit/ed574ea833797f71e28fb2abd060fee7d305376c#diff-981ccd25a498c16f8b2887442812b346) ended up accidentally adding spaces to a few regex patterns used by the Reader to parse images. This almost entirely broke support for preview images that weren't set as featured. I undid the breakage in 0379e37 (and added some tests).

### 2. Adding support for `srcset` and `data-large-file` `img` tag attributes

The above fix improves things a bit but still fails to recognize images in several kinds of posts. This is especially noticeable in the site view. As explained in #12894 the `/read/feed/` and `/read/sites/` endpoints return different content sometimes. For some reason one of the differences is that some images have no width attribute in the `/read/sites/` endpoint, so what works in the feed doesn't work sometimes in the site view.

To mitigate this, I added support for a few new `img` tag attributes that seem pretty common in WP posts: [srcset](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images) and data-large-file. These are checked for in a fallback style.

### To test:

With and without this patch, follow a few sites with a variety of posts containing images and compare.

Mainly looking for posts with a story block, and posts with image blocks (but no featured image set). (I can recommend a few.)

(For Automatticians: please proxy on your device to fully test posts with the Story block, since depending on the endpoint those won't work correctly otherwise. See pbArwn-14B-p2.)

You can also check the regex fix individually by checking out 86dedbe first and running the (failing) unit tests added.

Notes:
* Some stories (where the first slide is a video) still won't show a preview image - I'll look into that separately
* Detail view of stories is broken right now

### PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
@@ -105,8 +105,16 @@ public class ReaderImageScanner {
             if (width > currentMaxWidth) {
                 currentImageUrl = imageUrl;
                 currentMaxWidth = width;
-            } else if (currentImageUrl == null && hasSuitableClassForFeaturedImage(imageTag)) {
-                currentImageUrl = imageUrl;
+            } else {
+                // Inspect the image tag and look for clues for an image of the right size
+
+                // Look through the srcset attribute if set for the image best matching minImageWidth
+                currentImageUrl = ReaderHtmlUtils.getSrcsetImageForTag(imageTag, minImageWidth);
+
+                // Check if the image tag's class suggests it's a good enough size
+                if (currentImageUrl == null && hasSuitableClassForFeaturedImage(imageTag)) {
+                    currentImageUrl = imageUrl;
+                }
             }
         }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
@@ -115,6 +115,11 @@ public class ReaderImageScanner {
                 if (currentImageUrl == null && hasSuitableClassForFeaturedImage(imageTag)) {
                     currentImageUrl = imageUrl;
                 }
+
+                // Look for a data-large-file attribute if set and use the associated url
+                if (currentImageUrl == null) {
+                    currentImageUrl = ReaderHtmlUtils.getLargeFileAttr(imageTag);
+                }
             }
         }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
@@ -100,30 +100,33 @@ public class ReaderImageScanner {
             String imageTag = imgMatcher.group(0);
             String imageUrl = imgMatcher.group(1);
 
+            // Primary source: check the width attribute.
             int width = Math.max(ReaderHtmlUtils.getWidthAttrValue(imageTag),
                                  ReaderHtmlUtils.getIntQueryParam(imageUrl, "w"));
             if (width > currentMaxWidth) {
                 currentImageUrl = imageUrl;
                 currentMaxWidth = width;
-            } else {
-                // Inspect the image tag and look for clues for an image of the right size
+            }
 
-                // Look through the srcset attribute (if set) for the largest available size of this image
-                SrcsetImage bestFromSrcset = ReaderHtmlUtils.getLargestSrcsetImageForTag(imageTag);
-                if (bestFromSrcset != null && bestFromSrcset.getWidth() > currentMaxWidth) {
-                    currentMaxWidth = bestFromSrcset.getWidth();
-                    currentImageUrl = bestFromSrcset.getUrl();
-                }
+            // Look through the srcset attribute (if set) for the largest available size of this image.
+            SrcsetImage bestFromSrcset = ReaderHtmlUtils.getLargestSrcsetImageForTag(imageTag);
+            if (bestFromSrcset != null && bestFromSrcset.getWidth() > currentMaxWidth) {
+                currentMaxWidth = bestFromSrcset.getWidth();
+                currentImageUrl = bestFromSrcset.getUrl();
+            }
 
-                // Check if the image tag's class suggests it's a good enough size
-                if (currentImageUrl == null && hasSuitableClassForFeaturedImage(imageTag)) {
-                    currentImageUrl = imageUrl;
-                }
+            // Check if the image tag's class suggests it's a good enough size.
+            // Only do this if we don't already have a winner, since we can't be sure of the width
+            // and shouldn't replace an image we know for sure is larger than [minImageWidth].
+            if (currentImageUrl == null && hasSuitableClassForFeaturedImage(imageTag)) {
+                currentImageUrl = imageUrl;
+            }
 
-                // Look for a data-large-file attribute if set and use the associated url
-                if (currentImageUrl == null) {
-                    currentImageUrl = ReaderHtmlUtils.getLargeFileAttr(imageTag);
-                }
+            // Look for a data-large-file attribute if set and use the associated url.
+            // Only do this if we don't already have a winner, since we can't be sure of the width
+            // and shouldn't replace an image we know for sure is larger than [minImageWidth].
+            if (currentImageUrl == null) {
+                currentImageUrl = ReaderHtmlUtils.getLargeFileAttr(imageTag);
             }
         }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
@@ -109,7 +109,11 @@ public class ReaderImageScanner {
                 // Inspect the image tag and look for clues for an image of the right size
 
                 // Look through the srcset attribute if set for the image best matching minImageWidth
-                currentImageUrl = ReaderHtmlUtils.getSrcsetImageForTag(imageTag, minImageWidth);
+                SrcsetImage bestFromSrcset = ReaderHtmlUtils.getSrcsetImageForTag(imageTag, minImageWidth);
+                if (bestFromSrcset != null) {
+                    currentMaxWidth = bestFromSrcset.getWidth();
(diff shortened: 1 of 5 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
```diff
@@ -105,8 +105,21 @@ public String getLargestImage(int minImageWidth) {
             if (width > currentMaxWidth) {
                 currentImageUrl = imageUrl;
                 currentMaxWidth = width;
-            } else if (currentImageUrl == null && hasSuitableClassForFeaturedImage(imageTag)) {
-                currentImageUrl = imageUrl;
+            } else {
+                // Inspect the image tag and look for clues for an image of the right size
+
+                // Look through the srcset attribute if set for the image best matching minImageWidth
```
malinajirka: The method is called `getLargestImage` but `getSrcsetImageForTag` returns the smallest image from the set. Is that correct? If it is, we might want to consider renaming the method, wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
```diff
@@ -105,8 +105,21 @@ public String getLargestImage(int minImageWidth) {
...
+                currentImageUrl = ReaderHtmlUtils.getSrcsetImageForTag(imageTag, minImageWidth);
```
malinajirka: We are still in the `while (imgMatcher.find()) {` loop. AFAIU `currentImageUrl` holds the most suitable (largest) image we've found so far. Won't this call always override it even when we don't want to - eg. it will set it to null when there is no `srcset` attribute. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
```diff
@@ -105,8 +105,21 @@ public String getLargestImage(int minImageWidth) {
...
+
+                // Check if the image tag's class suggests it's a good enough size
+                if (currentImageUrl == null && hasSuitableClassForFeaturedImage(imageTag)) {
+                    currentImageUrl = imageUrl;
+                }
+
+                // Look for a data-large-file attribute if set and use the associated url
+                if (currentImageUrl == null) {
+                    currentImageUrl = ReaderHtmlUtils.getLargeFileAttr(imageTag);
```
malinajirka: Similar to the above comment - the method is called `getLargestImage` -> it seems we are checking the "largeFile" attribute only when we haven't found any images so far. Shouldn't we always be looking if the tag has `largeFile` attribute, assuming it might contain the largest image? Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
```diff
@@ -105,8 +105,21 @@ public String getLargestImage(int minImageWidth) {
             if (width > currentMaxWidth) {
                 currentImageUrl = imageUrl;
                 currentMaxWidth = width;
-            } else if (currentImageUrl == null && hasSuitableClassForFeaturedImage(imageTag)) {
-                currentImageUrl = imageUrl;
+            } else {
+                // Inspect the image tag and look for clues for an image of the right size
```
malinajirka: The else branch never updates the `currentMaxWidth`. Is it intentional? We are still in the while loop so if we don't update it we might override the `currentImageUrl` with a smaller image because `if (width > currentMaxWidth)` will be using out-of-date `currentMaxWidth`. Wdyt?

WordPress/src/test/java/org/wordpress/android/util/ReaderHtmlUtilsTest.kt
```diff
@@ -0,0 +1,175 @@
+package org.wordpress.android.util
+
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.wordpress.android.ui.reader.utils.ReaderHtmlUtils
+
+class ReaderHtmlUtilsTest {
```
malinajirka, line 9: ❤️

malinajirka (commented):
Thanks @aforcier! I haven't tested the app yet. I've reviewed the regexes and they all look great. 
I've a few questions regarding the code changes though - all of them are related to a single method. There is a pretty high chance I'm missing something obvious. Let me know what you think :).

WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
aforcier: You're right, things are working at cross purposes here a bit 😅

I was going for optimizing the filesize but it's unnecessary, we always [Photonize the URL](https://github.com/wordpress-mobile/WordPress-Android/blob/a3988b1b990862b2e7ee16df6737c449a43349f1/libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/PhotonUtils.java#L49) according to the screen width at the end anyway.

Updated `getSrcsetImageForTag` to return the biggest image instead in 7a5c5cf.

WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
aforcier: 👍 Yeah - I fixed this in e872a64, it'll only update the `currentImageUrl` if the new width is bigger than `currentMaxWidth`.

WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
aforcier: I think we shouldn't rely on the `data-large-file` attribute, because it's arbitrary - we have no way of knowing what the actual width the image is short of downloading it and seeing. So there's no guarantee that it won't actually be smaller than the `width` attribute if it's set. So I think it makes sense to keep it as a fallback, if we aren't able to pin down a URL with an actual known width.

I'll make a top-level comment laying out how I think the logic of this function should play out overall.

WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
aforcier: I think it makes sense for the two width-less checks (`hasSuitableClassForFeaturedImage` and `getLargeFileAttr`) not to update the `currentMaxWidth`, since we don't know the actual width of the photos in those cases (this is also how `hasSuitableClassForFeaturedImage` already worked I believe). Let me know if you think otherwise.

However the `srcset` check can have that information, so it updates the `currentMaxWidth` as of 7a5c5cf.

aforcier:
> Thanks @aforcier! I haven't tested the app yet. I've reviewed the regexes and they all look great.
I've a few questions regarding the code changes though - all of them are related to a single method. There is a pretty high chance I'm missing something obvious. Let me know what you think :).

Thanks @malinajirka! I got a bit too in the weeds there and neglected some of the bigger picture of what that function was trying to do.

To summarize the approach I'm taking, I think the `getLargestImage` function should obey these rules:
1. Choose an image whose size is known (as long as it's larger than the given `minImageWidth`) over one whose size isn't known.
2. All things being equal, keep the first image found. This means:
    * If there are multiple images with the same (maximum) width, keep the first one.
    * Don't replace an existing URL with a fallback one (`hasSuitableClassForFeaturedImage` and `getLargeFileAttr`). An existing URL is either a known-width image (so the first rule applies), or we got it from an earlier fallback (in which case we should keep the first image since we don't know the real width of either).

The current state should follow those rules properly now. Ready for a second pass 🙁

WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
malinajirka: I see, thanks for the clarification ;)

WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderImageScanner.java
malinajirka: Perfect, thanks!

malinajirka (approved):
Thanks for the changes ;)! I've tested the app and it works great. Great catch ;)

Issue #12821, opened 2020-08-28
Add missing unit tests for ReaderImprovements Phase 2


malinajirka:
BackLog

In Progress
- [ ] Placeholder
- [x] ReaderPostCardActionHandler (@malinajirka)
- [x] ReaderSiteFollowUseCase (@ashiagr )

Done
- [x] ReaderDiscoverViewModel (@malinajirka)
- [X] DiscoverDataProvider @zwarm 
- [x] PostLikeUseCase (@zwarm)

PR #12898, opened 2020-09-07
RI: Follow use case tests

ashiagr:
Parent issue #12821

This PR adds tests for Follow use case.

To test:
Run Test

PR submission checklist:


WordPress/src/test/java/org/wordpress/android/ui/reader/usecases/ReaderSiteFollowUseCaseTest.kt
```diff
@@ -0,0 +1,256 @@
...
+        }
+    }
+
+    private fun <T> testWithFailedResponseForFollowedSitePost(block: suspend CoroutineScope.() -> T) {
+        test {
+            readerPost.isFollowedByCurrentUser = true
+            whenever(readerPostTableWrapper.isPostFollowed(readerPost)).thenReturn(true)
+            whenever(readerBlogActionsWrapper.followBlogForPost(anyOrNull(), anyBoolean(), anyOrNull())).then {
+                (it.arguments[2] as ActionListener).onActionResult(false)
+                false
```
malinajirka: I believe the code returns true from `followBlogForPost` and invokes `onActionResult(false)` when a request fails. We are returning false even from `followBlogForPost` here - is it intentional?

malinajirka (commented):
Thanks @ashiagr! I've left one question but it looks great otherwise. Thanks ;)!

WordPress/src/test/java/org/wordpress/android/ui/reader/usecases/ReaderSiteFollowUseCaseTest.kt
ashiagr: We do not use the value returned from `followBlogForPost`, instead 
- first emit `PostFollowStatusChanged` for optimistic update [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderSiteFollowUseCase.kt#L38)
- then use the return (boolean) value of the [last suspension](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderSiteFollowUseCase.kt#L61-L67) point [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderSiteFollowUseCase.kt#L45) 

So it seems return value from `followBlogForPost` doesn’t matter in the refactored code. But I see your point and have made a change here: 3912425

ashiagr:

I've addressed your comment. Additionally, added two more commits:
1. 2a5bbed - Use un-followed site to test follow site action on request failure
2. 60c245c - Add few more tests

Ready for another round 🙇‍♀️.

malinajirka (approved):
Perfect, thanks for all the improvements ;)! LGTM 🚢

PR #12904, opened 2020-09-08
Issue/12028 add tests for discover vm

malinajirka:
Parent issue #12821 

Adds some missing unit tests for the Discover view model.

To test:
No need to test anything as long as CI jobs pass

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
@@ -2,9 +2,11 @@ package org.wordpress.android.ui.reader.discover
 
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MediatorLiveData
+import androidx.lifecycle.Observer
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.launch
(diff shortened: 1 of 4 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
```diff
@@ -86,6 +89,14 @@ class ReaderDiscoverViewModel @Inject constructor(
     private val photonWidth: Int = (displayUtilsWrapper.getDisplayPixelWidth() * PHOTON_WIDTH_QUALITY_RATION).toInt()
     private val photonHeight: Int = (photonWidth * FEATURED_IMAGE_HEIGHT_WIDTH_RATION).toInt()
 
+    private val communicationChannelObserver = Observer { data: Event<ReaderDiscoverCommunication>? ->
```
malinajirka, line 92: I've moved the observer into a field so we can remove it in onCleared method.

zwarm: I like the way you moved the communicationChannel stream logic to handlers!

zwarm (approved):
Changes look good and make the code easier to read. Nice work.

PR #12909, opened 2020-09-08
Try Firebase with API 28

jkmassel:
Attempts to fix failing Firebase tests

PR submission checklist:

- [ ] I have considered adding unit tests where possible.
- [ ] I have considered adding accessibility improvements for my changes.
- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #12908, opened 2020-09-08
Use API 27 for Connected Tests

jkmassel:
Fixes an issue where CI was failing because of WPStories integration.

**To test:**
- Ensure CI passes

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #12845, opened 2020-08-31
WP Stories: add permission denied dialog interface

mzorz:
Fixes https://github.com/Automattic/stories-android/issues/418

This PR implements the `PermanentPermissionDenialDialogProvider` interface from the Stories library to show a consistent explanation to users when they try to access the Stories capture mode but then deny the permissions request and ask not to show again.

To test:
See test instructions in https://github.com/Automattic/stories-android/pull/511

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #12910, opened 2020-09-09
e2e tests: Use Espresso method to find overflow menus

aforcier:
Fixes (I think) the current e2e test breakage caused by changing the emulator used to Pixel API 28, by changing the way the tests detect overflow menus.

### Background

1. https://github.com/wordpress-mobile/WordPress-Android/pull/12845 caused connected the tests to break due to an Android OS bug in stock Android API 26. (See https://github.com/wordpress-mobile/WordPress-Android/pull/12908#pullrequestreview-484342586 for a full explanation.)
2. https://github.com/wordpress-mobile/WordPress-Android/pull/12908 changed the connected tests to use a Pixel 2 API 27, but the above OS bug was still present on that emulator
3. https://github.com/wordpress-mobile/WordPress-Android/pull/12909 updated the API level to 28 (still Pixel 2)

After 3, it looks like the original issue introduced by https://github.com/wordpress-mobile/WordPress-Android/pull/12845 was resolved, but one test (`e2e.EditorTests.testPublishFullPost`) now fails continuously.

The point of failure is [here](https://github.com/wordpress-mobile/WordPress-Android/blob/5d1661cbf21c52c9b29f785737e5bf9597dda642/WordPress/src/androidTest/java/org/wordpress/android/e2e/pages/EditorPage.java#L90) - the test is trying to open the overflow menu in the editor, and appears not to be finding it.

The way we're finding it is a bit error-prone - we're looking for a UI element with content description "More options", which is set by the system and not by us (and subject to change).

### The fix

I haven't been able to confirm this directly (my own API 28 Pixel 2 emulator runs the tests successfully without this branch), but I suspect Firebase's emulator does not have "More options" as the content description for overflow menus. Maybe it's missing, or it's called something else, or the system UI is in French 🙂 (Actually it looks like it's missing for certain versions of [ActionBarActivityCompat](https://android.googlesource.com/platform/frameworks/testing/+/android-support-test/espresso/core/src/main/java/android/support/test/espresso/Espresso.java#146).)

It turns out that there's an Espresso function we can use instead: [openActionBarOverflowOrOptionsMenu](https://developer.android.com/training/testing/espresso/recipes#matching-view-inside-action-bar). This does the same job but seems to cover more cases (from looking at [the implementation](https://android.googlesource.com/platform/frameworks/testing/+/android-support-test/espresso/core/src/main/java/android/support/test/espresso/Espresso.java#146) it looks like it'll target by class name if "More options" fails). With that change I was able to get the tests to pass on Firebase. I replaced all the usages of this I found in the tests.

### To test:

I'm triggering the connected tests for this PR (I think it's the same job as runs on merge to develop?), but you also like to:

1. Run the test suite on firebase yourself, like so:

```
$ ./gradlew WordPress:assembleVanillaDebug WordPress:assembleVanillaDebugAndroidTest

$ gcloud firebase test android run \
--type instrumentation \
--app WordPress/build/outputs/apk/vanilla/debug/org.wordpress.android-vanilla-debug.apk \
--test WordPress/build/outputs/apk/androidTest/vanilla/debug/org.wordpress.android-vanilla-debug-androidTest.apk \
--device model=Pixel2,version=28,locale=en,orientation=portrait \
--test-targets="notPackage org.wordpress.android.ui.screenshots" \
--project api-project-108380595987 \
--no-record-video \
--timeout 10m
```

(I have and things seem to be working)

or

2. Merge the PR and watch the automated connected test run - they're broken right now anyway 😅 

#### PR submission checklist:

### DIFF
--- a/WordPress/src/androidTest/java/org/wordpress/android/e2e/pages/BlockEditorPage.java
+++ b/WordPress/src/androidTest/java/org/wordpress/android/e2e/pages/BlockEditorPage.java
@@ -1,22 +1,22 @@
 package org.wordpress.android.e2e.pages;
 
+import androidx.test.core.app.ApplicationProvider;
 import androidx.test.espresso.ViewInteraction;
 import androidx.test.espresso.action.ViewActions;
 
 import org.wordpress.android.R;
 
 import static androidx.test.espresso.Espresso.onView;
+import static androidx.test.espresso.Espresso.openActionBarOverflowOrOptionsMenu;
 import static androidx.test.espresso.action.ViewActions.typeText;
 import static androidx.test.espresso.assertion.ViewAssertions.matches;
 import static androidx.test.espresso.matcher.ViewMatchers.isDisplayed;
-import static androidx.test.espresso.matcher.ViewMatchers.withContentDescription;
 import static androidx.test.espresso.matcher.ViewMatchers.withHint;
 import static androidx.test.espresso.matcher.ViewMatchers.withId;
 import static org.wordpress.android.support.WPSupportUtils.clickOn;
 import static org.wordpress.android.support.WPSupportUtils.waitForElementToBeDisplayed;
 
 public class BlockEditorPage {
-    private static ViewInteraction optionsButton = onView(withContentDescription("More options"));
     private static ViewInteraction titleField = onView(withHint("Add title"));
 
     private ViewInteraction mEditor;
@@ -35,7 +35,7 @@ public class BlockEditorPage {
     }
 
     public void switchToClassic() {
-        clickOn(optionsButton);
+        openActionBarOverflowOrOptionsMenu(ApplicationProvider.getApplicationContext());
         clickOn("Switch to classic editor");
     }
 }
--- a/WordPress/src/androidTest/java/org/wordpress/android/e2e/pages/EditorPage.java
+++ b/WordPress/src/androidTest/java/org/wordpress/android/e2e/pages/EditorPage.java
@@ -1,11 +1,13 @@
 package org.wordpress.android.e2e.pages;
 
+import androidx.test.core.app.ApplicationProvider;
 import androidx.test.espresso.ViewInteraction;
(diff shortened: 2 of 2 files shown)


mzorz (approved):
LGTM!

1. installed google cloud CLI https://firebase.google.com/docs/test-lab/android/command-line
https://cloud.google.com/sdk/docs/downloads-versioned-archives
2. run steps as provided on `develop`
3. verified it failed with `09-09 05:33:54.780: E/ViewAssertions(8362): 'is displayed on the screen to the user' check could not be performed because view 'with content description text: is "More options"' was not found.`
```
More details are available at [https://console.firebase.google.com/project/api-project-108380595987/testlab/histories/bh.b3d1e88d8b09c8be/matrices/7111675873554974567].
┌─────────┬───────────────────────┬────────────────────────────────────────────┐
│ OUTCOME │    TEST_AXIS_VALUE    │                TEST_DETAILS                │
├─────────┼───────────────────────┼────────────────────────────────────────────┤
│ Failed  │ Pixel2-28-en-portrait │ 1 test cases failed, 116 passed, 1 skipped │
└─────────┴───────────────────────┴────────────────────────────────────────────┘
```
4. applied this PR's branch and ran the test again
https://console.firebase.google.com/project/api-project-108380595987/testlab/histories/bh.b3d1e88d8b09c8be/matrices/7561308975691691204
5. verified it worked
```
More details are available at [https://console.firebase.google.com/project/api-project-108380595987/testlab/histories/bh.b3d1e88d8b09c8be/matrices/7561308975691691204].
┌─────────┬───────────────────────┬──────────────────────────────────┐
│ OUTCOME │    TEST_AXIS_VALUE    │           TEST_DETAILS           │
├─────────┼───────────────────────┼──────────────────────────────────┤
│ Passed  │ Pixel2-28-en-portrait │ 117 test cases passed, 1 skipped │
└─────────┴───────────────────────┴──────────────────────────────────┘
```

:shipit:

Issue #12834, opened 2020-08-31
Reader Improvements: Sites/Blogs you might like


ashiagr:
Project Issue: https://github.com/wordpress-mobile/WordPress-Android/issues/12899

### Goal
Display a list of recommended sites to follow in the new "Discover" feed on the "Reader" tab, and allow the user to follow a site directly in the feed or tap it to see all the posts for the site. 

<img width="50%" alt="Screen Shot 2020-08-26 at 2 35 18 PM" src="https://user-images.githubusercontent.com/793774/91342704-43fa3780-e790-11ea-9855-e51a8b0abc07.png">

**Note** 
New "Discover" feed is currently hidden under a runtime feature flag which can be enabled as per steps described in #12050.

### Designs
Specs: [Figma](https://www.figma.com/file/RjrZZjXCtNYXWsPt5i0jha/Sites-You-Might-Like?node-id=0%3A1)

### Tasks

**BackLog**

**In Progress**

**Done**
- [x] Display recommended blogs card item in Discovery tab, add click action #12911
- [x] Design review #12960
- [x] Integrate Follow action #12960

AdamGrzybkowski:
Some question about the feature:
1. I can see that the "Sites to follow" card is displayed somewhere between other cards. Is there a specified/fixed position for it? If not, is there a rule on how it should be calculated? 
2. How should the app behave when:
       - there are no suggestions (is this even possible with the API?)
       - there are more than 3 suggestions
3. Should there be a message displayed to the user if "Follow/Unfollow" action after the icon click fails? If yes, is there a guideline on how you handle such messages (Snackbar etc.) and is there a copy for such message defined?

ashiagr:
Great questions! Here are the answers:

> 1. I can see that the "Sites to follow" card is displayed somewhere between other cards. Is there a specified/fixed position for it? If not, is there a rule on how it should be calculated?

We should follow the order returned by [the API](https://public-api.wordpress.com/wpcom/v2/read/tags/cards?tags=android) and display cards with “recommended_blogs” card type under “Sites to follow”. Note that “recommended_blogs” are not parsed yet unlike other card types. We can discuss it more when you've gone through the entire flow.

> 2. How should the app behave when: 
> - there are no suggestions (is this even possible with the API?)

If there are no suggestions, we shouldn’t display “Sites to follow” section.

  > - there are more than 3 suggestions

Let’s restrict to just 3 suggestions in a section. I’ll confirm the exact count.

[...]

PR #12911, opened 2020-09-09
Recommended blogs card item in Discovery tab

AdamGrzybkowski:
Partially fixes: #12834

This PR adds a new item to the Discovery tab - "Recommended blogs". 
Following the blog will be done in a separate PR. 

Screenshots

Portrait
|  Light | Dark  |
|---|---|
| Screenshot_20200910-212344 | Screenshot_20200910-212435 |

Landscape
|  Light | Dark  |
|---|---|
| Screenshot_20200910-212357 | Screenshot_20200910-212445 |

GIF
recommended_blogs

To test:
- Go to the Discovery tab
- Scroll down to see "Sites to follow"

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/ParseDiscoverCardsJsonUseCase.kt
```diff
@@ -36,6 +40,16 @@ class ParseDiscoverCardsJsonUseCase @Inject constructor() {
         return interestTags
     }
 
+    fun parseRecommendedBlogsCard(cardJson: JSONObject): List<ReaderCardRecommendedBlog> {
+        return cardJson.optJSONArray(ReaderConstants.JSON_CARD_DATA)
+                ?.let { jsonRecommendedBlogs ->
+                    List(jsonRecommendedBlogs.length()) { index ->
+                        gson.fromJson(jsonRecommendedBlogs[index].toString(), ReaderCardRecommendedBlog::class.java)
```
AdamGrzybkowski, line 47: Any reasons why other cards are parsed manually?

WordPress/src/main/java/org/wordpress/android/ui/reader/services/discover/ReaderDiscoverLogic.kt
```diff
@@ -157,6 +159,14 @@ class ReaderDiscoverLogic(
                     val post = parseDiscoverCardsJsonUseCase.parsePostCard(cardJson)
                     cards.add(ReaderPostCard(post))
                 }
+                JSON_CARD_RECOMMENDED_BLOGS -> {
+                    cardJson?.let {
+                        val recommendedBlogs = parseDiscoverCardsJsonUseCase.parseRecommendedBlogsCard(it)
+                        if (recommendedBlogs.isNotEmpty()) {
```
AdamGrzybkowski: Having "recommended_blogs" card with an empty list seems like a weird behavior so I wanted to handle this case in the data layer so that presentation doesn't have to care about that.

WordPress/src/main/java/org/wordpress/android/ui/utils/RecyclerViewExtensions.kt
```diff
@@ -0,0 +1,28 @@
...
+import android.view.View
+import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
+
+fun RecyclerView.addItemDivider(drawable: Drawable) {
+    if (layoutManager !is LinearLayoutManager) return
+    addItemDecoration(DividerItemDecorator(drawable))
+}
+
+class DividerItemDecorator(private val divider: Drawable) : RecyclerView.ItemDecoration() {
```
AdamGrzybkowski, line 14: I saw that other lists have a divider in each item, but I think that ItemDecarotor is the way to go for that. You can reuse it without adding extra properties to each UiItem.

WordPress/src/main/java/org/wordpress/android/ui/utils/ViewExtensions.kt
```diff
@@ -0,0 +1,11 @@
+package org.wordpress.android.ui.utils
+
+import android.view.View
+
+fun View.show(isVisible: Boolean) {
```
AdamGrzybkowski: I know you have a Utility class but extensions seem more Kotlinish? Maybe it's worth migrating slowly?

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/ParseDiscoverCardsJsonUseCase.kt
ashiagr: Great question!  Simple reason - its a legacy code :). Models (like `ReaderPost`, `ReaderBlog`) already exist along with the parsing logic, we simply reused those classes/ logic. That said, we’re discussing about refactoring/migrating entire backend code for the Reader. I’m not sure if we’d like to introduce `gson` for a small amount of change at the moment.

WordPress/src/main/java/org/wordpress/android/ui/utils/ViewExtensions.kt
ashiagr: Wondering if we can use [ViewUtils -> View.setVisible](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/util/ViewUtils.kt#L13) 🤔. I think it serves the same purpose.

AdamGrzybkowski: Yes! I missed it :)

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/ParseDiscoverCardsJsonUseCase.kt
AdamGrzybkowski: yeah, I thought about reusing [this](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/models/ReaderRecommendedBlog.java) but the models are a bit different so I decided to create a new one and wanted to use that opportunity to write it that way. I did use GSON as it is already in the project (I would prefer to use Moshi as it works better with Kotlin in my opinion but that would be just another lib, so it didn't make sense)

ashiagr: Not sure if we'll need new model `ReaderCardRecommendedBlog`. As discussed in the p2 post, we'll see where the implementation will lead us. 

EDIT: Let's make a decision for the best model class later and keep what we have for now.

WordPress/src/main/java/org/wordpress/android/models/ReaderCardRecommendedBlog.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import com.google.gson.annotations.SerializedName
+
+data class ReaderCardRecommendedBlog(
+    @SerializedName("name") val name: String,
+    @SerializedName("url") val url: String,
+    @SerializedName("blog_id") val blogId: Long,
+    @SerializedName("description") val description: String,
+    @SerializedName("icon") val iconUrl: String?,
+    @SerializedName("feed_id") val feedId: Long?
```
AdamGrzybkowski, line 11: `feedId` is nullable for now but I'm waiting for confirmation that this is how API behaves (and not e.g. setting 0 when not available)

WordPress/src/main/res/layout/reader_interest_card.xml
```diff
@@ -22,7 +22,7 @@
             app:layout_constraintTop_toBottomOf="@id/interests_you_may_like_header" />
 
         <org.wordpress.android.widgets.WPTextView
-            style="@style/ReaderTextView.Interests.Header"
+            style="@style/ReaderTextView.Card.Header"
```
AdamGrzybkowski, line 25: The `recommended_blogs' card header should match the style of the `interests` card so I decided to reuse the style. I renamed it so it is a bit more clear.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderRecommendedBlogsAdapter.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    override fun onBindViewHolder(holder: ReaderRecommendedBlogViewHolder, position: Int) {
+        holder.onBind(getItem(position))
+    }
+
+    class RecommendedBlogsDiffUtil : DiffUtil.ItemCallback<ReaderRecommendedBlogUiState>() {
+        override fun areItemsTheSame(
+            oldItem: ReaderRecommendedBlogUiState,
+            newItem: ReaderRecommendedBlogUiState
+        ): Boolean = oldItem.blogId == newItem.blogId
```
ashiagr: We might want to include `feedId` comparison here after integrating `feedId` changes 😄. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderRecommendedBlogsCardViewHolder.kt
```diff
@@ -0,0 +1,32 @@
...
+
+class ReaderRecommendedBlogsCardViewHolder(
+    parentView: ViewGroup,
+    imageManager: ImageManager
+) : ReaderViewHolder(parentView, R.layout.reader_recommended_blos_card) {
+    private val recommendedBlogsAdapter = ReaderRecommendedBlogsAdapter(imageManager)
+
+    init {
+        if (recommended_blogs.adapter == null) {
+            recommended_blogs.adapter = recommendedBlogsAdapter
```
ashiagr: Wondering if we need a null check here 🤔. Wouldn't it be always null during init?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderRecommendedBlogsCardViewHolder.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.wordpress.android.ui.reader.discover.ReaderCardUiState.ReaderRecommendedBlogsCardUiState
+import org.wordpress.android.ui.reader.discover.ReaderRecommendedBlogsAdapter
+import org.wordpress.android.ui.utils.addItemDivider
+import org.wordpress.android.util.AppLog
+import org.wordpress.android.util.image.ImageManager
+
+class ReaderRecommendedBlogsCardViewHolder(
+    parentView: ViewGroup,
+    imageManager: ImageManager
+) : ReaderViewHolder(parentView, R.layout.reader_recommended_blos_card) {
```
ashiagr: Typo :) can we rename `blos` to `blogs`?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderRecommendedBlogViewHolder.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val imageManager: ImageManager,
+    override val containerView: View = LayoutInflater.from(parent.context)
+            .inflate(R.layout.reader_recommended_blog_item, parent, false)
+) : RecyclerView.ViewHolder(containerView), LayoutContainer {
+    fun onBind(uiState: ReaderRecommendedBlogUiState) {
+        with(uiState) {
+            site_name.text = name
+            site_url.text = url
+            site_description.text = description
+            site_description.setVisible(description.isNotEmpty())
```
ashiagr: Can we set visibility state in the `ReaderRecommendedBlogUiState` and access it here? Might be easier to write a unit test for visibility.

WordPress/src/main/res/layout/reader_recommended_blog_item.xml
```diff
@@ -0,0 +1,93 @@
...
+        app:layout_constraintGuide_begin="@dimen/margin_extra_large" />
+
+    <androidx.constraintlayout.widget.Guideline
+        android:id="@+id/guideline_bottom"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:orientation="horizontal"
+        app:layout_constraintGuide_end="@dimen/margin_extra_large" />
+
+    <org.wordpress.android.widgets.WPTextView
```
ashiagr: We try to use material versions of widgets like `MaterialCardView` , `MaterialButton`, `MaterialTextView` now. Can we replace `WPTextView` with `MaterialTextView`?

Let me know if you need any examples in the app how we use it. I'll happy to help :).

WordPress/src/main/res/values-night/colors.xml
```diff
@@ -61,6 +61,9 @@
     <color name="reader_welcome_card_background">@color/background_dark</color>
     <color name="reader_welcome_card_text">@color/white</color>
 
+    <!--Reader Recommended Blogs Card -->
+    <color name="reader_recommended_blogs_card_divider">@color/white_translucent_12</color>
+
```
ashiagr: We try to avoid adding separate colors for light and dark modes wherever possible. 

Currently we have

Dark mode
`<color name="reader_recommended_blogs_card_divider">@color/white_translucent_12</color>`
Light mode
`<color name="reader_recommended_blogs_card_divider">@color/black_translucent_12</color>`

These can be replaced by using `colorOnSurface` + `disabled_alpha` (.12) which will give the same output. Wondering if we can use an [already existing](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/res/color/on_surface_divider.xml) color here 🤔. Lmkwyt.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
```diff
@@ -264,6 +273,10 @@ class ReaderDiscoverViewModel @Inject constructor(
         }
     }
 
+    private fun onRecommendedBlogItemClicked(blogId: Long, feedId: Long?) {
+        _navigationEvents.postValue(Event(ShowBlogPreview(blogId, feedId ?: 0)))
```
ashiagr, line 293: Can we include a unit test for click action?

WordPress/src/main/res/layout/reader_recommended_blog_item.xml
```diff
@@ -0,0 +1,93 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/reader_interest_container"
```
ashiagr: Seems like a copy-paste mistake 😊. Can we rename it to represent recommended blog container or remove if we don't need it?

WordPress/src/main/res/layout/reader_recommended_blos_card.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8"?>
+<com.google.android.material.card.MaterialCardView xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/interest_container"
```
ashiagr: `interest_container` doesn't seem to represent this container correctly. Seems like we can remove it. Wdyt?

ashiagr (commented):
Great work @AdamGrzybkowski ! I really liked how you've paid attention to small details like 

- checking for possibility of having feeds in the response and proactively asking for details
- using an item decorator, even adding an extension file for future use
- checking for blank results

and most importantly integrating entire flow: endpoint integration -> VM logic ->  UI display with the limited knowledge of the app in this PR.

I've left minor suggestions/ observations. Feel free to ping me if you need any example from the app e.g. for the material theming comment.

WordPress/src/main/java/org/wordpress/android/models/ReaderCardRecommendedBlog.kt
AdamGrzybkowski: Currently, both `feed_id` and `icon` fields would be omitted but it will be changed and they will be null if not available. 
Gson will handle both these cases correctly so the model is correct and the comment can be resolved.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderRecommendedBlogsAdapter.kt
AdamGrzybkowski: If we can have two objects with the same `blogId` and different `feedId` than definitely. I treated the `blogId` as the main identifier for the object.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderRecommendedBlogsCardViewHolder.kt
AdamGrzybkowski: correct, only one ViewHolder is created per View 👍

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderRecommendedBlogsCardViewHolder.kt
AdamGrzybkowski: sure! :)

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/viewholders/ReaderRecommendedBlogViewHolder.kt
AdamGrzybkowski: yeah, that's a good idea. I will add a unit test as well :)

WordPress/src/main/res/layout/reader_recommended_blog_item.xml
AdamGrzybkowski: ohh sure, I thought you want to use the custom text view. 

BTW. [TextView and some other views](https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/theme/MaterialComponentsViewInflater.java) are automatically replaced with Material components when using the Material theme, but this might now work with the Bridge theme 🤔

WordPress/src/main/res/values-night/colors.xml
AdamGrzybkowski: ohh that's smart! We can definitely use it :)

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
AdamGrzybkowski: yeah we could, but we would have to change the visibility modifier to public. 

I thought about testing it indirectly by using `ReaderPostUiStateBuilder` as it should return the UiState model with an `onItemClick` function from the ViewModel, but this is a mock so it won't work that way 🤔

WordPress/src/main/res/layout/reader_recommended_blog_item.xml
AdamGrzybkowski: ohh yes we can remove it, good catch! 👍

WordPress/src/main/res/layout/reader_recommended_blos_card.xml
AdamGrzybkowski: same here, we can remove it

AdamGrzybkowski:
@ashiagr I did push one additional commit with a fix. I noticed that I was filtering out the RecommendedBlogsCard with an empty list in the wrong place before storing it locally.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
ashiagr: It's bit tricky 😅. We might want to use `InvocationOnMock` - [example](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/test/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModelTest.kt#L118-L130). Let me know if it helps.

WordPress/src/main/res/layout/reader_recommended_blog_item.xml
ashiagr: Yeah, I'm not an expert but seems like they might not work with the Bridge theme. Thank you for replacing them with material components!

WordPress/src/main/res/values-night/colors.xml
ashiagr: Now that we've reused this existing color we can remove unused
`black_translucent_12`
`white_translucent_12`

AdamGrzybkowski: done

ashiagr:
Thank you for all the changes @AdamGrzybkowski ! 

There are just two outstanding issues:
https://github.com/wordpress-mobile/WordPress-Android/pull/12911#discussion_r487678855
https://github.com/wordpress-mobile/WordPress-Android/pull/12911#discussion_r487675488
and a merge conflict (latest develop is merged to the feature branch).

Once these are resolved, this PR should be ready for merge.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
AdamGrzybkowski: Ok, this is a TIL for me :P I rarely rely on mocks that heavily so I haven't known about this! 

Should be good to go now :)

ashiagr: Fixed in 471f509

ashiagr:
I've reviewed the code and tests. All looks good to me 👍.

PR #12912, opened 2020-09-09
Add paging and refresh to media picker

planarvoid:
This PR adds to media picker:
- the swipe to refresh - you can now refresh the data that are visible by pulling down. This doesn't make much sense with the current implementation but in theory the media in the library could change and you could reload it.
- pagination - this cannot be tested right now because we always load all of the media from the device. This is more of a preparation for the implementation of the WP media library. 

There are a few details that are worth noting:
- I've changed the data source to the Flux pattern. This means that now there are 2 methods on the data source. One triggers the data loading and returns a success/failure result if the operation succeeds/fails. The second operation loads the current cached data. This approach is more flexible because it supports database (if needed) as a single source of truth. The current implementation uses in-memory cache because it's cheap and fast to load the media from the provider. The main reason for this change is that the filtering didn't make sense in the `MediaLoader` so it's now moved to the `MediaSource` as part of the `get` operation.
- There is some simplification to the `MediaLoader` now based on these changes. 
- Instead of sending supported media types to the `MediaLoader` with the `Start` event they are now set when the loader is built. I think it makes sense because it's more of a setup of the loader and it never changes.
- The `MediaPickerUiItem`s now don't share as many fields as before. This is because now we have a `NextPageLoader` item which is not selectable. This requires some changes to the diff callback but I think the result is better now.

To test:
- Test that the media picker behaves as expected
- Try pull to refresh on the media picker screen
- It shows the "Refreshing" item which quickly disappears
- More tests will be necessary with the future changes

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/mediapicker/DeviceListBuilder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/mediapicker/DeviceListBuilder.kt
@@ -44,21 +44,23 @@ class DeviceListBuilder
         mediaTypes: Set<MediaType>,
         offset: Int,
         pageSize: Int?
-    ) = withContext(bgDispatcher) {
-        val result = mutableListOf<MediaItem>()
-        val deferredJobs = mediaTypes.map { mediaType ->
-            when (mediaType) {
-                IMAGE -> async { addMedia(Media.EXTERNAL_CONTENT_URI, IMAGE) }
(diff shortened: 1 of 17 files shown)


develric (approved):
Hi @planarvoid 👋 !

> There are a few details that are worth noting...

All these seems reasonable to me and agree with the approach 🙇 

I tested the swipe to refresh (also manually adding a file in the middle of the process) and it works well. As you said it's still not possible to test the pagination but the code looks good; also since we are having all this still behind a feature flag we will further test this functionality when all pieces are in place for it.

Good job 👍 ! :shipit:

PR #12917, opened 2020-09-09
Add tests

zwarm:
Parent #12821 

This PR adds tests for PostLikeUseCase

To test:
Run tests.

PR submission checklist:

### DIFF
--- a/WordPress/src/test/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeUseCaseTest.kt
+++ b/WordPress/src/test/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeUseCaseTest.kt
@@ -48,8 +48,6 @@ class PostLikeUseCaseTest {
 
     private lateinit var useCase: PostLikeUseCase
 
-    private val readerPost = createDummyReaderPost()
-
     @Before
     fun setUp() {
         val account = AccountModel()
@@ -69,7 +67,7 @@ class PostLikeUseCaseTest {
     @Test
     fun `NoNetwork returned when no network found`() = test {
         // Given
-        whenever(networkUtilsWrapper.isNetworkAvailable()).thenReturn(false)
+        val readerPost = init(isNetworkAvailable = false)
 
         // When
         val result = useCase.perform(readerPost, true).toList(mutableListOf())
@@ -81,15 +79,7 @@ class PostLikeUseCaseTest {
     @Test
     fun `unchanged returned when already liked and requesting to like`() = test {
         // Given
-        readerPost.isLikedByCurrentUser = true
-        whenever(networkUtilsWrapper.isNetworkAvailable()).thenReturn(true)
-        whenever(
-                readerPostActionsWrapper.performLikeActionLocal(
-                        anyOrNull(),
-                        anyBoolean(),
-                        anyLong()
-                )
-        ).thenReturn(false)
+        val readerPost = init(isLikedByCurrentUser = true, localSucceeds = false)
 
         // When
         val result = useCase.perform(readerPost, true).toList(mutableListOf())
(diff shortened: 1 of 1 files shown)


WordPress/src/test/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeUseCaseTest.kt
```diff
@@ -31,60 +48,273 @@ class PostLikeUseCaseTest {
...
+    fun `Post views NOT bumped when asking to unlike`() =
+            testWithLikedPost {
+                // Act
+                useCase.perform(readerPost, false).toList(mutableListOf())
+
+                // Assert
+                verify(readerPostActionsWrapper, never()).bumpPageViewForPost(anyOrNull())
+            }
+
+    private fun <T> testWithUnlikedPost(block: suspend CoroutineScope.() -> T) {
```
malinajirka: We might want to consider replacing all these methods with 

```
private fun init(
        isLikedByCurrentUser: Boolean = false,
        isNetworkAvailable: Boolean = true,
        succeeds: Boolean = true
    ) {
        readerPost.isLikedByCurrentUser = isLikedByCurrentUser
        whenever(networkUtilsWrapper.isNetworkAvailable()).thenReturn(isNetworkAvailable)
        whenever(readerPostActionsWrapper.performLikeActionLocal(anyOrNull(), anyBoolean(), anyLong()))
                .thenReturn(true)
        whenever(readerPostActionsWrapper.performLikeActionRemote(anyOrNull(), anyBoolean(), anyLong(), anyOrNull()))
                .then {
                    (it.arguments[3] as ActionListener).onActionResult(succeeds)
                }
    }
```
We'd call 
`init(isLikedByCurrentUser = false)` for `testWithUnlikedPost`
`init(isLikedByCurrentUser = false, succeeds = false)` for `testWitFailurehUnlikedPost`
etc.

Wdyt?

WordPress/src/test/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeUseCaseTest.kt
```diff
@@ -31,60 +48,273 @@ class PostLikeUseCaseTest {
 
     private lateinit var useCase: PostLikeUseCase
 
+    private val readerPost = createDummyReaderPost()
```
malinajirka: We might want to consider re-creating the model in each test - using a shared state might bite us in the future as the order of the tests can affect the result. Wdyt?

malinajirka (commented):
Thanks @zwarm! I think we are testing all possible combinations - great job! I've left a couple of suggestions, let me know what you think ;).

WordPress/src/test/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeUseCaseTest.kt
zwarm: As in using an init() in your previous comment. Sounds good! I'll take care of that in a jiffy.

WordPress/src/test/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeUseCaseTest.kt
zwarm: Sure, sounds good. In commit 9a6f6e8

WordPress/src/test/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeUseCaseTest.kt
zwarm: Same commit as above.

WordPress/src/test/java/org/wordpress/android/ui/reader/repository/usecases/PostLikeUseCaseTest.kt
```diff
@@ -33,58 +50,182 @@ class PostLikeUseCaseTest {
...
+                val readerPost = init(isLikedByCurrentUser = true)
+
+                // Act
+                useCase.perform(readerPost, false).toList(mutableListOf())
+
+                // Assert
+                verify(readerPostActionsWrapper, never()).bumpPageViewForPost(anyOrNull())
+            }
+
+    private fun <T> testWithBlock(block: suspend CoroutineScope.() -> T) {
```
malinajirka: I believe we can remove this `testWithBlock` and use `test` everywhere, right? The `test` function [runs the whole provided block](https://github.com/wordpress-mobile/WordPress-Android/blob/860a5a723842fe3d178aec639735b527538a1838/WordPress/src/test/java/org/wordpress/android/CoroutinesUtils.kt#L18) in "runBlocking" so that should be enough, right?

zwarm: Replaced in 6f295d1

malinajirka (approved):
LGTM, thanks ;)!

PR #12919, opened 2020-09-09
Add WP media source to media picker

### LINKED ISSUE #12640: Create WP Media library source for the consolidated media picker
Based on the the result of [this](https://github.com/wordpress-mobile/WordPress-Android/issues/12638) we need to create an implementation to support loading of images from the WP media library

planarvoid:
Fixes #12640  

This PR introduces the WP media library as another source in the consolidated media picker. There were several obstacles which I'll try to describe here.
- Identifiers - the local media uses URI as an identifier. The media items in the media library used remote ID (and they don't have URIs). To get around this I've created a class called `Identifier` that has 2 implementations - `LocalUri` for local files and `RemoteId` for items in the media library. These identifiers replace the selected URIs and are used in the whole component. There is a subclass called Parcel used to save the selected identifiers into the saved instance state
- the `load` method now uses `loadMore` parameter instead of passing in page size and offset

What's missing:
- Figure out tracking

To test:
- Go to GB editor/Add image block/Click "Add media"
- Select "WordPress Media Library"
- Check the selector behaviour (multiselection is allowed)
- Select a few items
- Confirm selection
- Multiple Image blocks are added to the post (for each selected items)

To test 2:
- Check that the previous behaviour is not broken with the "Choose from device" option

PR submission checklist:


develric:
Hey @planarvoid , I did a first pass here and the code it's looking great 👍! Since there are a few conflicts after we merged the #12884 if it's fine with you I would do a deeper testing round after those conflicts are fixed! Hope it makes sense, thanks 🙇

planarvoid:
thanks for the review @develric, I've resolved all the conflicts so this should be ready for another check 👍

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerUiItem.kt
```diff
@@ -29,7 +31,7 @@ sealed class MediaPickerUiItem(
     ) : MediaPickerUiItem(VIDEO)
 
     data class FileItem(
-        val uri: UriWrapper? = null,
+        val identifier: Identifier,
```
develric, line 34: I could not figure out an issue here actually, but I see in all these classes the uri was nullable before and now the identifier (and containing value: UriWrapper in LocalUri) is not. I'm just mentioning in case it can ring some bell in your mind and to be sure it was done fully on purpose 🙇

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerActivity.kt
```diff
@@ -301,11 +288,11 @@ class MediaPickerActivity : LocaleAwareActivity(), MediaPickerListener {
...
 
     private fun doMediaIdsSelected(
-        mediaIds: ArrayList<Long>?,
+        mediaIds: List<Long>?,
         source: MediaPickerMediaSource
     ) {
-        if (mediaIds != null && mediaIds.size > 0) {
-            if (browserType == WP_STORIES_MEDIA_PICKER) {
+        if (mediaIds != null && mediaIds.isNotEmpty()) {
+            if (browserType.canMultiselect()) {
```
develric, line 295: browserType.canMultiselect() seems to be a wider category than before 

```
public boolean canMultiselect() {
        return this == EDITOR_PICKER
                || this == AZTEC_EDITOR_PICKER
                || this == GUTENBERG_IMAGE_PICKER
                || this == GUTENBERG_VIDEO_PICKER
                || this == WP_STORIES_MEDIA_PICKER;
    }
```

It actually seems ok since this case seems linked to multiple mediaIds; maybe something you already considered but we should keep this in mind at least for tracking purposes I guess.

develric (commented):
Really like how you architected the media sources, good job @planarvoid  👍 !

I left one comment doubt re the uri: UriWrapper? being nullable before for your cross confirmation.

Also while testing I noticed the following, let me know wdyt thanks 🙇 

- When selecting from WP Media Lib as per test steps, I get the following error if I try to edit images from the media picker `9525-9525/org.wordpress.android.beta E/ss.android.bet: Invalid ID 0x00000000.`; this happens both if I select 1 or multiple images.
- This is not related to this PR I think: also when selecting from WP Media Lib we have the Browse For Items icon and if we select broswing from device the image is not added to the GB block. I was wondering if maybe we should not allow to pick from device when selecting from WP Media Lib? Wdyt? (I guess we can also address this in a dedicated PR eventually)

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerUiItem.kt
planarvoid: It's hard to say why it was nullable before, I think it's correctly non-null now because every item has to have an identifier.

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerActivity.kt
planarvoid: this is intentional. The same case now handles all the cases where the user can multiselect and there are multiple IDs. Looking at the code, I don't think we need the browser type check here. It should be enough that we have more than 1 ID to enter this case. 
We'll figure out tracking once we get there.

planarvoid:
thanks for the review @develric 

> When selecting from WP Media Lib as per test steps, I get the following error if I try to edit images from the media picker 9525-9525/org.wordpress.android.beta E/ss.android.bet: Invalid ID 0x00000000.; this happens both if I select 1 or multiple images.

edit: I've fixed this by removing the option to edit images from the media library (it wasn't allowed before)

> This is not related to this PR I think: also when selecting from WP Media Lib we have the Browse For Items icon and if we select broswing from device the image is not added to the GB block. I was wondering if maybe we should not allow to pick from device when selecting from WP Media Lib? Wdyt? (I guess we can also address this in a dedicated PR eventually)

I've fixed this and I've also removed the storage permissions check for the WP Media library. New testing steps for this scenario are: 

- Start with a fresh install of the app
- Log in -> Add GB post -> Add image block -> Add media
- Choose from WordPress media library
- You don't see the request for storage permissions
- You don't see the System picker in the toolbar
- You can select an image
- Go back
- Click on Add Media -> Choose from device
- You see the request for storage permissions
- Allow it
- You see the System picker in the toolbar

develric (approved):
Thanks for the changes @planarvoid 🙇 ! Looks good and works well! :shipit:

Issue #12544, opened 2020-07-28
Reader doesn't respect the Media & Text block settings to "Crop image to fill entire column" or "Show image on right" 


designsimply:
Steps to reproduce:

1. Create a new post.
1. Add a Media & Text block.
1. Tap "Add image or video".
1. Tap "Choose from Free Photo Library".
1. Search for, select, and insert an image.
1. Tapp "Show image on right" button.
1. Tap the cog icon to open block settings.
1. Turn on the "Crop image to fill entire column" option. 
1. Publish the post.
1. View the post in the Reader.

**Result:** in the Reader, the image is not cropped to fit the entire column and the image isn't shown on right (with text stacked above in a small screen view) as they are in a web view. ([2m42s](https://cloudup.com/cLUNl47N3Qw))

"Show image on right" setting|"Crop image to fill entire column" setting|Preview|Reader
---|---|---|---
<img width="480" alt="Screen Shot 2020-07-28 at Jul 28 5 15 22 PM" src="https://user-images.githubusercontent.com/1119271/88738572-96472a80-d0f6-11ea-8aec-76f77e95b8ef.png">|<img width="480" alt="Screen Shot 2020-07-28 at Jul 28 5 15 32 PM" src="https://user-images.githubusercontent.com/1119271/88738613-99dab180-d0f6-11ea-83a2-4712f515056d.png">|<img width="480" alt="Screen Shot 2020-07-28 at Jul 28 5 15 51 PM" src="https://user-images.githubusercontent.com/1119271/88738643-9c3d0b80-d0f6-11ea-878e-074db8333051.png">|<img width="480" alt="Screen Shot 2020-07-28 at Jul 28 5 21 10 PM" src="https://user-images.githubusercontent.com/1119271/88739039-c393d880-d0f6-11ea-9455-4831f91eed32.png">

<p style="clear:both;"><sup>Tested with WPAndroid 15.4-rc-1 on Pixel 3 Android 10.</sup></p>

zwarm:
@designsimply 
I wanted to recap what I am reading
1 - The text is not top stacking when right align (There is a prior rule for screens smaller than 660px that ignores left/right alignment requests - we've discussed this rule when we did the work for social media link icons)
2 - The image doesn't fill entire column

[...]

PR #12929, opened 2020-09-10
Add style for figure

zwarm:
Parent Issue #12544 

This PR adds a style for the `figure` element with 0 left/right margins. 
The problem in the parent issue manifested itself because the default user agent style was being used for `figure` which added the unnecessary left/right spacing

Before|After
---|---
|

To test:
Setup
- Create a new post.
- Add a Media & Text block.
- Tap "Add image or video".
- Tap "Choose from Free Photo Library".
- Search for, select, and insert an image.
- Tap the cog icon to open block settings.
- Turn on the "Crop image to fill entire column" option.
- Publish the post.

View the post in the reader
- With an account that is following the published post above
- Tap the Discover option
- Tap the following tab
- Navigate to the post created above
- Tap on the post to launch the detail view
- The image should fill the entire column

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostRenderer.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostRenderer.java
@@ -267,7 +267,6 @@ public class ReaderPostRenderer {
      */
     private String getPostContent() {
         String content = mPost.shouldShowExcerpt() ? mPost.getExcerpt() : mPost.getText();
-
         // some content (such as Vimeo embeds) don't have "http:" before links
         content = content.replace("src=\"//", "src=\"http://");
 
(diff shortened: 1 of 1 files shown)


malinajirka:
@zwarm Have we considered putting the style into the shared mobile css file?

zwarm:
> @zwarm Have we considered putting the style into the shared mobile css file?

This is only an issue on Android, which is why I went this way. I do know that iOS already overrides some user agent defaults locally, but I'll ask the iOS folks and get their thoughts on using the shared css. If moving to shared I will delete this PR, but will add a note if it remains here. Thanks for the suggestion.

@malinajirka - iOS wants to keep it local because they have stuff specific for safari browsers.

aforcier:
Thanks for the ping - I gave this a try combined with my in-progress changes and tested Stories since they also use a wrapping `figure` tag. All looks good on that front 👍

malinajirka (approved):
LGTM! Thanks for the fix and for considering using the shared css ;)

Issue #12900, opened 2020-09-07
Reader Improvements: Reader Detail Updates 


ashiagr:
Project Issue: https://github.com/wordpress-mobile/WordPress-Android/issues/12899

### Goal
Update the Reader Detail design

### Screenshots
![Reader Detail View (Android)](https://user-images.githubusercontent.com/1405144/92388963-c215f100-f135-11ea-8140-bfbc9d8a9a3f.png)

### Designs
- [Explorations](https://www.figma.com/file/a5O8CmzVbGbWBpqzrPUyiC/Reader-Detail-View?node-id=0%3A1)
- [Specs](https://www.figma.com/file/yG8ieha0yESnr8wCtplr29/Post-Detail-View?node-id=8%3A477) 

### Tasks

~~Change content type style (to be discussed)~~ No action needed

**BackLog**

Low Priority
- Recheck featured image visibility condition (https://github.com/wordpress-mobile/WordPress-Android/pull/12931#pullrequestreview-504709956)

Related to this PR: https://github.com/wordpress-mobile/WordPress-Android/pull/10611
Will fix separately. Tracked in #13239.

- Check for 2C & 2D in https://github.com/wordpress-mobile/WordPress-Android/pull/13058#issuecomment-705022900

Tests
Created separate issue to track it https://github.com/wordpress-mobile/WordPress-Android/issues/13156

**In Progress**

**Done**

- [x] App bar & Featured Image #12931
   - App bar
      - Show white icons (until scroll)
      - On scroll: white background with black icons
   - Featured Image
      - Show featured image edge-edge: No top, left, or right margins 
      - Add black to transparent linear gradient at the top of the image
      - Apply parallax effect on scroll (try adding featured image in a `CollapsingToolbarLayout`)
- [x] Post header #13042
   - Reposition post title(move from details section to header), follow button (next to blog section)
   - Blog section 
      - Present site URL more clearly along with the site name, date
      - Separate blog post author 
- [x] Tags #13049
   - Apply new topic pill styles (show tags in expandable tags view)
- [x] Post Header UI Polishing #13058
   - Apply new (icon only) follow button style
   - Post title: Update style
- [x] Sticky Action Bar #13106
   - Make the action bar sticky so that it doesn't disappear/reappear on scroll
   - Separate icons evenly
- [x] Remove the "liked by.." "more in.." sections #13107
- [x] Add the overflow/ellipsis menu to the header #13114
[...]

PR #12931, opened 2020-09-10
 Reader Detail Updates (App bar & Featured Image)

ashiagr:
Task: #12900

This PR includes below changes on the Reader Details screen:
   - App bar
      - Show white icons (until scroll) 01c43fc
      - On scroll: white background with black icons 01c43fc
      - Remove top app bar from the activity and define it in the details fragment - 141f15a
(Top app bar now includes collapsing toolbar that wraps a featured image wherever available. Since it changes across fragments based on featured image availability, it is moved from the activity to the individual ViewPager fragments.)
   - Featured Image
      - Show featured image edge-edge, parallax effect on scroll 0447ecf, 7de3f6c, da7a34d
      - Add black to transparent linear gradient at the top of the image d4f677d

[The PR is still in draft because of the known issues listed at the bottom. Looking for feedback for the rest of changes.]

featured_image_in_app_bar

Screenshots

Light Mode

Portrait | Landscape
--------|-----------
device-2020-09-15-105327|device-2020-09-15-105300

Dark Mode

Portrait | Landscape
--------|-----------
device-2020-09-15-105703|device-2020-09-15-105715

Known Issues

1. ~Status bar color - status bar is not reverted to original state (clicking search after returning from details screen shows status bar in white background instead of blue in light mode)~ - Fixed in 863d143

2. ~Featured image [height](https://github.com/wordpress-mobile/WordPress-Android/blob/863d14308cc5ec397526caeaaa9f43c5ab8dca1d/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostDetailFragment.kt#L1439) - should we change it to 16:9 aspect ratio? Since details screen is shown in full width in the landscape mode, we might want to retain fixed height so that it doesn't cover entire screen. Tentatively kept it equal to 40% of screen height similar to media settings screen.~ - retaining fixed height as per comment https://github.com/wordpress-mobile/WordPress-Android/pull/12931#issuecomment-703475513

3. Try to remove [isToolbarShowing](https://github.com/wordpress-mobile/WordPress-Android/blob/043bff53202eacd66a5d9c2fee33b20f853e6ae2/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostDetailFragment.kt#L176) - it is not dependent on the new toolbar but is still used to show/hide footer - will handle with new comments entry point changes

4. ~Yet to remove toolbar title~ - Fixed in 9d00059

PR submission checklist:


zwarm (commented):
@ashiagr - I really love the way this came together. Moving to toolbar to the fragment was key - the smooth transition feels like a unit. Nice work!

Only a couple of things to report
1. If I create a post with a featured image AND include a media text or image block, the featured image does not show. See `FeaturedImageUtils.showFeaturedImage` [line 31].
`postText.contains(featuredImagePath) && featuredImagePath != featuredImageFile -> false` 
I know you didn't add this check, but wanted to call it out because I'm not 100% sure what it needs to be checking for and this one probably needs to be addressed.

2. There is a brief moment when the toolbar title says "Wordpress reader". This is when you are swiping from right to left. It only happens the first time you swipe back to the post. Not sure you need to do anything here, but wanted to bring it up. See the attached gif.


Light

Other than that all is working as expected. Looks good on light/dark backgrounds. 

As for
> Since details screen is shown in full width in the landscape mode, we might want to retain fixed height so that it doesn't cover entire screen.

I am a +1 on retaining a fixed height, so that we don't disrupt the experience. I would double check on this one with Jirka.

malinajirka:
> Since details screen is shown in full width in the landscape mode, we might want to retain fixed height so that it doesn't cover entire screen.

> > I am a +1 on retaining a fixed height, so that we don't disrupt the experience. I would double check on this one with Jirka.

Sounds good to me ;). 

Great job!!!

ashiagr:
Thank you  for the review! Ready for another look @zwarm , @malinajirka 

> 1. If I create a post with a featured image AND include a media text or image block, the featured image does not show. See FeaturedImageUtils.showFeaturedImage [line 31].
postText.contains(featuredImagePath) && featuredImagePath != featuredImageFile -> false
I know you didn't add this check, but wanted to call it out because I'm not 100% sure what it needs to be checking for and this one probably needs to be addressed.

I'll check with Shakti on this, AFAIU we just need to move the existing featured image (with existing conditions) from the content section to the app bar. 

> 2. There is a brief moment when the toolbar title says "Wordpress reader". This is when you are swiping from right to left. It only happens the first time you swipe back to the post. Not sure you need to do anything here, but wanted to bring it up.

I've removed toolbar title as per design in 9d00059 which fixes this issue:

no-toolbar-title

Additionally, fixed another issue when toolbars were messed up on fragments accessed from notifications and comments activity in c4541a2.

zwarm (commented):
@ashiagr 
> I've removed toolbar title as per design 

Looks good and works as expected. When there is no featured image the missing title looks sparse, but that's just me experiencing the change. 

> I'll check with Shakti on this, AFAIU we just need to move the existing featured image (with existing conditions) from the content section to the app bar.

Just a bit more information on this one ... it has something to do with the path having `/wp` - if the path for the featured image includes `/wp`, and you add an image/media block and that image file has `/wp` in it's path, then it assumes it's the same file and `FeaturedImageUtils.showFeaturedImage` returns false. I don't know why it works off that. I don't even see the featured image in the content. Again, not something you did in this PR.

Everything else is looking good. 👍

ashiagr:
Thanks @zwarm !

> When there is no featured image the missing title looks sparse, but that's just me experiencing the change.

Hmm, Shakti is reviewing it in [this PR](https://github.com/wordpress-mobile/WordPress-Android/pull/13058). I'll check with her.

> Just a bit more information on this one ... it has something to do with the path having `/wp` - if the path for the featured image includes `/wp`, and you add an image/media block and that image file has `/wp` in it's path, then it assumes it's the same file and `FeaturedImageUtils.showFeaturedImage` returns false. I don't know why it works off that. I don't even see the featured image in the content. Again, not something you did in this PR.

I've added it to backlog [here](https://github.com/wordpress-mobile/WordPress-Android/issues/12900). I'll verify it later.

I think we can merge this PR to target branch as this condition wasn't updated in this PR. Lmkwyt.

zwarm (approved):
All good to merge. 👍  Again, nice work.

PR #12932, opened 2020-09-10
[Modal Layout Picker] Populate UI for WPCOM sites with data from the API

antonis:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2443

**Related PRs:**
- **Depends on**: https://github.com/wordpress-mobile/WordPress-Android/pull/12573
- **API call**: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1682

## Description

This PR populates the MLP UI with data fetched from `/wpcom/v2/sites/{site}/block-layouts`. It also implements a loading state where a view skeleton is displayed and a simple error state.

## To test:

### Regular Flow
1. With a good network connection
1. Open the app and select a WPCOM site
1. Open the Modal Layout Picker
1. **Expect** to see a brief ghost of while the data is fetched
1. **Expect** to see the screen updated with the fetched layouts
    - Layouts in the filter bar should be in alphabetical order
    - The sections in the table view should match the order of the categories in the filter bar
1. You should be able to select a layout from a category. 

### Slow Network
1. With a slow network connection (e.g. by changing the cellular configuration in an emulator).
1. Open the app and select a WPCOM site
1. Open the Modal Layout Picker.
1. **Expect** to see a ghost of empty cells while the data is fetched.
    - You should not be able to select a cell or category while the data is being fetched.
1. **Expect** to see the screen updated with the fetched layouts
    - Layouts in the filter bar should be in alphabetical order
    - The sections in the table view should match the order of the categories in the filter bar
1. You should be able to select a layout from a category. 

### No Network
1. Disable your device's connection ( e.g. airplane mode)
1. Return to the app and select a WPCOM site
1. Open the Modal Layout Picker.
1. **Expect** to see a ghost of empty cells while the data is fetched.
    - You should not be able to select a cell or category while the data is being fetched.
1. You should be able to still create a blank page
    - 📝We will be using cached data for this flow in the future.

## Screenshots:
| Light | Dark  |
|---|---|
|light|dark|

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/viewmodel/mlp/ModalLayoutPickerViewModel.kt
```diff
@@ -34,28 +46,62 @@ class ModalLayoutPickerViewModel @Inject constructor(
     private val _uiState: MutableLiveData<UiState> = MutableLiveData()
     val uiState: LiveData<UiState> = _uiState
 
+    private val contentUiState: ContentUiState
```
malinajirka: Could you please elaborate a bit why you decided to introduce this non-nullable field? AFAIU _uiState already holds the current state - which might not be ContentUiState -> won't this (contentUiState) field be out-of-sync when for example `uiState` is `LoadingUiState`? It'll technically hold an empty/invalid state, right?

WordPress/src/main/java/org/wordpress/android/viewmodel/mlp/ModalLayoutPickerViewModel.kt
```diff
@@ -3,27 +3,39 @@ package org.wordpress.android.viewmodel.mlp
...
  * Implements the Modal Layout Picker view model
  */
 class ModalLayoutPickerViewModel @Inject constructor(
+    private val dispatcher: Dispatcher,
+    private val siteStore: SiteStore,
+    private val appPrefsWrapper: AppPrefsWrapper,
     @Named(UI_THREAD) private val mainDispatcher: CoroutineDispatcher
 ) : ScopedViewModel(mainDispatcher) {
-    private val layouts: GutenbergPageLayouts = GutenbergPageLayoutFactory.makeDefaultPageLayouts()
+    private var layouts: GutenbergPageLayouts = GutenbergPageLayouts()
```
malinajirka: Could this field perhaps be `lateinit var`? Just asking so we limit mutability of the states as much as possible. Wdyt?

WordPress/src/main/java/org/wordpress/android/viewmodel/mlp/ModalLayoutPickerViewModel.kt
```diff
@@ -34,28 +46,62 @@ class ModalLayoutPickerViewModel @Inject constructor(
...
+
     fun init() {
-        updateUiState(ContentUiState())
-        loadLayouts()
-        loadCategories()
-        updateButtonsUiState()
+        fetchLayouts()
+    }
+
+    private fun fetchLayouts() {
```
malinajirka, line 75: We might want to consider moving this to a bg/io thread, wdyt?

WordPress/src/main/java/org/wordpress/android/viewmodel/mlp/ModalLayoutPickerViewModel.kt
```diff
@@ -34,28 +46,62 @@ class ModalLayoutPickerViewModel @Inject constructor(
...
+        } else {
+            layouts = GutenbergPageLayouts(event.layouts, event.categories)
+            loadLayouts()
+            loadCategories()
+        }
     }
 
     private fun loadLayouts() {
-        val state = uiState.value as ContentUiState
+        val state = contentUiState
```
malinajirka: We might want to consider moving this method to a bg/io thread as well, wdyt?

malinajirka (commented):
Thanks @antonis!

It looks good overall. I've left a couple of in-code comments. I also tested the app and I encounter a couple of scenarios about which I'm not sure about - I think they are expected, but just to be sure.

1. Should we perhaps always scroll to the top when the user changes a filter? It feels a bit weird when I'm scrolled to the bottom that items appear at the top of the list, wdyt?


2. I can select layouts when the image hasn't finished loading yet. I think this is fine, just wanted to double check it's expected.



3. I can select only one layout per category on some of my sites - is this expected?

WordPress/src/main/java/org/wordpress/android/viewmodel/mlp/ModalLayoutPickerViewModel.kt
antonis: I tried to avoid repeating the cast/check every time I wanted the `ContentUiState`. Thinking about this again making it non-nullable was not a good ideal and may cause more problems than it solves in the future.
Fixed with d968dd97eb86d7d85e7dc8069c84d45bccfe9d4e

antonis:
Hello @malinajirka 👋 ,

thank you for reviewing this 🙇 

> 1. Should we perhaps always scroll to the top when the user changes a filter? It feels a bit weird when I'm scrolled to the bottom that items appear at the top of the list, wdyt?

I agree that it feels more natural this way. Fixed with c5c5ebdd0aae1e5494fb1e184954c7e82d4a79c0

> 2. I can select layouts when the image hasn't finished loading yet. I think this is fine, just wanted to double check it's expected.

I was thinking of this too, but left it this way since the user may be able to preview the layout even if the thumbnail has not loaded.
Since you feel too that this needs handling I tried to cover it with d0127344944b1a4f26cc550c423c8d429c730b4d

> 3. I can select only one layout per category on some of my sites - is this expected?

The data are loaded from the server for each site. I guess this is normal but I will double check. Note that in a next PR I will filter out layouts that are not applicable so this "one layout per category" may be a common case.

WordPress/src/main/java/org/wordpress/android/viewmodel/mlp/ModalLayoutPickerViewModel.kt
antonis: I agree. Fixed with 45717af5b19f60ef468295e7d04d7321fcb28368

WordPress/src/main/java/org/wordpress/android/viewmodel/mlp/ModalLayoutPickerViewModel.kt
antonis: I agree. Fixed with 05ca9af80c966cdba96a30b4335a4a04bf31901e

WordPress/src/main/java/org/wordpress/android/viewmodel/mlp/ModalLayoutPickerViewModel.kt
antonis: As above fixed with 05ca9af80c966cdba96a30b4335a4a04bf31901e

WordPress/src/main/java/org/wordpress/android/viewmodel/mlp/ModalLayoutPickerViewModel.kt
malinajirka: It seems the code is still running on the main thread. We invoke `updateUiState(state.copy(layoutCategories = listItems))` which directly sets value to uiState livedata and the app doesn't crash. Since we extend `ScopedViewModel` I think we can drop `viewModelScope` before the `launch`. We'll need to wrap "updateUiState" with "withContext(main)". Wdyt?

WordPress/src/main/java/org/wordpress/android/viewmodel/mlp/ModalLayoutPickerViewModel.kt
malinajirka: Same as above

malinajirka:
Thanks @antonis! I've left one comment but the changes look great otherwise. I tried to move the execution to a bg thread, but the app didn't work as expected. So I'll wait for your fixes and test it after. Thanks!

> Note that in a next PR I will filter out layouts that are not applicable so this "one layout per category" may be a common case.

We are targeting develop in this PR, is it intentional?

WordPress/src/main/java/org/wordpress/android/viewmodel/mlp/ModalLayoutPickerViewModel.kt
antonis: You are right. Please check b3fc5e3e39f01fa443c7d3264b28da7df6da74d2

WordPress/src/main/java/org/wordpress/android/viewmodel/mlp/ModalLayoutPickerViewModel.kt
antonis: As above

antonis:
Hello @malinajirka 👋 ,
thank you for your feedback 🙇 

> We are targeting develop in this PR, is it intentional?

Yes. This feature is disabled with `BuildConfig.MODAL_LAYOUT_PICKER` by default (enabled only for `wasabi` and `jalapeno`)

malinajirka (approved):
> Yes. This feature is disabled with BuildConfig.MODAL_LAYOUT_PICKER by default (enabled only for wasabi and jalapeno)

:D I should wake up before I post comments next time :P. 

Thanks for all the changes @antonis! LGTM

PR #12935, opened 2020-09-11
Issue/12915 permission label overlap

### LINKED ISSUE #12915: Overlapped labels when accessing new media picker without permissions active.
This was noted while working on the #12884 PR and is related to the new media picker

### Expected behavior
After a clean install, when opening the media picker for the first time we should get the following message that asks for permissions.

<p align="center">
  
</p>

### Actual behavior
After a clean install, when opening the media picker for the first time we get the following overlapped message that asks for permissions.

<p align="center">
  
</p>

- Activate the CONSOLIDATED_MEDIA_PICKER flag for wasabi
- Clean install the WasabiDebug flavor and create a new post
- Add a Image block and open the picker to add an image
- note you the the following image with overlapped labels

<p align="center">
  
</p>

develric:
Fixes #12915 

### To test

- Make a clean install, 
- Create a new post and add an Image Block
- Open the media picker to add an image from device
- Check you get the following image without overlapping labels
<p align="center">
  
</p>

- Press the `Allow` button and grant permission
- Note the empty view is briefly shown while background is loading images
- Note the images are loaded and empty view is hidden
- Use the search and type a string with some matched images. Note the matched images are shown
- Use the search and type a string with no matched images. Note the empty state image is shown

<p align="center">
  
</p> 

- Smoke test the picker

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerFragment.kt
@@ -260,11 +260,16 @@ class MediaPickerFragment : Fragment() {
                         requestStoragePermission()
                     }
                 }
+
+                recycler.setEmptyView(null)
+                actionable_empty_view.visibility = View.GONE
+
                 soft_ask_view.visibility = View.VISIBLE
             }
             is SoftAskViewUiModel.Hidden -> {
                 if (soft_ask_view.visibility == View.VISIBLE) {
                     AniUtils.fadeOut(soft_ask_view, MEDIUM)
+                    recycler.setEmptyView(actionable_empty_view)
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerFragment.kt
```diff
@@ -260,11 +260,16 @@ class MediaPickerFragment : Fragment() {
                         requestStoragePermission()
                     }
                 }
+
+                recycler.setEmptyView(null)
+                actionable_empty_view.visibility = View.GONE
```
planarvoid: I think the correct solution would be to move this logic to the `PhotoListViewModel`. We can introduce a new subclass of the `PhotoListUiModel` (we have `Data`, `Empty` and we can add `Hidden`). The `build` method can decide to set it to `Hidden` based on the same check used `if (softAskRequest != null && softAskRequest.show)`. This moves the decision out of the fragment. WDYT?

develric: I think you are right 👍 !

develric:
Hey @planarvoid , thanks for the review. I made changes as per your feedback 👍 ! 

Re the 

```
fun `empty state is emitted when no items in picker`()
```

I needed to modify the flow behavior in the `setupViewModel` function. Not sure it is the best approach tbh so let me know wdyt 🙇

planarvoid (approved):
Looks good 👍 . I've tested the flow and it works well, thanks!

PR #12939, opened 2020-09-11
Mobile Stories block

mzorz:
This PR adds support for the jetpack Story block to be rendered on gutenberg-mobile, and provides the bridge implementation on the WPAndroid side of things to connect the block actions to the Story composer.

Related Jetpack PR: https://github.com/Automattic/jetpack/pull/17140
Related Gutenberg PR: https://github.com/WordPress/gutenberg/pull/25242
Related Gutenberg-mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/2611

#### test block rendering
1. create a story
2. publish it
3. now go to the posts list
4. open it
5. observe the block appears (it's not unsupported but an actual story block).

#### Test block editing
CASE A: the story is in memory (no prefs are used)
1. create a story
2. publish it
3. open the post on gutenberg
4. tap on the block's tool icon
5. it should load on the StoryComposer from memory

CASE B: the story has been created on this device, not in memory
1. create a story
2. publish it
3. kill the app and open it again
4. open the post on gutenberg
5. tap on the block
6. it should load on the StoryComposer from reviving it from prefs (added views should be editable)

CASE C: the story has not been created on this device
1. create a story
2. publish it
3. log in on another device (or uninstall the app and then re-install it)
4. open the post on gutenberg
5. tap on the block
6. it should load on the StoryComposer with the flattened views only (no added views, only background media)

CASE C1: story created on another device, media not present on current device
1. go to a device A and create a story, publish it
2. go to a device B and, WITHOUT refreshing the media section of the site, go straight to Posts list -> refresh (or wait) and tap on the newly created Post
3. observe the post loads
4. when you tap on the tool icon for the story block, it gets loaded and the "Limited story editing" dialog appears.

CASE C2: story created on another device, media not present on current device, NO NETWORK
1. go to a device A and create a story, publish it
2. go to a device B and, WITHOUT refreshing the media section of the site, go straight to Posts list -> refresh (or wait) until it appears on the list
3. turn airplane mode ON
4. tap on the newly created Post (if you're debugging, a red screen will appear on Gutenberg because it can't connect to metro - at this point, turn airplane mode OFF again, as it would have gotten past the media fetching code, now tap on the "Reload" option at the bottom right of the screen and allow the block to render the remote site http URL even if MediaModels haven't been fetched)
5. observe the post loads
6. when you tap on the tool icon for the story block, it cannot load and the following Dialog appears.

Screen Shot 2020-09-18 at 09 53 51

PR submission checklist:


.idea/codeStyles/Project.xml
```diff
@@ -32,16 +32,7 @@
     <JetCodeStyleSettings>
       <option name="PACKAGES_TO_USE_STAR_IMPORTS">
         <value>
-          <package name="kotlinx.android.synthetic" alias="false" withSubpackages="true" />
```
jd-alexander, line 35: Hi @mzorz do we want to include these changes here? They are related to how different versions of Android Studio manipulates this file right?

mzorz: you're right! I think it's handled in a later PR as well but we can definitely remove these changes from this PR in particular if it's creating noise

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
```diff
@@ -719,6 +722,41 @@ public static void addNewStoryWithMediaUrisForResult(
         activity.startActivityForResult(intent, RequestCodes.CREATE_STORY);
     }
 
+    public static void editStoryWithMediaIdsForResult(
```
jd-alexander, line 725: I can see where this method comes into use in the follow-up PR 👍

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -3049,6 +3065,51 @@ public void onTrackableEvent(TrackableEvent event, Map<String, String> propertie
         mEditorTracker.trackEditorEvent(event, mEditorFragment.getEditorName(), properties);
     }
 
+    @Override public void onStoryComposerLoadRequested(ArrayList<Object> mediaFiles, String blockId) {
```
jd-alexander, line 3069: Hi @mzorz I can see where this PR and the follow up is adding more code to this activity. Let's try to refactor some of this logic out into another class so we don't make this class any bigger since it's in the 3000s now 😅 All the `dispatcher`/ `EventBus` related logic can be placed in another class like how we did it [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/posts/PostListEventListener.kt)
If you have a reason why it has to be placed here then we can leave it but if we can extract then that would be good. Wdyt? 😄

.idea/codeStyles/Project.xml
jd-alexander: Okay if it handled in the later one then it's fine since we are merging these changes into the feature branch 👍

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -613,6 +621,13 @@ protected void onCreate(Bundle savedInstanceState) {
         }
 
         if (!mIsNewPost) {
+            // if we are opening an existing Post, and it contains a Story block, pre-fetch the media in case
```
jd-alexander, line 625: I would want to move this logic and the `fetchMediaList` function into the class we could use to consolidate all the dispatcher calls and events etc. LMK 😄

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -3049,6 +3065,51 @@ public void onTrackableEvent(TrackableEvent event, Map<String, String> propertie
...
+                allStorySlidesAreEditable = result.getAllStorySlidesAreEditable();
+            }
+            storyIndex = result.getStoryIndex();
+        }
+
+        if (!failedLoadingOrReCreatingStory) {
+            // Story instance loaded or re-created! Load it
+            ActivityLauncher.editStoryForResult(this, mSite, storyIndex, allStorySlidesAreEditable, true);
+        } else {
+            // unfortunately we couldn't even load the remote media Ids indicated by the StoryBLock so we can't allow
```
jd-alexander:
```suggestion
            // unfortunately we couldn't even load the remote media Ids indicated by the StoryBlock so we can't allow
```

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -3049,6 +3065,51 @@ public void onTrackableEvent(TrackableEvent event, Map<String, String> propertie
...
+                mMediaStore,
+                this
+        );
+        ArrayList<String> mediaIds =
+                loadStoryFromStoriesPrefsUseCase.getMediaIdsFromStoryBlockBridgeMediaFiles(mediaFiles);
+        boolean allStorySlidesAreEditable = loadStoryFromStoriesPrefsUseCase.areAllStorySlidesEditable(mSite, mediaIds);
+        boolean failedLoadingOrReCreatingStory = false;
+
+        // now look for a Story in the StoryRepository that has all these frames and, if not found, let's
+        // just build the Story object ourselves to keep these files arrangement
```
jd-alexander: When you say `build the Story object ourselves to keep these files arrangement` what do you mean by the `files arrangement` part?

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -3081,6 +3142,13 @@ public void onMediaUploaded(OnMediaUploaded event) {
 
     // FluxC events
 
+    @SuppressWarnings("unused")
+    @Subscribe(threadMode = ThreadMode.MAIN)
+    public void onMediaListFetched(OnMediaListFetched event) {
```
jd-alexander, line 3141: Good choice 👍

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/AddLocalMediaToPostUseCase.kt
```diff
@@ -67,6 +67,14 @@ class AddLocalMediaToPostUseCase @Inject constructor(
                 optimizeMediaResult.optimizedMediaUris
         )
 
+        // here we pass a map of "old" (before optimisation) Uris to the new MediaModels which contain
+        // both the mediaModel ids and the optimized media URLs.
+        // this way, the listener will be able to process from other models potining to the old URLs
```
jd-alexander:
```suggestion
        // this way, the listener will be able to process from other models pointing to the old URLs
```

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -3049,6 +3065,51 @@ public void onTrackableEvent(TrackableEvent event, Map<String, String> propertie
         mEditorTracker.trackEditorEvent(event, mEditorFragment.getEditorName(), properties);
     }
 
+    @Override public void onStoryComposerLoadRequested(ArrayList<Object> mediaFiles, String blockId) {
+        LoadStoryFromStoriesPrefsUseCase loadStoryFromStoriesPrefsUseCase = new LoadStoryFromStoriesPrefsUseCase(
```
jd-alexander: I would make this use case injectable instead of instantiating it here to improve testability. WDYT?

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -3206,6 +3274,17 @@ private void refreshEditorTheme() {
         mDispatcher.dispatch(EditorThemeActionBuilder.newFetchEditorThemeAction(payload));
     }
 
+    private void fetchMediaList() {
+        // do not refresh if there is no network
+        if (!NetworkUtils.isNetworkAvailable(this)) {
```
jd-alexander, line 3274: Just a note here, I realized that I don't see many usages of this approach where we check to see if network connectivity exists before utilizing the `dispatcher`. Do you know why this is so?

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
```diff
@@ -0,0 +1,128 @@
...
+
+import android.content.Context
+import android.net.Uri
+import androidx.preference.PreferenceManager
+import com.wordpress.stories.compose.story.StoryFrameItem
+import com.wordpress.stories.compose.story.StoryFrameItem.BackgroundSource.FileBackgroundSource
+import com.wordpress.stories.compose.story.StoryFrameItem.BackgroundSource.UriBackgroundSource
+import com.wordpress.stories.compose.story.StorySerializerUtils
+
+object StoriesPrefs {
```
jd-alexander: Hi @mzorz any specific reason why this class was made a Singleton since we can utilize Dagger to create a `@Singleton` scope and we could make it testable? LMK 🙏

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
```diff
@@ -0,0 +1,128 @@
...
+            val inputStream = context.contentResolver.openInputStream(uri)
+            if (inputStream != null) {
+                inputStream.close()
+                return true
+            }
+        } catch (e: java.lang.Exception) {
+            e.printStackTrace()
+        }
+        return false
+    }
```
jd-alexander: Nitpick : This could probably be simplified like so 😉 
```suggestion
        context.contentResolver.openInputStream(uri).use { stream -> 
            return stream != null
        }
```

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
```diff
@@ -0,0 +1,128 @@
...
+        val editor = PreferenceManager.getDefaultSharedPreferences(context).edit()
+        editor.remove(slideIdKey)
+        editor.apply()
+    }
+
+    fun deleteSlideWithRemoteId(context: Context, siteId: Long, mediaId: RemoteMediaId) {
+        val slideIdKey = buildSlideKey(siteId, mediaId)
+        val editor = PreferenceManager.getDefaultSharedPreferences(context).edit()
+        editor.remove(slideIdKey)
+        editor.apply()
```
jd-alexander: We can simplify these with a bit of kotlin goodness :) 
```suggestion
        PreferenceManager.getDefaultSharedPreferences(context).edit().apply { 
            remove(slideIdKey)
            apply()
        }
```

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    data class RemoteMediaId(val mediaId: Long)
+    data class LocalMediaId(val id: Long)
```
jd-alexander: Could the `LocalOrRemoteId` based in FluxC be used instead of creating these? Just checking 😄 

Also if we are sticking with these we could make the parameter names consistent, meaning both would be either `mediaId` or `id`

WordPress/src/main/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCase.kt
```diff
@@ -0,0 +1,98 @@
...
+import javax.inject.Inject
+
+@Reusable
+class LoadStoryFromStoriesPrefsUseCase @Inject constructor(
+    private val storyRepositoryWrapper: StoryRepositoryWrapper,
+    private val site: SiteModel,
+    private val mediaStore: MediaStore,
+    private val context: Context
+) {
+    fun getMediaIdsFromStoryBlockBridgeMediaFiles(mediaFiles: ArrayList<Object>): ArrayList<String> {
```
jd-alexander:
```suggestion
    fun getMediaIdsFromStoryBlockBridgeMediaFiles(mediaFiles: ArrayList<Any>): ArrayList<String> {
```

WordPress/src/main/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCase.kt
```diff
@@ -0,0 +1,98 @@
...
+                    .toLong()
+            val mediaIdString = mediaIdLong.toString()
+            mediaIds.add(mediaIdString)
+        }
+        return mediaIds
+    }
+
+    fun areAllStorySlidesEditable(site: SiteModel, mediaIds: ArrayList<String>): Boolean {
+        for (mediaId in mediaIds) {
+            if (!isValidSlide(context, site.getId().toLong(), RemoteMediaId(mediaId.toLong()))) {
```
jd-alexander:
```suggestion
            if (!isValidSlide(context, site.id.toLong(), RemoteMediaId(mediaId.toLong()))) {
```

WordPress/src/main/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCase.kt
```diff
@@ -0,0 +1,98 @@
...
+                allStorySlidesAreEditable = false
+
+                // for this missing frame we'll create a new frame using the actual uploaded flattened media
+                val tmpMediaIdsLong = ArrayList<Long>()
+                tmpMediaIdsLong.add(mediaId.toLong())
+                val mediaModelList: List<MediaModel> = mediaStore.getSiteMediaWithIds(
+                        site,
+                        tmpMediaIdsLong
+                )
+                if (mediaModelList.size == 0) {
```
jd-alexander:
```suggestion
                if (mediaModelList.isEmpty()) {
```

jd-alexander (changes requested):
Thanks for these changes @mzorz Great work on this integration so far. I performed all the test cases you described above using two emulators and the expected results were realized. 🔥 🔥 🔥 

- [x] test block rendering 
- [x] Test block editing
    - [x] CASE A: the story is in memory (no prefs are used)
    - [x] CASE B: the story has been created on this device, not in memory
    - [x] CASE C: the story has not been created on this device
    - [x] CASE C1: story created on another device, media not present on current device
    - [x] CASE C2: story created on another device, media not present on current device, NO NETWORK

For CASE C2, I was wondering if the dialog is providing enough context for the user. It's stating that the story was created on a device and can't be edited at this moment even though if you go back and refresh the list and re-enter it works. I would put a dialog that says a connectivity error ocurred and put a retry button so the media can be fetched without using leaving the editor. Let me know your thoughts on this.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
mzorz: That's a smart idea Joel! (using the lifecycleobserver and externalizing the logic for eventbus register/unregister and listening out to a different class, as shown in the PostsListEvent example). In this case, `onStoryComposerLoadRequested` is not part of EventBus / dispatcher code, it's an interface implementation which connects Gutenberg to the containing Activity (EditPostActivity). So, the interface override I'd say needs to stay there, but we could move the logic within that method override out, to keep EditPostActivity shorter.

Done in 0f4d87c25b 👍

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
mzorz: I understand the motivations as said above - however having moved the essential logic of loading from prefs in https://github.com/wordpress-mobile/WordPress-Android/commit/0f4d87c25bde9da952d91fa9a5902a189acc0ed2, I'd argue there's little to nothing to gain from moving this call to a new class, for which also its results we don't care about listening to anyway (see how the [FluxC event listener is empty and only contains a comment](https://github.com/wordpress-mobile/WordPress-Android/pull/12939/files#diff-38b6aea9cfff5f3e0e4b7389b44de602R3123-R3127), and we could even completely remove if we wanted to). Let me know if you had something different in mind, I may be misunderstanding. 🙏

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
mzorz: so basically is the newly re-created Story should match the same order in which the media files being passed came in in the first place. That's what I meant by files arrangement

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
mzorz: Good idea, done in https://github.com/wordpress-mobile/WordPress-Android/commit/0f4d87c25bde9da952d91fa9a5902a189acc0ed2

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
mzorz: I think I thought of doing something else in this case but, since we don't really need to do anything in particular in this regard (it's just a background refresh), we could also just attempt and let it fail. No big deal 🤷

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
mzorz: the reason is: I keep forgetting, thanks a ton for reminding me an being on top of this, covering our back :D 
Made the needed changes in f6d92d293c

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
mzorz: good idea; also noted we need to call `inputStream.close()` before returning so, ended up doing this in 34d4effeac:

```
            context.contentResolver.openInputStream(uri)?.let {
                it.close()
                return true
            }
```

WordPress/src/main/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCase.kt
mzorz: changed in 99efe00984 👍

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
mzorz: Good catch - done in ddcda8b069 👍

mzorz:
Thanks a lot for your very insightful review @jd-alexander ! Addressed all your comments, ready for another round 🙇 .

> For CASE C2, I was wondering if the dialog is providing enough context for the user. It's stating that the story was created on a device and can't be edited at this moment even though if you go back and refresh the list and re-enter it works. I would put a dialog that says a connectivity error ocurred and put a retry button so the media can be fetched without using leaving the editor. Let me know your thoughts on this.

I think what the dialog says is still true: a) the story has been created on another device (well, not entirely true if you re-installed, but most likely), and b) we can't allow editing at this moment.
So regarding the second part "can't be edited at this moment" is still valid if you refresh and enter back again and now are able to edit it, right?

I understand the case for ambiguity you're pointing out; so I wouldn't tie this to connectivity issues because it could be something else as well. 

What about `This story may have been created on a different device. We'll be able to allow limited editing as soon as media related to this Story has been refreshed on this device. Please, try again later`. ? also cc @aforcier

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
mzorz: forgot to mention -  looking into the other PR, I'm creating the suggested class and refactor the listeners there in #12982 👍

mzorz: the suggested refactor effectively made available in the PR after this one (#12982) in this commit https://github.com/wordpress-mobile/WordPress-Android/pull/12982/commits/a6dee817a13eddd0aecfbda821b5d0767dc8f508 @jd-alexander 👍

aforcier:
> What about This story may have been created on a different device. We'll be able to allow limited editing as soon as media related to this Story has been refreshed on this device. Please, try again later. ? also cc @aforcier

If I understand the situation, the 'created on a different device' part seems not important here - we can't let the user edit the story because we aren't able to download the media. (If it had been created on the device, we'd have the media, but that's probably not a connection the user will/should make.)

So what about just:

> Unable to load media for this story. Check your internet connection.

You did mention that

> I understand the case for ambiguity you're pointing out; so I wouldn't tie this to connectivity issues because it could be something else as well.

What else could cause this state? Unless this is often flat wrong, I think pointing to the connection is a good hint at what would have to change before this screen actually works, and we can add more specific messaging if we later have other cases that lead to this state. WDYT?

mzorz:
> So what about just:
> "Unable to load media for this story. Check your internet connection."

When this situation is reached (media indicated by the story block can't be found), they will only be able to successfully retrieve the media if **both** these conditions are met:

a) they go out of the Editor and back in, with proper connectivity to their site
b) the media pointed to by the Story block still exists (they could have been deleted from the site altogether)

So I would be extra cautious (that's why I said I wouldn't tie this to connectivity issues only because it could be something else as well and we'd be asking them to refresh forever when it may actually be an unrecoverable state).

What about this:

Title: `Can't edit story` (as it is)
Message: `This story may have been created on a different device, and as such the related media couldn't be found. We'll be able to allow limited editing after this Story's media is downloaded on this device. Verify your internet connection and if the problem persists, try going to the Media section on the app and refresh the list there.`

This is still long I know 😅 , if you can come up with something shorter that still conveys the message politely that'd be great 👍 🙏 


**EDIT**
To make it clear, these are the possible situations:
1. we have both serialized slides and local background sources: we allow editing everything (no dialog is presented)
2. we don't have the serialized slides, but we have access to the media pointed to by the block: we allow "limited story editing" (dialog shows up in the Story composer because things suggest you created this on another device or re-installed the app)
3. we don't have the serialized slides nor can we find the media referenced by the block: we show this dialog.

We could also for example catch the `fetchMediaList()` FluxC's `OnMediaListFetched` event in EditPostActivity when we tried refreshing and, if it all went well and we still can't find the media, tell them this is a permanent error (there's really nothing we can do at this point), and if there was an error in the fetch, still tell them about the connectivity issues as expressed above (I'd go with some shorter version of my message or, the original message would just suffice at this point?).

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
jd-alexander: @mzorz understood. Thank you. I think it would be good to use this explanation in the comment instead of `files arrangement` as it's more clear. So something like 

```
 // now look for a Story in the StoryRepository that has all these frames and, if not found, let's
 // just build the Story object ourselves to match the order in which the media files were passed. 
```

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
jd-alexander: Okay true! 😄

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
jd-alexander: Nice. Thank you 🙏

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
jd-alexander: Oh yes, I totally agree with this. After looking at 0f4d87c I understand your motivations and this can remain.

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
jd-alexander: More than happy to help! 🙇

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
jd-alexander: nice. Good going 👍

jd-alexander:
Thanks for expounding in more detail here @aforcier @mzorz 

> So I would be extra cautious (that's why I said I wouldn't tie this to connectivity issues only because it could be something else as well and we'd be asking them to refresh forever when it may actually be an unrecoverable state).

I totally get this 🙏 As you described further in your comment, what we can do is show the dialog with errors we are sure of, and then for the "something else" case we just have a default dialog. I agree with @aforcier about adding more specific messages about what leads to a particular state. Just to further the discussion some more, can we do any type of logging to figure out the other cases that would cause this to fail? Also, we want to track how many failures are taking place here as well versus successes so we can make incremental improvements to have all cases covered.

> This story may have been created on a different device, and as such the related media couldn't be found. We'll be able to allow limited editing after this Story's media is downloaded on this device. Verify your internet connection and if the problem persists, try going to the Media section on the app and refresh the list there.

This would be good for a "More Info" action but not for the dialog itself as you have stated it's too long 😅 but yes, this is helpful. As a user using the story functionality I wouldn't want to go to the media of the site and refresh the list there. Is there anything we can do from within the `EditPostActivity` to do that retry mechanism automatically? 

> We could also for example catch the fetchMediaList() FluxC's OnMediaListFetched event in EditPostActivity when we tried refreshing and, if it all went well and we still can't find the media, tell them this is a permanent error (there's really nothing we can do at this point), and if there was an error in the fetch, still tell them about the connectivity issues as expressed above (I'd go with some shorter version of my message or, the original message would just suffice at this point?).

This approach sounds like a good one because one thing I know for sure is that once it comes to media downloads and syncing there can be several issues and we wouldn't want any of them to cause the user too much frustration when they have created their story. So optimizing the error handling and it's communication as much as possible will definitely save us in the long term. 

To summarize my thoughts, we can utilize short & specific error messages based on the error that led to a particular state eg. 

1. @aforcier 's  `Unable to load media for this story. Check your internet connection.`
2. `We couldn't find the media for your story on this site.` ( for this we would need to probably exit the editor 😢 ) 

Doing this will allow us to have good fallback mechanisms and context so users can figure out what's going wrong with the story they are created so it can be resolved and lead them to successful edits in the future.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
jd-alexander: Nice work @mzorz. Thanks for this change 🙇

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
mzorz: done in cadf22b4df

mzorz:
Alright! Thank you for all the brainstorming @jd-alexander @aforcier 🙇 

I implemented the following logic in 62fec15ade:

1. when the Editor shows up we issue a request to update the site's media list. If there's no network prior to that (so we can't make sure this is a fresh item) or in the case the result event from FluxC is an error, we set a boolean flag internally. Then when the user actually decides to try and edit a Story block, we check this flag and show @aforcier's suggested message, slightly tweaked `Unable to load media for this story. Check your internet connection and try again in a moment.`. Basically: we didn't find the backing media but it may still be the case that the media exists on the remote site but we just haven't been able to refresh locally. **Also**: when the user presses OK on the dialog, we issue another fetchMediaList request _(we could potentially implement the UI for retrying and showing progress to the user but I think it's overkill for a corner case they can circumvent also by exiting the editor and re-entering it, provided they do have connectivity this time)._

2. in the case the flag didn't return any errors but still we can't find the media Id being pointed to by the block, we do now show the second error as suggested by @jd-alexander 
`We couldn't find the media for your story on this site.` . Note we don't necessarily need to exit the editor, as people may still remove the block or add other blocks, etc.


To test:

For point 1 above (potentially retriable error):
1. go to a device A and create a story, publish it
2. go to a device B and, WITHOUT refreshing the media section of the site, go straight to Posts list -> refresh (or wait) until it appears on the list
3. turn airplane mode ON
4. tap on the newly created Post (if you're debugging, a red screen will appear on Gutenberg because it can't connect to metro - at this point, turn airplane mode OFF again, as it would have gotten past the media fetching code, now tap on the "Reload" option at the bottom right of the screen and allow the block to render the remote site http URL even if MediaModels haven't been fetched)
5. observe the post loads
6. when you tap on the tool icon for the story block, it cannot load and the following Dialog appears.

Screen Shot 2020-10-08 at 16 18 48


For point 2 above (permanent error) - after doing the above:
1. on device A, now go to the media section of the app and delete the backing media for the story you created on step 1 before.
2. go to device B and make sure you have airplane mode OFF now, exit the editor and on the Posts list, tap on the same Post again to open it.
3. The media list should have been refreshed now in the background
4. now select the Story block and tap on block's tool icon
5. the app won't find the slides nor backing media and as such can't proceed to allow editing; this is the dialog shown:

Screen Shot 2020-10-08 at 16 21 48

Should be ready for another round, and let me know your thoughts 🙇 @jd-alexander @aforcier

WordPress/src/main/res/values/strings.xml
```diff
@@ -2882,6 +2882,13 @@
     <string name="mlp_selected_description">%s selected</string>
     <string name="mlp_notselected_description">%s</string>
 
+    <!-- Stories -->
+    <string name="dialog_edit_story_limited_title">Limited Story Editing</string>
+    <string name="dialog_edit_story_limited_message">This story was edited on a different device and the ability to edit certain objects may be limited.</string>
+    <string name="dialog_edit_story_unavailable_title">Can\'t edit Story</string>
+    <string name="dialog_edit_story_unavailable_message">Unable to load media for this story. Check your internet connection and try again in a moment.</string>
+    <string name="dialog_edit_story_unrecoverable_title">Can\'t edit Story</string>
+    <string name="dialog_edit_story_unrecoverable_message">We couldn\'t find the media for your story on this site.</string>
```
aforcier: I think we should avoid 'your', this isn't necessarily the author of the story. How about this, also more consistent with the other error:

```suggestion
    <string name="dialog_edit_story_unrecoverable_message">We couldn\'t find the media for this story on the site.</string>
```

aforcier (changes requested):
I think the solution for the messaging is great, nicely done 👏

I tried out the new changes and re-tested the overall flow, everything seems to be working smoothly 🎉 Just made one wording suggestion.

I'll defer to @jd-alexander for the final :shipit:

WordPress/src/main/res/values/strings.xml
jd-alexander: This is a good point @aforcier Thanks for spotting 🙇

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -3081,6 +3136,16 @@ public void onMediaUploaded(OnMediaUploaded event) {
 
     // FluxC events
 
+    @SuppressWarnings("unused")
+    @Subscribe(threadMode = ThreadMode.MAIN)
+    public void onMediaListFetched(OnMediaListFetched event) {
+        // no op - we don't need to check anything just now, but declaring the method so it's
+        // clear we make a request to FetchMedia in this class.
```
jd-alexander: @mzorz do we still want this comment here like how we have logic below? LMK 🙏

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
```diff
@@ -0,0 +1,129 @@
...
+                    source.contentUri?.let {
+                        return isUriAccessible(it)
+                    } ?: return false
+                }
+            }
+        }
+        return false
+    }
+
+    private fun isUriAccessible(uri: Uri): Boolean {
```
jd-alexander, line 59: I can't help but think that this probably doesn't belong here 😅 I know we have `UriUtils`  in `WordPressUtils` so I am wondering if it's best to put it there? You would lose the kotlin goodness 😉 but it would be in a place where it's reusable. WDYT?

mzorz: That's certainly something that crossed my mind as well, but given it also involves some administrativia with other repos and is not particularly needed anywhere else (can you think of a real use case where this would be needed elsewhere in WP?), I think it's fine to keep it here where there's a meaning to checking for whether the background source is accessible or not (it really helps the decision making about allowing editing or not in our case). If you'd like to, we can open an issue in the meantime instead so to focus on the rest of PRs, does that make sense? 🙏

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
mzorz: good catch  - removed in commit 13e5bfd2ca

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
jd-alexander: Understood! We can keep it here!

jd-alexander:
@mzorz I am giving this a test now. I just saw the hash update so I will pull that down to ensure I have the latest changes 🙏

jd-alexander:
@mzorz I was trying to do the final review here but for some reason, the content of my post is of the initial HTML so I am wondering if I did something is off with the way how I approached this. I will revisit.

<kbd></kbd>

mzorz:
> @mzorz I was trying to do the final review here but for some reason, the content of my post is of the initial HTML so I am wondering if I did something is off with the way how I approached this. I will revisit.

Thanks for spotting that one - that's fixed in https://github.com/wordpress-mobile/gutenberg-mobile/pull/2705 - will cherry pick and update this branch here so you can continue reviewing 👍

jd-alexander (approved):
Thanks for these changes @mzorz tested both paths and they work as expected. This is a great improvement. LGTM 🚢

PR #12940, opened 2020-09-11
Login: Replace endpoint that fetches site info when using a site address to login

### LINKED ISSUE #12832: Unified Login: site address login fails with "Error while adding site. Error code GENERIC_ERROR" for private custom-domain WordPress.com sites using a 2FA login
### Expected behavior

I would expect that I could log in with the new unified login flow using my site address for a private WordPress.com site with a custom domain or be directed in the correct direction to be able to login.

### Actual behavior

When I try to log in with my site address for a WordPress.com site using a custom domain, I get a generic error message. 

1. Install the app fresh.
2. Use the Enter Site Address of a WordPress.com private site using a custom domain (this may require an account with 2FA enabled)
3. Enter your log in credentials in the next screen and see the Generic Error screen.
Image 2020-08-29 at 1 54 03 pm

Note that on iOS I got a pop-up error at this same stage with the notice `this account has two-step authentication enabled. Please configure an application-specific password to use with this application` so I tried doing this on Android and was able to log in to the site using the application-specific password. 

On all other types of sites I tried this worked normally:

- WordPress.com subdomain simple site (private or public)
- WordPress.com public custom domain simple site

As a side-note, it's also odd that on iOS when I log in with the application-specific password I am logged into my WordPress.com account, but on Android, although I see all my sites listed and can manage them in the app if I click on 'Me' I'm not logged into my WordPress.com account which means there is no way for me to log out of these sites or WordPress.com without first logging into WordPress.com and logging out again.

renanferrari:
Fixes #12832

When logging in using a site address, we used to make a call to [`/sites/$site`](https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/) to fetch more info about the site and to determine if we should treat it as a WordPress.com site or a self-hosted one.

That endpoint did not work well with private sites, which would cause issues for users that had custom domains and 2FA enabled (you can read more about it in the linked issue or in this internal ref: pbArwn-16v-p2).

This PR replaces the usage of that endpoint with [`/connect/site-info`](https://public-api.wordpress.com/rest/v1.1/connect/site-info/help), which allows us to better handle those situations. This endpoint was already being implemented in the LoginFlow library, but was only used by the Woo app. I tried to reuse some of that code, while also making sure to keep the existing behavior regarding how to treat different site types (WordPress.com, Jetpack and self-hosted).

## To test

You can test the following scenarios by creating sites with [jurassic.ninja](https://jurassic.ninja) or by using a [pre-configured test site](https://do.wpmt.co/).

### Main Scenario

**Private WordPress.com site with custom domain and 2FA enabled**

0. Clear app data.
1. Tap **Enter your site address**.
1. On the Site Address screen, enter a site address for a private WordPress.com site with a custom domain and 2FA enabled.
1. Tap **Continue**.
1. Notice the Username/Password screen.
1. Enter the username and password associated with the site.
1. Tap **Continue**.
1. Notice that the 2FA screen was displayed with no errors.
1. Enter the two-factor authentication code.
1. Tap **Continue**.
1. Complete the login flow normally.

### Secondary Scenarios

**WordPress.com, Self-Hosted or Jetpack site**

0. Clear app data.
1. Tap **Enter your site address**.
1. On the Site Address screen, enter either a WordPress.com, a Self-Hosted or a Jetpack site address.
1. Tap **Continue**.
1. Notice the Username/Password screen.
1. Enter the username and password associated with the site.
1. Tap **Continue**.
1. Complete the login flow normally.

Note: While working on this PR, it was noted that Atomic and Jetpack sites are being treated as Self-Hosted sites as opposed to WP.com sites, which is the expected behavior. This will be addressed in a follow-up PR.

**HTTP Auth**

0. Clear app data.
1. Tap **Enter your site address**.
1. On the Site Address screen, enter http://do.wpmt.co/httpauth/.
1. Tap **Continue**.
1. Notice the Authorization Required dialog asking for the username/password pair for HTTP Basic Auth.

**Self-Signed SSL**

0. Clear app data.
1. Tap **Enter your site address**.
1. On the Site Address screen, enter https://do2.wpmt.co/vanilla/.
1. Tap **Continue**.
1. Notice the Invalid SSL Certificate dialog asking for a confirmation of the self-signed SSL certificate.

---

PR submission checklist:


khaykov (approved):
Test scenarios work as expected! I'm not super familiar with the login codebase, but nothing particularly stands out to me 👍

PR #12948, opened 2020-09-14
Fix build and tests on JDK11

### LINKED ISSUE #12739: Add support for JDK 11
The latest version of Circle CI requires our App support JDK 11. See paqN3M-c9-p2

malinajirka:
Fixes #12739

CircleCI Android docker images were updated to JDK 11 on August 17. Since the project was failing we decided to pinpoint a docker image which was still using JDK8 as a temporary workaround. This PR fixes the root cause of the failures on JDK11 and reverts this temporary workaround.

Java started using CLDR for localization in JDK9+. This, for example, changes the format of dates in some locales. Some tests in StatsDateFormatterTest were failing - I updated the tests. This change will have an unpleasant impact that the tests will fail for devs using JDK8. 

The build was also failing with various java.lang.NoClassDefFoundError errors. Updating Robolectric to version 4.3.1 fixed the issue but introduced another issue - tests in TenorProviderTest started failing. Updating Robolectric to version 4.4 fixed the tests, but introduced yet another issue. Transitive dependency bcprov-jdk15on  version 1.65 was failing with the following exception `Failed to transform 'bcprov-jdk15on-1.65.jar' using Jetifier. Reason: null`. I decided to force version 1.64 on test builds. 
All tests and the build was working, however, lint was failing with `bcprov-jdk15on-1.64.jar: Error: Invalid package reference in library; not included in Android: javax.naming.directory`. This is a known issue of the library. I decided to suppress this warning by modifying lint.xml. 

To test
- CircleCI build
- Smoke test the app
- Run Connected tests - either locally or on CI

PR submission checklist:


loremattei (approved):
This looks good to me. 
I smoke tested the app and verified that the Connected Tests work fine on CI. 

Thanks for your work on this @malinajirka!

PR #12950, opened 2020-09-14
Fix comment render crash

### LINKED ISSUE #12775: Crashing when opening a comment from notification during block testing
While testing opening comments from notifications for blocks, it seems that either one of the comments is causing the app to crash, or just having to scroll through so many comments is causing the app to crash. The post with all the test block comments can be found here: pceC9O-60-p2

### Expected behavior
The comment is scrolled to and displayed successfully.

### Actual behavior
The app crashes. 

1. Click on a comment notification for the above p2 post
2. App will load the reader and you'll see scrolling, but then crash.

- Pixel 4 (10)
- S20 5g (10)
- Pixel 3 emulator

zwarm:
Fixes #12775 

This PR addresses a crash that occurred when trying to parse comment html. In the case of pceC9O-60-p2, the `` tag was missing the `src` parameter. The method `android.text.fromHtml` will throw an exception if required parameters are null/missing. Instead of assuming we can fill in the empty src`, I added an additional try/catch wrap to handle the `IllegalArgumentException` and set the html to empty. The calling classes will pass along the message that they want to show in place of the comment that we can't parse. The default message is "Unable to show comment"

Note: I have not been able to recreate this crash, so it's my hope that this is a one-off situation ; however the fix will stop future IllegalArgumentException fromHtml crashes.

Before | After |
---|---|
NPE - Crash |  

 
To test:
1. Log in to app
2. Navigate to the post listed above
3. Tap on the comments
4. Scroll down to the bottom and notice no crash.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/comments/CommentDetailFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/comments/CommentDetailFragment.java
@@ -684,7 +684,8 @@ public class CommentDetailFragment extends Fragment implements NotificationFragm
                 WordPress.getContext()));
 
         int maxImageSz = getResources().getDimensionPixelSize(R.dimen.reader_comment_max_image_size);
-        CommentUtils.displayHtmlComment(mTxtContent, mComment.getContent(), maxImageSz);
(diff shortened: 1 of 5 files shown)


malinajirka (approved):
LGTM, thanks!

PR #12956, opened 2020-09-15
Update chart library

planarvoid:
This PR updates the MPAndroidChart library from the alpha version to the stable 3.1.0. The only change necessary was the the use of `ValueFormatter` abstract class instead of the interfaces.

To test:
- Go to stats - granular view
- Check that the views graph looks as expected
- Check on a site with high numbers that the values on the left axis get properly formatted (10,000 -> 10k)

PR submission checklist:

### DIFF
--- a/WordPress/build.gradle
+++ b/WordPress/build.gradle
@@ -345,7 +345,7 @@ dependencies {
 
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:$coroutinesVersion"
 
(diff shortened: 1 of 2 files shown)


develric (approved):
Hey @planarvoid 👋 ! I checked this PR on Pixel 3 Emu API 29, comparing on 2 emulators the behaviour before and after the library update and did not find behaviour deviations; also big values were formatted correctly 👍 ! LGTM :shipit:

PR #12958, opened 2020-09-15
Issue/12951 report post action

### LINKED ISSUE #12951: Reader: Add a 'Report This Site' button to the reader post overflow menu
Screen Shot 2020-09-11 at 4 56 23 PM

Mirroring the action on Calypso, we should also add this option to the mobile apps. 

### Specs
- Add a report button to the reader post's overflow menu
- It should appear wherever the 'Block This Site' option appears
- Upon tapping on the action, it should open a webview at the following URL: `https://wordpress.com/abuse/?report_url={SITE_URL}`
- Upon tapping on the action, we should also track `reader_post_reported` with the following properties:
`<blog_id: {BLOG_ID}, is_jetpack: {IS_JETPACK}, post_id: {POST_ID}>`
- Report should appear under the Block action mirroring Calypso

zwarm:
Fixes #12951 

This PR adds a "Report this Post" option to the Overflow menu on reader post cards.

Note: With this implementation the user has to return from the webview themselves using the back button. In addition there will be no message in app that the report request was successful. A success or failure message will be shown in the webview. A future implementation may introduce the _report this post_  functionality directly in to the app using an API call instead of a webview.

To test:

1. Open the App
2. Navigate to reader "Following"
3. Tap on the overflow menu
4. Tap on the Report this Post option
5. An internal browser will open showing the abuse form for the post selected
6. Repeat 2-5, but switch "Discover" for "Following" tab

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
@@ -12,6 +12,7 @@ import org.wordpress.android.R
 import org.wordpress.android.analytics.AnalyticsTracker.Stat.APP_REVIEWS_EVENT_INCREMENTED_BY_OPENING_READER_POST
 import org.wordpress.android.analytics.AnalyticsTracker.Stat.FOLLOWED_BLOG_NOTIFICATIONS_READER_ENABLED
 import org.wordpress.android.analytics.AnalyticsTracker.Stat.READER_ARTICLE_VISITED
+import org.wordpress.android.analytics.AnalyticsTracker.Stat.READER_POST_REPORTED
 import org.wordpress.android.analytics.AnalyticsTracker.Stat.READER_SAVED_LIST_SHOWN
 import org.wordpress.android.analytics.AnalyticsTracker.Stat.READER_SAVED_POST_OPENED_FROM_OTHER_POST_LIST
 import org.wordpress.android.analytics.AnalyticsTracker.Stat.SHARED_ITEM_READER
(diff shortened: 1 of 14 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
```diff
@@ -141,6 +145,17 @@ class ReaderPostCardActionsHandler @Inject constructor(
         }
     }
 
+    suspend fun handleReportPostClicked(post: ReaderPost) {
```
malinajirka, line 148: We might want to consider re-using this code even in the legacy fragment. We re-use most (maybe even all) the other actions. Wdyt? If you decide to make this change, feel free to make it in develop for 15.8.

WordPress/src/main/res/values/strings.xml
```diff
@@ -1700,6 +1700,7 @@
     <!-- menu text -->
     <string name="reader_menu_tags">Edit tags and sites</string>
     <string name="reader_menu_block_blog">Block this site</string>
+    <string name="reader_menu_report_post">Report this post</string>
```
malinajirka, line 1703: We need to let Lorenzo know that we changed/added a new string - I believe we don't run the localization task on betafix builds.

malinajirka (approved):
Thanks @zwarm! Looks good to me. I noticed that "Report this post" and "Block this site" have the same icon - I assume that's intentional, right? I've left one in-code comment, but that's not urgent. Feel free to merge this PR;).

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
zwarm: @malinajirka - already have a note to update, because legacy functionality is different in 15.8. I'll drop a task on the RI board too! Thanks

WordPress/src/main/res/values/strings.xml
zwarm: ooh, I did not know this. Thank you!!

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
malinajirka: ohh, I see - I forgot about that. Thanks;)

PR #12959, opened 2020-09-15
Issue/remove code related to status bar color update

khaykov:
This PR removes the manual setting of the status bar color. Since AppBar refresh feature got merged, we use light status bar by default, and there is no need to switch between white and blue colors anymore.

In addition, I was also safe to remove it from MediaGridFragment, since action mode related code was moved into the new media picker.

@malinajirka @ashiagr most prominent location for status bar color update was in Reader, so I would really appreciate if you could make sure everything is ok there :)

To test:
- Toggle between My Site, Reader, and Notification tabs and make sure the status bar color does not change.
- Navigate to Site Preview and confirm that the status bar color is white in light mode.

PR submission checklist:


ashiagr (approved):
Thanks @khaykov ! Changes look good to me 👍.

PR #12960, opened 2020-09-15
Follow recommended blog from "recommended_blogs" discovery card

AdamGrzybkowski:
Task: #12834

This PR introduces `follow` action from `recommended_blogs` card in the new Discover tab. 
I've tried to reuse as much `follow` logic as possible. I had to slightly modify the use-case to make it more versatile and not tightly coupled with the `Post` object.  

The first implementation was storing all the information about the recommended blogs in the `tbl_discover_cards` table. After discussing it with @ashiagr we came to the conclusion that it makes more sense to reuse `tbl_recommended_blogs` and store the `ReaderBlog` object there, and only store simplified JSON in the `tbl_recommended_blogs`. This is how it was done for the `Post` card, so following the same pattern makes it easier to follow. This commit https://github.com/wordpress-mobile/WordPress-Android/pull/12960/commits/731b18803623ae0c613a111e7998f744c79a98ca contains all the changes that were needed to do that.

Questions:
1. Should we show "enable notification" snack when following the Site from the "recommended_blogs" card?

| Light  | Dark  |
|---|---|
| follow_site_light  |  follow_site_dark |
| Screenshot_20200924-201815 | Screenshot_20200924-201744 |
| Screenshot_20200924-201834 | Screenshot_20200924-201856 |

To test:
- open Discovery tab
- scroll to 'Sites to follow' card
- click follow/unfollow IconButton
- verify site was followed/unfollowed

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/models/discover/RecommendedBlog.kt
```diff
@@ -0,0 +1,11 @@
+package org.wordpress.android.models.discover
+
+data class RecommendedBlog(
```
AdamGrzybkowski: previous conversation https://github.com/AdamGrzybkowski/WordPress-Android/pull/1#discussion_r487567023

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -1610,35 +1609,6 @@ public void onEventMainThread(ReaderEvents.SearchPostsEnded event) {
         }
     }
 
-    /*
-     * called when user taps follow item in popup menu for a post
-     */
-    private void toggleFollowStatusForPost(final ReaderPost post) {
```
AdamGrzybkowski, line 1596: I would have to update this ReaderBlogActions.followBlogForPost but this whole method was unused so I simply removed it

WordPress/src/main/java/org/wordpress/android/ui/reader/services/discover/ReaderDiscoverLogic.kt
```diff
@@ -131,8 +130,8 @@ class ReaderDiscoverLogic(
         val fullCardsJson = json.optJSONArray(JSON_CARDS)
 
         // Parse the json into cards model objects
-        val cards = parseCards(fullCardsJson)
-        insertPostsIntoDb(cards.filterIsInstance<ReaderPostCard>().map { it.post })
+        val postCards = parsePostCards(fullCardsJson)
+        insertPostsIntoDb(postCards.map { it.post })
```
AdamGrzybkowski: I've decided to rename the function and modify it to return only ReaderPostCard objects.
It did confuse me before and other types are not used at all here so the code was unnecessary.

WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderSiteFollowUseCase.kt
```diff
@@ -83,4 +97,24 @@ class ReaderSiteFollowUseCase @Inject constructor(
...
+            override val blogName: String = post.blogName
+        }
+
+        data class RecommendedSite(
+            override val blogId: Long,
+            override val feedId: Long,
+            override val blogName: String,
+            val isAskingToFollow: Boolean
+        ) : Param()
+    }
```
AdamGrzybkowski: In order to reuse this use case, I decided to create a sealed class for a param object that is passed to the toggleFollow function.
This allows us to pass different properties and based on the type map it to the correct data.

WordPress/src/main/res/layout/reader_recommended_blog_item.xml
```diff
@@ -89,4 +76,29 @@
...
+        app:layout_constraintTop_toBottomOf="@+id/site_icon"
+        tools:text="@tools:sample/lorem/random" />
+
+    <org.wordpress.android.ui.reader.views.ReaderFollowButton
+        android:id="@+id/site_follow_icon"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:contentDescription="@string/reader_btn_follow"
+        android:padding="@dimen/margin_extra_large"
+        app:iconTint="@color/secondary_gray_20_selector"
```
AdamGrzybkowski: @ashiagr I was able to use `ReaderFollowButton` 
the only thing I had to modify was an `iconTint`. For now, I just used it here directly but we could:
- create a different style
add
- add a custom attribute to set the color 
Do you have a preference?

WordPress/src/main/java/org/wordpress/android/ui/reader/views/ReaderFollowButton.kt
```diff
@@ -45,6 +45,8 @@ class ReaderFollowButton @JvmOverloads constructor(
     private fun hideCaptionAndEnlargeIcon(context: Context) {
         text = null
         iconSize = context.resources.getDimensionPixelSize(R.dimen.reader_follow_icon_no_caption)
+        iconPadding = context.resources.getDimensionPixelSize(R.dimen.reader_follow_icon_padding_no_caption)
+        iconGravity = ICON_GRAVITY_TEXT_START
```
AdamGrzybkowski, line 49: Without this, the icon is not centered when text is disabled.

WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderSiteFollowUseCase.kt
```diff
@@ -28,44 +30,56 @@ class ReaderSiteFollowUseCase @Inject constructor(
...
-
-            performAction(post, isAskingToFollow)
+            val isAskingToFollow = when (param) {
+                is FromPost -> !readerPostTableWrapper.isPostFollowed(param.post)
+                is RecommendedSite -> param.isAskingToFollow
+            }
+            val showEnableNotification = when (param) {
+                is FromPost -> !readerUtilsWrapper.isExternalFeed(param.blogId, param.feedId) &&
+                        !param.post.isFollowedByCurrentUser
+                is RecommendedSite -> param.isAskingToFollow // todo: figure out logic for recommended site
```
AdamGrzybkowski: What should be the logic for the recommended site here?

AdamGrzybkowski: According to [Chris answer](https://appsp2.wordpress.com/2020/09/04/trial-task-sites-to-follow-discovery-card-thoughts-ideas/#comment-2386) this logic is fine for recommended blogs because API won't return external feeds. 

I think we could simplify it to sth like 
```kotlin
val showEnableNotification = !readerUtilsWrapper.isExternalFeed(param.blogId, param.feedId) && isAskingToFollow
```

WDYT?

WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderSiteFollowUseCase.kt
```diff
@@ -28,44 +30,56 @@ class ReaderSiteFollowUseCase @Inject constructor(
...
-            val isAskingToFollow = !readerPostTableWrapper.isPostFollowed(post)
-            val showEnableNotification = !readerUtilsWrapper.isExternalFeed(post.blogId, post.feedId) &&
-                    !post.isFollowedByCurrentUser
-            emit(PostFollowStatusChanged(post.blogId, isAskingToFollow, showEnableNotification))
-
-            performAction(post, isAskingToFollow)
+            val isAskingToFollow = when (param) {
+                is FromPost -> !readerPostTableWrapper.isPostFollowed(param.post)
+                is RecommendedSite -> param.isAskingToFollow
+            }
```
AdamGrzybkowski: I wonder if we could simplify it to:

```suggestion
            val isAskingToFollow = !readerBlogTableWrapper.isSiteFollowed(param.blogId, param.siteId)
```

WordPress/src/main/java/org/wordpress/android/datasets/ReaderBlogTableWrapper.kt
```diff
@@ -11,4 +11,13 @@ class ReaderBlogTableWrapper
...
             ReaderBlogTable.setNotificationsEnabledByBlogId(blogId, isEnabled)
+    fun isSiteFollowed(blogId: Long, feedId: Long?): Boolean {
+        return if (blogId != 0L) {
+            ReaderBlogTable.isFollowedBlog(blogId)
+        } else if (feedId != null && feedId != 0L) {
+            ReaderBlogTable.isFollowedFeed(feedId)
+        } else {
+            false
+        }
+    }
```
AdamGrzybkowski: Now that I know `blogId` can be indeed empty for an external feed and the fact that `feedId` is assigned to `blogId` in such case I think this should be 
```suggestion
    fun isSiteFollowed(blogId: Long, feedId: Long?): Boolean {
        return if (feedId != null && feedId != 0L) {
            ReaderBlogTable.isFollowedFeed(feedId)
        } else if (blogId != 0L) {
            ReaderBlogTable.isFollowedBlog(blogId)
        } else {
            false
        }
    }
```

WDYT?

WordPress/src/main/java/org/wordpress/android/models/discover/RecommendedBlog.kt
ashiagr: Thank you for linking original discussion :). 

Some more thoughts:

> We would still need to store some information about the recommended_blogs card though.

We can continue to save simplified json into `tbl-discover-cards` similar to the format used for posts. Something like this: 
`{"type":"recommended_blogs","data":[{"blog_id":11184476,"feed_id":24204169},…}`

AFAIU` tbl-discover-cards` is used only to save simplified json with info enough to identify a record in the DB. If you notice actual fields for posts are saved into a table. I think this was done with the same  “single source of truth” principle in mind.

You’re right about using a single source of truth (`tbl-blog-info`) for the follow state. I’m bit worried that by storing blog fields like `name`, `description`, `icon` fields as json into `tbl-discover-cards`, we would violate the same principle. Lmkwyt.

As suggested, we can ask to include a blog model in the cards API and use the same flow as for posts. This can bring consistency in the code and it might be easier to understand in the future. 

> and we are fine with storing a lot of blogs over time

I think we’ll be storing lot of recommended blogs json data as well over time from the discover cards API into `tbl-discover-cards`. Wdyt?

WordPress/src/main/res/layout/reader_recommended_blog_item.xml
ashiagr: We have a style for "icon + caption" follow new button [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/res/values/reader_styles.xml#L241). Thinking in terms of consistency, we can use a different style for "icon only" button.

WordPress/src/main/java/org/wordpress/android/datasets/ReaderBlogTableWrapper.kt
ashiagr: Looks good to me.

WordPress/src/main/java/org/wordpress/android/models/discover/RecommendedBlog.kt
AdamGrzybkowski: > I think we’ll be storing lot of recommended blogs json data as well over time from the discover cards API into tbl-discover-cards. Wdyt?

hmm tbh I assumed it's cleaned when refresh is done, but haven't confirmed it in the code, so you might be right! :) I will check it later though 

Overall I agree with what you wrote :) 
I'm still not sure If I will be able to reuse current models. I mean If we get full blog object I will have to use `ReaderBlog` instead of `ReaderRecommendedBlog`, and it feels like the second one should be used to parse recommended blogs 🤔  so this could bring some confusion as well. 
One thing that comes to my mind that could be a good example this solution might not be future proof is if the design changed and we would have to show e.g. `score` we couldn't do that by using `ReaderBlog`. But maybe that's just going a bit too far?

ashiagr: > I’m still not sure If I will be able to reuse current models. I mean If we get full blog object I will have to use `ReaderBlog` instead of `ReaderRecommendedBlog`

Using `ReaderBlog` is completely fine -> we can completely remove `ReaderRecommendedBlog` if we won’t use it - which is fine as well.  I think it is currently used on "Reader -> Following tab -> Filter (Gear icon) -> Manage Sites and Tags -> Sites you may like" but might be removed after new changes. Let's not worry about it now.

> One thing that comes to my mind that could be a good example this solution might not be future proof is if the design changed and we would have to show e.g. `score` we couldn’t do that by using `ReaderBlog`. But maybe that's just going a bit too far?

I agree that it can happen, but it’s not something we expect - we’ll have to deal with it. Again, let's not worry about it now :).

AdamGrzybkowski: ok! Thanks for all the answers and discussion @ashiagr  :) 
I will ask for a full blog object

WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderSiteFollowUseCase.kt
AdamGrzybkowski: I simplified this, so no need for sealed class anymore

WordPress/src/main/res/layout/reader_recommended_blog_item.xml
AdamGrzybkowski: Done :)

WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderSiteFollowUseCase.kt
ashiagr: LGTM 👍

mbshakti:
This is so exciting to see, works smoothly & looks awesome, thanks so much for working on it! 

Some questions:
- I noticed we are only showing 2 sites instead of 3. Is that due to the API?
- What color are we using for the site URL? I know I had "on background" in the specs, but would "on surface" with a high alpha be lighter? Ideally I'd like to use the same color as the site URL on the blog/site page.

One request:
- Can we increase the padding beneath the "sites to follow" heading so it matches the "you might like" (topics) heading. In the specs, I had 12px, but I think it should be 16px. Sorry about that!

Thanks again!

AdamGrzybkowski:
@mbshakti thanks for the review :) 

1. Yes, the API returns 2 blogs. I haven't yet seen a card with a different number. 
2. currently the URL styling is: `font style: body2; color: OnBackground`, URL on blog/site page has a bit different style: `font style: caption; color: OnSurface`. Should I match it? 

> Can we increase the padding beneath the "sites to follow" heading so it matches the "you might like" (topics) heading. In the specs, I had 12px, but I think it should be 16px. Sorry about that!

Yup, I will do that :)

mbshakti:
Gotcha. Yes, let's match the URL on the blog/site page. If possible, would love to see one last screenshot with the URL style change & added padding.

Thank you @AdamGrzybkowski!

AdamGrzybkowski:
> If possible, would love to see one last screenshot with the URL style change & added padding.

Sure, I will ping you when it's ready :)

AdamGrzybkowski:
@mbshakti The PR is updated with fixed styling :) Let me know if it's ok now!

| Before  | After  |
|---|---|
| Screenshot_20200917-211607 | Screenshot_20200919-173309 |

mbshakti:
Looks awesome!! Thank you :)

WordPress/src/main/java/org/wordpress/android/datasets/ReaderBlogTableWrapper.kt
```diff
@@ -11,4 +11,18 @@ class ReaderBlogTableWrapper
...
+            ReaderBlogTable.getBlogInfo(blogId)
+        }
+    }
+    fun isSiteFollowed(blogId: Long, feedId: Long): Boolean {
+        return if (feedId != 0L) {
+            ReaderBlogTable.isFollowedFeed(feedId)
+        } else {
+            ReaderBlogTable.isFollowedBlog(blogId)
+        }
+    }
```
AdamGrzybkowski, line 27: This `if else` statement is duplicated now, I wonder if we could handle this in the DB level by using an SQL `SELECT * FROM tbl_blog_info WHERE blog_id=? AND feed_id=?` expression. The pair of `blog_id` and `feed_id` should be unique, right?
Any thoughts on this?

WordPress/src/main/java/org/wordpress/android/models/ReaderBlog.java
```diff
@@ -25,6 +26,28 @@
     private String mImageUrl;
     private String mFeedUrl;
 
+    public static ReaderBlog fromRecommendedBlogJson(@NotNull JSONObject json) {
```
AdamGrzybkowski: Moved this to a separate method to not increase the complexity of `ifs` statements more.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCase.kt
```diff
@@ -24,51 +25,54 @@ class GetDiscoverCardsUseCase @Inject constructor(
...
+    private val readerBlogTableWrapper: ReaderBlogTableWrapper,
     private val appLogWrapper: AppLogWrapper,
     private val appPrefsWrapper: AppPrefsWrapper,
     @Named(IO_THREAD) private val ioDispatcher: CoroutineDispatcher
 ) {
-    suspend fun get(): ReaderDiscoverCards =
-            withContext(ioDispatcher) {
-                val cardJsonList = readerDiscoverCardsTableWrapper.loadDiscoverCardsJsons()
-                val cards: ArrayList<ReaderDiscoverCard> = arrayListOf()
+    suspend fun get(): ReaderDiscoverCards = withContext(ioDispatcher) {
```
AdamGrzybkowski, line 33: I moved `withContext(ioDispatcher)` up, as the indentation in this file was starting to make it hard to fit anything in one line.

WordPress/src/main/java/org/wordpress/android/ui/reader/repository/usecases/GetDiscoverCardsUseCase.kt
```diff
@@ -24,51 +25,54 @@ class GetDiscoverCardsUseCase @Inject constructor(
...
+                            continue@forLoop
+                        }
+                    }
+                    ReaderConstants.JSON_CARD_RECOMMENDED_BLOGS -> {
+                        cardJson?.let {
+                            val recommendedBlogs = parseDiscoverCardsJsonUseCase.parseSimplifiedRecommendedBlogsCard(it)
+                                    .mapNotNull { (blogId, feedId) ->
+                                        readerBlogTableWrapper.getReaderBlog(blogId, feedId)
+                                    }
+                            cards.add(ReaderRecommendedBlogsCard(recommendedBlogs))
```
AdamGrzybkowski, line 70: This is the actual change in this file

WordPress/src/main/java/org/wordpress/android/ui/reader/services/discover/ReaderDiscoverLogic.kt
```diff
@@ -176,6 +183,12 @@ class ReaderDiscoverLogic(
         ReaderPostTable.addOrUpdatePosts(ReaderTag.createDiscoverPostCardsTag(), postList)
     }
 
+    private fun insertBlogsIntoDb(blogs: List<ReaderBlog>) {
+        blogs.forEach { blog ->
+            ReaderBlogTable.addOrUpdateBlog(blog)
+        }
```
AdamGrzybkowski, line 189: I've decided to reuse `ReaderBlogTable .addOrUpdateBlog()` function because we don't store a lot of blogs at this point, but I can add a new method that will store all the blogs in one transaction. WDYT?

malinajirka: I think the current approach is fine ;)

WordPress/src/main/java/org/wordpress/android/models/ReaderBlog.java
malinajirka: It seems that the only difference between `fromRecommendedBlogJson` and `fromJson` is that "is_jetpack" is called "jetpack", right? I'm a bit worried that introducing a new method and copy-pasting the code might increase the complexity even more. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostUiStateBuilder.kt
```diff
@@ -175,18 +174,21 @@ class ReaderPostUiStateBuilder @Inject constructor(
...
-        onItemClicked: (Long, Long?) -> Unit
+        recommendedBlogs: List<ReaderBlog>,
+        onItemClicked: (Long, Long) -> Unit,
+        onFollowClicked: (ReaderRecommendedBlogUiState) -> Unit
     ): ReaderRecommendedBlogsCardUiState = withContext(bgDispatcher) {
         recommendedBlogs.take(READER_RECOMMENDED_BLOGS_LIST_SIZE_LIMIT)
                 .map {
                     ReaderRecommendedBlogUiState(
-                            name = StringEscapeUtils.unescapeHtml4(it.name),
+                            name = it.name,
```
malinajirka, line 184: Just so I get a full context - how come we can remove this "unescapeHtml"?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderCardUiState.kt
```diff
@@ -84,12 +86,20 @@ sealed class ReaderCardUiState {
...
-            val feedId: Long?,
+            val feedId: Long,
             val description: String,
             val iconUrl: String?,
-            val onItemClicked: (Long, Long?) -> Unit
+            val isFollowed: Boolean,
+            val onItemClicked: (Long, Long) -> Unit,
+            val onFollowClicked: (ReaderRecommendedBlogUiState) -> Unit
         ) {
             val isDescriptionVisible: Boolean = description.isNotEmpty()
```
malinajirka: We usually make the description nullable and when it's null we hide the view -> there is a helper method in UiHelpers.setTextOrHide. We might want to consider using it. Wdyt?

malinajirka (commented):
Great job @AdamGrzybkowski! I haven't done a full review, but I checked most of the code changes. It looks good overall. I'll leave the final review to Ashita as we discussed. However, I've encountered an issue while I was testing the changes.

When I click on the new "follow" button, the request keeps failing for some blogs - eg. "9to5google.com". When I open a detail of that blog, I'm able to follow it. So it seems it's not a server/blog issue. Any guesses what might be the root cause of the issue - I thought we are sharing the same code for the follow action, right?



Logs: ```feed follow failed - {"info":"no_access","subscribed":false} - read/following/mine/new?source=android&url=http%3A%2F%2F9to5google.com%2Ffeed```

New Card: Failing request: `https://public-api.wordpress.com/rest/v1.1/read/following/mine/new?source=android&url=http%3A%2F%2F9to5google.com%2Ffeed&locale=en_US`

Blog detail: Working request: `https://public-api.wordpress.com/rest/v1.1/sites/146091833/follows/new?source=android&locale=en_US`

Random guess: If I recall correctly Annmarie and Ashita had some troubles recognising rss feeds. They introduced `ReaderUtils.isExternalFeed` method. It might not be related at all but I thought I'd better mention it in case it'd be helpful.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostUiStateBuilder.kt
AdamGrzybkowski: Previously I used Gson without any kind of custom adapter, so I've decided to unescape the String purely in the presentation layer. Now that I reused the existing model `ReaderBlog` and I'm parsing the JSON manually I've resued the existing utility class `JSONUtils.getStringDecoded` in the data layer (this is how it's done in other places)

WordPress/src/main/java/org/wordpress/android/models/ReaderBlog.java
AdamGrzybkowski: These are also different

`feed_id` / `feed_ID`
`url` / `URL`
`blog_id` / `blog_ID or ID`
`feed_url` / `feed_URL`

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderCardUiState.kt
AdamGrzybkowski: hmm yeah, I think I would prefer to have description nullable instead of an empty string and handle this when building the UI model. I did that as I kind of stuck to the way API works (it sends empty description instead of null)

AdamGrzybkowski:
@malinajirka 
Thanks for the bug report! I guess I must've incorrectly handled the `blogId`/`feedId` logic? At least that's the first thing that comes to my mind, I will test it later. 

> Random guess: If I recall correctly Annmarie and Ashita had some troubles recognising rss feeds. They introduced ReaderUtils.isExternalFeed method. It might not be related at all but I thought I'd better mention it in case it'd be helpful.

from what I know, external feeds are not included in the `recommended_blogs` card

malinajirka:
> from what I know, external feeds are not included in the recommended_blogs card

Yep, I agree with that statement. However, it seems that the failing request is using an endpoint for external feeds, right? - `https://public-api.wordpress.com/rest/v1.1/read/following/mine/new?source=android&url=http%3A%2F%2F9to5google.com%2Ffeed&locale=en_US`. Take it with a grain of salt - I'm not sure this endpoint is supposed to be used only with external feeds. Wdyt?

WordPress/src/main/java/org/wordpress/android/models/ReaderBlog.java
AdamGrzybkowski: My reasoning: 
Usually, when I work with multiple backends I'm creating separate models that correspond to each API response body. The advantage is that if one API's response changes, changing just this one model is enough. 
Here I would say is a similar problem, but instead of creating a new model I've decided to add a new method with that parsing logic. If the API response changes I don't need to modify the if logic, I will just update this one method that is responsible for this.

AdamGrzybkowski:
> Yep, I agree with that statement. However, it seems that the failing request is using an endpoint for external feeds, right? - https://public-api.wordpress.com/rest/v1.1/read/following/mine/new?source=android&url=http%3A%2F%2F9to5google.com%2Ffeed&locale=en_US. 

That's a very good hint, thanks :) I will investigate it later! 

> Take it with a grain of salt - I'm not sure this endpoint is supposed to be used only with external feeds. Wdyt?

Me neither, but I will find out :D

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderCardUiState.kt
malinajirka: We might want to consider transforming the empty string to null in the VM/ui state builder layer. Wdyt?

AdamGrzybkowski: Yeah I agree :) That's what I tried to say in the previous comment :D

malinajirka: Ohh I see :D I misread your previous comment. My bad.

AdamGrzybkowski:
@malinajirka I've been scrolling through the discovery tab but so far I couldn't find a similar situation. Could you possibly try to extract the `blogId` and `feedId` for this recommended blog from `tbl_discover_cards` table? You can do this easily in the AS 4.1-rc version by using a Database inspector. 

If I understand correctly this is a blog (not feed) and my guess is that `feed_id` was not null(nor 0) in the recommended blogs card response. Because of that [this code](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/reader/actions/ReaderBlogActions.java#L250) tries to follow a feed and not a blog. But that's just my guess...

I wonder if @bluefuton could help with that, whether my reasoning makes sense or not.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderCardUiState.kt
AdamGrzybkowski: unresolved it to not forget to fix it :)

WordPress/src/main/java/org/wordpress/android/models/ReaderBlog.java
malinajirka: I see and your reasoning makes sense. At the same time I feel like if we have an inconsistent field names across the endpoints we might end up with tens of different parsing logics which are almost identical. It might get quite difficult to get oriented in and it would probably be highly error-prone. Since this is a new endpoint I'd consider pinging someone from the API and asking them if they considered making the endpoints consistent. Perhaps, there is a reason why the attribute names differ, but if there isn't I think introducing such inconsistency might bite us in the future. Wdyt?

Either case, I wouldn't block this PR on it. We can "refactor" the code if we decide to make the changes on the API.

AdamGrzybkowski:
Ok I found a similar problem with a blog [`reclining standards`](https://public-api.wordpress.com/rest/v1.1/sites/62875176)

In the `recommended_cards` card response this site has `blogId=62875176` & `feedId=96380324` so as I suspected the code will execute follow feed as `feedId != 0L`. 

The question is now how should I know that I should use the `blogId` in that case?

@bluefuton any thought how to handle it?

malinajirka:
The tbl_discover_cards table got cleared when I re-opened the app. However, I checked tbl_blog_info and this is the entry:

```146091833 | 0 | https://9to5google.com | https://9to5google.com/wp-content/uploads/sites/4/2017/03/favicon-bg-none-face-white.png |   | 9to5Google | Google news, Pixel, Android, Home, Chrome OS, more```

However, it's possible that this entry got saved when I opened "blog detail" so it's not relevant. But unfortunately, I wasn't able to recover more.

> The question is now how should I know that I should use the blogId in that case?

Isn't this exactly what ReaderUtils.isExternalFeed takes care of? 
```
public static boolean isExternalFeed(long blogId, long feedId) {
         return (blogId == 0 && feedId != 0) || blogId == feedId;
    }
```

How come it works on the detail screen - is the code different or the response from the api differs?

**Comment by AdamGrzybkowski** [comment]
`ReaderUtils.isExternalFeed` is used only to determine if we should show `enable notification` snackbar not which endpoint to use, or at least that's what I've found

> How come it works on the detail screen - is the code different or the response from the api differs?

Hard to say now, as I can't compare the responses but that would be my guess. 

```
public static boolean isExternalFeed(long blogId, long feedId) {
         return (blogId == 0 && feedId != 0) || blogId == feedId;
    }
```

this will always return false for recommended blog according to @bluefuton [comment here](https://appsp2.wordpress.com/2020/09/04/trial-task-sites-to-follow-discovery-card-thoughts-ideas/#comment-2341)

**Comment by AdamGrzybkowski** [comment]
> When I open a detail of that blog, I'm able to follow it.

it didn't work for me 😕  I was able to successfully follow a site from the detail screen only after manually removing this blog form the `tbl_blog_info`


I will ask Chris about that in case he misses the Github notification.

File: WordPress/src/main/java/org/wordpress/android/models/ReaderBlog.java (hunk 11 above)
Comment (AdamGrzybkowski) [reply WordPress/src/main/java/org/wordpress/android/models/ReaderBlog.java]: > Since this is a new endpoint I'd consider pinging someone from the API and asking them if they considered making the endpoints consistent.

I will do that. Tbh I assumed this is the new API and they don't want to be restricted by the old one? But is this true? 

 > We can "refactor" the code if we decide to make the changes on the API.

👍

**Comment by AdamGrzybkowski** [comment]
@ashiagr @malinajirka 
I went through the code and I'm a bit confused so I will need your input on this :) This comment will be long so I will do my best to describe it as clear as possible. 
I did check 3 different places where you can follow the Site 

| 1. Post Header | 2. Site details | 3. Recommended Blogs |
| --- | --- | --- |
| Zrzut ekranu 2020-09-27 o 18 56 56 | Zrzut ekranu 2020-09-27 o 18 57 11 | Zrzut ekranu 2020-09-27 o 18 57 22 |


**1. Post Header**

This one is only using `feedId` do determine whether a Site is a blog or feed with the following code:
```kotlin
if (feedId != 0) {
    return followFeedById(feedId, isAskingToFollow, actionListener);
} else {
    return followBlogById(blogId, isAskingToFollow, actionListener);
}
```


**2. Site details**

This one is using above-mentioned by @malinajirka util function to determine the SIte type 
```kotlin
if (ReaderUtils.isExternalFeed(mBlogId, mFeedId)) { 
    result = ReaderBlogActions.followFeedById(mFeedId, isAskingToFollow, listener);
} else {
    result = ReaderBlogActions.followBlogById(mBlogId, isAskingToFollow, listener);
}

 /**
    * isExternalFeed identifies an external RSS feed
     * blogId will be empty for feeds and in some instances, it is explicitly
     * setting blogId equal to the feedId  
*/
public static boolean isExternalFeed(long blogId, long feedId) {
    return (blogId == 0 && feedId != 0) || blogId == feedId;
}
```

**3.  Recommended Blogs**

Same as point 1 as I did reuse the same use case to follow a Site. 



Now, assuming a Site with `blogId=123` & `feedId=321`(a similar case was failing with both ids):
**1** and **3** Will call `followFeedById`
**2** Will call `followBlogById`
(I did test that, we are actually calling different endpoints for the same Site)
Both will most of the time work but I guess there is a case where following a feed will fail. 


One solution to this problem could be simply using the util method from the point nr 2 instead of checking only the `feedId` value. This would effectively mean that we would always call `followBlogById` from the `recommended_blogs` card because as Chris wrote:
> So the only possible combinations for recommended_blogs cards should be:
>
>blog_id = valid ID and feed_id = valid ID (these IDs will be different)
>blog_id = valid ID and feed_id = null 

Let me know your thoughts on this :) 


BTW. It seems like this is not an issue with a different response from the backend as I originally suspected, but your input @bluefuton might still be very valuable. If you know why following the site using the `feed` endpoint could fail only in some scenarios, even if the `feedId` was not null, let us know! :)

WordPress/src/main/java/org/wordpress/android/models/ReaderBlog.java
bluefuton: Hi folks - thanks for the feedback on this. On reflection I think it's probably best to keep the naming consistent with the v1 API. I'll mirror the key naming of `/v1.1/read/site/:site`.

bluefuton:
> However, it seems that the failing request is using an endpoint for external feeds, right?

@malinajirka admittedly this is a bit confusing!

Every site that we crawl in Reader will have a corresponding feed ID, including internal sites and Jetpack sites. You can follow any feed using the `/read/following/mine/new` endpoint.

However, it is also possible to follow a blog_id directly, even if we don't have a corresponding feed for it.

> If you know why following the site using the feed endpoint could fail only in some scenarios, even if the feedId was not null

@AdamGrzybkowski if you could provide an example request and response to the follows API endpoint that fails, that'd be a great help to diagnose this.

WordPress/src/main/java/org/wordpress/android/models/ReaderBlog.java
bluefuton: I've tackled this in D50244-code and put that up for review 👍 The response will look like this:

Screen Shot 2020-09-28 at 14 52 21

The posts returned by the endpoint use the v1-style key naming, so it does feel right to be consistent. Thanks for raising it!

Just to provide some background on why I originally considered a different path here, my main historical gripes are:
* site and blog (and site_id/blog_id) are used interchangeably - generally the API refers to `site` but in the wpcom codebase it's usually `blog` and `site_id` means something else entirely.
* the key naming of v1 is a bit weird - mostly camel case, except for ID and URL (so we have prefer_feed but then also feed_ID and feed_URL) - we have to override our PHP linter to ignore these inconsistencies.
* we tend to include unnecessary site info in similar API responses that's never used by the client (for example, `/read/site/:site` returns four flags about which editor experience is enabled, which we never use), and I was aiming for the minimum in this case to keep the response as small as we can.

WordPress/src/test/java/org/wordpress/android/ui/reader/discover/ReaderPostUiStateBuilderTest.kt
```diff
@@ -820,23 +824,31 @@ class ReaderPostUiStateBuilderTest {
...
+                { }
+        )
 
         // Assert
         assertThat(uiState.blogs.size).isEqualTo(3)
     }
 
     @Test
-    fun `blog description is not visible when description is empty`() = test {
+    fun `blog description is not null when description is empty`() = test {
```
ashiagr: Seems like we want to write blog description is null when description is empty. Can you please double check?

ashiagr (commented):
Thank you everyone for working on this persistently! Couple of more observations, we're very close to merging it 🎉. 

1. Noticed a crash
cc @AdamGrzybkowski 

```
E/AndroidRuntime: FATAL EXCEPTION: DefaultDispatcher-worker-4
    Process: org.wordpress.android.prealpha, PID: 4514
    org.json.JSONException: Value null at feed_id of type org.json.JSONObject$1 cannot be converted to long
        at org.json.JSON.typeMismatch(JSON.java:101)
        at org.json.JSONObject.getLong(JSONObject.java:528)
        at org.wordpress.android.ui.reader.services.discover.ReaderDiscoverLogic.createSimplifiedRecommendedBlogsCardJson(ReaderDiscoverLogic.kt:247)
        at org.wordpress.android.ui.reader.services.discover.ReaderDiscoverLogic.createSimplifiedJson(ReaderDiscoverLogic.kt:207)
        at org.wordpress.android.ui.reader.services.discover.ReaderDiscoverLogic.handleRequestDiscoverDataResponse(ReaderDiscoverLogic.kt:145)
        at org.wordpress.android.ui.reader.services.discover.ReaderDiscoverLogic.access$handleRequestDiscoverDataResponse(ReaderDiscoverLogic.kt:56)
        at org.wordpress.android.ui.reader.services.discover.ReaderDiscoverLogic$requestDataForDiscover$1$listener$1$1.invokeSuspend(ReaderDiscoverLogic.kt:114)
```

2. Noticed same recommendations in a block - same entries were present in the `tbl_discover_cards`:
`{"type":"recommended_blogs","data":[{"blog_id":146091833,"feed_id":86180988},{"blog_id":146091833,"feed_id":86180988}]}`

Can we ensure we get unique recommendations?
cc @bluefuton 

repeat recommendation

AdamGrzybkowski:
@bluefuton 
I thought I will be able to write more than @malinajirka [here](https://github.com/wordpress-mobile/WordPress-Android/pull/12960#pullrequestreview-496317948) but there's really not much more. 
but here's what I have: 

`blogId = 54534204`
`feedId = 69293169`

calling `read/following/mine/new?source=android&url=http%3A%2F%2Ftheprettycanary.blog`
fails with response
```
{"info":"invalid_feed","subscribed":false}
```

calling `sites/54534204/follows/new?source=android`
succedes with response
```
{"success":true,"is_following":true,"meta":{"links":{"site":"https:\/\/public-api.wordpress.com\/rest\/v1.1\/sites\/54534204","help":"https:\/\/public-api.wordpress.com\/rest\/v1.1\/sites\/54534204\/follows\/new\/help","posts":"https:\/\/public-api.wordpress.com\/rest\/v1.1\/sites\/54534204\/posts\/"}}}
```

So the `feedId` is not null but it still fails when the app calls `read/following/mine/new`. This doesn't happen for all Sites.


@ashiagr 
Interesting findings with the duplicated blog 🤔  This `9to5Google` is a real edge case :D

WordPress/src/test/java/org/wordpress/android/ui/reader/discover/ReaderPostUiStateBuilderTest.kt
AdamGrzybkowski: fixed, thanks for noticing this, I've missed it :)

WordPress/src/main/java/org/wordpress/android/models/ReaderBlog.java
bluefuton: Just a heads up that this change is going live today 🚢

AdamGrzybkowski:
@bluefuton I've just found out that unfollowing a feed with `read/following/mine/delete?url=http%3A%2F%2F9to5mac.com%2Ffeed` works, but following fails

ashiagr:
> @bluefuton I've just found out that unfollowing a feed with `read/following/mine/delete?url=http%3A%2F%2F9to5mac.com%2Ffeed` works, but following fails

@AdamGrzybkowski can you please create a separate GH issue for it? It seems like an edge case and I don't want to block this PR because of it.

WordPress/src/main/java/org/wordpress/android/models/ReaderBlog.java
AdamGrzybkowski: Thanks @bluefuton 💪 ! 
PR updated @ashiagr & @malinajirka

AdamGrzybkowski:
@ashiagr yes, sure!

AdamGrzybkowski:
@ashiagr done https://github.com/wordpress-mobile/WordPress-Android/issues/13027

WordPress/src/main/java/org/wordpress/android/models/ReaderBlog.java
```diff
@@ -26,28 +25,6 @@
...
-        blog.isPrivate = JSONUtils.getBool(json, "is_private");
-        blog.numSubscribers = json.optInt("subscribers_count");
-        JSONObject jsonNotification = JSONUtils.getJSONChild(json, "subscription/delivery_methods/notification");
-        if (jsonNotification != null) {
-            blog.isNotificationsEnabled = JSONUtils.getBool(jsonNotification, "send_posts");
-        }
-        JSONObject jsonIcon = json.optJSONObject("icon");
-        if (jsonIcon != null) {
-            blog.setImageUrl(JSONUtils.getString(jsonIcon, "img"));
-        }
```
ashiagr: @AdamGrzybkowski now that we've removed this parsing method, I think we can move this icon parsing logic to the existing block.

AdamGrzybkowski: hmm I'm not sure I follow, there is a logic already to parse the `icon` in the `fromJson()` method 🤔

ashiagr: oh yes! It exists outside else block. I noticed icon was not shown for a recommended site but it was on site details screen. I'll see if I can reproduce it.

ashiagr: icon

AdamGrzybkowski: hmm could you check the logs if there's something? 
I will try to reproduce it myself

ashiagr: BlogId, FeedId:
`{"ID":160909219,"feed_ID":97888741}`

Logs:

```
W/Glide: Load failed for https://startupandroid.com/wp-content/uploads/2019/11/cropped-19ccf390-b650-40ca-8ad8-994e509860c8.jpg with size [110x110]
    class com.bumptech.glide.load.engine.GlideException: Failed to load resource
    There was 1 cause:
    com.android.volley.NoConnectionError(java.net.UnknownHostException: Unable to resolve host "startupandroid.com": No address associated with hostname)
     call GlideException#logRootCauses(String) for more detail
      Cause (1 of 1): class com.bumptech.glide.load.engine.GlideException: Fetching data failed, class java.io.InputStream, REMOTE
    There was 1 cause:
    com.android.volley.NoConnectionError(java.net.UnknownHostException: Unable to resolve host "startupandroid.com": No address associated with hostname)
     call GlideException#logRootCauses(String) for more detail
        Cause (1 of 1): class com.bumptech.glide.load.engine.GlideException: Fetch failed
```

Seems like it was a jetpack site and `https://startupandroid.com` wasn't reachable 🤔.

AdamGrzybkowski: > Seems like it was a jetpack site

What does this mean exactly?

Why could this be loaded on the Site details screen if it wasn't reachable? 🤔

ashiagr: Site details response includes image url in a different format

https://i1.wp.com/startupandroidcom.wpcomstaging.com/wp-content/uploads/2019/11/cropped-19ccf390-b650-40ca-8ad8-994e509860c8.jpg?fit=512%2C512&ssl=1

Screenshot 2020-09-29 at 1 41 23 PM

Not sure if icon url included for recommended site should be formatted in some manner. Looks like it is a question for @bluefuton .

AdamGrzybkowski: interesting 🤔 
thanks @ashiagr for spotting this!

ashiagr: Looks like there's a base url difference
https://startupandroidcom.wpcomstaging.com/
vs
https://startupandroidcom.com/

First one is reachable, second one is not.

AdamGrzybkowski: yeah... I guess we have to wait for @bluefuton answer, as I don't think we can do much here

bluefuton: I looked into this. It's a Jetpack/Atomic site and `/sites/:site` endpoint contacts the site directly for these details (including the icon). The `/read/site/:site` relies on the information in the shadow/cache blog. I've tried a full sync but that hasn't made a difference.

The background is: this site hasn't launched yet and the domain is inactive. I don't think it should have been recommended, so I've created a task for us to filter out sites with *.wpcomstaging.com domains.

ashiagr: Yes we shouldn't recommend such sites. Thank you for this update 🙇‍♀️.

PR #12961, opened 2020-09-15
Fixing connected tests issues

khaykov:
This PR fixes connected test issues:
- Custom Site title layout update loop prevented My Site screen from becoming idle.
- `scrollTo` was not working with nested scroll views.

To test (single line Site Title has a bigger font size then multiline Site Title):
- Set a long site title, so it would wrap to the second line. Confirm that the site title size is reduced compared to a single-line title.
- Set a long site title, so it would wrap to the second line. Rotate your device into a landscape orientation, so the title will take a single line. Confirm that the title size is bigger.

PR submission checklist:


jkmassel (approved):
I see connected tests passing, and this looks like a neat addition! I'm ok to :shipit:

Thanks for handling this @khaykov!!

PR #12969, opened 2020-09-16
Add topic follow and unfollow track event

zwarm:
This PR adds tracking for `reader_reader_tag_followed` and `reader_reader_tag_unfollowed` for folloiwng/unfollowing from the topic header.

To test:
1. Tap on a topic chip to open up topic detail
2. Tap on the follow button
3. Check that a `reader_reader_tag_followed` event is triggered with a `tag` property
🔵 Tracked: reader_reader_tag_followed, Properties: {"tag":"slug"}

4. Tap on the unfollow button
5. Check that a 'reader_reader_tag_unfollowed" event is triggered with a `tag` property
🔵 Tracked: reader_reader_tag_unfollowed, Properties: {"tag":"slug"}

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderPostAdapter.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderPostAdapter.java
@@ -285,7 +285,6 @@ public class ReaderPostAdapter extends RecyclerView.Adapter<RecyclerView.ViewHol
             return;
         }
         Function0<Unit> onFollowButtonClicked = () -> {
-            // todo: annmarie remove this comment
             toggleFollowButton(tagHolder.itemView.getContext(), currentTag, tagHolder);
             return Unit.INSTANCE;
         };
@@ -331,7 +330,6 @@ public class ReaderPostAdapter extends RecyclerView.Adapter<RecyclerView.ViewHol
             renderTagHeader(currentTag, tagHolder, true);
         };
 
-        // todo: annmarie here remove this comment
         boolean success;
         boolean isLoggedIn = mAccountStore.hasAccessToken();
         if (isAskingToFollow) {
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderPostAdapter.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderPostAdapter.java
@@ -19,6 +19,7 @@ import androidx.recyclerview.widget.RecyclerView;
 import org.jetbrains.annotations.NotNull;
 import org.wordpress.android.R;
 import org.wordpress.android.WordPress;
+import org.wordpress.android.analytics.AnalyticsTracker;
 import org.wordpress.android.datasets.ReaderPostTable;
 import org.wordpress.android.datasets.ReaderTagTable;
 import org.wordpress.android.fluxc.store.AccountStore;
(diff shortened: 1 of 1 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderPostAdapter.java
```diff
@@ -283,6 +285,7 @@ private void renderTagHeader(
             return;
         }
         Function0<Unit> onFollowButtonClicked = () -> {
+            // todo: annmarie remove this comment
```
malinajirka: Is this comment intentional?

WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderPostAdapter.java
```diff
@@ -309,15 +312,26 @@ private void toggleFollowButton(
...
+                        new HashMap<String, String>() { { put("tag", slugForTracking); }});
+                } else {
+                    AnalyticsTracker.track(AnalyticsTracker.Stat.READER_TAG_UNFOLLOWED,
+                        new HashMap<String, String>() { { put("tag", slugForTracking); }});
+                }
             }
             renderTagHeader(currentTag, tagHolder, true);
         };
 
+        // todo: annmarie here remove this comment
```
malinajirka: Is this comment intentional?

WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderPostAdapter.java
```diff
@@ -309,15 +312,26 @@ private void toggleFollowButton(
...
 
+        final String slugForTracking = currentTag.getTagSlug();
+
         ReaderActions.ActionListener listener = succeeded -> {
             if (!succeeded) {
                 int errResId = isAskingToFollow ? R.string.reader_toast_err_add_tag
                         : R.string.reader_toast_err_remove_tag;
                 ToastUtils.showToast(context, errResId);
+            } else {
+                if (isAskingToFollow) {
```
malinajirka, line 322: Just to double check: Do we really want to track whether the action succeeds/fails and not the button clicks? We usually track user interaction with the app (touch events) not results of actions. However, if it's in sync with iOS, we can keep it as is. Wdyt?

malinajirka (approved):
I've tested the app and it LGTM. Feel free to merge it ;). Thanks

WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderPostAdapter.java
zwarm: ha - no - let me get rid of that

WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderPostAdapter.java
zwarm: Removed in 0daa506

WordPress/src/main/java/org/wordpress/android/ui/reader/adapters/ReaderPostAdapter.java
zwarm: >  However, if it's in sync with iOS, we can keep it as is. Wdyt?

@malinajirka - iOS confirmed it's tracked on the success of the operation. All is ready to go!

PR #12971, opened 2020-09-17
Enable like and comments action on bookmark list

malinajirka:
This PR enables comments and like actions on the saved tab in Reader. These actions were disabled in the past to make the cards on this tab as simple as possible. However, we started showing even disabled actions on post cards a few months ago (disabled actions were hidden before that). Since showing disabled actions doesn't make the card simpler we decided to enable these actions on the saved tab.

To test:
1. Open Reader
2. Bookmark a few posts from the Following and Discover tabs
3. Click on "Saved" tab
4. Notice comments and like actions are enabled - if they are disabled make sure the post has them disabled even on the original tab (following/discover)
5. Like + unlike some posts
6. Open comments of some posts

PR submission checklist:


zwarm (approved):
Works like a charm, nice work!
I tested instances of disabled comments and with comments too.
👍

PR #12978, opened 2020-09-17
Login: Properly handle Simple, Atomic and Jetpack sites addresses

### LINKED ISSUE #12957: Login: When logging in with a site address, Atomic and Jetpack sites behave like self-hosted sites instead of WP.com ones
### Expected behavior

When logging in with a site address for an Atomic or Jetpack site, I'd expect to be logged into their associated WP.com account at the end of the flow.

### Actual behavior

Instead, I'm either unable to login or they behave just like self-hosted sites, not allowing me, for example, to access Stats and Notifications or to see other sites associated with the same WP.com account.

0. Clear app data.
1. Tap **Enter your site address**.
1. On the Site Address screen, enter either a Jetpack or an Atomic site address*.
1. Tap **Continue**.
1. Notice the Username/Password screen.
1. Enter the username and password associated with the site**.
1. Tap **Continue**.
1. Complete the login flow normally.
1. On the Epilogue screen, notice how it doesn't display any other site associated with the same WordPress.com account.
1. Tap **Done**.
1. On the Main screen, tap **Notifications**.
1. Notice how it shows the "Install Jetpack" button.

\* Private Atomic sites will fail with an error here, which is also not expected.
 
\** Using WP.com credentials here won't work, which is another problem related to this. For a Jetpack site, it's possible to use site credentials (the same used to login to `wp-admin`), but 2FA won't be supported then (just like with self-hosted sites). As for Atomic sites, site credentials need to be generated via `wp-admin > Users > All Users > Select User > Generate Password`.

renanferrari:
Fixes #12957

This PR takes advantage of the endpoint used by #12940 to correctly detect Simple, Atomic and Jetpack sites on the login flow and then redirect the user to the Get Started screen, just like if they had tapped the "Continue with WordPress.com" button in the Prologue screen, where they will be able to enter an email address or continue with a Google account.

This new behavior is demonstrated in the video below:

video-giffer

Notes:
- When a user is redirected to the Get Started from the site address flow **they won't be able to signup**, meaning an error will be shown if they try to continue with a non-existing account.
- Another smaller difference in the Get Started screen is that we're changing the main text label from "Enter your email address to log in or create a WordPress.com account." to "Log in with WordPress.com to connect to yoursiteaddress.com".
- If the user provides a site address that has Jetpack installed and/or active, but not connected to a WordPress.com account, then they will proceed to the Username/Password screen, where they'll be able to login to the site using their site credentials, just like before. In this case the site will behave just like a self-hosted one, and the user won't be able to access Stats and Notifications or to see other sites associated with the same WP.com account until they setup the Jetpack connection.

## To test

0. Clear app data.
1. Tap **Enter your site address**.
1. On the Site Address screen, enter either a Simple, Atomic or Jetpack site address.
1. Tap **Continue**.
1. Notice the Get Started screen.
1. Enter an email address associated with a WordPress.com account.
1. Tap **Continue**.
1. Complete the login flow normally.

Repeat for different types of sites and verify they all work as expected.

---

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
@@ -656,9 +656,8 @@ public class LoginActivity extends LocaleAwareActivity implements ConnectionCall
 
     @Override
     public void gotWpcomSiteInfo(String siteAddress) {
-        LoginUsernamePasswordFragment loginUsernamePasswordFragment =
-                LoginUsernamePasswordFragment.newInstance(siteAddress, siteAddress, null, null, true);
-        slideInFragment(loginUsernamePasswordFragment, true, LoginUsernamePasswordFragment.TAG);
+        LoginEmailFragment loginEmailFragment = LoginEmailFragment.newInstance(false, false, true);
+        slideInFragment(loginEmailFragment, true, LoginEmailFragment.TAG);
     }
 
     @Override
--- a/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
@@ -726,8 +726,8 @@ public class LoginActivity extends LocaleAwareActivity implements ConnectionCall
     }
 
     @Override
-    public void addGoogleLoginFragment() {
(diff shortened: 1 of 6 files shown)


khaykov (approved):
Looks good, and works as expected 👍 I tried with wpcom, jetpack, self-hosted and non-wp sites.

PR #12984, opened 2020-09-18
Issue/12865 media picker action mode badge

### LINKED ISSUE #11518: Replace system image picker with custom image picker in gutenberg

### LINKED ISSUE #12865: Update Action mode in Media Picker
Based on the designs the action mode should have grey background and the "Edit" icon should show the number of items that are selected. 

The task of this PR is to implement these two changes and generally update the action mode to look according to the designs.

khaykov:
Fixes #12865

This PR adds a badge to the edit button in Media PIcker's Action Mode. The original design was for the older-looking version of the app, so I haven't modified the ActionMode toolbar color in favor of the global ActionMode style.

One minor problem I encountered is is the color of the badge ring in dark mode - I wasn't able to exactly match it to the ActionMode BG, and while I got pretty close the logic for generating elevated colors and applying them to shape drawable turned out to be quite complicated, so I opted to use surface color for the ring. (I also tried to achieve this by using elevated UI components to simulate ring around the badge, but there is a set of whole different issues with that approach).

The badge is only visible when multi-select enabled.

[Image from Gyazo](https://gyazo.com/2f423b6219c72a5e7e9d74ec4117b814)

[Image from Gyazo](https://gyazo.com/211811d9afc2e96d2aca3570fee90f9e)

[Image from Gyazo](https://gyazo.com/1147d3bcae5473e748f84e26fafe792e)

[Image from Gyazo](https://gyazo.com/28be46e23dfc998f72aa67e09096de19)

The badge can fit up to 99999 and then ellipsizes:

[Image from Gyazo](https://gyazo.com/af07bc9dd72c1076280dbd0195bd5b0c)
To test:

PR submission checklist:


WordPress/src/main/res/layout/media_picker_customize_action_layout.xml
```diff
@@ -0,0 +1,50 @@
...
+    android:background="?android:attr/actionBarItemBackground"
+    android:clickable="true"
+    android:focusable="true">
+
+    <ImageView
+        android:id="@+id/customize_icon"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:contentDescription="@null"
+        android:padding="@dimen/me_avatar_padding"
```
develric: Wonder if we should create a semantical dimen here, like `image_padding_small` (or maybe better than this 🙃 ), but feel free to ignore...

WordPress/src/main/res/layout/media_picker_customize_action_layout.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="@dimen/toolbar_height"
+    android:layout_height="@dimen/toolbar_height"
```
develric: Unless it was done on purpose to give it some extra finger hit space, I was wondering if we should use a 48dp dimension here (like `min_touch_target_sz`) otherwise the edit icon falls a bit too on left; wdyt?

| OLD | NEW |
|---|---|
| image | image |

develric (commented):
I'm reviewing this since I was starting looking into 12865 but Vojta made me note I was overlooking this PR 😊 (thanks @planarvoid for the ping and @khaykov for working on this 🙇 ).

I tested it with Pixel 3 Emu and API 28 and it looks beautiful 👍 ! I left a couple of comments inline with the code. Also at least in my emu, when I tap on the Edit icon the `override fun onActionItemClicked` function is not triggered; I wonder if we should add a click listener to the actionView or similar. Let me know wdyt, thanks 🙇 

PS: I'm sure you will see it but just noticed there is a small conflict with develop that came out 😄

WordPress/src/main/res/layout/media_picker_customize_action_layout.xml
khaykov: I realized we actually don't need it here - it's a copy-paste artifact :)

WordPress/src/main/res/layout/media_picker_customize_action_layout.xml
khaykov: Good catch 👍

khaykov:
@develric Thanks for the review! Can't believe I missed that the button is not clickable 🤦 Should be fixed now!

develric (approved):
Hey @khaykov 👋  ! Thanks for the changes 🙇 ! Works well 👍 ! 

> Can't believe...

I don't know you but it happens to me 😄  and this is why we do reviews, one plus one is more than two 😄

PR #12995, opened 2020-09-19
Add a test to create the 'My site' screenshot

designsimply:
Adds a UI Test to create a screenshot for the Play Store with heading "Keep tabs on your site" showing the "My site" screen according to the design.

android-phone-1

### Challenges

* Running screenshot tests against `develop` is currently failing with the following error. I asked for help fixing the tests earlier but have been unable to get them unstuck.

`java.lang.NullPointerException: Attempt to invoke virtual method 'boolean androidx.recyclerview.widget.RecyclerView.isComputingLayout()' on a null object reference`

### What was tried?

* Tried commenting out `PostsListPage.java:20` with `scrollToTopOfRecyclerView()` because the tests appear to be failing when attempting to scroll in either the block editor or the posts list page—the tests still failed so that must be used elsewhere.
* Tried adding a breakpoint to `PostsListPage.java:25` to try to see if that's where the failure happens and `pager.getChildAt(pager.getCurrentItem()).findViewById(R.id.recycler_view);` fails as expected but `pager.getChildAt(pager.getCurrentItem());` works and I'm not sure why it wouldn't be able to find `recycler_view` in this context.

### Notes

* For the sake of the project (and speed), we decided I should go ahead and commit and _then_ get some help figuring out how to fix the broken tests before moving forward. 
* In this first iteration, the "Create a post or page" bubble that shows up when you first install and open the app is showing on the My Site screen but it's not in the design. I can add it, but it also seems fine (and simpler) to leave it in and I'm wondering what people think about whether to remove it.

Example with the "Create a post or page" bubble|Example without it
---|---
Screenshot_20200918-181539|Screenshot_20200918-181553

To test:

1. Apply these changes to the 15.6 tag.
1. Select `WPScreensotTest` as the run configuration.
1. Run the test and confirm it passes.

PR submission checklist:

### DIFF
--- a/WordPress/src/androidTest/java/org/wordpress/android/ui/screenshots/WPScreenshotTest.java
+++ b/WordPress/src/androidTest/java/org/wordpress/android/ui/screenshots/WPScreenshotTest.java
@@ -82,7 +82,7 @@ public class WPScreenshotTest extends BaseTest {
 
         waitForElementToBeDisplayedWithoutFailure(R.id.row_blog_posts);
 
-        takeScreenshot("4-screenshot-my-site");
+        takeScreenshot("4-keep-tabs-on-your-site");
     }
 
     private void editBlogPost() {
--- a/WordPress/src/androidTest/java/org/wordpress/android/ui/screenshots/WPScreenshotTest.java
+++ b/WordPress/src/androidTest/java/org/wordpress/android/ui/screenshots/WPScreenshotTest.java
@@ -65,6 +65,7 @@ public class WPScreenshotTest extends BaseTest {
         idleFor(1000);
         takeScreenshot("1-build-and-manage-your-website");
 
+        navigateMySite();
         editBlogPost();
         manageMedia();
         navigateNotifications();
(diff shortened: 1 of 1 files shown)


WordPress/src/androidTest/java/org/wordpress/android/ui/screenshots/WPScreenshotTest.java
```diff
@@ -75,6 +76,15 @@ public void wPScreenshotTest() {
         logoutIfNecessary();
     }
 
+    private void navigateMySite() {
+        // Click on the "Sites" tab and take a screenshot
+        clickOn(R.id.nav_sites);
+
+        waitForElementToBeDisplayedWithoutFailure(R.id.row_blog_posts);
+
+        takeScreenshot("4-screenshot-my-site");
```
JavonDavis: Let's go ahead and change this to `4-keep-tabs-on-your-site`

JavonDavis (changes requested):

WordPress/src/androidTest/java/org/wordpress/android/ui/screenshots/WPScreenshotTest.java
designsimply: Aha. Yes, I meant to keep the names matching the previous pattern for this edit and then update them all at once later with better names matching what's in each screenshot instead of the heading later so that change happens all in one go.

JavonDavis:
Thanks @designsimply! 

> java.lang.NullPointerException: Attempt to invoke virtual method 'boolean androidx.recyclerview.widget.RecyclerView.isComputingLayout()' on a null object reference

I started to encounter this issue when i recently refreshed develop. It seems for some reason the layout at the time the ReceyclerView does not have an id. I've tested your changes though and it works as expected I just commented out the steps following your changes.

JavonDavis (approved):

PR #13002, opened 2020-09-22
Reader: Support Stories in detail view

aforcier:
Tweaks the Reader to support the Story block in detail view:

device-2020-09-22-122031

The bulk of the work for this has been done in the shared CSS file in Calypso and already deployed: https://github.com/Automattic/wp-calypso/pull/45397

Extra changes that were needed for WPAndroid:
* The in-app custom CSS forces auto height/width for divs, which breaks with the Story markup. Stories already have set, mobile-friendly width/height params, so I [excluded Stories from the rule](https://github.com/wordpress-mobile/WordPress-Android/compare/feature/stories-reader-detail?expand=1#diff-01ae2d5b004de5f0cdafb0f21217c146R388).
* I [excluded Stories from the img tag re-building](https://github.com/wordpress-mobile/WordPress-Android/compare/feature/stories-reader-detail?expand=1#diff-01ae2d5b004de5f0cdafb0f21217c146R128) we do, as it strips class attributes important for the CSS. Again, we can just rely on the attributes provided by the Story markup.
* Tapping on a story directly [opens the external browser](https://github.com/wordpress-mobile/WordPress-Android/compare/feature/stories-reader-detail?expand=1#diff-0ce39e7d0d99df91b63bcf95cdeb14caR1545) instead of the in-app web viewer. This is because the in-app web viewer has a lot of navigational chrome and doesn't provide as a good a 'full-screen' experience as the browser does.

iOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14965

### To test:

1. Proxy WP.com on your device (because the Story block is currently a beta block and poorly supported by the API)
2. Clear device data to make sure you get a fresh copy of the `reader-mobile.css` and aren't hit by the 5-day cache
3. Follow a site or two with Stories (DM me for some samples)
4. Check that opening a story in Reader detail shows a properly formatted story, like the screenshot above
5. Check that tapping the story opens it in the browser, and it automatically plays once the page loads

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostDetailFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostDetailFragment.kt
@@ -10,6 +10,7 @@ import android.content.pm.PackageManager
 import android.content.res.Resources
 import android.graphics.Rect
 import android.graphics.drawable.Drawable
+import android.net.Uri
 import android.os.AsyncTask
 import android.os.Bundle
 import android.text.Html
@@ -1537,6 +1538,11 @@ class ReaderPostDetailFragment : ViewPagerFragment(),
     private fun shouldOpenExternal(url: String): Boolean {
         // open YouTube videos in external app so they launch the YouTube player
         if (ReaderVideoUtils.isYouTubeVideoLink(url)) {
+            return true
+        }
+
+        // Open Stories links in external browser so they have more fullscreen play real estate
+        if (Uri.parse(url).queryParameterNames.any { it.contains("wp-story") }) {
             return true
         }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostRenderer.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostRenderer.java
@@ -379,9 +379,11 @@ public class ReaderPostRenderer {
               .append(" p, div" + (renderAsTiledGallery ? ":not(." + galleryOnlyClass + ")" : "")
                       + ", li { line-height: 1.6em; font-size: 100%; }")
               .append(" h1, h2, h3 { line-height: 1.6em; }")
-              // counteract pre-defined height/width styles, expect for the tiled-gallery divs when rendering as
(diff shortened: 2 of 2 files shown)


malinajirka (commented):
Thanks @aforcier! It all looks great overall. I've encountered one small issue though. When I click on a story on a .com site the system prompts me to choose an app I want to open it in -> when I choose WordPress Android it just re-opens Reader Post Detail view. I'm not sure we can modify the intent filter somehow. Wdyt?

aforcier:
Yikes, good catch - I'm not sure why this doesn't trigger on my emulators, but I can confirm it on an actual device.

Looks like the intent filter needs to be modified, yes - I'll look into it and follow up, thanks 👍

aforcier:
So here's what I've found out.

We already [disable the Reader deeplink when opening an external URL](https://github.com/wordpress-mobile/WordPress-Android/blob/eb2497843d3ceb49567eb7292cde7cb703e97ec8/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java#L1091), so that the app doesn't appear as an option. We then [re-enable it](https://github.com/wordpress-mobile/WordPress-Android/blob/eb2497843d3ceb49567eb7292cde7cb703e97ec8/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java#L1111) (there and in a couple of other places for safety).

As far as I can tell, on certain APIs and devices the re-enabling of the Reader deeplink component in that `finally` block happens before the `startActivity` call resolves the intent, so the Reader component which was removed is readded to the list.

This affects the Reader deeplinks in general (try hitting the globe button in the detail page instead for example) and isn't specific to Stories, but affects Stories a bit worse since they effectively link to themselves.

Reproduced on:
A Pixel 3a emulator, API 29 (@malinajirka's report above)
A Pixel 3 device, API 30

Could not reproduce on:
A Pixel 3XL emulator, API 29
A Pixel 3 emulator, API 28

I'll open an issue and resolve this in a separate PR - I think the change will be somewhat deep and I don't want to bury it in an apparently unrelated PR.

**Edit:** Filed in #13010, fixed in #13012.

malinajirka (approved):
Thanks, LGTM and I can confirm the fix works as expected.

PR #13004, opened 2020-09-22
Refactor to make variable more in line with it's meaning.

develric:
While reviewing another PR we noticed that the `onlyImagesSelected` was used as if it was a sort of `notOnlyImagesSelected`, so we decided to improve it.

### To test

Unit test should pass; moreover the behaviour should not have changed (it was correct even before), so:
- Open Media > + > Choose from Device
- In the media picker try to select only images and check the edit icon is shown
- Try to select other kind of files (and also images together with those other kind of files) and check the Edit icon is hidden
- Smoke test

PR submission checklist:


planarvoid (approved):
Looks good and works well 👍  thanks

PR #13007, opened 2020-09-23
Remove browser type from media picker

planarvoid:
This PR removes the browser type from the MediaPickerActivity. There are 2 tricky places where it was still used and I hope I fixed them correctly.

I've moved this part:
```
if (browserType == FEATURED_IMAGE_PICKER) {
                    featuredImageHelper.trackFeaturedImageEvent(
                            IMAGE_PICKED,
                            localPostId
                    )
                }
```
to the place where the results are processed in the `EditPostSettingsFragment`. I've tested it and I think it's tracking the upload correctly.

I've removed the `isWPStoriesMediaBrowserTypeResult(data))` from the stories handler. As far as I can tell it wasn't necessary. I've checked all the usage and I don't think there is any branch we're skipping or missing if the check failed (I don't think it could've failed, it was just a precaution). Let me know if it makes any sense. 

I've squeezed in `LiveDataUtils` update into this PR because it will be necessary for the stock media data source. I hope that's fine. 

To test:
- Set the ConsolidatedMediaPickerConfig to `true` in the App settings
- Check the featured image upload
- Check that an event like this `WordPress-STATS: 🔵 Tracked: featured_image_picked_post_settings, Properties: {"post_id":2}` is tracked correctly when the image is picked (before the upload finishes)
- Check that you can pick media for stories

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
@@ -2391,6 +2391,7 @@ public class EditPostActivity extends LocaleAwareActivity implements
                 case RequestCodes.MULTI_SELECT_MEDIA_PICKER:
                 case RequestCodes.SINGLE_SELECT_MEDIA_PICKER:
                 case RequestCodes.PHOTO_PICKER:
+                case RequestCodes.STORIES_PHOTO_PICKER:
                 case RequestCodes.STOCK_MEDIA_PICKER_SINGLE_SELECT:
                 case RequestCodes.MEDIA_LIBRARY:
                 case RequestCodes.PICTURE_LIBRARY:
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
@@ -1571,7 +1571,7 @@ public class EditPostActivity extends LocaleAwareActivity implements
                     FluxCUtils.mediaFileFromMediaModel(media));
         } else if (media != null && media.getMarkedLocallyAsFeatured() && media.getLocalPostId() == mEditPostRepository
                 .getId()) {
-            setFeaturedImageId(media.getMediaId());
+            setFeaturedImageId(media.getMediaId(), false);
         }
     }
 
@@ -2341,9 +2341,9 @@ public class EditPostActivity extends LocaleAwareActivity implements
         }
     }
 
-    private void setFeaturedImageId(final long mediaId) {
+    private void setFeaturedImageId(final long mediaId, final boolean imagePicked) {
         if (mEditPostSettingsFragment != null) {
-            mEditPostSettingsFragment.updateFeaturedImage(mediaId);
+            mEditPostSettingsFragment.updateFeaturedImage(mediaId, imagePicked);
         }
     }
 
@@ -2421,7 +2421,7 @@ public class EditPostActivity extends LocaleAwareActivity implements
                     // user chose a featured image
                     if (data.hasExtra(MediaPickerConstants.EXTRA_MEDIA_ID)) {
                         long mediaId = data.getLongExtra(MediaPickerConstants.EXTRA_MEDIA_ID, 0);
-                        setFeaturedImageId(mediaId);
+                        setFeaturedImageId(mediaId, true);
                     } else if (data.hasExtra(MediaPickerConstants.EXTRA_MEDIA_QUEUED)) {
                         if (mEditPostSettingsFragment != null) {
                             mEditPostSettingsFragment.refreshViews();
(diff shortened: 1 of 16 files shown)


WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -2421,7 +2421,7 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
                     // user chose a featured image
                     if (data.hasExtra(MediaPickerConstants.EXTRA_MEDIA_ID)) {
                         long mediaId = data.getLongExtra(MediaPickerConstants.EXTRA_MEDIA_ID, 0);
-                        setFeaturedImageId(mediaId);
+                        setFeaturedImageId(mediaId, true);
```
develric, line 2425: Probably I'm just missing something stupid but is this case for future steps we still need to cover? it seems I cannot reach here with the current capabilities we have for setting the featured image 🤔

WordPress/src/main/java/org/wordpress/android/ui/stories/StoriesMediaPickerResultHandler.kt
```diff
@@ -26,7 +25,7 @@ class StoriesMediaPickerResultHandler {
                         PagePostCreationSourcesDetail.STORY_FROM_MY_SITE
                 )
                 return true
-            } else if (isWPStoriesMediaBrowserTypeResult(data)) {
```
develric, line 29: I think we actually still need to discriminate if we are coming here from a stories picker or not, otherwise we risk to trigger the stories path even when not needed. For example:

- In My Site change the site icon
- In the picker select an image for the site icon and Confirm
- In this case we are landed to the story path but we shouldn't

develric (commented):
Hey @planarvoid , thanks for this PR 🙇 ! I did a first round of code review and testing. 

I think we have an issue with the `isWPStoriesMediaBrowserTypeResult` (see [comment inline](https://github.com/wordpress-mobile/WordPress-Android/pull/13007#discussion_r493933807)).

Also I get the rationale with the `IMAGE_PICKED` event but I wonder if that could change the numbers we track. I think we risk to trace the successful upload for the featured image instead of user picking a featured image (potentially a lower number even if I would not be able to say how much lower tbh, nor if it actually can make a relevant difference, but I thought better to mention). If I'm not wrong this case is related to the `EditPostActivity > setFeaturedImageId(mediaId, true);` code but tbh I was not able to trigger that code path 🤔 

I think I need some more code-review/testing time for this one but I wanted to anticipate some of the findings/thoughts, hope it helps 🙇

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
planarvoid: edit: You're right. I assume this will be used again once we implement other sources to the featured image picker.

WordPress/src/main/java/org/wordpress/android/ui/stories/StoriesMediaPickerResultHandler.kt
planarvoid: this is a great catch 👍 . I've solved this using 2 approaches for when the flag is on/off:
- When the flag is off, the behaviour should be exactly as it was before
- When the flag is on, there is a new RequestCode called STORIES_PHOTO_PICKER which is triggered only when we're creating a story. This was we stop mixing up the normal photo picker and the stories photo picker and the need for the check to `isWPStoriesMediaBrowserTypeResult` disappears. 

To test it check the flow both with and without the flag.

planarvoid:
thanks for the review @develric ! I've fixed your suggestion about the wp stories picker. 

> Also I get the rationale with the IMAGE_PICKED event but I wonder if that could change the numbers we track. I think we risk to trace the successful upload for the featured image instead of user picking a featured image (potentially a lower number even if I would not be able to say how much lower tbh, nor if it actually can make a relevant difference, but I thought better to mention). 

If I see it correctly, the only difference between the code now and before is that before it was tracked right before triggering the intent and now it's moved onActivityResult which happens right after the intent so we should be safe here.

I think this PR should be ready for another check.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
develric: In the case of flag off (like in Vanilla) I think we trigger the tracking event two times if we leave the `setFeaturedImageId`to true. I know we will need to review the tracking for this feature overall, but what if we replace the true with a check on if the flag is enabled? wdyt?

planarvoid: that sounds correct! good catch
I've added the check to the `org.wordpress.android.ui.posts.EditPostSettingsFragment#updateFeaturedImage`

develric (approved):
Thanks for the changes @planarvoid ; code looks good and works as per testing steps. Also your comment on IMAGE_PICKED event makes sense to me (tested on OnePlus 7T - Android 10). :shipit:

PR #13008, opened 2020-09-23
Media picker edit button tweaks

develric:
After having merged the #12984, while testing another PR I noticed a couple of tweaks we could add to the Media Picker edit button that I didn't notice during review (sorry @khaykov 🙇 ).

This PR do the following:

1. Adds a tooltip to the edit button when long pressed
2. Review the margin of the edit icon badge to account for the fact that we reduced the layout from 64dp to 48dp

### To test
- In media picker long press the edit icon and notice you get the tooltip

| image |
|---|

- In media picker select some images and note the badge is correctly aligned on right top of the icon

| image |
|---|

PR submission checklist:


khaykov (approved):
Thank's for the fixes! Looks good 👍

PR #13012, opened 2020-09-24
Reader: Fix deeplink activity disable/enable

### LINKED ISSUE #13010: Reader: Links opened from Reader sometimes trigger the deeplink component
Discovered in #13002 - this behavior is extra bad for Stories since they effectively link to themselves.

### Expected behavior

When the Reader attempts to open a URL externally, the WordPress app shouldn't be included in the list of possible apps to handle the link.

### Actual behavior

On certain devices/API levels, the WordPress app itself is one of the options.

### Background

We already [disable the Reader deeplink when opening an external URL](https://github.com/wordpress-mobile/WordPress-Android/blob/eb2497843d3ceb49567eb7292cde7cb703e97ec8/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java#L1091), so that the app doesn't appear as an option. We then [re-enable it](https://github.com/wordpress-mobile/WordPress-Android/blob/eb2497843d3ceb49567eb7292cde7cb703e97ec8/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java#L1111) (there and in a couple of other places for safety).

This behavior has been in place for years, and I'm not sure how recently it started to break down. As far as I can tell, on certain APIs and devices the re-enabling of the Reader deeplink component in that `finally` block happens before the `startActivity` call resolves the intent, so the Reader component which was removed is re-added to the list.

#### Reproduced on:
A Pixel 3a emulator, API 29 (@malinajirka's report above)
A Pixel 3 device, API 30

#### Could not reproduce on:
A Pixel 3XL emulator, API 29
A Pixel 3 emulator, API 28

aforcier:
Fixes #13010. Adjusts the way we disable/re-enable the Reader deeplink activity when opening an external URL so that it works correctly on all devices.

Relies on #13002 since that's where the issue was noticed, but I didn't want to bury this deep change in an apparently unrelated PR. That one should be mergable independent of this fix.

The first step to address the issue was to remove the re-enabling logic from [this `finally` block](https://github.com/wordpress-mobile/WordPress-Android/compare/feature/stories-reader-detail...issue/13010-fix-reader-deeplink-registration?expand=1#diff-cb472478231b297b0399fbbe2bde9f32L1111), which it seems is too soon for some devices, and we end up with the deeplink activity being re-enabled before the intent is handled and appearing when it was supposed to be removed. (My best guess is that this has been happening for a while and we just didn't notice, and/or that perhaps behavior changed in recent Android API levels so that `startActivity` unblocks the main thread earlier than before.)

However, this introduced a different issue: as soon as the chooser menu was shown, the app would close. This is because the `pm.setComponentEnabledSetting()` with `COMPONENT_ENABLED_STATE_DISABLED` call causes the system to close the activity in question (`ReaderPostPagerActivity`) in some cases. If the external URL was opened from the reader detail view, that's actually the current activity, and the app closes.

```
I/WindowManager:   Force finishing activity ActivityRecord{ea34428 u0 org.wordpress.android.beta/org.wordpress.android.ui.reader.ReaderPostPagerActivity t28021}
I/WindowManager:   Force finishing activity ActivityRecord{fadbc24 u0 org.wordpress.android.beta/org.wordpress.android.ui.main.WPMainActivity t28021}
```

You may be able to confirm this by checking out fb58104.

To address this, I moved the `intent-filter` rules for `ReaderPostPagerActivity` to an [activity-alias](https://developer.android.com/guide/topics/manifest/activity-alias-element). That way we enable/disable the alias activity for deeplinking instead, and the system never tries to force finish the real `ReaderPostPagerActivity`.

Finally, I also added a new call to re-enable deeplinking, when the app is backgrounded (we already do this when `WPMainActivity` is created or resumed). This should replace the work the `finally` block was doing, ensuring that deeplinks are re-enabled as the user leaves the app. This way links visited outside the app can still be picked up by the WordPress app and open the Reader, as before.

### To test:

Note that if you have multiple different builds of the WordPress app installed, the other builds will always show - that's expected and not something a regular user will experience.

1. Visit an external link from the Reader (the globe icon in the action bar in detail view works, but also check a Story block on WordPress.com).
2. Confirm that there is either no chooser screen (if you only have one browser installed), or that if there is a chooser the WordPress app isn't an option
3. Dismiss the chooser, visit another link, confirm the same behavior
4. Choose to open a link in the browser
5. Without returning to the app, visit a blog post link on a WordPress.com site
6. Confirm that the chooser appears, allowing you to open the post in the Reader

PR submission checklist:

### DIFF
--- a/WordPress/src/main/AndroidManifest.xml
+++ b/WordPress/src/main/AndroidManifest.xml
@@ -459,6 +459,10 @@
             android:label="@string/reader_title_deeplink"
             android:theme="@style/WordPress.NoActionBar"
             android:windowSoftInputMode="adjustResize" >
+        </activity>
+        <activity-alias
+            android:name=".WPComPostReaderActivity"
+            android:targetActivity=".ui.reader.ReaderPostPagerActivity">
             <intent-filter>
                 <data
                     android:host="wordpress.com"
@@ -513,7 +517,7 @@
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
             </intent-filter>
-        </activity>
+        </activity-alias>
         <activity
             android:name=".ui.reader.ReaderCommentListActivity"
             android:label="@string/reader_title_comments"
--- a/WordPress/src/main/java/org/wordpress/android/WordPress.java
+++ b/WordPress/src/main/java/org/wordpress/android/WordPress.java
@@ -115,6 +115,7 @@ import org.wordpress.android.util.SiteUtils;
 import org.wordpress.android.util.UploadWorker;
 import org.wordpress.android.util.UploadWorkerKt;
 import org.wordpress.android.util.VolleyUtils;
+import org.wordpress.android.util.WPActivityUtils;
 import org.wordpress.android.util.analytics.AnalyticsUtils;
 import org.wordpress.android.util.config.AppConfig;
 import org.wordpress.android.util.image.ImageManager;
@@ -928,6 +929,9 @@ public class WordPress extends MultiDexApplication implements HasServiceInjector
             properties.putAll(mReaderTracker.getAnalyticsData());
 
             mReaderTracker.onAppGoesToBackground();
+
+            // Ensure that the deeplinking activity is re-enabled.
+            WPActivityUtils.enableReaderDeeplinks(getContext());
 
             AnalyticsTracker.track(AnalyticsTracker.Stat.APPLICATION_CLOSED, properties);
             AnalyticsTracker.endSession(false);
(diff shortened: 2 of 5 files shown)


malinajirka (approved):
TIL about activity-aliases ;).

The changes look good to me. Great job on figuring it out!!!

SideNote: I found one super minor issue, but I personally wouldn't worry about it. When you open the "choose app" dialog and rotate the device, the app appears in the dialog.

aforcier:
> SideNote: I found one super minor issue, but I personally wouldn't worry about it. When you open the "choose app" dialog and rotate the device, the app appears in the dialog.

😀 yeah agreed, let's just stop here, good thinking though 😉

PR #13026, opened 2020-09-29
Issue/fix my site layout on bigger screens

khaykov:
This PR fixes issues with My Site screen layout present on tablets and bigger screens, according to this design:

proposed-design-1

What changed:
- Content area of My Site is limited to 500dp width (we eliminated screen size specific padding).
- Quick Action buttons have an equal distance between them withing a 500dp container (they were spaced equally before)

Phone layout:
[Image from Gyazo](https://gyazo.com/6b37503a63e0cf34c3b7bb696f3365a3)

Domain CTA:
[Image from Gyazo](https://gyazo.com/462a0ba0a4a03518230077f4300b8d07)

Tablet:
[Image from Gyazo](https://gyazo.com/514f8acac9a2b79554f7fa2f78e9d1e5)

To test:
- Make sure My Site layout resembles the proposed design and nothing looks out of place.

PR submission checklist:


osullivanchris (commented):
Hey @khaykov ! Thanks for such a quick fix. It looks so much better! I only have a phone to test the APK on. But it looks great on landscape and portrait, and the example screenshots look good too.

Are you happy with the solution? Is 500dp an ok number to use or is it conflicting with anything else?

I have one further tweak. A little tricky to explain so I'll try break it down.

- In the build, the site-switcher chevron and the Quick Action button are right aligned to the edge of the container.

Screenshot 2020-09-29 at 13 47 01

- This is correct of course. However, I feel that the chevron optically looks better with a little bit of extra right-margin inside the container. I actually had it in designs but wanted to see how the build panned out first. 
- Below is how it looks with an extra 4dp right-margin/padding on the site switcher icon. Both with and without guide lines

Screenshot 2020-09-29 at 13 49 20

I think it just looks better to the eye. So that's the first part, just adding 4dp padding. However there's a second tweak, I'm not sure about this one. 

- If you look at the following spec, the chevron now _almost_ lines up centrally with the profile image, but not quite. I also notice that the content area has a slightly large margin to the right than the profile image (the profile image overlapping the green margin area)
Screenshot 2020-09-29 at 13 53 39


- In the below image, if we nudge the profile picture in about 7-8dp, it has both the same right alignment as the content area, and lines up centrally with the (new position) site switcher. Again, showing the same thing with and without guidelines.

Screenshot 2020-09-29 at 13 53 23

I'm not so sure about this second change because
- the profile picture might be a set location by the system app bar
- this will only apply at most to phone anyway, as the app bar is outside of our content area and will float right per the tablet examples
- Even if you can get this to work, I'm not sure if it will line up on all phones of different DPI

let me know what you think 😅  Even the first tweak would be enough. Just wanted to consider this.

khaykov:
@osullivanchris I don't mind 500dp context area - before we used to apply extra padding at 600dp and brekpoints 700dp, so it's pretty close.

I aligned everything according to your mocks. I had to eyeball the avatar alignment, but I think it looks mostly centered. :)

[Image from Gyazo](https://gyazo.com/cf6aa48f100941aad8c583120de7b96c)

I also found that the long quick action button labels mess up with spacing, so I limited them to the width of the button (so they will wrap like this)

[Image from Gyazo](https://gyazo.com/673d7b61564c6f149ac26a08c78d071e)

Just a piece of info about the chevron - it has a 48dp bounding box so we can have a bigger touch area, like this:

[Image from Gyazo](https://gyazo.com/d9260c792fd85acf90100244a73c2469)

osullivanchris:
> I aligned everything according to your mocks. I had to eyeball the avatar alignment, but I think it looks mostly centered. :)

first, thanks for humouring my very specific request!! Second, looks good. Can I find it in the build if I re-download, or something like that? Also will that alignment hold true for different device sizes and resolutions, if you had to eyeball it?

> I also found that the long quick action button labels mess up with spacing, so I limited them to the width of the button (so they will wrap like this)

Good thinking. You could also allow the label width to be something like 4dp wider than the Quick Action. Its not so long that it will look disconnected from the Quick Action, but it just gives a bit of wiggle room to get an extra character or two in before it wraps.

Screenshot 2020-09-30 at 12 36 29

> Just a piece of info about the chevron - it has a 48dp bounding box so we can have a bigger touch area, like this:

Ah yes. Ok. That's good to know. The exact values / layout I was proposing didn't translate exactly then. Per point 1 I'm happy to go with your eyeball-ed version.

khaykov:
@osullivanchris 
> first, thanks for humouring my very specific request!! Second, looks good. Can I find it in the build if I re-download, or something like that? Also will that alignment hold true for different device sizes and resolutions, if you had to eyeball it?


Np! Yes, I think you can just redownload it and the changes should be reflected. Let me know if they are not there! 
Alignment of chevron will remain the same on all the devices, but the avatar will not be aligned when we reach the +500dp width, since the avatar is part of the AppBar and not the content area.

> Good thinking. You could also allow the label width to be something like 4dp wider than the Quick Action. Its not so long that it will look disconnected from the Quick Action, but it just gives a bit of wiggle room to get an extra character or two in before it wraps.

Good idea! Will do 👍

osullivanchris:
> Np! Yes, I think you can just redownload it and the changes should be reflected. Let me know if they are not there!

downloaded and it worked. The alignment looks good on my device too. Really happy with this now 👌 

> Alignment of chevron will remain the same on all the devices, but the avatar will not be aligned when we reach the +500dp width, since the avatar is part of the AppBar and not the content area.

cool. Yes I suppose it should carry across different resolutions fine if the value is in DP. Yeah expected on +500dp width, understood!

> Good idea! Will do 👍

Sweet!

Very happy with this now. Thanks for the tweaks.

khaykov:
Ready for the code review 👍 I recommend hiding whitespace changes for better readability.

develric (approved):
Hey @khaykov , @osullivanchris , really like how it looks 😊 ! Code LGTM and checked that screens appear as per design pics. Tested with a Pixel 3 Emu API 29 and a 10.1 WXGA Tablet API 26 and didn't notice deviations. :shipit:

designsimply:
LGTM! Tested with WPAndroid 15.9-rc-2 on Samsung Galaxy Tab4 (SM-T330NU), Android 5.1.1.

Screenshot_2020-10-13-12-48-18 Screenshot_2020-10-13-12-48-53

osullivanchris:
Not specific to this issue but the default avatar thing in @designsimply 's screenshot looks wrong. Is that a custom avatar or is something broken?

designsimply:
> Not specific to this issue but the default avatar thing in @designsimply 's screenshot looks wrong. Is that a custom avatar or is something broken?

It's a custom avatar! Not broken. 🙂

PR #13034, opened 2020-09-30
Alleviate time taken to populate media list in media picker

### LINKED ISSUE #12996: Delay in displaying images from the phone
### Expected behavior
The images from the phone should be displayed immediately

### Actual behavior
We need to wait for a few seconds for the images from the phone to get listed on the screen

1. In the post/page editor, add an image block
2. Choose option add image > choose from device.
3. You will need to wait for a few seconds(~10 seconds) to get the images from the phone listed on the screen.

mzorz:
Fixes #12996 

This PR intends to make the time taken by the filter more relaxed by moving the isSupportedMimeType call to the viewHolder bind() method, instead of running through the whole data set at first attempt.

Now instead of directly not showing items that we know are not supported, we paint a "disabled" overlay and remove / avoid setting any click listeners on such items.

To test:
1. open a post, add an image block, tap on ADD IMAGE
2. on the bottom sheet, tap Add media from device
3. observe the media gets populated quicker than before
4. observe the unsupported items have an overlay

Also note / side effect:
- scrolling is not smooth so, leaving this as a draft PR as of  now

PR submission checklist:


mzorz:
Closing, given this served just as a PoC

PR #12824, opened 2020-08-28
Filtering media items in picker to only show supported mime types

mzorz:
Fixes #12822

I checked the FluxC code to see what is considered a valid MIME type, the check comes from [this here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/fluxc/src/main/java/org/wordpress/android/fluxc/utils/MimeTypes.kt) based on the Wordpress documentation available here https://wordpress.com/support/accepted-filetypes/

I've tested an unsupported format such as `webp` to double check
 nevertheless on a self-hosted site and a .com site and was able to confirm the media section there does not allow you to upload other than supported file extensions.

Screen Shot 2020-08-28 at 10 30 53

### Before
In these gifs, the first image to the left is a `webp` (unsupported mime type) image
Aztec:
mimetypenofilter_aztec

Gutenberg:
mimetypenofilter_gb

### After
As you can see, the image with an unsupported mime type is filtered out, not allowing it to be inserted
Aztec:
mimetypefilter_aztec

Gutenberg:
mimetypefilter_gb

To test:
1. download a webp or another WP-unsupported media file on Chrome on your device  / emulator. I used the examples given here https://developers.google.com/speed/webp/gallery1
2. go to the WP app, start a Post, and try uploading an image
3. observe the just downloaded image (as per step 1) is not shown

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13035, opened 2020-09-30
Removed the `contentResolver.getMimeType - isSupportedMimeType() `  check for populating the picker

### LINKED ISSUE #12996: Delay in displaying images from the phone

mzorz:
Fixes #12996 

After discussion and attempting an alternative solution in #13034, this PR effectively unmakes the change introduced in https://github.com/wordpress-mobile/WordPress-Android/pull/12824 with which we wanted to offer users only those media item types that were supported on their WP sites.

To test:
1. create a GB post, add an image block
2. tap on Add Image
3. on the bottom sheet, choose "Choose from device"
4. observe the picker shows up and loads image thumbnails without issues

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/mediapicker/DeviceListBuilder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/mediapicker/DeviceListBuilder.kt
@@ -14,7 +14,6 @@ import android.webkit.MimeTypeMap
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.async
 import kotlinx.coroutines.withContext
-import org.wordpress.android.fluxc.utils.MediaUtils
 import org.wordpress.android.fluxc.utils.MimeTypes
 import org.wordpress.android.modules.BG_THREAD
 import org.wordpress.android.ui.mediapicker.MediaSource.MediaLoadingResult
@@ -108,9 +107,7 @@ class DeviceListBuilder
                         getMimeType(uri),
                         dateModified
                 )
-                if (MediaUtils.isSupportedMimeType(context.contentResolver.getType(uri))) {
-                    result.add(item)
-                }
+                result.add(item)
             }
         } finally {
             SqlUtils.closeCursor(cursor)
--- a/WordPress/src/main/java/org/wordpress/android/ui/photopicker/DeviceMediaListBuilder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/photopicker/DeviceMediaListBuilder.kt
@@ -7,7 +7,6 @@ import android.provider.MediaStore.Images.Media
 import android.provider.MediaStore.Video
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.withContext
-import org.wordpress.android.fluxc.utils.MediaUtils
 import org.wordpress.android.modules.BG_THREAD
 import org.wordpress.android.ui.media.MediaBrowserType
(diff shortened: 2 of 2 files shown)


WordPress/src/main/java/org/wordpress/android/ui/mediapicker/DeviceListBuilder.kt
```diff
@@ -108,9 +107,7 @@ class DeviceListBuilder
                         getMimeType(uri),
                         dateModified
                 )
-                if (MediaUtils.isSupportedMimeType(context.contentResolver.getType(uri))) {
-                    result.add(item)
-                }
+                result.add(item)
```
planarvoid, line 110: I'm not sure why we haven't done that before but we're already calling the `getMimeType(uri)` 3 lines above this code so we could be using the mime type loaded there. However, we can do that in another PR. I think the `MediaUtils.isSupportedMimeType` is quite inefficient. We can figure out something better. 

edit: I missed that this is the updated picker, not the one that's currently released :-)

planarvoid (approved):
Looks good, thanks for the quick fix!

develric (approved):
Jumping in to confirm I tested this one with my scenario (OnePlus 7T with 8000+ media on device) and works as expected (no delay when opening picker). Thanks for the quick fix @mzorz and also for exploring the solution in #13034 🙇

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/DeviceListBuilder.kt
mzorz: > I'm not sure why we haven't done that before but we're already calling the getMimeType(uri) 3 lines above this code so we could be using the mime type loaded there. However, we can do that in another PR. I think the MediaUtils.isSupportedMimeType is quite inefficient. We can figure out something better.

Agreed it is inefficient and as such, contributing to time consumption while building a list seems like something we don't want. 

A proposed solution is to only check in the entry points once the item has been selected, as described here https://github.com/wordpress-mobile/WordPress-Android/issues/13036

I'd say this is something to take into account and we should probably remove the `context.contentResolver.getType(uri)` call from this class altogether as well - other solutions (for example involving pre-fetching) might apply but seem potentially beyond the effort for an issue we haven't heard many times from users anyway. Let me know your thoughts @planarvoid 👍

planarvoid: I was thinking about implementing pagination to solve this issue in the new media picker. I don't think it's very efficient to load all the images from the device anyway. I can imagine some users having tens of thousands of items and that could definitely cause other issues.

mzorz: Ah definitely a great idea! That passed over my head.
I was leaning on checking in the entry points given we may also need to filter out things per entry point / consumer of the picker to have a more granular control over the broad "supported by WP site" or the "give me video" or "give me images" sort of flags (although can't think of a good enough use case right now anyway, and if such a need appears it would still make more sense to filter in the picker right away)
Looking forward to seeing the new picker 💯

PR #13047, opened 2020-10-01
Feature/mark referrers as spam

### LINKED ISSUE #11818: Stats: Referrers: Not Possible to Mark a Referrer as Spam
In Calypso, referrers which are not whitelisted can be marked spam by tapping the three dots to the right of the referrer. This is not the case in the WordPress Android app (or iOS).

Expected:
I expected to be able to mark referrers as spam from within the app.

Actual: 
The option to mark referrers as spam doesn’t exist within the app.

Stats referrers in WPAndroid version alpha-220 and Calypso web (same site) on moto e5 play phone:
androidreferrers

##### Tested on Moto e5 Play, WPAndroid alpha-220
This came up in #2934822-zen when a user wanted to mark some referrers as spam but couldn't.

marco-altran:
Fixes #11818 

Depends on: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1704

This PR adds the possibility to mark and unmark a referrer as spam by accessing the `reportReferrerAsSpam` and `unreportReferrerAsSpam` methods from `ReferrersStore`. The functionally was added through a menu popup that is activated on long press:

wordpress-referrer-3

The desired state is that when marked as spam, the referrer is greyed out and the user still has the ability to unmark as spam. If he does a pull to refresh, the referrer disappears (except if it's a search engine where the WP API ignores the report). This is very similar to the existing behavior in wordpress.com.

To test:

Go to Stats -> select the period (select YEARS for maximum referrals) -> go to referrers -> long press -> mark/unmark as spam

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/StatsListViewModel.kt
```diff
@@ -27,6 +27,7 @@ import org.wordpress.android.ui.stats.refresh.lists.StatsListViewModel.StatsSect
 import org.wordpress.android.ui.stats.refresh.utils.ItemPopupMenuHandler
 import org.wordpress.android.ui.stats.refresh.utils.NewsCardHandler
 import org.wordpress.android.ui.stats.refresh.utils.StatsDateSelector
+import org.wordpress.android.ui.stats.refresh.utils.StatsDateSelector.Factory
```
planarvoid: I think I'd revert this change. It removes a little bit of readability and you're not touching this file for any other reason.

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
```diff
@@ -124,6 +124,7 @@ sealed class BlockListItem(val type: Type) {
         val showDivider: Boolean = true,
         val textStyle: TextStyle = TextStyle.NORMAL,
         val navigationAction: NavigationAction? = null,
+        val menuAction: ((View) -> Boolean)? = null,
```
planarvoid: I know it's not consistent with the `navigationAction` but I think I'd call this one `longClickAction` and (and that's for another PR) I'd rename the `navigationAction` to something like `clickAction`. The reason is that the UI doesn't need to know that the action navigates somewhere or opens a menu. It's just necessary to know what user action should trigger it. Does it make sense? However, this is not super important :-)

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/granular/usecases/ReferrersUseCase.kt
```diff
@@ -180,7 +193,12 @@ constructor(
         return items
     }
 
-    private fun buildIcon(iconUrl: String?): Int? {
+    private fun buildTextStyle(spam: Boolean) = if (spam) LIGHT else TextStyle.NORMAL
+
+    private fun buildIcon(iconUrl: String?, spam: Boolean): Int? {
+        if (spam) {
```
planarvoid: I think I'd merge this condition into the `when` bellow into something like:
```
when {
spam -> ...
iconUrl == null -> ...
...
}

File: WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/BlockListItemViewHolder.kt [hunk 4]
```diff
@@ -49,6 +49,7 @@ open class BlockListItemViewHolder(
                 imageManager.load(this, item.icon)
             }
             item.iconUrl != null -> {
+                this.imageTintList = null
```
planarvoid, line 52: 👍

libs/analytics/WordPressAnalytics/src/main/java/org/wordpress/android/analytics/AnalyticsTracker.java
```diff
@@ -102,6 +102,12 @@
         STATS_POSTS_AND_PAGES_ITEM_TAPPED,
         STATS_REFERRERS_VIEW_MORE_TAPPED,
         STATS_REFERRERS_ITEM_TAPPED,
+
+        // FIXME Verify p2 post before adding these stats
+        STATS_REFERRERS_ITEM_LONG_PRESSED,
```
planarvoid, line 105: Make sure you also add the mapping to the correct value to `org.wordpress.android.analytics.AnalyticsTrackerNosara#getEventNameForStat`. Usually we just map the event to lowercase string.

planarvoid (commented):
It's looking good and works really well! I have a few comments for the code.

Looking at the WordPress.com functionality, I think it would be better if the `mark as spam` option was enabled on the group of referrers instead the children. If you mark a child of a group as a spam, the whole group disappears anyway. What do you think? We can have it enabled also on children but I think it makes sense to have it on the parent. 

I was trying to find out how the `Search Engines` and the `Android App` items are disabled for the `mark as spam` feature in WordPress.com and from the response I have no idea. I'm expecting some kind of a hack.

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
marco-altran: I totally agree, it was super confusing to me that the `navigationAction` actually is the `onItemClick` I was looking for. I will make this change.

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/granular/usecases/ReferrersUseCase.kt
marco-altran: Good thinking. I will make the change

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/StatsListViewModel.kt
marco-altran: I did some modifications in this class that was later rolled back. I kept this one because it solved an Android Studio warning. I am fine with both, please let me know what you think it's best.

libs/analytics/WordPressAnalytics/src/main/java/org/wordpress/android/analytics/AnalyticsTracker.java
marco-altran: Got it, I will make this change.

marco-altran: Fixed in [af23b7e](https://github.com/wordpress-mobile/WordPress-Android/pull/13047/commits/af23b7e98f96188ee4c047c25e92eb9f8d586c3c)

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/granular/usecases/ReferrersUseCase.kt
marco-altran: Fixed in [af23b7e](https://github.com/wordpress-mobile/WordPress-Android/pull/13047/commits/af23b7e98f96188ee4c047c25e92eb9f8d586c3c)

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/BlockListItem.kt
marco-altran: Fixed in [af23b7e](https://github.com/wordpress-mobile/WordPress-Android/pull/13047/commits/af23b7e98f96188ee4c047c25e92eb9f8d586c3c)

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/StatsListViewModel.kt
marco-altran: Fixed in [af23b7e](https://github.com/wordpress-mobile/WordPress-Android/pull/13047/commits/af23b7e98f96188ee4c047c25e92eb9f8d586c3c)

marco-altran:
> It's looking good and works really well! I have a few comments for the code.
> 
> Looking at the WordPress.com functionality, I think it would be better if the `mark as spam` option was enabled on the group of referrers instead the children. If you mark a child of a group as a spam, the whole group disappears anyway. What do you think? We can have it enabled also on children but I think it makes sense to have it on the parent.

Fixed with [5c226f6](https://github.com/wordpress-mobile/WordPress-Android/pull/13047/commits/5c226f6825ec1e0e7ec158f2e72e0678cbcee86f) and [9b4893e in FluxC](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1704/commits/9b4893e13db6571a9116d5748d910d5e51ea9874)

Keep in mind that the functionality for marking a child as spam still exists, I just deactivated through the removal of `onMenuAction`. It will be easier to change the behavior if we want to, but I am happy to completely rollback these changes if you think it's better that way.

> I was trying to find out how the `Search Engines` and the `Android App` items are disabled for the `mark as spam` feature in WordPress.com and from the response I have no idea. I'm expecting some kind of a hack.

I was wrong in thinking that the behavior was caused by some groups having an URL and others don't. Indeed we have groups with and without URLs, but it has nothing to do but it had no correlation with the behavior. I noticed empirically that all groups that can be marked as spam can be resolved to a valid URL, so I added a validation using `UrlUtils.isValidUrlAndHostNotNull("https://${group.name}")` and a small modification in FluxC to replicate this behavior. It's a hack, but it's working exactly like in wordpress.com. Let me know what you think about it.

planarvoid (approved):
Thanks for all the changes! It's looking really great, good job 👍 

It's looking good and works really well! I have a few comments for the code.


> Keep in mind that the functionality for marking a child as spam still exists, I just deactivated through the removal of onMenuAction. It will be easier to change the behavior if we want to, but I am happy to completely rollback these changes if you think it's better that way.

This is more than enough 👍 

> I was wrong in thinking that the behavior was caused by some groups having an URL and others don't. Indeed we have groups with and without URLs, but it has nothing to do but it had no correlation with the behavior. I noticed empirically that all groups that can be marked as spam can be resolved to a valid URL, so I added a validation using UrlUtils.isValidUrlAndHostNotNull("https://${group.name}") and a small modification in FluxC to replicate this behavior. It's a hack, but it's working exactly like in wordpress.com. Let me know what you think about it.

Good catch 👍 , it's working well.

planarvoid (approved):
It looks good and works well! Thanks for the updates 👍

PR #13048, opened 2020-10-02
Introduce pagination for device library

planarvoid:
This PR is solving a problem where we want to add pagination to the "Choose from device" list. This issue is actually quite complicated because we have 4 independent sources (all of which need to be paginated) and then we have to make sure we load the merged data in the correct order. The pagination is necessary because the content resolver mime type check is a heavy operation and if you have a lot of media files, it could show you an empty list for quite some time. 

Because the logic is complicated, I've added a lot of tests to cover all the possible scenarios (and extracted the android specific logic into a new file). Let me know what you think!

To test:
- Stress test the "Choose from device" loader
- try devices with many files/images, try that the pagination works as expected and nothing jumps around as you scroll
- try devices with no images
- check that loading isn't slow
- set the PAGE_SIZE to something smaller to make it easier to test pagination

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mediapicker/loader/DeviceMediaLoader.kt
```diff
@@ -0,0 +1,141 @@
...
+                dateCondition ?: filterCondition
+            }
+            cursor = context.contentResolver.query(
+                    baseUri,
+                    projection,
+                    condition,
+                    null,
+                    "$ID_DATE_MODIFIED DESC LIMIT ${(pageSize + 1)}"
+            )
+        } catch (e: SecurityException) {
```
develric: Wondering if we should catch on a wider exception also, wdyt? Or maybe this is done on purpose to test against the content resolver? 🤔

develric (commented):
Hi @planarvoid , I like the idea behind this pagination. We will need to full proof if the access to the content resolver penalty is attenuated enough, but idea as said looks good 👍 !

I would like to make some additional tests but starting leaving a couple of notes below:

1. with a physical device running API29, I get the following (but I guess it could be the same exception for which we released a beta update isn't it? If so it should be like expected I think until we merge back that fix)

```
21196-21337/org.wordpress.android.beta E/AndroidRuntime: FATAL EXCEPTION: DefaultDispatcher-worker-2
    Process: org.wordpress.android.beta, PID: 21196
    java.lang.IllegalStateException: storagePublicDirectory.listFiles() must not be null
        at org.wordpress.android.ui.mediapicker.loader.DeviceMediaLoader.loadDocuments(DeviceMediaLoader.kt:99)
        at org.wordpress.android.ui.mediapicker.loader.DeviceListBuilder$loadDownloads$2.invokeSuspend(DeviceListBuilder.kt:106)
        at org.wordpress.android.ui.mediapicker.loader.DeviceListBuilder$loadDownloads$2.invoke(Unknown Source:10)
        at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatchedOrReturn(Undispatched.kt:91)
        at kotlinx.coroutines.BuildersKt__Builders_commonKt.withContext(Builders.common.kt:160)
        at kotlinx.coroutines.BuildersKt.withContext(Unknown Source:1)
        at org.wordpress.android.ui.mediapicker.loader.DeviceListBuilder.loadDownloads(DeviceListBuilder.kt:101)
        at org.wordpress.android.ui.mediapicker.loader.DeviceListBuilder$load$2$invokeSuspend$$inlined$map$lambda$2.invokeSuspend(DeviceListBuilder.kt:50)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:738)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```

2. on an emu with API28 if I upload 48 test images into Download folder in the emu, the picker suddenly cannot load images at all (possibly it can be related to the set of test images I used or the emulator that is pretty slow, I will try to narrow the reason of this and come back here 🙇  )

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/loader/DeviceMediaLoader.kt
planarvoid: I wasn't sure why this exception was caught here so I checked and I think it was just copy & paste from the photo picker. I've removed it so any problem here doesn't get lost. WDYT?

planarvoid:
hey @develric , thanks for the review! 

> with a physical device running API29, I get the following (but I guess it could be the same exception for which we released a beta update isn't it? If so it should be like expected I think until we merge back that fix)

I don't think our fix is merged back to develop yet so I've added the legacy flag to android manifest. This should work now 👍 

> on an emu with API28 if I upload 48 test images into Download folder in the emu, the picker suddenly cannot load images at all (possibly it can be related to the set of test images I used or the emulator that is pretty slow, I will try to narrow the reason of this and come back here 🙇 )

I think this is because of the emulator not adding the images correctly to the media store. There is nothing we can do about this (I've experienced the same thing)

develric (approved):
Hi @planarvoid 😄 

> I don't think our fix is merged back to develop yet so I've added the legacy flag to android manifest. This should work now 👍

Confirm now it works well also in API 29 OnePlus 7T. Checked it having 8000+ including photos/videos/pdf and AFAIU pagination works well 👍 

> on an emu with API28 if I upload 48 test images into Download folder in the emu, the picker suddenly cannot load images at all 

Checked this again also, and adding the images not passing for AS but downloading items to emulator from google drive. In this case I confirm that the picker worked well also 👍 

:shipit:

PR #13062, opened 2020-10-06
Add empty screen handling to media picker

planarvoid:
This PR adds an empty screen to all the data sources. There are two empty screens for the pexel data source - one is shown before the search and one when the search result is empty. The other two data sources use standard empty screens:

Pexel initial screen:
Screenshot_1601987969
Empty search results:
Screenshot_1601987976

To test:
- Try to add media from the stock library
- Notice the empty screen before you enter query
- Enter a query that doesn't return any results
- Notice a different empty screen
- Try search for non-sense queries in the media library and the device library
- Notice the empty screens

PR submission checklist:


develric (commented):
Hey @planarvoid , thanks for this PR. Code looks good and works well 👍 . Only comment I have is related to the empty state when search is not matched. I think we should place it above the keyboard like I see for example in Tenor for the initial empty state with image (that is slided up when opening the keyboard), wdyt?

| Empty state hidden by keyboard | Empty state not hidden by keyboard |
|---|---|
| image | image |

planarvoid:

> Only comment I have is related to the empty state when search is not matched. I think we should place it above the keyboard like I see for example in Tenor for the initial empty state with image (that is slided up when opening the keyboard), wdyt?

I've used the `updateLayoutForSearch` method so now it should all work as you describe and not hide the empty view.

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerFragment.kt
```diff
@@ -379,8 +382,29 @@ class MediaPickerFragment : Fragment() {
...
+                    actionable_empty_view.subtitle.visibility = View.GONE
+                }
+
+                if (uiModel.image != null) {
+                    actionable_empty_view.image.setImageResource(uiModel.image)
+                    actionable_empty_view.image.visibility = View.VISIBLE
+                } else {
+                    actionable_empty_view.image.visibility = View.GONE
+                }
+                actionable_empty_view.updateLayoutForSearch(uiModel.isSearching, 0)
```
develric: hi @planarvoid, I tested this again but I'm now getting empty state without image in Stock media. Tried to move the `updateLayoutForSearch` call right after the `is Empty -> {` block and it worked; so probably that call effect is dependent on call order. Let me know wdyt 🙂 .

| From this PR | Moving the updateLayoutForSearch call at the beginning of the block |
|---|---|
| image | image |

planarvoid: good catch @develric, it's fixed in 77619bab3eea81e483bb896c43bf96217f844d59

planarvoid:
I've added tenor empty screen in c1122747602aafb59d6b0fa7788155d49988cd11

Screenshot_1602599521

develric (approved):
Hi @planarvoid 👋 . Thanks for the changes (and for adding empty state for tenor 🙇 ). LGTM :shipit:

PR #13063, opened 2020-10-06
Copy content provider files into local memory to enable upload on API 29

### LINKED ISSUE #13055: Permission issue while reading images from device
While reviewing the #13009 PR I noticed the issue described below. Tracking here since it is something to further investigate.

### Expected behavior
I expect to be able to add an image to a GB image block from device.

### Actual behavior
At least with code from the above mentioned PR when trying to add an image from device to the GB image block in logcat I get the error `7103-7103/org.wordpress.android.beta E/BitmapFactory: Unable to decode stream: java.io.FileNotFoundException: /storage/emulated/0/Download/amazing-animal-beautiful-beautifull.jpg: open failed: EACCES (Permission denied)` and the following error is shown in the UI

<p align="center">
  
</p>

- In GB editor add an image block
- Try to add an image from device
- Notice you get the error as per above image and log

##### Tested on [Pixel 3 Emu], Android [API 29], WPAndroid [13009 PR]

**NOTE:** as suspected by @planarvoid it seems related to API 29 permissions management. With API 28 the issue is not present.

planarvoid:
Fixes #13055

From my investigation I think it's no longer possible to work with the content provider URIs in the `File` class ([link](https://developer.android.com/training/data-storage/shared/media#open-file)). That's why we have to copy the file into local memory in order to upload it correctly to the media library.

The `storagePublicDirectory.listFiles()` can be null on API 29 so I've added check for that as well.

Since API 29 update also broke taking photos and videos, I've added the fix to this PR. 

To test (with a new device running API 28, 29 and 30):
- Go to Gutenberg editor
- Add image block
- Click on "Add Media"/"Choose from device"
- Select an item
- Confirm it
- Notice it gets uploaded correctly

To test (with a new device running API 28, 29 and 30):
- Go to Gutenberg editor
- Add image block
- Click on "Add Media"/"Take photo"
- Take a photo
- Notice it gets uploaded correctly

To test (with a new device running API 28, 29 and 30):
- Go to Gutenberg editor
- Add video block
- Click on "Add Media"/"Take video"
- Take a video
- Notice it gets uploaded correctly

To test (with a new device running API 28, 29 and 30):
- Go to Media library
- Click on "Take a photo"
- Take a photo
- Notice it gets uploaded correctly

To test (with a new device running API 28, 29 and 30):
- Go to Media library
- Click on "Take a video"
- Take a video
- Notice it gets uploaded correctly

PR submission checklist:


develric (commented):
Hi @planarvoid , AFAICT the changes LGTM, also the linked resource and motivation makes sense. I'm also not getting the NPE in `DeviceListBuilder` anymore (as we would now expect). Some more things I would like to mention just to check they make sense to you:

1. When adding an image from device to GB image block I still get the error `15502-15502/org.wordpress.android.beta E/BitmapFactory: Unable to decode stream: java.io.FileNotFoundException: /storage/emulated/0/Download/pexels-photo-531767.jpeg: open failed: EACCES (Permission denied)`

develric (commented):
Hi @planarvoid , AFAICT the changes LGTM, also the linked resource and motivation makes sense. I'm also not getting the NPE in `DeviceListBuilder` anymore (as we would now expect). Some more things I would like to mention just to check they make sense to you:

1. When adding an image from device to GB image block I still get the error 
`15502-15502/org.wordpress.android.beta E/BitmapFactory: Unable to decode stream: java.io.FileNotFoundException: /storage/emulated/0/Download/pexels-photo-531767.jpeg: open failed: EACCES (Permission denied)`
but the image is loaded correctly into the block
2. When trying to add from device an image into the media library I get
`15502-15502/org.wordpress.android.beta E/BitmapFactory: Unable to decode stream: java.io.FileNotFoundException: /storage/emulated/0/Download/pexels-photo-531767.jpeg: open failed: EACCES (Permission denied)`
and it fails to upload with the following snackbar

<p align="center">
  
</p>

Since you are not replicating it, I'm still wondering if it could be something related to my Pixel 3 Emu API 29 setup in any way; let align tomorrow in case we can make some additional check on my setup 👍

planarvoid:
Hey @develric , thanks for the testing! I was able to get the same results as you. I've fixed it by using the `CopyMediaToAppStorageUseCase` directly in the `DeviceListInsertUseCase`. As part of this change I've removed the conditional `queueResults` part because it no longer makes sense. Now we're copying all the files. I've also removed tracking because I'll be moving it to the view model in another PR. I've checked GB and media library and now it works in both places for me.

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/CopyMediaToAppStorageUseCase.kt
```diff
@@ -20,18 +20,13 @@ class CopyMediaToAppStorageUseCase @Inject constructor(
    * Some media providers (eg. Google Photos) give us a limited access to media files just so we can copy them and then
    * they revoke the access. Copying these files must be performed on the UI thread, otherwise the access might be
    * revoked before the action completes. See https://github.com/wordpress-mobile/WordPress-Android/issues/5818
+   *
+   * From API 29 we need to download also the files from the media store
    */
     suspend fun copyFilesToAppStorageIfNecessary(uriList: List<Uri>): CopyMediaResult {
```
develric: Maybe we could rename it to `copyFilesToAppStorage` now; wdyt?

develric (commented):
Thanks for this PR @planarvoid 🙇 ! I tested the scenarios and they worked well. I'm leaving a small np in line if you think it's worth. One more thing to mention is that while testing this PR I got an issue with adding image from taking a photo both in adding to the Media Library and to an image GB block. I don't think it's related to this PR but wanted to mention in case you think different. Let me know if you prefer me to open an issue for that 👍 !

Update: it seems the add by taking a photo is not working (at least for me) also in current 15.9 beta. Could not look into code today but if you can confirm the behaviour and that's not related to this set of changes (as I think), let sync on this tomorrow so we can evaluate if makes sense to ping anyone and think to a beta update 🙇

planarvoid:
Thanks for the review @develric , I've fixed taking the photo and video in GB editor, photo picker (for example in site icon selection) and in the media library. Once we test and approve this, we can probably release a beta update.

planarvoid:
I'm closing this PR cause it's no longer necessary

PR #13042, opened 2020-10-01
Reader Detail Updates (Post Header)

ashiagr:
Task: #12900

This PR refactors post details header view with following changes:

Child PRs:

PR #13060 - Kotlin conversion
PR #13061 - `uiState` extracted for blog section and follow button

- Post header view updated using `uiState` extracted in PR #13061 03db5a1
- Post title moved from `reader_include_post_detail_content.xml` 8790672
- Separated blog post author 8790672

**Merge instructions**

1. Make sure #13060, #13061 are merged
2. Change branch to "issue/12900-rip3-detail-updates-main"
3. Remove "Not Ready for Merge"
4. Merge this PR

#### To test

- That UI elements are more or less in the same position as per [Figma design](https://www.figma.com/file/yG8ieha0yESnr8wCtplr29/Post-Detail-View?node-id=8%3A477)
- Clicking blog section opens blog preview screen

|Post Details Header|
|-------------------|
|Screenshot 2020-10-01 at 10 59 59 AM|

**Note**
- Follow action refactoring and UI polishing will be done in separate PRs.
- ~Most of the changes are due to extraction and reuse of blog section from `ReaderPostUiState` (2c6fac5, 8dbc4f8). I can split this PR into smaller ones if anyone finds it difficult to review.~ - #13060,  #13061 split from this PR

PR submission checklist:

- [ ] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13071, opened 2020-10-07
Default to (Untitled) on the reader post card when blogName is empty

ashiagr:
Fixes blank site title issue reported in https://github.com/wordpress-mobile/WordPress-Android/pull/13042#pullrequestreview-502998546

#### To test

- Go to Reader
- Find a post with empty site title
- Verify that header positioning is not broken and blog name defaults to (Untitled)

image

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderCardUiState.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderCardUiState.kt
@@ -20,7 +20,7 @@ sealed class ReaderCardUiState {
         val blogId: Long,
         val dateLine: String,
         val title: UiString?,
-        val blogName: String?,
+        val blogName: UiString?,
         val excerpt: String?, // mTxtText
         val blogUrl: String?,
         val tagItems: List<TagUiState>,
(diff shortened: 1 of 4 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderCardUiState.kt
```diff
@@ -20,7 +20,7 @@ sealed class ReaderCardUiState {
         val blogId: Long,
         val dateLine: String,
         val title: UiString?,
-        val blogName: String?,
+        val blogName: UiString?,
```
malinajirka: I'm wondering if we could make this field non-nullable now, wdyt?

malinajirka (approved):
LGTM, thanks! I've left one minor comment - I'm not sure about it.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderCardUiState.kt
ashiagr: You're right, fixed in b40758d

PR #13075, opened 2020-10-07
Media picker tracking

### LINKED ISSUE #12646: Define and implement tracking mechanism for the media picker
The task is to:
- define all the events we want to track in the consolidated media picker.
- make sure we're tracking comparable events both in the old and the updated media picker so we can compare results with an A/B test. 
- implement the updated events

planarvoid:
Fixes #12646

This PR contains few fixes:
- it adds tracking to all user interactions with the media picker - all the clicks and the screen view. 
- I've consolidated all the navigation events into one object. I think it makes the code less cluttered. 
- I've added preview to stock library and WP media library options (it was previously missing)

To test:
- Check that the preview works in all 3 data sources
- Go through the flows and check that all user interactions are tracked (using Logcat and the "Tracked" filter)
- As part of ^ check that all the navigation actions work (preview, edit, insert media, click on icons, exit the screen)

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerFragment.kt
```diff
@@ -226,38 +230,42 @@ class MediaPickerFragment : Fragment() {
...
+                        ActivityLauncher.openImageEditor(activity, inputData)
+                    }
+                    is InsertMedia -> listener?.onItemsChosen(navigationEvent.identifiers)
+                    is IconClickEvent -> listener?.onIconClicked(navigationEvent.action)
+                    Exit -> {
+                        val activity = requireActivity()
+                        activity.setResult(Activity.RESULT_CANCELED)
+                        activity.finish()
+                    }
+                }
```
develric, line 268: ❤️

develric (commented):
Hi @planarvoid 👋 ! Checked this PR and code looks good. Also the track events seems to be triggered consistently for me 👍 ! I had the following issue I documented in #13132 when trying to edit/preview a selected item. Doesn't seems to be linked to this PR tbh but I'm investigating it better first and hopefully creating a fix PR before to merge this one so we can properly check the edit/preview paths as well. Hope it's fine for you 🙇

planarvoid:
hey @develric , this should be ready for another check

develric (approved):
Hi @planarvoid did a second pass here after merging the not directly related [fix](https://github.com/wordpress-mobile/WordPress-Android/pull/13138) as per note [here](https://github.com/wordpress-mobile/WordPress-Android/pull/13075#pullrequestreview-508300651) and was able to final check the preview and edit path also. LGTM  :shipit:

PR #13076, opened 2020-10-07
Fix/2692 unsupported block editor cannot save changes

marecar3:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/2692

To test:

1. On the web, add any block not yet supported in Gutenberg Mobile (i.e. Business Hours, Contact Form, YouTube Embed).
2. Save or publish changes.
3. On the app, open the post created in step 1.
4. Tap the unsupported block.
5. Tap the unsupported block again.
6. Tap the "Edit using web editor" option.
7. Try making any change to the block.
8. Tap the checkmark button at the top right.
9. You should return to Blog Post
10. Open Unsupported block again on web or mobile
11. You should see the latest changes

PR submission checklist:

### DIFF
--- a/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
+++ b/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
@@ -154,6 +154,7 @@ public class GutenbergEditorFragment extends EditorFragmentAbstract implements
 
         if (getGutenbergContainerFragment() == null) {
             GutenbergPropsBuilder gutenbergPropsBuilder = getArguments().getParcelable(ARG_GUTENBERG_PROPS_BUILDER);
+            mLatestGutenbergPropsBuilder = gutenbergPropsBuilder;
 
             FragmentManager fragmentManager = getChildFragmentManager();
             FragmentTransaction fragmentTransaction = fragmentManager.beginTransaction();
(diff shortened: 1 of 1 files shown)


designsimply (approved):
I tested with WPAndroid pr-13076-build-75553 and I am able to save changes to an unsupported block without seeing a crash in the editor now. 🎉

Testing notes: I tested two different WordPress.com Premium sites logged in as different users (made4testing0318.blog logged in as testuser0313 and madefortesting062420.wordpress.com logged in as madefortesting062420) and the blocks I tested changing were the YouTube embed block, the audio block,  and the Business Hours block. I used Pixel 3 Android 11 for testing.

mchowning (approved):
Looks good and works well in my tests. Nice job @marecar3 ! :shipit: 

It might be worth marking the mCurrentGutenbergProps field as `@Nullable` to give a hint that we need to be careful when using it. In practice, it really shouldn't ever be null for us going forward, so I'm not sure whether that would be all that helpful (of course we thought it wouldn't ever be null before now as well). Just thinking out loud, I don't think that change needs to go in this PR, but I think it's worth keeping in mind and considering in the future.

marecar3:
> Looks good and works well in my tests. Nice job @marecar3 ! :shipit:
> 
> It might be worth marking the mCurrentGutenbergProps field as `@Nullable` to give a hint that we need to be careful when using it. In practice, it really shouldn't ever be null for us going forward, so I'm not sure whether that would be all that helpful (of course we thought it wouldn't ever be null before now as well). Just thinking out loud, I don't think that change needs to go in this PR, but I think it's worth keeping in mind and considering in the future.

Good thinking from your side, I like it, thanks for thinking loud!

PR #13078, opened 2020-10-07
Prologue: Background update

renanferrari:
This PR updates the Prologue screen background to match the design spec (internal ref: pbArwn-193-p2). 

Here's how it looks like:

Note: this PR doesn't make the status bar transparent as defined by the design spec. This work will be done separately later on.

The goal here was to make the background look like a single element that gives us an effect of continuity when swiping through the Prologue pages, rather than five separate backgrounds for each page. To achieve that, I first added a parallax effect to the `ViewPager` to make the foreground elements move a bit faster than the background when swiping pages. As of right now, this is the configuration I went with:

- Background: Normal speed
- Title: 1.25x normal speed
- Content: 1.5x normal speed

As for the background itself, I decided to have a single circle image and then size, tint and position multiple copies of it individually for each page. I also experimented with having a single wide image with all the background circles pre-positioned in it and then make it scroll along with the pages, and while that approach was easier to implement, it didn't look good on different screen sizes, as the circles were positioned independently of the available screen space.

To position the circles in each page, I decided to use a `ConstraintLayout` and have two guidelines for each circle to indicate where they would be positioned, both vertically and horizontally. A [guideline](https://developer.android.com/reference/androidx/constraintlayout/widget/Guideline) needs to be bound to one of edges of the layout: vertical guidelines are bound to either the left or the right side, while horizontal guidelines are bound to either the top or the bottom. We then define a fixed distance in `dp` to indicate how far from that edge the guideline is going to be positioned. Finally, we specify how we want a certain circle drawable to be positioned in relation to that guideline. 

The example below demonstrates a horizontal guideline (in blue) bound to the top of the layout (as indicated by its arrow icon pointing up) and distant from it by `60dp`. The bottom of the top-left circle is positioned using this guideline.

Since we're using fixed distance values, the circles will always be positioned at the exact same place, independently of the screen size:

|Small|Medium|Large|Wide|Tablet|
|-----|-----|-----|-----|-----|
||||||

Note: please, disregard the page content as it has not been optimized for different screen sizes yet.

This is a bit different from what we have on the design spec, where the circles seem to be more visible when there's more screen space available. One way of addressing this would be to make the guidelines use percentages of the width or the height of a layout instead of a fixed distance value. I did experiment with this a bit but the results seemed to be more unpredictable, but I'm totally open to revisit this approach again if necessary.

## To test

1. Open the app.
1. Log out if necessary. 
1. On the Prologue screen, swipe between the pages and confirm that the background looks and behaves as expected.

Make sure to test on devices with different screen sizes/orientations and Android versions.

---

PR submission checklist:


damemm56:
@glung

damemm56 (approved):

damemm56 (changes requested):
@damemm56 ### [``](

- [ ] url
[``](url)

)

osullivanchris:
It looks brilliant. Great job on making this work, and thanks for explaining how you put it together.

> This is a bit different from what we have on the design spec, where the circles seem to be more visible when there's more screen space available. One way of addressing this would be to make the guidelines use percentages of the width or the height of a layout instead of a fixed distance value. I did experiment with this a bit but the results seemed to be more unpredictable, but I'm totally open to revisit this approach again if necessary.

I think, judging by the screenshots above, its probably fine the way you have done it. 

> To achieve that, I first added a parallax effect to the ViewPager to make the foreground elements move a bit faster than the background when swiping pages. As of right now, this is the configuration I went with:

The background looks great. I'm not 100% sure on the speed and direction of the title and image. Have you seen the animation I made in i3 (pbArwn-15E-p2)? There, even though the screens are sliding across, I moved elements _up_ into frame. Of course, going sideways is pretty natural too, and more an extension of what the user is doing. What I did might feel too much like scroll-jacking and going against the users manipulation.

If sticking with what you've done I'd love to see it with a couple of other values just to compare. How would it behave
- if title and content were both at 1.5x speed
- if title and content were 2x speed
- if title and content were both at normal speed, but the circles moved _slower_

I am unsure by the way if 2x means slower (twice as long) or faster (twice as quick). Can you clarify so I know we are speaking the same language? 😄

osullivanchris (commented):
forgot to use the 'add review' button

frosty:
@renanferrari and @khaykov, this is looking amaaaazing 😍

renanferrari:
Hey @osullivanchris, thanks a lot for reviewing this.

>Have you seen the animation I made in i3 (pbArwn-15E-p2)?

Oh my, I had totally seen that animation before, but actually completely forgot about it when implementing this 🤦‍♂️ Sorry! I can definitely build another version for you to try it out and see what works better, it shouldn't be hard to do – although I believe it would be better to do so in a separate PR.

>I am unsure by the way if 2x means slower (twice as long) or faster (twice as quick). Can you clarify so I know we are speaking the same language? 😄

Absolutely! "2x normal speed" would mean twice as fast. I'm going to try it out with the values you suggested and post an update here with the results.

osullivanchris:
> Oh my, I had totally seen that animation before, but actually completely forgot about it when implementing this 🤦‍♂️ Sorry! I can definitely build another version for you to try it out and see what works better, it shouldn't be hard to do – although I believe it would be better to do so in a separate PR.

That's no problem. You don't have to do it immediately. Just wanted you to see the ideas 😃 

> Absolutely! "2x normal speed" would mean twice as fast. I'm going to try it out with the values you suggested and post an update here with the results.

ah ok...I think I was thinking twice as long as in slower haha. Let me write in a better way...some things I'd consider.

- title and content coming in slower, but at the same speed
- title and content coming in extra slow delay 
- title and content coming in at regular speed, but the background panning slower

I can also play around with it in one of my prototyping tools tomorrow or next week if its very intensive to play around with the code.

renanferrari:
@osullivanchris and I chatted a bit via Slack about the parallax effect and we realized that referring to the foreground elements speed using terms like "1.25x normal speed" can a be quite ambiguous: while I had intended for these values to represent the speed of the views when _moving out_ of the screen, Chris thought they represented the speed of the views when _moving into_ the screen. 

This can be very confusing, since the logic we're currently using for the parallax effect _increases_ the speed of views that are _exiting_ the screen and _decreases_ the speed of views that are _entering_ the screen by the same factor:

|View speeds when exiting the screen|View speeds when entering the screen|
|-----|-----|
|||

So basically, Chris and I were referring to the same goal when we said "twice as fast" or "twice as long", even though we were not referring to the exact same thing.

I have updated [the parallax logic and its documentation](https://github.com/wordpress-mobile/WordPress-Android/blob/4c76f7d8d9e864ebaa8172809ce33ba97a01766b/WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginPrologueFragment.kt#L85-L117) to hopefully prevent similar misunderstandings from happening in the future.

khaykov (approved):
Looks beautiful! Thank you so much for implementing this 🙇‍♂️

osullivanchris:
> Looks beautiful! Thank you so much for implementing this 🙇‍♂️

+1 to this, its lovely!

PR #13081, opened 2020-10-08
Update debug build crash message

malinajirka:
Updates crash message.

To test:
No need to test anything.
PR submission checklist:


malinajirka:
We differentiate between the expected scenario (*) and unexpected scenario (**) [here](https://github.com/wordpress-mobile/WordPress-Android/blob/2ea185764149d988ba2ed752716199480a6aed6f/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L1661). So if the messages says [this](https://github.com/wordpress-mobile/WordPress-Android/blob/9b7448853cd773e4d7c71d459787afa41e6120b0/WordPress/src/main/java/org/wordpress/android/ui/posts/SaveOnExitException.kt#L24), it's not expected and we should investigate why gutenberg didn't invoke autosave mechanism.

`*` User makes changes in the last 500ms and there is a pending autosave
`**` User make changes and there isn't any pending autosave

PR #13090, opened 2020-10-08
Support multiple media sources

### LINKED ISSUE #13089: Support multiple media sources in media picker
It is currently possible to switch media sources in the media picker while selecting featured image and the site icon. The featured image can be selected from the device, the WP library, the gif library and the stock library, the site icon can be selected from the device and the WP library. 

We need to implement the same functionality in the consolidated media picker.

planarvoid:
Fixes #13089

This PR adds the support for multiple data sources. It is currently used in the featured image selection and in the site icon selection (I don't know if it's necessary somewhere else, I couldn't find any other place where it could be used). I think we'll replace this solution by tabs soon so this should be temporary.

I've added a WP media library as an optional icon, it's visible if it fits the screen. The stock media and gif selection are always collapsed in the menu.

Tenor selection is not yet enabled because it's not yet implemented.

Screenshot 2020-10-08 at 16 30 30
Screenshot 2020-10-08 at 16 30 37
Screenshot 2020-10-08 at 16 30 55

To test:
- Turn on the Consolidated media picker flag
- Go to a post editor/settings/set featured image
- Notice that there is now the extra menu in the toolbar
- Notice that it contains 3 options to select from WP Media library, stock library and gif library
- Click on WP Media library
- Notice a new activity is opened 
- Notice the extra menu is now gone
- Select an image and confirm it
- The featured image gets selected

To test:
- Turn on the Consolidated media picker flag
- Go to a post editor/settings/set featured image
- Notice that there is now the extra menu in the toolbar
- Notice that it contains 3 options to select from WP Media library, stock library and gif library
- Click on the Stock library
- Notice a new activity is opened 
- Notice the search is now enabled as default
- Search for an image
- Select an image and confirm it
- The featured image gets selected

To test:
- Turn on the Consolidated media picker flag
- Go to my site/click on site icon/change icon
- Notice that there is now the WP media option available
- Select a site icon
- Check that it's now selected
 

PR submission checklist:


planarvoid:
I've added support for tenor as an additional data source in 39c70c9ddf67974af1edc0eaf276376d6ff96fdc

This can be tested by:
- Go to Editor/Settings/Choose featured image
- From the overflow menu select "Choose from Tenor"
- Search for a gif and select it
- Notice it gets set as the post featured image

develric (commented):
Hi @planarvoid 👋 ; went through the test scenarios and I had the following issues (using Pixel 3 Emu API 29); not sure if I'm doing anything wrong, let me know if you cannot replicate 🙇 :

- I get the WP Stories FAB visible while adding feature image for a regular blog post in all the picker sources (AFAIU it should not be present for a regular post, is it correct?). The FAB is present for me also when picking for setting a site icon

| Main picker | Switched to WP lib |
|---|---|
| image | image |

- Site icon selection fails for me with the following exception  
`Unexpected error - Site icon upload failed, because there wasn't any site selected.`
when selecting the site icon from WP lib source the icon is not updated but I do not get the exception.

- When selecting feature image after switching source (it happened for every sources for me) the 
featured image is not updated but apparently I do not get relevant exceptions in logcat

planarvoid:

> I get the WP Stories FAB visible while adding feature image for a regular blog post in all the picker sources (AFAIU it should not be present for a regular post, is it correct?). The FAB is present for me also when picking for setting a site icon

This is on purpose - it's the FAB to take a photo. It's not doing anything yet though :-). The functionality is in the next PR. I've changed the behaviour so the FAB is now hidden after switching sources in bf4c3cf

> Site icon selection fails for me with the following exception
> Unexpected error - Site icon upload failed, because there wasn't any site selected.
> when selecting the site icon from WP lib source the icon is not updated but I do not get the exception.

> When selecting feature image after switching source (it happened for every sources for me) the
featured image is not updated but apparently I do not get relevant exceptions in logcat

As discusses, these two are caused by the "Don't keep activities" settings. Try turning it off and let me know if it works.

develric (approved):
Hey @planarvoid 👋 .

> This is on purpose - it's the FAB to take a photo. 

Was not aware of this, thanks for sharing 🙇 

> As discusses, these two are caused by the "Don't keep activities" settings. 

Yep 👍 , as per slack discussion I confirm I was using an Emu with `don't keep activities` setting active, that I needed to check a few edge cases while developing code for another PR. Tested scenarios again without that setting and all LGTM. As a side note, probably still not super we are not resilient to that (even if bit extreme) scenario, but has totally nothing to do with this PR or the MP (I think also some editor behaviours could be affected with that setting on); maybe something to review at a general level if it makes sense. 

So safe to go here 😄  . Thanks @planarvoid 🙇  . :shipit:

PR #13094, opened 2020-10-09
Feature/prologue refresh

khaykov:
This PR introduces a new login prologue.

It might seem really big, but most of it just layout changes, which you can confirm just by running an app.

Portrait:
[Image from Gyazo](https://gyazo.com/df09dcfec998cb3a17046e21c8c3b184)

Landscape:
[Image from Gyazo](https://gyazo.com/67af944cc399c6d79abf72c15ccc5285)

Tablet:
[Image from Gyazo](https://gyazo.com/763cc9cb847c81aec5d5341276c70082)

To make the content of each slide look right on each screen size we calculate a scaling factor and scale the layout accordingly.

The transparent status bar is supported starting from API 23.

To test:
- Clear the app data and install the app from this PR.
- Check that the login prologue is visually solid.
- Navigate into login flow and make sure the status bar looks ok in both Light and Dark modes and in Landscape/Portrait orientations.
- Perform the same checks on a device with API 21.

PR submission checklist:


khaykov:
@renanferrari Not too critical, but it looks like the page transformer makes the shadows on the cards behave a little bit funny.

[Image from Gyazo](https://gyazo.com/a9fffdcbe8dcf709a599ade55facb241)

I found a [couple](https://stackoverflow.com/questions/39068133/pagetransformer-has-snapping-effect-when-animating-elevation) of [mentions](https://stackoverflow.com/questions/41763863/shadows-created-with-elevation-flicker-when-changing-fragments-in-a-viewpager) of these issues, but no solution so far.

osullivanchris (changes requested):
Hey its looking really great!

A couple of small issues

- I still think the heading needs to move down approx 24dp. 
- Not sure, but maybe the image could move up 16dp. Let me know on those whether its gonna break something that I'm not considering
- There is a flickering on the shadows on swipe which I know I mentioned to @renanferrari . I don't think we can release it with the animation unless the shadows are resolved TBH! 😬 
- The new user avatars should be used on screen 3. Currently, they are on screen 5. Screen 5 has different assets. Let me know if you need clarification

cc @mattmiklic with all your work on login it would be good to get your eyes on this as it gets closer to release.

renanferrari:
>it looks like the page transformer makes the shadows on the cards behave a little bit funny.

>There is a flickering on the shadows on swipe which I know I mentioned to @renanferrari . I don't think we can release it with the animation unless the shadows are resolved TBH! 😬

@osullivanchris did mention this to me, but I didn't realize it was related to `PageTransformer` at the time. So thank you @khaykov for pointing that out. I'm gonna use the links you posted above and go from there to investigate this issue.

khaykov:
@osullivanchris 

> I still think the heading needs to move down approx 24dp.

Right, I just realized that because of the clear status bar we need that extra 24dp :)

> Not sure, but maybe the image could move up 16dp. Let me know on those whether its gonna break something that I'm not considering

No problem here! Just wanted to make sure if it's 16dp only or you want to account for the fact that we pushed the image down with the status bar and make it 40dp (16+24). Hope this makes sense :)

On this screenshot, I pushed it up by 40dp to account for the extra 24dp at the top.
(before-after)
[Image from Gyazo](https://gyazo.com/0263cf57345fede2e264faf5007f907f)

> The new user avatars should be used on screen 3. Currently, they are on screen 5. Screen 5 has different assets. Let me know if you need clarification

Good catch! Fixed it 👍

renanferrari:
It seems that using `ViewPager2` managed to solve the shadow issue. @khaykov Thanks a lot for taking care of this!

osullivanchris:
> Right, I just realized that because of the clear status bar we need that extra 24dp :)

Sweet! Clear status bar looks good too.

> On this screenshot, I pushed it up by 40dp to account for the extra 24dp at the top.

It looks good now! 😄 

> Good catch! Fixed it 👍

thanks!

osullivanchris (approved):

mattmiklic:
Thanks for the ping @osullivanchris. This is looking really nice! Seeing it on a device brings out some of the subtle details I hadn't noticed previously. 

On my Pixel 3 XL the headlines felt a little too close to the top edge of the screen. Mostly because of this device's huge notch and the fact that there's more space below the headline than above it. This may look fine on a more standard shaped device. In landscape mode the headline felt too close to the left edge.

I think there may be a typo: the last screen says "discover new _ads_" instead of reads. 😄

Screenshots: [Portrait](https://user-images.githubusercontent.com/1396552/96310978-7711b800-0fd6-11eb-872e-a8085a2d67c7.png), [Landscape](https://user-images.githubusercontent.com/1396552/96310980-77aa4e80-0fd6-11eb-9611-268d9bf52c93.png)

mattmiklic:
A couple of things I noticed while using TalkBack:

- It speaks the screen title "Login", would probably be better without it if possible. Could it speak "WordPress" instead?
- There isn't a way to navigate between screens. Focus goes from the headline to the Continue with WordPress and Enter your Site Address buttons.
- On the illustrations that include text, the text ("Madison Ruiz liked your post", etc.) is read out loud when moving through the screen content. Can we provide an alternate text description instead?

khaykov:
@mattmiklic 

> discover new ads

Chris mentioned this one before, but I totally forgot about it 🤦 

> On the illustrations that include text, the text ("Madison Ruiz liked your post", etc.) is read out loud when moving through the screen content. Can we provide an alternate text description instead?

Should we just announce the header text for the whole slide (eg. Follow your favorite sites and discover new reads.)? I think this will make more sense.

> There isn't a way to navigate between screens. Focus goes from the headline to the Continue with WordPress and Enter your Site Address buttons.

It actually works as intended - when the multipage element is selected, you navigate to the next page by swiping with two fingers.

> It speaks the screen title "Login", would probably be better without it if possible. Could it speak "WordPress" instead?

I'll see if we can do it 👍

mattmiklic:
> Should we just announce the header text for the whole slide (eg. Follow your favorite sites and discover new reads.)? I think this will make more sense.

It already does, so maybe that's all we need. After swiping past the header text it just currently reads out all the illustration text before getting to the buttons.

> It actually works as intended - when the multipage element is selected, you navigate to the next page by swiping with two fingers.

Oh perfect, it does! Should there be an audible indicator that you're on a multipage element? Or maybe that's what some of the sound effects indicate?

libs/login/WordPressLoginFlow/src/main/res/values-v23/dimens.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<resources>
+
+    <dimen name="status_bar_height_offset">24dp</dimen>
```
renanferrari: Since we're assuming the status bar is always going to have a `24dp` height, I believe this is causing the issue @mattmiklic mentioned with the Pixel 3XL notch:

Captura de Tela 2020-10-16 às 19 54 55

Another thing, if I'm not mistaken this is supposed to be used with the new `Theme.LoginFlow.TransparentStatusBar`, right? If so, since this change is being made within the Login Library, we also need to preserve the behavior for when this theme is not being used, otherwise it might cause some unexpected results in the Woo app, for example.

renanferrari (commented):
I tested this quite a bit and only found the same issue that was reported by @mattmiklic above. I left a comment below with my thoughts about it. 

Other than that, everything else seems to be working as expected and the code is looking pretty good 👍

renanferrari:
I'm going to be AFK next week, so I'm removing my assignment from this PR in case anyone wants to do a second round of review before I come back. If not, I'm happy to do it myself the week after the next one :)

osullivanchris:
> On my Pixel 3 XL the headlines felt a little too close to the top edge of the screen. Mostly because of this device's huge notch and the fact that there's more space below the headline than above it. This may look fine on a more standard shaped device. In landscape mode the headline felt too close to the left edge.

Thanks for the feedback @mattmiklic . I've been trying to tweak this as well which you can see in [these images](https://github.com/wordpress-mobile/WordPress-Android/pull/13094#issuecomment-708064858). I'm actually trying to figure out if your reference screenshot is from the before or after before I get @khaykov to change it again! It looks like yours might be from before the fix...but I'm not sure as they're different device sizes and different carousel screens.

khaykov:
@renanferrari Sorry for the delay, I was also mostly AFK last week. The notch is a bit tricky, and the only solution to it that I found results in content moving below it when we transition from prologue to login screen. I tried to get rid of this transition, but no luck. I confirmed with @osullivanchris that it's ok.

[Image from Gyazo](https://gyazo.com/c49d978431f7e938fb55543db3651792)

A couple of other fixes:
- Fixed the typo in one of the slide titles.
- Added start margin to the slide title in the landscape.
- TalkBack: Announcing screen as WordPress instead of Login.
- TalkBack: the content of slides is not longer focusable with talkback.

> Another thing, if I'm not mistaken this is supposed to be used with the new Theme.LoginFlow.TransparentStatusBar, right? If so, since this change is being made within the Login Library, we also need to preserve the behavior for when this theme is not being used, otherwise, it might cause some unexpected results in the Woo app, for example.

I removed the status bar offset and moved the full-screen related logic to the login prologue fragment, so if I understand this correctly this should not have any effect on different clients (They can implement Theme.LoginFlow.TransparentStatusBar which will do exactly what the name implies without any extra effects).

renanferrari (approved):
Thanks for the changes @khaykov, everything seems to be working pretty well now. Great job with this!

LGTM :shipit:

PR #13095, opened 2020-10-09
Check if site is not null in onSiteChanged in SiteSettingsFragment

### LINKED ISSUE #13045: NullPointerException: Attempt to invoke virtual method 'int org.wordpress.android.fluxc.model.SiteModel.getId()' on a n...
**Sentry Url**: https://sentry.io/share/issue/7cacd4f401ea4058bf8b971fa13399eb/
**User Count**: 760
**Count**: 941
**First Release**: org.wordpress.android@15.1-rc-1+881
**First Seen**: 2020-07-03T09:14:52Z
**Last Seen**: 2020-10-01T07:48:08Z
**24 Hours**: 7
**30 Days**: 282

```
NullPointerException: Attempt to invoke virtual method 'int org.wordpress.android.fluxc.model.SiteModel.getId()' on a null object reference
     at org.wordpress.android.ui.prefs.SiteSettingsFragment.onSiteChanged(SiteSettingsFragment.java:1976)
...
(1 additional frame(s) were not displayed)
```

planarvoid:
Fixes #13045 

The site shouldn't be null if the fragment is properly initialized. However, it seems that this doesn't always happen. I'm following the approach that's used in the rest of the fragment where we always check whether the site is present when we want to access it.

To test:
- I don't know how to reproduce this but the fix should be clear enough

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/prefs/SiteSettingsFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/prefs/SiteSettingsFragment.java
@@ -1966,7 +1966,7 @@ public class SiteSettingsFragment extends PreferenceFragment
 
     @Subscribe(threadMode = ThreadMode.MAIN)
     public void onSiteChanged(OnSiteChanged event) {
-        if (!event.isError()) {
+        if (!event.isError() && mSite != null) {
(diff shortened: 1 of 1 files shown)


khaykov (approved):
Looks good 👍 And hope it works 🤞

PR #13101, opened 2020-10-10
Mobile stories block - part 3: (rename refactor)

mzorz:
This PR contains needed work to rename component to `BlockMediaUpdateProgress`

Related PRs:

Gutenberg: https://github.com/WordPress/gutenberg/pull/26005

Jetpack: https://github.com/Automattic/jetpack/pull/17456

Gutenberg mobile: https://github.com/wordpress-mobile/gutenberg-mobile/pull/2706

To test:
- same as per https://github.com/wordpress-mobile/WordPress-Android/pull/12939

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13103, opened 2020-10-11
Mobile stories block - part 4: error handling

mzorz:
This PR builds on top of https://github.com/wordpress-mobile/WordPress-Android/pull/13101

This contains all the work needed in order to provide error situation handling when saving or uploading any of the media related to a Story block.

errorhandling2

To test:
CASE A: error in saves
1. apply this gist https://gist.github.com/mzorz/83b64b49d7a6160cc7beb7c3fd329d15 (this creates an exception the second time the service runs into an image slide to be saved (could be the same 1-slide story being saved more than once or a different slide on the same story)
2. run WPAndroid
3. create a story with 1 image slide and at least 1 view, it should be saved and published correctly
4. now go to the Posts list open the post on mobile gutenberg
5. select the story block, then tap again or tap on the edit icon button on the block to launch the Story composer
6. do edits like, move the added view around or add another view
7. tap on DONE, this should exit the composer and trigger the saving process
8. observe the process runs into an error and signals Gutenberg to show the failed overlay
9. tapping on the block should launch the story composer again, in error-wrangling mode
10. tap on RETRY to attempt a new save (this time it'll work)
11. now tap on DONE
12. observe the block gets updated on mobile gutenberg
13. if you want to see the changes reflected on your site, don't forget to tap on the UPDATE button on the editor as well.

CASE B: error in upload.
1. create a story with a few slides (3-4 so it takes enough time)
2. once its uploaded, edit it
3. tap on the story block and launch the story composer, now edit some views and tap DONE
4. the story will start uploading
5. turn airplane mode ON while uploads are still happening
6. observe the failed upload error overlay is shown
7. turn airplane mode oFF
8. tap on the story block failed overlay and observe the dialog appears "Couldn't upload file"
9. tap RETRY
10. observe the uploads resume and everything ends up correctly

Related PRs:

Gutenberg mobile: https://github.com/wordpress-mobile/gutenberg-mobile/pull/2708
Gutenberg: https://github.com/WordPress/gutenberg/pull/26008
Jetpack: https://github.com/Automattic/jetpack/pull/17458

PR submission checklist:


aforcier:
I didn't dig into the implementation, but I gave the flow a try and confirmed that both test cases work as advertised for me 👍 Nice work @mzorz !

loremattei:
I'm moving this to 16.1 because 16.0 has been cut. Please feel free to ping me if you want this to make it to 16.0.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -3308,6 +3295,69 @@ public void onTrackableEvent(TrackableEvent event, Map<String, String> propertie
         }
     }
 
+    @Override public void onRetryUploadForMediaCollection(ArrayList<Object> mediaFiles) {
```
jd-alexander, line 3268: Hi @mzorz could these methods be moved to another class for testability and line reduction reasons? LMK 👍

mzorz: 👍 addressed this request in this PR https://github.com/wordpress-mobile/WordPress-Android/pull/13198 targeting this one here (so that one should be reviewed and merged first, then you could continue reviewing this one properly here)

jd-alexander: Nice. This is such a great improvement 👍

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
```diff
@@ -510,39 +512,46 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
...
-                                            mediaFile = it,
-                                            temporaryId = "$storyIndex-$frameIndex"
-                                    )
-                            frame.id = storyMediaFileData.id
+                        if (it.startsWith(TEMPORARY_ID_PREFIX)) {
+                            val storyMediaFileData = buildStoryMediaFileDataForTemporarySlide(
+                                    frame,
+                                    it
+                            )
+                            // frame.id = storyMediaFileData.id
```
jd-alexander: hi @mzorz do we want to remove this line or is it going to be utilized in the future?

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
```diff
@@ -510,39 +512,46 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
...
                             storyMediaFileDataList.add(storyMediaFileData)
+                        } else {
+                            val mediaModel = mediaStore.getSiteMediaWithId(site, it.toLong())
+                            val mediaFile = fluxCUtilsWrapper.mediaFileFromMediaModel(mediaModel)
+                            mediaFile?.let {
+                                val storyMediaFileData =
+                                        saveStoryGutenbergBlockUseCase.buildMediaFileDataWithTemporaryId(
+                                                mediaFile = it,
```
jd-alexander: could we give the parameter `it` a name so that the outer one isn't being shadowed? 🙏

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
```diff
@@ -510,39 +512,46 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
...
+                            val storyMediaFileData = buildStoryMediaFileDataForTemporarySlide(
+                                    frame,
+                                    it
+                            )
+                            // frame.id = storyMediaFileData.id
                             storyMediaFileDataList.add(storyMediaFileData)
+                        } else {
+                            val mediaModel = mediaStore.getSiteMediaWithId(site, it.toLong())
+                            val mediaFile = fluxCUtilsWrapper.mediaFileFromMediaModel(mediaModel)
+                            mediaFile?.let {
```
jd-alexander: Just a quick question here. If this `mediaFile` is null, what should happen here or is it a case where it's highly unlikely that will happen so we don't have to handle that case in any way?

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
```diff
@@ -551,6 +560,20 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
         return storyMediaFileDataList
     }
 
+    private fun buildStoryMediaFileDataForTemporarySlide(frame: StoryFrameItem, tempId: String): StoryMediaFileData {
+        val storyMediaFileData =
```
jd-alexander: This variable can be eliminated and the logic below inlined since it's only used in the return :)

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
mzorz: nice catch! removed in 5e55e168f6

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
```diff
@@ -127,7 +206,7 @@ class StoriesPrefs @Inject constructor(
         }
     }
 
-    fun replaceLocalMediaIdKeyedSlideWithRemoteMediaIdKeyedSlide(
+    fun replaceLocalMediaIdKeyedSlideWithRemoteMediaIdKeyedSlide_Phase2(
```
jd-alexander: could we remove the underscore from this method name and the one below since function names shouldn't contain them? Also could you expound a bit on why we need the `phase1` and `phase2` in the names? Is there anyway we can make it clearer? Probably some comments here? LMK

WordPress/src/main/java/org/wordpress/android/ui/stories/media/StoryMediaSaveUploadBridge.kt
```diff
@@ -211,23 +221,23 @@ class StoryMediaSaveUploadBridge @Inject constructor(
...
+                    event.storyIndex,
+                    story.frames
+            )
+
+            // only trigger the bridge preparation and the UploadService if the Story is now complete
+            // otherwise we can be receiving successful retry events for individual frames we shouldn't care about just
+            // yet.
+            if (isStorySavingComplete(event) && !event.isRetry) {
+                // only remove it if it was successful - we want to keep it and show a snackbar once when the user
+                // comes back to the app if it wasn't, see MySiteFrament for details.
```
jd-alexander:
```suggestion
                // comes back to the app if it wasn't, see MySiteFragment for details.
```

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
mzorz: 👍 done in 4d1eaa16d6

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
```diff
@@ -17,7 +18,9 @@ class StoriesPrefs @Inject constructor(
     private val context: Context
 ) {
     companion object {
+        private val KEY_STORIES_SLIDE_INCREMENTAL_ID = "incremental_id"
         private val KEY_PREFIX_STORIES_SLIDE_ID = "story_slide_id-"
+        private val KEY_PREFIX_TEMP_MEDIA_ID = "t-"
         private val KEY_PREFIX_LOCAL_MEDIA_ID = "l-"
         private val KEY_PREFIX_REMOTE_MEDIA_ID = "r-"
```
jd-alexander: These values could be marked as constants :) using `const`

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
```diff
@@ -32,13 +35,60 @@ class StoriesPrefs @Inject constructor(
...
+    fun getNewIncrementalTempId(): Long {
+        var currentIncrementalId = getIncrementalTempId()
+        currentIncrementalId++
+        val editor = PreferenceManager.getDefaultSharedPreferences(context).edit()
+        editor.putLong(KEY_STORIES_SLIDE_INCREMENTAL_ID, currentIncrementalId)
+        editor.commit()
+        return currentIncrementalId
+    }
+
+    fun getIncrementalTempId(): Long {
```
jd-alexander:
```suggestion
    private fun getIncrementalTempId(): Long {
```

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
```diff
@@ -32,13 +35,60 @@ class StoriesPrefs @Inject constructor(
                 KEY_PREFIX_LOCAL_MEDIA_ID + mediaId.value.toString()
     }
 
+    private fun buildSlideKey(siteId: Long, tempId: TempId): String {
+        return KEY_PREFIX_STORIES_SLIDE_ID + siteId.toString() + "-" +
+                KEY_PREFIX_TEMP_MEDIA_ID + tempId.id
+    }
+
+    @SuppressLint("ApplySharedPref")
```
jd-alexander, line 43: Could you expound a bit on why you suppressed using apply here Is it because apply does it in the background and you want this to be done immediately?

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
```diff
@@ -32,13 +35,60 @@ class StoriesPrefs @Inject constructor(
...
+                        0
+                )
+    }
+
     fun checkSlideIdExists(siteId: Long, mediaId: RemoteId): Boolean {
         val slideIdKey = buildSlideKey(siteId, mediaId)
         return PreferenceManager.getDefaultSharedPreferences(context).contains(slideIdKey)
     }
 
+    fun checkSlideIdExists(siteId: Long, tempId: TempId): Boolean {
```
jd-alexander:
```suggestion
    private fun checkSlideIdExists(siteId: Long, tempId: TempId): Boolean {
```

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
mzorz: right - as you said, it is highly unlikely for this to happen, given the frame has an id it means it has previously been assigned from a MediaModel, and as such getting a mediaFile from it is trivial and guaranteed to succeed. Given this is only called when the Save button is pressed when editing, the mediaModel exists (and things are recreated from there at this point).

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
```diff
@@ -32,13 +35,60 @@ class StoriesPrefs @Inject constructor(
...
+        val slideIdKey = buildSlideKey(siteId, tempId)
+        return PreferenceManager.getDefaultSharedPreferences(context).contains(slideIdKey)
+    }
+
+    fun checkSlideIdExists(siteId: Long, localId: LocalId): Boolean {
```
jd-alexander:
```suggestion
    private fun checkSlideIdExists(siteId: Long, localId: LocalId): Boolean {
```

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
mzorz: 👍 done in 6bb0e31cf3

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
```diff
@@ -32,13 +35,60 @@ class StoriesPrefs @Inject constructor(
...
+
+    fun checkSlideOriginalBackgroundMediaExists(siteId: Long, tempId: TempId): Boolean {
+        val storyFrameItem: StoryFrameItem? = getSlideWithTempId(siteId, tempId)
+        return checkSlideOriginalBackgroundMediaExists(storyFrameItem)
+    }
+
+    fun checkSlideOriginalBackgroundMediaExists(siteId: Long, localId: LocalId): Boolean {
+        val storyFrameItem: StoryFrameItem? = getSlideWithLocalId(siteId, localId)
+        return checkSlideOriginalBackgroundMediaExists(storyFrameItem)
+    }
```
jd-alexander: Nitpick comments here :) We could make these private and also when I utilize a sealed class for a type I normally make the variable name more descriptive since the type gives enough context. so I would go with `mediaId: LocalId` so it's easy to see that we are representing the local media id here because `localId` on it's own isn't as clear. Same thought process for the `checkSlideIdExists` above. Again, this is a Nitpick 😅

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
mzorz: Excellent idea, I didn't know how to enforce that enough so I thought calling the attention of the reader by making that underscore postfix name on purpose would make it "special enough" but ofc that goes with moving away from standards, which we don't want 👍 
Addressed as suggested in 9b80af491a "renamed _Phase1 and _Phase2 postfixed method names and added comments to illustrate how id assignment for media works in two phases"

jd-alexander (changes requested):
Thanks for these changes @mzorz I tested using your the guideline you described (including apply the gist 😉 ) and the behavior was as expected. I left a few some thoughts that you can address as you see but things are looking good here so once you have addressed those and responded we can ship this 🙏

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
mzorz: 👍 done in 695fed6d0b

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
jd-alexander: Yes, exactly so :) Thank you.

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
mzorz: done in 2bcc13e08d

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
jd-alexander: Those comments are beautifully done @mzorz . Everything is all clear now 🤝

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
mzorz: exactly :) - I need to block the call to make sure it's saved and then we can continue on.
Just in case of curiosity, I've [plans tracked here](https://github.com/Automattic/stories-android/issues/445) to move to [Jetpack Datastore](https://developer.android.com/topic/libraries/architecture/datastore) afterwards - aim is to finish this current implementation for now 👍

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
mzorz: Oh I see! the tricky part here is that the only actual `remoteId` is `mediaId` itself :D - as is called in MediaModel. But then the `id` can be either local or temporary, and as such I wanted to make super sure the user is passing a localId or temporary id as asked by the parameters.
That said, it really looks better when you use the same conceptual representation for `mediaId`, leaving the type enforce the context part (being the name `mediaId`, we now know what this is, and then which type we're being told by the type itself).

All this to say I agree 😅 - done in 798da18e17

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
jd-alexander: Okay, awesome. Thank you. Looking forward to seeing Jetpack datastore in action.

WordPress/src/main/java/org/wordpress/android/ui/stories/prefs/StoriesPrefs.kt
jd-alexander: haha okay. You always make a good argument so I see what you are saying as well but yes, lets your changes look good and I still get the same point based on the type in use 😄

jd-alexander:
I will have a look at test cases tomorrow @mzorz

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
```diff
@@ -0,0 +1,288 @@
...
+    private lateinit var siteModel: SiteModel
+    @Mock lateinit var storiesPrefs: StoriesPrefs
+    @Mock lateinit var context: Context
+    @Mock lateinit var postStore: PostStore
+
+    @InternalCoroutinesApi
+    @Before
+    fun setUp() {
+        saveStoryGutenbergBlockUseCase = SaveStoryGutenbergBlockUseCase(storiesPrefs)
+        siteModel = SiteModel()
```
jd-alexander: I think we can also mock this `siteModel` as well :) we can probably even remove it since it's just being used in one place.

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
```diff
@@ -0,0 +1,288 @@
...
+    fun `slides are saved locally to storiedPrefs`() {
+        // Given
+        val frames = ArrayList<StoryFrameItem>()
+        frames.add(getOneStoryFrameItem("1"))
+        frames.add(getOneStoryFrameItem("2"))
+        frames.add(getOneStoryFrameItem("3"))
+
+        // When
+        saveStoryGutenbergBlockUseCase.saveNewLocalFilesToStoriesPrefsTempSlides(
+                siteModel,
```
jd-alexander: So yes, with the comment above, I was saying we could just mock this to not distract the reader with an object that doesn't have any relevance to the test.
```suggestion
                mock(),
```

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
```diff
@@ -0,0 +1,288 @@
...
+                            caption = ""
+                    )
+                    mediaFiles.add(oneMediaFile)
+                }
+                return mediaFiles
+            }
+        }
+    }
+
+    companion object {
```
jd-alexander, line 227: Could the constants in this companion object use the uppercase naming convention. When I was looking at their usages in the tests I ended up looking for a variable within the test itself. Thanks :) 

`blockWithEmptyMediaFiles` -> `BLOCK_WITH_EMPTY_MEDIA_FILES`

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
```diff
@@ -0,0 +1,288 @@
...
+                mock(),
+                postStore,
+                mock(),
+                TEST_DISPATCHER,
+                TEST_DISPATCHER
+        )
+    }
+
+    @Test
+    fun `post with empty Story block is set given an empty mediaFiles array`() {
```
jd-alexander:
```suggestion
    fun `post with empty Story block is given an empty mediaFiles array`() {
```

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
```diff
@@ -0,0 +1,288 @@
...
+                editPostRepository,
+                mediaFiles
+        )
+
+        // Then
+        Assertions.assertThat(editPostRepository.content).isEqualTo(blockWithEmptyMediaFiles)
+    }
+
+    @Test
+    fun `post with non-empty Story block is set given a non-empty mediaFiles array`() {
```
jd-alexander, line 64:
```suggestion
    fun `post with non-empty Story block is given a non-empty mediaFiles array`() {
```

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
```diff
@@ -0,0 +1,288 @@
...
+        // Then
+        Assertions.assertThat(mediaFileData.alt).isEqualTo("")
+        Assertions.assertThat(mediaFileData.id).isEqualTo(TEMPORARY_ID_PREFIX + 1)
+        Assertions.assertThat(mediaFileData.link).isEqualTo(
+                "https://testsite.files.wordpress.com/2020/10/wp-0000000.jpg"
+        )
+        Assertions.assertThat(mediaFileData.type).isEqualTo("image")
+        Assertions.assertThat(mediaFileData.mime).isEqualTo(mediaFile.mimeType)
+        Assertions.assertThat(mediaFileData.caption).isEqualTo("")
+        Assertions.assertThat(mediaFileData.url).isEqualTo(mediaFile.fileURL)
```
jd-alexander: I'm not a big fan of multiple Assertions within one test method 😅 To note, this is learnt behavior from Jirka 😆 so yes, he taught me that a unit test should have one single reason to fail, hence the name `unit` 😉 

So the assertion I am expecting here based on the test name is that `mediaFileData` is not null and was returned from the use case. All the other assertions would be separate tests as they are testing different things.

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
```diff
@@ -0,0 +1,288 @@
...
+        oneMediaFile.fileURL = "https://testsite.files.wordpress.com/2020/10/wp-0000000.jpg"
+        return oneMediaFile
+    }
+
+    private fun getOneStoryFrameItem(id: String): StoryFrameItem {
+        return StoryFrameItem(
+                source = mock(),
+                id = id
+        )
+    }
```
jd-alexander: Could we remove the `one` in the names of the functions/variables here? If not then let me know. Thanks :)

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
```diff
@@ -0,0 +1,288 @@
...
+        )
+
+        // Then
+        verify(storiesPrefs, times(3)).saveSlideWithTempId(any(), any(), any())
+    }
+
+    private fun setupFluxCMediaFiles(
+        emptyList: Boolean
+    ): ArrayList<MediaFile> {
+        when (emptyList) {
```
jd-alexander: we can lift the return out of the when here :)

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
```diff
@@ -0,0 +1,288 @@
...
+                    oneMediaFile.mimeType = "image/jpeg"
+                    oneMediaFile.fileURL = "https://testsite.files.wordpress.com/2020/10/wp-0000000.jpg"
+                    mediaFiles.add(oneMediaFile)
+                }
+                return mediaFiles
+            }
+        }
+    }
+
+    private fun getOneMediaFile(id: Int): MediaFile {
```
jd-alexander: We can put a default value of 1 here for this `id` since it's always 1 in this test.

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
mzorz: 👍 done in 6b27610f4e thanks for the suggestion!

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
mzorz: excellent yes done in 6b27610f4e

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
mzorz: 😅 🤕 🤦 IDK how I missed that one 🤷‍♂️ - done in 0575171172

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
mzorz: 👍 done 87d8f7b4e1

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
mzorz: > he taught me that a unit test should have one single reason to fail, hence the name unit

I see, this is understandable 👍 

> is that mediaFileData is not null and was returned from the use case

This got me thinking - `buildMediaFileDataWithTemporaryId` returns a new object and cannot return null. Hence, it doesn't make sense to write a test to check the result is not null. It will always be not null, right? That's why I'm checking for the properties of the object as I know the resulting object should look like `This` given the source object looks like `That` , does that make sense? Wondering how we could write things here. Thoughts?

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
mzorz: 👍 252b338763

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
```diff
@@ -0,0 +1,288 @@
...
+
+        // When
+        saveStoryGutenbergBlockUseCase.replaceLocalMediaIdsWithRemoteMediaIdsInPost(
+                onePost,
+                mediaFile
+        )
+
+        // Then
+        Assertions.assertThat(onePost.content).isEqualTo(blockWithNonEmptyMediaFilesWithOneRemoteId)
+    }
```
jd-alexander: Can we just call this `post` instead of `onePost`? 😄

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
mzorz: I prefer to keep it explicit if we needed to extend it, since for example there's a correlation between the id given in the constructor param and the way things are setup in [this line here](https://github.com/wordpress-mobile/WordPress-Android/pull/13103/files#diff-55f5e21ccf6cc6baa9e15eb33285d543008f79657547a539be6b68ef329d3283R115) where the `+ 1`   in the `TEMPORARY_ID_PREFIX + 1` is relevant and needs to match the value passed in the constructor. Sounds good?

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
mzorz: s-ure, d-one 🤣 828d5b9ea8

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
```diff
@@ -0,0 +1,190 @@
...
+    private lateinit var siteModel: SiteModel
+
+    @Before
+    fun setUp() {
+        loadStoryFromStoriesPrefsUseCase = LoadStoryFromStoriesPrefsUseCase(
+                storyRepositoryWrapper,
+                storiesPrefs,
+                mediaStore
+        )
+        siteModel = SiteModel()
```
jd-alexander: We can mock this `siteModel` as well :)

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
```diff
@@ -0,0 +1,190 @@
...
+        val mediaFiles: ArrayList<HashMap<String, Any>> = setupMediaFiles(emptyList = false)
+
+        // When
+        val mediaIds = loadStoryFromStoriesPrefsUseCase.getMediaIdsFromStoryBlockBridgeMediaFiles(
+                mediaFiles as ArrayList<Any>
+        )
+
+        // Then
+        Assertions.assertThat(mediaIds).isNotEmpty
+        Assertions.assertThat(mediaIds).containsExactly("1", "2", "3", "4", "5", "6", "7", "8", "9", "10")
```
jd-alexander: Can we split these assertions into two tests are remove one of them? Again, this advise is based on the discussion above 😄

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
mzorz: 👍 4d00ef2ce0

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    @Test
+    fun `verify all story slides are editable with temporary ids`() {
+        // Given
+        val mediaIdsTemp = setupTestSildes(sayValid = true, useTempPrefix = true, useRemoteId = false)
```
jd-alexander: can we rename this to `tempMediaIds`?

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
```diff
@@ -0,0 +1,190 @@
...
+                                siteModel.id.toLong(),
+                                RemoteId(mediaId.toLong())
+                        )
+                ).thenReturn(sayValid)
+            } else {
+                whenever(storiesPrefs.isValidSlide(
+                            siteModel.id.toLong(),
+                            LocalId(mediaId.toInt())
+                        )
+                ).thenReturn(sayValid)
```
jd-alexander: how about `isValidSlide` instead of `sayValid`?

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
mzorz: Removed the `isNotEmpty` check (given it is redundant) in 204dc93975

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
```diff
@@ -0,0 +1,190 @@
...
+
+        // Then
+        Assertions.assertThat(result).isFalse()
+    }
+
+    private fun setupMediaFiles(
+        emptyList: Boolean
+    ): ArrayList<HashMap<String, Any>> {
+        when (emptyList) {
+            true -> return ArrayList<HashMap<String, Any>>()
```
jd-alexander:
```suggestion
            true -> return ArrayList()
```

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
```diff
@@ -0,0 +1,190 @@
...
+                val mediaFiles = ArrayList<HashMap<String, Any>>()
+                for (i in 1..10) {
+                    val oneMediaFile = HashMap<String, Any>()
+                    oneMediaFile.put("mime", "image/jpeg")
+                    oneMediaFile.put("link", "https://testsite.files.wordpress.com/2020/10/wp-0000000.jpg")
+                    oneMediaFile.put("type", "image")
+                    oneMediaFile.put("id", i.toString())
+                    mediaFiles.add(oneMediaFile)
+                }
+                return mediaFiles
```
jd-alexander: Let's lift this return out of the when.

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
mzorz: 👍 f51842da4f

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
```diff
@@ -0,0 +1,190 @@
...
+    private fun setupMediaFiles(
+        emptyList: Boolean
+    ): ArrayList<HashMap<String, Any>> {
+        when (emptyList) {
+            true -> return ArrayList<HashMap<String, Any>>()
+            false -> {
+                val mediaFiles = ArrayList<HashMap<String, Any>>()
+                for (i in 1..10) {
+                    val oneMediaFile = HashMap<String, Any>()
+                    oneMediaFile.put("mime", "image/jpeg")
```
jd-alexander: This could be simplified like so 
`oneMediaFile["mime"]` This is a nitpick so in the interest of time it can be left.

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
```diff
@@ -0,0 +1,190 @@
...
+                    oneMediaFile.put("type", "image")
+                    oneMediaFile.put("id", i.toString())
+                    mediaFiles.add(oneMediaFile)
+                }
+                return mediaFiles
+            }
+        }
+    }
+
+    private fun setupTestSildes(
```
jd-alexander: Typo in the name here 😉 `setupTestSildes` -> `setupTestSlides`

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
mzorz: I intended to mean the parameter is telling the test setup to `sayValid` (i.e. say that something is valid whenever a certain check is made), as opposed to checking whether something isX or not. I'm ok changing it though, but wondering if it makes sense to you with this clarification?

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
jd-alexander: Indeed. I totally understand this. Normally I would just duplicate the method logic meaning the parts under the given and when and then verify each property of the object but honestly, in this case, we can probably just leave this as is to not cause too much duplicate logic. So yes, it's up to you :)

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
jd-alexander: 😅

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
mzorz: done in e8a1f558d7

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
mzorz: done in e8a1f558d7, alone with renaming `oneMediaFile` to `mediaFile`

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
mzorz: nice catch, fixed in d9b260f046

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
mzorz: Thanks for the explanation! I think it makes sense to verify that the properties of the source object have been mapped to the target object in the way defined by each of the assertion statements. If any of these fail, I want the test to fail 👍

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
jd-alexander: Thanks. Could we do the method names as well?

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
jd-alexander: Indeed. Makes sense 🙇 Could we make the name of the test, something like `verify that all properties of mediaFileData that are created from buildMediaFileDataWithTemporaryId are correct`? I want it to be a bit a clearer that these assertions are intention.

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
jd-alexander: Okay, I get what you are you saying now. How about `markAsValid` ?

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
mzorz: yes, renamed in 10a8ee0562 - also dropped the second word `that` because of line length reasons

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
mzorz: sure, done in f324acdce3

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
jd-alexander: oh I get you. In that case let's put the 1 in a variable and the temp in one as well for reusability and clarity. 

so something like 
```
        // Given
        val mediaFileId = 1
        val mediaFile = getOneMediaFile(mediaFileId)
       val tempMediaFileId = TEMPORARY_ID_PREFIX + mediaFileId 
```

Please let me know if this is helpful. Thank you.

WordPress/src/test/java/org/wordpress/android/ui/stories/usecase/LoadStoryFromStoriesPrefsUseCaseTest.kt
mzorz: 👍 05ba52ee3e

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
jd-alexander: 😆

WordPress/src/test/java/org/wordpress/android/ui/stories/SaveStoryGutenbergBlockUseCaseTest.kt
mzorz: > Please let me know if this is helpful. 

It is! done in 621abbf388

jd-alexander (approved):
Thanks for these tests @mzorz all is well from my side. LGTM 🚢

PR #13104, opened 2020-10-11
Adding Gif data source for new media picker

develric:
This PR is part of the implementation for #12642 

The code here does the following:

1. Adds a GifMediaDataSource to the new media picker
2. Adds a GifMediaInsertUseCase that manages the insertion of selected Gif items in Media Browser and GB blocks
3. Adding possibility to use different strings in upload dialog depending on insert use case

Note that I also restored in the `MediaPickerFragment` recycler view the onSaveInstanceState/onRestoreInstanceState when loading data. This was done since without it, the recycler view does not keep positions (like the top alignment) when loading items for example while the user is taping search string (like type `cats` then delete the final `s` to see this effect in action; if it's not clear what I mean let me know 🙇  ). This effect is visible also in the Stock Media picker for example.

Things not covered in this PR that will be addressed in subsequent PRs:

- Adding debounce while user types in search strings (something like [here](https://github.com/wordpress-mobile/WordPress-Android/blob/05001d3a1a3aa95fadbe2e90bca3fbfaedf32c57/WordPress/src/main/java/org/wordpress/android/viewmodel/gif/GifPickerViewModel.kt#L385) ). This probably should be done for all data sources implementations with search (like Stock Media also)
- Add Unit Testing
- Tracking events review
- Evaluate to implement a retry mechanism on failed gif loading (something similar to the approach [here](https://github.com/wordpress-mobile/WordPress-Android/blob/05001d3a1a3aa95fadbe2e90bca3fbfaedf32c57/WordPress/src/main/java/org/wordpress/android/viewmodel/gif/GifPickerDataSource.kt#L128) )

NOTE: after merging from develop (in this [commit](https://github.com/wordpress-mobile/WordPress-Android/pull/13104/commits/bcbbe62119978e4a56271e5aa59ce23afc6b5f4f)) I tried to adapt the `GifMediaInsertUseCase -> insert` coroutine code to the new flow approach. Possibly that one can be simplified, happy to get your advice there 🙇 .

## To test
For all the below enable the `CONSOLIDATED_MEDIA_PICKER` feature flag in wasabi.

### Media Browser insert
- In media browser, + menu -> Choose from Tenor
- Fill in search string(s) in the search and check the gifs are loaded (check they are consistent with those loaded with same string(s) in the Tenor current vanilla implementation)
- Scroll down the loaded GIFs and check the pagination works (check loaded GIFs and pages are consistent with those loaded with same string(s) in the Tenor current vanilla implementation)
- Select more than one gifs and check the selection progressive counter is consistent
- Try also unselecting some of the gifs to check the counter number stays progressive
- Confirm selection and check the upload dialog is presented with the string `Uploading gif media`
- Check you are brought back to media browser and the gifs are loaded correctly
- Smock test (like check you can preview the uploaded GIFs and similar)

### GB Image/Gallery block
- In GB add Image/Gallery block -> Choose from Tenor
- Insert a search string
- Select GIF and accept
- Note the GIF is uploaded and added to the block
- Smock test

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaItem.kt
```diff
@@ -17,20 +18,61 @@ data class MediaItem(
...
+            val remoteId: String?,
+            val mediaModel: MediaModel?,
+            val thumbnailUri: Uri?,
+            val previewImageUri: Uri?,
+            val largeImageUri: Uri?,
+            val title: String?
+        ) : Identifier()
 
-        fun toParcel() = Parcel((this as? LocalUri)?.value?.uri, (this as? RemoteId)?.value)
+        fun toParcel() = Parcel(
```
planarvoid: I was wondering if we could leverage the `@Parcelize` annotation instead of this code. The children would have to be annotated, the parent could implement a `Parcelable` interface. The `Parcel` object below could then be either like this:
```
@Parcelize
data class Parcel(val localUri: LocalUri? = null, val remoteId: RemoteId? = null.....)
```
the other option would be something like this:
```
@Parcelize
data class Parcel(val type: NewEnumType, val identifier: Identifier)
```
where we would do some kind of conditional parsing. Option A sounds simpler and it will reduce the amount of changes necessary here. WDYT?

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaItem.kt
```diff
@@ -17,20 +18,61 @@ data class MediaItem(
         data class LocalUri(val value: UriWrapper) : Identifier()
         data class RemoteId(val value: Long) : Identifier()
         data class StockMediaIdentifier(val url: String?, val name: String?, val title: String?) : Identifier()
+        data class GifMediaIdentifier(
+            val remoteId: String?,
+            val mediaModel: MediaModel?,
+            val thumbnailUri: Uri?,
```
planarvoid: Maybe you can use `UriWrapper` instead of `Uri` to help with testing?

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerFragment.kt
```diff
@@ -407,7 +407,9 @@ class MediaPickerFragment : Fragment() {
                         1
                     }
                 }
+        val recyclerViewState = recycler.layoutManager?.onSaveInstanceState()
         adapter.loadData(items)
+        recycler.layoutManager?.onRestoreInstanceState(recyclerViewState)
```
planarvoid, line 411: I think it's a good idea to have this back 👍

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerViewModel.kt
```diff
@@ -353,7 +353,7 @@ class MediaPickerViewModel @Inject constructor(
                     is InsertModel.Progress -> {
                         progressDialogJob = launch {
                             delay(100)
-                            _showProgressDialog.value = Visible(R.string.media_uploading_stock_library_photo) {
+                            _showProgressDialog.value = Visible(it.title) {
```
planarvoid, line 356: 👍

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/insert/GifMediaInsertUseCase.kt
```diff
@@ -0,0 +1,92 @@
...
+                }
+            }
+        )
+    }
+
+    private fun fetchAndSaveAsync(
+        scope: CoroutineScope,
+        gifIdentifier: GifMediaIdentifier,
+        site: SiteModel
+    ): Deferred<MediaModel?> = scope.async(Dispatchers.IO) {
```
planarvoid: I think we should use the injected dispatcher here instead of `Dispatchers.IO` for testing purposes.

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/insert/GifMediaInsertUseCase.kt
```diff
@@ -0,0 +1,92 @@
...
+    override val actionTitle: Int
+        get() = R.string.media_uploading_gif_library_photo
+
+    override suspend fun insert(identifiers: List<Identifier>) = flow {
+        emit(InsertModel.Progress(actionTitle))
+        emit(coroutineScope {
+                return@coroutineScope try {
+                    val mediaIdentifiers = identifiers.mapNotNull { identifier ->
+                        (identifier as? GifMediaIdentifier)?.let {
+                            fetchAndSaveAsync(this, it, site).await().let { mediaModel ->
```
planarvoid: This approach will not work as expected :-). I think you want the work to be done in parallel, right? However, this will call the `await` right after the `async` for each item. This means that it will wait until that item is fetched before  jumping to the next one. In order to do parallel execution you need to call the async on all the items. When you call `async`, the execution starts right away in the background. After it's called on all the items, you can take the list of deferred calls and start calling `await` on all the items. 
What happens now:
1. Map takes identifier 1
2. Fetching of 1 starts in `async` asynchronously
3. We await the result 1
4. Map takes identifier 2
...

What we need:
1. Map takes identifier 1
2. Fetching of 1 starts in `async` asynchronously
3. Map takes identifier 2
4. Fetching of 2 starts in `async` asynchronously
5. We take list of deferred items (which are being fetched in the background)
6. We await item 1 
7. We await item 2
...

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaItem.kt
```diff
@@ -17,20 +18,61 @@ data class MediaItem(
         data class LocalUri(val value: UriWrapper) : Identifier()
         data class RemoteId(val value: Long) : Identifier()
         data class StockMediaIdentifier(val url: String?, val name: String?, val title: String?) : Identifier()
+        data class GifMediaIdentifier(
```
planarvoid, line 28: I think this doesn't need all these fields, right? Only the fields necessary in the Insert use case need to be part of the identifier, are the other fields ever read from this class? It could be only `largeImageUri` and the `title`.

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/loader/GifMediaDataSource.kt
```diff
@@ -0,0 +1,131 @@
...
+@Inject constructor(
+    private val context: Context,
+    private val tenorClient: IApiClient
+) : MediaSource {
+    private var nextPosition: Int = 0
+    private var items = mutableListOf<MediaItem>()
+    private var lastFilter: String? = null
+
+    override suspend fun load(forced: Boolean, loadMore: Boolean, filter: String?): MediaLoadingResult {
+        if (lastFilter != filter) {
```
planarvoid: I was thinking about this approach (I know I used it in the stock data source). I think it's enough to check here whether the `loadMore == false`. We want to clear or the previous data every time when we are not loading the next page. Does it make sense?

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/loader/GifMediaDataSource.kt
```diff
@@ -0,0 +1,131 @@
...
+import kotlin.coroutines.resume
+import kotlin.coroutines.suspendCoroutine
+
+class GifMediaDataSource
+@Inject constructor(
+    private val context: Context,
+    private val tenorClient: IApiClient
+) : MediaSource {
+    private var nextPosition: Int = 0
+    private var items = mutableListOf<MediaItem>()
```
planarvoid: It can either be `var` or mutable list, you don't need both. Either rewrite the variable or use `clear` and `addAll` on a mutable list.

planarvoid (commented):
Thanks for this PR, it's working well! I have a few comments about the code :-) let me know what you think 👍

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/loader/GifMediaDataSource.kt
develric: ahah, totally right 😄 ! Changed in 7a5d7f94cc965e6135ae82d67391bf8d96a46c1b 👍

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/loader/GifMediaDataSource.kt
develric: Since tenor seems to have finite streams of items per each filter string, I was thinking this to be more robust in the edge case when you reach the end of list, but tbh I just tried the edge case and it works as expected, so happy to change it 👍  (changed in 7a5d7f94cc965e6135ae82d67391bf8d96a46c1b)

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaItem.kt
develric: I left it like this after originally mapping the fields from `data class MutableGifMediaViewModel` and then forgot to simplify 🤦‍♂️! Changed in 546cc427e28d0554813d57c93a7ffd6de24a1817 👍

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/insert/GifMediaInsertUseCase.kt
develric: That makes sense 👍 ; thanks for catching it 🙇 ! Changed in 24bb569c4229d6c105b9dd1f2e1e5de3083d7793

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/insert/GifMediaInsertUseCase.kt
develric: Changed in 57b8cf50f3fa70c61514390808339a829010dd78

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaItem.kt
develric: Tried to simplify the parcelize and GifMediaIdentifier in 546cc427e28d0554813d57c93a7ffd6de24a1817

develric:
Hi @planarvoid, thanks for the review 🙇 ! I should have followed up to your comments, should be ready for another pass, thanks 😄 .

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaItem.kt
```diff
@@ -13,24 +14,21 @@ data class MediaItem(
...
-                }
-            }
-        }
+        @Parcelize
+        data class StockMediaIdentifier(val url: String?, val name: String?, val title: String?) : Identifier()
 
         @Parcelize
-        data class Parcel(val uri: Uri? = null, val remoteId: Long? = null) : Parcelable
+        data class GifMediaIdentifier(
+            val mediaModel: MediaModel?,
```
planarvoid, line 29: I was thinking about this. Do you think it might make sense to split the `GifMediaIdentifier` into 2 classes, one without the `MediaModel` and the other called something like `LocalId` (we already have `RemoteId`)? And map one to the other in the Insert use case? It might even work to change the functionality from using `LocalId` to `RemoteId` since we're using that for the other cases (even stock library). This would probably require a little more work (and we can do that in a separate PR) but might be a good step towards consistency.

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/insert/GifMediaInsertUseCase.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    override suspend fun insert(identifiers: List<Identifier>) = flow {
+        emit(InsertModel.Progress(actionTitle))
+        emit(coroutineScope {
+                return@coroutineScope try {
+                    val mediaIdentifiers = identifiers.mapNotNull { identifier ->
+                        (identifier as? GifMediaIdentifier)?.let {
+                            fetchAndSaveAsync(this, it, site)
+                        }
+                    }.map { it.await() }
```
planarvoid, line 45: NIT: I've found a useful method `awaitAll` which we can use here.

planarvoid (approved):
It's looking really good :+1:, good job. I have 2 minor comments so I'm approving this PR. Feel free to ping me if you decide to change something

develric:
Hey @planarvoid , thanks for the review 😊. I think I will open pretty soon a PR with a small refactor to allow testability of data source, so if you agree we could merge this one and I can address your suggestions there, this one is already pretty big. let me know 🙇

PR #13058, opened 2020-10-06
Reader Detail Updates (Post Header UI Polishing)

ashiagr:
Task: #12900

This PR polishes post header UI, particularly

- Applies new follow button style
- Updates post title style

Looking for design review for UI updates so far (except for the backlog items in #12900 )

**Mobile**

| Light Mode Following | Light Mode Follow | Dark Mode Following | Dark Mode Follow |
|-----|-----|-----|-----|
|device-2020-10-06-125504|device-2020-10-06-125513|device-2020-10-06-125548|device-2020-10-06-125602|

**Mobile Landscape**
| Following | Follow | 
|-----|-----|
|device-2020-10-06-125418| device-2020-10-06-125449

**Tablet Light Mode** 

| Following | Follow | 
|-----|-----|
|device-2020-10-06-125831| device-2020-10-06-125848|

**Tablet Dark Mode** 

| Following | Follow |
|-----|-----|
|device-2020-10-06-130145| device-2020-10-06-130153|

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13106, opened 2020-10-12
Reader Detail Updates (Sticky Action Bar)

ashiagr:
Task: #12900

This PR 
- Makes the action bar sticky so that it doesn't disappear/reappear on scroll
- Makes footer layout changes (converts to flat constraint layout) to separates icons evenly 

sticky_action_bar

**Merge instructions**

1. Make sure #13058 is merged
4. Merge this PR

Note:

I want to refactor code for the footer actions to reuse `ReaderPostCardActionsHandler`. I'll try to do it in a separate PR once other changes are implemented.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostDetailFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostDetailFragment.kt
@@ -104,7 +104,6 @@ import org.wordpress.android.ui.reader.utils.FeaturedImageUtils
 import org.wordpress.android.ui.reader.utils.ReaderUtils
 import org.wordpress.android.ui.reader.utils.ReaderUtilsWrapper
 import org.wordpress.android.ui.reader.utils.ReaderVideoUtils
-import org.wordpress.android.ui.reader.views.ReaderBookmarkButton
 import org.wordpress.android.ui.reader.views.ReaderFollowButton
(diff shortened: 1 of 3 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostDetailFragment.kt
```diff
@@ -997,11 +980,12 @@ class ReaderPostDetailFragment : ViewPagerFragment(),
                 }
             }
         } else {
-            reblogButton.visibility = View.GONE
+            reblogButton.isEnabled = false
             reblogButton.setOnClickListener(null)
         }
 
         if (canShowCommentCount()) {
+            countLikes.isEnabled = true
```
malinajirka: Could you please double check this line: Should this be "countComments"?

malinajirka (commented):
Thanks @ashiagr! Great job - I've left one comment but it looks good otherwise.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostDetailFragment.kt
ashiagr: Nice catch, thank you! Fixed in 7bd6f38

malinajirka (approved):

PR #13118, opened 2020-10-13
Implement debounce on search and add test for search

planarvoid:
This PR adds a simple debounce mechanism to search. This is useful when the user is typing in the search field. In this case we don't want a request and result for each letter typed.

I'm not sure how to test the debouncer so I just added a simple test for search

To test:
- Turn on the feature
- Go to Media library/Add from stock library
- Start search
- Type "Cat"
- Notice the search results
- Switch to "Cats"
- Notice different results
- Switch fast between Cat and Cats and notice only the last option gets searched for

PR submission checklist:


develric (approved):
Simple and effective 😄 . Works well for me 👍

PR #13123, opened 2020-10-13
Update mocks for screenshots

JavonDavis:
Updates the mocks for the app store screenshots

To test: Run screenshot tests

PR submission checklist:


designsimply (approved):
I was able to get tests to work after updating to ddf31668bc! 🎉

The Discover screenshot looks amaaaazing. Yay! 😍

Nitpick: there's a gray overlay on the stats screen that isn't there when I browse with the app directly. Note that this may not matter if the overlay isn't present when generating screenshots at the fastlane level (I observed this in the test directly).

Screenshot_1602693727

Nitpick: images in the media screenshot don't match the ones in the figma. This also might not matter but we could double check with Megs on the design if you want!

Screenshot_1602700812

PR #13124, opened 2020-10-13
13043/improve save progress dialog handling

### LINKED ISSUE #13043: IllegalArgumentException: [EditPostActivity] not attached to window manager
**Sentry Url**: https://sentry.io/share/issue/03fe12bff0a8429d936665b30910c200/
**User Count**: 200
**Count**: 246
**First Release**: org.wordpress.android@15.7-rc-3+920
**First Seen**: 2020-09-17T11:37:04Z
**Last Seen**: 2020-10-01T07:54:50Z
**24 Hours**: 42
**30 Days**: 246

```
IllegalArgumentException: View=DecorView@ec8315a[EditPostActivity] not attached to window manager
     at org.wordpress.android.editor.gutenberg.GutenbergEditorFragment.hideSavingProgressDialog(GutenbergEditorFragment.java:935)
     at org.wordpress.android.ui.posts.EditPostActivity.lambda$updateAndSavePostAsyncOnEditorExit$16(EditPostActivity.java:1601)
     at org.wordpress.android.ui.posts.-$$Lambda$EditPostActivity$EdCXO7W5A5ZADiksnSdosr5m3QE.onPostUpdatedFromUI(null:6)
     at org.wordpress.android.ui.posts.EditPostActivity.lambda$updateAndSavePostAsync$15(EditPostActivity.java:1571)
     at org.wordpress.android.ui.posts.-$$Lambda$EditPostActivity$ZG_Jw9txlRYaZ-qVEi4LxIeTO28.invoke(null:6)
     at org.wordpress.android.ui.posts.EditPostRepository$updateAsync$1.invokeSuspend(EditPostRepository.kt:128)
...
(10 additional frame(s) were not displayed)
```

mchowning:
Fixes #13043 by improving the handling of the Gutenberg fragment's save in progress dialog.

## To Test

**Before these tests, enable "Don't keep activities" in your developer preferences.** Also, be prepared for some really severe slowness with the editor--it does not handle posts as large as the test steps call for very well at all.

#### a. Happy Path Save-On-Exit
1. Open a very long post ([for example](https://gist.github.com/mchowning/61f1dadc6d3829436ba8531c1e1803ab)) in the gutenberg editor
2. Exit the editor by tapping the back arrow
3. Observe the save dialog appear briefly and then close when you are returned to the previous screen

#### b. Save-On-Exiting HTML mode
Be prepared for some pretty significant slowness when switching to the HTML editor with a long post (see https://github.com/wordpress-mobile/gutenberg-mobile/issues/2716) 😞 
1. Open a very long post ([for example](https://gist.github.com/mchowning/61f1dadc6d3829436ba8531c1e1803ab)) in the gutenberg editor
2. Tap the HTML mode button in the three dot menu
3. Wait for HTML mode to open (it is [going to take a while with a long post](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2716))
4. Exit the editor by tapping the back arrow
5. Observe the save dialog appear briefly and then close when you are returned to the previous screen

#### c. Background During Save-On-Exit
1. Open a very long post ([for example](https://gist.github.com/mchowning/61f1dadc6d3829436ba8531c1e1803ab)) in the gutenberg editor
2. Tap the back arrow to exit the editor
3. Immediately background the app (you should have seen the save in progress dialog popup)
4. Check the logs for 15 seconds or so and you should not see any exceptions relating to the window management from our app (i.e., no (1) `IllegalArgumentException View=DecorView@7daa612[EditPostActivity] not attached to window manager` or (2) `android.view.WindowLeaked: Activity org.wordpress.android.ui.posts.EditPostActivity has leaked window DecorView@ec49822[EditPostActivity]...` errors). I was not able to reproduce the "not attached" error noted in [the Sentry issue](https://sentry.io/share/issue/03fe12bff0a8429d936665b30910c200/) consistently even without this fix, but it did occur for me sometimes.
5. Reopen the app, and you should be returned to the post with no problems. Ideally, we would have moved the user out of the editor here (since they pressed the back button before backgrounding the app), but this seems like a minor issue to me that does not need to be resolved in this PR.

#### d. Rotate During Save-On-Exit
1. Open a very long post ([for example](https://gist.github.com/mchowning/61f1dadc6d3829436ba8531c1e1803ab)) in the gutenberg editor
2. Tap the back arrow to exit the editor
3. Immediately rotate the device to change orientation
4. Observe that until the save is complete and the editor closes, the save in progress dialog is presented--even after rotation. In other words, the save in progress dialog should block you from editing the post after rotation. Note that this test only works if it takes a decently long time to save the post.

#### a. Autosave Happy Path
1. Open a very long post ([for example](https://gist.github.com/mchowning/61f1dadc6d3829436ba8531c1e1803ab)) in the gutenberg editor
2. Make some changes to the post
3. Wait a couple of seconds so the autosave is triggered
4. You should **not** see the save dialog appear

#### b. Rotate During Autosave
1. Open a very long post ([for example](https://gist.github.com/mchowning/61f1dadc6d3829436ba8531c1e1803ab)) in the gutenberg editor
2. Make some changes to the post
3. Immediately rotate the device to change orientation
4. Observe that even if the autosave is not complete the save in progress dialog is not presented. In other words, the save-in-progress dialog only presents itself after rotation if the editor is doing a save-on-exit save and not an autosave save.

PR submission checklist:


libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
```diff
@@ -357,6 +359,19 @@ public void run() {
...
     }
 
+    private void initializeSavingProgressDialog() {
+        if (mEditorFragmentListener != null) {
+            mEditorFragmentListener.getSavingInProgressDialogVisibility()
+                                   .observe(getViewLifecycleOwner(), visibility -> {
+                                       if (DialogVisibility.Showing == visibility) {
+                                           showSavingProgressDialogIfNeeded();
+                                       } else {
+                                           hideSavingProgressDialog();
```
mchowning, line 373: It doesn't seem like we actually _have_ to hide the dialog _both_ here and in `onDestroy`, but I decided to go with the double safety of using both since this is a blocking dialog. Also, it makes the code clearer if the view model is sending both signals (show and hide) and we just also hide the dialog in `onDestroy` as a fallback.

WordPress/src/test/java/org/wordpress/android/ui/posts/editor/StorePostViewModelTest.kt
```diff
@@ -229,4 +232,26 @@ class StorePostViewModelTest : BaseUnitTest() {
...
+
+        viewModel.showSaveProgressDialog()
+        expected.add(Showing)
+        assertThat(actual).isEqualTo(expected)
+
+        val randomNonNullFinishState = SAVED_LOCALLY
+        viewModel.finish(randomNonNullFinishState)
+        expected.add(Hidden)
+        assertThat(actual).isEqualTo(expected)
+    }
```
mchowning, line 256: I was a bit conflicted about whether the ViewModel changes really needed unit tests since they're just basically passed along method calls (i.e., if `a()` gets called, make sure it calls `b()`), but I do like to test view models, so I split the difference by writing this test to kind-of, sort-of test the "lifecycle" of the progress dialog's live data. I'm fine changing this to a more typical set of unit tests if you prefer (i.e., separate tests for initialization, showing the dialog, and hiding the dialog).

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -1659,12 +1661,8 @@ private void updateAndSavePostAsyncOnEditorExit(@NonNull final OnPostUpdatedFrom
         // Store this before calling updateAndSavePostAsync because its value can change before the callback returns
         boolean isAutosavePending = mViewModel.isAutosavePending();
 
-        mEditorFragment.showSavingProgressDialogIfNeeded();
+        mViewModel.showSaveProgressDialog();
         updateAndSavePostAsync((result) -> {
-            if (mEditorFragment != null) {
-                mEditorFragment.hideSavingProgressDialog();
```
mchowning, line 1727: Since we only show the saving progress dialog when we are exiting the Editor (at least that's what it looks like to me, let me know if you think I might be mistaken), so the `LiveData` listening for `EditPostActivity` to call the view model's finish method and the `GutenbergEditorFragment::onDestroy` can take care of dismissing the dialog.

WordPress/src/test/java/org/wordpress/android/ui/posts/editor/StorePostViewModelTest.kt
malinajirka: LGTM ;)

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
malinajirka: Yep, I agree ;). Thanks

malinajirka (approved):
LGTM, thanks @mchowning!

I noticed an unrelated issue which I'm not sure if we already discussed in the past or not. When I open a published post -> close the editor -> the post contains some local changes even though I haven't changed anything. I'll submit an issue, since this behavior might result in a lot of version conflicts.

cc @loremattei I'm targeting 16.0 as we discussed. Thanks for waiting 🙇

PR #13128, opened 2020-10-14
Reader Detail Updates (Update footer actions using the VM)

ashiagr:
Task: #12900

This PR 
- Updates footer actions using the VM
- Removes now unused footer action methods (`refreshIconCount`, `setPostLike`, `togglePostLike` , `toggleBookmark`) from the fragment

**To Test**

1. Click a reader post
2. Verify that **click action on footer icons** behave correctly as before
3. Toggle like for the opened post from the website and verify that like counts are refreshed correctly on **pull to refresh** 
4. Verify that a **direct operation** (operation to perform automatically when opened via deeplinking) to like a post behaves correctly as before
(You can test it by replacing `directOperation` by `DirectOperation.POST_LIKE` on [this line](https://github.com/wordpress-mobile/WordPress-Android/blob/2455d927a42315687155e4e754d97026365a43d2/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostDetailFragment.kt#L1406)
`args.putSerializable(ReaderConstants.ARG_DIRECT_OPERATION, directOperation)`)

**Merge instructions**

1. Make sure #13129 is merged
4. Merge this PR

// TODO (added to backlog)
Add missing tracking
READER_POST_SAVED_FROM_DETAILS
READER_POST_UNSAVED_FROM_DETAILS
READER_SAVED_LIST_VIEWED_FROM_POST_DETAILS_NOTICE

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostDetailFragment.kt
```diff
@@ -1531,7 +1332,7 @@ class ReaderPostDetailFragment : ViewPagerFragment(),
      * can we show the footer bar which contains the like & comment counts?
      */
     private fun canShowFooter(): Boolean {
-        return canShowLikeCount() || canShowCommentCount() || canShowBookmarkButton()
+        return canShowLikeCount() || canShowCommentCount() || canShowBookmarkButton() || canBeReblogged()
```
malinajirka: AFAIK we never hide the actions in the post list - is this condition ever false? For what types of posts do we need to hide the footer?

malinajirka (approved):
Thanks @ashiagr! LGTM, great job!

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostDetailFragment.kt
ashiagr: This condition was already present, I've not tested extensively if it can ever happen. I just didn't want to break existing logic by removing it.
That said I also think that it is fine to display footer with all actions disabled if it ever happens 🤷‍♀️.

Should I remove `canShowFooter` logic?

malinajirka: Yeah, I understand why you didn't want to just remove it. At the same time, since we just worked on the post list logic I think no-one will have a better context than we do right now. So I'd just remove it.

ashiagr: Removed in d6cc863

malinajirka:
LGTM, thanks!

PR #13131, opened 2020-10-14
Introduce loading and error states in the media picker

planarvoid:
This PR was much more complicated than I thought because there are so many cases that needed to be covered. The main changes I did were to implement 3 new states:
- Loading state - it's shown when the data is being loaded. This could potentially be improved because it says "Loading" even during search.
- Full screen error - this is shown when the initial loading fails. This needed to include the Retry action so there is now Retry mechanism in the `MediaLoader` which remembers the last performed action and on `Retry` retries it. 
- Next page loading error - this required me to add the current data to the `Failure` loading result. This data is present only if loading of the next page fails. 

As part of this PR I've implemented check whether network is present to WP Media library, stock library and gif library. 

To test (everything with the flag turned on):
- Slow down your internet
- Go to media library
- Add a post/Add image block/Add media
- Choose your WP media library
- Notice the `Loading` screen
- Search for something - Notice the loading screen
- Turn off the internet on your device
- Search for something
- Notice a full screen error
- Turn on the internet
- Click on "Retry"
- Notice the previous search gets loaded
- Turn off the internet
- Scroll down
- Notice the Next page fails and a "Retry" button appears bellow the results
- Turn on the internet
- Click on "Retry"
- Notice the next page is loaded

To test (everything with the flag turned on) - with Tenor library and stock library:
- Go to media library
- Add a post/Add image block/Add media
- Choose your source
- Search for something - Notice the loading screen
- Turn off the internet on your device
- Search for something
- Notice a full screen error
- Turn on the internet
- Click on "Retry"
- Notice the previous search gets loaded
- Turn off the internet
- Scroll down
- Notice the Next page fails and a "Retry" button appears bellow the results
- Turn on the internet
- Click on "Retry"
- Notice the next page is loaded

PR submission checklist:


develric (commented):
Hey @planarvoid, this is looking good, good job 👍 . During testing I found the following that I'm reporting for your feedback 🙇 

- first test case step 5 `Notice the Loading screen` is not triggered for me even with slow connection; I directly get the items loaded without passing (at least visually) for the loading state. Tried with EDGE connection on emulator to slow things down.
- When setting slow connection (like EDGE) active (so not airplane mode) I'm getting some images not loading even if no error seems triggered from glide (at least this is the effect, I need to dig further into the code to be sure). Maybe I can open an issue for this since I think it's not related directly to this PR (or maybe the MP specifically also); let me know wdyt 🙇 

<p align="center">

</p>

planarvoid:
hey @develric , thanks for the review 👍 

> first test case step 5 Notice the Loading screen is not triggered for me even with slow connection; I directly get the items loaded without passing (at least visually) for the loading state. Tried with EDGE connection on emulator to slow things down.

I've fixed this in 5e0dd7c. You can check the PR :-), for some reason I was explicitly not emitting the Loading event on start.

> When setting slow connection (like EDGE) active (so not airplane mode) I'm getting some images not loading even if no error seems triggered from glide (at least this is the effect, I need to dig further into the code to be sure). Maybe I can open an issue for this since I think it's not related directly to this PR (or maybe the MP specifically also); let me know wdyt 🙇

I think this is unrelated to the media picker so we can create an issue for a possible improvement. Maybe change it visually or add retry buttons.

develric (approved):
Hey @planarvoid thanks for the follow up; LGTM, I'm merging this one 👍

PR #13136, opened 2020-10-14
Gutenberg : Html mode switching after image upload cancellation

jd-alexander:
Fixes # https://github.com/wordpress-mobile/gutenberg-mobile/issues/2048

## Findings
Based on my investigation it seemed the root cause of this issue involved the `localMediaId` being used to track the progress of each upload being removed as the wrong type from the `mUploadingMediaProgressMax`ConcurrentHashMap which caused no removal at all. Also, when the upload is canceled via the dialog, the `EditPostActivity` would still send another progress event after the cancellation which would put a progress value back into the `mUploadingMediaProgressMax`ConcurrentHashMap so even though the positive action of the cancel dialog was removing the progress value for the `localMedia` id from the map it would be put back in. Hence when you clicked on the `Html mode` action in the overflow menu, it would still think that's there's still an upload in progress even though it was canceled.

## Solution
The solution was for this issue was two-fold. First, the `localMediaIds` being used as as `Integers` in places that were expecting a `String` were converted to a `String` and a `Set<String>` was incorporated to track the media ids that were canceled so progress events for them could be ignored. 

Further comments annotate the code below.

## Testing

Before testing, you can lessen the network speed of your emulator by clicking on the menu option of the emulator, and choosing the lowest Network type and Signal strength to ensure the upload doesn't happen too quickly. 

1. Start a Gutenberg post and add an image block.
2. Add media by selecting "Choose from device". Before the upload finishes tap on the image to trigger the "Stop uploading" dialog and select "yes" to cancel the upload.
3. Tap on the 3-dot menu and select the "HTML mode" option
4. The editor changes mode and a toast message come up to announce that the switch occurred.

<kbd></kbd>

-----------------

1. Start a Gutenberg post and add an image block.
2. Add media by selecting "Choose from device". Before the upload finishes tap on the 3-dot menu and select the "HTML mode" option.
3. The editor should show a toast message that the switch can't occur during upload. 

<kbd></kbd>

------------------

1. Start a Gutenberg post and add an image block.
2. Add media by selecting "Choose from device". 
3.After the upload finishes tap on  the 3-dot menu and select the "HTML mode" option
4. The editor changes mode and a toast message come up to announce that the switch occurred.

<kbd></kbd>

## Reviewing 

## Submitter Checklist


libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
```diff
@@ -962,8 +964,13 @@ public void onMediaUploadSucceeded(final String localMediaId, final MediaFile me
...
     @Override
     public void onMediaUploadProgress(final String localMediaId, final float progress) {
-        mUploadingMediaProgressMax.put(localMediaId, progress);
-        getGutenbergContainerFragment().mediaFileUploadProgress(Integer.valueOf(localMediaId), progress);
+        if(!mCancelledMediaIds.contains(localMediaId)) {
+            mUploadingMediaProgressMax.put(localMediaId, progress);
+            getGutenbergContainerFragment().mediaFileUploadProgress(Integer.valueOf(localMediaId), progress);
+        }
+        else {
+            mCancelledMediaIds.remove(localMediaId);
```
jd-alexander: I am wondering if this approach of removing the `localMediaId` once it's within the `mCancelledMediaIds` is appropriate. So far during my testing only only one `onMediaUploadProgress` event is fired after cancellation is requested via the event sent from the `EditPostActivity` to the `UploadService`. Let me know your thoughts on this. I did this to ensure that if the same image with the same local media id is being reuploaded this does not prevent the progress from being propagated to the editor.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
```diff
@@ -707,19 +709,19 @@ private void toggleHtmlMode() {
...
         mEditorFragmentListener.onTrackableEvent(TrackableEvent.HTML_BUTTON_TAPPED);
-        mEditorFragmentListener.onHtmlModeToggledInToolbar();
 
         // Don't switch to HTML mode if currently uploading media
         if (!mUploadingMediaProgressMax.isEmpty() || isActionInProgress()) {
             ToastUtils.showToast(getActivity(), R.string.alert_action_while_uploading, ToastUtils.Duration.LONG);
             return;
         }
 
+        mEditorFragmentListener.onHtmlModeToggledInToolbar();
```
jd-alexander, line 721: I put this call after the `return` above because in the gifs generated and also the gif of the primary issue on `gutenberg-mobile` you can see where the Snackbar is saying that the switch was made even when it wasn't so once the upload is still in progress this line shouldn't be called hence the parent `EditPostActivity` should not be able to invalidate the menu which triggers the snackbar that shows the `Switchted to X Mode` behavior.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
mzorz: yeah I get what you mean. I was thinking if the event comes rapidly after the cancellation, then we can disregard it. But if it comes say, after 2+ seconds then we can assume with some level of confidence that  this may be treated as a "new attempt". So we'd have to record the time when the user cancelled it along with its id, instead of it being a HashSet it could be a HashMap that links the time of cancellation to the localMediaid as key, then when onMediaUploadProgress is called we'd check if this is contained in mCancelledMediaIds.keys, and if yes we check the time elapsed before deciding to remove or not.
A proper signal would be to check when the user inserted the image in the first place, but I'm not sure we have a signal for this through the bridge. It may be appropriate to add a bridge method, but it also may be overkill for a corner case bug anyway.
I think adding the time check adds extra security and is not that much code to add, without going for the full blown solution, wdyt?

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
mzorz: 👍 sounds good

mzorz (approved):
Than you for these changes @jd-alexander !

I think this PR is mostly right - I would probably add a time check as left in the comments for extra security, but leaving the decision up to you. 👍 

Let me know your thoughts

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
jd-alexander: Yes, I ran another test on this and realized I would need to change the logic in the `EditPostActivity` a bit because the  `MenuItem` triggers the snackbar saying the switch has happened before the logic in the `GutenbergEditorFragment` is run and it's determined that a switch can be made based on the upload status or not.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
jd-alexander: Awesome. This sounds like a really sound idea. Thanks, much @mzorz I did an implementation in f60b6fd so I would like your feedback on my approach.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
jd-alexander: I did the change here so everything works as expected now a1f54c7

To test you will realize that if you try to switch to HTML mode during upload it will no longer show the snackbar. It will only show the toast telling you that the upload is in progress.

mzorz (approved):
Thanks for this work and updates @jd-alexander 🙇 
:shipit:

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
mzorz: thanks for this!

PR #13144, opened 2020-10-15
Reader Detail Updates (Comments entry view)

ashiagr:
Task: #12900

This PR adds comments entry view at the bottom of reader details view as per [Figma design](https://www.figma.com/file/yG8ieha0yESnr8wCtplr29/Post-Detail-View?node-id=13%3A30).

Light mode | Dark mode
------------|------------
device-2020-10-15-142117|device-2020-10-15-142149

**To test**

1. Open a reader post which allows adding comments 
2. Click "Add comment"
3. Verify that Comments screen is opened

Note: Grey border on the top and bottom are not added because of the margin on the sides. iOS mockup also do not include those borders, let me know if it is not ok. 

**Merge instructions**

1. Make sure #13128 are merged
4. Merge this PR

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostDetailFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostDetailFragment.kt
@@ -44,6 +44,7 @@ import com.google.android.material.dialog.MaterialAlertDialogBuilder
 import com.google.android.material.elevation.ElevationOverlayProvider
 import com.google.android.material.snackbar.Snackbar
 import kotlinx.android.synthetic.main.reader_fragment_post_detail.*
+import kotlinx.android.synthetic.main.reader_include_post_detail_comments_entry.*
 import kotlinx.android.synthetic.main.reader_include_post_detail_footer.*
 import org.greenrobot.eventbus.EventBus
 import org.greenrobot.eventbus.Subscribe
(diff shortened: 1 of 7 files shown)


zwarm (commented):
@ashiagr Nice work here!

I have a couple of questions
1. Should the comment bar (for lack of a better word) be fixed at the bottom of the screen or does it float with the detail? Maybe this was already discussed with @mbshakti and I'm just late to the game. When I pull up a long post and don't scroll to the bottom, I wouldn't have known I can comment from this view. And when I view a short post, the comment bar can be just about anywhere. Felt a little awkward to me.

2. The comment count did not update after I returned from entering a comment. Should there be a refresh upon return?

mbshakti:
Awesome, thanks for this Ashita. It looks exactly as the design :) 

@zwarm it's meant at the bottom of the reader detail, so not sticky or anything. 

>When I pull up a long post and don't scroll to the bottom, I wouldn't have known I can comment from this view. And when I view a short post, the comment bar can be just about anywhere. Felt a little awkward to me.

I think you are onto something here Annmarie. Android screens are quite vertically wide, and the font sizes are small, so this comment entry looks quite awkward with short posts. I don't think we encountered this problem with iOS (probably because the screens aren't so vertically wide / font sizes generally are pretty big). 

I'm wondering whether it makes sense to have this entry point now.....

ashiagr:
> I'm wondering whether it makes sense to have this entry point now.....

We already have comments action in the footer that does the same thing so it doesn't seem to add more value. 
Let me know what needs to be done here.
@mbshakti , @malinajirka

ashiagr:
@malinajirka can you approve this PR? If we want to remove it we can do it in a separate PR.

malinajirka (approved):

PR #13146, opened 2020-10-15
Task/adjust choose topics

malinajirka:
Users were complaining that they can't access the following tab unless they select an interests. We decided to rework when and where is the select interest screen shown.

Props to @zwarm for kicking of these changes

Previously, if the user wasn't following any tags, the select interest screen was shown over the Reader fragment. Users couldn't access the Reader at all unless they started following a topic.
After this change, users can access Reader no matter if they follow any topics. An empty layout is shown on the discover tab if the user isn't following any topics. The select interest screen is shown only after the user clicks on a button on the empty layout.

Responsibilities:
ReaderFragment/VM - fragment which contains a container for the select interest screen. It observes open/close interest screen events and acts accordingly.
ReaderDiscoverFragment/VM - Checks whether the user follows any tags and displays either discover feed or empty layout with "select interests" button which leads to the "select interests" screen.
ReaderInterestFragment/VM - The select interest screen itself.

To test:
1. Unfollow all tags
2. Restart the app
3. Open Reader
4. Open Discover tab
5. Notice an empty layout is shown
6. Click on the button
7. Select a few interests
8. Click on done and verify that the discover feed is refreshed automatically
9. Go to Following Tab
10. Click on the gear icon
11. Unfollow all tags
12. Go back to discover tab
13. Notice the empty layout is shown again

+ anything else you can think of

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
```diff
@@ -113,15 +120,20 @@ class ReaderDiscoverViewModel @Inject constructor(
...
-                        _uiState.value = ContentUiState(
-                                convertCardsToUiStates(posts),
-                                reloadProgressVisibility = false,
-                                loadMoreProgressVisibility = false
-                        )
+                val userTags = getFollowedTagsUseCase.get()
+                if (userTags.isEmpty()) {
+                    _uiState.value = ShowFollowInterestsEmptyUiState
+                } else {
+                    if (posts != null && posts.cards.isNotEmpty()) {
```
zwarm, line 127: @malinajirka - Is there ever a case in which the posts can be null or empty? In the future should we think about adding a generic "empty" view ? Not related to this PR, but while I was testing I forced empty data and was shown a blank screen.

zwarm (approved):
LGTM! I see the interests picker when needed and am able to navigate back. 👍

mbshakti:
It's working well, thanks a ton for making this change so quickly. 

Only two things:
1. Can we change the "follow topics to discover new blogs..." copy on the empty state of Discover to the on surface color? On light mode, it looks blue (which makes it look like a link), but I think it should be black. 

2. On dark mode, I'm not able to see the back button on the interest picker. Screenshot below. 

Screen Shot 2020-10-15 at 8 22 10 AM

malinajirka:
@mbshakti I've fixed the issues. Let me know if it looks better now;). Thanks!

| Screenshot_1602767433 | Screenshot_1602767596 | Screenshot_1602767422 |
|-----|----|---|

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
malinajirka: AFAIK it can't happen. I've noticed this potential issue as well. I think showing an empty view with "refresh" button sounds like a good solution just to make sure.

mbshakti:
Looks good, thanks @malinajirka!

PR #13147, opened 2020-10-15
Add unit tests for MediaLibraryDataSource.kt

planarvoid:
This PR adds the missing tests for the media library source and fixes 2 small bugs in the source.
The bugs were probably not noticeable but they were:
- we were ignoring the date from the media model, we were using 0 and ordering the items by ID - I think this usually works but it might cause weird effect while paging through the list
- we were not correctly returning data only when the next page loading fails but with all failures (again, this probably worked because of the store implementation)

To test:
- Check that the media library selection works as expected
- Let the tests run

PR submission checklist:


develric (approved):
Hey @planarvoid , code LGTM and it works as described. :shipit:

Issue #13149, opened 2020-10-15
Prepublishing nudges: Write unit tests for Categories


zwarm:
Write unit tests for 
Prepublishing Nudges 
- Select Categories
- Add Categories

Issue #12166, opened 2020-06-11
Prepublishing nudges: bottom sheet transition is very slow


mbshakti:
Just a quick note that this issue is related to bottom sheets specifically.

### Expected behavior
Expected the "fade in" transition to be faster on the bottom sheet

### Actual behavior
It currently takes some time to transition to the new bottom sheet, and the opacity seems to reduce very gradually, which seems a bit awkward. It's especially more pronounced when you tap on the "Tags" row.

### Steps to reproduce the behavior
1. Hit publish on a new post
2. See bottom sheet
3. Tap on any row in the bottom sheet

### Screencapture
[Link to video](https://drive.google.com/file/d/1PkwGMEpA6r3Dj085BUXoh5DqMQRwsfeB/view?usp=sharing)

##### Tested on [device], Android [version], WPAndroid [version]
Android version 10 using the build you shared with me https://55447-9306568-gh.circle-artifacts.com/0/Artifacts/WordPress-pr-12149-build-55447.apk

dangermattic:
Thanks for reporting! 👍

PR #13150, opened 2020-10-15
Issue/13066 ppn categories

### LINKED ISSUE #13066: Prepublishing Nudge: Add Categories to bottom sheet
Introduce categories to the PPN bottom sheet.

This feature will allow users to 
- Pick categories from existing categories 
- Add new categories

Screen Shot 2020-10-06 at 4 08 34 PM

zwarm:
Part of #13066 

This PR adds support for
1. Selecting categories for a post
2. Adding a new category

This PR is labeled WIP/do not merge. I am looking for a review before moving forward with the open issues (see list below).

The meat of the PR can be found in:
`PrepublishingAddCategoryFragment` and `PrepublishingAddCategoryViewModel`
- This pair is responsible for the adding of a new category 

`PrepublishingCategoriesFragment and PrepublishingCategoriesViewModel`
- This pair is responsible for selecting/unselecting categories and for kicking off the Add Category action.

`PrepublishingBottomSheetFragment` (preexisting) is the traffic cop and handles adding/removing fragment transactions to the bottom sheet container. This was updated to handle the AddCategory and Categories fragments. Each of the individual actions in the bottom sheet are standalone fragments and the behavior is identical for all.

I was on the fence with regards to suspendable functions.
- They are used for update/add/fetch server calls in the `UpdatePostCategoriesUseCase`, `AddCategoryUseCase`, and `GetCategoriesUseCase`. I followed existing architecture for calls to TaxonomyStore and EditPostRepository. I am open to suggestions.

It’s a rather larger PR, but I broke it up by commits.
f8a263b => layout files, animation files, strings, `AppComponent`/`ViewModelModule`, plugging into existing architecture
740ce31 => Changes to `PrepublishingBottomSheetFragmentPrepublishingHomeViewModel`, `PrepublishingViewModel`
7c5012d => Add category related
3d28bb4 => Select category related

Home|Select|Add
---|---|---
||

**This PR has open issues** and they are almost all related to BottomSheet behavior (scrolling the biggest one)
1. Showing a snackbar in a bottom sheet (using toast for now)
~~2. Recyclerview: scrolling within a bottom sheet (this also affects PTR)~~  ****Update**: Removed the PTR in commit cd27ec0 and scrolling works just fine. 
3. There is an existing [issue](https://github.com/wordpress-mobile/WordPress-Android/issues/12166) for the janky transitions between the bottom sheet fragments. I started on some new transition files and things were looking positive, but ultimately decided to push this issue off until all the bottom sheet scrolling issues are resolved.
4. There is a separate [issue](https://github.com/wordpress-mobile/WordPress-Android/issues/13149) for writing unit test 
I am open to any suggestions for solving the scrolling/recyclerview issues in a bottom sheet. 😉 

To test:
Prep
- Launch the app
- Select Blog posts
- Select Edit on a post

Categories are shown on the Categories row
- If the post has categories assigned, you will see them listed on the home view (figure 1 above). If not, you will see Not Set

Select Categories is shown when user taps on Categories
- Tap the categories row to launch the Select categories view (fig 2)

Add category is shown and user can save a new category
- Tap the Add New Category button on the bottom of the select categories view (fig 3)
- Enter a new category and tap the <- Back Arrow

Newly Added category is shown in the categories list
- Take note of the toast which will tell you if the action succeeded or failed. 
- Back on the select category view, the category you have added is visible and checked

Categories can be checked/unchecked 
- Select or unselect some categories

Categories are saved
- Tap the <- Back arrow to save your options

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
@@ -1,0 +1,56 @@
+package org.wordpress.android.ui.posts
+
+import dagger.Reusable
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.withContext
+import org.apache.commons.text.StringEscapeUtils
+import org.wordpress.android.fluxc.Dispatcher
+import org.wordpress.android.fluxc.generated.TaxonomyActionBuilder
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.model.TermModel
+import org.wordpress.android.fluxc.store.TaxonomyStore
+import org.wordpress.android.models.CategoryNode
+import org.wordpress.android.modules.IO_THREAD
+import javax.inject.Inject
+import javax.inject.Named
+
+@Reusable
+class GetCategoriesUseCase @Inject constructor(
+    private val taxonomyStore: TaxonomyStore,
+    private val dispatcher: Dispatcher,
+    @Named(IO_THREAD) private val ioDispatcher: CoroutineDispatcher
+) {
+    // todo: annmarie should these all be in suspendable functions?
+    fun getPostCategories(editPostRepository: EditPostRepository, siteModel: SiteModel): String? {
+        val post = editPostRepository.getPost() ?: return null
+        val categories: List<TermModel> = taxonomyStore.getCategoriesForPost(
+                post,
+                siteModel
+        )
+        return formatCategories(categories)
+    }
+
+    private fun formatCategories(categoryList: List<TermModel>): String? {
+        if (categoryList.isEmpty()) return null
+
+        val formattedCategories = categoryList.joinToString { it -> it.name }
+        return StringEscapeUtils.unescapeHtml4(formattedCategories)
+    }
+
+    fun getCategoryLevels(siteModel: SiteModel): ArrayList<CategoryNode> {
+        val rootCategory = CategoryNode.createCategoryTreeFromList(
+                getCategoriesForSite(siteModel)
+        )
+        return CategoryNode.getSortedListOfCategoriesFromRoot(rootCategory) ?: arrayListOf()
+    }
+
+    private fun getCategoriesForSite(siteModel: SiteModel): List<TermModel> {
+        return taxonomyStore.getCategoriesForSite(siteModel)
+    }
+
+    suspend fun fetchNewCategories(siteModel: SiteModel) {
+        withContext(ioDispatcher) {
+            dispatcher.dispatch(TaxonomyActionBuilder.newFetchCategoriesAction(siteModel))
+        }
+    }
+}
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
@@ -11,6 +11,7 @@ import org.wordpress.android.fluxc.model.TermModel
 import org.wordpress.android.fluxc.store.TaxonomyStore
 import org.wordpress.android.models.CategoryNode
 import org.wordpress.android.modules.IO_THREAD
+import java.util.ArrayList
 import javax.inject.Inject
 import javax.inject.Named
 
@@ -20,14 +21,32 @@ class GetCategoriesUseCase @Inject constructor(
     private val dispatcher: Dispatcher,
     @Named(IO_THREAD) private val ioDispatcher: CoroutineDispatcher
 ) {
-    // todo: annmarie should these all be in suspendable functions?
-    fun getPostCategories(editPostRepository: EditPostRepository, siteModel: SiteModel): String? {
+    fun getPostCategoriesString(
+        editPostRepository: EditPostRepository,
+        siteModel: SiteModel
+    ): String? {
         val post = editPostRepository.getPost() ?: return null
         val categories: List<TermModel> = taxonomyStore.getCategoriesForPost(
                 post,
                 siteModel
         )
         return formatCategories(categories)
+    }
+
+    fun getPostCategories(editPostRepository: EditPostRepository, siteModel: SiteModel) =
+            editPostRepository.getPost()?.categoryIdList ?: listOf()
+
+    fun getSiteCategories(siteModel: SiteModel): ArrayList<CategoryNode> {
+        val rootCategory = CategoryNode.createCategoryTreeFromList(
+                getCategoriesForSite(siteModel)
+        )
+        return CategoryNode.getSortedListOfCategoriesFromRoot(rootCategory) ?: arrayListOf()
+    }
+
+    suspend fun fetchSiteCategories(siteModel: SiteModel) {
+        withContext(ioDispatcher) {
+            dispatcher.dispatch(TaxonomyActionBuilder.newFetchCategoriesAction(siteModel))
+        }
     }
 
     private fun formatCategories(categoryList: List<TermModel>): String? {
@@ -37,20 +56,7 @@ class GetCategoriesUseCase @Inject constructor(
         return StringEscapeUtils.unescapeHtml4(formattedCategories)
     }
 
-    fun getCategoryLevels(siteModel: SiteModel): ArrayList<CategoryNode> {
-        val rootCategory = CategoryNode.createCategoryTreeFromList(
-                getCategoriesForSite(siteModel)
-        )
-        return CategoryNode.getSortedListOfCategoriesFromRoot(rootCategory) ?: arrayListOf()
-    }
-
     private fun getCategoriesForSite(siteModel: SiteModel): List<TermModel> {
         return taxonomyStore.getCategoriesForSite(siteModel)
-    }
-
-    suspend fun fetchNewCategories(siteModel: SiteModel) {
-        withContext(ioDispatcher) {
-            dispatcher.dispatch(TaxonomyActionBuilder.newFetchCategoriesAction(siteModel))
-        }
     }
 }
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
@@ -9,6 +9,7 @@ import org.wordpress.android.fluxc.model.TermModel
 import org.wordpress.android.fluxc.store.TaxonomyStore
 import org.wordpress.android.fluxc.utils.AppLogWrapper
 import org.wordpress.android.models.CategoryNode
+import org.wordpress.android.models.wrappers.CategoryNodeWrapper
 import org.wordpress.android.util.AppLog.T.PREPUBLISHING_NUDGES
 import java.util.ArrayList
 import javax.inject.Inject
@@ -17,7 +18,8 @@ import javax.inject.Inject
 class GetCategoriesUseCase @Inject constructor(
     private val taxonomyStore: TaxonomyStore,
     private val dispatcher: Dispatcher,
-    private val appLogWrapper: AppLogWrapper
+    private val appLogWrapper: AppLogWrapper,
+    private val categoryNodeWrapper: CategoryNodeWrapper
 ) {
     fun getPostCategoriesString(
         editPostRepository: EditPostRepository,
@@ -39,10 +41,10 @@ class GetCategoriesUseCase @Inject constructor(
             editPostRepository.getPost()?.categoryIdList ?: listOf()
 
     fun getSiteCategories(siteModel: SiteModel): ArrayList<CategoryNode> {
-        val rootCategory = CategoryNode.createCategoryTreeFromList(
+        val rootCategory = categoryNodeWrapper.createCategoryTreeFromList(
                 getCategoriesForSite(siteModel)
         )
-        return CategoryNode.getSortedListOfCategoriesFromRoot(rootCategory) ?: arrayListOf()
+        return categoryNodeWrapper.getSortedListOfCategoriesFromRoot(rootCategory)
     }
 
     fun fetchSiteCategories(siteModel: SiteModel) {
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
@@ -7,20 +7,27 @@ import org.wordpress.android.fluxc.generated.TaxonomyActionBuilder
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.TermModel
 import org.wordpress.android.fluxc.store.TaxonomyStore
+import org.wordpress.android.fluxc.utils.AppLogWrapper
 import org.wordpress.android.models.CategoryNode
+import org.wordpress.android.util.AppLog.T.PREPUBLISHING_NUDGES
 import java.util.ArrayList
 import javax.inject.Inject
 
 @Reusable
 class GetCategoriesUseCase @Inject constructor(
     private val taxonomyStore: TaxonomyStore,
-    private val dispatcher: Dispatcher
+    private val dispatcher: Dispatcher,
+    private val appLogWrapper: AppLogWrapper
 ) {
     fun getPostCategoriesString(
         editPostRepository: EditPostRepository,
         siteModel: SiteModel
     ): String? {
-        val post = editPostRepository.getPost() ?: return null
+        val post = editPostRepository.getPost()
+                if (post == null) {
+                    appLogWrapper.d(PREPUBLISHING_NUDGES, "Post is null in EditPostRepository")
+                    return null
+                }
         val categories: List<TermModel> = taxonomyStore.getCategoriesForPost(
                 post,
                 siteModel
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
@@ -22,11 +22,11 @@ class GetCategoriesUseCase @Inject constructor(
     fun getPostCategoriesString(
         editPostRepository: EditPostRepository,
         siteModel: SiteModel
-    ): String? {
+    ): String {
         val post = editPostRepository.getPost()
                 if (post == null) {
                     appLogWrapper.d(PREPUBLISHING_NUDGES, "Post is null in EditPostRepository")
-                    return null
+                    return ""
                 }
         val categories: List<TermModel> = taxonomyStore.getCategoriesForPost(
                 post,
@@ -49,8 +49,8 @@ class GetCategoriesUseCase @Inject constructor(
             dispatcher.dispatch(TaxonomyActionBuilder.newFetchCategoriesAction(siteModel))
     }
 
-    private fun formatCategories(categoryList: List<TermModel>): String? {
-        if (categoryList.isEmpty()) return null
+    private fun formatCategories(categoryList: List<TermModel>): String {
+        if (categoryList.isEmpty()) return ""
 
         val formattedCategories = categoryList.joinToString { it -> it.name }
         return StringEscapeUtils.unescapeHtml4(formattedCategories)
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
@@ -1,8 +1,6 @@
 package org.wordpress.android.ui.posts
 
 import dagger.Reusable
-import kotlinx.coroutines.CoroutineDispatcher
-import kotlinx.coroutines.withContext
 import org.apache.commons.text.StringEscapeUtils
 import org.wordpress.android.fluxc.Dispatcher
 import org.wordpress.android.fluxc.generated.TaxonomyActionBuilder
@@ -10,16 +8,13 @@ import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.TermModel
 import org.wordpress.android.fluxc.store.TaxonomyStore
 import org.wordpress.android.models.CategoryNode
-import org.wordpress.android.modules.IO_THREAD
 import java.util.ArrayList
 import javax.inject.Inject
-import javax.inject.Named
 
 @Reusable
 class GetCategoriesUseCase @Inject constructor(
     private val taxonomyStore: TaxonomyStore,
-    private val dispatcher: Dispatcher,
-    @Named(IO_THREAD) private val ioDispatcher: CoroutineDispatcher
+    private val dispatcher: Dispatcher
 ) {
     fun getPostCategoriesString(
         editPostRepository: EditPostRepository,
@@ -43,10 +38,8 @@ class GetCategoriesUseCase @Inject constructor(
         return CategoryNode.getSortedListOfCategoriesFromRoot(rootCategory) ?: arrayListOf()
     }
 
-    suspend fun fetchSiteCategories(siteModel: SiteModel) {
-        withContext(ioDispatcher) {
+    fun fetchSiteCategories(siteModel: SiteModel) {
             dispatcher.dispatch(TaxonomyActionBuilder.newFetchCategoriesAction(siteModel))
-        }
     }
 
     private fun formatCategories(categoryList: List<TermModel>): String? {
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
@@ -137,7 +137,7 @@ class PrepublishingAddCategoryFragment : Fragment(R.layout.prepublishing_add_cat
         val siteModel = requireArguments().getSerializable(WordPress.SITE) as SiteModel
         val retainedSelectedCategoryPosition = savedInstanceState?.getInt(
                 SELECTED_PARENT_CATEGORY_POSITION)
-        viewModel.start(siteModel, !needsRequestLayout, retainedSelectedCategoryPosition )
+        viewModel.start(siteModel, !needsRequestLayout, retainedSelectedCategoryPosition)
     }
 
     private fun startObserving() {
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
@@ -2,6 +2,7 @@ package org.wordpress.android.ui.posts
 
 import android.content.Context
 import android.os.Bundle
+import android.util.Log
 import android.view.View
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.Observer
@@ -56,6 +57,7 @@ class PrepublishingAddCategoryFragment : Fragment(R.layout.prepublishing_add_cat
         initBackButton()
         initCloseButton()
         initInputText()
+       // initSpinner()
         initViewModel()
         super.onViewCreated(view, savedInstanceState)
     }
@@ -80,6 +82,19 @@ class PrepublishingAddCategoryFragment : Fragment(R.layout.prepublishing_add_cat
         category_name.requestFocus()
         ActivityUtils.showKeyboard(category_name)
     }
+
+// private fun initSpinner() {
+// todo: annmarie - get back to this after Android Studio is working again
+// parent_category.onItemSelectedListener = object : AdapterView.OnItemSelectedListener {
+// override fun onNothingSelected(parent: AdapterView<*>?) {
+//  Log.i(javaClass.simpleName, "***=> implement nothing selected")
+//  }
+//
+// override fun onItemSelected(parent: AdapterView<*>?, view: View?, position: Int, id: Long) {
+// Log.i(javaClass.simpleName, "***=> implement item selected")
+// }
+// }
+// }
 
     private fun initViewModel() {
         viewModel = ViewModelProviders.of(this, viewModelFactory)
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
@@ -2,8 +2,8 @@ package org.wordpress.android.ui.posts
 
 import android.content.Context
 import android.os.Bundle
-import android.util.Log
 import android.view.View
+import android.widget.AdapterView
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.Observer
 import androidx.lifecycle.ViewModelProvider
@@ -28,6 +28,8 @@ class PrepublishingAddCategoryFragment : Fragment(R.layout.prepublishing_add_cat
     @Inject lateinit var viewModelFactory: ViewModelProvider.Factory
     private lateinit var viewModel: PrepublishingAddCategoryViewModel
     @Inject lateinit var uiHelpers: UiHelpers
+
+    var spinnerTouched: Boolean = false
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
@@ -57,8 +59,9 @@ class PrepublishingAddCategoryFragment : Fragment(R.layout.prepublishing_add_cat
         initBackButton()
         initCloseButton()
         initInputText()
-       // initSpinner()
-        initViewModel()
+        initAdapter()
+        initSpinner()
+        initViewModel(savedInstanceState)
         super.onViewCreated(view, savedInstanceState)
     }
 
@@ -83,23 +86,58 @@ class PrepublishingAddCategoryFragment : Fragment(R.layout.prepublishing_add_cat
         ActivityUtils.showKeyboard(category_name)
     }
 
-// private fun initSpinner() {
-// todo: annmarie - get back to this after Android Studio is working again
-// parent_category.onItemSelectedListener = object : AdapterView.OnItemSelectedListener {
-// override fun onNothingSelected(parent: AdapterView<*>?) {
-//  Log.i(javaClass.simpleName, "***=> implement nothing selected")
-//  }
-//
-// override fun onItemSelected(parent: AdapterView<*>?, view: View?, position: Int, id: Long) {
-// Log.i(javaClass.simpleName, "***=> implement item selected")
-// }
-// }
-// }
+    private fun initAdapter() {
+        val categoryLevels = arrayListOf<CategoryNode>()
+        categoryLevels.add(
+                0, CategoryNode(
+                0, 0,
+                getString(R.string.top_level_category_name)
+        ))
 
-    private fun initViewModel() {
+        val categoryAdapter = ParentCategorySpinnerAdapter(
+                activity,
+                R.layout.categories_row_parent,
+                categoryLevels
+        )
+        parent_category.adapter = categoryAdapter
+    }
+
+    private fun initSpinner() {
+        parent_category.setOnTouchListener { v, event ->
+            spinnerTouched = true
+            v.performClick()
+            false
+        }
+
+        parent_category.onItemSelectedListener = object : AdapterView.OnItemSelectedListener {
+            override fun onNothingSelected(parent: AdapterView<*>?) {
+            }
+
+            override fun onItemSelected(
+                parent: AdapterView<*>?,
+                view: View?,
+                position: Int,
+                id: Long
+            ) {
+                if (spinnerTouched) {
+                    viewModel.parentCategorySelected(position)
+                    spinnerTouched = false
+                }
+            }
+        }
+    }
+
+    private fun initViewModel(savedInstanceState: Bundle?) {
         viewModel = ViewModelProviders.of(this, viewModelFactory)
                 .get(PrepublishingAddCategoryViewModel::class.java)
+
         startObserving()
+
+        val needsRequestLayout = requireArguments().getBoolean(PrepublishingTagsFragment.NEEDS_REQUEST_LAYOUT)
+        val siteModel = requireArguments().getSerializable(WordPress.SITE) as SiteModel
+        val retainedSelectedCategoryPosition = savedInstanceState?.getInt(
+                SELECTED_PARENT_CATEGORY_POSITION)
+        viewModel.start(siteModel, !needsRequestLayout, retainedSelectedCategoryPosition )
     }
 
     private fun startObserving() {
@@ -129,30 +167,19 @@ class PrepublishingAddCategoryFragment : Fragment(R.layout.prepublishing_add_cat
                 when (uiState) {
                     is ContentUiState -> {
                         loadCategories(uiState.categories)
+                        if (uiState.selectedParentCategoryPosition != parent_category.selectedItemPosition) {
+                            parent_category.setSelection(uiState.selectedParentCategoryPosition)
+                        }
                     }
                 }
             with(uiHelpers) {
                 updateVisibility(close_button, uiState.closeButtonVisible)
             }
         })
-
-        val needsRequestLayout = requireArguments().getBoolean(PrepublishingTagsFragment.NEEDS_REQUEST_LAYOUT)
-        val siteModel = requireArguments().getSerializable(WordPress.SITE) as SiteModel
-        viewModel.start(siteModel, !needsRequestLayout)
     }
 
     private fun loadCategories(categoryLevels: ArrayList<CategoryNode>) {
-        categoryLevels.add(
-                0, CategoryNode(
-                0, 0,
-                getString(R.string.top_level_category_name)
-        ))
-        val categoryAdapter = ParentCategorySpinnerAdapter(
-                activity,
-                R.layout.categories_row_parent,
-                categoryLevels
-        )
-        parent_category.adapter = categoryAdapter
+        (parent_category.adapter as? ParentCategorySpinnerAdapter)?.updateItems(categoryLevels)
     }
 
     private fun SnackbarMessageHolder.showToast() {
@@ -163,9 +190,19 @@ class PrepublishingAddCategoryFragment : Fragment(R.layout.prepublishing_add_cat
         )
     }
 
+    override fun onSaveInstanceState(outState: Bundle) {
+        viewModel.uiState.value?.let { state ->
+            if (state is ContentUiState) {
+                outState.putInt(SELECTED_PARENT_CATEGORY_POSITION, state.selectedParentCategoryPosition)
+            }
+        }
+        super.onSaveInstanceState(outState)
+    }
+
     companion object {
         const val TAG = "prepublishing_add_category_fragment_tag"
         const val NEEDS_REQUEST_LAYOUT = "prepublishing_add_category_fragment_needs_request_layout"
+        const val SELECTED_PARENT_CATEGORY_POSITION = "selected_parent_category_position"
         @JvmStatic fun newInstance(
             site: SiteModel,
             needsRequestLayout: Boolean
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
@@ -1,0 +1,180 @@
+package org.wordpress.android.ui.posts
+
+import android.content.Context
+import android.os.Bundle
+import android.view.View
+import androidx.fragment.app.Fragment
+import androidx.lifecycle.Observer
+import androidx.lifecycle.ViewModelProvider
+import androidx.lifecycle.ViewModelProviders
+import com.google.android.material.snackbar.Snackbar
+import kotlinx.android.synthetic.main.add_category.*
+import kotlinx.android.synthetic.main.prepublishing_add_category_fragment.*
+import kotlinx.android.synthetic.main.prepublishing_toolbar.*
+import org.wordpress.android.R
+import org.wordpress.android.R.layout
+import org.wordpress.android.WordPress
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.models.CategoryNode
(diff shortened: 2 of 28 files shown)


WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
```diff
@@ -0,0 +1,62 @@
...
+class GetCategoriesUseCase @Inject constructor(
+    private val taxonomyStore: TaxonomyStore,
+    private val dispatcher: Dispatcher,
+    @Named(IO_THREAD) private val ioDispatcher: CoroutineDispatcher
+) {
+    fun getPostCategoriesString(
+        editPostRepository: EditPostRepository,
+        siteModel: SiteModel
+    ): String? {
+        val post = editPostRepository.getPost() ?: return null
```
malinajirka: We might want to log an exception - since I believe the null state is not expected, right?

WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    fun getPostCategories(editPostRepository: EditPostRepository, siteModel: SiteModel) =
+            editPostRepository.getPost()?.categoryIdList ?: listOf()
+
+    fun getSiteCategories(siteModel: SiteModel): ArrayList<CategoryNode> {
+        val rootCategory = CategoryNode.createCategoryTreeFromList(
+                getCategoriesForSite(siteModel)
+        )
+        return CategoryNode.getSortedListOfCategoriesFromRoot(rootCategory) ?: arrayListOf()
```
malinajirka: We might want to create a wrapper around CategoryNode, since the static access is not testable. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    suspend fun fetchSiteCategories(siteModel: SiteModel) {
+        withContext(ioDispatcher) {
```
malinajirka: The dispatch action is asynchronous. We might want to consider removing the "suspend and withContext". The context switching is most probably more expensive than the action itself. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
```diff
@@ -0,0 +1,62 @@
...
+            dispatcher.dispatch(TaxonomyActionBuilder.newFetchCategoriesAction(siteModel))
+        }
+    }
+
+    private fun formatCategories(categoryList: List<TermModel>): String? {
+        if (categoryList.isEmpty()) return null
```
malinajirka: I don't have the full context, but could we return an empty string here and change the return type to non-nullable string?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
```diff
@@ -0,0 +1,165 @@
...
+
+    override fun onDetach() {
+        super.onDetach()
+        closeListener = null
+    }
+
+    override fun onResume() {
+        val needsRequestLayout = requireArguments().getBoolean(NEEDS_REQUEST_LAYOUT)
+        if (needsRequestLayout) {
+            requireActivity().window.decorView.requestLayout()
```
malinajirka, line 53: Could you please elaborate why we need to manually request layout? We might want to consider adding a comment here. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
```diff
@@ -0,0 +1,165 @@
...
+        }
+        super.onResume()
+    }
+
+    companion object {
```
malinajirka: NitPick: We usually keep the companion object at the bottom of the file.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
```diff
@@ -0,0 +1,165 @@
...
+        initBackButton()
+        initCloseButton()
+        initInputText()
+        initViewModel()
+        super.onViewCreated(view, savedInstanceState)
+    }
+
+    private fun initBackButton() {
+        back_button.setOnClickListener {
+            viewModel.addCategory(
```
malinajirka, line 70: IMO the view layer shouldn't be responsible for such logic. It should just propagate user and system events to the VM. 

Ideally we'd just invoke `viewModel.onBackButtonClicked()` and the VM would do whatever is necessary. The VM should already know the state of editable fields. If it doesn't I guess we could propagate it as a parameter in "onBackButtonClicked" but it'd still feel a bit weird.

Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
```diff
@@ -0,0 +1,165 @@
...
+                    category_name.text.toString(),
+                    (parent_category.selectedItem as CategoryNode)
+            )
+            viewModel.onBackButtonClicked()
+        }
+    }
+
+    private fun initCloseButton() {
+        close_button.setOnClickListener {
+            viewModel.onBackButtonClicked()
```
malinajirka, line 80: I'd consider renaming "onBackButtonClicked" to something else - since we use it for both "back" and "close". Or we can introduce a new method for "onCloseButtonClicked". Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
```diff
@@ -0,0 +1,165 @@
...
+import android.os.Bundle
+import android.view.View
+import androidx.fragment.app.Fragment
+import androidx.lifecycle.Observer
+import androidx.lifecycle.ViewModelProvider
+import androidx.lifecycle.ViewModelProviders
+import kotlinx.android.synthetic.main.add_category.*
+import kotlinx.android.synthetic.main.prepublishing_toolbar.*
+import org.wordpress.android.R
+import org.wordpress.android.R.layout
```
malinajirka: Lint is having hard time with all "R.xyz" imports - it doesn't find them and complains after, that a resource is unused. I'd consider removing it. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
```diff
@@ -0,0 +1,165 @@
...
+                categoryLevels
+        )
+        parent_category.adapter = categoryAdapter
+    }
+
+    private fun SnackbarMessageHolder.showToast() {
+        val message = uiHelpers.getTextOfUiString(requireContext(), this.message).toString()
+        ToastUtils.showToast(
+                requireContext(), message,
+                LONG
```
malinajirka, line 189: Is there any particular reason why we want to go with "LONG" instead of "SHORT" here? AFAIK we use SHORT for most usecases in the app. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryViewModel.kt
```diff
@@ -0,0 +1,124 @@
+package org.wordpress.android.ui.posts
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MediatorLiveData
+import androidx.lifecycle.MutableLiveData
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
+import org.wordpress.android.R.string
```
malinajirka: Same as one of the comments above - I'd consider removing the ".string" part from the import.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
```diff
@@ -0,0 +1,223 @@
...
+    private lateinit var siteModel: SiteModel
+    private var updateCategoriesJob: Job? = null
+
+    private val _navigateToHomeScreen = MutableLiveData<Event<Unit>>()
+    val navigateToHomeScreen: LiveData<Event<Unit>> = _navigateToHomeScreen
+
+    private val _navigateToAddCategoryScreen = MutableLiveData<Event<Unit>>()
+    val navigateToAddCategoryScreen: LiveData<Event<Unit>> = _navigateToAddCategoryScreen
+
+    private val _toolbarTitleUiState = MutableLiveData<UiString>()
```
malinajirka, line 44: Out of curiosity - is there any reason why you prefer having multiple uiStates (toolbarUiState, refreshingUiState, uiState) over having a single uistate? Both approaches (single vs multiple) are valid and used in our codebase. 

I always prefer having a single UIState as it's extremely simple to make sure that an invalid state even cannot be created. It's also way easier to write tests for it since there are clearly defined combinations of possible states.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
```diff
@@ -0,0 +1,223 @@
...
+    val snackbarEvents: LiveData<Event<SnackbarMessageHolder>> = _snackbarEvents
+
+    private val _uiState: MutableLiveData<UiState> = MutableLiveData()
+    val uiState: LiveData<UiState> = _uiState
+
+    private val _refreshingUiState = MutableLiveData<SwipeToRefreshUiState>()
+    val refreshingUiState: LiveData<SwipeToRefreshUiState> = _refreshingUiState
+
+    fun start(editPostRepository: EditPostRepository, siteModel: SiteModel) {
+        this.editPostRepository = editPostRepository
```
malinajirka, line 54: Won't this throw an exception? Both editPostRepository and siteModel are lateinit vars, and if I recall correctly, those can't be re-assigned. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
```diff
@@ -0,0 +1,223 @@
...
+        val siteCategories = getSiteCategories()
+        val postCategories = getPostCategories()
+        _uiState.value = ContentUiState(
+                siteCategories = siteCategories,
+                selectedCategoryIds = postCategories.toHashSet()
+        )
+    }
+
+    private fun setToolbarTitleUiState() {
+        _toolbarTitleUiState.postValue(UiStringRes(string.prepublishing_nudges_toolbar_title_categories))
```
malinajirka, line 75: Just out of curiosity - any reason why you use "postValue" instead of "value = " here?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
```diff
@@ -0,0 +1,223 @@
...
+
+    override fun onCleared() {
+        super.onCleared()
+        updateCategoriesJob?.cancel()
+    }
+
+    fun updateCategories() {
+        if (!hasChanges()) return
+
+        if (!networkUtilsWrapper.isNetworkAvailable()) {
```
malinajirka, line 94: AFAIU we call this method from "backPressed" and that's the only path how to save the changes. When the user presses "back" in offline, won't we just throw away their changes?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
```diff
@@ -0,0 +1,223 @@
...
+        _navigateToHomeScreen.postValue(Event(Unit))
+    }
+
+    fun onAddNewCategoryClicked() {
+        _navigateToAddCategoryScreen.postValue(Event(Unit))
+    }
+
+    override fun onCleared() {
+        super.onCleared()
+        updateCategoriesJob?.cancel()
```
malinajirka, line 88: Just to double check: Do we really want to cancel the saving operation when the system destroys the VM?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
```diff
@@ -0,0 +1,223 @@
...
+            )
+            return
+        }
+
+        val categoryList = (uiState.value as? ContentUiState)?.selectedCategoryIds ?: return
+        updateCategoriesJob?.cancel()
+        updateCategoriesJob = launch(bgDispatcher) {
+            updatePostCategoriesUseCase.updateCategories(categoryList.toList(), editPostRepository)
+        }
+        trackCategoriesChangedEvent()
```
malinajirka, line 106: This line will be executed no matter if the updateCategoriesJob started/finished. Is this intentional? Don't we want to track the event after the categories are saved?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    fun addSelectedCategory(categoryId: Long) {
+        uiState.value.let { state ->
+            if (state is ContentUiState) {
+                state.selectedCategoryIds.add(categoryId)
```
malinajirka, line 112: Have you considered using an immutable collection? Using a mutable uiState intentionally feels a bit risky. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
```diff
@@ -0,0 +1,223 @@
...
+                _uiState.value = state.copy()
+            }
+        }
+    }
+
+    fun removeSelectedCategory(categoryId: Long) {
+        uiState.value.let { state ->
+            if (state is ContentUiState) {
+                state.selectedCategoryIds.add(categoryId)
```
malinajirka: Same as above.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryViewModel.kt
```diff
@@ -0,0 +1,124 @@
...
+
+    private val _uiState: MutableLiveData<UiState> = MutableLiveData()
+    val uiState: LiveData<UiState> = _uiState
+
+    fun start(
+        siteModel: SiteModel,
+        closeKeyboard: Boolean = false
+    ) {
+        this.closeKeyboard = closeKeyboard
+        this.siteModel = siteModel
```
malinajirka, line 59: Similar as above - we are re-assigning a lateinit var.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryViewModel.kt
```diff
@@ -0,0 +1,124 @@
...
+        super.onCleared()
+        addCategoryJob?.cancel()
+    }
+
+    private fun loadCategories() {
+        val newUiState = ContentUiState(categories = getCategoryLevels())
+        updateUiState(newUiState)
+    }
+
+    fun addCategory(categoryText: String, parentCategory: CategoryNode?) {
```
malinajirka, line 97: Similar as in the previous VM
- AFAIU we'll just throw away user changes in offline
- do we really want to cancel the saving operation in `onCleared` ?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryViewModel.kt
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    private fun getCategoryLevels(): ArrayList<CategoryNode> =
+            getCategoriesUseCase.getSiteCategories(siteModel)
+
+    private fun trackCategoryAddedEvent() {
+        analyticsTrackerWrapper.trackPrepublishingNudges(Stat.EDITOR_POST_CATEGORIES_ADDED)
+    }
+
+    private fun updateUiState(uiState: UiState) {
```
malinajirka, line 123: We might want to consider inlining this method - it's used just from a single place and it's body is pretty self-explanatory. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingBottomSheetFragment.kt
```diff
@@ -190,12 +210,10 @@ class PrepublishingBottomSheetFragment : WPBottomSheetDialogFragment(),
     private fun fadeInFragment(fragment: Fragment, tag: String) {
         childFragmentManager.let { fragmentManager ->
             val fragmentTransaction = fragmentManager.beginTransaction()
-            fragmentManager.findFragmentById(R.id.prepublishing_content_fragment)?.run {
```
malinajirka, line 193: Could you please elaborate why we are removing this line?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesAdapter.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    override fun getItemId(position: Int): Long {
+        return categoryNodeList[position].categoryId
+    }
+
+    override fun getItemCount(): Int {
+        return categoryNodeList.size
+    }
+
+    private fun isSameCategoryList(categoryList: List<CategoryNode>): Boolean {
```
malinajirka, line 56: Have you considered implementing DiffUtils instead of a custom setter?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesAdapter.kt
```diff
@@ -0,0 +1,116 @@
...
+    private fun containsNode(categoryNode: CategoryNode): Boolean {
+        categoryNodeList.forEach {
+            if (it.categoryId == categoryNode.categoryId) {
+                return true
+            }
+        }
+        return false
+    }
+
+    class CategoriesViewHolder(
```
malinajirka, line 82: We might want to consider extracting this class into a different file. I think it'd be consistent with the other VH. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesAdapter.kt
```diff
@@ -0,0 +1,116 @@
...
+                    horizontalPadding,
+                    verticalPadding
+            )
+
+            prepublishing_category_text.text = StringEscapeUtils.unescapeHtml4(
+                    row.name
+            )
+
+            prepublishing_category_check.isChecked = false
+            if (selectedCategoryIds.contains(row.categoryId)) {
```
malinajirka, line 112: Have you considered keeping the "check/not-checked" state within the UiState? Currently the view layer is responsible for checking if `selectedCategoryIds` contains the item which is about to be rendered. It feels a bit weird. Ideally the VM would create the UiStates and the only job of the VH would be to render the UI state. The UIState could even contain `vertical padding = R.dimen.margin_large`, `horizontalPadding = R.dimen.margin_extra_large` and `start padding multiplier = row.level`. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesFragment.kt
```diff
@@ -0,0 +1,239 @@
...
+    private fun initAddCategoryButton() {
+        add_new_category.setOnClickListener {
+            viewModel.onAddNewCategoryClicked()
+        }
+    }
+
+    private fun initRecyclerView() {
+        recycler_view.layoutManager = LinearLayoutManager(context, RecyclerView.VERTICAL, false)
+        recycler_view.adapter = PrepublishingCategoriesAdapter(
+                onCheckChangeListener = { id, checked ->
```
malinajirka, line 100: Same as the above comments - the view layers shouldn't be responsible for such logic. Have you considered storing the listener in the uiState of the items?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesFragment.kt
```diff
@@ -0,0 +1,239 @@
...
+        val message = uiHelpers.getTextOfUiString(requireContext(), this.message).toString()
+        ToastUtils.showToast(
+                requireContext(), message,
+                LONG
+        )
+    }
+
+    @SuppressWarnings("unused")
+    @Subscribe(threadMode = MAIN)
+    fun onTaxonomyChanged(event: OnTaxonomyChanged) {
```
malinajirka: Have you considered observing these changes in the VM layer not the view layer?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesFragment.kt
```diff
@@ -0,0 +1,239 @@
...
+        initAddCategoryButton()
+        initRecyclerView()
+        initSwipeToRefreshHelper()
+        initViewModel()
+        super.onViewCreated(view, savedInstanceState)
+    }
+
+    private fun initBackButton() {
+        back_button.setOnClickListener {
+            viewModel.updateCategories()
```
malinajirka, line 86: Same as for the other fragment -  we might want to consider handling this within the VM. The view layer shouldn't tell the VM what to do (eg. it shouldn't tell the VM to updateCategories). It's not testable. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesFragment.kt
```diff
@@ -0,0 +1,239 @@
...
+    private fun initViewModel() {
+        viewModel = ViewModelProviders.of(this, viewModelFactory)
+                .get(PrepublishingCategoriesViewModel::class.java)
+        startObserving()
+    }
+
+    private fun initSwipeToRefreshHelper() {
+        swipeToRefreshHelper = WPSwipeToRefreshHelper.buildSwipeToRefreshHelper(ptr_layout,
+                SwipeToRefreshHelper.RefreshListener {
+                    if (!NetworkUtils.checkConnection(requireContext())) {
```
malinajirka: Have you considered moving this logic into the VM?

WordPress/src/main/res/anim/prepublishing_fragment_fade_in.xml
```diff
@@ -1,10 +1,14 @@
...
-
     <alpha
+        android:duration="@android:integer/config_mediumAnimTime"
+        android:interpolator="@android:anim/decelerate_interpolator"
         android:fromAlpha="0"
-        android:toAlpha="1"
-        android:startOffset="200"
-        android:duration="200" />
+        android:toAlpha="1" />
+<!--<alpha-->
```
malinajirka: Is this change intentional - if it is, we might want to remove the commented out code. Wdyt?

WordPress/src/main/res/anim/prepublishing_fragment_fade_out.xml
```diff
@@ -1,10 +1,14 @@
...
+android:fillAfter="true" >
     <alpha
+        android:duration="@android:integer/config_mediumAnimTime"
+        android:interpolator="@android:anim/decelerate_interpolator"
         android:fromAlpha="1"
-        android:toAlpha="0"
-        android:startOffset="200"
-        android:duration="200" />
+        android:toAlpha="0" />
+<!--<alpha-->
```
malinajirka: Same as above.

malinajirka (commented):
Thanks @zwarm! 

I've done my best reviewing the PR, but it's huge. Please try to keep PRs focused. This PR could have been easily split into 4-6 single focused PRs and the review would be way easier.

It seems you've decided not to re-use any part of the existing "add category" screen from the Post settings. It was probably a good decision. However, I'm wondering -  have we considered using this new code even for that screen? 

> I am looking for a review before moving forward with the open issues (see list below).

I'd suggest fixing the issues in separate PRs. Wdyt?

> Recyclerview: scrolling within a bottom sheet (this also affects PTR)
>There is an existing issue for the janky transitions between the bottom sheet fragments. I started on some new transition files and things were looking positive, but ultimately decided to push this issue off until all the bottom sheet scrolling issues are resolved.

This is the cost of mis-using bottom sheets for something they weren't designed for :D. I'd suggest not providing PTR in bottom sheets. Wdyt?




Bugs:
1. The list items are not clickable - I need to click right on the check box.
2. I can't de-select categories and save the changes. When I de-select them and click on back, they are still there.


3. When I select/de-select a category and rotate my device, the change is lost.

4. When I set parent category and rotate my device, the change is lost.



5. When I enter a new category and go back -> it redirects me to the list of categories, but the new category is not there and after +-1s it redirects me back to "add new category screen" when I click back again, the category appears in the list.




6. (@mbshakti) When I clicked on the close icon on the "add category screen" I expected the whole bottom sheet would get closed. If we want to treat it as "discard" I'd consider a different icon/text. Wdyt?

(Emulator Pixel 3a, API 29)

zwarm:
> I've done my best reviewing the PR, but it's huge. Please try to keep PRs focused. This PR could have been easily split into 4-6 single focused PRs and the review would be way easier.

Yeah, I don't know what I was thinking. I committed on the branch to switch context earlier in the week and from there it went down hill fast. I'll switch this merge from develop to a feature branch. It will be easier to track.

> It seems you've decided not to re-use any part of the existing "add category" screen from the Post settings. It was probably a good decision. However, I'm wondering - have we considered using this new code even for that screen?

Yes and yes. I did lift layout files from old code, but didn't want to lock us into listview. My next hack week project is going to be refactoring the existing Post Settings categories and bring it in line with this. 
 
> I'd suggest fixing the issues in separate PRs. Wdyt?
Yes. I'll change the branch name to merge this PR into and go from there.

> This is the cost of mis-using bottom sheets for something they weren't designed for :D. I'd suggest not providing PTR in bottom sheets. Wdyt?
 I agree with you 100%. Even recycler views are crappy in bottom sheets. And since each fragment is not it's own bottom sheet, I'll need to add hooks back to the main bottom sheet fragment for any "on-the-fly" behavior changes. I have some thoughts about making the 'Add New Category' a static bottom component. I'll spend some time on that today. Goodbye PTR!

> Bugs:
> 1. The list items are not clickable - I need to click right on the check box.

I did not make them clickable, but I will.

> 2. I can't de-select categories and save the changes. When I de-select them and click on back, they are still there.
> 3. When I select/de-select a category and rotate my device, the change is lost.

Thanks, I thought I got these all

> 4. When I set parent category and rotate my device, the change is lost.

Will take another look at this one

> 1. When I enter a new category and go back -> it redirects me to the list of categories, but the new category is not there and after +-1s it redirects me back to "add new category screen" when I click back again, the category appears in the list.

That one is odd, I see the new category fairly quickly.

@malinajirka  Thanks for the set of eyes

WordPress/src/main/res/anim/prepublishing_fragment_fade_out.xml
zwarm: In 89a9945

WordPress/src/main/res/anim/prepublishing_fragment_fade_in.xml
zwarm: In 89a9945

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesFragment.kt
zwarm: Going to remove PTR all together because it doesn't belong in a bottom sheet. 👍

WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
zwarm: Makes sense in 3d8941d.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
zwarm: Will add comment

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
zwarm: I switched to Long because the message doesn't show in the view in which the action was submitted. I wanted the user to have more time to notice if there was an issue.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesAdapter.kt
zwarm: Yes, it's on my short list, but didn't transition it over before submitting 1st review.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingBottomSheetFragment.kt
zwarm: This was a fantastic catch. I removed it by accident because I was rushing to swap out the transitions at the last moment.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesAdapter.kt
zwarm: 🤔  this is a good suggestion. I like it.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
zwarm: This was introduced when Stories were added. Specific to PPN it's so the RecyclerView is re-calculated and visible when coming from stories. The functionality was added to AddCategory and Category fragments to maintain usability for coming from stories.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
zwarm: in 9216cca

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
zwarm: Comment added in 9216cca

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingBottomSheetFragment.kt
zwarm: In 646567b

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
zwarm: Fixed in 646567b

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryViewModel.kt
zwarm: Fixed in 646567b

WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
zwarm: Exception is now logged in c42360b

WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
zwarm: Returning empty string in 3055751

WordPress/src/main/java/org/wordpress/android/ui/posts/GetCategoriesUseCase.kt
zwarm: Wrapper created in c389ca3

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
zwarm: I lifted the logic for the toolbar/navigate from the existing PPN fragments/viewmodels and in the interest of time, used them as they were written. I'll jots this down as a nice-to-change list when I work on some overall architecture changes. Not part of this PR

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
zwarm: Same comment as above: I lifted the logic for the toolbar/navigate from the existing PPN fragments/viewmodels and in the interest of time, used them as they were written.  Noted in the issue #13066.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
zwarm: I followed the strategy in the other PPN actions figuring I should match what is already being done. If we change here, should we change the other PPN actions to match? @malinajirka

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryViewModel.kt
zwarm: Same as above, I followed the strategy in the other PPN actions figuring I should match what is already being done. If we change here, should we change the other PPN actions to match?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
zwarm: The close button is being removed and a true Submit button will be added. See issue ticket for details #13066

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
zwarm: Moved to #13066  ticket to track as a separate item outside of this PR

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
zwarm: Moved to #13066  ticket to track as a separate item outside of this PR

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
zwarm: Ugh, I suppose I can move it to the home fragment since we don't wait around for success or failure. The same would be true for add category - the capture would have to occur in the select logic. Bleh. Noted in the issue #13066.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
zwarm: Noted in the issue #13066 under arch changes.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
zwarm: Noted in the issue #13066 under arch changes.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryViewModel.kt
zwarm: Noted in the issue #13066 under arch changes.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
zwarm: Noted in #13066 under arch changes.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
zwarm: I'm not sure what the desired behavior is and interested in what your thoughts are? I'm going to check how the setting work, maybe that will provide a hint. The PPN tags updates each keypress (with a delay), so that doesn't happen on back button. I didn't want to do that here because we might overload categories. wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryViewModel.kt
zwarm: Noted in the issue #13066 under arch changes, but it might be moot if the tracking location changes to the select categories view.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesAdapter.kt
zwarm: Noted in the issue #13066 under arch changes.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesAdapter.kt
zwarm: Noted in the issue #13066 under arch change

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesFragment.kt
zwarm: Noted in the issue #13066 under arch change

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesFragment.kt
zwarm: Noted in the issue #13066 under arch change

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesFragment.kt
zwarm: Separate task in the issue #13066.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesFragment.kt
zwarm: Noted in the issue #13066 under arch change

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
zwarm: Actually @malinajirka I think those are fine - hence the use of the `this` keyword. They are being initialized not reassigned. I've seen this pattern in [Interests](https://github.com/wordpress-mobile/WordPress-Android/blob/2476981600502ddb9a8c73207e937fc1edb48707/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/interests/ReaderInterestsViewModel.kt#L53), [Stories](https://github.com/wordpress-mobile/WordPress-Android/blob/2476981600502ddb9a8c73207e937fc1edb48707/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt#L64), [PostList](https://github.com/wordpress-mobile/WordPress-Android/blob/2476981600502ddb9a8c73207e937fc1edb48707/WordPress/src/main/java/org/wordpress/android/ui/posts/PostListMainViewModel.kt#L227), [Tags](https://github.com/wordpress-mobile/WordPress-Android/blob/2476981600502ddb9a8c73207e937fc1edb48707/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingTagsViewModel.kt#L39)

I mean, we could all be wrong 🤣  The other option is setting them var xxx = null? Thoughts?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesFragment.kt
zwarm: in commit cd27ec0
After I removed the SwipeRefreshLayout, miraculously the scrolling of the recyclerview works like a charm. 🥳

zwarm:
> 1. The list items are not clickable - I need to click right on the check box.

I made the rows clickable in 09250ff

> 2. I can't de-select categories and save the changes. When I de-select them and click on back, they are still there.

Fixed in 09250ff

> 3. When I select/de-select a category and rotate my device, the change is lost.

Fixed in 09250ff

> 4. When I set parent category and rotate my device, the change is lost.

Fixed in 701759b

> 5. When I enter a new category and go back -> it redirects me to the list of categories, but the new category is not there and after +-1s it redirects me back to "add new category screen" when I click back again, the category appears in the list.

Fixed in 040f65b

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
malinajirka: Isn't the `this` keyword used because of the name collision - both the parameter and the field are called `editPostRepository`?

> They are being initialized not reassigned.

But start() can be called multiple times, right? So the first call is "initialization" but any subsequent calls are "reassignments".

Either way, I've created a unit test and it seems "lateinit vars" can be re-assigned. Which kind of make sense, otherwise, they'd be called "lateinit val". So we are good - even though it's probably not a great practice since we are opening doors for race conditions. Having said that it's completely fine to use it here - we already use it at other places as you mentioned.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
malinajirka: It works the same way in the settings, so I think we are good. Having said that, I think it's not a great UX - we probably should display a dialog and let the user confirm that they are ok loosing their changes. Either way, it's out of scope.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
malinajirka: If it's consistent with the rest I think it's okay to keep it - we'd ideally update all the actions, but it's above the scope. Thanks for the clarification!

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesFragment.kt
malinajirka: Which of the items in the issue is referring to this comment?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
malinajirka: Could you please double check this is in the list? I can't find it there. I think we use it in some others VMs too, but it might bite us.

malinajirka (approved):
Thanks @zwarm! Thanks for the updates, great job ;)! 

I've found just one issue and I don't think it's vital.
1. Select some categories
2. Click on "Add new category"
3. Either press back right way or type something and press back
4. Notice the categories selected in step `1` are not selected.

I think this will get fixed when we include "check/unchecked" state in the uiState of each item. Wdyt?

Feel free to merge this PR. I left some comments, but they are mostly clarification questions, not blockers.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
zwarm: Ah, yes, I forgot about `start` being called over and over again. It does seem like a poor pattern to rely on, I'll drop it on my list for hack week when I consolidate editPostSettings in. Thanks

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
zwarm: > I think it's not a great UX

I'm with you on this ... no notifications success/failure ... I get it, but we are all over the place in the app and we probably won't come to a standard. 🤷

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
zwarm: FYI: I pulled it out in my refactors :)

zwarm:
> I've found just one issue and I don't think it's vital.
> I think this will get fixed when we include "check/unchecked" state in the uiState of each item. Wdyt?

Yep, cleaned up in the refactor code. I'm going to merge and get to a stabler place for further reviews. Thanks so much @malinajirka for the good feedback.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
malinajirka: Awesome, thanks!! ❤️

mbshakti:
From a design POV, looks good to me. Thanks @zwarm !

PR #13159, opened 2020-10-16
[Modal Layout Picker] Retry view

antonis:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2740

## Description
Adds a retry view to handle the case where the loading of layouts fails and there are no cached data

## To test:

1. Fresh install the app or clear storage (Device Settings>Apps>WordPress Beta>Storage & cache>Clear storage)
1. Login
1. **Disable the internet** connection (e.g. flight mode)
1. Open the Modal Layout Picker 
1. Verify that you see the **error/retry screen** (check screenshot below)
1. Press the *Retry* button
1. Verify that nothing happens
1. **Enable the internet** connection
1. Press the **Retry** button
1. Verify that the layouts did load

## Screenshot

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mlp/ModalLayoutPickerFragment.kt
```diff
@@ -97,6 +96,10 @@ class ModalLayoutPickerFragment : BottomSheetDialogFragment() {
         previewButton.setOnClickListener {
             viewModel.onPreviewPageClicked()
         }
+        actionableEmptyView.button.setOnClickListener {
+            errorLayout.setVisible(false)
```
malinajirka: We might want to consider calling "viewModel.onRetryClicked()" and updating the uiState (errorLayout visibility) from there. So the UiState is in sync with the UI and the VM layer is responsible for building it. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/mlp/ModalLayoutPickerFragment.kt
```diff
@@ -174,19 +177,19 @@ class ModalLayoutPickerFragment : BottomSheetDialogFragment() {
             when (uiState) {
                 is LoadingUiState -> {
                     setTitleVisibility(uiState.isHeaderVisible)
```
malinajirka: We might want to consider pulling "setTitleVisibility", "setContentVisibility" out of the `when`. Since it's used in all the cases. Wdyt?

Note: I usually try to build the uiState so the only thing that is in the `when` is content of that specific layout. So for example, layout visibility needs to be set to true/false for all the states so it wouldn't be in the `when`. Whereas text of a retry button on error layout needs to be set only when the error layout is visible so it'd be in the `when`.

malinajirka (commented):
Thanks @antonis!

I've left two minor comments. I also encountered an issue on a small device. The "retry" button doesn't fit the screen and the layout is not scrollable.

WordPress/src/main/java/org/wordpress/android/ui/mlp/ModalLayoutPickerFragment.kt
antonis: I agree. Fixed with 3239d4a0c955ccd0f8d80803b7502f0679eaed75

WordPress/src/main/java/org/wordpress/android/ui/mlp/ModalLayoutPickerFragment.kt
antonis: I agree. Fixed with 3239d4a0c955ccd0f8d80803b7502f0679eaed75

antonis:
Hey @malinajirka 👋 
Thank you once again for the prompt feedback 🙇 

I implemented your suggestions along the [design feedback](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2563#issuecomment-712297796) with 3239d4a0c955ccd0f8d80803b7502f0679eaed75

> The "retry" button doesn't fit the screen and the layout is not scrollable.

The [new design](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2563#issuecomment-712703544) moves the retry button to the bottom toolbar where it should be accessible in smaller devices. It also removes the large headers in error state making it more suitable for small devices.

malinajirka (approved):
LGTM, thanks

PR #13160, opened 2020-10-16
Merge release 15.9 into develop

loremattei:
Merge release 15.9 into develop

PR submission checklist:


AliSoftware (commented):
Haven't yet received my Android physical device (should only arrive tomorrow) so was only able to smoke-test by installing the apk in an emulator on my Mac.

Wasn't really sure what to smoke test exactly, so basically what I tried:
 - I logged in successfully
 - Opened my AT test website
 - Navigated to all the tabs (My Site, Reader, Discover) and to each section 1 level deeper (each entry in My Site, each top tab in Reader and Notifications) as well as in Me screens
 - Created a new post
 - Edited a "normal" post
 - Edited a post containing classic blocks, trying to reproduce the gutenberg-mobile issue 2695 mentioned in the RELEASE-NOTES.txt in the process
 - Did the same with my non-AT site test site

All that worked without any issue.

I don't feel that's a super large smoke-test coverage though and most importantly given that I could only test on emulator and not on real device, I'm not sure I'm confident enough to be able to put my Approve stamp without anyone else smoke-testing it too?

WordPress/build.gradle
```diff
@@ -92,9 +92,9 @@ android {
             dimension "buildType"
             // Only set the release version if one isn't provided
             if (!project.hasProperty("versionName")) {
-                versionName "15.9-rc-2"
+                versionName "15.9"
```
AliSoftware, line 95: Side note: do we already have a Peril rule to check that we don't commit `rc` versions on release PRs? (Just wondering if we're supposed to double-check that manually on release PRs or if Peril already takes care of it)

loremattei: We don't have a Peril rule for this. We also merge back `rc`s after every beta so that changes land there quickly. We can think about how to create a Peril rule.

PR #13167, opened 2020-10-18
Issue/13066 ppn add category button

zwarm:
Part of #13066 

This PR removes the Close "X" click button on the Add Category toolbar and replaces it with a Submit button. The `PrepublishingAddCategoryViewModel` and `PrepublishingAddCategoryFragment` have been refactored based on PR comments in #13150 

***Merge Instructions***
- Note: This PR will be merged into a feature branch
- Make sure #13150 has been merged
- Remove the Not Ready For Merge label
- Merge as normal

Before|After
---|---
|

To test:

- Launch the app
- Select Blog posts
- Select Edit on a post
- From the edit post view, tap the UPDATE action to launch the bottom sheet
- From the select categories view, tap the Add New Category link on the bottom of the select categories view 
- Enter a new category and tap the submit button
- Newly Added category is shown in the categories list

Take note of the toast which will tell you if the action succeeded or failed.

PR submission checklist:

- [X] I have considered adding unit tests where possible.
- [X] I have considered adding accessibility improvements for my changes.
- [X] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13168, opened 2020-10-19
Issue/13066 ppn category select

zwarm:
Part of #13066 

This PR is a refactor of `PrepublishingCategoriesViewModel`, PrepublishingCategoriesAdapter  and to a lesser extent `PrepublishingCategoriesFragment`. It addresses a bunch of architectural recommendations in PR #13150. 

in `PrepublishingCategoriesViewModel` `PrepublishingCategoriesListItemUiState` was added to maintain consistency.  The SwipeRefreshLayout was removed and that cleared up the scrolling issue. In `PrepublishingCategoriesAdapter` the ViewHolder was extracted out to a separate class file and binding is done with a uiState object. 

**Merge Instructions**
Note: This PR will be merged into a feature branch
Make sure #13150 has been merged
Make sure #13167 has been merged
Remove the Not Ready For Merge label
Merge as normal

**To test**
- Launch the app
- Select Blog posts
- Select Edit on a post
- Categories are shown on the Categories row
(If the post has categories assigned, you will see them listed on the home view. If not, you will see Not Set)
- Tap the categories row to launch the Select categories view
- Tap the <- Back arrow to save your options

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesAdapter.kt
```diff
@@ -1,121 +1,77 @@
...
+    override fun getItemCount(): Int = items.size
 
-        categoryNodeList.forEach {
-            if (!containsNode(it)) {
-                return false
-            }
-        }
-        return true
+    override fun getItemId(position: Int): Long {
+        return position.toLong()
```
malinajirka, line 25: Is it safe to use position as id? Won't the id/position change when you insert a new item in the list? 

I guess it could be safe as long as we use "hasStableIds = false", just asking to be sure.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesAdapter.kt
```diff
@@ -1,121 +1,77 @@
...
     }
 
-    private fun isSameSelectedList(selectedCategoryIdList: HashSet<Long>): Boolean {
-        return selectedCategoryIdList == selectedCategoryIds
+    override fun getItemViewType(position: Int): Int {
+        return position
```
malinajirka: AFAIU this means that each item has it's own viewType - is this really what we want?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesAdapter.kt
```diff
@@ -1,121 +1,77 @@
...
-            ViewCompat.setPaddingRelative(
-                    prepublishing_category_text,
-                    horizontalPadding * row.level,
-                    verticalPadding,
-                    horizontalPadding,
-                    verticalPadding
-            )
+            return ((oldItem).categoryNode.categoryId ==
+                    (newItem).categoryNode.categoryId) &&
+                    (oldItem.checked == newItem.checked)
```
malinajirka: Could you please elaborate why we want to compare "checked/unchecked" in "areItemsTheSame"? It feels like a comparison suitable for "areContentsTheSame". Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewHolder.kt
```diff
@@ -0,0 +1,63 @@
...
+                onCategorySelected?.invoke(adapterPosition)
+            }
+            categoryCheckBox.setOnClickListener {
+                onCategorySelected?.invoke(adapterPosition)
+            }
+        }
+
+        override fun onBind(uiState: PrepublishingCategoriesListItemUiState) {
+            onCategorySelected = requireNotNull(uiState.onItemTapped) { "OnItemTapped is required" }
+            categoryName.text = StringEscapeUtils.unescapeHtml4(uiState.categoryNode.name)
```
malinajirka, line 44: We might want to move the call to "unescapeHtml4" to the VM. The VH should just set values to views and propagate events back to the VM, nothing more. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
```diff
@@ -98,33 +154,34 @@ class PrepublishingCategoriesViewModel @Inject constructor(
...
+    private fun getSiteCategories() =
+            getCategoriesUseCase.getSiteCategories(siteModel)
+
+    private fun getPostCategories() =
+            getCategoriesUseCase.getPostCategories(editPostRepository, siteModel)
+
+    private fun hasChanges(): Boolean {
+        val postCategories = getPostCategories()
+        val stateSelectedCategories = getSiteCategories()
+        return (stateSelectedCategories != postCategories)
```
malinajirka, line 196: It seems we are comparing two different data types + the comparison is order dependant, is it intentional?

WordPress/src/main/java/org/wordpress/android/ui/posts/UpdatePostCategoriesUseCase.kt
```diff
@@ -1,14 +1,26 @@
...
+class UpdatePostCategoriesUseCase @Inject constructor(
+    private val analyticsTrackerWrapper: AnalyticsTrackerWrapper
+) {
     fun updateCategories(categoryList: List<Long>, editPostRepository: EditPostRepository) {
         editPostRepository.updateAsync({ postModel ->
             postModel.setCategoryIdList(categoryList)
             true
+        }, { postModel: PostImmutableModel?, result: UpdatePostResult ->
+            if (result == Updated) {
+                analyticsTrackerWrapper.trackPrepublishingNudges(Stat.EDITOR_POST_CATEGORIES_ADDED)
```
malinajirka: This tracking makes this usecase not-reusable, wdyt?

malinajirka (commented):
Thanks for the improvements @zwarm! I've left some comment, let me know what you think.

WordPress/src/main/java/org/wordpress/android/ui/posts/UpdatePostCategoriesUseCase.kt
zwarm: > This tracking makes this usecase not-reusable, wdyt?

You are correct, but it was the lesser of two evils. The save happens on <- back button press and no one is listening for success/failure. Now that I am no longer rushing for this release, I have some time to figure out how to get the notification from the closed view to the previous.

malinajirka: I'd consider inlining the usecase. It doesn't do much anyway, wdyt?

zwarm: Easier to mock the useCase for testing, but let me check out inlining and see what happens. Thanks for the suggestion.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewHolder.kt
zwarm: > unescapeHtml4" to the VM

I kept it here because my model is the categoryNode and I didn't want to add duplicate properties. I'll look into adjusting it. I'm going to leave it as is for now though.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesAdapter.kt
zwarm: > I guess it could be safe as long as we use "hasStableIds = false"

That is correct, although it works for both right now, but it is always better to go with false if getItemId is using postion.toLong. in a983929

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
zwarm: Nope, that was "you've been at the computer for 10 hours on a Sunday" ... and that's why we have code reviews. Thanks! Fixed in a983929

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesAdapter.kt
zwarm: It does belong in `areContentsTheSame` - changed in a983929

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesAdapter.kt
zwarm: > AFAIU this means that each item has it's own viewType - is this really what we want?

Nope, this is not what I wanted. Just a left over toying with another idea. Gone in a983929

WordPress/src/main/java/org/wordpress/android/ui/posts/UpdatePostCategoriesUseCase.kt
zwarm: Moved inline in 0d17dc5 👍

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesAdapter.kt
malinajirka: > That is correct, although it works for both right now

If it works for both, I guess I don't understand the "hasStableIds" correctly. It shouldn't work when an item changes its id when you change its position, right? I mean it can work in some cases but won't work in other cases when the adapter uses optimization techniques when re-arranging items. Wdyt?

Either way, keeping it as `false` seems like a solution. Thanks

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewHolder.kt
malinajirka: Oh I see, make sense. Ideally, we wouldn't be re-using real models as uistates, but we do that all over the place and it might be an over optimization for this case. So feel free to leave it.

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesViewModel.kt
malinajirka: :D

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesAdapter.kt
```diff
@@ -1,121 +1,68 @@
...
-            prepublishing_category_check.isChecked = false
-            if (selectedCategoryIds.contains(row.categoryId)) {
-                prepublishing_category_check.isChecked = true
-            }
-
-            prepublishing_category_check.setOnCheckedChangeListener { _, isChecked ->
-                onCheckedChangeClickListener.invoke(row.categoryId, isChecked)
-            }
+        override fun areContentsTheSame(oldItemPosition: Int, newItemPosition: Int): Boolean {
+            return oldItems[oldItemPosition] == newItems[newItemPosition] &&
```
malinajirka: I think we don't need to compare "checked" state manually since we already compare if the two data classes are equal, wdyt?

I missed this yesterday, sorry. I think the comparison might not work in some cases. The data class has `var onItemTapped: ((Int) -> Unit)?` as one of the parameters of the constructor - so the equals method will check equality of the lambda. However, two lambdas are not equal even when they perform the exact same actions. I think we need to move the lambda out of the constructor and make it a regular field of the  `PrepublishingCategoriesListItemUiState`. Wdyt?

malinajirka (commented):
The changes look great overall. I found just one potential issue in the code. 

I also noticed [this issue](https://github.com/wordpress-mobile/WordPress-Android/pull/13167#pullrequestreview-511620150) still doesn't work. It behaves a bit differently than yesterday (the existing category gets selected and no new category gets added).

WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingCategoriesAdapter.kt
zwarm: > the existing category gets selected and no new category gets added).

I have not been able to reproduce this, please let me know what device & OS you are testing on. Thanks.



>  I think the comparison might not work in some cases 

Okay, so I ran a bunch of tests and it is working as expected. I believe it has to do with using `copy`, but I am going to do more test too. I did make two changes (1) Removed the isChecked comparison (2) changed the onItemTapped to a val from a var. 

Thanks to your video, I caught an issue with the dropdownlist, I fixed that in the same commit.

ashiagr (approved):
Thanks @zwarm. I tested mentioned issues and could not reproduce in the latest changes. Also confirm that "check -> uncheck -> uncheck ignored" issue that we discussed is resolved.

Good job 👍. Merging it to feature branch.

PR #13171, opened 2020-10-19
Add FLAG_UPDATE_CURRENT flag to notification pending intent

### LINKED ISSUE #8363: Rare NPE crash in ActivityLogListFragment

planarvoid:
Fixes #8363

It seems that this issue happens because the pending intent sometimes doesn't pass the extras. It seems that the system (because of efficiency) reuses the previous pending intent without the extras ([the issue](https://stackoverflow.com/questions/18037991/pendingintent-does-not-send-intent-extras)). 

The suggested fix seems to be to use the `PendingIntent.FLAG_UPDATE_CURRENT` flag when creating the pending intent. 

The second fix in this PR is a proper error message when the `SiteModel` is null in the `PostsListActivity`

To test:
- I don't know how to reproduce this issue

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
@@ -496,9 +496,7 @@ public class ActivityLauncher {
             ToastUtils.showToast(context, R.string.posts_cannot_be_started, ToastUtils.Duration.SHORT);
             return;
         }
-        Intent intent = new Intent(context, PostsListActivity.class);
-        intent.putExtra(WordPress.SITE, site);
-        context.startActivity(intent);
+        context.startActivity(PostsListActivity.buildIntent(context, site));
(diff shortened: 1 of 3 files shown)


mchowning (approved):
Looks good to me. I tried to recreate the crash, but I wasn't able to either. Your fix makes sense and seems to be working well in my testing. 

I'm very curious to see if this does fix the crash since this seems like a particularly nasty bit of behavior by Android.

planarvoid:
If it doesn't work, we can look into replacing `PendingIntent.FLAG_ONE_SHOT | PendingIntent.FLAG_UPDATE_CURRENT` with just with `PendingIntent.FLAG_UPDATE_CURRENT` and see.

PR #13174, opened 2020-10-19
Mobile stories block - part 5: empty block editing

mzorz:
This PR builds on top of #13103

This adds the ability to edit an empty story block just added via the block picker, as shown in the gif below.

emptyblockediting

To test:
1. create a post (either regular or a story post)
2. open in in gutenberg
3. add a Story block
4. tap on the empty Story block to trigger the StoryComposer camera capture mode to add a frist slide to your Story.
5. when one, tap on the arrow button to the top-right
6. observe the story slides get saved in the block.

Related PRs:
Gutenberg-mobile https://github.com/wordpress-mobile/gutenberg-mobile/pull/2747
Jetpack https://github.com/Automattic/jetpack/pull/17539

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
```diff
@@ -480,7 +480,12 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
             val blockId = intent.extras?.getString(ARG_STORY_BLOCK_ID)
             savedContentIntent.putExtra(ARG_STORY_BLOCK_ID, blockId)
 
-            val storyIndex = intent.getIntExtra(KEY_STORY_INDEX, DEFAULT_NONE_SELECTED)
+            // check if story index has been passed through intent
+            var storyIndex = intent.getIntExtra(KEY_STORY_INDEX, DEFAULT_NONE_SELECTED)
+            if (storyIndex == DEFAULT_NONE_SELECTED) {
+                // if not, let's use the current Story
+                storyIndex = storyRepositoryWrapper.getCurrentStoryIndex()
```
jd-alexander, line 487: @mzorz is it redundant to set this since it's `DEFAULT_NONE_SELECTED ` as well? Let me know if I am missing something. Thanks 🙏

jd-alexander (approved):
Thanks for these changes @mzorz works as expected! I had an emulator mishap that occurred due to the operation in progress toast message shown when the camera resource is behaving wonky. I ended up seeing the message below even after saving the post and reopening. it's probably expected since the camera probably created something incorrect file or something. Sharing nonetheless.

<kbd></kbd>

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
mzorz: Good catch! it actually is not redundant, because it may or may be not `DEFAULT_NONE_SELECTED`. (it may happen the value _is actually_ set in extras as -1, for example). Si yeah we need to check for it 👍

jd-alexander: Okay understood! I get what you are saying now because the currentStoryIndex is definitely not always the default as it starts out since it's within a singleton. thanks for expounding 🙏

PR #13181, opened 2020-10-20
Media and text from WP lib picker

develric:
This PR is based on findings while final testing the #13131

- Checking the expected extra when picking from wp media lib for media and text block
- Filtering media with null url when picking from wp media lib seems correct when picking an image/video. This condition can happen for example when a gif upload doesn't complete correctly (in this case the upload is done in 2 phases, phase 1 creates a scaffolding entry in the fluxc MediaModel table, phase 2 completes the upload and populate the table entry).

### Media & Text block from WP lib 
- Go to GB editor and add a Media&Text block -> Add image or video
- Select Choose from media library
- Select an image/video
- Check the image/video uploads correctly in the GB block

PR submission checklist:


planarvoid (approved):
Looks good 👍  and works well :shipit:

PR #13183, opened 2020-10-20
Use updated referrers store

### LINKED ISSUE #10057: SQLiteBlobTooBigException: Row too big to fit into CursorWindow requiredPos=0, totalRows=1
https://sentry.io/share/issue/b99a555f8217430aa5c6b2246659dfe5/

planarvoid:
Fixes #10057 

In order to lower the load there are some changes in the FluxC - https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1728

I'm also lowering the number of results on the `View all` page. I think 300 is more than enough results and I don't see anyone needing to scroll lower on a mobile device. Unfortunately, the endpoint doesn't offer pagination. 

To test:
- Go to a site with loads of referrers 
- Open stats/year stats
- Scroll to the Referrers block
- Notice it's showing the correct data (compare to calypso)
- Click on "View more"
- Notice the full screen view is showing the correct data

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/detail/PostAverageViewsPerDayUseCase.kt
```diff
@@ -24,7 +24,7 @@ import javax.inject.Inject
 import javax.inject.Named
 
 private const val BLOCK_ITEM_COUNT = 6
-private const val VIEW_ALL_ITEM_COUNT = 1000
+private const val VIEW_ALL_ITEM_COUNT = 300
```
mchowning: Just curious if there is a particular reason we don't have these use cases referencing a shared constant for this.

planarvoid: I think the idea was that we could tweak the number of visible items per use case. I think it makes sense to use a common constant though (since it's all the same). I'll change it

mchowning (approved):
LGTM! :shipit:

PR #13186, opened 2020-10-20
Adding slides to story from photo.

develric:
This PR is based on findings while final testing the #13131

When creating a story, first time you add a slide by taking a photo all works as expected but when adding additional slides from photo camera the camera is not opening up.

## To test
- Create a story
- Tap on photo FAB and notice the camera opens
- Take a photo and notice it is added as slide in the story composer
- Tap to add a new slide; the media picker is opened. Tap on the photo FAB
    - Behaviour before: camera does not open and you are back to the story composer; no slide is added
    - Behaviour after: the camera opens; take a photo and notice it is added as slide in the story composer

PR submission checklist:


planarvoid (approved):
I've tested the code and the stories work as expected 👍 , thanks for the fix!

PR #13188, opened 2020-10-21
Test: Verify auto-uploaded Bintray dependency works

mokagio:
This is just a test PR, not meant to be merged.

If CI can build, then we're good.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
     ext.kotlin_ktx_version = '1.2.0'
     ext.androidx_work_version = "2.0.1"
     ext.buildGutenbergMobileJSBundle = 1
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -50,6 +50,7 @@ allprojects {
         google()
         jcenter()
         maven { url "https://dl.bintray.com/wordpress-mobile/maven" }
+        maven { url "https://dl.bintray.com/mokagio/maven" }
```
mokagio, line 53: Version `40-b5e3198` is available on [my Bintray Maven repo](https://bintray.com/beta/#/mokagio/maven/utils-experiment), which I setup to avoid polluting the official one.

If we remove this repo from the list, Gradle won't find that version and the build will fail.

_QED_

PR #13192, opened 2020-10-21
[Home Page Picker] Create Base Container

antonis:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2717

## Description

- Adds Feature Flag for enabling Home Page Picker
- Adds header with title, subtitle and animation reused from Page Layout Picker with a *Skip* button at top-left
- Adds footer toolbar sliding up from the bottom

*Internal design reference: pb3aDo-xN*

### Note
[Extraction of resources common with the Modal Layout Picker screen may be revisited](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2725#issuecomment-711913791). Since reusing common resources will also occur in the following PRs and to avoid extending the scope, this is not yet handled.

## To test:

To start the **Home Page Picker** screen:

### Feature flag
1. Start the *Home Page Picker*
1. Verify that the new *Choose a design* screen appears

#### Elements
1. Start the *Home Page Picker*
1. Verify that the *Choose a design* title is shown
1. Verify that the *Pick your favourite...* subtitle is shown
1. Verify that the *Back* button is shown
1. Verify that the *Skip* button is shown
#### Collapsing behavior
1. Start the *Home Page Picker*
1. Scroll the empty preview list
1. Verify that the *Choose a design* title is collapsed
#### Rotation
1. Start the *Home Page Picker*
1. Rotate the device to landscape mode
1. Verify that the *Choose a design* title is collapsed

### Footer
1. Start the *Home Page Picker*
1. Press the *Skip* button (This is action is only for demo purposes)
1. Verify that the footer toolbar is sliding up from the bottom

## Screenshot

|Initial state|Sliding Toolbar |Collapsed state|
|---|---|---|
|device-2020-10-23-114919|device-2020-10-23-114930|device-2020-10-23-114947|

|Landscape|Landscape with toolbar|
|---|---|
|device-2020-10-23-115013|device-2020-10-23-115026|

PR submission checklist:


WordPress/build.gradle
```diff
@@ -124,6 +127,7 @@ android {
             applicationId "org.wordpress.android.prealpha"
             dimension "buildType"
             buildConfigField "boolean", "MODAL_LAYOUT_PICKER", "true"
+            buildConfigField "boolean", "HOME_PAGE_PICKER", "true"
         }
     }
```
antonis, line 133: Note ([ref](https://github.com/wordpress-mobile/WordPress-Android/pull/12482/files#r462066562)):
- `wasabi` is for developers
- `jalapeno` is for reviewers (automatic CI builds on PRs - needs to be manually downloaded and installed)
- `zalpha` is for alpha testers (closed track distributed through Google Play)
- `vanilla` is for beta testers and users/release (open track + release distributed through Google Play)

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerFragment.kt
```diff
@@ -0,0 +1,119 @@
...
+        backButton.setOnClickListener {
+            requireActivity().onBackPressed() // FIXME: This is temporary for PR #13192
+            viewModel.onBackPressed()
+        }
+        setScrollListener()
+    }
+
+    private fun setScrollListener() {
+        if (DisplayUtils.isLandscape(requireContext())) return // Always visible
+        val scrollThreshold = resources.getDimension(R.dimen.mlp_header_scroll_snap_threshold).toInt()
```
mkevins: NP: since we are using this more generally, should we change the name of this threshold the be more generic?

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerFragment.kt
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    private fun setupViewModel(savedInstanceState: Bundle?) {
+        viewModel = ViewModelProviders.of(requireActivity(), viewModelFactory)
+                .get(HomePagePickerViewModel::class.java)
+
+        viewModel.uiState.observe(viewLifecycleOwner, Observer { uiState ->
+            setTitleVisibility(uiState.isHeaderVisible)
+            AniUtils.animateBottomBar(bottomToolbar, uiState.isToolbarVisible)
+            when (uiState) {
```
mkevins, line 71: Just noting (as I'm not sure what will eventually go here) that we can omit the `else` clause when using the `when` expression, instead `when` as a statement. This might be moot in our case, but mentioning here since it's one of those subtleties of compiler feedback :smiley:

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerViewModel.kt
```diff
@@ -0,0 +1,73 @@
...
+        open val isToolbarVisible: Boolean = false
+    ) {
+        object Loading : UiState()
+
+        data class Content(
+            override val isHeaderVisible: Boolean = false,
+            override val isToolbarVisible: Boolean = false
+        ) : UiState()
+
+        class Error : UiState()
```
mkevins, line 71: Not suggesting a change here, but curious why we use a singleton for `Loading` and not for `Error`. Is this perhaps in anticipation of handling of different error states?

WordPress/src/main/res/layout/home_page_picker_bottom_toolbar.xml
```diff
@@ -0,0 +1,42 @@
...
+        android:layout_height="@dimen/mlp_bottom_shadow_height"
+        android:background="@drawable/modal_layout_picker_bottom_shadow" />
+
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:background="?attr/categoriesBackground"
+        android:gravity="center"
+        android:orientation="horizontal"
+        android:paddingBottom="@dimen/mlp_bottom_button_vertical_margin"
```
mkevins: Same comment as before: Do you think these dimensions generic enough now to rename them?

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerFragment.kt
```diff
@@ -0,0 +1,119 @@
...
+import org.wordpress.android.util.DisplayUtils
+import org.wordpress.android.util.setVisible
+import javax.inject.Inject
+
+private const val NUM_COLUMNS = 2
+
+/**
+ * Implements the Home Page Picker UI
+ */
+class HomePagePickerFragment : Fragment() {
```
mkevins, line 34: Please ignore if I've missed some context on this: I wonder if we should subclass from `SiteCreationBaseFormFragment`? It seems that class has the help menu item with an abstract `onHelp` method we could implement here. Is there something that makes this incompatible / inconvenient for re-use? Wdyt?

mkevins (commented):
Tested via Pixel 3a (real device) and everything is working beautifully :smile: . Nice work Antonis! Just left a few questions / comments on the code.

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerFragment.kt
antonis: Thank you for taking the time to review this PR @mkevins 🙇 

Good point 👍 
I refactored this along the other similar change below with 963619e2ac8f22eaa9fc8efb273f61ab5633c3bb
I will probably rename more resources as I move forward with the implementation.

WordPress/src/main/res/layout/home_page_picker_bottom_toolbar.xml
antonis: As above fixed with 963619e2ac8f22eaa9fc8efb273f61ab5633c3bb

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerViewModel.kt
antonis: > Is this perhaps in anticipation of handling of different error states?

Exactly. I expect that this might need to hold an error message depending on the error type like [this](https://github.com/wordpress-mobile/WordPress-Android/blob/0a82bc8f2b53896f7a9038d1c1a46321f150f50f/WordPress/src/main/java/org/wordpress/android/viewmodel/mlp/ModalLayoutPickerViewModel.kt#L354).

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerFragment.kt
antonis: I agree 👍 Thanks for pointing out 🙇 
I'll leave it as a placeholder for now and [handle this in the next PR as this part of the code takes shape](https://github.com/wordpress-mobile/WordPress-Android/pull/13218/files#diff-390502960224826eee61d0e68f9d1ffddfd5f7c9b63237b2d27f1b55ef7bfaceL71). Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerFragment.kt
antonis: I agree with this @mkevins. I started implementing this way but since there are a lot of hook points with the existing UI (and refactoring in the existing code that need to be done) I decided to keep this PR small (under 500 lines). I will eventually need to handle this [in a next issue](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2721) that seemed more targeted to this. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerFragment.kt
mkevins: Absolutely! :smiley:

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerFragment.kt
mkevins: That makes sense. Handling in the next PR is fine, and your explanation is sensible.

mkevins (approved):
Nice work Antonis, LGTM!

PR #13198, opened 2020-10-21
Mobile stories block - error handling 4.1  - move code out of EditPostActivity

mzorz:
Comes from https://github.com/wordpress-mobile/WordPress-Android/pull/13103/files#r508824133

This PR moves logic for Stories-related load/retry/cancel handling from EditPostActivity to dedicated class StoriesEventListener

To test:
Follow up on #13103 test steps

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
@@ -171,7 +171,6 @@ import org.wordpress.android.ui.stockmedia.StockMediaPickerActivity;
 import org.wordpress.android.ui.stories.StoryRepositoryWrapper;
 import org.wordpress.android.ui.stories.prefs.StoriesPrefs;
 import org.wordpress.android.ui.stories.usecase.LoadStoryFromStoriesPrefsUseCase;
-import org.wordpress.android.ui.stories.usecase.LoadStoryFromStoriesPrefsUseCase.ReCreateStoryResult;
 import org.wordpress.android.ui.uploads.PostEvents;
 import org.wordpress.android.ui.uploads.UploadService;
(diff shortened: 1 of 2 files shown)


jd-alexander:
@mzorz this is probably not related to this PR but I noticed that the error notification that gets created when a story isn't saved remains even after the save was successful so we probably want to cancel that notification in the logic somewhere. Let me know if you are able to reproduce. I was using `CASE A: error in saves`

mzorz:
> @mzorz this is probably not related to this PR but I noticed that the error notification that gets created when a story isn't saved remains even after the save was successful so we probably want to cancel that notification in the logic somewhere. Let me know if you are able to reproduce. I was using CASE A: error in saves

Thank you @jd-alexander  - tracked here https://github.com/Automattic/stories-android/issues/579

jd-alexander:
I was retesting Case B and got this error during the flow:

```
   kotlin.UninitializedPropertyAccessException: lateinit property editorMediaListener has not been initialized
        at org.wordpress.android.ui.posts.editor.media.EditorMedia.access$getEditorMediaListener$p(EditorMedia.kt:52)
        at org.wordpress.android.ui.posts.editor.media.EditorMedia$retryFailedMediaAsync$1.invokeSuspend(EditorMedia.kt:233)
```

This occurred after:

```
tap on the story block failed overlay and observe the dialog appears "Couldn't upload file"
tap RETRY
```
I relaunched the app and the upload continued flawlessly. This might have been due to some emulator issue. Not sure but sharing.

jd-alexander (approved):
Thanks for these changes @mzorz the error resolution and overall fallback mechanisms of your logic and the existing logic in the `UploadService` works really well. LGTM 🚢 once you are fine with this.

mzorz:
Thanks for your tests and the reported error! Will now merge this one and the base one 👍

PR #13205, opened 2020-10-22
Encode media filenames before upload

planarvoid:
Fixes #13194

Related FluxC PR - https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1731

This PR updates FluxC version with the fix that encodes the filename with UTF-8 encoding before uploading.

To test:

Download in your device/emu this image
Create a post and add an image block
try to upload into the image block the downloaded image above
it works

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13209, opened 2020-10-22
Merge release 15.9.1 into 16.0

loremattei:
Merge the 15.9.1 hotfix into 16.0

PR submission checklist:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -129,7 +129,7 @@ ext {
     androidxWorkVersion = "2.0.1"
 
     daggerVersion = '2.22.1'
-    fluxCVersion = '949a0579683e58bf93b7f0bf806a1cc3ac1f8e2c'
+    fluxCVersion = '1.6.22.1'
 
     appCompatVersion = '1.0.2'
     coreVersion = '1.2.0'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -129,7 +129,7 @@ ext {
     androidxWorkVersion = "2.0.1"
 
     daggerVersion = '2.22.1'
-    fluxCVersion = '1.6.24'
+    fluxCVersion = '1.6.24.1'
```
AliSoftware, line 132: Just ot be sure, is this expected that we poin to 1.6.24.1 here while the hotfix in #13205 was pointing to 1.6.22.1, right? I think it is, as I see 1.6.24.1 of FluxC got released since indeed, but I also see that https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1732 was still pending merge, so just want to double-check the git workflow was applied right here. (if we do follow semver correctly, 1.6.24.1 should be compatible with 1.6.22.1 anyway… but do we? 😅 )

AliSoftware (approved):
LGTM with just a confidence check about versioning used

PR #13217, opened 2020-10-23
Open local previews for atomic sites

### LINKED ISSUE #13093: Post Previews: Previewing changes to an already published post on an Atomic site shows the published version
### Expected behavior
When editing an already published post and previewing changes, we expect to see a preview of those changes appear in a web browser. 

### Actual behavior
The published version of the post is shown rather than the autosave. 

Edit a post on an Atomic site.  Do not use an ephemeral site as they do not correctly report their atomic status via the REST API.
After making some changes, attempt to preview the changes via the Preview option. 

Pixel 2, 14.8-rc-1  There is some more context and proposed fix at https://github.com/wordpress-mobile/WordPress-iOS/pull/15061#issuecomment-705840222

planarvoid:
Fixes #13093

This is a tricky issue. First I tried using the WPWebViewActivity but there was a lot of redirecting happening and I always ended up having the `wp-admin` site opened in the browser (outside of the web view). In the end I've tried using the browser for this very specific usecase and it works. It's a bit inconsistent but I think that's enough for this edge case. 

Since it's an Atomic site, we can't do any of the authentication we normally do - Auth token only works on wp.com sites, we don't have username and password. We need to let the user login in the browser. They only have to do it once. It's not super convenient but it's better than what's happening now. I'm open to suggestions for improvements!

To test:
- Go to an Atomic site with a plugin installed (so you can be sure it's fully transformed)
- Go to an existing post
- Make a change 
- Click on Menu/Preview
- Notice the browser is opened
- Log in to your site in the browser
- The post is opened and your local changes are visible
- Go back to the app
- Open another post
- Make local changes
- Click on Menu/Preview
- Notice the post with changes is opened in the browser
- Notice no login was necessary 
- Test with a post without changes to check it's opened in the in-app WebView activity

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
@@ -879,12 +879,13 @@ public class ActivityLauncher {
                     shareSubject,
                     true,
                     startPreviewForResult);
-        } else if (site.isWPComAtomic() && !site.isPrivateWPComAtomic()) {
+        } else if (remotePreviewType == RemotePreviewType.REMOTE_PREVIEW_WITH_REMOTE_AUTO_SAVE && site.isWPComAtomic()
+                   && !site.isPrivateWPComAtomic()) {
             openAtomicBlogPostPreview(
-                            context,
-                            url,
-                            site.getLoginUrl(),
-                            site.getFrameNonce());
+                    context,
+                    url,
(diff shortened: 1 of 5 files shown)


khaykov (approved):
Thanks for the fix, @planarvoid ! And sorry for the delayed review - I was AFK for some time.

Works as expected, and you said, the preview experience is not optimal, but I don't have better alternatives myself 😅

PR #13218, opened 2020-10-23
[Home Page Picker] Populate thumbnail cells of the Home Page Picker Container

antonis:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2718 and https://github.com/wordpress-mobile/gutenberg-mobile/issues/2719

**Related PRs:**
Depends on: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1743

## Description

This PR creates the UI of the layout cells using data pulled from the API. It also handles the selected states for the layouts.

*Internal design reference: pb3aDo-xN*

## To test:

<details>
<summary><strong>To disable or enable the development version of Home Page Picker</strong></summary>

- Open the app from the build that allows FeatureFlags such as a PR build or a local development build
    - By default, the modal layout picker is enabled only in development builds.
    - From the site page:
        - Click on your Gravatar.
        - Click on App Settings.
        - Click on Test feature configuration.
        - Toggle "HomePagePickerFeatureConfig" to enable or disable the picker
        - Restart the App
</details>

To start the **Home Page Picker** screen:

### Data loading
1. Start the **Home Page Picker**
1. Expect to see a loading animation
    - Note: [Ghost view will be handled by a separate PR](https://github.com/wordpress-mobile/WordPress-Android/pull/13288)
1. Once the data are fetched expect to see the thumbnails to be populated
    - Note: Some thumbnails my appear with a *Generating preview* image. A separate PR will handle polling the server till thumbnail is ready

### Error
1. Turn off the internet
1. Start the **Home Page Picker**
1. Expect to see an error view
1. Turn on the internet
1. Press the **Retry** button
1. Expect to see the data loading

### Selection
1. Start the **Home Page Picker** in portrait mode
1. Wait for the data to load
1. Tap on a design
1. Expect to see a selection overlay with a checkmark
1. Expect to see the toolbar to appear from the bottom
1. Tap on the selected design again
1. Expect to see a selection overlay disappear
1. Expect to see the toolbar hiding at the bottom

#### Phone in portrait
1. Start the **Home Page Picker** in portrait mode
1. Wait for the data to load
1. Expect to see two columns filling the screen as described in *Internal design reference: pb3aDo-xN*
#### Tablet or Phone in landscape
1. Start the **Home Page Picker**
1. Wait for the data to load
1. Rotate the device to landscape
1. Expect to see a different layout according to *Internal design reference: pb3aDo-xN*

## Screenshot

|Phone portrait  initial state|Phone portrait scrolled and selected state|
|---|---|
|device-2020-11-02-131233|device-2020-11-02-131511|

|Phone landscape initial state|Phone landscape scrolled and selected state|
|---|---|
|device-2020-11-02-131338|device-2020-11-02-131351|

|Tablet portrait initial state|Tablet portrait scrolled and selected state|
|---|---|
|Screenshot_1603904843|Screenshot_1603904356|

|Tablet landscape initial state|Tablet landscape scrolled and selected state|
|---|---|
|Screenshot_1603904331|Screenshot_1603904343|

PR submission checklist:


antonis:
Hello @iamthomasbishop and @SylvesterWilmott 👋 ,
thought this is still work in progress (no API integration yet) the UI of the Home Page Picker is ready for the most part.

I've attached some screenshots in the description and you can find the latest APK [above](https://github.com/wordpress-mobile/WordPress-Android/pull/13218#issuecomment-715397509).

Any early feedback on this is welcome :)

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/usecases/FetchHomePageLayoutsUseCase.kt
```diff
@@ -0,0 +1,42 @@
...
+    private val thumbDimensionProvider: ThumbDimensionProvider
+) {
+    private var continuation: Continuation<OnStarterDesignsFetched>? = null
+
+    suspend fun fetchStarterDesigns(): OnStarterDesignsFetched {
+        if (continuation != null) {
+            throw IllegalStateException("Fetch already in progress.")
+        }
+        val payload = FetchStarterDesignsPayload(
+                thumbDimensionProvider.width.toFloat(),
```
chipsnyder, line 28: I noticed we're differing a bit on how we're sending the width to the REST call. I see here we're using the [phone's width](https://github.com/wordpress-mobile/WordPress-Android/blob/7712041199a52d979803f77018d90f9c6475b3da/WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/ThumbDimensionProvider.kt#L38-L39), on my Pixel 3a it's asking for preview images with a width of 440 in portrait and 474 in landscape. 

On iOS, I'm using the size of the cell that display's the thumbnail, which makes a preview width of 320 on all devices with a scale of @2x and 480 with a scale of @3x. I think using the thumbnail's size might provide a better experience because then we can cache the data from portrait to landscape. And on larger devices like a tablet it might help keep the thumbnail sizes smaller. WDYT?

chipsnyder (commented):
This is looking really good @antonis! 

I had one question for you in the comments. I also noticed that when rotating after selecting a layout the selection is lost. Is this expected?

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerViewHolder.kt
```diff
@@ -1,15 +1,40 @@
...
+        uiState: LayoutGridItemUiState,
+        imageManager: ImageManager
+    ) {
+        imageManager.loadWithResultListener(preview, ImageType.THEME, uiState.preview, FIT_CENTER, null,
+                object : RequestListener<Drawable> {
+                    override fun onLoadFailed(e: Exception?, model: Any?) {
+                    }
+
+                    override fun onResourceReady(resource: Drawable, model: Any?) {
+                        uiState.onThumbnailReady.invoke()
```
mkevins: Not suggesting a change here, just curious: why use `invoke()` instead of directly calling like: `uiState.onThumbnailReady()`?

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerViewModel.kt
```diff
@@ -36,20 +100,49 @@ class HomePagePickerViewModel @Inject constructor(
...
-        }
+        // TODO
     }
 
     fun onBackPressed() {
         // TODO
     }
 
+    fun onRetryClicked() {
+        if (networkUtils.isNetworkAvailable()) {
```
mkevins, line 113: I wonder whether we should display a toast here when the network is not available :thinking: wdyt?

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/ThumbDimensionProvider.kt
```diff
@@ -0,0 +1,47 @@
...
+    private val displayUtilsWrapper: DisplayUtilsWrapper
+) {
+    private val containerWidth: Int
+        get() = min(getPixels(dimen.hpp_maximum_container_width), displayUtilsWrapper.getDisplayPixelWidth())
+
+    private val isTabletOrLandscape: Boolean
+        get() = displayUtilsWrapper.isTablet() || displayUtilsWrapper.isLandscape()
+
+    private val largeScreenColumns: Int
+        get() = (containerWidth - (2 * getPixels(dimen.hpp_card_margin_outer)) - getPixels(dimen.hpp_card_padding)) /
```
mkevins, line 30: NP: I was thinking it might be good to add a code comment to explain this calculation. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/ThumbDimensionProvider.kt
```diff
@@ -0,0 +1,47 @@
...
+                (getPixels(dimen.hpp_card_width) + getPixels(dimen.hpp_card_padding))
+
+    private val phoneColumns: Int
+        get() = 2
+
+    private val largeScreenWidth: Int
+        get() = getPixels(dimen.hpp_card_width)
+
+    private val phoneWidth: Int
+        get() = (containerWidth - ((phoneColumns + 1) * getPixels(dimen.hpp_card_padding))) / phoneColumns
```
mkevins, line 43: Same, re: code comment

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/ThumbDimensionProvider.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    private val isTabletOrLandscape: Boolean
+        get() = displayUtilsWrapper.isTablet() || displayUtilsWrapper.isLandscape()
+
+    private val largeScreenColumns: Int
+        get() = (containerWidth - (2 * getPixels(dimen.hpp_card_margin_outer)) - getPixels(dimen.hpp_card_padding)) /
+                (getPixels(dimen.hpp_card_width) + getPixels(dimen.hpp_card_padding))
+
+    private val phoneColumns: Int
+        get() = 2
```
mkevins: Does this need to be a getter?

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerFragment.kt
```diff
@@ -116,4 +118,7 @@ class HomePagePickerFragment : Fragment() {
             AniUtils.fadeOut(title, Duration.SHORT, View.INVISIBLE)
         }
     }
+
+    private fun isPhoneLandscape() = DisplayUtils.isLandscape(requireContext()) &&
```
mkevins: Can we use [this](https://github.com/wordpress-mobile/WordPress-Android/pull/13218/files#diff-c21c814a771a8de931459e462bb91a30bd9224cbe1488f5027e231f17d483da7R13) here?

mkevins (approved):
Tested via the described flows on Pixel 3a (real) and Pixel C (emulator), and everything is working beautifully as descibed! Nice work Antonis! :tada: 

I've left a few comments (non are blockers, IMO).

antonis:
Thank you for your feedback on this @chipsnyder 🙇 

> I also noticed that when rotating after selecting a layout the selection is lost. Is this expected?

Good catch 👍  I added handling this with fd94213dfe44965e5f910f2ca57cfaa9aa1283ff

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerViewHolder.kt
antonis: Thank you for taking the time to review this @mkevins 🙇 

> Not suggesting a change here, just curious: why use invoke() instead of directly calling like: uiState.onThumbnailReady()?

Good catch 👍 
There is no reason for this. I originally used a nullable `onThumbnailReady` type. After removing the nullability I didn't clean up properly. Fixed with bbe22199689cee2cb06a35cf083dd427145dce45

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerViewModel.kt
antonis: That would be a good addition. Fixed with aa989f2a0a92c209b4a0e5c380ec6c53f0f153f1

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/ThumbDimensionProvider.kt
antonis: Not really ;) Fixed with 2ac7f9eac466c7a03fb7d1009e46918883b6a982

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerFragment.kt
antonis: Good idea. Fixed with cda4fac3a662a4b863367ff6aa0778dc7a7f0b9c

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/usecases/FetchHomePageLayoutsUseCase.kt
antonis: I feel my solution is over-engineered but I couldn't find a simpler way to encode all the [design constrains](https://mobilegutenpagesp2.wordpress.com/2020/10/14/home-page-layout-picker-i2/).
* On **tablet and phone landscape** I'm using an `160dp` fixed width and translate that to pixels for the request.
* On **phone-portrait mode** thumbnail size dynamically to fill the full device width with two thumbnails per row.
* I always use `scale = 1.0` since I use pixels and not logical units in the call

> On iOS, I'm using the size of the cell that display's the thumbnail, which makes a preview width of 320 on all devices with a scale of @2x and 480 with a scale of @3x

I'm not sure if I can do the same on Android since the screen sizes are not standard.
Keeping a fixed preview width (especially in *phone-portrait mode*) will make things a lot simpler but will probably degrade the image quality in high resolution devices since the two thumbnails need to fill the full device width. Wdyt?

> we can cache the data from portrait to landscape

With the [addition of handling saved state](fd94213dfe44965e5f910f2ca57cfaa9aa1283ff) I prevent the data from reloading and use the already fetched images. This is not completely correct since it will use portrait size images for landscape and vice versa (depending on which is requested first) but seems better for user experience.

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/ThumbDimensionProvider.kt
antonis: Added with 07dc71ed360e4492384eed9b3361fe0ff28c1228

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/ThumbDimensionProvider.kt
antonis: As above

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/usecases/FetchHomePageLayoutsUseCase.kt
chipsnyder: > On tablet and phone landscape I'm using an 160dp fixed width and translate that to pixels for the request.
On phone-portrait mode thumbnail size dynamically to fill the full device width with two thumbnails per row.

Thanks for adding this context. I think I missed that part originally when looking at the code 👍

chipsnyder (approved):
This is looking good to me. Thanks for taking care of that rotation bug and the explanation on the request size :)

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerViewHolder.kt
mkevins: Ah, that makes sense. Thanks for the explanation!

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerViewModel.kt
mkevins: Awesome!

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerFragment.kt
mkevins: Nice!

PR #13175, opened 2020-10-19
Issue/13066 ppn move add category action

zwarm:
Part of #13066 

This PR removes the "Add New Category" link from the select categories bottom sheet and adds a "+" to the toolbar/title area. It also removes "bolding" from the Add Category button on the Add Category view.

Before|After|Before|After
---|---|---|---
|||

**Merge Instructions**
Note: This PR will be merged into a feature branch
Make sure #13167 has been merged (this PR may need a sync after 13167 has been merged - let me know)
Remove the Not Ready For Merge label
Merge as normal

**To test:**

- Launch the app
- Select Blog posts
- Select Edit on a post
- From the edit post view, tap the UPDATE action to launch the bottom sheet
- Tap the categories row to launch the Select categories view
- Notice the "Add new category" link is gone and there is a "+" in the toolbar area
-Tap on the "+" to ensure the Add Category view is launched
- Notice the "Add Category" button text is no longer bold

PR submission checklist:

- [X] I have considered adding unit tests where possible.
- [X] I have considered adding accessibility improvements for my changes.
- [X] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13219, opened 2020-10-23
Issue/13066 add category behavior

zwarm:
Part of #13066

This PR changes the submit behavior in Add Category. The "add category" request will no longer take place in `PrepublishingAddCategoryFragment`. The logic has been moved to `PrepublishingCategoriesFragment`, so that we can better control the action. The extra request is passed along as a `Bundle?` to the `PrepublishingScreenClosedListener.onBackClicked()`. Using a bundle keeps it generic so that it can be used for any current or future additions to the PPN bottom sheet family.

This PR also introduces a progress spinner to the select categories view. This keeps the user informed that an action is taking place and the list is in the progress of being updated. This aligns the feature more closely with how the EditPostSettings architecture works (which is the action is passed to the previous fragment/activity for submittal).

To test:

- Launch the app
- Navigate to Posts and choose a post to edit
- Tap the Update button to launch the select categories view
- Tap the Add Category + symbol on the upper right of the tool bar
- Enter a new category (with or without a parent - I'd try both)
- Tap the Add Category button
- Notice that you are taken back to the select categories view and the progress indicator is visible.

To test:

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryFragment.kt
@@ -148,10 +148,8 @@ class PrepublishingAddCategoryFragment : Fragment(R.layout.prepublishing_add_cat
             }
         })
 
-        viewModel.navigateBack.observe(viewLifecycleOwner, Observer { event ->
-            event?.applyIfNotHandled {
-                closeListener?.onBackClicked()
-            }
+        viewModel.navigateBack.observe(viewLifecycleOwner, Observer { bundle ->
+            closeListener?.onBackClicked(bundle)
         })
 
         viewModel.toolbarTitleUiState.observe(viewLifecycleOwner, Observer { uiString ->
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingAddCategoryViewModel.kt
@@ -1,11 +1,10 @@
 package org.wordpress.android.ui.posts
 
+import android.os.Bundle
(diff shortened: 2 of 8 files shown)


ashiagr (commented):
> This PR also introduces a progress spinner to the select categories view. This keeps the user informed that an action is taking place and the list is in the progress of being updated.

Nice improvement 👍


> The "add category" request will no longer take place in PrepublishingAddCategoryFragment. The logic has been moved to PrepublishingCategoriesFragment, so that we can better control the action. The extra request is passed along as a Bundle? to the PrepublishingScreenClosedListener.onBackClicked(). Using a bundle keeps it generic so that it can be used for any current or future additions to the PPN bottom sheet family.
…
This aligns the feature more closely with how the EditPostSettings architecture works (which is the action is passed to the previous fragment/activity for submittal).

LGTM 👍. Since we're already using `PrepublishingScreenClosedListener` for communicating back action to the `PrepublishingBottomSheetFragment`, I think is is fine to pass the bundle via `onBackClicked`. 

Separate to this, noticed an "existing API bug" related to my comment earlier here:
https://github.com/wordpress-mobile/WordPress-Android/pull/13175#issuecomment-714508307

Sometimes API responds with a success response containing `null` name if we try to add a subcategory with an existing title - [Video](https://cloudup.com/c2qM9VgHapG). 
It is also reproducible on Post settings screen. We might want to add a check for null category DB insertion. I’d suggest we create a separate low priority GH issue for it as it looks like an edge case and it is not introduced as part of current changes. Lmkwyt.

zwarm:
> I’d suggest we create a separate low priority GH issue for it as it looks like an edge case and it is not introduced as part of current changes. Lmkwyt.

Created a new [issue](https://github.com/wordpress-mobile/WordPress-Android/issues/13230) . imo it should be fixed at the API level, but I'll add create a task to prevent a crash until then. The _sometimes_ explains why I never experienced it. Thanks for the second set of eyes.

PR #13222, opened 2020-10-24
Issue/13066 select category behavior

zwarm:
Part of #13066 

This PR addresses some wonkiness within the category selection behavior. The PPN bottomsheet architecture is to always replace fragments instead of reusing them. Since the fragment is always reinstantiated we do not have access to data, as even the VM is restarted, which was causing the selected/not selected issues when moving to/from select/add. Rewriting that fragment behavior sits outside of the scope of this PR and is on track for my next hack week project.

To support keeping category selection alive, the data is passed along in a bundle to the `PrepublishingAddCategoryFragment` fragment and then the `PrepublishingAddCategoryFragment` passes that same bundle back to the `PrepublishingCategoriesFragment`. Which is then sent along when the new fragment instance is created.

This PR also adds a liveData observer to intercept the device backbutton press. It it observed from  `PrepublishingCategoriesFragment` & `PrepublishingAddCategoryFragment` , so they can save data/package up the bundle before requesting a back navigation.

`PrepublishingActionClickedListener.onActionClicked` & `PrepublishingScreenClosedListener.onBackClick` have been modified to both accept a `(bundle: Bundle?)`

The category selection view now has a progress spinner. The spinner is shown upon return from Add Category while the save is taking place. It is also shown on toolbar or device back button tap while the categories selected are updated in the post. This prevents the Home view from showing out-of-date categories.

**Merge Instructions**
Note: This PR will be merged into a feature branch
Make sure #13219  has been merged 
Remove the Not Ready For Merge label
Merge as normal

**To test:**
- Launch the app
- Select Blog posts
- Select Edit on a post
- Categories are shown on the Categories row

If the post has categories assigned, you will see them listed on the home view. If not, you will see Not Set or Uncategorized
Select Categories is shown when user taps on Categories

- Tap the categories row to launch the Select categories view 
- Select/unselect a few categories - but note which ones 
- Tap the Add New Category button "+" on the title bar
- Enter a new category and tap "Add Category Button" 
Note: A new category is NOT saved on back press - the explicit tap on the action button is needed

- Back on the select category view, the category you have added is visible and checked
- Take note of the toast which will tell you if the action succeeded or failed.
- The categories you unselected/selected will be shown properly
- Tap the <- Back arrow or the device back button to save your options

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/AddCategoryUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/AddCategoryUseCase.kt
@@ -1,31 +1,24 @@
 package org.wordpress.android.ui.posts
 
 import dagger.Reusable
-import kotlinx.coroutines.CoroutineDispatcher
-import kotlinx.coroutines.withContext
 import org.wordpress.android.fluxc.Dispatcher
 import org.wordpress.android.fluxc.generated.TaxonomyActionBuilder
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.TermModel
 import org.wordpress.android.fluxc.store.TaxonomyStore
 import org.wordpress.android.fluxc.store.TaxonomyStore.RemoteTermPayload
-import org.wordpress.android.modules.IO_THREAD
 import javax.inject.Inject
-import javax.inject.Named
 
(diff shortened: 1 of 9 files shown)


ashiagr (approved):
LGTM 👍.

PR #13242, opened 2020-10-27
Add from wp media lib for stories.

### LINKED ISSUE #13229: Missing add from WP Media Lib for stories
Noticed during internal call for testing (comment here pbMoDN-1oy-p2#comment-2766)

### Expected behavior
When adding media to stories I expect to be able to add from WP Media Library.

### Actual behavior
When adding media to stories the option to add from WP Media Library is not available.

- Create a new story
- In the picker that opens notice you cannot add from WP Media Lib.

develric:
Fixes #13229 

This PR:
- Adds the possibility to add from WP Media library for stories.
- Avoids showing actions when search is expanded or we are asking for permissions.

### test 1
- Create a new story
- In the picker that opens notice you can add from WP media picker; tap on WP media picker icon and add from there
- Smock test the picker and stories scenario

### test 2
- Create a new story
- In the picker that opens notice you can add from WP media picker
- Tap on search icon and notice both system picker and WP media picker icons are not visible with search expanded
- Make the same check in changing site icon scenario.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerViewModel.kt
@@ -143,8 +143,10 @@ class MediaPickerViewModel @Inject constructor(
     private fun buildBrowseMenuUiModel(softAskRequest: SoftAskRequest?, searchExpanded: Boolean?): BrowseMenuUiModel {
         val isSoftAskRequestVisible = softAskRequest?.show ?: false
         val isSearchExpanded = searchExpanded ?: false
-        val showSystemPicker = mediaPickerSetup.systemPickerEnabled && !isSoftAskRequestVisible && !isSearchExpanded
-        return if (showSystemPicker || mediaPickerSetup.availableDataSources.isNotEmpty()) {
+        val showActions = !isSoftAskRequestVisible && !isSearchExpanded
+        val showSystemPicker = mediaPickerSetup.systemPickerEnabled && showActions
+
+        return if (showActions && (showSystemPicker || mediaPickerSetup.availableDataSources.isNotEmpty())) {
             val actions = mutableSetOf<BrowseAction>()
             if (showSystemPicker) {
                 actions.add(SYSTEM_PICKER)
--- a/WordPress/src/main/java/org/wordpress/android/ui/photopicker/MediaPickerLauncher.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/photopicker/MediaPickerLauncher.kt
@@ -302,7 +302,7 @@ class MediaPickerLauncher @Inject constructor(
         }
         return MediaPickerSetup(
                 primaryDataSource = DEVICE,
-                availableDataSources = setOf(),
+                availableDataSources = if (browserType.isWPStoriesPicker) setOf(WP_LIBRARY) else setOf(),
                 canMultiselect = browserType.canMultiselect(),
                 requiresStoragePermissions = true,
                 allowedTypes = allowedTypes,


aforcier:
Hey @develric and @planarvoid ! I noticed this PR and gave the changes a try.

It looks like it's bringing back the option to add media from the WP library by reverting the stories picker to the previous version with bottom tabs:

media-picker-stories-fab-overlap

That's unfortunately still not ideal since the FAB shouldn't overlap the bottom buttons like that. The expected move to a 'source' menu I mentioned in the call for testing (pbMoDN-1oy-p2#comment-2766) would have also resolved that.

That would still be a good solution, but there are also a few other viable options if there's an issue with that, here are all of them together:

* Use the source button from before this PR, but have it open a drop-down menu that includes the WP media library instead of opening the device file picker directly (see pbMoDN-1oy-p2#comment-2766)
* Move the FAB up so it doesn't overlap with the bottom bar
* Move to top tabs (this is likely a future improvement and not a good quick fix option)

cc @megsfulton @osullivanchris 

Incidentally, for PRs affecting UI it's worth adding a screenshot or two, it makes it much easier to tell what the intended changes are.

planarvoid:
hey @aforcier , it looks like you haven't turned on the `ConsolidatedMediaPickerConfig` flag in the App Settings/Test feature config. It's not turned on by default yet. When you turn it on, it looks like this: 

Screenshot_1603875082

planarvoid (approved):
I've tested the flow and it's looking good 👍 . I'll leave this open for Alex to check it with the flag turned on 👍

aforcier:
> hey @aforcier , it looks like you haven't turned on the ConsolidatedMediaPickerConfig flag in the App Settings/Test feature config. It's not turned on by default yet. When you turn it on, it looks like this:

Ah that makes more sense, thanks 👍

Hadn't thought of adding it to the action bar, three items seems a bit busy to me but I see it's used elsewhere as well (the site image picker for example). It's definitely an improvement over the overlapping FAB though and could perhaps tide us over to the top bar approach if we do that eventually - cc @megsfulton WDYT?

megsfulton:
Hmm, I think it might be trying to do a lot with icons. It's hard to know what they're going to do without labels. 

I'm wondering if maybe a dropdown from the title might be clearest? 
Drop Down Title

Could also do it from the icon, but it's probably less intuitive. 
Drop Down Icon


@osullivanchris what are you thoughts?

planarvoid:
@megsfulton the previous solution was using icons as well, only in the bottom bar. I think at the moment we can either:
- release what we have
- use the previous picker (with the bottom bar) for stories 

In both cases we should start working on the suggested improvements with tabs for multiple sources
I personally don't like the solution with the spinner in the toolbar. It's not consistent with the rest of the app and not super intuitive.

megsfulton:
>release what we have

By release what we have, are you referring to the implementation in your comment [here](https://github.com/wordpress-mobile/WordPress-Android/pull/13242#issuecomment-717789144)? 

>use the previous picker (with the bottom bar)

If we go with this approach can we move the FAB button up so it doesn't overlap with the bottom bar?

aforcier:
Weighing the options with @megsfulton, how about we go for the previous (bottom tab) behavior, but bumping up the FAB so it doesn't overlap @planarvoid @develric? Seems like the best solution for now.

> In both cases we should start working on the suggested improvements with tabs for multiple sources

Agreed, definitely seems like a better long-term solution for this screen 👍

planarvoid:
> If we go with this approach can we move the FAB button up so it doesn't overlap with the bottom bar?

we can definitely do that 👍  is there a big difference for the UX if the icons are in the bottom bar and if they are in the toolbar?

osullivanchris:
Personally, what I think works long term is either:
- presenting the different sources in top tabs per my proposal on pbArwn-1cr-p2 (attaching sketch below)
- presenting the different sources in a dropdown menu as Megs suggested. But off the top of my head, I am not sure if that will work for all scenarios. 

I do agree that the icons at the top right are a bit cluttered and may not be understood, but thought it might be a step towards this type of future.

If we want to move back, per 
 
> If we go with this approach can we move the FAB button up so it doesn't overlap with the bottom bar?

as long as we do that fix, its fine by me. But the behaviour within stories with these bottom tabs is a little different to what we'll have elsewhere. I think it would need to be tidied up at some stage later.

sketch
proposed

frosty:
> But the behaviour within stories with these bottom tabs is a little different to what we'll have elsewhere. I think it would need to be tidied up at some stage later.

My two cents is that I feel like it would be good to have consistency between the different areas of the app, even if we're going to iterate on this again (part of the reason behind the Media Picker Consolidation) project was to align all our different uses of the media picker.

planarvoid:
> presenting the different sources in a dropdown menu as Megs suggested

The current app behaviour is that you open a "Choose from device" picker which allows you to open other subpickers. You can't switch from (let's say) gif selector to stock media selector. You have to go back to the "Choose from device" and select another source.
With a dropdown menu you'd just be switching content of the screen back and forth and that's currently not supported. That doesn't mean we can't implement it but it's not a trivial task.

osullivanchris:
Thanks for the explanation there @planarvoid - fair point

I'm in danger of just voting for my own design here! But I do think the tabs solves some things.

In the meantime - the main issue seems to be that the icons top right are not clear. However, the same information is presented when they're at the bottom. There's no labels. So its not necessarily any clearer. 

I think I'd be fine with what Vojta has here, at least as a v1

develric:
Hi all 👋  , a bit late to the party sorry; not super sure we had a final decision here 🤔 ; 

> In the meantime - the main issue seems to be that the icons top right are not clear. However, the same information is presented when they're at the bottom. There's no labels. So its not necessarily any clearer.
I think I'd be fine with what Vojta has here, at least as a v1

@osullivanchris am I right if I say we are currently fine to ship this v1 PR as it is (see this images @vojta [shared](https://github.com/wordpress-mobile/WordPress-Android/pull/13242#issuecomment-717789144) ), or do we need any actions before?

PS: hi @aforcier , also agree including images of UI changes is a (the) good practice, sorry for overlooking it here 🙇

megsfulton:
I don't agree that the icons in the top tab is a good approach but I am willing to disagree and commit for a v1 here. And while I understand that the goal was to consolidate the media picker around all the different use cases, the stories use case was not particularly well addressed in this first iteration. I'm looking forward to the clearer future iteration that @osullivanchris has proposed and hope that can be addressed soon.

PR #13243, opened 2020-10-27
CONTRIBUTING.md: Add information about our PR merge policy

AliSoftware:
Similar addition to our contributing guidelines as https://github.com/wordpress-mobile/WordPress-iOS/pull/15170, but adapted to the Android workflow

PR submission checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,6 +34,14 @@ The core team monitors and reviews all pull requests. Depending on the changes, 
 
 We do our best to respond quickly to all pull requests. If you don't get a response from us after a week, feel free to reach out to us via Slack.
 
+Note: If you are part of the org and have the permissions on the repo, don't forget to assign yourself to the PR, and add the appropriate GitHub label and Milestone for the PR
+
+### PR merge policy
+
+* PRs require one reviewer to approve the PR before it can be merged to the base branch
+* We keep the PR git history when merging (merge via "merge commit")
+* The reviewer who approved the PR will merge it right after approval (without waiting for the PR author) if all checks are green.
+
 ### Development Practices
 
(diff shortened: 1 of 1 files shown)


AliSoftware:
_Side note: Using this PR to test the new `@wordpress-mobile/release-managers` team we just created. Note that this new team should usually be used only for Release PRs while we'd still continue to use the `@wordpress-mobile/platform-9` for Tooling PRs like this one to affect three random P9 members, instead of every single MIE in P9 like `release-managers` does)_

mokagio (approved):
Thank you for working on this @AliSoftware, and for creating the "release managers" team ❤️ 

I like how clear and explicit this is. On the other hand, other than the last part regarding how a reviewer how approved the PR should merge it, everything else is automated, can be automated (a Peril check to make sure PRs have an assignee?), or configured in the repo settings (disable "squash and merge" and "rebase and merge").

I don't mind the verbosity of it, but I feel like we could keep the document lighter if we only left what we cannot (yet?) enforce.

What do you think? I'm fine either way, so I'm leaving a ✅ to unblock this (_although I realize I'm not following the convention by not merging it immediately_ 😳😅)

AliSoftware:
Yeah good point.

I actually thought about not mentioning the "1 reviewer approval needed" and "merge commit" given that they are (or can be) enforced by repo settings, but I feel like it's still nice to have this written down rather than any contributor wondering about our policy and not having admin access to our settings to check. I mean they don't need to if it's enforced, but that's still something that is interesting to share and document imho just for clarity and user documentation instead of the user wondering about it until they actually open the PR and investigate to see what's allowed to do?

As for labels and milestones, they are already enforced by Peril but I still thing it's worth mentioning them, because it's always nicer to know what you're supposed to do in advance before being yelled at by Peril that you forgot to do it. Peril is there to ensure you didn't forget to do it, but doesn't mean we shouldn't document what we expect so that contributors do the right thing right from the start. #CommunicationIsOxygen 😛

mokagio:
> Peril is there to ensure you didn't forget to do it, but doesn't mean we shouldn't document what we expect so that contributors do the right thing right from the start. #CommunicationIsOxygen

Fair point 👍

PR #13162, opened 2020-10-16
Reader Improvements: Sites/Blogs you might like

ashiagr:
Project Issue: #12899
Fixes #12834

This is a main branch for the **Reader Improvements: Sites/Blogs you might like** task.
List of related PRs can be found in #12834.

Light mode | Dark mode
------------|------------
device-2020-10-17-182144|device-2020-10-17-182125

**To test**
Bulk of PRs reviewed as part of #12834, needs one more round of testing after [recent recommended blogs API changes](https://github.com/wordpress-mobile/WordPress-Android/issues/13027#issuecomment-706770430).

To test:

1. Open Discovery tab
2. Scroll to 'Sites to follow' card
3. Click follow/unfollow button
4. Verify site was followed/unfollowed

**Open Questions**

1. Now that we have a recommended blogs section in the man discover feed, should we remove "Sites you may like" section from "Manage Topics and Sites" and corresponding `ReaderRecommendedBlog` table?

TODO

1. ~~Need to clear recommended blogs on pull-to-refresh~~ [here](https://github.com/wordpress-mobile/WordPress-Android/blob/9e59a46e33ea0fe7e79b7d5eaa96bfd8a7866fcc/WordPress/src/main/java/org/wordpress/android/ui/reader/services/discover/ReaderDiscoverLogic.kt#L216-L219) - Done in #13172
2. Add suggested site tracking (`reader_suggested_site_toggle_follow`, `reader_suggested_site_visited`) - Done in #13173

**Review/ Merge Instructions**

1. Review #13172,  #13173
2. Merge #13172,  #13173 to this PR
3. Test the app for all changes in this PR 
4. Merge this PR to develop

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13246, opened 2020-10-28
Remove recommended sites from manage topics and sites

ashiagr:
Since we now display recommended sites in the reader discover feed in https://github.com/wordpress-mobile/WordPress-Android/pull/13162, this PR removes "Sites you may like" tab (and associated logic) from the "Manage Topics and Sites" screen to have consistent behaviour with iOS app.

### To test

- Go to Reader
- Click Settings icon on the app bar
- Notice that "Sites you may like" section is no longer visible

Also notice that `tbl_recommended_blogs` does not exist in the `wpreader.db `using the AS -> database inspector.

Before | After
-------|------
before | after

PR submission checklist:


zwarm (approved):
LGTM! A nice clean removal. The tab no longer shows in the view and the table was removed from wpreader.db

PR #13247, opened 2020-10-28
Simplify MediaPickerActivity.kt

planarvoid:
I was investigating the complex way the media picker activity handles activity results and passes them back to the original caller and I think it could be much simplified. In most cases we can just pass full intent back to the caller without having to parse it. The exceptions are the system picker and the camera (where we have to load the selected/taken files). The calls to the other sources now don't need different request codes cause they are ignored anyway so all use the `PHOTO_PICKER` now. All the calls to the system picker now have one request code - `MEDIA_LIBRARY`. 

As an additional step I've decided to add all the results to the intent. This is a step that allows us in the future to have only one place that handles the result based on what identifiers it contains. Let me know what you think of this approach. Instead of a boolean flag with `queue_results` I've changed it to have a separate list of URIs that should be queued. 

To test:
- Test featured image picker other sources to check the redirecting works
- Test featured image default source to check the queueing works
- Take a photo with a camera in site icon picker 
- Check whatever you feel is necessary :-) 

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaItem.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaItem.kt
@@ -28,7 +28,7 @@ data class MediaItem(
     }
 
     sealed class Identifier(val type: IdentifierType) : Parcelable {
-        data class LocalUri(val value: UriWrapper) : Identifier(LOCAL_URI)
+        data class LocalUri(val value: UriWrapper, val queued: Boolean = false) : Identifier(LOCAL_URI)
 
         data class RemoteId(val value: Long) : Identifier(REMOTE_ID)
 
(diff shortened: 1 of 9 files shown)


develric (commented):
Hi @planarvoid, I tested this one using the steps from the internal call for testing (`pbMoDN-1oy-p2`). Vast majority is working fine 👍 , reporting below some scenarios that didn't work at least for me; let me know 🙇 .

- Media library- Add one edited image from device with media
- Site Icon Picker - Choose from device
- Site Icon Picker - Take photo
- Featured image picker - Take photo
- Gravatar selection - Take photo

planarvoid:
Thanks for the testing @develric !

I've fixed all of the issues. They were mostly related to incorrectly queueing the URIs. 

The site icon bug was fixed by 09359c6 - we had incorrectly set the `queueResults` parameter to `true` there. I think it should only be true when picking the featured image.

The other issues are fixed by 75fa1d0 where we now correctly set the uris as QUEUED_URIs when the `queueResults` flag is on (for taking photos and system picker). 

I've tested all the 5 scenarios + selecting featured image from system picker (which was also broken) and all seem to work well now. Let me know what you think!

develric (approved):
Thanks for the fixes @planarvoid 🙇 ! All worked well for me! Also love the usage you made of the Intent, good job 👍 ! :shipit:

Issue #13248, opened 2020-10-28
IllegalArgumentException org.wordpress.android.editor.gutenberg.GutenbergPropsBuilder in build


loremattei:
Sentry Url: https://sentry.io/share/issue/c98a69d0e21641bf885f0910eeae2650/
User Count: 426
Count: 1.8K
First Release: org.wordpress.android@15.8-rc-1 (925)
First Seen: Sep 24, 2020 6:51:47 PM UTC
Last Seen: Oct 28, 2020 10:16:55 AM UTC
24 Hours: 29
30 Days: 1.8K
```
org.wordpress.android.editor.gutenberg.GutenbergPropsBuilder in build at line 2
org.wordpress.android.editor.gutenberg.GutenbergContainerFragment in updateCapabilities at line 215
org.wordpress.android.editor.gutenberg.GutenbergEditorFragment in updateCapabilities at line 696
org.wordpress.android.ui.posts.EditPostActivity in onSettingsUpdated at line 680
org.wordpress.android.ui.prefs.SiteSettingsInterface$4 in run at line 1155
...
(40 additional frame(s) were not displayed)
```

emilylaguna:
The sentry url 404's for me so here's a correct one: https://sentry.io/share/issue/c98a69d0e21641bf885f0910eeae2650/

emilylaguna:
Unable to reproduce this on my end, reached out to @marecar3 who mentioned this maybe similar to https://github.com/wordpress-mobile/gutenberg-mobile/issues/2692 but since this is happening in 15.9.1 he's going to take another look. 

Thank you @marecar3!

marecar3:
Thank you @emilylaguna! 👍

PR #13254, opened 2020-10-28
Hotfix/13248_IllegalArgumentException org.wordpress.android.editor.gutenberg.GutenbergPropsBuilder in build

marecar3:
Fixes https://github.com/wordpress-mobile/WordPress-Android/issues/13248

To test:

At this moment it's not clear the exact scenario of how this crash can be reproduced.
I can assume that maybe as we are retaining `GutenbergEditorFragment` which is creating `GutenbergContainerFragment` it survives rotation while activity is null?

@mchowning @jd-alexander if you have ideas on how to reproduce the crash it would be helpful. Thanks.

PR submission checklist:

### DIFF
--- a/WordPress/build.gradle
+++ b/WordPress/build.gradle
@@ -92,9 +92,9 @@ android {
             dimension "buildType"
             // Only set the release version if one isn't provided
             if (!project.hasProperty("versionName")) {
-                versionName "15.9.1"
+                versionName "15.9.2"
             }
-            versionCode 938
+            versionCode 15.9.2
             buildConfigField "boolean", "ME_ACTIVITY_AVAILABLE", "false"
             buildConfigField "boolean", "TENOR_AVAILABLE", "false"
             buildConfigField "long", "REMOTE_CONFIG_FETCH_INTERVAL", "3600"
--- a/WordPress/build.gradle
+++ b/WordPress/build.gradle
@@ -56,7 +56,7 @@ android {
         } else {
             versionName "alpha-249"
         }
-        versionCode 934
+        versionCode 943
         minSdkVersion rootProject.minSdkVersion
         targetSdkVersion rootProject.targetSdkVersion
 
@@ -94,7 +94,7 @@ android {
             if (!project.hasProperty("versionName")) {
                 versionName "15.9.2"
             }
-            versionCode 15.9.2
+            versionCode 943
             buildConfigField "boolean", "ME_ACTIVITY_AVAILABLE", "false"
             buildConfigField "boolean", "TENOR_AVAILABLE", "false"
             buildConfigField "long", "REMOTE_CONFIG_FETCH_INTERVAL", "3600"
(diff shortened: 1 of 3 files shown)


libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergContainerFragment.java
```diff
@@ -212,7 +212,11 @@ public void updateTheme(Bundle editorTheme) {
     }
 
     public void updateCapabilities(GutenbergPropsBuilder gutenbergPropsBuilder) {
-        GutenbergProps gutenbergProps = gutenbergPropsBuilder.build(getActivity(), mHtmlModeEnabled);
-        mWPAndroidGlueCode.updateCapabilities(gutenbergProps);
+        // We want to make sure that activity isn't null
+        // as it can make this crash to happen: https://github.com/wordpress-mobile/WordPress-Android/issues/13248
+        if (getActivity() != null) {
+            GutenbergProps gutenbergProps = gutenbergPropsBuilder.build(getActivity(), mHtmlModeEnabled);
```
mchowning: Perhaps extract `getActivity()` out to a variable to just insure that we never hit the scenario where the call on line 217 an object, but the second `getActivity()` call on line 218 to returns `null`.

```
final Activity activity = getActivity();
if (activity !=null) {
  GutenbergProps gutenbergProps = gutenbergPropsBuilder.build(activity, mHtmlModeEnabled);
  ...
}
```

mchowning (approved):
I also wasn't able to find a way to recreate the crash either. I tested this by doing some smoke-testing both with your changes and with everything inside the if-not-null check commented out (to do a rough simulation of what might happen if the activity was `null`), and I didn't see any issues.

I did leave one small suggested change, but I do not think it is critical.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergContainerFragment.java
marecar3: Ah good one. Fixed!

WordPress/build.gradle
```diff
@@ -92,9 +92,9 @@ android {
             dimension "buildType"
             // Only set the release version if one isn't provided
             if (!project.hasProperty("versionName")) {
-                versionName "15.9.1"
+                versionName "15.9.2"
             }
-            versionCode 938
+            versionCode 943
```
mokagio, line 97: The latest `versionCode` on the Play Console was 942.

The issue that made CI fail [here](https://app.circleci.com/pipelines/github/wordpress-mobile/WordPress-Android/17778/workflows/dc3da90b-844d-4784-9234-33320a4e5918/jobs/79995) was that `versionCode` was the SemVer style value `15.9.2` instead of an integer.

I'm not sure how this happened in this instance, but I did the same thing once because I didn't read the prompt from the hotfix lane properly and input the `versionName` value when asked for `versionCode`.

Sounds like something we might want to guard against in the release toolkit. I had a quick look at the time and saw that Fastlane allows you to specify whether an input parameter should be an integer. But, when I tried to do it, the validation failed from CLI input. I haven't had the time to look deeper into it, yet.

cc @wordpress-mobile/release-managers.

PR #13259, opened 2020-10-29
Merge hotfix 15.9.2 into develop with manual conflicts resolution

mokagio:
I bumped the `versionCode` value to make sure it's clear that `versionCode` 943 belonged to 15.9.2 and has nothing to do with the current version.

I'm not 100% sure that's appropriate... Maybe I should have bumped `versionName`, too, shipping a new RC for 16.0?

Let me know and I'll update if necessary.
Thanks :smile:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,5 +1,9 @@
 16.0
 -----
+
+15.9.2
+-----
+* [*] Block Editor: Fix for a crash that can occur when activity is null [https://github.com/wordpress-mobile/WordPress-Android/issues/13248]
  
 15.9
 -----
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -3,7 +3,7 @@
 
 15.9.2
 -----
-* [*] Block Editor: Fix for a crash that can occur when activity is null [https://github.com/wordpress-mobile/WordPress-Android/issues/13248]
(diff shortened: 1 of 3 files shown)


loremattei (commented):
Hey @mokagio! 

There's no a right thing to do here. Usually, I find it better to accept the versionCode and versionName in develop, so that there won't be another conflict when merging the release branch in develop.

Even better, IMHO, I usually skip this merge PR. We already merge into the release branch and then the release branch back into develop, so the changes go to develop anyway, but we have to solve the merge conflicts only ones and I find the history clearer this way. 

For example, the changes in this PR are going to develop also with this PR: https://github.com/woocommerce/woocommerce-ios/pull/3082

AliSoftware:
> Even better, IMHO, I usually skip this merge PR

It could still be nice to have the hotfix changes in develop ASAP (with these kind of PRs) tho, especially in case the hotfix in question change a couple of files that are likely to be changed again by a new feature people are planning to work on and branch off of develop in the current sprint?

loremattei:
> It could still be nice to have the hotfix changes in develop ASAP (with these kind of PRs) tho, especially in case the hotfix in question change a couple of files that are likely to be changed again by a new feature people are planning to work on and branch off of develop in the current sprint?

On Android, you have to release new betas at the same time of the hotfix, because otherwise the testers will get the hotfix instead of the beta. So, the PR which merges the hotfix into develop and the one which merges the beta into develop land at the same time.

mokagio:
Thanks for the detailed explanation @loremattei. 

I'm happy closing this as discussed in https://github.com/wordpress-mobile/WordPress-Android/pull/13264#discussion_r514366073.

PR #13261, opened 2020-10-29
Merge hotfix 15.9.2 into release 16.0 with manual conflicts resolution

mokagio:
I bumped the `versionCode` value to make sure it's clear that `versionCode` 943 belonged to 15.9.2 and has nothing to do with the current version.

I'm not 100% sure that's appropriate... Maybe I should have bumped `versionName`, too, shipping a new RC for 16.0?

Let me know and I'll update if necessary.
Thanks :smile:

### DIFF
--- a/WordPress/build.gradle
+++ b/WordPress/build.gradle
@@ -92,9 +92,9 @@ android {
             dimension "buildType"
             // Only set the release version if one isn't provided
             if (!project.hasProperty("versionName")) {
-                versionName "15.9.1"
+                versionName "15.9.2"
             }
-            versionCode 938
+            versionCode 15.9.2
             buildConfigField "boolean", "ME_ACTIVITY_AVAILABLE", "false"
             buildConfigField "boolean", "TENOR_AVAILABLE", "false"
             buildConfigField "long", "REMOTE_CONFIG_FETCH_INTERVAL", "3600"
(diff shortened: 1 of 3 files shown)


WordPress/build.gradle
```diff
@@ -94,7 +94,7 @@ android {
             if (!project.hasProperty("versionName")) {
                 versionName "16.0-rc-3"
             }
-            versionCode 941
+            versionCode 944
```
loremattei, line 97: This is right, as 15.9.2 has versionCode 943.

WordPress/build.gradle
```diff
@@ -56,7 +56,7 @@ android {
         } else {
             versionName "alpha-252"
         }
-        versionCode 942
+        versionCode 944
```
loremattei, line 59: This is wrong. 16.0-rc-3 is set to 944, so the alpha has to be 945.

loremattei (approved):
Thank you @mokagio. 
Yes, we need to release a new beta after every hotfix on Android, otherwise beta testers will get the hotfix instead of the beta/alpha version. 
To make things cleaner, we usually handle the version conflicts in two ways:
- we accept the `versionCode` and `VersionName` of the beta branch and update the versions in the beta branch itself after the merge.
_or_  
- we bump the `versionCode` and `versionName` when fixing the conflict, so that the beta branch is ready for the release after the merge. 

This PR is a mix, where the versionCode has been updated, but the versionCode hasn't. 

This is not a big deal, and I'm merging this anyway and fixing the versions in the beta branch.

mokagio:
Noted. Thanks.

PR #13265, opened 2020-10-29
[TESTING] Unit Controls and web units

lukewalczak:
Fixes #

To test:

PR submission checklist:


iamthomasbishop:
@lukewalczak Just took the latest Android build for a spin, here's my feedback:

- The width of the text inputs seem to be static, so if there's a value of 4 or more digits, the label overflow is hidden ([screenshot](https://cloudup.com/cY1tsNs9ImM))
- There are some issues with the styling and behavior of the new `Menu` component: 
	- It doesn't seem to support dark mode on Android ([screenshot](https://cloudup.com/cQ9sKR8T3qM))
	- I don't think it's using our default sans-serif font (Roboto). Looks like it's using Open Sans, maybe? ([screenshot](https://cloudup.com/c0bBhHmBpFO))
	- Label colors also don't seem correct — can we have these use our standard/primary label color? ([screenshot](https://cloudup.com/c0bBhHmBpFO))
	- It looks like the `Menu` box shadow is getting cut off by the system home indicator bar ([screenshot](https://cloudup.com/cUQXiZ54Sjr))
	- I'm not sure if this is a bug or just a behavior of this specific  `Menu` component, but the entrance animation looks kinda off to me. It appears to always be entering from the bottom right of the screen ([screenshot](https://cloudup.com/cfbK-WnYcfo)), whereas I would expect it to start/be anchored to the bottom right of the element that triggered it. So for example, if I tap on a `unitControl` that is halfway up the screen, I expect the Menu to enter over the top of the `unitControl` that was tapped, as shown in [this example](https://cloudup.com/cFd59J-_CL9) from the Material Guidelines.
- Can we refine the spacing a little bit between the stepper buttons and `unitControl`? I think the padding of the `unit` segment makes it seem like there's more spacing on between it and the right-most stepper button. I think just removing the spacing between them will make it look more visually aligned. ([example](https://cldup.com/rP9rf6YPic.png))
- This isn't directly related to this work, but I'd like to increase the size of the stepper buttons (on Android only). I'm not sure what size they are, but I'd like to make them maybe 1.5x. ([screenshot](https://cloudup.com/cooD4kdvA55))

PR #13281, opened 2020-11-03
Media editing: Fix Glide cache issue for crop output file

ashiagr:
This PR fixes a Glide cache issue resulting in a wrong preview image after crop action.
Issue can be seen in [this video](https://cloudup.com/cWBbtx9ynoj).

Related Glide issues:
https://rb.gy/n0etgo
https://rb.gy/n0etgo

Output file path previously depended only on the input fie path hash. If the same input file path is re-chosen for editing, already present old cached image for that file is loaded on the preview screen.

To overcome this issue, a timestamp is appended to get a unique output file path. Another solution is to associate a signature in the Glide request. 

#### To test

1. Launch app and go to My Site
2. Edit a post (Gutenberg editor) and add a Gallery block
3. Click "Add Media" in the block -> Choose from device
4. Choose more than 1 images (so that preview screen can be seen in the next step) 
5. Click Edit 
6. Crop an image
7. Notice image preview 
8. Click back on the preview screen to return to the image chooser screen
9. Click Edit again on the image chooser screen (with the same images selected)
10. Crop the same image again 
11. Notice that the preview screen is loaded with the newly crop image and not the one from the old cache.

PR submission checklist:

### DIFF
--- a/libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/crop/CropViewModel.kt
+++ b/libs/image-editor/ImageEditor/src/main/kotlin/org/wordpress/android/imageeditor/crop/CropViewModel.kt
@@ -71,7 +71,8 @@ class CropViewModel : ViewModel() {
                 Uri.fromFile(
                     File(
                         mediaEditingDirectory,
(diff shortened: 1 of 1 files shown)


malinajirka (approved):
LGTM, thanks!

PR #13285, opened 2020-11-04
Use previous media picker in stories

planarvoid:
We've decided to not use the consolidated media picker for the Stories project. This PR disables it (instead of relying on the flag) but keeps the functionality in place in case we want to go back. 

As part of this project I'm moving the FAB in the old photo picker up so it doesn't overlap the bottom bar. The FAB was previously used only in the stories so there shouldn't be any issue with moving it in the layout. 

To test:
- Turn on the Consolidated picker flag in App Settings/Test feature config
- Go to My Site/Add Story
- Notice the previous UI is visible
- Notice the FAB is moved up

Screenshot_1604480715

PR submission checklist:


aforcier (approved):
Looks good @planarvoid, thanks!

I did notice one separate issue though, I guess that since we're not using the new media picker for now for Stories this fix is missing: https://github.com/wordpress-mobile/WordPress-Android/pull/12892. Can we apply the same change to the old media picker as well for now?

planarvoid:
Good point @aforcier, I've implemented this in 739fd05 (it was a one line change). Here is a video of the behaviour after the change:


stories_fab

aforcier:
Thanks @planarvoid, the FAB behavior works great 👍 

I found one more issue though 😬

1. Enable the consolidated media picker feature flag (the issue doesn't happen otherwise)
2. Enter the Story creation flow
3. Select any number of images and proceed

If you try this from the My Site FAB, the flow proceeds to get you to crop the image into a square (it looks like it's trying to update the site icon). If you try it from the Post List FAB, you just get taken back to the post list.

Here's the My Site flow:

media-picker-consolidated-stories-broken-flow

planarvoid:
Good catch @aforcier , this was because of a solution we implemented to avoid the need to use browser type when handling a story result. This is no longer necessary so I've removed the additional logic and I've kept the previous behaviour.

aforcier:
> Good catch @aforcier , this was because of a solution we implemented to avoid the need to use browser type when handling a story result. This is no longer necessary so I've removed the additional logic and I've kept the previous behaviour.

Sounds good @planarvoid - I think you may have forgotten to push the change though 😀

planarvoid:
oh god you're right @aforcier 🤦 , I've pushed 89a1495

aforcier:
Thanks @planarvoid - confirmed that creating stories from the my site view works now (media picker flag on or off), but I believe a similar fix is needed for the post list, as this is still happening (when the media picker flag is on):

> If you try it from the Post List FAB, you just get taken back to the post list.

planarvoid:
I've decided to remove all the code related to both consolidated media picker and the stories to make sure we're not breaking anything. I've tested:
- Add a story from My Site FAB without the flag
- Add a story from Post list FAB without a flag
- Add a story from My Site FAB with the flag
- Add a story from Post list FAB with a flag

Are these all the scenarios for the stories? 

All should work now. I think we can just reimplement the code if we decide for a future solution. The changes are in aa0bc75039fa4c3469a8c727d65b9727cba93b91 @aforcier

aforcier:
Great @planarvoid - confirmed that all those scenarios work as well 👍 

There's one more entry point (really recently added), which is adding the Story block to mobile Gutenberg directly from the block picker and then adding media. That brings up the camera and swiping up opens the media picker. I tested that and it's working fine with and without the flag, so we're all good there. (I merged `develop` into this branch locally when I tested since those changes are very recent.)

Thanks for making all the changes 🙇

:shipit:

PR #13291, opened 2020-11-04
Update OkHTTP library and FluxC

planarvoid:
Fixes #https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/919

This PR is here to test the OkHTTP update of FluxC. This was previously stopped because it breaks the self-signed SSL login. I've made some changes to the way the certificate is handled and I think it's working correctly now.

Let me know if there are any more testing steps that will confirm the behaviour!

To test:
- Use fresh install of the app
- Log in with a site with a self-signed SSL certificate (use the one from FluxC tests.properties)
- Notice the dialog that let's you confirm the certificate
- Go to "Blog posts"
- Notice the posts are loaded
- Go to "Site pages"
- Notice they are loaded
- Create a post
- Publish it
- It gets published correctly

PR submission checklist:


aforcier:
@planarvoid this is working well for me - though it needs to be updated to point to the new tip of https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1748.

(I confirmed that everything works on debug builds - it's broken on release builds unless I update the FluxC hash to the latest of that PR, in which case all is well 👍)

One more person testing out the app before updating okhttp would be a good idea though.

planarvoid:
good catch @aforcier , I've updated the FluxC version to point to the latest one with the fix and added one more person to check the PRs

malinajirka (approved):
LGTM, feel free to merge it when the CI finishes! Thanks!

aforcier (approved):
(Edited) I commented on the wrong PR, but still the approval makes sense 🙃

PR #13295, opened 2020-11-05
Update photo view version

malinajirka:
Updates PhotoView library version from 1.2.4 to 2.3.0. We need to update from 1.2.4 so we can disable jetifier in the future.

Before this change we'd manually attach PhotoViewAttacher to an ImageView. This approach isn't supported anymore, so I replaced ImageViews with PhotoView which seems to work just fine.

There shouldn't be any user facing changes.

To test:
1. Open Reader
2. Open a post with a photo
3. Click on the photo and make sure the fullscreen image loads correctly and that you can "pinch zoom"
------------
1. Open My Site
2. Open Media Library
3. Open media detail
4. Click on the image
5. make sure the fullscreen image loads correctly and that you can "pinch zoom"

PR submission checklist:


nbradbury (approved):
:shipit:

PR #13298, opened 2020-11-05
WP Stories : Prepublishing fragment container no longer cut off by nav bar in Android 10

jd-alexander:
Fixes # https://github.com/Automattic/stories-android/issues/537

## Findings

In Android 10, the bottom insets are not applied to the bottom sheet/fragment container. However, we are able to access the insets via `systemWindowInsetBottom` and dynamically adjust our layout so it sits on top of the navigation bar. 

## Solution
We get the `systemWindowInsetBottom` and we apply this as the `marginBottom` of our `fragmentContainer` so that the submit button no longer gets cut off.

**Note :** Within the attached [issue](https://github.com/Automattic/stories-android/issues/537) there is another keyboard bug where the story title focusing does not open the keyboard. 
That is when the keyboard is covering the view in its entirety. I think this solution might not be a fix for that specific issue and I will try to reproduce it in a clean branch and resolve it. Screenshot below. 

<kbd></kbd>

### Android 10 

1. Create a story post on an Android 10/ API 29 device. 
2. Click Next. 
3. Observe that the publish button is no longer cut off.

Before | After 
--------|-------
<kbd></kbd> |    <kbd></kbd>

Before | After 
--------|-------
 <kbd></kbd>   |    <kbd></kbd>
      
-----------------
### Other versions of Android 

1. Create a story post on an Android 10/ API 29 device. 
2. Click Next. 
3. Observe that the publish button is not cut off and the margin of the fragment container has not been changed.

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingBottomSheetFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PrepublishingBottomSheetFragment.kt
@@ -2,6 +2,8 @@ package org.wordpress.android.ui.posts
 
 import android.content.Context
 import android.content.DialogInterface
+import android.os.Build.VERSION
+import android.os.Build.VERSION_CODES
 import android.os.Bundle
 import android.view.KeyEvent
 import android.view.LayoutInflater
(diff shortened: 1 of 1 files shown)


mzorz (approved):
I confirm this works like a charm! 
Thanks for wrangling this @jd-alexander 
Also, I like that this is narrowed down to Android 10 - we can broaden the reach of the solution if we see this behavior happens on other versions as well, but I like that it's contained. 👍 
LGTM :shipit:

Issue #12491, opened 2020-07-21
Prepublishing Nudges : Publish button for stories cut off on Android 10


jd-alexander:
### Expected behavior
The publish button should be able to the keyboard and navigation bar.
Keyboard closed | keyboard open 
--------|-------
<kbd><img src="https://user-images.githubusercontent.com/1509205/88078731-6ae69d80-cb42-11ea-9296-116fe0d07bd9.png" width="320"></kbd>   |      <kbd><img src="https://user-images.githubusercontent.com/1509205/88078721-67ebad00-cb42-11ea-95e8-fea312484bea.png" width="320"></kbd>
 
### Actual behavior
The button is cut off by the status bar and keyboard. 

### Steps to reproduce the behavior
1. Create a story post. 
2. Press Publish. 
3. See the behavior.

Keyboard closed | Keyboard open 
--------|-------
  <kbd><img src="https://user-images.githubusercontent.com/1509205/88078518-2eb33d00-cb42-11ea-91dd-ced02c340bf7.png" width="320"></kbd>   |       <kbd><img src="https://user-images.githubusercontent.com/1509205/88078532-3246c400-cb42-11ea-9973-c545be63e85d.png" width="320"></kbd>

Tested on Google Pixel 4 XL, Android 10, WPAndroid `wp_alpha2` ()

ericamQA:
Another instance of this reported in zd-3607193

jd-alexander:
Thanks for reporting @wpericam I will look into this.

reginabally:
Another report in 3638952-zen with a video recording attached in the ticket.

Device: Xiaomi Redmi Note 9 Pro ( India ) 
OS: Android 10 
App version: WPAndroid 16.3

jd-alexander:
Thanks for reporting @reginabally we are looking into this issue.

jd-alexander:
I am thinking that the issue might be occurring because a 1-second delay or so might be needed to catch the keyboard opening on this specific device. The problem is I don’t have any immediate way to verify this. I am seeing where some peeps in the mobile division have Xiaomi devices, so I will try my best to reproduce it on the devices I own and if that's not possible, I will seek another way to address this. 

cc @mzorz

mzorz:
> I am thinking that the issue might be occurring because a 1-second delay or so might be needed to catch the keyboard opening on this specific device.

[...]

PR #13300, opened 2020-11-06
Prepublishing bottom sheet : fixed an issue where the keyboard overlays fragment container

### LINKED ISSUE #13299: Prepublishing bottom sheet : Keyboard overlays the fragment container
### Expected behavior
The Prepublishing Bottom sheet's content should always be above the keyboard.

### Actual behavior
Sometimes when you press the Next button on the Story Composer screen the keyboard overlays the fragment container. 

1. My Site
2. Press add new fab in the bottom right corner. 
3. Select Story post. 
4. Choose a media item. 
5. Press and close the bottom sheet. Eventually, the keyboard will overlay the content as seen below. 

<kbd></kbd>

jd-alexander:
Fixes #13299 

## Findings
The `KeyboardResizeViewUtil` is a layout hack used to add padding to a container layout's padding-bottom since we can't use the `adjustResize` of `android:windowSoftInputMode ` because the `StoryComposerActivity` is full screen. It seems the issue arises when the `OnGlobalLayoutListener` within the `KeyboardResizeViewUtil` is called and the `contentView` is not ready as yet hence the padding is not set.
 
## Solution

I thought of this solution because the property of a view not being set and reflected is a common occurrence when the main thread is handling a lot of work and these properties are modified before the view is properly attached to the window. So my solution is to utilize the `contentView.post` method so that we can queue the padding changes being applied on the `contentView`'s message loop, so once the view gets attached to the window, the padding will go into effect. 

## Testing
1. Create a new story post. 
2. Press the Next button multiple times after closing the bottom sheet and it will never overlap the content.

Before | After 
--------|-------
<kbd></kbd> | <kbd></kbd>

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/util/KeyboardResizeViewUtil.java
+++ b/WordPress/src/main/java/org/wordpress/android/util/KeyboardResizeViewUtil.java
@@ -35,30 +35,32 @@ public class KeyboardResizeViewUtil {
     ViewTreeObserver.OnGlobalLayoutListener mOnGlobalLayoutListener = new ViewTreeObserver.OnGlobalLayoutListener() {
         @Override
         public void onGlobalLayout() {
-            Rect r = new Rect();
-            // r will be populated with the coordinates of your view that area still visible.
-            mDecorView.getWindowVisibleDisplayFrame(r);
+            mContentView.post(() -> {
+                Rect r = new Rect();
+                // r will be populated with the coordinates of your view that area still visible.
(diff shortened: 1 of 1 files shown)


mzorz (approved):
Nice job @jd-alexander ! 🎉 
LGTM :shipit:

jd-alexander:
@mzorz It seems this issue is still occurring on some devices. This is also being tracked here https://github.com/wordpress-mobile/WordPress-Android/issues/12491 I am going to be looking into it.

PR #13311, opened 2020-11-06
Exclude support-annotations dependency from zendesk to skip jetifier

malinajirka:
This PR excludes support-annotations from the Zendesk dependency. It's the only dependency in zendesk which requires Jetifier. I believe we can simply drop it since support-annotations is used just for code inspection.

To test:
1. Smoke test Zendesk in the app

PR submission checklist:


antonis (approved):
Tested with the generated installable build and the app works as expected.
LGTM 🎉

PR #13315, opened 2020-11-06
Updating issue-triage to reference 'medium' priority instead of 'normal'

frosty:
This PR simply updates the references to 'normal' priority for issues to 'medium' priority to match the labels we use.


diegoreymendez (approved):
Checked and the doc looks good.  Thanks for taking care of the change!

Issue #13268, opened 2020-10-30
Jetpack Section: Activity Log


malinajirka:
https://github.com/wordpress-mobile/WordPress-Android/issues/13267

![activity_log_filtering](https://user-images.githubusercontent.com/2261188/100442209-ef749a00-30a7-11eb-8e9d-6980198a1f85.png)

#### Date Filter
- [x] Add support for a date range filter to the getActivities endpoint (https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/1744)
- [x] Add support for passing an input and an output date range (https://github.com/wordpress-mobile/WordPress-Android/pull/13324)
- [x] Add the calendar component (https://github.com/wordpress-mobile/WordPress-Android/pull/13324)
- [x] Add date range chip to Activity Log Screen
- [x] Polish the UI
- [ ] Fix status bar color (it's black now) = seems like a bug in the library (reported [here](https://github.com/material-components/material-components-android/issues/1966))
- [x] Fix timezone issue (https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1816 and https://github.com/wordpress-mobile/WordPress-Android/pull/13655)

#### Activity Type Filter
- [x] Create a new endpoint for fetching available activity types
- [x] Fix date format on the fetchActivityTypes endpoint - it seems the API ignores our requests
- [x] Create a skeleton vm, fragment, RecyclerView, ViewHolder and tests (https://github.com/wordpress-mobile/WordPress-Android/pull/13420)
- [x] Introduce UIStates (https://github.com/wordpress-mobile/WordPress-Android/pull/13439)
- [x] Show a loading screen while available activity types are being fetched (https://github.com/wordpress-mobile/WordPress-Android/pull/13440)
- [x] Show an error with retry when the fetch fails (https://github.com/wordpress-mobile/WordPress-Android/pull/13440)
- [x] Start showing a list of items with checkboxes (https://github.com/wordpress-mobile/WordPress-Android/pull/13469 and https://github.com/wordpress-mobile/WordPress-Android/pull/13470)
- [x] Introduce "clear" and "apply/save" buttons (https://github.com/wordpress-mobile/WordPress-Android/pull/13483 and https://github.com/wordpress-mobile/WordPress-Android/pull/13485)
- [x] Add activity type chip to Activity Log Screen
- [X] Polish the UI
- [x] Empty state (https://github.com/wordpress-mobile/WordPress-Android/pull/13707)

[...]

PR #13270, opened 2020-10-30
Introduce activity log filters feature flag

malinajirka:
Related to #13268 

This PR introduces a local feature flag for the Activity Log filters feature. When the feature flag is enabled a "Date range" chips is shown in the activity log.

Note: It doesn't matter if it gets merged into 16.1 or 16.2.

To test:
1. Open Activity Log and make sure the "Date range" chip is hidden
2. My Site -> Avatar icon in the toolbar -> App Settings -> Test feature configuration
3. Enable ActivityLogFiltersFeatureConfig and click on "Restart app"
4. Go back to Activity Log and notice "Date range" chip is shown

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13324, opened 2020-11-09
Issue/13268 introduce date range picker screen in Activity Log

malinajirka:
Partially fixes #13268 

Merge instructions: 
1. Review this PR
2. Make sure #13270 is merged
4. Merge this PR

This PR introduces Material Fullscreen Date Range picker component in Activity Log. It overrides some styles of the material component to facilitate future styling. The design of the date range picker hasn't been finalised yet so ignore the overall styling for now.

Unit tests will be added in a separate PR.

To test:
1. My Site -> Avatar icon in the toolbar -> App Settings -> Test feature configuration
2. Enable ActivityLogFiltersFeatureConfig and click on "Restart app"
3. Open Activity Log
4. Click on the "Date range" chip
5. Notice the fullscreen date range picker is shown
6. Select a date range
7. Click on "Cancel"
8. Click on the "Date range" chip and notice no date range is pre-selected
9. Select a date range
10. Click on "Save"
11. Click on the "Date range" chip and notice the date range from step 9 is pre-selected 

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
```diff
@@ -136,6 +146,14 @@ class ActivityLogListFragment : Fragment() {
         }
     }
 
+    private fun showDateRangePicker(initialDateRange: DateRange?) {
```
ParaskP7, line 170: Minor: I would suggest that the whole `MaterialDatePicker` import is used instead of starting with `Builder`. It might seem a bit confusing next time someone looks at that since `Builder` can seem too generic.

`MaterialDatePicker.Builder.dateRangePicker()`

Idea: What about having just one variable instead, and make this function a bit more simple? Example below:
```
        val picker = MaterialDatePicker.Builder
                .dateRangePicker()
                .setSelection(initialDateRange)
                .build()
        picker.addOnPositiveButtonClickListener { viewModel.onDateRangeSelected(it) }
        picker.show(parentFragmentManager, DATE_PICKER_TAG)
```

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -30,6 +31,8 @@ import org.wordpress.android.viewmodel.activitylog.ActivityLogViewModel.Activity
 import javax.inject.Inject
 import javax.inject.Named
 
+typealias DateRange = Pair<Long, Long>
```
ParaskP7, line 34: Love it! 🌟

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -151,6 +159,15 @@ class ActivityLogViewModel @Inject constructor(
         }
     }
 
+    fun dateRangePickerClicked() {
+        _showDateRangePicker.value = ShowDateRangePicker(initialSelection = currentDateRangeFilter)
```
ParaskP7, line 167: Minor: Since this data class has just one field in it I suggest that the named parameter can be emitted.

WordPress/src/main/res/values/styles.xml
```diff
@@ -93,6 +93,51 @@
...
+        <item name="materialCalendarHeaderTitle">@style/WordPress.MaterialCalendarHeaderTitle</item>
+        <item name="materialCalendarHeaderLayout">@style/WordPress.MaterialCalendarHeaderLayout</item>
+        <item name="materialCalendarHeaderSelection">@style/WordPress.MaterialCalendarHeaderSelection</item>
+        <item name="materialCalendarHeaderToggleButton">@style/WordPress.MaterialCalendarHeaderToggleButton</item>
+        <item name="materialCalendarHeaderConfirmButton">@style/WordPress.MaterialCalendarHeaderConfirmButton</item>
+        <item name="materialCalendarHeaderDivider">@style/WordPress.MaterialCalendarHeaderDivider</item>
+        <!-- Used to color the cancel button which doesn't use any overridable styles -->
+        <item name="colorOnPrimary">?attr/colorOnSurface</item>
+    </style>
+
```
ParaskP7, line 116: I know the styles are not ready, so I am just going to comment on the structure.

OCD Warning; What do you think about keeping the order of defining the styles within `WordPress.MaterialCalendarFullscreenTheme` above and the individual internal styles below consistent. For example, having the `...HeaderTitle`, then the `HeaderLayout` below it, etc. Now, they are a bit up and down as when reviewing it by clicking on the styles above and jumping on the actual definition seems a bit odd.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
```diff
@@ -100,6 +106,10 @@ class ActivityLogListFragment : Fragment() {
             uiHelpers.updateVisibility(date_range_picker, visibility)
         })
 
+        viewModel.showDateRangePicker.observe(viewLifecycleOwner, Observer { event ->
+            showDateRangePicker(event.initialSelection)
+        })
```
ParaskP7, line 132: Idea: After reviewing the whole solution and especially `dateRangePickerClicked()` and the new ViewModel `showDateRangePicker` state, I was thinking maybe this can become a bit simpler. What do you think about the below solution:

```
    var picker: MaterialDatePicker<DateRange>? = null

        ...
        picker = MaterialDatePicker.Builder
                .dateRangePicker()
                .build()
        picker?.addOnPositiveButtonClickListener { viewModel.onDateRangeSelected(it) }
        date_range_picker.setOnClickListener { picker?.show(parentFragmentManager, DATE_PICKER_TAG) }
        ...
    }
```

With the above, you don't need the `dateRangePickerClicked()` and the `showDateRangePicker` solution surrounding it, thus removing the `ShowDateRangePicker` data class and its nullable `initialSelection` DataRange field. In the end the result is the same and the second time the picker gets clicked it holds the previous state.

PS: I am not sure whether you are thinking a bit ahead and I don't know how and whether you would need this `showDateRangePicker` later on, so please take this purely as a small suggestion.
PSS: It all started with me not liking the nullable `initialSelection`! 😄

ParaskP7 (commented):
Cool stuff, now we have a material picker that we can play with! 🌟 

PS: I didn't use emoji this time, but I did use a suffix on each of my comments to give you some context of what to expect (like `Minor: Xyx` or `Idea: Xyz`). Let me know your thoughts.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
malinajirka: I agree, good suggestions ;) - fixed in 20d94ba41769ca5dfbe1e412c47f5963adbd5a27

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
malinajirka: Tbh I prefer to keep the naming parameters since you don't need to navigate to the class and you can clearly see what you are setting. If the variable was called `initialSelection` I wouldn't have used it, but since it's called `currentDateRangeFilter` I still find the naming parameter useful. Wdyt?

ParaskP7: What you described makes sense! Please feel free to keep it as is. Sometime the ide would actually do that for you as well, giving you the named parameter for free, but this way it is more explicit. Usually, I am using named parameters as a builder, or when I want to change the order, but this makes sense as well.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
malinajirka: I haven't tested it, but I think with this solution the selection won't survive configuration change. Also it'd store the selection even when the user clicked on the "cancel" button, right?

------

I kind of like when the VM layer is the only place which makes any decisions. In this proposal, the view layer knows that we want to show the picker when you click on "date_range_picker" chip - imo it's not responsibility of that layer. The view layer should ideally just propagate user and system events to the VM layer and all the decision should happen there. All this makes testing easier since we can "fake" only system/user events and we can test all the scenarios.

We might also have troubles with tracking - we'd either need to have all tracking in the view layer or we'd need to call something like "viewModel.trackDateRangeChipClicked()" in the on click listener. Which would introduce yet another decision to the view layer.

Wdyt?

> PSS: It all started with me not liking the nullable initialSelection! 😄

Yeah, I don't like that either. We could pass Pair(-1,-1), but that feels even worse.

WordPress/src/main/res/values/styles.xml
malinajirka: I agree, I'll update that tomorrow.

malinajirka:
Thanks for the review @ParaskP7! 


> PS: I didn't use emoji this time, but I did use a suffix on each of my comments to give you some context of what to expect (like Minor: Xyx or Idea: Xyz). Let me know your thoughts.

I like it. We actually already use "NitPick:" prefix on minor comments - eg. [this comment](https://github.com/wordpress-mobile/WordPress-Android/pull/13042#discussion_r500308455).

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
ParaskP7: Thank you so much for the detailed response @malinajirka !
> I haven't tested it, but I think with this solution the selection won't survive configuration change. Also it'd store the selection even when the user clicked on the "cancel" button, right?

Actually, you are 💯 right! The suggested by me solution will not survive configuration change, unless more work is done, which I don't think will create a better solution in the end. Also, indeed, the `cancel` button will store the selection, which I don't believe we want, you are right again.

-----

About having the VM layer make all the decisions, I again totally agree. Especially, what you described with tracking clicked. We would anyway need something like that on the VM layer when we will be adding the tracking related code.

> Yeah, I don't like that either. We could pass Pair(-1,-1), but that feels even worse.

I agree, let's keep it as is for now. If I think something better I'll let you know. Passing `Pair(-1, -1) feel worse, I agree.

WordPress/src/main/res/values/styles.xml
malinajirka: Fixed in 6f5e34d898f03d3d6f066090db7daf7514ae9bb5

ashiagr:
Material date range picker looks great! Good job 👍.
Just noticed one issue when the last selected date range was not pre-selected the next time. 

Steps to reproduce:

- Open the date picker 
- Change orientation 
- Select a date range 
- Click save
- Reopen the date picker 
- Notice that the last selected date range is not pre-selected


date-range

malinajirka:
Nice catch @ashiagr! Thanks for the review ;)

Fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/13324/commits/29b9c1760fd357ca08a4b8e5d93cf257de85ddfd and https://github.com/wordpress-mobile/WordPress-Android/pull/13324/commits/b7c44fece6d72b9005638fc84d6a47a6226c74e4.

ashiagr (approved):
LGTM 👍. I've tested and confirm that last selected data range is saved properly after save action is performed after orientation change.

Issue #13326, opened 2020-11-09
Jetpack Section: Scan


ashiagr:
#13267

![scan-i2](https://user-images.githubusercontent.com/1405144/98535220-be285b80-22ab-11eb-9d8e-149fcb288c8d.png)

## Designs
Add ability to start scans, see status, and manage threats. 

## Tasks

- [x] Introduce scan feature flag (use a temporary scan menu item as an entry point that displays when scan feature flag is enabled) - #13334

# High priority (must have)
 #### Scan
- [x] Show a loading screen while current scan status is being fetched (@ashiagr) #13920
- [x] Show an error with retry when the fetch fails (@ashiagr) #13920
- [x] Here to help -> link to contact support (@ashiagr) #13937
- [x] Polish the UI (@ashiagr) In Progress #13949, #13950
- [x] Filter out invalid states in `ScanStateListItemsBuilder` (@ashiagr) - #13923
- [x] "Something went wrong" screen/response from the server - if the app doesn't crash consider this medium priority (@ashiagr) #13923
- [x] Scan menu item shouldn't be shown for sites with business plan (@malinajirka) (https://github.com/wordpress-mobile/WordPress-Android/pull/13962)

#### Threat Details
- [x] Get a free estimate (sync with iOS) (@malinajirka) (https://github.com/wordpress-mobile/WordPress-Android/pull/13938)
- [x] Use Firebase config for the hardcoded url
- [x] Polish the UI (@ashiagr )
- [x] Sync with iOS about all supported threat types (@malinajirka)
- [x] Fix and Ignore threat action can fail on missing server credentials -> check how the app handles it (check even Fix All action on Scan screen) (@malinajirka) (https://github.com/wordpress-mobile/WordPress-Android/pull/13931)

#### Push Notification
- [x] Update the “threat found” notification so it redirects the user to the new screen (@malinajirka) (https://github.com/wordpress-mobile/WordPress-Android/pull/13941)

#### Scan History
- [x] Scan history -> threat details screen -> fix/ ignore action should show scan screen with updated threats status. (@malinajirka) (https://github.com/wordpress-mobile/WordPress-Android/pull/13924)

#### Tracks
- [x] Add tracking (@malinajirka) (https://github.com/wordpress-mobile/WordPress-Android/pull/13954)

#### Finishing Touches

- [x] Check Small Devices (@malinajirka)
- [x] Check RTL languages (@malinajirka)
- [x] Design Review (@ashiagr)

[...]

PR #13334, opened 2020-11-10
Jetpack Mobile: Scan - Introduce feature flag

ashiagr:
#13326

This PR introduces a local feature flag for the Scan feature. When the feature flag is enabled a "Scan" menu is shown on the My Site tab (currently irrespective of the site plan).

#### To test

Prerequisite:  Ensure `ENABLE_FEATURE_CONFIGURATION` is enabled in the `build.gradle` for the build variant you're testing.

1. Launch app and go to My Site tab
2. Notice that "Scan" menu is hidden
3. Go to My Site -> Avatar icon in the toolbar -> App Settings -> Test feature configuration
4. Enable ScanFeatureConfig and click on "Restart app"
5. Go back to My Site  and notice "Scan" menu is shown

device-2020-11-10-105404

PR submission checklist:

### DIFF
--- a/WordPress/build.gradle
+++ b/WordPress/build.gradle
@@ -76,6 +76,7 @@ android {
         buildConfigField "boolean", "WP_STORIES_AVAILABLE", "true"
         buildConfigField "boolean", "ANY_FILE_UPLOAD", "true"
         buildConfigField "boolean", "CONSOLIDATED_MEDIA_PICKER", "false"
+        buildConfigField "boolean", "SCAN_AVAILABLE", "false"
         buildConfigField "boolean", "ENABLE_FEATURE_CONFIGURATION", "true"
     }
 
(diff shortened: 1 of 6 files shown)


ParaskP7 (commented):
Great job! 🌟 

Tested the changes, I was able to see `Scan` row after enabling `ScanFeatureConfig`.

PS: The only thing I can point is that the `fillColor` on the scan vector asset is not white, just like the others are, especially since the resource name suggest it is white indeed. I understand that the tint color anyway changes from the style afterwards, but we might want to keep consistency on those vector icons. Wdyt Ashita?
PPS: Also, I believe this file can be formatted further.

ashiagr:
Thanks for the review @ParaskP7 !

> The only thing I can point is that the fillColor on the scan vector asset is not white, just like the others are, especially since the resource name suggest it is white indeed.

> PPS: Also, I believe this file can be formatted further.

You're absolutely right, I grabbed a temporary scan icon, I plan to replace it when we have some Figma designs ready - it allows to directly export icons. Can we wait till then @ParaskP7 ? I'll add an action item in the feature tasks list. Wdyt?

ParaskP7:
> You're absolutely right, I grabbed a temporary scan icon, I plan to replace it when we have some Figma designs ready - it allows to directly export icons. Can we wait till then @ParaskP7 ? I'll add an action item in the feature tasks list. Wdyt?

Of course! I didn't understood this is a temp icon indeed. An action item will suffice, thank YOU Ashita! 🌟

malinajirka (approved):
LGTM, and works as expected;). Thank you both!

PR #13352, opened 2020-11-10
Fix importing media from external sources i.e. google photos

### LINKED ISSUE #585: Importing from Google Photos crash

mzorz:
Fixes https://github.com/Automattic/stories-android/issues/585

There was a misconception that we needed to run the copy process of the files URI[ in the UI thread ](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/CopyMediaToAppStorageUseCase.kt#L20-L22) given the permissions for the passed Uris were temporary. Actually, the valid lifespan of these Uris is only for as long as [the context in which these were requested lasts](https://github.com/wordpress-mobile/WordPress-Android/pull/12670), so it's not really about the UI thread but rather about keeping the same context in which things were requested.
It makes sense since performing heavy operations such as i/o in the UI thread is counterintuitive.

This PR then removes the pre-existing code that made this operation run in the UI thread, and also performs the copy in the PhotoPickerActivity where it initially triggers the `startActivityForResult` to call the native picker, so the processing can be performed in this context.

While this was made apparent while testing Stories, I was able to observe the problem happening on Gutenberg and Aztec (although less frequently) as well.

To test:
On a handset where you have Google Photos installed and are logged in to:
1. start a new Story
2. when the media picker shows up, tap on the left-bottom button
3. then select `Select videos from device`
4. when the native picker appears, tap on the hamburger menu icon
5. look for Google Photos
6. choose a video that is long enough that would otherwise fail as per the reported issue.
7. verify the `Adding media...`  progress dialog appears on the PhotoPickerActivity and does not hang the UI, while the files are getting copied locally, and the app doesn't crash
8. verify the video gets finally added to the Story in the Story creator.

9. Repeat the same with an already created Story (i,e, add another slide to the Story you just created, and verify the same steps still work)

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoPickerActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoPickerActivity.java
@@ -1,6 +1,7 @@
 package org.wordpress.android.ui.photopicker;
 
 import android.app.Activity;
+import android.app.ProgressDialog;
 import android.content.Intent;
 import android.net.Uri;
 import android.os.Bundle;
@@ -29,7 +30,14 @@ import org.wordpress.android.ui.media.MediaBrowserActivity;
 import org.wordpress.android.ui.media.MediaBrowserType;
 import org.wordpress.android.ui.posts.FeaturedImageHelper;
 import org.wordpress.android.ui.posts.FeaturedImageHelper.EnqueueFeaturedImageResult;
+import org.wordpress.android.ui.posts.ProgressDialogHelper;
+import org.wordpress.android.ui.posts.ProgressDialogUiState.HiddenProgressDialog;
+import org.wordpress.android.ui.posts.ProgressDialogUiState.VisibleProgressDialog;
 import org.wordpress.android.ui.posts.editor.ImageEditorTracker;
+import org.wordpress.android.ui.posts.editor.media.CopyMediaToAppStorageUseCase;
+import org.wordpress.android.ui.posts.editor.media.CopyMediaToAppStorageUseCase.CopyMediaResult;
+import org.wordpress.android.ui.utils.UiHelpers;
+import org.wordpress.android.ui.utils.UiString.UiStringRes;
 import org.wordpress.android.util.AppLog;
 import org.wordpress.android.util.ListUtils;
 import org.wordpress.android.util.ToastUtils;
@@ -39,6 +47,8 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Timer;
+import java.util.TimerTask;
 
 import javax.inject.Inject;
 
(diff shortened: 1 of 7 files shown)


WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoPickerActivity.java
```diff
@@ -310,13 +325,40 @@ public void doNext(Uri uri) {
...
+            VisibleProgressDialog visibleProgressDialog = new VisibleProgressDialog(
+                    new UiStringRes(R.string.add_media_progress),
+                    false,
+                    true
+            );
+            mAddingMediaToEditorProgressDialog = mProgressDialogHelper
+                    .updateProgressDialogState(this, mAddingMediaToEditorProgressDialog, visibleProgressDialog,
+                            mUiHelpers);
+
+            new Timer().schedule(new TimerTask() {
```
mzorz, line 338: This needs to be run after a delay to allow the UI to get updated with the progress dialog, for which we could use `Handler().postDelayed`, but  we  need it not run in the UI thread, hence we're using a java `Thread`, scheduled to run with a `Timer`/ `TimerTask`. We could also have used AsyncTask but it's deprecated. The best option would be to use Kotlin Coroutines for this, probably a good fit for the picker rewrite effort.

aforcier:
Confirming that I tried the testing steps (with multiple lengthy videos from Google Photos) and everything worked smoothly. 👍

WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoPickerActivity.java
planarvoid: There are coroutines used in the PhotoPickerFragment and the PhotoPickerViewModel. Since we'll still be using this old photo picker for the stories, do you think it would make sense to use them? This approach doesn't work if you rotate your device when the progress dialog is shown (in that case the operation is cancelled).

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/CopyMediaToAppStorageUseCase.kt
```diff
@@ -2,43 +2,39 @@ package org.wordpress.android.ui.posts.editor.media
...
    * Some media providers (eg. Google Photos) give us a limited access to media files just so we can copy them and then
-   * they revoke the access. Copying these files must be performed on the UI thread, otherwise the access might be
-   * revoked before the action completes. See https://github.com/wordpress-mobile/WordPress-Android/issues/5818
+   * they revoke the access. Copying these files must be performed within the context (Activity) that requested the
+   * files, otherwise the access might be revoked before the action completes.
+   * See https://github.com/wordpress-mobile/WordPress-Android/issues/5818
    */
-    suspend fun copyFilesToAppStorageIfNecessary(uriList: List<Uri>): CopyMediaResult {
-        return withContext(mainDispatcher) {
+    fun copyFilesToAppStorageIfNecessary(uriList: List<Uri>): CopyMediaResult {
```
planarvoid, line 20: I think I'd keep it `suspend` and change the withContext to use background dispatcher (if it doesn't have to run on the main thread)

planarvoid (commented):
The code works mostly well 👍 , however there is an issue with the dialog and device being rotated. I think we should use the view model approach to fix this.

mzorz:
Closing in favor of https://github.com/wordpress-mobile/WordPress-Android/pull/13385 👍

PR #13356, opened 2020-11-11
Introduce skeleton activity type filter screen in Activity Log

malinajirka:
Partially fixes #13268 

This PR introduces a skeleton activity type filter screen. Only a dummy ui is used just to showcase that the dialog is shown on button click.

Big part of this PR was written during a pair programming session with @ParaskP7. Thanks 🙇 

To test:
1. My Site -> Avatar icon in the toolbar -> App Settings -> Test feature configuration -> **enable** "ActivityLogFiltersFeatureConfig" and restart the app
2. My Site -> Activity Log
3. Notice the "Activity Type" chip is shown
4. Click on the chip and notice a dialog is shown (it's not fullscreen yet)
5. Rotate your device and make sure the dialog survives the rotation
6. My Site -> Avatar icon in the toolbar -> App Settings -> Test feature configuration -> **disable** "ActivityLogFiltersFeatureConfig" and restart the app 
7. My Site -> Activity Log
8. Make sure the "Activity Type" chip is hidden

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/modules/AppComponent.java
+++ b/WordPress/src/main/java/org/wordpress/android/modules/AppComponent.java
@@ -32,6 +32,7 @@ import org.wordpress.android.ui.accounts.signup.SignupEpilogueFragment;
 import org.wordpress.android.ui.activitylog.detail.ActivityLogDetailFragment;
 import org.wordpress.android.ui.activitylog.list.ActivityLogListActivity;
 import org.wordpress.android.ui.activitylog.list.ActivityLogListFragment;
+import org.wordpress.android.ui.activitylog.list.filter.ActivityLogTypeFilterFragment;
 import org.wordpress.android.ui.comments.CommentAdapter;
 import org.wordpress.android.ui.comments.CommentDetailFragment;
 import org.wordpress.android.ui.comments.CommentsActivity;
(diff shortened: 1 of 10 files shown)


ParaskP7 (commented):
Great job and a big 🤗  for the pair programming session yesterday! 🌟

Tested the changes, I was able to see `Activity Type` chip after enabling `ActivityLogFiltersFeatureConfig`.

The only thing I can add here are the below:
1) The `Date Range` and `Activity Type` chips are too close to each other, maybe we might want to add some extra space in between. I know the pixel-perfect design is not ready, but just in case we want to have something closer to it for later.
1) When clicking on the `Activity Type` chip, we might want to make the dialog a full screen. This can happen later on, but maybe it is better if we have it ready now.
1) I have some suggestions on the commit splits. Maybe we can pick on that during our next pair programming session. Just adding that here so remind us.

malinajirka:
Thanks for the review @ParaskP7! 

> The Date Range and Activity Type chips are too close to each other, maybe we might want to add some extra space in between. I know the pixel-perfect design is not ready, but just in case we want to have something closer to it for later.

I'm completely ignoring the designs for now to save time. I agree it looks ugly, but I don't think it's important for this PR. I feel like we'll need to update it anyway and it might be better to do that in a separate PR and polish the UI as a whole + we'll want to request a design review on that PR.

> When clicking on the Activity Type chip, we might want to make the dialog a full screen. This can happen later on, but maybe it is better if we have it ready now.

Similar to the above comment. Moreover, I considered making it fullscreen on this PR, but decided against it since there are different approaches on how to make it fullscreen and I didn't want to block this PR on the related discussion. 

I introduced the button just to let us test that the fragment and vm is correctly initialised and that dagger works.

> I have some suggestions on the commit splits. Maybe we can pick on that during our next pair programming session. Just adding that here so remind us.

Looking forward to them ;)!

ParaskP7:
Thank you for the detailed explanations above @malinajirka ! 🤗

zwarm (approved):
All is working as expected, nice start! 👍

PR #13022, opened 2020-09-28
[Modal Layout Picker] "Preview" button functionality

antonis:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2452

**Related PRs:**

- Depends on: https://github.com/wordpress-mobile/WordPress-Android/pull/13018

- `gutenberg`:
1. https://github.com/WordPress/gutenberg/pull/25661
1. https://github.com/WordPress/gutenberg/pull/25930
- `gutenberg-mobile`:
1. https://github.com/wordpress-mobile/gutenberg-mobile/pull/2674
1. https://github.com/wordpress-mobile/gutenberg-mobile/pull/2696
- `WordPress-FluxC-Android`
1. https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1715

To test:

Layout Picker should show [when creating a new page from *My Site* or *Site Pages* the *Modal Layout Picker* appears](https://github.com/wordpress-mobile/WordPress-Android/pull/12482).

- Start by navigating to the Modal Layout Picker

### Preview a layout
1. Select a layout from the list
1. Select Preview
- **Expect** to see a read-only editor load that shows a preview of the selected layout.

### Preview a layout, then a different one
1. Select a layout from the list
1. Select Preview
- **Expect** to see a read-only editor load that shows a preview of the selected layout.
1. Select Back
1. Select a different layout from the list
1. Select Preview
- **Expect** to see a read-only editor load that shows a preview of the newly selected layout.

### Preview a layout, then cancel
1. Select a layout from the list
1. Select Preview
- **Expect** to see a read-only editor load that shows a preview of the selected layout.
1. Select Back
- Or close the modal by swiping down from the top
1. Close the modal via the ✖️ 

### Preview a layout, then create a page
1. Select a layout from the list
1. Select Preview
- **Expect** to see a read-only editor load that shows a preview of the selected layout.
1. Select Create Page
- **Expect** to see the page editor load with the selected layout content populated.
- **Expect** to see the page editor load with a title that relates to the content. Example a layout from the "About Pages" should have a title like "About"

### Spot check for regressions
Open an existing page or blog, expect the content to be loaded and for normal editing to take place.

### Spot check Modal Layout Picker Disabled
1. Disable Modal Layout Picker
1. Create a new page
- **Expect** to see the existing Starter Page Template Picker
1. Select an option
- **Expect** to see the Preview with "Close" and "Apply" as actions
1. Try Each option, expect it to behave as it previously did. 

## NOTES

[...]

PR #13358, opened 2020-11-11
[Modal Layout Picker] Prevents layout preview from creating a duplicate page

antonis:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2797

Relates with: https://github.com/wordpress-mobile/WordPress-Android/pull/13022

## Description

This PR prevents saving post locally in preview mode

### Preview and Create Page
1. Go to My Site > Site Pages > + and select any layout.
1. Tap Preview at bottom left.
1. Tap the Create page button.
1. Make some changes to the title and content.
1. Tap the back arrow at top left.
1. Verify that no duplicate post is created in drafts

### Preview only
1. Go to My Site > Site Pages > + and select any layout.
1. Tap Preview at bottom left.
1. Tap the back arrow at top left.
1. Verify that no local post is created in drafts

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
@@ -860,7 +860,9 @@ public class EditPostActivity extends LocaleAwareActivity implements
             return null;
         }));
         mEditPostRepository.getPostChanged().observe(this, postEvent -> postEvent.applyIfNotHandled(post -> {
-            mViewModel.savePostToDb(mEditPostRepository, mSite);
+            if (!mIsPreview) {
+                mViewModel.savePostToDb(mEditPostRepository, mSite);
+            }
             return null;
         }));
     }


malinajirka (approved):
Thanks for the fix @antonis! It works as expected.

This is heading exactly in the direction I was expecting when we decided to re-use EditPostActivity for previews. I don't think re-using the activity in the current state is maintainable in the long run. 

My bet is that as we keep updating the class (even unrelated changes) we'll keep introducing more and more issues. The code is already waaaay too complex and having an extra flag which completely changes how the activity behaves makes it even more complex and error prone.

I think we'll need to create a real preview activity in the future anyway. It'll just be more difficult since there will already be a ton of complex code and we won't have as much context as we do now. I'd personally consider allocating the time now or in the near future so it's handled before it's too late. (cc @bummytime)

Just my two cents.

antonis:
Thank you for taking the time to review this and your feedback @malinajirka. I really appreciate this 🙇 

> I think we'll need to create a real preview activity in the future anyway. It'll just be more difficult since there will already be a ton of complex code and we won't have as much context as we do now. I'd personally consider allocating the time now or in the near future so it's handled before it's too late. (cc @bummytime)

Do you think this should be based in editor instance (as we do now) or completely change direction with something like a WebView (I guess this would require server side work too)?

malinajirka:
> Do you think this should be based in editor instance (as we do now) or completely change direction with something like a WebView (I guess this would require server side work too)?

As I mentioned on one of the previous PRs. We decided to remove local previews because they weren't showing how the page will actually look like - the look in gutenberg isn't 100% precise. So I'd personally vote for a webview. But it's a bit more complex than that, since webview might not work on self-hosted sites. I'd need to do more research on that to give you a full answer. Either way, even if we decided to re-use gutenberg-mobile, handling previews in a clean almost empty class would be definitely better than the current state (even if we had to leave tons of callback methods empty because they wouldn't be needed for preview).

chipsnyder:
> handling previews in a clean almost empty class

This is the approach we took on the iOS side for previews. I know there is talk about improving the Gutenberg preview but I'm not sure where it is in the pipeline. Outside of that, since this static content, we might be able to stand up a demo site for each one similar to what we have for starter-designs on the home page picker.

Issue #13342, opened 2020-11-10
Setup presentation logic for My Site


planarvoid:
As part of this task 
- create a ViewModel for ImprovedMySiteFragment
- create a UI model for the my site using exposed LiveData (could be a list of UI model items)
- Link Adapter to the observed data

renanferrari:
Closed by #13360.

PR #13360, opened 2020-11-11
Setup UI layer for new MySite screen

### LINKED ISSUE #13341: Setup UI layer for My Site
Setup Adapter, Base ViewHolder and a parent UI model for the improved my site. These should all be extensible. The UI model should have an enum type. There doesn't need to be an actual view holder implementation yet.

planarvoid:
Fixes #13341 and #13342 

I've used this PR to handle 2 issues (cause they seemed really related). I've created an Adapter, a DiffCallback, ViewModel, put everything into the Fragment so it should be ready for implementation of all the basic list items. 

To test:
- Turn on the MySiteImprovements flag (App settings/Test feature configuration)
- Restart the app
- Notice the empty My Site fragment
- Notice nothing is crashing

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
```diff
@@ -1,9 +1,78 @@
...
+    }
+
+    private fun loadData(items: List<MySiteItem>) {
+        if (recycler_view.adapter == null) {
+            recycler_view.adapter = MySiteAdapter()
+        }
+        val adapter = recycler_view.adapter as MySiteAdapter
+        val recyclerViewState = recycler_view.layoutManager?.onSaveInstanceState()
+        adapter.loadData(items)
+        recycler_view.layoutManager?.onRestoreInstanceState(recyclerViewState)
```
renanferrari: I'm not sure I understand why we need to call `onSaveInstanceState()` and `onRestoreInstanceState()` here.

renanferrari (commented):
Thanks for the changes! I tested the app with the flag enabled and can confirm everything works as expected. Code also looks good, but I left a question below.

WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
planarvoid: hmm, good question. Generally it's there to keep scroll position after data changes but I think in this case it's not necessary (cause we don't expect massive changes to the data). I'll remove it.

renanferrari (approved):
LGTM! :shipit:

PR #13362, opened 2020-11-11
Copy media from system picker before they are passed down

### LINKED ISSUE #585: Importing from Google Photos crash

planarvoid:
Fixes Automattic/stories-android#585

When we use the system picker to select media from Google photos, the media has to be handled in the same context (the activity that requested them has to be alive). This PR handles this problem in the new media picker by taking the result of the system picker and passing it through the insert use case to copy the selected file to the app local storage.

To test:
- Turn on ConsolidatedMediaPicker flag in App Settings/Test feature configuration
- Follow steps in the following PR - https://github.com/wordpress-mobile/WordPress-Android/pull/13352

To test:
- Go to an existing post
- Go to post settings
- Click on "Choose featured image"
- Click on "System picker" icon in the Toolbar
- Select an image from a system picker
- The media gets selected and uploaded correctly
- Repeat the same steps not using System picker

To test:
- Go to a post in GB
- Add an image block
- Click on "Add Media"
- Click on "System picker" icon in the toolbar
- Select an item
- The item gets correctly added to a block
- Repeat with a video block and a video item

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerActivity.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerActivity.kt
@@ -24,7 +24,6 @@ import org.wordpress.android.ui.gif.GifPickerActivity
 import org.wordpress.android.ui.media.MediaBrowserActivity
 import org.wordpress.android.ui.mediapicker.MediaItem.Identifier
 import org.wordpress.android.ui.mediapicker.MediaPickerActivity.MediaPickerMediaSource.ANDROID_CAMERA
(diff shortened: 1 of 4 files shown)


mzorz (approved):
Tested with various scenarios, with and without permissions, found a funny race condition on a Samsung S5e tablet that would let the loadstate run and crash when no permissions given (fixed in bb0fa512c04e7f6ac285d1b3a11bc27014088ca9)
Also observed the progress dialog wouldn't show up in some situations and instead no feedback was painted on the UI (fixed 99bbe7d19aac27811d0acd0eb71771c35774b17d) 
Verified no further issues have been found related to this PR and it works well for me now
LGTM :shipit:

PR #13366, opened 2020-11-12
Stories: Update minimum supported Jetpack version

aforcier:
9.1 is the first Jetpack release with the Story block as a production block (and not a beta block), so it will be the earliest version we support when we lift the feature flag and ship to users on WPAndroid. (See https://github.com/Automattic/jetpack/pull/17536.)

### To test:

* In the app, select a site running Jetpack 9.1
* Confirm that the stories option appears from the My Site and post list FABs
* Switch to a site running a Jetpack with version earlier than 9.1
* Confirm that you're unable to access the stories option from the FABs.

PR submission checklist:


aforcier:
cc @bjtitus to do the same over on WPiOS 🙏  (no hurry)

mzorz (approved):
LGTM :shipit:

PR #13368, opened 2020-11-12
Enable consolidated media picker

planarvoid:
This PR turns the feature flag for consolidated media picker ON and makes it possible to control it remotely (turn it off when necessary).

To test:
- Install the app
- Make sure you haven't manually overriden the flag 
- Check that the new media picker is used everywhere except stories (where we still use the old picker)

Let's do a smoke test of the media picker as part of this PR

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/util/config/ConsolidatedMediaPickerFeatureConfig.kt
+++ b/WordPress/src/main/java/org/wordpress/android/util/config/ConsolidatedMediaPickerFeatureConfig.kt
@@ -1,15 +1,21 @@
 package org.wordpress.android.util.config
 
 import org.wordpress.android.BuildConfig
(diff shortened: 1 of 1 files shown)


develric (approved):
Hi @planarvoid 👋 !
I tested using the apk from this CI build (I think it should be fine as a check right?). 
I used:
- the call for testing steps in `pbMoDN-1oy-p2`plus 
- some smoke testing 

I tested on:
- my real device OnePlus 7T API 29 and
- on Emulators (Pixel 2 Emu API 23 and Pixel 2 Emu API 26). 

I only found a minor issue I reported [here](https://github.com/wordpress-mobile/WordPress-Android/issues/13370), but for the rest all was smooth! Good job 👍 ! 

I'm approving this PR from my side but not merging in case anyone else can have a test round as well before we close it 🙇 .

PR #13371, opened 2020-11-12
Merge Release/16.1-rc-4 into develop

loremattei:
Merge Release/16.1-rc-4 into develop

PR submission checklist:


WordPress/build.gradle
```diff
@@ -93,9 +93,9 @@ android {
             dimension "buildType"
             // Only set the release version if one isn't provided
             if (!project.hasProperty("versionName")) {
-                versionName "16.1-rc-3"
+                versionName "16.1-rc-4"
             }
-            versionCode 952
+            versionCode 954
```
AliSoftware, line 98: Is it expected that `versionCode` for `vanilla` and `defaultConfig` don't match (off-by-one, both before and after the change)?

mokagio: I can't find the comment where @loremattei explained this to me, so I could be remembering wrong, but basically we send out two different builds so that testers (or translators? 🤔 ) can get the right one.

Lore can correct me if I'm wrong.

mokagio (approved):
Approving but not merging while waiting for an explanation on the version numbers.

WordPress/build.gradle
loremattei: Yup! It's expected and required :-) As you can see by the associated `versionName`, `defaultConfig` is the configuration for the `alpha` version, and `vanilla` is the configuration for the `beta` version. They must have different versionCodes as they are different binaries.

AliSoftware: 👍 Cool (might be worth a comment in the gradle file at some point to remember that rationale 🙂  )

AliSoftware (approved):

PR #13372, opened 2020-11-12
Fix NPE in Reader tags

planarvoid:
Related FluxC PR https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1765

Updated OkHttp library doesn't support invalid POST requests with a null body. This PR replaces null body with an empty body which fixes this issue.

To test:
- Go to reader/cog wheel
- Unfollow a tag
- There is no crash and the tag gets unfollowed
- Follow a tag
- There is no crash and the tag gets followed

PR submission checklist:


malinajirka (approved):
LGTM, thanks!

PR #13373, opened 2020-11-12
Fix UI tests after media picker update

planarvoid:
We don't use MediaBrowserActivity any more so the tests don't work as expected. This PR should replace the view IDs with the new ones used.

To test:
- Rely on the UI tests build

PR submission checklist:


mzorz (approved):
:shipit:

PR #13376, opened 2020-11-12
Fix configuration for Connected Tests and Screenshots

loremattei:
This PR attempts a fix for an issue where some tasks run out of memory:

- it updates the memory configuration to leave more free memory to spawned tasks.
- it splits some critical jobs in smaller parts.

To test:
- Make sure CI is green

PR submission checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,10 +23,14 @@ commands:
           name: Setup gradle.properties
           command: cp gradle.properties-example gradle.properties && cp libs/utils/WordPressUtils/gradle.properties-example libs/utils/WordPressUtils/gradle.properties
   update-gradle-memory:
+      parameters:
+        jvmargs:
+          type: string
+          default: "Xmx2048m"
       steps:
         - run:
             name: Update memory setting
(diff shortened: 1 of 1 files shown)


loremattei:
Note that it looks like Connected Tests are failing right now in develop, likely because PRs have been merged, but verifying them was difficult because the Connected Tests-2 job failed during the build. 

This PR doesn't fix the test failure itself, but it fixes the build step so that the other failures can be investigated. 

When reviewing it, please make sure the build step is successful for the Connected Test-2 job.

mzorz (approved):
Since the errors seen were about memory issues, hope these parameters changes augmenting run memory work alright!
Will merge once all is green
LGTM :shipit:

renanferrari:
Just a heads up that we might still see an error here, probably related to the fix in #13373. Considering that PR needs the changes introduced here to make the CI job run successfully, I'd probably just go ahead and merge this anyway.

mzorz:
Sounds good @renanferrari ! Let's merge this one and then I'll update #13373 with `develop` right after 👍

PR #13382, opened 2020-11-13
Issue/13312 fix infinite loading dialog

### LINKED ISSUE #13312: Reader: Discover keeps showing an infinite progress screen
### Expected behavior
When `/cards` endpoint used for the discover tab in Reader returns an empty list of cards, the app should display an empty/error view.

### Actual behavior
When `/cards` endpoint used for the discover tab in Reader returns an empty list of cards, the app keeps showing an infinite loading screen. The user can leave the screen, but the UX is confusing.

1. Open Reader
2. Either make sure you are following tags for which the API doesn't return any cards or fake the api response.
3. Open Discover tab
4. Notice an infinite loading - restarting the app or even re-install won't help.

Emulator SDK 29

Originally reported by @planarvoid

malinajirka:
Partially fixes #13312

Not ready for merge: **There are two PRs for 13312 - they need to be merged into `develop` together.** 

We currently don't handle an empty response from the /cards endpoint correctly. When there are no data in database, the apps keeps showing an infinite progress screen since the DiscoverViewModel completely ignores empty ReaderDiscoverCards list. This PR fixes the issue, however the design of the new "empty" view is not polished - that's coming in the next PR.

I recommend reviewing this PR by commits. It might seem overwhelming but I think all the changes should be more or less clear. Let me know if you want me to clarify something.
- https://github.com/wordpress-mobile/WordPress-Android/commit/ded7315e595c97c60b0ba8fc9dc572830d74c8e6 - When the only card is the welcome card, we don't want to show it. Before this PR, this was handled in the DiscoverViewModel. This commit moves this logic into GetDiscoverCardsUseCase.
- https://github.com/wordpress-mobile/WordPress-Android/commit/51ab64c045d5f80a6562574abec502c9324afc68 and https://github.com/wordpress-mobile/WordPress-Android/commit/d39755dd1bece52d957c7e21cf2b72a8db909848 - these commits update the "TagUpdated" flag only when the response from the server contains some cards, when it doesn't the flag is cleared (When the response is empty, we want to re-try on the next app startup)
- https://github.com/wordpress-mobile/WordPress-Android/commit/784041ae515b7b8b7fd452da64889f0078f1b7aa - when there are no data in the database we don't want to propagate empty data set to the vm/ui before the fetch cards request finishes.
- https://github.com/wordpress-mobile/WordPress-Android/commit/4feab315ce5705f1e947999e9b828ba97bb480e9 - removes invalid "R.string" import. Lint is having troubles recognizing if a resource is used when it's referenced using "string.xyz" without the "R." prefix.
- https://github.com/wordpress-mobile/WordPress-Android/commit/9b18ad0f8a6ca2e0a422342bd1e7d5a6ece74c65 - Renames the existing empty state to make its name more clear.
- https://github.com/wordpress-mobile/WordPress-Android/commit/f2eb0548af52f393dce258f773aa2d74ce2d2ccd - We updated the business logic so we show a new empty view when the data from "discoverFeed" (data provider) is empty.
- https://github.com/wordpress-mobile/WordPress-Android/commit/b46159dc6970cffdd43c19f90600b3ad53565215 - fixes a newly introduced issue where the new empty screen is sometimes shown for a few ms.
- https://github.com/wordpress-mobile/WordPress-Android/commit/d3cf5600f56824e41416666ff833edcf41875f79 - introduces a new string resource for the new empty screen
- https://github.com/wordpress-mobile/WordPress-Android/commit/7a88a3ddc464a43c3765f5403dbd4e9a0c980f6a - adds a unit test for the newly introduced business logic
- https://github.com/wordpress-mobile/WordPress-Android/commit/c82568b6ec8a45069cc3bafeae874e3ec5a6e1dc and https://github.com/wordpress-mobile/WordPress-Android/commit/2ca9c7293dc35b3b251923c56a433b15d21354e4 -  makes ".throttle" testable so we can fix the existing unit tests.
- https://github.com/wordpress-mobile/WordPress-Android/commit/abd022ca38641f59bddba763ee6b51012038b96d - fixes existing and introduces new unit tests for the logic introduced in https://github.com/wordpress-mobile/WordPress-Android/commit/784041ae515b7b8b7fd452da64889f0078f1b7aa

To test:
I recommend testing the follow up PR. This PR contains just a dummy UI and we'll want to retest all the flows in the upcoming PR anyway. 

To test:

PR submission checklist:


zwarm (approved):
All works as expected, I was able to test in the next PR. The throttle was a nice addition. 👍  @malinajirka

PR #13385, opened 2020-11-13
Fix importing media from external sources i.e. google photos

### LINKED ISSUE #585: Importing from Google Photos crash

planarvoid:
Fixes https://github.com/Automattic/stories-android/issues/585

I'm copying the description from [this PR](https://github.com/wordpress-mobile/WordPress-Android/pull/13352): 

There was a misconception that we needed to run the copy process of the files URI[ in the UI thread ](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/CopyMediaToAppStorageUseCase.kt#L20-L22) given the permissions for the passed Uris were temporary. Actually, the valid lifespan of these Uris is only for as long as [the context in which these were requested lasts](https://github.com/wordpress-mobile/WordPress-Android/pull/12670), so it's not really about the UI thread but rather about keeping the same context in which things were requested.
It makes sense since performing heavy operations such as i/o in the UI thread is counterintuitive.

I've moved the logic to ViewModel in this PR and it seems to work well.

To test:
On a handset where you have Google Photos installed and are logged in to:
1. start a new Story
2. when the media picker shows up, tap on the left-bottom button
3. then select `Select videos from device`
4. when the native picker appears, tap on the hamburger menu icon
5. look for Google Photos
6. choose a video that is long enough that would otherwise fail as per the reported issue.
7. verify the `Adding media...`  progress dialog appears on the PhotoPickerActivity and does not hang the UI, while the files are getting copied locally, and the app doesn't crash
8. verify the video gets finally added to the Story in the Story creator.

9. Repeat the same with an already created Story (i,e, add another slide to the Story you just created, and verify the same steps still work)

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/CopyMediaToAppStorageUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/CopyMediaToAppStorageUseCase.kt
@@ -4,7 +4,7 @@ import android.net.Uri
 import dagger.Reusable
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.withContext
-import org.wordpress.android.modules.UI_THREAD
+import org.wordpress.android.modules.BG_THREAD
 import org.wordpress.android.util.AppLog
 import org.wordpress.android.util.AppLog.T.UTILS
 import org.wordpress.android.util.MediaUtilsWrapper
@@ -14,7 +14,7 @@ import javax.inject.Named
 @Reusable
 class CopyMediaToAppStorageUseCase @Inject constructor(
     private val mediaUtilsWrapper: MediaUtilsWrapper,
-    @Named(UI_THREAD) private val mainDispatcher: CoroutineDispatcher
+    @Named(BG_THREAD) private val bgDispatcher: CoroutineDispatcher
 ) {
     /*
    * Some media providers (eg. Google Photos) give us a limited access to media files just so we can copy them and then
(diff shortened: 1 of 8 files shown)


mzorz (approved):
Thanks a lot for these changes @planarvoid ! A discussed over DM, tested on emu and a tablet with Google Photos, verified it works with several and feels much more responsive and safer :) 

LGTM! :shipit:

WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/CopyMediaToAppStorageUseCase.kt
```diff
@@ -14,15 +14,18 @@ import javax.inject.Named
...
    * Some media providers (eg. Google Photos) give us a limited access to media files just so we can copy them and then
    * they revoke the access. Copying these files must be performed on the UI thread, otherwise the access might be
    * revoked before the action completes. See https://github.com/wordpress-mobile/WordPress-Android/issues/5818
+   * they revoke the access. Copying these files must be performed within the context (Activity) that requested the
+   * files, otherwise the access might be revoked before the action completes.
+   * See https://github.com/wordpress-mobile/WordPress-Android/issues/5818
    */
     suspend fun copyFilesToAppStorageIfNecessary(uriList: List<Uri>): CopyMediaResult {
-        return withContext(mainDispatcher) {
+        return withContext(bgDispatcher) {
```
mzorz, line 28: AH! before we merge, can we add [this check](https://github.com/wordpress-mobile/WordPress-Android/pull/13352/files#diff-c85806f6820a41ec9a43ee9f4b74536efb93928240e47de7668f342813aeff33R23-R25) here?

```
  if (!mediaUtilsWrapper.isInMediaStore(mediaUri) &&
                                // don't copy existing local files
                                !mediaUtilsWrapper.isFile(mediaUri)) {
[...]
```

because once it's copied as a local file, it may be copied once more when added through this same use case being reused in other use cases such as `addNewMediaToEditorAsync` in `AddLocalMediaToPostUseCase` and such

lmk 🙏

mzorz: Thanks for the change in 1e9844f! Verified it works as expected 👍

PR #13389, opened 2020-11-13
Merge Release/16.1 back into develop

loremattei:
Merge Release/16.1 back into develop.

The `Release Build` task is currently failing on the CI. It's shown in the report, but it's not required for merging the PR and the error itself is most likely a weird bug in the release-toolkit, so the code is ok as long as the other jobs are green.

PR submission checklist:


AliSoftware (approved):
> The Release Build task is currently failing on the CI. It's shown in the report, but it's not required for merging the PR and the error itself is most likely a weird bug in the release-toolkit, so the code is ok as long as the other jobs are green.

Yeah weird. Approved as I think it shouldn't be a blocker indeed

mokagio (approved):

PR #13390, opened 2020-11-13
Release/16.1

loremattei:
Merge release/16.1 into trunk.

The Release Build task is currently failing on the CI. It's shown in the report, but it's not required for merging the PR and the error itself is most likely a weird bug in the release-toolkit, so the code is ok as long as the other jobs are green.

PR submission checklist:


AliSoftware (approved):
LGTM

> The Release Build task is currently failing on the CI. It's shown in the report, but it's not required for merging the PR and the error itself is most likely a weird bug in the release-toolkit, so the code is ok as long as the other jobs are green.

Could it be related to the fact that it was recently re-enabled (according to this diff at least) and hasn't been used for a while so we missed a specific change to report on the script?

mokagio (approved):

loremattei:
> Could it be related to the fact that it was recently re-enabled (according to this diff at least) and hasn't been used for a while so we missed a specific change to report on the script?

I used it to build all the betas on this release and it worked as usual. It's been failing with a weird behaviour only on the final build, which differed from the previous beta only because of the translations 🤷‍♂️.

I'll investigate this later today.

PR #13355, opened 2020-11-11
Update gutenberg-mobile with non-dev Story block

aforcier:
Adds the change from https://github.com/wordpress-mobile/gutenberg-mobile/pull/2795, adding the Story block to the supported list for all builds (not just development builds).

PR #13130, opened 2020-10-14
Mobile stories block

mzorz:
This PR contains all the work done for integrating the Jetpack Story block in mobile-gutenberg, and making the move from gutenberg to the story composer in the stories library integrated as well.

This is the feature branch PR.

All of the PRs have been tested.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13391, opened 2020-11-13
Revert "Update gutenberg-mobile with non-dev Story block"

cameronvoell:
Reverts wordpress-mobile/WordPress-Android#13355

That PR updated the gutenberg-mobile reference on WPAndroid develop after our [1.41.0 gutenberg-mobile release](https://github.com/wordpress-mobile/WordPress-Android/pull/13322) (targeting 16.2) was already cut, so that PR is bringing in unintentional gb-mobile changes. 

If we would like to still bring the changes from wordpress-mobile/WordPress-Android#13355 to WPAndroid 16.2, we will need to create a gutenberg-mobile 1.41.1 betafix release.


mchowning:
Would we also need to revert the changes in https://github.com/wordpress-mobile/WordPress-Android/pull/13130 @cameronvoell ? 😬

PR #13397, opened 2020-11-14
Create job that fails build if gutenberg-mobile submodule hash changes

mchowning:
WIP

This is a job we could turn on and off as appropriate to prevent changes to the gb-mobile submodule hash from being updated on `develop` when a gb-mobile release is in process.

To test:

PR submission checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -555,7 +555,7 @@ jobs:
               curl "$url" | jq '.base.ref' | tr -d '"'
             )
 
-            if [ -z "$target_branch" ] || [ "$target_branch" != 'develop' ]; then
+            if [ "$target_branch" != 'develop' ]; then
               echo "Not targeting develop branch. No need to run gutenberg submodule check."
               exit 0
             fi
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,7 @@
 orbs:
   # Using 1.0 of the Orbs means it will use the latest 1.0.x version from https://github.com/wordpress-mobile/circleci-orbs
   android: wordpress-mobile/android@1.0
+  jq: circleci/jq@2.2.0
   git: wordpress-mobile/git@1.0
   bundle-install: toshimaru/bundle-install@0.3.1
   slack: circleci/slack@3.4.2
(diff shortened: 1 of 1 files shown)


mchowning:
It feels like we would need a "global" variable we could change on circle ci to control whether this job is active (i.e., whether it fails if there is a submodule change). I don't think we could use a variable defined within this repo for that because nothing would prevent someone from merging code where the job was deactivated (i.e., code they branched off of develop when a gb-mobile release was not in testing) to develop at a later point in time, when the gb-mobile release is being tested.

We could use set a [project-wide environment variable](https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-project) in circle ci (or perhaps use a variable in a circleci context) to globally activate/deactivate this job, but that's adding more process to the releases, which is obviously not ideal. If we [updated the environment variable via an API call though](https://circleci.com/docs/api/v2/#operation/createEnvVar), that might not be so bad. We could just add it to our release scripts. 🤔

mchowning:
Thinking about this some more, one weakness of this approach would be that CI doesn't have to run right before the PR is merged. For example, someone could open a PR on day 1, not get approvals until day 14, and then merge the PR, with CI only having been run on day 1. From a practical standpoint, this seems pretty unlikely though since pretty much any PR that updates the gb-mobile submodule hash would be doing so immediately after a gb-mobile PR merge, which would mean that this PR would need to be updated with the merged gb-mobile hash before it could be merged. Still, this does seem like a weakness for any CI-based approach.

aforcier:
This is a great idea @mchowning. If flagging the builds to fail has some kinks to work out/might add more process to the release, we could consider a simpler interim solution of a Peril message on all PRs that update the gutenberg-mobile submodule reference.

The message might be something like 'This PR is updating the gutenberg-mobile submodule reference. Please make sure you're aware of \[gutenberg-mobile's independent release timeline](link-to-release-process-doc).'. Folks doing a release/targeting the right branch can just ignore it, but it would warn occasional contributors and the like that they might be missing something.

We do something like this for subtrees: there's a message on all PRs modifying the `utils` (and I believe `login`) subtrees with a prompt to sync your changes to the remote, like this:

https://github.com/wordpress-mobile/WordPress-Android/pull/13352#issuecomment-725012460

mchowning:
> we could consider a simpler interim solution of a Peril message on all PRs that update the gutenberg-mobile submodule reference.

I like your idea @aforcier . 👍  It's easier to implement and maintain, less fragile, and may be all we need to prevent most, if not all, of these issues. And if we find out down the road that we really do need a stronger guard against changes to the submodule hash, we can always take another look at having CI block PRs.

Any thoughts @hypest , @cameronvoell , @ceyhun ?

hypest:
> Any thoughts @hypest 

My personal pref is to go for a blocking mechanism so people have less things to read, remember, or pay attention to. Also, it can help the release wranglers worry about one thing less (uncoordinated merges) while the editor code freeze is in progress.

I understand that the solution in this PR is not bullet proof yet so, I wouldn't mind a simpler one (like the idea for a Peril nudge) so we don't turn this into a whole project at this point (since not many incidents have happened so far).

ceyhun:
I agree with moving forward with a Peril message on the PR for now. Later on maybe we can leverage the [milestones data from gutenberg-mobile](https://api.github.com/repos/wordpress-mobile/gutenberg-mobile/milestones) and `due_on` field to check if a release is going on and fail the job like @jd-alexander mentioned before.

mchowning:
Closing this since it is no longer relevant due to the fact that we have [dropped the use of the submodule for the Gutenberg dependency](https://github.com/wordpress-mobile/WordPress-Android/pull/14175).

Issue #13379, opened 2020-11-13
Publishing empty post result in undismissable progress dialog and unresponsive app


khaykov:
##### Steps to reproduce the behavior
- Create a new empty post.
- Tap PUBLISH in the toolbar.
- Notice the "Saving" progress dialog and that "Can't publish empty post Toast appears for a moment"
- Confirm that dialog is not disappearing and can't be dismissed.

Might not be clear from GIF - but the progress dialog is never dismissed by itself.
[![Image from Gyazo](https://i.gyazo.com/ffc0ae6ed7f20f29455eede2b80eceed.gif)](https://gyazo.com/ffc0ae6ed7f20f29455eede2b80eceed)

hypest:
Wondering if this related to recent work on the save-on-exit. @mchowning , @malinajirka can you please have a look? Thanks!

hypest:
Fixed with https://github.com/wordpress-mobile/WordPress-Android/pull/13381

Issue #12570, opened 2020-07-30
First preview attempt does not work for changes to previously saved drafts


designsimply:
Steps to reproduce:

1. Create a new draft.
1. Add a title and content.
1. Go to More > Preview.
1. Expect the preview to work.
1. Make any change to the content.
1. Go to More > Preview.
1. Observe the preview does not open.
1. Go to More > Preview again.

<img width="270" alt="Screen Shot 2020-07-30 at Jul 30 3 58 01 PM" src="https://user-images.githubusercontent.com/1119271/88979132-42ffe400-d27e-11ea-9612-9879426898e6.png">

**Result:** the preview does not open the first time but does work if you try to preview a 2nd time. ([1m28s](https://cloudup.com/caXfaQVEu67))

<p style="clear:both;"><sup>Tested with WPAndroid 15.3 (production) and 15.4-rc-1 (TestFlight beta) on Pixel 3 Android 10 using a WordPress.com Premium site with a custom domain.</sup></p>

designsimply:
I noticed this again while beta testing with 15.6-rc-1—in this case, it's timing out most often and the editor stays open but sometimes the app freezes.

Video: [44s](https://cloudup.com/cseXSXWMmpE)
<p style="clear:both;"><sup>Tested with WPAndroid 15.6-rc-1 on Pixel 3 Android 10.</sup></p>

<details>
<summary>App logs for the timestamp matching the video</summary>

```
WordPress - 15.6-rc-1 - Version code: 910
Android device name: Google Pixel 3

01 - [Aug-25 17:46 API] Dispatching action: ListAction-LIST_DATA_INVALIDATED
02 - [Aug-25 17:46 API] Dispatching action: ListAction-LIST_DATA_INVALIDATED
03 - [Aug-25 17:46 API] Dispatching action: PostAction-UPDATE_POST
04 - [Aug-25 17:46 API] Dispatching action: ListAction-LIST_DATA_INVALIDATED
05 - [Aug-25 17:46 API] Dispatching action: ListAction-LIST_DATA_INVALIDATED
06 - [Aug-25 17:46 API] Dispatching action: PostAction-UPDATE_POST
07 - [Aug-25 17:46 API] Dispatching action: ListAction-LIST_DATA_INVALIDATED
08 - [Aug-25 17:46 API] Dispatching action: ListAction-LIST_DATA_INVALIDATED
[...]

Issue #13361, opened 2020-11-11
Preview Out-Of-Date if Save is In Progress


mchowning:
### Expected behavior

A post preview should always represent the current content in the editor.

### Actual behavior

If an autosave is in progress when the user attempts to preview the post, the currently-being-saved changes will not be uploaded and included in the preview.

### Steps to reproduce the behavior

1. Open a very long post, ideally on a slow device (so it will be slow to save)
2. Add some content, 
3. Wait at least 0.5 seconds (so the autosave delay has passed), but not much more than that (because we don't want the save to finish) and tap the "Preview" menu button.
3. Observe that the post is _not_ uploaded (assuming there was no other updated content saved apart from what was added just before the preview was opened) and the preview does _not_ include the just-added content.

![preview-save-in-progress mp4](https://user-images.githubusercontent.com/4656348/98831097-1d4fb280-2409-11eb-82b4-dba7d5cff453.gif)

mchowning:
When investigating this, we should keep in mind that there is another bug that sometimes causes the preview to not actually open: https://github.com/wordpress-mobile/WordPress-Android/issues/12570.

hypest:
Added the "Priority High" lavel since this essentially means the Preview flow is broken since one cannot preview the changes currently in the editor. Please dive in @mchowning , thanks!

malinajirka:
<p>I looked into this issue. I added bunch of logs and I think the bug might be in Gutenberg.</p>



<p>When the bug happens, the editor doesn't emit change </p>



<pre class="wp-block-preformatted">mEditorFragment.getTitleOrContentChanged()</pre>



<p>Moreover, <a href="https://github.com/wordpress-mobile/WordPress-Android/blob/b739d8a436e6b2efbeff6584c065409e939b1e28/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L1795">this method</a> keeps returning the state before the last change (even 5 seconds after the change) -> so even if the EditPostActivity manually tries to sync the content with the UI, it doesn't help, because the EditorFragment keeps returning the old content.</p>



Here are logs when the preview works as expected: (I changed title from "A" to "ABC")
```
14:22:22.675 - D/WordPress-MAIN: malinjir: change in editor detected.
14:22:38.308 - D/WordPress-MAIN: malinjir: updating from editor: ABC
14:22:38.320 - D/WordPress-MAIN: malinjir: Right before the post is saved into db: ABC
[...]

Issue #13169, opened 2020-10-19
Editor: Post contains local changes even when the user doesn't modify the post


malinajirka:
### Expected behavior
When I open a post in the gutenberg editor and I close it I expect the post won't contain any local changes.

### Actual behavior
When I open a post in the gutenebrg editor and I close it the post contains local changes.

### Steps to reproduce the behavior 
1. My Site
2. Blog Posts
3. Open a published post (eg. [this long post](https://gist.github.com/mchowning/61f1dadc6d3829436ba8531c1e1803ab)) - it happens on some posts only after you click into the content. However, it happens on this post even when I just open it and close it)
4. Press back
5. Notice the post contains "Local changes" label

##### Tested on [device], Android [version], WPAndroid [version]
Emulator API 29

cc @mchowning you recently played with saving/exiting editor
cc @planarvoid you also have context. Would you have time to look into it during groundskeeping?

hypest:
Looks related to the ticket opened a few days ago in the gb-mobile repo: https://github.com/wordpress-mobile/gutenberg-mobile/issues/2714

malinajirka:
Thanks for the info, I've missed that. They look similar, however, I haven't even need to scroll + this issue isn't about autosave but about save on exit. Having said all that there is a pretty high chance these two tickets are related.

planarvoid:
Since you're also @mchowning doing groundskeeping this week, is it ok if I'll leave this to you?

mchowning:
@mkevins has noted a similar-sounding issue [here](https://github.com/wordpress-mobile/WordPress-Android/pull/13398#pullrequestreview-530992031):

> Another observation [...] is that simply opening and existing the post (without making any changes) and then closing it results in the autosave mechanism kicking in.

mchowning:
I have spent some time investigating this, and it appears to be due to certain blocks having outdated or otherwise unexpected (but valid) formatting in the post content. I (obviously) still need to do some more investigation into this, and I do plan to follow-up on this.

mchowning:
I suspect this may be related to https://github.com/wordpress-mobile/gutenberg-mobile/issues/894.

malinajirka:
A note which might be useful - I noticed the EditorFragment.getTitleOrContentChanged() is invoked when you open a post. It's also interesting that it is invoked on the same post every time you open it (even if you open it after it already has the "local changes" label).

mchowning:
[...]

PR #13398, opened 2020-11-16
Add timeout to save dialog

mchowning:
The save dialog when exiting the editor should always be dismissed, but it's a blocking dialog, so with this PR I want to make it so that if we don't dismiss it for some reason, the user isn't left with an unresponsive app (like what happened with https://github.com/wordpress-mobile/WordPress-Android/issues/13379).

I found that saving [a very long post](https://gist.github.com/mchowning/366c011db0ccfefaec703d0e4779525f) on my Pixel 1 took 10-16 seconds on release builds (debug builds took roughly twice as long 😱 ), so I set the timeout here for 20 seconds. I don't feel strongly about that number at all. If the timeout passes before the save is finished, the save dialog just gets dismissed and the user is able to use the editor until the save is finished, at which point the editor resumes whatever action was waiting on the save (probably exiting the editor).

I set this up so that if the timeout is used we send an event to Sentry. That way we can both (a) be able to get an idea if our timeout is too short, and (b) be able to see if there is a sudden spike in the usage of the timeout, which would indicate that we have a bug where we're failing to properly dismiss the save dialog. The exception that is thrown will look like [this in Sentry](https://sentry.io/share/issue/5487f9dec99746f0979f0e223f2c1ef9/) (ignore the fact that the timeout in the Sentry exception is different from the timeout in this PR--I was testing some different timeouts).

Note that I have included [@malinajirka 's fix for the blocking save dialog when previewing](https://github.com/wordpress-mobile/WordPress-Android/pull/13381) in order to facilitate testing this since that fix [has not been merged from the 16.1 release back to develop yet](https://github.com/wordpress-mobile/WordPress-Android/pull/13389).

### To test

Because it is helpful to trigger the save-on-exit behavior when testing this, and the save-on-exit causes a crash in Debug builds (the app is set up to crash if the save-on-exit is triggered because that is an indicator that there may be a problem with the autosave mechanism) I suggest creating a VanillaRelease build to test this PR. Alternatively, you could avoid that crash by creating a debug build with [the code that throws the save-on-exit exception](https://github.com/wordpress-mobile/WordPress-Android/blob/d05207745d0fc034b40af8aa8ff86edab22c39a2/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L1781-L1784) commented out.

#### 1. Non-firing timeout causes no issues
1. Make a VanillaRelease build with this PR
2. Open a post
3. Make some changes to the post and quickly (before the 0.5 second autosave activates) preview the post  (if you open a saved post you'll encounter #12570, which means the first time you preview the post it will only save the post and not actually open the preview)
4. Observe the save dialog is presented and dismissed appropriately
5. Make some more changes and quickly exit the editor before the autosave activates.
6. Observe that the save dialog is presented and dismissed appropriately.
7. Reopen the post
8. Make some changes
9. Allow time for the autosave to finish
10. Exit the post
11. Observe the save dialog is presented and (quickly) dismissed.

#### 2. Firing timeout causes no issues
1. Create a VanillaRelease build where you set the timeout to be very short (maybe 500ms) so that it will fire every time.
2. Open a post (a longer one will help make sure you hit the timeout).
3. Make some changes
4. Quickly exit the post before the autosave kicks in
5. Observe the save dialog appears and is dismissed due to the timeout, after which you are left on the editor until the save finishes, at which point the editor closes.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
@@ -2038,6 +2038,7 @@ public class EditPostActivity extends LocaleAwareActivity implements
             AccountModel account = mAccountStore.getAccount();
             // prompt user to verify e-mail before publishing
             if (!account.getEmailVerified()) {
+                mViewModel.hideSavingDialog();
                 String message = TextUtils.isEmpty(account.getEmail())
                         ? getString(R.string.editor_confirm_email_prompt_message)
                         : String.format(getString(R.string.editor_confirm_email_prompt_message_with_email),
@@ -2059,6 +2060,7 @@ public class EditPostActivity extends LocaleAwareActivity implements
                 return;
             }
             if (!mPostUtils.isPublishable(mEditPostRepository.getPost())) {
+                mViewModel.hideSavingDialog();
                 // TODO we don't want to show "publish" message when the user clicked on eg. save
                 mEditPostRepository.updateStatusFromPostSnapshotWhenEditorOpened();
                 EditPostActivity.this.runOnUiThread(() -> {
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
@@ -1883,6 +1883,9 @@ public class EditPostActivity extends LocaleAwareActivity implements
                     AppLog.e(T.EDITOR, s);
                 }
             });
+        } else if (mEditorFragment instanceof GutenbergEditorFragment) {
+            GutenbergEditorFragment gutenbergEditorFragment = (GutenbergEditorFragment) mEditorFragment;
+            gutenbergEditorFragment.setExternalLogger(e -> mCrashLogging.reportException(e));
         }
     }
 
--- a/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
+++ b/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
@@ -104,7 +104,7 @@ public class GutenbergEditorFragment extends EditorFragmentAbstract implements
 
     private static final int UNSUPPORTED_BLOCK_REQUEST_CODE = 1001;
 
-    private static final int SAVE_DIALOG_TIMEOUT_DURATION = 15000;
+    private static final int SAVE_DIALOG_TIMEOUT_DURATION = 20000;
 
     private boolean mHtmlModeEnabled;
 
--- a/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
+++ b/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
@@ -104,10 +104,26 @@ public class GutenbergEditorFragment extends EditorFragmentAbstract implements
 
     private static final int UNSUPPORTED_BLOCK_REQUEST_CODE = 1001;
 
+    private static final int SAVE_DIALOG_TIMEOUT_DURATION = 15000;
+
     private boolean mHtmlModeEnabled;
 
     private Handler mInvalidateOptionsHandler;
     private Runnable mInvalidateOptionsRunnable;
+
+    @Nullable private GutenbergExternalLogger mExternalLogger;
+
+    private final Handler mProgressDialogTimeoutHandler = new Handler(Looper.getMainLooper());
+    private final Runnable mProgressDialogTimeoutRunnable = () -> {
+        // This timeout only fires when the dialog has been stuck open for a long time, so
+        // let Sentry know something went wrong
+        if (mExternalLogger != null) {
+            SaveDialogTimeoutException exception = new SaveDialogTimeoutException(SAVE_DIALOG_TIMEOUT_DURATION);
+            mExternalLogger.reportException(exception);
+        }
+
+        hideSavingProgressDialog();
+    };
 
     private LiveTextWatcher mTextWatcher = new LiveTextWatcher();
     private int mStoryBlockEditRequestCode;
@@ -1064,11 +1080,13 @@ public class GutenbergEditorFragment extends EditorFragmentAbstract implements
             mSavingContentProgressDialog.setIndeterminate(true);
             mSavingContentProgressDialog.setMessage(getActivity().getString(R.string.long_post_dlg_saving));
         }
+        mProgressDialogTimeoutHandler.postDelayed(mProgressDialogTimeoutRunnable, SAVE_DIALOG_TIMEOUT_DURATION);
         mSavingContentProgressDialog.show();
         return true;
     }
 
     private boolean hideSavingProgressDialog() {
+        mProgressDialogTimeoutHandler.removeCallbacks(mProgressDialogTimeoutRunnable);
         if (mSavingContentProgressDialog != null && mSavingContentProgressDialog.isShowing()) {
             mSavingContentProgressDialog.dismiss();
             return true;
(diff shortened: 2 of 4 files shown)


mkevins (commented):
Reviewed, and this looks good from a code perspective. :+1: 

I also tested this on a Pixel 3a (physical device), and for some reason, when I got to step 6 of the first test, I did not observe that the changes were saved (it also didn't seem to consume all 20 seconds of the timeout :man_shrugging: ). I'm not sure whether this is related to this PR though. In the case where I reduced the timeout to 500 ms, the autosave seemed to work (after the timeout).

Autosave fails on fast exist | Autosave completes after fast timeout
-|-
autosave-fails-on-fast-exit|autosave-completes-after-timeout

Another observation (and quite likely not related to this PR, but mentioning for completeness) is that simply opening and existing the post (without making any changes) and then closing it results in the autosave mechanism kicking in. Is this expected?

loremattei:
Hey! I'm moving this to 16.3 has 16.2 has been cut. If you want this to make it to 16.2, please feel free to ping me!

mchowning:
Thanks for the review @mkevins ! 🙇 

> I also tested this on a Pixel 3a (physical device), and for some reason, when I got to step 6 of the first test, I did not observe that the changes were saved (it also didn't seem to consume all 20 seconds of the timeout 🤷‍♂️ ). I'm not sure whether this is related to this PR though. In the case where I reduced the timeout to 500 ms, the autosave seemed to work (after the timeout).

That's interesting. I'm having trouble recreating that. I'm guessing (based on the fact that you were able to record the issue) you were able to reproduce it pretty consistently. **Could you give me some more details about the setup where you recreated the issue?** I doubt it's related to the changes in this PR, but it is quite concerning, and I want to follow-up on it to make sure.

> Another observation (and quite likely not related to this PR, but mentioning for completeness) is that simply opening and existing the post (without making any changes) and then closing it results in the autosave mechanism kicking in. Is this expected?

Nice catch. That sounds a lot like https://github.com/wordpress-mobile/WordPress-Android/issues/13169. Either way, I agree that this issue isn't related to this PR, but let me know if you think it's a different issue from the one I linked.

mkevins:
> you were able to reproduce it pretty consistently

Yeah, it was consistently reproducible. I was testing on a physical device (Pixel 3a), with a checkout at `b48a8260cdf01b4b6de37a3ffda1e34abcc0aa8d`, and the following changes:

```diff
diff --git a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
index e522c1f9b9..764e1dc239 100644
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
@@ -1779,9 +1779,9 @@ public class EditPostActivity extends LocaleAwareActivity implements
             if (result == Updated.INSTANCE) {
                 SaveOnExitException exception = SaveOnExitException.Companion.build(isAutosavePending);
                 if (BuildConfig.DEBUG) {
-                    throw new RuntimeException(
-                            "Debug build crash: " + exception.getMessage() + " This only crashes on debug builds."
-                    );
+//                    throw new RuntimeException(
+//                            "Debug build crash: " + exception.getMessage() + " This only crashes on debug builds."
+//                    );
                 } else {
                     mCrashLogging.reportException(exception, T.EDITOR.toString());
                     AppLog.e(T.EDITOR, exception.getMessage());
```

I observed this on small posts, but also with the example content you provided [here](https://gist.github.com/mchowning/366c011db0ccfefaec703d0e4779525f). What was surpising about that is that the "saving" dialog didn't wait for 20 seconds, instead only briefly appearing and dismissing (even though the content was apparently not saved).

For the second test, I changed the following code:

```diff
diff --git a/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java b/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
index cefa2b18c4..4d9f41131f 100644
--- a/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
+++ b/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
@@ -104,7 +104,7 @@ public class GutenbergEditorFragment extends EditorFragmentAbstract implements
 
     private static final int UNSUPPORTED_BLOCK_REQUEST_CODE = 1001;
 
-    private static final int SAVE_DIALOG_TIMEOUT_DURATION = 20000;
+    private static final int SAVE_DIALOG_TIMEOUT_DURATION = 500;
 
     private boolean mHtmlModeEnabled;
 
```

Changing and exiting before autosave could complete resulted in the dialog appearing and dismissing, and as expected per the description, the editor remains open until the saves completes, and then closes on its own.

Let me know if there's anything else I can do to troubleshoot the issue.

mchowning:
> Yeah, it was consistently reproducible. I was testing on a physical device (Pixel 3a), with a checkout at b48a8260cdf01b4b6de37a3ffda1e34abcc0aa8d, and the following changes

Thanks for the details @mkevins ! I've tried this on a hardware Pixel 1, hardware Pixel 3, and a Pixel 3a emulator, and I still can't seem to recreate the issue. 😞  What would you think about testing this on a build from `develop`, and if you can recreate the issue there then at least we know it's not an issue with this PR?

mkevins:
> if you can recreate the issue there then at least we know it's not an issue with this PR

I've recreated the issue on `develop` (at `077b419e41e51d1bfb998ec4735f04ef7c42d1ee`), and also noted that it is not very consistent.

My steps were:

1. Open the large post mentioned above
2. Add the word "test" to the heading near the top of the post
3. Quickly exit out of the post before the autosave is triggered

Then: observe the summary in the post list, and see whether the content was modified

[Here is an example of the save on exit successfully completing](https://cloudup.com/i4AdlbiyG0X), and [here is an example of the save failing to complete](https://cloudup.com/ink58TQ9q_E).

When I first attempted this flow on `develop`, the save on exit failed the first time, then succeeded every other time afterwards. When attempting to record this, I was unable to reproduce the failure until I reinstalled the app. After reinstalling, I attempted the steps 3 times, and it failed each time (observed in the failure video linked above :point_up_2: ). On a second attempted recording (I was aiming to show in one "take" the inconsistency), the save was successful (the successful video linked above).

I'm not sure if there is a timing issue (perhaps some race condition) or whether it has anything to do with the state of the app upon a fresh install, but I felt those details have the potential to be clues.

mchowning:
Thanks so much for the continued investigation (and videos 😄) @mkevins ! Since you've also seen the issue on `develop`, what do you think about going ahead and getting this PR merged (assuming you have no other questions or concerns), and opening an issue for the failure-to-save problem with your findings? I definitely think we need to follow-up on that, I'm just not sure it needs to be in this PR.

mkevins:
> what do you think about going ahead and getting this PR merged (assuming you have no other questions or concerns)

My only concern is that I wasn't able to complete the testing steps as a result of the issue(s) described. I agree that those issues can be investigated / solved in a separate PR, but I'm not sure it makes sense to go forward here without confirmation of the testing steps. Is there a way to test these changes in a way that works around the other issues?

mchowning:
Thanks @mkevins . Like we discussed in Slack, I didn't realize this was consistently failing to save for you. That's very concerning! 😨 I spent some more time trying to recreate this without any luck (with/without metro server, fresh install, etc.). 

Would you mind creating an apk where you see the issue and sending it my way? That may help me narrow down if we're seeing different results because of something different about our builds, or if it's a device issue.

loremattei:

mkevins:
Hey @mchowning :wave: :smile: I'm going to take a closer look at this and [a related issue](https://github.com/wordpress-mobile/WordPress-Android/issues/13361) this week (during groundskeeping).

mchowning:
I'm not actively working on this, so I'm closing this for the time being.

PR #13414, opened 2020-11-18
Provide different error message when input is empty

mchowning:
Previously, if a user tried to "submit" from the suggestions UI when they had typed nothing in the suggestions field a toast would appear with the message "+ is not a valid crosspost", which struck me as a bit confusing. The message makes sense when the user has typed something: "+something is not a valid crosspost", so this PR just adds a different message ("Please type to filter the list of suggestions.") to be used when the suggestions filter is empty.

This PR relies on the changes in the `xposts_suggestions` branch, so I'm targeting that branch. I just wanted to extract these changes out so it would be easy to review them independently before adding them to the `xposts_suggestions` branch.

### To Test

This behavior should be the same for both xposting and @-mentions

1. Open a site that is capable of suggestions
2. Open the suggestions UI
3. Without typing anything, tap the Enter/Checkmark key on the on-screen keyboard
4. Observe a toast is presented with the text: "Please type to filter the list of suggestions"
5. Add some text to filter the list of suggestions so that there are either 0 or more than 1 suggestions (i.e., anything but a single matching suggestion)
6. Tap the Enter/Checkmark key on the on-screen keyboard
7. Observe a toast is presented informing that the entered text is not a valid suggestion
8. Update the text so it matches exactly 1 suggestion
9. Tap the Enter/Checkmark key on the on-screen keyboard
10. Observe the suggestion is added to the post.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/suggestion/SuggestionActivity.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/suggestion/SuggestionActivity.kt
@@ -81,7 +81,8 @@ class SuggestionActivity : LocaleAwareActivity() {
             setOnKeyListener { _, keyCode, event ->
                 if (event.action == KeyEvent.ACTION_UP) {
                     if (keyCode == KeyEvent.KEYCODE_ENTER) {
-                        return@setOnKeyListener exitIfOnlyOneMatchingUser()
+                        exitIfOnlyOneMatchingUser()
+                        return@setOnKeyListener true
                     }
                 }
                 false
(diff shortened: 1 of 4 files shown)


guarani (approved):
Looks good to me, I have one question though: Is it expected that the message (toast) causes the keyboard to close? I presume the keyboard is closed to make room for the toast message, but want to make sure.

| Screenshot showing keyboard closing when toast is shown  |
| - |

mchowning:
> Is it expected that the message (toast) causes the keyboard to close? I presume the keyboard is closed to make room for the toast message, but want to make sure.

Good catch @guarani ! I think that should be working better now if you'd care to give it another go.

guarani:
> I think that should be working better now if you'd care to give it another go.

@mchowning sorry for not getting to this sooner, I just noticed today that I still hadn't given it another test 🤦 .

I tested the latest build 83915 and the keyboard is no longer closing and all else looks good to me!

mchowning:
Thanks for testing this @guarani ! 🙇 I think it's going to be simpler if I just cherry-pick these changes into the `xpost_suggestions` branch, so I'll close this PR.

guarani:
Sounds good @mchowning!

PR #13416, opened 2020-11-18
Fix duplicate media upload issue

### LINKED ISSUE #11722: Image uploads multiple times even after it is uploaded once
### Expected behavior
If I accidentally close the app while an image is uploading, the app should attempt to upload the image on next re-launch. Once uploaded there shouldn't be any further attempt.

### Actual behavior
If I accidentally close the app while an image is uploading, the app attempts to upload the image on next re-launch. Once uploaded, the app re-attempts to upload that image on next relaunch resulting in multiple images upload.

[Link to video](https://www.dropbox.com/s/0gs943y1tizkafa/Screen%20Recording%202020-04-22%20at%201.22.13%20PM.mov?dl=0)

ashiagr:
Fixes #11722

This PR removes media in `MediaUploadState.UPLOADED` state from the list of (to be uploaded) media received in the `UploadService` on app restart.

#### Findings

When a media is uploaded for a post and the app is restarted, sometimes `UploadService` receives an intent that contains the original media in the `MediaUploadState.QUEUED` state. The app start re-uploading it as it is still not in the `MediaUploadState.UPLOADED` state.
This results in duplicate media in the site's Media section while the post still contains only the original image. 

https://cloudup.com/i05aJ0_x9nd

|Timestamp | Description |
|------------|-------------|
|`00:19 - 00:23` | Image is uploaded  |
|`00:23 - 00:27` | App is killed  |
|`00:32 - 00:36` | App starts re-uploading the image on relaunch |
|`00:51` | Media section displays duplicate images |

#### Solution

Media received in the `UploadService` intent is now reloaded from the DB to check the latest upload status and if found in the `MediaUploadState.UPLOADED` state, it is removed from the (to be uploaded) media list.

#### To test

- Launch the app
- Upload an image on a post using the Gutenberg editor
- After upload completes, wait for few seconds and kill the app
- Restart the app 
- Notice that uploading arrow is not shown in the status bar
- Go to media browser for the site
- Notice that it contains no unneeded duplicate images 

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadService.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadService.java
@@ -205,15 +205,18 @@ public class UploadService extends Service {
             // if this media belongs to some post, register such Post
             registerPostModelsForMedia(mediaList, intent.getBooleanExtra(KEY_SHOULD_RETRY, false));
 
-            ArrayList<MediaModel> toBeUploadedMediaList = new ArrayList<>(mediaList);
+            ArrayList<MediaModel> toBeUploadedMediaList = new ArrayList<>();
             for (MediaModel media : mediaList) {
-                // Reload media and if already uploaded, remove it from the media list to be uploaded
                 MediaModel localMedia = mMediaStore.getMediaWithLocalId(media.getId());
-                if (localMedia != null
-                    && MediaUploadState.fromString(localMedia.getUploadState()) == MediaUploadState.UPLOADED) {
-                    toBeUploadedMediaList.remove(media);
-                    continue;
+                boolean notUploadedYet = localMedia != null
+                                         && MediaUploadState.fromString(localMedia.getUploadState())
+                                            != MediaUploadState.UPLOADED;
+                if (notUploadedYet) {
+                    toBeUploadedMediaList.add(localMedia);
                 }
+            }
+
+            for (MediaModel media : toBeUploadedMediaList) {
                 mMediaUploadHandler.upload(media);
             }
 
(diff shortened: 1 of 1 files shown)


WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadService.java
```diff
@@ -205,10 +205,21 @@ private void unpackMediaIntent(@NonNull Intent intent) {
             // if this media belongs to some post, register such Post
             registerPostModelsForMedia(mediaList, intent.getBooleanExtra(KEY_SHOULD_RETRY, false));
 
+            ArrayList<MediaModel> toBeUploadedMediaList = new ArrayList<>(mediaList);
```
malinajirka: I'm wondering if we could improve clarify of this code. There are some things I found a bit confusing while parsing it:
- "toBeUploadedMediaList" is initialized with media we might not want to upload
- "toBeUploadedMediaList" is not used to decide what we want to upload, it's used only to show the notification

I'd consider the following:
- initialize "toBeUploadedMediaList" as an empty list
- if the media was already uploaded, don't even add it to the list
- iterate over "toBeUploadedMediaList" and enqueue all items from this list

Something like this:
```
ArrayList<MediaModel> toBeUploadedMediaList = new ArrayList<>();
            for (MediaModel media : mediaList) {
                MediaModel localMedia = mMediaStore.getMediaWithLocalId(media.getId());
                boolean notUploadedYet = localMedia != null
                                               && MediaUploadState.fromString(localMedia.getUploadState())
                                                  != MediaUploadState.UPLOADED;
                if (notUploadedYet) {
                    toBeUploadedMediaList.add(localMedia);
                }
            }

            for (MediaModel media : toBeUploadedMediaList) {
                mMediaUploadHandler.upload(media);
            }

            if (!toBeUploadedMediaList.isEmpty()) {
                mPostUploadNotifier.addMediaInfoToForegroundNotification(toBeUploadedMediaList);
            }
```
Wdyt?

malinajirka (commented):
Really great job finding the root cause of this issue @ashiagr! I've left a suggestion, let me know what you think.

WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadService.java
ashiagr: Makes sense :)! Done in b78ebcc.

malinajirka (approved):
Thanks, works as expected ;)

PR #13420, opened 2020-11-18
Issue/13268 activity type filter screen skeleton

malinajirka:
Partially fixes #13268 

This PR updates Activity type filter screen
- makes the fragment fullscreen
- inits a toolbar
- inits adapter and view holder skeletons

Note: The adapter doesn't contain any data so the screen is completely empty (except of the "X" button)

To test - make sure the ActivityLogFilters feature flag is on in App settings
1. My site -> Activity Log
2. Click on "Activity Type" chip
3. Notice a fullscreen fragment with "X" is shown

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterAdapter.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterAdapter.kt
@@ -12,11 +12,11 @@ class ActivityLogTypeFilterAdapter(private val uiHelpers: UiHelpers) : Adapter<A
         return ActivityLogTypeFilterViewHolder(parent, uiHelpers)
     }
 
-    override fun getItemCount(): Int = items.size
-
     override fun onBindViewHolder(holder: ActivityLogTypeFilterViewHolder, position: Int) {
         holder.onBind(items[position])
     }
+
+    override fun getItemCount(): Int = items.size
 
     fun update(newItems: List<ItemUiState>) {
         items.clear()
--- a/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterAdapter.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterAdapter.kt
@@ -1,0 +1,26 @@
+package org.wordpress.android.ui.activitylog.list.filter
+
+import android.view.ViewGroup
+import androidx.recyclerview.widget.RecyclerView.Adapter
+import org.wordpress.android.ui.activitylog.list.filter.ActivityLogTypeFilterViewModel.ItemUiState
+import org.wordpress.android.ui.utils.UiHelpers
+
+class ActivityLogTypeFilterAdapter(private val uiHelpers: UiHelpers) : Adapter<ActivityLogTypeFilterViewHolder>() {
+    private val items = mutableListOf<ItemUiState>()
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ActivityLogTypeFilterViewHolder {
+        return ActivityLogTypeFilterViewHolder(parent, uiHelpers)
+    }
+
+    override fun getItemCount(): Int = items.size
+
+    override fun onBindViewHolder(holder: ActivityLogTypeFilterViewHolder, position: Int) {
+        holder.onBind(items[position])
+    }
+
+    fun update(newItems: List<ItemUiState>) {
+        items.clear()
+        items.addAll(newItems)
+        notifyDataSetChanged()
+    }
+}
--- a/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
@@ -51,13 +51,12 @@ class ActivityLogTypeFilterFragment : DialogFragment() {
     }
 
     private fun initToolbar(view: View) {
-        val toolbar = view.findViewById(R.id.toolbar_main) as MaterialToolbar
-        toolbar.navigationIcon = ColorUtils.applyTintToDrawable(
-                toolbar.context, R.drawable.ic_close_white_24dp,
-                toolbar.context.getColorResIdFromAttribute(R.attr.colorOnSurface)
+        toolbar_main.navigationIcon = ColorUtils.applyTintToDrawable(
+                toolbar_main.context, R.drawable.ic_close_white_24dp,
+                toolbar_main.context.getColorResIdFromAttribute(R.attr.colorOnSurface)
         )
-        toolbar.setNavigationContentDescription(R.string.close_dialog_button_desc)
-        toolbar.setNavigationOnClickListener { dismiss() }
+        toolbar_main.setNavigationContentDescription(R.string.close_dialog_button_desc)
+        toolbar_main.setNavigationOnClickListener { dismiss() }
     }
 
     private fun initRecyclerView() {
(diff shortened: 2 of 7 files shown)


WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterAdapter.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.wordpress.android.ui.utils.UiHelpers
+
+class ActivityLogTypeFilterAdapter(private val uiHelpers: UiHelpers) : Adapter<ActivityLogTypeFilterViewHolder>() {
+    private val items = mutableListOf<ItemUiState>()
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ActivityLogTypeFilterViewHolder {
+        return ActivityLogTypeFilterViewHolder(parent, uiHelpers)
+    }
+
+    override fun getItemCount(): Int = items.size
```
ParaskP7: Minor: I suggest moving the `getItemCount` below the `onBindViewHolder(...)` function for the below reasons:
1) The `getItemCount` function is on line 7165 of the parent function, while the rest two are on top, which will help preserve the general order. Wdyt?
2) Having `onCreateViewHolder(...)` and `onBindViewHolder(...)` close to each other, and thus grouped, helps with adapter readability. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
```diff
@@ -31,6 +39,35 @@ class ActivityLogTypeFilterFragment : DialogFragment() {
     }
 
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
-        return inflater.inflate(layout.activity_log_type_filter_fragment, container, false)
+        return inflater.inflate(R.layout.activity_log_type_filter_fragment, container, false)
+    }
+
+    override fun getTheme(): Int {
```
ParaskP7: 1) Minor: This function can be converted to an expression body, thus winning us a couple of lines.
1) Minor: I also suggest moving the `getTheme` above the `onCreateView(...)` function for the same reasons I suggested the same for the adapter in my previous comment.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
```diff
@@ -31,6 +39,35 @@ class ActivityLogTypeFilterFragment : DialogFragment() {
...
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        initToolbar(view)
+        initRecyclerView()
+    }
+
+    private fun initToolbar(view: View) {
+        val toolbar = view.findViewById(R.id.toolbar_main) as MaterialToolbar
```
ParaskP7: Question: For `recycler_view` the Kotlin Android Extensions synthetic is used. However, in this case `findViewById(...)` is used instead. Why is that? I would suggest one or the other for consistency purposes. Wdyt?

PS: I am inclining more on using synthetics for now (although deprecated) in order to make any future migration a bit easier. Also, this will help us be consistent with the rest of the codebase. I am guessing synthetics are now being utilised more than `findViewById(...)`, right? If not, then the opposite might be better. Not sure. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
```diff
@@ -31,6 +39,35 @@ class ActivityLogTypeFilterFragment : DialogFragment() {
...
+        toolbar.navigationIcon = ColorUtils.applyTintToDrawable(
+                toolbar.context, R.drawable.ic_close_white_24dp,
+                toolbar.context.getColorResIdFromAttribute(R.attr.colorOnSurface)
+        )
+        toolbar.setNavigationContentDescription(R.string.close_dialog_button_desc)
+        toolbar.setNavigationOnClickListener { dismiss() }
+    }
+
+    private fun initRecyclerView() {
+        recycler_view.layoutManager = LinearLayoutManager(context, RecyclerView.VERTICAL, false)
```
ParaskP7, line 62: Question: Is there a reason why setting a layout manager is not done on the xml level instead?

WordPress/src/main/res/layout/activity_log_type_filter_fragment.xml
```diff
@@ -1,13 +1,30 @@
...
+
+        <com.google.android.material.appbar.MaterialToolbar
+            android:id="@+id/toolbar_main"
+            style="@style/WordPress.ToolBar"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content" />
+
+    </com.google.android.material.appbar.AppBarLayout>
+
+    <androidx.recyclerview.widget.RecyclerView
```
ParaskP7: Minor: I am not sure why is that, but on my IDE, when reformat file is run on this layout it moves the `RecyclerVIew` on top of the `AppBarLayout`. Can you try it on yours as well?

If that is the case maybe it is better to go with what reformat suggests so that future reformats become easier.

WordPress/src/main/res/layout/activity_log_type_filter_item.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout
```
ParaskP7: Minor: This file can be reformatted.

ParaskP7 (commented):
Good job @malinajirka and nice commits! 🌟 

Thanks for including me in this PR! 🙏 

Only minor and questions from my side (as always)!

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterAdapter.kt
malinajirka: Good call - fixed in 0bb021dd572584d1b349c60bdf00da21cc0a904e

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
malinajirka: Fixed in 559082bd9999f98a991a84c76fbd8be6ac542b8c

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
malinajirka: Copy pasting mistake :) - fixed in 052cb4497a9cdc6a8883808f084520e98848a69d

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
malinajirka: I don't think there is any specific reason, but we mostly use this programatic approach throughout the codebase.

WordPress/src/main/res/layout/activity_log_type_filter_fragment.xml
malinajirka: Yeah, I noticed this while working on the feature. I wanted to keep the order consistent with other layouts. But I agree that having to format the file manually is probably worse than having it a bit inconsistent with the rest. Tbh I'm not sure why the formatter prefers this order 🤷 . Fixed in 3e1d186f208a8981f3894e5b2039605f0636e4c6

WordPress/src/main/res/layout/activity_log_type_filter_item.xml
malinajirka: Fixed in cdd2759f98e31aefebe936d5b6e23a2039b1badc

malinajirka:
@ParaskP7 Thanks so much for the review and great tips!

ParaskP7:
OMG @malinajirka , you are the Mr. Speedy Gonzales indeed, thank you so much for taking care of all that! ❤️

ashiagr (approved):
Thanks both of you! All looks good to me 👍.

PR #13421, opened 2020-11-18
Load media library without storage permissions

planarvoid:
We've implemented a fix to not start loading data when the app doesn't have the storage permissions (this was crashing on one testing Samsung tablet). However, this didn't count with the use cases that don't require storage permissions. That's why when you have a fresh install of the app, the WP media library is not populated.

To test:
- Start with a freshly installed app
- Add a new post in GB editor
- Add an image block
- Click on "Add Image" 
- Click on "WordPress Media Library"
- Notice the images are populated

PR submission checklist:


develric (approved):
Hey @planarvoid modification makes sense and works as described. 

I reproduced the issue using develop and checked that the fix works using the CI build from this PR on Pixel 3 Emu API 29. :shipit:

Issue #13329, opened 2020-11-09
Jetpack Section: Backup Download


zwarm:
Parent #13267 

<img width="1779" alt="Screen Shot" src="https://user-images.githubusercontent.com/506707/100615590-cb9e9780-32e5-11eb-8793-b8394e82e36c.png">

**Design**
TBD

**Tasks**

- [x] Introduce "prepare for download endpoint" - [FluxC Issue](https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/1764) - [FluxC PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1766) 

- [x] Introduce support for checking status of a backup download preparation - [FluxC Issue ](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1766) [FluxC PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1772)
- [x] Create BackupDownloadProgressChecker 
- [x] Create BackupDownloadStatusService
- [x] Add backup feature flag
- [x] Turn choose items into recyclerview
- [x] Temporary entry point for backup download feature
- [x] Implement more menu in activity log for restore/backup choices
- [x] FluxC - migration backup table details into WellSqlConfig
- [x] FluxC - Update progress endpoints to include file message and current entry
- [x] FluxC - Enhance fetch backup download state to account for empty response [WordPress-Flux-Android#1853](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1853)
- [x] Activity Log List - Show backup download progress item #13926
- [x] Activity Log List - Show backup download and restore progress item at the same time #13926
- [x] Change "additional information" names to `footnote` to better describe the item #13912
- [x] On unchecking all checkbox items on the "Details" screen, we might want to disable the action button.  #13957
- [x] Add error logging anywhere an exception is deliberately thrown #13910 
- [x] Clear DB on null Fetch from API to support edge case where we are caught between running and complete. FluxC [PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1860) and WPAndroid [PR](https://github.com/wordpress-mobile/WordPress-Android/pull/13944)
~~- [ ] Make sure timestamp is shown on backup progress row when triggered from the web ~~ -- Issue has been moved outside of the Jetpack Section package #14083
- [x] Add notice to download the backup to ActivityLog & Backup list views. See issue #14030  @zwarm 
[...]

PR #13425, opened 2020-11-18
Add backup download progress checker

zwarm:
Parent #13329 

This PR introduces `BackupDownloadProgressChecker` and `BackupDownloadProgressCheckerTest` which handles checking status of the downloadable backup. Future PRs will use this class to maintain the UI progress indicator. It is modeled after RewindProgressChecker. This is the base logic, there may be changes to the logic once the UI development is underway.

**To test:** 
Feature is not being used yet, CI test success will be sufficient until UI is implemented.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/activitylog/BackupDownloadProgressChecker.kt
```diff
@@ -0,0 +1,61 @@
...
+import org.wordpress.android.fluxc.store.ActivityLogStore.FetchBackupDownloadStatePayload
+import org.wordpress.android.fluxc.store.ActivityLogStore.OnBackupDownloadStatusFetched
+import org.wordpress.android.modules.DEFAULT_SCOPE
+import javax.inject.Inject
+import javax.inject.Named
+import javax.inject.Singleton
+
+@Singleton
+class BackupDownloadProgressChecker
+@Inject constructor(
```
malinajirka: Nitpick: I'd consider removing this line wrap. Unfortunately not all, but most classes in the project put the `constructor` keyword on the same line as the class name. The RewindProgressChecker is not consistent so we might want to update that in this PR too since it's a very minor change. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/activitylog/BackupDownloadProgressChecker.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val activityLogStore: ActivityLogStore,
+    @param:Named(DEFAULT_SCOPE) private val defaultScope: CoroutineScope
+) {
+    companion object {
```
malinajirka: Nitpick: Again something which is consistent with the RewindProgressChecker but AFAIK not with the rest of the codebase. We usually create constants as private vals above the class, not in companion objects.

WordPress/src/test/java/org/wordpress/android/ui/activitylog/BackupDownloadProgressCheckerTest.kt
```diff
@@ -0,0 +1,144 @@
...
+            rewindId = "rewindId",
+            backupPoint = Date(),
+            startedAt = Date(),
+            progress = null,
+            downloadCount = 0,
+            validUntil = Date(),
+            url = "url"
+    )
+
+    private val backupDownloadStatusModel = BackupDownloadStatusModel(
```
malinajirka: Since the previous model is called "finishedBackDownload", we might want to consider naming this one something like "ongoing/runningBackupDownload". 

Also I don't see the purpose of `finishedBackupDownload` and `finishedBackupDownloadStatus` -> why do we need two fields when they contain the exact same values. Could we just remove `finishedBackupDownloadStatus` and use `finishedBackupDownload` instead?

Wdyt?

WordPress/src/test/java/org/wordpress/android/ui/activitylog/BackupDownloadProgressCheckerTest.kt
```diff
@@ -0,0 +1,144 @@
...
+        )
+
+        assertEquals(
+                OnBackupDownloadStatusFetched(FETCH_BACKUP_DOWNLOAD_STATE),
+                onBackupDownloadStatusFetched
+        )
+    }
+
+    @Test
+    fun `on start triggers fetch if backup download in progress`() = runBlocking {
```
malinajirka, line 78: I'd consider adding a one more test -> "on start does not trigger fetch if backup download not in progress" as otherwise this test technically doesn't test the "if backup download in progress" part? Wdyt?

malinajirka (commented):
Thanks @zwarm! I've left a couple of comments.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/BackupDownloadProgressChecker.kt
zwarm: Will do 👍

WordPress/src/main/java/org/wordpress/android/ui/activitylog/BackupDownloadProgressChecker.kt
zwarm: I'll tidy that up. 👍

WordPress/src/test/java/org/wordpress/android/ui/activitylog/BackupDownloadProgressCheckerTest.kt
zwarm: Agree. I was actually considering adding the sealed classes since we don't get a progress = 100. Some of this _may_ be refactored once I get into the actual UI pieces, but for now I will get rid of it. Thanks

WordPress/src/test/java/org/wordpress/android/ui/activitylog/BackupDownloadProgressCheckerTest.kt
zwarm: In 1c79f0d

malinajirka (approved):
Thanks for all the changes! LGTM

PR #13430, opened 2020-11-19
Extract basic dialog with view model

planarvoid:
In order to make the SiteInfoBlock PR smaller I'm extracting this basic dialog and the view model into a separate PR (and adding some tests). This class is not used anywhere yet so it can't really be tested. 

The idea behind this change is to stop using the current BasicFragmentDialog which requires us to pass information back to the fragment through the activity and thus clutters the WPMainActivity.

To test:
- Nothing to test here

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/posts/BasicDialog.kt
```diff
@@ -0,0 +1,93 @@
...
+import androidx.lifecycle.ViewModelProviders
+import com.google.android.material.dialog.MaterialAlertDialogBuilder
+import dagger.android.support.AndroidSupportInjection
+import org.wordpress.android.ui.posts.BasicDialogViewModel.BasicDialogModel
+import javax.inject.Inject
+
+/**
+ * Basic dialog fragment with support for 1,2 or 3 buttons.
+ */
+internal class BasicDialog : AppCompatDialogFragment() {
```
renanferrari: Why do we need this to be `internal`?

WordPress/src/main/java/org/wordpress/android/ui/posts/BasicDialogViewModel.kt
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    fun onDismissByOutsideTouch(tag: String) {
+        _onInteraction.postValue(Event(Dismissed(tag)))
+    }
+
+    @Parcelize
+    data class BasicDialogModel(
+        val tag: String,
+        val title: String?,
```
renanferrari: Since this is nullable, maybe we should set `null` as the default value like the other optional arguments?

```suggestion
        val title: String? = null,
```

renanferrari (approved):
Thanks for the changes, the code looks pretty good and I really like how we're using the `BasicDialogViewModel` here. That said, I left a couple of minor question below. Feel free to merge this after answering them 🙂

planarvoid:
thanks for the comments @renanferrari , I've implemented your suggestions 👍

PR #13431, opened 2020-11-19
Create site info block item

### LINKED ISSUE #13343: Create My Site item and ViewHolder

planarvoid:
Fixes #13343 

I'm sorry for the size of this PR. I've tried extracting unrelated steps into separate PRs and not implementing parts of the logic but the rest seems really coupled together. Thinking back I probably could've left out the items interaction for another PR. What this PR does is that it introduces the `SiteInfoBlock`. That's the block that contains the icon, site name, URL and the `switchSite` button. 
- The icon is clickable and it opens a dialog the same way as it does in the current implementation. The dialog doesn't do anything yet and crashes the app on `TODO` items
- The title is clickable and opens a dialog to change the site title. The dialog interaction doesn't do anything yet and crashes the app on `TODO` items.
- The site URL opens the browser with the URL
- The switch site button opens the Switch Site activity (the result is not yet handled).

To test:
- Turn on the Improved My Site flag and restart the app
- Go to My Site fragment
- Notice the Site Info block is now visible
- Click on the URL 
- Notice the site URL opens in a browser
- Click on Switch site button 
- Notice the switch site activity is opened
- Click on the Title 
- Notice the title change dialog opens (dismiss will crash the app)
- Click on the site icon
- Notice the dialog opens (interaction will crash the app)

PR submission checklist:


renanferrari (commented):
Thanks for the changes! Code looks pretty good and everything seems to be working as expected, but I found a small inconsistency in the size of the site title, which used to be a bit bigger than it is now, as you can see below:

|Before|After|
|----|----|
|before|after|

planarvoid:
good catch @renanferrari , it took me some time to figure this out. It seems that we had to do kind of a hacky layout setup previously to manage the autosizing. In the end I've fixed it with setting min height on the title view in d4885f2 and it seems to work well. Make sure you try also changing the title to be very long (so it takes 2 lines of code instead). In that case it should downsize.

Screenshot 2020-11-23 at 16 16 25

renanferrari (approved):
Thanks for the updates, @planarvoid! Everything seems to be working as expected now. 

LGTM :shipit:

PR #13437, opened 2020-11-20
[Gutenberg] File Block -  FileBlockProcessor for upload media url transformation

jd-alexander:
## Findings
This PR is the Android side of the [iOS implementation for updating the HTML content](https://github.com/wordpress-mobile/WordPress-iOS/pull/15339) of the File Block without Gutenberg being open. 

## Solution
The `FileBlockProcessor` replaces all the `<a href>` that contain local file URLs with remote media URLs once the file upload is complete. The id of the File Block is also updated with a remote variant as well. 

@malinajirka while working on this I saw this use case [CleanUpMediaToPostAssociationUseCase](https://github.com/wordpress-mobile/WordPress-Android/blob/7ed34a51eab27dd77aa1b87b948bc296b478deb8/WordPress/src/main/java/org/wordpress/android/ui/posts/editor/media/CleanUpMediaToPostAssociationUseCase.kt) 

Since a file is basically a media item should we modify the regex for finding media related blocks below to include the file block so the use case mentioned above is run for it as well? 

https://github.com/wordpress-mobile/WordPress-Android/blob/e693397bc0cb879ec4fc8ea4efeb6114bb6dd44d/WordPress/src/main/java/org/wordpress/android/ui/posts/PostUtils.java#L444

## Testing

1. Add a file block to a new post.
2. Upload a file to the block (I  lessened the Cellular Network type and Signal strength to replicate a slow network so I  had enough time to close the editor).
3. Close the editor by pressing the back button.
4. Let the upload to finish in the background.
5. When the file is uploaded and the post saved, open the post again (on web or mobile).

Verify that the behavior is functioning correctly by :
  - Checking that the block renders properly.
  - Checking that all references to href are updated to the remote URL.
  - Checking that the file ID has been updated to the remote one.

## Reviewing 

## Submitter Checklist


malinajirka:
>@malinajirka while working on this I saw this use case CleanUpMediaToPostAssociationUseCase
> Since a file is basically a media item should we modify the regex for finding media related blocks below to include the file block so the use case mentioned above is run for it as well?

Great question! I'm not familiar with the file upload flow. However, if we use MediaModel (or similar) which can have a link to a post I agree that re-using the cleanUp usecase makes sense.

mkevins (commented):
Nice work here @jd-alexander :smiley: :+1: 

The code looks good, and I've tested this on a Pixel 3a (physical device), and it is working as expected. :tada: 

The only addition I'd suggest is to add a few unit tests (similar to the ones [here](https://github.com/wordpress-mobile/WordPress-Android/tree/gutenberg/file_block_media_processor/WordPress/src/test/java/org/wordpress/android/ui/posts/mediauploadcompletionprocessors)).

mkevins (approved):
Nice work Joel! Tests look good, and are passing. Once lint errors are fixed, LGTM!

jd-alexander:
Thanks for all the reviews @mkevins were you the one who wrote the regex here?
https://github.com/wordpress-mobile/WordPress-Android/blob/e693397bc0cb879ec4fc8ea4efeb6114bb6dd44d/WordPress/src/main/java/org/wordpress/android/ui/posts/PostUtils.java#L444

Let me know. I need to add regex for the File block but I am still a beginner with regex and I tend to take a bit longer time to write them so in the interest of time I am asking for some help to create one for the File block, Let me know if that's possible. 😄

mkevins:
> Thanks for all the reviews @mkevins were you the one who wrote the regex here?
> https://github.com/wordpress-mobile/WordPress-Android/blob/e693397bc0cb879ec4fc8ea4efeb6114bb6dd44d/WordPress/src/main/java/org/wordpress/android/ui/posts/PostUtils.java#L444
> 
> Let me know. I need to add regex for the File block but I am still a beginner with regex and I tend to take a bit longer time to write them so in the interest of time I am asking for some help to create one for the File block, Let me know if that's possible.

I didn't write that, but I've worked with similar regex in the past, so I'll give it a shot. I've created [a PR targeting this one](https://github.com/wordpress-mobile/WordPress-Android/pull/13458) with what I think the needed changes are. cc: @malinajirka , please let me know if anything else is needed for this.

PR #13439, opened 2020-11-20
Init UiState for activity type filter screen

malinajirka:
Partially fixes #13268 

This PR initializes uistate for activity type filter screen and adds some basic logic. Since the endpoint for getting available activity types isn't implemented yet, it's just a dummy implementation.

Merge instructions:
1. Make sure https://github.com/wordpress-mobile/WordPress-Android/pull/13420 is merged
2. Merge this PR

To test:
Nothing to test, there is no UI yet.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
@@ -88,20 +88,20 @@ class ActivityLogTypeFilterViewModel @Inject constructor(
     }
 
     sealed class UiState {
+        open val contentVisibility = false
         open val loadingVisibility = false
-        open val items: List<ListItemUiState>? = null
-        open val primaryAction: Action? = null
-        open val secondaryAction: Action? = null
 
         object FullscreenLoading : UiState() {
             override val loadingVisibility: Boolean = true
         }
 
         data class Content(
-            override val items: List<ListItemUiState>,
-            override val primaryAction: Action,
-            override val secondaryAction: Action
-        ) : UiState()
+            val items: List<ListItemUiState>,
+            val primaryAction: Action,
+            val secondaryAction: Action
+        ) : UiState() {
+            override val contentVisibility = true
+        }
 
         data class Action(val label: UiString) {
             var action: (() -> Unit)? = null
--- a/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
@@ -17,6 +17,27 @@ class ActivityLogTypeFilterViewModel @Inject constructor(
         isStarted = true
     }
 
+    sealed class UiState {
+        open val loadingVisibility = false
+        open val items: List<ListItemUiState>? = null
+        open val primaryAction: Action? = null
+        open val secondaryAction: Action? = null
+
+        object FullscreenLoading : UiState() {
+            override val loadingVisibility: Boolean = true
+        }
+
+        data class Content(
+            override val items: List<ListItemUiState>,
+            override val primaryAction: Action,
+            override val secondaryAction: Action
+        ) : UiState()
+
+        data class Action(val label: UiString) {
+            var action: (() -> Unit)? = null
+        }
+    }
+
     sealed class ListItemUiState {
         data class SectionHeader(
             val title: UiString
--- a/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
@@ -3,14 +3,24 @@ package org.wordpress.android.ui.activitylog.list.filter
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import okhttp3.internal.immutableListOf
+import org.wordpress.android.R
+import org.wordpress.android.modules.BG_THREAD
 import org.wordpress.android.modules.UI_THREAD
+import org.wordpress.android.ui.activitylog.list.filter.ActivityLogTypeFilterViewModel.UiState.Action
+import org.wordpress.android.ui.activitylog.list.filter.ActivityLogTypeFilterViewModel.UiState.Content
 import org.wordpress.android.ui.activitylog.list.filter.ActivityLogTypeFilterViewModel.UiState.FullscreenLoading
 import org.wordpress.android.ui.utils.UiString
+import org.wordpress.android.ui.utils.UiString.UiStringRes
+import org.wordpress.android.ui.utils.UiString.UiStringText
 import org.wordpress.android.viewmodel.ScopedViewModel
 import javax.inject.Inject
 import javax.inject.Named
 
 class ActivityLogTypeFilterViewModel @Inject constructor(
+    @Named(BG_THREAD) private val bgDispatcher: CoroutineDispatcher,
     @Named(UI_THREAD) private val mainDispatcher: CoroutineDispatcher
 ) : ScopedViewModel(mainDispatcher) {
     private var isStarted = false
@@ -23,6 +33,33 @@ class ActivityLogTypeFilterViewModel @Inject constructor(
         isStarted = true
 
         _uiState.value = FullscreenLoading
+        fetchAvailableActivityTypes()
+    }
+
+    private fun fetchAvailableActivityTypes() {
+        launch {
+            // TODO malinjir initiate the fetch
+            onActivityTypesFetched(immutableListOf(DummyActivityType, DummyActivityType, DummyActivityType))
+        }
+    }
+
+    private suspend fun onActivityTypesFetched(activityTypes: List<DummyActivityType>) {
+        _uiState.value = buildContentUiState(activityTypes)
+    }
+
+    private suspend fun buildContentUiState(activityTypes: List<DummyActivityType>): Content {
+        return withContext(bgDispatcher) {
+            // TODO malinjir replace "it.toString()" with activity type name
+            val activityTypeListItems: List<ListItemUiState.ActivityType> = activityTypes
+                    .map {
+                        ListItemUiState.ActivityType(title = UiStringText(it.toString()))
+                    }
+            Content(
+                    activityTypeListItems,
+                    primaryAction = Action(label = UiStringRes(R.string.activity_log_activity_type_filter_apply)),
+                    secondaryAction = Action(label = UiStringRes(R.string.activity_log_activity_type_filter_clear))
+            )
+        }
     }
 
     sealed class UiState {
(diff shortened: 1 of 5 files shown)


WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
```diff
@@ -1,20 +1,123 @@
...
         isStarted = true
+
+        _uiState.value = FullscreenLoading
+        fetchAvailableActivityTypes()
+    }
+
+    private fun fetchAvailableActivityTypes() {
+        launch {
+            // TODO malinjir initiate the fetch
+            onActivityTypesFetched(immutableListOf(DummyActivityType, DummyActivityType, DummyActivityType))
```
ParaskP7: Questions (❓ ): Why using the `immutableListOf(...)` from `OkHttp`?

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
```diff
@@ -1,20 +1,123 @@
...
+
+    private suspend fun buildContentUiState(activityTypes: List<DummyActivityType>): Content {
+        return withContext(bgDispatcher) {
+            // TODO malinjir replace the hardcoded header title
+            val headerListItem = SectionHeader(UiStringText("Test"))
+            // TODO malinjir replace "it.toString()" with activity type name
+            val activityTypeListItems: List<ListItemUiState.ActivityType> = activityTypes
+                    .map {
+                        ListItemUiState.ActivityType(title = UiStringText(it.toString()))
+                    }
```
ParaskP7: Minor (🔍 ): The `.map { ... }` can become a one-liner, just for readability purposes. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
```diff
@@ -1,20 +1,123 @@
...
+        // TODO malinjir save and dismiss
+    }
+
+    private fun onClearClicked() {
+        (_uiState.value as? Content)?.let { it ->
+            _uiState.value = it.copy(items = uncheckAllActivityTypeItems(it))
+        }
+    }
+
+    private fun uncheckAllActivityTypeItems(it: Content): List<ListItemUiState> {
```
ParaskP7: 1) Minor (🔍 ): The name of this function might suggest something different to what it does. What if this is renamed to `getAllActivityTypeItemsUnchecked(...)`?
2) Minor (🔍 ): Usually, I personally try to pass to a function only the exact amount of information needed so that this function can do its job. For example, in this case I would pass only the list of items and not the full content, since only that is needed for this function to perform its job. Wdyt?
3) Minor (🔍 ): Afterwards, this function can be further simplified by converting it to an expression body. Accordingly to `Kotlin Style Guide` this is still allowed and IMHO simplifies the function a lot:
```
    private fun getAllActivityTypeItemsUnchecked(items: List<ListItemUiState>) = items.map { item ->
        if (item is ListItemUiState.ActivityType) {
            item.copy(checked = false)
        } else {
            item
        }
    }
```

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
```diff
@@ -1,20 +1,123 @@
...
+
+        data class Content(
+            val items: List<ListItemUiState>,
+            val primaryAction: Action,
+            val secondaryAction: Action
+        ) : UiState() {
+            override val contentVisibility = true
+        }
+
+        data class Action(val label: UiString) {
```
ParaskP7: Minor (🔍 ): From Detekt: `MemberNameEqualsClassName: A member is named after a class. This might result in confusion.  Either rename the member or change it to a constructor.`

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
```diff
@@ -1,20 +1,123 @@
...
+            var action: (() -> Unit)? = null
+        }
+    }
+
+    sealed class ListItemUiState {
```
ParaskP7, line 102: Question (❓ ): Since the `Action` data class is part of the `UiState` sealed class, why not having this `ListItemUiState` sealed class as part of it as well? This way, it would be more consistent, don't you think? Else, maybe moving `Action` outside the `UiState` might be another alternative. Or is there a specific logic behind that decision?

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
```diff
@@ -1,22 +1,51 @@
...
+    fun `fullscreen loading shown, when screen initialized`() = test {
+        val uiStates = initObservers().uiStates
+
+        viewModel.start()
+
+        assertThat(uiStates[0]).isInstanceOf(UiState.FullscreenLoading::class.java)
+    }
+
+    @Test
+    fun `section header gets added as first item in the list`() {
```
ParaskP7: Minor (🔍 ): Seeing that you added the `... , when ...` convention on the above test, maybe consider adding the same on this test as well, just for consistency purposes. Wdyt?

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
```diff
@@ -1,22 +1,51 @@
...
+        val uiStates = initObservers().uiStates
```
ParaskP7: Minor (🔍 ): The `uiStates` val is unused here. Maybe just using `initObservers()` is enough for this test as a setup. Wdyt?

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
```diff
@@ -1,22 +1,51 @@
 package org.wordpress.android.ui.activitylog.list.filter
```
ParaskP7, line 1: Question (❓ ) How would the rest of this view model's functionality be tested?

ParaskP7 (commented):
It is so nice so see how the view model is being built, thanks for including me in this PR @malinajirka and great job! 🙏 🌟

Guess what... minors and questions from my side, nothing more, nothing less, as you would expect! 😄

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
malinajirka: Silly mistake, nice catch! ;) Fixed in bdd76bc8fe7da0926b45b6610741d6df6aec9256

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
malinajirka: Fix in 17735265601558f8800973ad76cf2298c187de42

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
malinajirka: I totally agree with all the points - I forgot to polish this method altogether. The `it` parameter name isn't great either :D.
Fixed in 17735265601558f8800973ad76cf2298c187de42

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
malinajirka: I decided to keep this as is. The only reason the parameter is not in the constructor is that two lambdas are not equal unless they reference the exact same instance. I want to use equals in the RV diffUtil, however, implementing a custom equals felt like an overkill.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
malinajirka: I moved it outside of the sealed class in 73eb6e713a71f4483f9ec6d238e3d5b761763130. This is consistent with the structure of the other UiStates in the repo.

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
malinajirka: Fixed in 416893abd3da953154503e839c77a482efbec525

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
malinajirka: Fixed in 416893abd3da953154503e839c77a482efbec525

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
malinajirka: I didn't want to test them in this PR since some of the logic can't be easily mocked yet. Bunch of tests were added in https://github.com/wordpress-mobile/WordPress-Android/pull/13440 and more are coming in the following PRs.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
ParaskP7: 😄 OMG, you are right, I forgot to mention the `it` parameter, all my comment started on that and I forgot to mention it in the end! 😄

PS: I just felt exactly as when I unlocked my phone to check on a notification only to figure out that after 5' minutes I locked my phone doing all those other thingies but never did action on the notification that made me unlock my phone in the first place! Then, I had to unlock my phone a second time to action on that notification, crazy social dilemma stuff! 😅

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
ParaskP7: Yes, yes, thank you! I saw those test in the next PR. ❤️

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
ParaskP7: Totally, I think we should not move it to constructor. The main purpose of this change is about renaming it to something else to avoid confusion. The changing it to a constructor is a secondary though and might not be applicable usually.

ashiagr (approved):
LGTM 👍.

PR #13440, opened 2020-11-20
Error handling on Activity Type filter screen

malinajirka:
Partially fixes #13268

This PR adds error state with retry action to the activity type filter screen.

To test:
There is no UI yet, so there isn't anything to test.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
```diff
@@ -5,48 +5,55 @@ import androidx.lifecycle.MutableLiveData
...
 import javax.inject.Named
+import kotlin.properties.Delegates
 
 class ActivityLogTypeFilterViewModel @Inject constructor(
+    private val dummyActivityTypesProvider: DummyActivityTypesProvider,
     @Named(BG_THREAD) private val bgDispatcher: CoroutineDispatcher,
     @Named(UI_THREAD) private val mainDispatcher: CoroutineDispatcher
 ) : ScopedViewModel(mainDispatcher) {
     private var isStarted = false
+    private var siteId by Delegates.notNull<Long>()
```
ParaskP7, line 29: Question (❓ ): This is very interesting, I never used that before! Is this being used instead of `lateinit`, and if so, how does it help?

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/DummyActivityTypesProvider.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import kotlinx.coroutines.delay
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * This class will be deleted as soon as we implement fetchAvailableActivityTypes endpoint in ActivityLogStore.
+ */
+@Singleton
+class DummyActivityTypesProvider @Inject constructor() {
```
ParaskP7, line 11: Love (❤️ ): Loving it, nice way of implementing stub functionality and proceed forward with the feature!

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
```diff
@@ -1,51 +1,131 @@
 package org.wordpress.android.ui.activitylog.list.filter
```
ParaskP7, line 1: Love (❤️ ): Loving the tests, and actually this PR response to my previous PR are testing view model functionality. Many thanks for that and the dummy provider!

PS: Just as an FYI, on some tests the `..., when ...` is missing.

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
```diff
@@ -1,51 +1,131 @@
...
-        val uiStates = initObservers().uiStates
+    fun `available activity types fetched, when screen initialized`() = test {
+        init()
 
-        viewModel.start()
+        viewModel.start(0L)
 
-        assertThat((viewModel.uiState.value as Content).items[0])
+        verify(dummyActivityTypesProvider).fetchAvailableActivityTypes(anyOrNull())
+        verifyNoMoreInteractions(dummyActivityTypesProvider)
```
ParaskP7: Question (❓ ): How much is `verifyNoMoreInteractions(...)` in general is utilized in our tests and when?

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
```diff
@@ -1,51 +1,131 @@
...
+        viewModel.start(0L)
+
+        assertThat((viewModel.uiState.value as UiState.Content).items[0])
                 .isInstanceOf(ListItemUiState.SectionHeader::class.java)
     }
 
-    private fun initObservers(): Observers {
+    @Test
+    fun `content shown, when fetch available activity types completes successfully`() = test {
+        val uiStates = init(successResponse = true).uiStates
```
ParaskP7: Minor (🔍 ): The `uiStates` is not used in this test, is it?

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
```diff
@@ -1,51 +1,131 @@
...
+
+        assertThat(viewModel.uiState.value).isInstanceOf(UiState.Error::class.java)
+    }
+
+    @Test
+    fun `available activity types fetched, when error retry action invoked`() = test {
+        init(successResponse = false)
+        viewModel.start(0L)
+
+        (viewModel.uiState.value as UiState.Error).retryAction.action!!.invoke()
```
ParaskP7, line 81: Warning (⚠️ ): Although this is test, I usually always suggest an alternative to double explanation marks (`!!`). What if just a question mark is used instead (`?`)? This way the test will not result in an exception (if something goes wrong) but instead it will result in failure. Unless, this is what the intention of the test is, to result in an exception in case something wrong. In that case please let me know and ignore this comment.

PS: The same applies for the test below.

ParaskP7 (commented):
Thanks a lot once again, and thanks for adding the dummy provider and tests, love both! ❤️ 

PS: I added a warning comment (⚠️) but this is test specific, so don't be alarmed. The rest are minors, questions and love giving! 😄

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
malinajirka: Lateinit cannot be used on primitives. However, I removed this in the following PR anyway, since I replaced it with "lateinit var siteId: RemoteId". :D

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
malinajirka: Not much - I add it when I want to test that a certain action results in certain re-action. Eg. when I invoke two methods in a test and I want to make sure that only one of them results in an API request. However, it doesn't make sense in this test - I think I kept updating the test and it stopped making sense after. 

Removed in c2e74b17d3b14ab62f601cd5e3c8e926dad58a64

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
malinajirka: Fixed in 4eda09587ac3a0703d0c0e09c58be08047141f23

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
malinajirka: In my experience "!!" has valid usecases and overusing "?" sometime leads to a harder to find bugs. 

For example, if we use "?" here and the action is null, the test will result in an error saying something like "Wanted 2 times but was 1 time".  The developer will try to understand why we expected 2 calls, they'll need to check if the action's body is invoking the correct method (fetchAvailableActivityTypes), etc.

However, if the tests fails with a NPE in the "act" block of the test, they can see right away what's wrong.

Wdyt?

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
malinajirka: Fixed in 4d51eec8084a0acf0c3ffec5e12f0011a84e255f

> Just as an FYI, on some tests the ..., when ... is missing.

I felt like that if the "Arrange" and "Act" sections don't have any "when", it's ok to omit the "when" in the name. However, it's probably better to always keep it ;).

ParaskP7: Yea, definitely, if your intention is to sometimes avoid the `when`, by all means,  just let me know.  My commenting is purely because I was thinking that it has split your mind by mistake.

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
ParaskP7: Great, great, thanks for the explanation. Yeap, I agree, I used to do this as well but lately I find myself doing less and less this kind of verification and only applying them when indeed it makes total sense and there is no ambiguity of its value. But, my general rule of thumb is that those are usually not needed much.

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
ParaskP7: Cool, cool, I was kind of thinking about that, thanks for letting me know that this was done by design!

I'll give it a though myself to weight pros and cons on the specifics of it later and will come back to you. But, since this how you want it to behave, please disregard my comment and proceed forward by ignoring my comment.

ashiagr (approved):
Nice work @malinajirka, good use of dummy data providers 👍  for the time being.

PR #13424, opened 2020-11-18
[Home Page Picker] Updates the design of the “Choose a Domain” screen

antonis:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2722

## Description
Updates the design of the “Choose a Domain” screen
*Internal reference: `pb3aDo-xN`*

## To test:

<details>
<summary><strong>To disable or enable the development version of Home Page Picker</strong></summary>

- Open the app from the build that allows FeatureFlags such as a PR build or a local development build
    - By default, the modal layout picker is enabled only in development builds.
    - From the site page:
        1. Click on your Gravatar.
        1. Click on App Settings.
        1. Click on Test feature configuration.
        1. Toggle "HomePagePickerFeatureConfig" to enable or disable the picker
        1. Restart the App
</details>

<details>
<summary><strong>To start the Home Page Picker screen</strong></summary>

1. Start the site creation flow (e.g. *Choose site > Add button*)
1. Select the *Create WordPress.com site* option
</details>

*Note: This PR introduces no change in the functional part of the choose domain screen. The tests below just validate that no functionality is broken*

### Search a domain
1. Start the site creation flow
1. Select a design and press *Choose* or press *SKIP*
1. Enter your search query in the input field
1. Validate that you get results

### Choose a domain
1. Follow the steps above
1. Choose a domain
1. Verify that the toolbar appears
1. Press *Choose*
1. Verify that the site is created

## Screenshots

|Initial Screen Before|Initial Screen After|
|---|---|
|before1|after1|

|Search State Before|Search State After|
|---|---|
|before2|after2|

|Select State Before|Select State After|
|---|---|
|before4|device-2020-11-18-221630|

|Landscape Before|Landscape After|
|---|---|
|before3|after3|

|Dark mode Initial Screen|Dark mode Select State|
|---|---|
|device-2020-11-18-220300|device-2020-11-18-221646|

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13409, opened 2020-11-17
[Home Page Picker] Enable the feature

antonis:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2727

## Description
Enables the new Home Page Picker for release

## To test:

*Note: An easy way to test the analytics is to debug the app and filter Logcat to look for the string `🔵 Tracked:`*

### Start the Home Page Picker screen

1. Start the site creation flow (e.g. *Choose site > Add button*)
1. Select the *Create WordPress.com site* option
1. Verify that you see the Choose Design screen
1. In the console expect to see:
`🔵 Tracked: enhanced_site_creation_site_design_viewed`

### Create a site with the default design
1. Start the Home Page Picker 
1. Press **SKIP**
1. In the console expect to see:
`🔵 Tracked: enhanced_site_creation_site_design_skipped`
1. Search for and Choose a domain
1. Tap **Create Site**
1. Expect Your site to be created with the default design

### Create a site with a specific design
1. Start the Home Page Picker 
1. Select a design
1. Press **Choose**
1. In the console expect to see something like (where `alves` is your selected design):
`🔵 Tracked: enhanced_site_creation_site_design_selected, Properties: {"template":"alves"}`
1. Search for and Choose a domain
1. Tap **Create Site**
1. Expect Your site to be created with the selected design

### Error tracking
1. Turn your internet access off (e.g. with flight mode)
1. Start the Home Page Picker
1. In the console expect to see something like:
`🔵 Tracked: enhanced_site_creation_error_shown, Properties: {"error_context":"design","error_description":"Error fetching designs","error_type":"unknown"}`

### Preview Design
1. Start Home Page Picker
1. Select a design
1. Press the **Preview** button on the bottom toolbar
1. Verify that the selected design preview loads
1. Verify that the following events are tracked (where *alves* the design you selected)
```
🔵 Tracked: enhanced_site_creation_site_design_preview_viewed, Properties: {"template":"alves"}
🔵 Tracked: enhanced_site_creation_site_design_preview_loading, Properties: {"template":"alves"}
🔵 Tracked: enhanced_site_creation_site_design_preview_loaded, Properties: {"template":"alves"}
```

### Choose Design from Preview
1. Follow the *Preview Design* steps above
1. Press Choose
1. Search and select a domain
1. Select create site
1. Verify that the created site has the selected design
1. Verify that the following events are tracked (where *alves* the design you selected)
```
[...]

PR #13443, opened 2020-11-20
[Home Page Picker] Move to Page sheet

antonis:
**Fixes** https://github.com/wordpress-mobile/WordPress-iOS/pull/15293#issuecomment-730520774

Depends on: https://github.com/wordpress-mobile/WordPress-Android/pull/13424

## Description
This PR applies a dismissible modal behavior to the site creation flow

## To test:

<details>
<summary><strong>To disable or enable the development version of Home Page Picker</strong></summary>

- Open the app from the build that allows FeatureFlags such as a PR build or a local development build
    - By default, the modal layout picker is enabled only in development builds.
    - From the site page:
        1. Click on your Gravatar.
        1. Click on App Settings.
        1. Click on Test feature configuration.
        1. Toggle "HomePagePickerFeatureConfig" to enable or disable the picker
        1. Restart the App
</details>

<details>
<summary><strong>To start the Home Page Picker screen</strong></summary>

</details>

1. Start the site creation flow
1. In each step verify that the flow is dismissible by swiping down from the top 

## Screenshots

|Normal Mode|Dark Mode|
|---|---|
|device-2020-11-20-210434|device-2020-11-20-210454|

PR submission checklist:


antonis:
Hey @kyleaparker and @SylvesterWilmott 👋 ,

After the discussion between @chipsnyder and @iamthomasbishop [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/15293#issuecomment-730520774) we decided to do a spike on the Android implementation as well.
This PR is still draft and has some rough edges but demonstrates the modal behavior applied. You can compare this with the flow without it [from this PR](https://github.com/wordpress-mobile/WordPress-Android/pull/13409).
In the later only the design preview screen is a dismissible modal.

While developing this I realised that the fact that this flow is inside a modal means that it can be dismissed at any time/step. The user may swipe it down even when the site creation is in progress or when you reach the preview screen. Consequently  the user may miss the result of the site creation (e.g. a failure). By dismissing the modal in the last step the user won’t switch to the newly created site and will stay in the “My Sites” screen.

Let me know if the behavior introduced with this PR is desirable in order to proceed with polishing the implementation.

kyleaparker:
>While developing this I realised that the fact that this flow is inside a modal means that it can be dismissed at any time/step. The user may swipe it down even when the site creation is in progress or when you reach the preview screen. Consequently the user may miss the result of the site creation (e.g. a failure). By dismissing the modal in the last step the user won’t switch to the newly created site and will stay in the “My Sites” screen.

Is it possible to prevent people dismissing the modal once you click the "Create Site" button on the domains screen? Or move to a new page after this step that isn't dismissible. I don't think you should be able to dismiss the flow from this point as you have intentionally created a new site and should see that through. 

If that isn't possible, I don't think the modal is the best approach as I could see it causing quite a bit of confusion. When testing I swiped the modal down while the site was being created and then wasn't sure how to access the site.

antonis:
Hello @kyleaparker 👋 
thank you for your prompt response 🙇 

> Is it possible to prevent people dismissing the modal once you click the "Create Site" button on the domains screen? Or move to a new page after this step that isn't dismissible. I don't think you should be able to dismiss the flow from this point as you have intentionally created a new site and should see that through.

At this point it seems difficult to prevent dismissing after the *Create Site* button is pressed because all the screens in the flow live in the same Activity (except the *Design Preview* that is a modal on top of that). 

> If that isn't possible, I don't think the modal is the best approach as I could see it causing quite a bit of confusion. When testing I swiped the modal down while the site was being created and then wasn't sure how to access the site.

Since achieving the desired behaviour would require non-trivial refactoring of the whole flow I suggest to further investigate this in a later stage to avoid introducing bugs in the *Round 1 deliverable*. Wdyt?
cc @Tug , @chipsnyder

kyleaparker:
>Since achieving the desired behaviour would require non-trivial refactoring of the whole flow I suggest to further investigate this in a later stage to avoid introducing bugs in the Round 1 deliverable. Wdyt?

Agreed!

antonis:
Hello @kyleaparker 👋 

Sorry for the delayed response on this. I revisited the modal transformation of the Site Creation flow.

> Is it possible to prevent people dismissing the modal once you click the "Create Site" button on the domains screen? Or move to a new page after this step that isn't dismissible. I don't think you should be able to dismiss the flow from this point as you have intentionally created a new site and should see that through.

I verified my comment [above](https://github.com/wordpress-mobile/WordPress-Android/pull/13443#issuecomment-732228954) that this would require a lot of refactoring to be achieved. Actually implementation wise the flow will break in two separate parts containing the steps that can and cannot be dismissed respectively.
Given that the site creation flow may change dynamically (with the addition of A/B testing in the order of steps) and future additions I feel that the modal transformation will add unnecessary complexity.

Furthermore the visual result of a modal vs a normal screen on Android is not so different (apart from the animation and the top handle) contrary to iOS that offers a different experience with the stacked screens.

|Android Normal Screen|Android Modal Screen|iOS Modal Screen|
|---|---|---|
|<image width="360" src="https://user-images.githubusercontent.com/304044/101359065-6b3ec580-38a4-11eb-9003-31f0568eeab8.png"/>|<image width="360" src="https://user-images.githubusercontent.com/304044/99840101-93c08300-2b74-11eb-8d78-28a41bbde051.png"/>|<image width="360" src="https://user-images.githubusercontent.com/304044/101358108-0e8edb00-38a3-11eb-80b3-9c97b1a0350e.png"/>|

> If that isn't possible, I don't think the modal is the best approach as I could see it causing quite a bit of confusion. When testing I swiped the modal down while the site was being created and then wasn't sure how to access the site.

My suggestion would be to drop this refactoring and keep the current implementation. Wdyt?

kyleaparker:
>My suggestion would be to drop this refactoring and keep the current implementation. Wdyt?

This works for me 👍

PR #13444, opened 2020-11-20
Add backup feature flag config

zwarm:
Part of #13329 

This PR add a local feature flag for the Jetpack Backup feature. This is the precursor step to using the feature flag during development. 

There is nothing to test in this PR.

PR submission checklist:

### DIFF
--- a/WordPress/build.gradle
+++ b/WordPress/build.gradle
@@ -80,6 +80,7 @@ android {
         buildConfigField "boolean", "SCAN_AVAILABLE", "false"
         buildConfigField "boolean", "ENABLE_FEATURE_CONFIGURATION", "true"
         buildConfigField "boolean", "MY_SITE_IMPROVEMENTS", "false"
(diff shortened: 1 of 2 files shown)


malinajirka (approved):
LGTM, thanks!

ParaskP7 (commented):

Question (❓ ): Based on the previous refactoring that we did on `FluxC`, should we maybe add the `Download` keyword everywhere missing within the definition of this feature flag as well?

zwarm:
>  Should we maybe add the `Download` keyword everywhere missing within the definition of this feature flag as well?

🤔  I don't have a preference either way. @malinajirka  want to weigh in?

malinajirka:
It depends on what exactly do we plan to use it for. If we plan to use it only for the "download backup" feature, then I think it's a good idea to include "download". However, if we plan to use it for the whole backup section, I think the current name is appropriate.

zwarm:
> plan to use it for the whole backup section

Yes, the entire feature. Thanks.

PR #13445, opened 2020-11-21
Issue/13433 add p2 flag to analytics with site

### LINKED ISSUE #13433: Analytics: Add a flag to identify P2s
We would like to be able to easily identify [new P2](https://wordpress.com/p2/) sites in tracks events. 
Let's update our tracker to include a `site_type` property anywhere we note a site's blog_id. (Probably a change to AnalyticsUtils.java)  If the site is a P2 assign the value "p2", otherwise assign "blog".  In the future this can be extended to other site types, e.g. "store".

Details:
Property name: site_type
Property values: p2, blog

Related https://github.com/wordpress-mobile/WordPress-iOS/issues/15344

khaykov:
Closes #13433 

iOS PR is [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/15362)

This PR adds a `site_type` property to all the events where we utilize SiteModel to get information about the site (like we do it with `is_jetpack` flag).

This PR does not add a `site_type` property to reader request, since PR with that flag is still in review. I will add it alongside other events.

To test:
- We don't really test Analytic Utils yet, but I added a test for the `AnalyticsSiteType` enum just in case.
- Confirm that the key and parameter names match the ones in the issue.
- Confirm that property added to correct methods.
- Setup a couple of breakpoints and confirm that the property is correctly attached to the tracking event.

cc @aerych 

PR submission checklist:


develric (approved):
Hey @khaykov 👋 . Code is clean and works well 👍 . Checked that site_type property is added and is consistent. 

Agree about the ReaderPost related methods (where we have ReaderPost we can leverage with the new is_wp_for_teams flag); not fully sure about how we better manage the `trackWithSiteId` method where we only have the blogId (not a ReaderPost) like when using it from `StatsDetailActivity->start`. 

Maybe you have already thought about that, otherwise let's keep it in mind for next PRs 🙇 . Let's move this ahead :shipit:

PR #13448, opened 2020-11-23
Fix site icon upload never starts 

### LINKED ISSUE #13442: Site icon upload never starts
### Expected behavior
- Change a site icon
- Select an image from your device
- Upload of the image starts
- The site icon gets updated

### Actual behavior
- Change a site icon
- Select an image from your device
- Upload of the image doesn't start
- The site icon is replaced with the progress bar

ashiagr:
Fixes #13442

#### To test

- Change a site icon
- Select an image from your device
- Upload of the image starts
- The site icon gets updated

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadService.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadService.java
@@ -209,10 +209,11 @@ public class UploadService extends Service {
             for (MediaModel media : mediaList) {
                 MediaModel localMedia = mMediaStore.getMediaWithLocalId(media.getId());
                 boolean notUploadedYet = localMedia != null
-                                         && MediaUploadState.fromString(localMedia.getUploadState())
-                                            != MediaUploadState.UPLOADED;
+                                         && (localMedia.getUploadState() == null
+                                             || MediaUploadState.fromString(localMedia.getUploadState())
+                                                != MediaUploadState.UPLOADED);
                 if (notUploadedYet) {
-                    toBeUploadedMediaList.add(localMedia);
+                    toBeUploadedMediaList.add(media);
                 }
             }
 


planarvoid (approved):
Thanks for the changes 👍 , I've tested the site icon change and I confirm it works now :shipit:

WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadService.java
```diff
@@ -209,10 +209,11 @@ private void unpackMediaIntent(@NonNull Intent intent) {
...
                 MediaModel localMedia = mMediaStore.getMediaWithLocalId(media.getId());
                 boolean notUploadedYet = localMedia != null
-                                         && MediaUploadState.fromString(localMedia.getUploadState())
-                                            != MediaUploadState.UPLOADED;
+                                         && (localMedia.getUploadState() == null
+                                             || MediaUploadState.fromString(localMedia.getUploadState())
+                                                != MediaUploadState.UPLOADED);
                 if (notUploadedYet) {
-                    toBeUploadedMediaList.add(localMedia);
+                    toBeUploadedMediaList.add(media);
```
malinajirka, line 216: @ashiagr Is this change intentional? Why do we want to pass a possibly out-dated item instead of an item which was just loaded from the db? Are there some fields which are not being saved into the db and we need them?

ashiagr: Yes this was intentional. For the site icon, attributes in the DB were set as null and upload didn't succeed - it failed with the message:
`MediaUploadHandler > Unexpected state, site is null. Skipping this request.`

null_state

Since we already skipped media in uploaded state, it felt fine to add original media for the other states.

PR #13450, opened 2020-11-23
Issue/13268 remote site

malinajirka:
Partially fixes #13268 

Merge instructions
1. Make sure https://github.com/wordpress-mobile/WordPress-Android/pull/13440 is merged
3. Merge this PR 

Small PR in which the "remote site id" is passed to the ActivityLogTypeFilterViewModel. Since the upcoming PR is related just to the UI itself, I thought it'd be better to create a separate PR.

To test:
There is nothing to test yet. If you really want to test it, you can put a breakpoint [here](https://github.com/wordpress-mobile/WordPress-Android/blob/696d9cf12abd947ea9c2794707dafead2ccd1067/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt#L39) and check the the id is not null. However, we'll test this before we remove the feature flag so it's not necessary.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -63,8 +64,8 @@ class ActivityLogViewModel @Inject constructor(
     val showRewindDialog: LiveData<ActivityLogListItem>
         get() = _showRewindDialog
 
-    private val _showActivityTypeFilterDialog = SingleLiveEvent<Unit>()
-    val showActivityTypeFilterDialog: LiveData<Unit>
+    private val _showActivityTypeFilterDialog = SingleLiveEvent<RemoteId>()
```
ashiagr, line 70: IIRC we don't use `SingleLiveEvent` in new classes. Can we replace it with [Event](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/viewmodel/Event.kt) class?

ashiagr (approved):
LGTM 👍, just left one minor comment.

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
malinajirka: This is not a new class and I wanted to stay consistent with the other fields in this class. I can replace all of the SingleLiveEvent usages in here, but we'll probably need to retest them all. I'm not sure it's worth it at the moment, wdyt?

ashiagr: Ah sure, thanks for clarification.

ashiagr:
LGTM 👍.
Please feel free to merge after resolving the conflict.

PR #13451, opened 2020-11-23
Selected site repository

planarvoid:
In order to implement the new my site fragment I need to pull out the selected site and the selected site settings logic into a separate class. The main reason behind this change is that right now we're relying on the callback from the top activity and it makes the code entangled. Hopefully with this change we'd be able to make the selected site logic separate from the rest of the code and hopefully move as much logic from the main activity as possible.

The functionality should be the same as it was before. The only piece of logic that might be a little bit fragile is the logic around the site icon progress bar. 

To test:
- Set site icon with an icon from your device
- Notice the progress bar
- The icon gets set
- Set site icon from WP media library
- The new icon is visible
- Change site title
- Change settings
- Switch sites
- Everything behaves as you'd expect

PR submission checklist:


renanferrari (approved):
Thanks for the changes! I've tested the app and everything seems to be working ok as far as I can see. Code is looking pretty good as well.

LGTM :shipit:

PR #13453, opened 2020-11-23
Create toolbar in the improved My Site fragment

### LINKED ISSUE #13344: Create a toolbar in improved My Site
Copy the current Toolbar with the Me button and the scroll behaviour to the new fragment

renanferrari:
Fixes #13344

This PR moves the existing collapsing toolbar and Me button logic to the new `ImprovedMySiteFragment`. It also introduces a `UiModel` class to `MySiteViewModel` in order to handle avatar changes.

To test:

1. Enable the Improved My Site flag and restart the app.
1. On the My Site fragment, notice the new toolbar.
1. Verify that both the title and the Me button are resized when the toolbar is collapses and expands.
1. Press the Me button and verify that the Me screen appears.
1. Change the account's avatar, go back to the My Site fragment and verify that the avatar was updated.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
```diff
@@ -60,6 +67,33 @@ class ImprovedMySiteFragment : Fragment(),
...
+                val modifierPx = (minSize.toFloat() - maxSize.toFloat()) * (percentage.toFloat() / 100) * -1
+                val modifierPercentage = modifierPx / minSize
+                val newScale = 1 + modifierPercentage
+
+                avatar.scaleX = newScale
+                avatar.scaleY = newScale
+            }
+        })
+
+        toolbar_main?.let { toolbar ->
```
planarvoid: do these synthetic accessors need to be nullable?

WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteViewModel.kt
```diff
@@ -58,6 +63,15 @@ class MySiteViewModel @Inject constructor(
         } else {
             listOf()
         }
+        UiModel(currentAvatarUrl.orEmpty(), items)
+    }
+
+    init {
+        updateAccountAvatarUrl()
```
planarvoid: I'm wondering if it could ever happen that the user doesn't have an avatar (so the URL is empty) and then selects a new avatar but this field never gets updated. In the previous version the avatar URL is updated in `onCreateOptionsMenu`. Maybe it would make sense to update it in `onResume` if it has changed?

planarvoid (approved):
It's looking good and works well 👍  Do you think you can add a test or two for the view model avatar loading? Thanks!

WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
renanferrari: Good catch! Probably not.

WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteViewModel.kt
renanferrari: Good question! I actually tested this scenario and thought it wasn't necessary as the avatar was being correctly updated after coming back from the Me screen, but your question made me realize that this was just a side effect of the call we are already making to `MySiteViewModel::updateSite` from the fragment's `onResume`. 

I've now moved the method to refresh the avatar url out of the ViewModel's `init` block so it can be called from the fragment's `onResume`, to make the intended behavior explicit.

renanferrari:
Thanks for the review, @planarvoid! I've addressed your questions above and also added some tests like you suggested.

WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteViewModel.kt
planarvoid: thanks, I've tested it as well and it works because the logic is strange. The avatar URL is always the same but it shows a different image in the Gravatar loader. I can't say I really understand why but I assume that the backend just replaces the image on the same URL. However, I think there could be a case where the avatar URL is empty and if that happens, I think we'd never recover. Thanks for the change!

planarvoid (approved):
Thanks for the changes! It's looking great 👍

PR #13466, opened 2020-11-24
Adds WPUtils as composite build

loremattei:
This PR takes over the excellent work @oguzkocer started in https://github.com/wordpress-mobile/WordPress-Android/pull/12796. 
The PR was opened some time ago and there are now conflicts with the Utils library. I tried to fix those, but due to the involved operations, the graph of the branch ended up to be very confusing, so I preferred to branch from https://github.com/wordpress-mobile/WordPress-Android/pull/12796/commits/f19af8bdba0f16a980c8179ab338a18aec377e01 which contains all the diffs to enable the composite build and re-do the deletion of the current `Utils` library in this branch. 

The changes in this PR can be split in three parts: 

1. changes required to enable the composite build. An easy way to review them is to look at this branch: https://github.com/wordpress-mobile/WordPress-Android/pull/13465 which is cut off https://github.com/wordpress-mobile/WordPress-Android/pull/12796/commits/f19af8bdba0f16a980c8179ab338a18aec377e01 (I'm keeping it open just to this review easier. I'll delete it without merging it). 

2. changes to remove the `utils` library subtree. They are done in the commits from https://github.com/wordpress-mobile/WordPress-Android/pull/13466/commits/b4490675f24db19c073425936ba485e1c194766f to https://github.com/wordpress-mobile/WordPress-Android/pull/13466/commits/45a13efb5bbf1a7a48e4dc50a18611fdd3fc50be.

3. changes to make lint work without errors after that. 

**To test:**
For testing the first two points above I'll just copy Oguz's instructions: 

This PR is a follow up to #13001 and makes https://github.com/wordpress-mobile/WordPress-Utils-Android a binary dependency and adds the option to make it a composite build for local development. It also sets up the repository to remove the libs/utils subtree since it's no longer used.

Please make sure to check out #13001 for details on what composite build is and how it works. Since the merge of this PR will be a bit involved, I am not duplicating that information here to keep things brief. Also note that, the PR currently includes changes from #13001 since it's not yet merged, but once it is merged, the changes should stop showing up. (sometimes it doesn't in which case, please change the base branch to another branch and change it back to develop which hopefully resolves the issue)

**To test:**

- Copy local-builds.gradle-example as local-builds.gradle and update localWPUtilsPath to your liking
- Use Navigate Class shortcut to open UrlUtils which is in WPUtils and add a new static method such as a summation
- Use Navigate Class shortcut to open ReaderPostLogic (any other class would be fine) which is in WPAndroid and use that function
- Run the project and verify that it works 🎉
- Comment out localWPUtilsPath in local-builds.gradle file and verify that the build fails for missing the new method from UrlUtils
- Discard the changes for ReaderPostLogic, build the app and verify that it works

**Lint errors**

I found a few lint errors after removing the subtree. I dug into those, but I couldn't find a lot of information about them. They are mostly about missing classes and wrong interfaces, for which I expected crashes if they were actual issues, but I've tested the app and I haven't experienced any crash, so I _think_ that they are mostly related to the linter not handling well composite builds, **but** I may be very wrong here, so, please, lay out any doubt you may have. 

1. The first error is here: https://github.com/wordpress-mobile/WordPress-Android/pull/13466/commits/e911e43f196bc2d54d8511e70de88e6a1499dead
This happens when running `gradlew lintVanillaRelease` in binary import mode. I downloaded `aar` file, decompiled it and verified that the parameter expects a `ColorInt`, so it looked good to me. So, I silenced it as a false positive.

2. The second group of errors is here: https://github.com/wordpress-mobile/WordPress-Android/pull/13466/commits/af3cbeff9b0d7d34ae313c77111350e88aac08ba. They happen when running composite builds with dependency substitution in local mode. I expected build errors if classes where really missing, but the app builds fine and testing it it seems to work ok. So, I silenced them as false positives. 

3. The third error is here: https://github.com/wordpress-mobile/WordPress-Android/pull/13466/commits/b70fed75cc64a89a05a3910b2cb243b8250f27ca. This is the only one I'm sure about why it happens as it's because we import the resource in our main `strings.xml` to have it translated. The fix for this one should be ok. 

4. The last group of errors is here: https://github.com/wordpress-mobile/WordPress-Android/pull/13466/commits/57449da95c0d685d97ea4c5191ef80e3e5b49352. Again, I think that it's because the lint doesn't compute the dependencies right with composite builds, but I'm not totally sure about why. I silenced them as false positives as the app seems to work well. 

PR submission checklist:


aforcier:
Nice work @loremattei (and of course Oguz 🙂)! I'll be glad for this to be merged and perhaps sometime soon never have to do `subtree push` again 😀 😅 😭 

I haven't quite finished reviewing this, but I started looking at the lint issues and have some stuff to report in the meantime.

It's clear the issues reported aren't actually breaking the build, so silencing them in `lint-baseline` seems sensible. However the issue fixed in af3cbef seems to affect all uses of utils classes from xml files, and the baseline additions are for existting cases, which means the lint error will strike whenever a new usage is added. And, if I understand what's happening correctly, CI won't fail, so the new error might not be noticed until the next time someone runs lint locally while using a local copy of `utils`. Not horrible but potentially confusing.

However I _think_ we might be able to avoid it. I was looking at the gradle build config and wondered if [the binary path not pointing to the actual utils module](https://github.com/wordpress-mobile/WordPress-Android/pull/13466/files#diff-7f825392aa37acd1cee0c2e7b9bb7366ad6eac64f3e6cdd816e156bcb69d30deR88) (`WordPressUtils`) might be a factor. I would have expected it to fail but the build seems to work with or without it 😅

So I checked out the last commit before your lint silencing changes (45a13ef), pointed to the module and rebuilt, and  after that`./gradlew lintVanillaRelease` is passing for me locally with a local utils checkout. And I pushed the change and opened a test PR, and CI seems happy with it: https://github.com/wordpress-mobile/WordPress-Android/pull/13484 (The actual change is here: https://github.com/wordpress-mobile/WordPress-Android/pull/13484/commits/4658ad8c22a947b22c2624982b4f46c224db0158).

Could you try that out locally too and let me know if it's resolved the issue you as well (or if I just missed something 🙃)?

loremattei:
Thank you very much for looking into this @aforcier!
It definitely solves the lint issues when importing the library as a binary dependency, both on CI and locally.

I still get lint errors locally when using the build substitution tho :-( 
I wonder if this is because I miss something when enabling the substitution: what are you putting in your `local-builds.gradle` to use the source code locally?

aforcier:
Hmm interesting 🤔 

My file just looks like this:

```
ext {
    //localFluxCPath = "../WordPress-FluxC-Android"
    localWPUtilsPath = "../WordPress-Utils-Android"
}

```

I also made sure my local copy of utils was pointing to latest `develop`.

loremattei:
This looks like a weird issue 🤔 I use the same configuration file, but I experienced the lint errors until I nuked my ~/.gradle cache and checked out a new copy of the WPAndroid (unfortunately, I did these two things together, so I'm not sure which one actually solved the problem). 

Anyway, this looks like a caching issue, so I've reverted the commits to silence the lint error and cherry-picked your change. 
Thank you for looking into this!

aforcier (approved):
This is looking good @loremattei! :shipit: from me.

I do have a few operational questions/suggestions:

1. Should we include instructions somewhere on how to work with the utils library going forward?
2. We can probably remove the utils Peril hook: https://github.com/wordpress-mobile/WordPress-Android/pull/13466#issuecomment-732887507
3. I was going to comment about having removed the utils unit test run job from CI, but I see we're also running that in the utils library, and now the subtree will be gone all utils changes will require a PR in that repo, so we're good there. 👍 

Finally, I recommend having one other person try out this PR in case there are more hiccups with the linter - if others need to clean out their gradle cache to get local builds working too we should probably put out a PSA. I know @malinajirka has probably seen a good few linter oddities, perhaps he can give the build a try 😀

loremattei:
Thank you @aforcier! 

> Should we include instructions somewhere on how to work with the utils library going forward?

Yes! I'm drafting a post about this. We can also add something to the library's README.

> We can probably remove the utils Peril hook:

Right! I'll take care of this!

malinajirka:
Thank you!!! Really great job on both sides:).

I tested running the app and lint and I encountered some issues. Here are steps I took:
1. Pulled the branch
2. Activated local build for WPUtils + synced with gradle files
3. Run the app - it worked without issues
4. I tried to run `./gradlew lintVanillaRelease` and it failed with the following error
```
 /Users/jirimalina/Code/Automattic/WordPress-Android-submodule/WordPress/src/main/java/org/wordpress/android/util/WPSwipeToRefreshHelper.java:33: Error: Expected a color resource id (R.color.) but received an RGB integer [ResourceType]
          return new SwipeToRefreshHelper(swipeRefreshLayout, listener, backgroundColor, primaryProgressColor,
```
5. Run clean + run lint -> same error
6. Run `./gradlew cleanBuildCache` -> run lint -> same error
7. Killed gradle daemons and removed the whole ./gradlew/caches folder -> It took me a while to fix the build but when I fixed it I'm still seeing the same lint error.
8. Killed gradle deamons and remove the whole ./gradlew folder -> same error
8. My AS is quite confused now and I can't manage to fix it. It can't find a bunch of dependencies and it keeps underlining most of the code. However, the build works.
9. I tried to replace `elevationOverlayProvider.compositeOverlayWithThemeSurfaceColorIfNeeded(appbarElevation);` in `WPSwipeToRefreshHelper` with `int backgroundColor = context.getResources().getColor(R.color.accent_100);` and the lint works. Are you sure the lint worked for you even when you didn't suppress this error (with the local build enabled)?

malinajirka:
I got rid of the broken Android Studio (step 9) by opening the project in a newer version. Tbh I don't understand why, but when I run the lint now everything works as expected. 

To summarize: There is a pretty high chance others will have troubles running lint locally as well. I can't tell how to fix the issue since I basically fixed it by accident. Either way, I would just mention it on a P2 and wouldn't worry about it too much. I don't think many people run lint locally anyway.

loremattei:
Thanks for your tests @malinajirka! This lint issue is so weird! 
I'm still pretty torn about silencing it or not. It really looks like a lint hiccup rather than an actual issue with the code. 
But I think the best for now is to move ahead with your suggestion and add a note about it when announcing the change. 
It it turns out to be too annoying or unsolvable for people, we can silence it with a follow up PR later.

jd-alexander:
Thanks for discovering this @malinajirka 

@loremattei this seems to be happing on `develop` as well. I just got a random failure with this error message on the [Gutenberg release branch](https://github.com/wordpress-mobile/WordPress-Android/pull/13560).

PR #13469, opened 2020-11-24
Issue/13268 bind ui

malinajirka:
Partially fixes #13268 

This PR maps data provided by the VM to the UI. 

Merge instructions:
1. Make sure https://github.com/wordpress-mobile/WordPress-Android/pull/13450 is merged
3. Merge this PR

Notes:
- The final designs are not ready yet, so the UI is not polished. 
- On item click logic is implemented in the next PR

To test:
- Make sure the ActivityLogFiltersFeatureConfig is enabled
1. My site -> Activity Log
2. Click on "Activity Type" chip
3. Notice a loading screen is shown
4. Notice the loading screen gets replaced with the content screen after 1s

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
```diff
@@ -65,6 +62,39 @@ class ActivityLogTypeFilterFragment : DialogFragment() {
         initAdapter()
     }
 
+    private fun initViewModel() {
+        viewModel = ViewModelProviders.of(this, viewModelFactory)
+                .get(ActivityLogTypeFilterViewModel::class.java)
+
+        viewModel.uiState.observe(viewLifecycleOwner, Observer { uiState ->
```
ashiagr, line 69: IDE highlights it (`Observer`) as redundant SAM-constructor, however I think we're not removing them yet in the code for compatibility with older AS versions, right?

ashiagr (approved):
I've tested both success and error (by changing provider data) flow and everything worked as expected 👍.
Left a nitpick + one question.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
```diff
@@ -48,7 +45,7 @@ class ActivityLogTypeFilterFragment : DialogFragment() {
         super.onViewCreated(view, savedInstanceState)
         initToolbar(view)
         initRecyclerView()
-        viewModel.start(remoteSiteId = RemoteId(requireNotNull(arguments).getLong(WordPress.REMOTE_SITE_ID)))
+        initViewModel()
     }
 
     private fun initToolbar(view: View) {
```
ashiagr: This is not related to changes in this PR and should've been caught earlier, can we remove unused view param since we've replaced findByViewId with Kotlin synthetics.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
malinajirka: Yes, I had the same thought that it'd be better to keep it for now.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
malinajirka: Fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/13469/commits/15ef9fdbd4a974ff07f0c4e755d89c63115b4179

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
ParaskP7: Hmmm... Is this due to compatibility reason with AS version or Kotlin versions? I am thinking the later, right? And if that is the case, I suggest having every develop be forced to update the AS Kotlin plugin to at least the Gradle's Kotlin version, currently `1.4.10`, as specified by our build files. This will make sure everyone is on the same boat. Wdyt?

malinajirka: Good point! I thought it was the AS lint, but it's the kotlin plugin. In that case I agree with you.

PR #13473, opened 2020-11-24
Follow unfollow post comments

### LINKED ISSUE #12602: Adding to the Reader P2 Follow/Unfollow comments capability
### Background

We already massively use P2 blogs internally in a8c to enforce teams project collaboration and we are going to launch the P2 Product [p7DVsv-8N3-p2] pretty soon. 

In a P2 web page we allow a user to Follow/Unfollow a stream of comments to a post by the `Follow` button available at the bottom of a post itself.

<p align="center">
  
</p>

The effect of Following a post comments stream is that the user gets notified by emails of new comments to that post.

### Task

With the above in mind we would like to add the capability in the app to Follow/Unfollow a P2 post comments stream from the Reader (will add some mocks related to this soon)

### Relevant endpoints

- [Subscribe to a post comments stream](https://developer.wordpress.com/docs/api/1.1/post/sites/%24site/posts/%24post/subscribers/new/)  
- [Unsubscribe from a post comments stream](https://developer.wordpress.com/docs/api/1.1/post/sites/%24site/posts/%24post/subscribers/mine/delete/)
- [Get subscription state](https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/posts/%24post/subscribers/mine/)

This will require some work on the [FluxC library](https://github.com/wordpress-mobile/WordPress-FluxC-Android) as well

NOTE: the new P2 Product blogs are marked as `"is_wpforteams_site": true,` in the site options response

<p align="center">
  
</p>

as of current understanding this does not apply to older a8c internal P2s. In case we cannot find a flag that marks an old P2 site as being a P2 we could eventually evaluate to assume that the blogs that are in the reader Automattic Tab are P2s (note those P2s are only available to Automatticians).

<p align="center">
  
</p>

develric:
Fixes #12602 

This PR adds support to follow/unfollow comments conversation in the app reader so you are notified by email when a new comment is placed in a specific post. Originally this was meant only for P2 sites but after some researches, given the API endpoints used are supported for all .com/JP/P2 sites, we decided to support this for all that kind of sites. Only exception, since that API cannot be accessed in this case, are not Jetpack connected pure self-hosted sites. (see internal `pbArwn-1n5-p2` for more context on this).

Follow button is available wherever the `ReaderCommentListActivity.java` is launched (reader card, reader post details, from comments notifications...) and can be used both if you follow the site or not.

**NOTE:** I'm using the already available `ReaderCommentListViewModel.kt`but in the `ReaderCommentListActivity.java` I changed the method to retrieve the VM from `create` to `get` since I needed to keep the VM state in between configuration changes. AFAIU this works correctly also for the scroll position it was used before, but I thought better to mention 😄 

### References
- Original design `pbArwn-Q1-p2`
- Some discussion on user audience `pbArwn-1n5-p2`

| Description | Image |
|---|---|
| Access comments from reader tapping on comments icon in the card (also possible to access from Reader Post Deatils) |  <p align=center ><image src="https://user-images.githubusercontent.com/47797566/100155201-1f088400-2ea7-11eb-8d8f-68a1bf58f2ed.png" width=300/></p> |
| Subscribed to the comments | <p align=center ><image src="https://user-images.githubusercontent.com/47797566/100152342-d6e76280-2ea2-11eb-8b0a-25fe4e0ed4c5.png" width=300/></p> |
| Unsuscribed from the comments | <p align=center ><image src="https://user-images.githubusercontent.com/47797566/100152391-ea92c900-2ea2-11eb-9464-234a458508dd.png" width=300/></p> |
| Shimmer effect on Follow button while loading status  | <p align=center ><image src="https://user-images.githubusercontent.com/47797566/100152228-a8698780-2ea2-11eb-9bf5-22cd79fc312b.png" width=300/></p> |
| Disabled state in error querying endpoints (like no network) - using `colorOnSurface` with `disabled_alpha [0.12]`  | <p align=center ><image src="https://user-images.githubusercontent.com/47797566/100153418-5fb2ce00-2ea4-11eb-826c-b1f143b2f5e6.png" width=300/></p> |
| Disabled state also accessing comments for posts in pure self-hosted sites  | <p align=center ><image src="https://user-images.githubusercontent.com/47797566/100395151-1c8d6200-3040-11eb-8d27-1a2422467975.png" width=300/></p> |
| Button not available if user is not logged in .com  | <p align=center ><image src="https://user-images.githubusercontent.com/47797566/100152859-a2c07180-2ea3-11eb-8080-727c0aa0c50b.png" width=300/></p> |

### Something I would need a design review for
- I should have used the same as iOS copy for button and snackbars; hope that's fine.
- Follow button (see images above for an example):
    - Since the call to recover subscription status can fail, we used a Shimmer while loading the status from API and the button is disabled if we cannot recover the status. Designer eyes needed here 😄 
    - Tried to use same styles as Follow site button but with 32dp height (as in iOS AFAIU) and adding a Disabled state when errors querying the subscribe endpoints happens (like for no network error; looking at the snackbar copy, I think we use that sentence already in other places but not fully sure the copy is effective maybe 🤔 )

For a live tour you should be able to use the apk built in this PR activating the feature in `My Site -> Me -> App Settings -> Test feature configuration` and activating the `FollowUnfollowCommentsFeatureConfig` (restart the app after activating)

<p align=center ><image src="https://user-images.githubusercontent.com/47797566/100155873-4c096680-2ea8-11eb-8b0b-d10bc6bbf29f.png" width=300/></p>

### Some open items that will be addressed in subsequent PRs
I left some topics for other PRs since this one was not so small already :)
- Adding tracking events
- Evaluate if to use `suspendCancellableCoroutine`and `withTimeout`instead of the pure `suspendCouroutine`
- Adding Unit Testing

### TEST CASE 1: Comments opened from reader card or reader post details (This and TEST CASE 2 should be checked with JP/.COM/ATOMIC sites (better if you own them so you can freely comment on posts just for testing 😄 . Refer to the images in the References above for context)

1. From the reader open a tab filter (FOLLOWING/DISCOVER etc...)
2. Tap on the comments icon from a reader post card
3. In the `ReaderCommentListActivity` check that the Follow Button shimmer effect is started and it then show the Follow button in the expected state (Following or not Following yet)
4. Set the conversation to Following conversation (a shimmer should start before reaching the final state)
5. Check the snackbar acknowledge you subscribed to the comments
6. From the web post a comment on the post and check your user email receives the notification about the new comment
7. Set the conversation to Follow conversation CTA 
8. Check the snackbar acknowledge you unsubscribed from the comments
9. From the web post a comment on the post and check your user email does not receive the notification about the new comment
10. Smoke test the `ReaderCommentListActivity` opening it from a system notification or app notification, reply to post comments or whatever else you can think of 😄 
11. Now open a post to access to the reader post details screen
12. Tap on comments icon in the bottom bar and repeat all steps from 3 to 10 above

### TEST CASE 2
1. From the reader open a tab filter (FOLLOWING/DISCOVER etc...)
2. Set your device to Airplane Mode
3. In the `ReaderCommentListActivity` check that the Follow Button is disabled and you get a snackbar saying to check your network

<p align=center ><image src="https://user-images.githubusercontent.com/47797566/100238473-89193b80-2f30-11eb-901e-6296d9d958ad.png" width=300/></p>

### TEST CASE 3: Comments opened from reader card or reader post details for pure self-hosted sites
Follow same steps above but notice the Follow button is dimmed after a shimmer effect

### TEST CASE 4: Comments opened when user is not logged in .com
- logout from .com and add a pure self-hosted site by address
- go to reader Discover tab and tap on comments from a post
- notice the follow button is not visible

### TEST CASE 5
Smoke test the comments scenario above disabling the feature flag so that we ensure there is no wrong behaviour when the FF is disabled.

PR submission checklist:


mattmiklic (approved):
> I should have used the same as iOS copy for button and snackbars; hope that's fine.

Yes, that makes sense to me.

> Since the call to recover subscription status can fail, we used a Shimmer while loading the status from API and the button is disabled if we cannot recover the status. Designer eyes needed here 😄

The shimmer effect makes sense as a loading state, and looks fine to me.

> Tried to use same styles as Follow site button but with 32dp height (as in iOS AFAIU) and adding a Disabled state when errors querying the subscribe endpoints happens

This also makes sense. It looks good and the disabled state is clear to me. I think this is good to go.

develric:
Thanks for the review @mattmiklic 🙇 . 

I updated the PR description with test instructions and should be ready for code review.

WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderCommentsFollowUseCase.kt
```diff
@@ -0,0 +1,124 @@
...
+
+        data class FollowStateChanged(
+            val blogId: Long,
+            val postId: Long,
+            val isFollowing: Boolean,
+            val isInit: Boolean = false,
+            val userMessage: UiString? = null
+        ) : FollowCommentsState()
+
+        data class NoNetwork(
```
planarvoid: I think I'd merge this and the `Failure` class. They are handled very similarly. It's just a type of a Failure

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderCommentListViewModel.kt
```diff
@@ -31,4 +65,77 @@ class ReaderCommentListViewModel
...
+        followStatusGetJob?.cancel()
+        followStatusGetJob = launch(bgDispatcher) {
+            followCommentsHandler.handleFollowCommentsStatusRequest(blogId, postId, isInit)
+        }
+    }
+
+    private fun buildFollowCommentsUiState(followCommentsState: FollowCommentsState): FollowCommentsUiState {
+        return if ((followCommentsState is NoNetwork && !followCommentsState.isGetStatus) ||
+                (followCommentsState is Failure && !followCommentsState.isGetStatus)) {
+            Nop
```
planarvoid: I think I'd personally move this check to the place where we set the value of the live data. For example we could filter it out here: 
```
_updateFollowStatus.addSource(followCommentsHandler.followStatusUpdate) { event ->
            _updateFollowStatus.value = event
        }
```

It feels a little out of place to set the field and not use the value in certain cases.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderFollowCommentsHandler.kt
```diff
@@ -0,0 +1,71 @@
...
+    @Named(BG_THREAD) private val bgDispatcher: CoroutineDispatcher
+) {
+    private val _snackbarEvents = MediatorLiveData<Event<SnackbarMessageHolder>>()
+    val snackbarEvents: LiveData<Event<SnackbarMessageHolder>> = _snackbarEvents
+
+    private val _followStatusUpdate = MediatorLiveData<FollowCommentsState>()
+    val followStatusUpdate: LiveData<FollowCommentsState> = _followStatusUpdate
+
+    suspend fun handleFollowCommentsClicked(blogId: Long, postId: Long, askSubscribe: Boolean) {
+        withContext(bgDispatcher) {
```
planarvoid: instead of this you should use `readerCommentsFollowUseCase.setMySubscriptionToPost(blogId, postId, askSubscribe).flowOn(bgDispatcher)`

planarvoid (commented):
I've looked into the flow related code and it's looking good 👍 , I've left a few comments but I haven't tested the app :-)

WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderCommentsFollowUseCase.kt
develric: Good point 👍 . Merged it in 015dd2a8c791d38eb3427bb80fbcb04d8ef3d329 . Also used `UpdateFollowCommentsUiState` as a simple data class removing the Nop state.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderCommentListViewModel.kt
develric: Agree. Removed the Nop completely here 015dd2a8c791d38eb3427bb80fbcb04d8ef3d329

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderFollowCommentsHandler.kt
develric: Thanks for sharing 🙇 . Done it in 612cc797d5ee09c2a1091ef9926fa0f47a35e476

develric:
Was hoping you could have time for an high level check especially on the flow side, thanks @planarvoid 😄 . I implemented your suggestions 🙇 . @khaykov this should be ready for another check round from your side now 🙇

WordPress/src/main/java/org/wordpress/android/ui/reader/FollowCommentsUiState.kt
```diff
@@ -0,0 +1,16 @@
+package org.wordpress.android.ui.reader
+
+data class UpdateFollowCommentsUiState(
```
khaykov: This is just a personal preference, but `Update` in the data class name hints that it does some sort of action, while it just holds a UI state. What do you think about calling this class just `FollowCommentsUiState` ?

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderCommentListViewModel.kt
```diff
@@ -31,4 +61,71 @@ class ReaderCommentListViewModel
...
+    private fun buildFollowCommentsUiState(followCommentsState: FollowCommentsState): UpdateFollowCommentsUiState {
+        return UpdateFollowCommentsUiState(
+                    type = when (followCommentsState) {
+                        Loading -> LOADING
+                        is FollowStateChanged -> VISIBLE_WITH_STATE
+                        is Failure -> DISABLED
+                        FollowCommentsNotAllowed -> GONE
+                    },
+                    showFollowButton = followCommentsState !is FollowCommentsNotAllowed,
+                    isFollowing = if (followCommentsState is FollowStateChanged)
```
khaykov: Tiny nitpick, but can we add curly brackets to if/else statements? Single line statements, like `if(!followUnfollowCommentsFeatureConfig.isEnabled()) return`  are ok, but multiline ones are easy to accidentally mess up, and it's very hard to debug later on.

WordPress/src/main/res/values/strings.xml
```diff
@@ -1345,6 +1345,11 @@
     <string name="reader_discover_empty_button_text">Get Started</string>
     <string name="reader_discover_no_posts_button_text">Follow topics</string>
     <string name="reader_author_with_blog_name" translatable="false">%1$s ▸ %2$s</string>
+    <string name="reader_follow_comments_null_response">Null response received</string>
```
khaykov: Since it's a user-facing string, I would use some other word than null (maybe `No response received`?)

WordPress/src/main/res/values/strings.xml
```diff
@@ -1345,6 +1345,11 @@
...
+    <string name="reader_follow_comments_bad_format_response">Bad format response received</string>
```
khaykov: And similarly here, maybe "Invalid response received" will sound a bit more friendly (if it's even possible with error messages :D)

WordPress/src/main/res/color/secondary_gray_20_disabled_selector.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:alpha="@dimen/disabled_alpha" android:color="?attr/colorOnSurface" android:state_enabled="false" android:state_selected="false" />
+    <item android:alpha="@dimen/disabled_alpha" android:color="?attr/colorOnSurface" android:state_enabled="false" android:state_selected="true" />
+    <item android:color="@color/gray_20" android:state_enabled="true" android:state_selected="true" />
```
khaykov: What do you think about using the material color attribute here, like wpColorOnSurfaceMedium? Looks very similar :)

khaykov (commented):
Thanks for a great feature, @develric ! It works perfectly, especially after your recent changes. I'm not 100% sure about the parts of code that use flow, but they look ok to me :) I guess I have to implement something similar myself to be able to tell exactly what is going on :)

Apart from this I only have a couple of really small nitpicks :)

WordPress/src/main/java/org/wordpress/android/ui/reader/FollowCommentsUiState.kt
develric: Makes sense 👍 . Refactored in 9205aa7e8c97d6f2905a6211abb3255f9be5dd27

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderCommentListViewModel.kt
develric: Changed here cca73232f8f233a2a71f1c2f7a35805a3d8d97a4

WordPress/src/main/res/values/strings.xml
develric: Good point 😄 . Changed in 498ddfad2d52c1f1bc7d3acf8a1974d26afe3e47

WordPress/src/main/res/values/strings.xml
develric: Changed in 498ddfad2d52c1f1bc7d3acf8a1974d26afe3e47

WordPress/src/main/res/color/secondary_gray_20_disabled_selector.xml
develric: ahah 😄 . It was the only state not using material attr 🤦 ! Good catch 👍 . Changed in fcf46ac15fe24b4c8de9facb92a7019aace15900

develric:
Thanks for the review @khaykov 🙇 ! Yep Flow can be really clean and powerful 😄 . I implemented your suggestions, ready for another pass 🙇 !

khaykov (approved):
Thanks for the changes! Looks good 🚀

PR #13475, opened 2020-11-25
Issue/13446 clean up comment related analytics

### LINKED ISSUE #13446: Clean up comment related analytics
As part of P2-related improvements, we need to clear up existing analytics events, align both platforms, and fix the double-counting that is currently happening where notification related comment events are bundled with non-notification comment actions.

khaykov:
Fixes #13446

This PR updates comment related analytics.

Since we are probably going to do a global rework of comment logic (which will also include analytics), I decided to not overthink it and integrated most of the tracking with existing AnalyticsUtils code.

I kept the existing events for now and fixed the issue with site comment events being bundled with notification ones.

What changed:
- We differentiating the source of comment action (eg. Notifications, Reader, Site Comments)
- Simplified names.
- Adding site type when tracking with Reader's Post (eg. p2 or blog).
- Added a couple of extra events (eg, Reply, Delete, View)

To test:
- Confirm the new events in `AnalyticsTracker`.
Place breakpoints in the following `AnalyticsUtils` methods:
- trackCommentActionWithSiteDetails
- trackCommentActionWithReaderPostDetails
- trackCommentReplyWithDetails (this will work for non-reader comment replies)
And confirm that they are called with correct parameters and at the correct timing.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderCommentListActivity.java
```diff
@@ -604,6 +606,8 @@ private void doDirectOperation() {
 
                                 AnalyticsUtils.trackWithReaderPostDetails(
                                         AnalyticsTracker.Stat.READER_ARTICLE_COMMENT_LIKED, mPost);
+                                AnalyticsUtils.trackCommentActionWithReaderPostDetails(Stat.COMMENT_LIKED,
+                                        AnalyticsCommentActionSource.READER, mPost);
```
develric, line 610: I see in this file at line 264 we have

```
AnalyticsUtils.trackWithReaderPostDetails(AnalyticsTracker.Stat.READER_ARTICLE_COMMENTS_OPENED, mPost);
```

Should we duplicate also there? wdyt?

libs/analytics/WordPressAnalytics/src/main/java/org/wordpress/android/analytics/AnalyticsTracker.java
```diff
@@ -690,6 +690,23 @@
         ENCRYPTED_LOGGING_UPLOAD_SUCCESSFUL,
         ENCRYPTED_LOGGING_UPLOAD_FAILED,
         READER_POST_REPORTED,
+        COMMENT_APPROVED,
+        COMMENT_UNAPPROVED,
+        COMMENT_SPAMMED,
+        COMMENT_UNSPAMMED,
+        COMMENT_FOLLOWED,
+        COMMENT_UNFOLLOWED,
```
develric: We correctly declare these 2 but if I'm not wrong we do not use them, do you agree? I see iOS has it, not sure what they trace but maybe something to dig a bit? wdyt?

libs/analytics/WordPressAnalytics/src/main/java/org/wordpress/android/analytics/AnalyticsTracker.java
```diff
@@ -690,6 +690,23 @@
...
+        COMMENT_LIKED,
+        COMMENT_UNLIKED,
+        COMMENT_TRASHED,
+        COMMENT_UNTRASHED,
```
develric, line 700: Not sure how untrashing a comment should work, trying it on notifications I see that for a trashed comment we have a `Restore` action where we usually have an Approve/Approved button. Using it I get a `comment_approved` event instead of a `comment_untrashed`. Maybe something similar is going on in backend logic like for SPAM/UNSPAM?

Update: in site comments, when trashing there is a snackbar that allows to UNDO the trashing, I think we don't trace that.

WordPress/src/main/java/org/wordpress/android/ui/comments/CommentDetailFragment.java
```diff
@@ -421,6 +440,8 @@ public void afterTextChanged(Editable s) {
     public void onResume() {
         super.onResume();
         ActivityId.trackLastActivity(ActivityId.COMMENT_DETAIL);
+        AnalyticsUtils.trackCommentActionWithSiteDetails(
+                Stat.COMMENT_VIEWED, mCommentSource.toAnalyticsCommentActionSource(), mSite);
```
develric: maybe we should avoid tracking this if we are coming back from another screen? wdyt?

libs/analytics/WordPressAnalytics/src/main/java/org/wordpress/android/analytics/AnalyticsTracker.java
```diff
@@ -690,6 +690,23 @@
...
+        COMMENT_REPLIED_TO,
+        COMMENT_EDITED,
+        COMMENT_VIEWED,
+        COMMENT_DELETED,
+        COMMENT_QUICK_ACTION_APPROVED,
+        COMMENT_QUICK_ACTION_LIKED,
+        COMMENT_QUICK_ACTION_REPLIED_TO,
```
develric, line 707: I think you did it on purpose, and not sure it's worth, but placing here as a reminder to discuss about it; we are missing a few events that are managed by iOS (unless they are not used anymore there, we should verify in tracks):

- followed_blog_notification_settings_comments_on
- followed_blog_notification_settings_comments_off

Also not sure it's worth here, but we should add to `stats_overview_type_tapped` a source as well (VIEWS/VISITORS/LIKES and COMMENTS). see pbArwn-1oJ-p2#comment-1969 for some more context.

develric (commented):
Hey @khaykov , good job with this 👍 . Code LGTM and I tried to test all the flows and it's looking ok. I left a few comments for your feedback before approving it. Let me know, thanks 🙇

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderCommentListActivity.java
khaykov: We decided to leave the "screen opened" events as is, so I removed the corresponding event from the proposed events list.

libs/analytics/WordPressAnalytics/src/main/java/org/wordpress/android/analytics/AnalyticsTracker.java
khaykov: Good catch! Looks like the original iOS event is called `WPAnalyticsStatNotificationsSiteFollowAction`, which should track site following, so I assume it got mixed into the events list by mistake.

libs/analytics/WordPressAnalytics/src/main/java/org/wordpress/android/analytics/AnalyticsTracker.java
khaykov: Think I figured this one out. Regarding the UNDO snackbar, we don't track that one, since the actual action happens when snackbar is dismissed, and undo is more like "I changed my mind" thing :)

WordPress/src/main/java/org/wordpress/android/ui/comments/CommentDetailFragment.java
khaykov: This was a bit tricky, but I think I made it work 👍

libs/analytics/WordPressAnalytics/src/main/java/org/wordpress/android/analytics/AnalyticsTracker.java
khaykov: After doing some reasearch, It looks like we do track the `followed_blog_notifications_settings_comments_on` and `followed_blog_notifications_settings_comments_off` events in Android already.

We will tackle adding a source to the stats event in a separate PR 👍

develric: >After doing some reasearch, It looks like we do track the followed_blog_notifications_settings_comments_on and followed_blog_notifications_settings_comments_off events in Android already.

ahah 😄 , I was missing the typo with `s` in notifications! good catch :). Agree for the rest 👍

develric (approved):
Thanks for the changes and the follow ups @khaykov 🙇 ! Agree makes sense to cover those couple of things in separate PR. Had another round here and it's working great 👍 . :shipit:

PR #13478, opened 2020-11-25
Site block icon interactions

### LINKED ISSUE #13343: Create My Site item and ViewHolder

planarvoid:
Fixes #13343 

This PR adds interactions to the new site icon. The dialog has 3 options - change icon, remove icon and dismiss. Change icon should open the media picker and correctly handle the selected icon. The remove icon should start showing the placeholder and dismiss just hides the dialog. I've had to use extract quite a lot of code for this. I've created a new view model to handle the upload success/error events. 

To test:
- Click on the site icon
- Select "CHANGE"
- Notice the media picker is opened
- Select an image from the "Choose from device" option
- Notice the image cropper is opened
- Crop your icon and confirm
- The site icon upload starts
- The site icon progress bar shows
- After the site icon upload finishes successfully the icon is updated and the progress bar is hidden
- Test this with the other media picker options - "Take a photo", "WP Media library" and "System picker"

To test 2:
- Click on the site icon
- Select "REMOVE"
- Notice the progress bar is shown for a short while
- Notice the icon gets removed and the placeholder is visible
- Click on the site icon again
- Notice the dialog is now different with only 2 options

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
```diff
@@ -1087,16 +1083,9 @@ public void onClick(View v) {
                 break;
             case RequestCodes.STORIES_PHOTO_PICKER:
             case RequestCodes.PHOTO_PICKER:
-                Fragment fragment = mBottomNav.getActiveFragment();
-                // TODO move this logic directly to the fragment
-                if (fragment instanceof MySiteFragment) {
-                    fragment.onActivityResult(requestCode, resultCode, data);
-                }
-                break;
+            case RequestCodes.SITE_ICON_PICKER:
```
renanferrari: Since we're launching the media picker from `ImprovedMySiteFragment`, wouldn't it be possible to handle this result directly into that fragment instead of having to handle it here first?

WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
```diff
@@ -155,15 +192,55 @@ class ImprovedMySiteFragment : Fragment(),
...
+                .withAspectRatio(1f, 1f)
+                .withOptions(options)
+                .start(requireActivity(), this)
     }
 
     override fun onResume() {
         super.onResume()
-        viewModel.updateSite((activity as? WPMainActivity)?.selectedSite)
+        selectedSiteRepository.updateSite((activity as? WPMainActivity)?.selectedSite)
+        selectedSiteRepository.updateSiteSettingsIfNecessary()
```
renanferrari: I'm not sure I'm missing something here, but why are we making calls to the repository directly from the fragment?

WordPress/src/main/java/org/wordpress/android/ui/mysite/SiteIconUploadViewModel.kt
```diff
@@ -0,0 +1,90 @@
...
+import org.wordpress.android.ui.uploads.UploadService.UploadErrorEvent
+import org.wordpress.android.ui.uploads.UploadService.UploadMediaSuccessEvent
+import org.wordpress.android.util.AppLog
+import org.wordpress.android.util.AppLog.T.MAIN
+import org.wordpress.android.util.EventBusWrapper
+import org.wordpress.android.util.analytics.AnalyticsTrackerWrapper
+import org.wordpress.android.viewmodel.Event
+import javax.inject.Inject
+
+class SiteIconUploadViewModel
```
renanferrari: What was the rationale behind introducing a new ViewModel just for the site icon upload? 

I don't disagree nor agree with this approach by the way, but unless I'm missing something, it seems that there is no technical obstacle that would prevent us from implementing this in `MySiteViewModel`, so I'm just trying to understand your motivations in case I have to make similar a decision.

renanferrari (commented):
Thanks for the changes. I've tested the app and everything seems to be working as expected. Code also looks pretty solid overall, I just left a few comments and questions below.

WordPress/src/main/java/org/wordpress/android/ui/mysite/SiteIconUploadViewModel.kt
planarvoid: I wanted to extract the upload events related logic to a separate class. However, I could probably create a view lifecycle aware component that would live inside `MySiteViewModel`. It could all live directly in the `MySiteViewModel` but there will be a lot of logic there so I'm trying to pull out as much as possible. When I'm thinking about it, I think it makes sense to use this directly in the `MySiteViewModel` 

edit: I've moved the view model to a handler in 9a94470

WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
planarvoid: Good catch, I've fixed it in ac9ab4d

WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
planarvoid: Good catch, I've fixed it in ac9ab4d

WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
planarvoid: hm, when thinking about it, it doesn't really sound like a good idea. I'll move it to the view model 👍 

edit: moved in 720ccbd

planarvoid:
thanks for the review @renanferrari ! I've implemented all your suggestions. Let me know what you think

WordPress/src/main/java/org/wordpress/android/ui/mysite/SiteIconUploadViewModel.kt
renanferrari: This is looking pretty good, now!

WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
```diff
@@ -155,16 +187,54 @@ class ImprovedMySiteFragment : Fragment(),
...
+                .withAspectRatio(1f, 1f)
+                .withOptions(options)
+                .start(requireActivity(), this)
     }
 
     override fun onResume() {
         super.onResume()
-        viewModel.updateSite((activity as? WPMainActivity)?.selectedSite)
-        viewModel.refreshAccountAvatarUrl()
+        viewModel.refresh((activity as? WPMainActivity)?.selectedSite)
```
renanferrari: Do we actually need to pass the selected site as a parameter here? `WPMainActivity::getSelectedSite` seems to just call `SelectedSiteRepository::getSelectedSite`, so couldn't we call that directly from the view model?

WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteViewModel.kt
```diff
@@ -120,11 +152,9 @@ class MySiteViewModel @Inject constructor(
         _onNavigation.value = Event(OpenSitePicker(site))
     }
 
-    fun updateSite(selectedSite: SiteModel?) {
-        _selectedSite.value = selectedSite
-    }
-
-    fun refreshAccountAvatarUrl() {
+    fun refresh(selectedSite: SiteModel?) {
+        selectedSiteRepository.updateSite(selectedSite)
```
renanferrari: To illustrate what I mentioned in the other comment, wouldn't calling `selectedSiteRepository.updateSite(selectedSiteRepository.getSelectedSite())` here essentially produce the same result?

I'm assuming we're doing this to force a refresh in the repository live data, in which case we could even consider introducing a `SelectedSiteRepository::forceRefresh`.

renanferrari (commented):
Thanks for the changes @planarvoid, they are looking pretty good! I still have one question about the call we make to update the selected site inside `onResume`. More details below.

Another thing I just realized is that the switch site functionality is not working. Correct if I'm wrong, but it looks like we haven't handled that in any past PR and I also don't see an open issue for it. If that's the case, when are we planning to address this?

WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
planarvoid: you are right 👍

WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteViewModel.kt
planarvoid: I don't think this call is necessary at the moment. It's a leftover from when we weren't using the repository. I'll remove it. Updating site settings should be enough 👍 

Both changes implemented in a7249fb

planarvoid:
thanks for the review @renanferrari . You were right about the selected site changes. The main purpose of the repository was that I wanted to remove the calls between the fragment and the activity. 

About the site picker, I wanted to create an issue but in the end it was a one line change so I did it in https://github.com/wordpress-mobile/WordPress-Android/pull/13481/files

It was fixed by adding `setSite` to the site picker.

renanferrari (approved):
Thanks for the changes and further clarifications, @planarvoid! 

LGTM :shipit: 

I'm gonna merge this + #13481 and #13497.

PR #13481, opened 2020-11-26
Site block title settings

### LINKED ISSUE #13343: Create My Site item and ViewHolder

planarvoid:
Fixes #13343 

This PR implements logic for setting site title and adds tests to cover it.

To test:
- Turn on the my site flag and restart the app
- Click on My Site title
- Notice the dialog
- Change your site title
- Save your change
- The title gets updates
- Open the dialog
- Cancel it 
- The app doesn't crash

PR submission checklist:


renanferrari (approved):
Thanks for the changes! I've tested the app and everything seems to be working as expected. Code also looks good.

LGTM! :shipit: 

I'm not going to merge this right away to prevent it from messing the diff in #13478.

PR #13413, opened 2020-11-18
[Gutenberg] File Block : Consolidated Media Picker + WP Media Source integration

jd-alexander:
`gutenberg`: https://github.com/WordPress/gutenberg/pull/26960
`gutenberg-mobile`: https://github.com/wordpress-mobile/gutenberg-mobile/pull/2805
`WPiOS`: https://github.com/wordpress-mobile/WordPress-iOS/pull/15291

## Solution
This solution seeks to address the ability to add a file to the File Block from the device or WordPress media library utilizing the media picker components that are present within the codebase. This PR brings consistent file picking behavior between both the Consolidated Media Picker and the Normal Picker. 

The primary changes are:

1.  introducing a `SINGLE_FILE_PICKER` type in the `MediaBrowserType` so the behavior that's specific to the File Block can be constrained by this picker type. 

2. For the `MediaBrowserActivity` when a single file is being selected the `mBrowserType.isSingleFilePicker()` flag is used to ensure the entire media list is loaded. 

3. `showFilePicker` now accepts `canMultiSelect` so that `onAddFileClicked` can pass the `allowMultipleSelection `

4. The GutenbergEditor's WP Media Library `FileButtonClicked` behavior was wired to the media picker so that a picker with the WP Media Library as it's source could be created.

## Testing
Before testing, I toggled the default value in the file `ConsolidatedMediaPickerFeatureConfig` to switch between each picker type.

#### Picking local files. 

1. Create a new post. 
2. Add a File Block. 
3. Press Choose File. 
4. Select “Choose from device”. 
5. The System Picker should be visible. 
6. You should be able to pick a file. 
7. You should see the file name and the download button activated along with an upload progress.
8. You can preview the post and download the file for further validation. 
Picking WP media library files

Select Option | See Picker 
--------|-------
  <kbd></kbd>  |       <kbd></kbd>

#### Picking WP media files. 

1. Create a new post. 
2. Add a File Block. 
3. Press Choose File. 
4. Select “WordPress Media Library”
5. Select a file. 
6. You should see the file name and the download button activated.
7. You can preview the post and download the file for further validation.

Select Option | See Picker 
--------           |-------
  <kbd></kbd>  | <kbd></kbd>

####   Picking local files. 

1. Create a new post. 
2. Add a File Block. 
3. Press Choose File. 
4. Select “Choose from device”. 
5. The File Picker should be visible. 
6. You should be able to pick a file. 
7. You should see the file name and the download button activated along with an upload progress.
[...]

PR #13482, opened 2020-11-26
[Gutenberg] Enable File block

etoledom:
This PR enables the File Block on the gutenberg block editor.

To test:

- Install the CI build on your device.
- Follow the manual tests document: TBD

PR submission checklist:


hypest:
Gave the File block feature a try via WPAndroid and here are some notes, not sure if those are reported or addressed elsewhere, let me know.

* 🎉 Awesome work on this block! Well done @etoledom and @jd-alexander!
* ✅ block options work nicely for me
* ✅ Having 2 file blocks in the post works nicely too
* ✅ File block worked on an WP.com Business (Atomic) site as well, including when GB is not the default there

* "Choose from device" has an image icon. Is that intentional? Seems to imply only media/image files can be chosen?

* The disabled state of the Download button is too faint to be readable. Perhaps it could be full-white or something? See the button in this screenshot from my Pixel 2XL Android 10:
Screenshot 2020-11-26 at 19 15 26

* Not a big issue at the moment but, also have a look at the various "frames" of the button evolution when I take a video of adding the File block (it happens very quickly for the eye to see the individual frames in realtime but you do notice that some flickering is going on):

<table>
<tr>
	<td>Screenshot 2020-11-26 at 19 24 32</td>
	<td>Screenshot 2020-11-26 at 19 24 39</td>
	<td>Screenshot 2020-11-26 at 19 24 46</td>
	<td>Screenshot 2020-11-26 at 19 24 56</td>
</table>

* Speaking of the "Download" button, it's resizing is rather visible when opening the post for editing. See the gif:
file_block_button_resize_on_open

* I see the "Link to" options being available both via the link toolbar button and via the block options (cog wheel) button as well. Is that by design? Not an issue, just wondering of this "2 ways to reach the same options".

etoledom:
Thanks for the preliminar preview @hypest !

> "Choose from device" has an image icon. Is that intentional? Seems to imply only media/image files can be chosen?

@iamthomasbishop mentioned about using what the other blocks use. But I see he mentioned a mobile phone icon (which we don't really use there). All other blocks (including Video block 🤔 ) use the image icon (even though for the video block I'd guess not intended).

I agree that it looks out of place or at least a bit confusing. I would go (for now) with the file icon as shown here: https://github.com/wordpress-mobile/WordPress-Android/pull/13413#issuecomment-732964377

> The disabled state of the Download button is too faint to be readable. Perhaps it could be full-white or something? See the button in this screenshot from my Pixel 2XL Android 10:

True that. On iOS the text remains white, but for some reason on Android, when the button is dimmed, the text color gets darker and more difficult to see. I'll increase the opacity to make it more readable.

> Not a big issue at the moment but, also have a look at the various "frames" of the button evolution when I take a video of adding the File block

It does look bad. Made a quick check and found an easy work-around to this. Will push it soon.

> I see the "Link to" options being available both via the link toolbar button and via the block options (cog wheel) button as well. Is that by design? Not an issue, just wondering of this "2 ways to reach the same options".

It's by design as shown in the blueprint.

--- 

I'll be adding these small fixes to this same PR.

jd-alexander:
@etoledom thanks for the changes in the Gutenberg PR that's supposed to address the flickering issue above. While testing, I can attest to that the fact that the flickering is now gone but when I open a saved post that has a File Block, the Download button takes a little while before it appears. I think it's because we are waiting on the max-width to be set to a value above 0 before rendering the download button.
I am wondering if there's any way to either layout the entire component when the layout has been called or simply use a hardcoded max-width that's the default and then when the new one is set it is overwritten. I am still thinking about this but just wanted to share my thoughts early for you to process them. 

below is a gif depicting what I am seeing. 

<kbd></kbd>

etoledom:
> I am wondering if there's any way to either layout the entire component when the layout has been called or simply use a hardcoded max-width

This is also a possibility, but in the time the button is not present, instead it will have a wrong visible size for any text of different length than the default text (or default max-width).

Testing on my (fairly low and mid range) device, the time the button is not visible is much shorter an in the simulator from the Gif.

Also, the button block, which this approach is based on, has a similar effect of not being visible until it's lied out properly, on a similar time frame. So I'd say this is acceptable. (There's where I got the idea from)

In the other hand, there are other components that depend on layout metrics, and its layout is visibly changing when they are loaded. So probably both approaches are OK.

I'd personally prefer to hide the visual glitch, but open to change it 👍

jd-alexander:
Thanks for your feedback on this @etoledom I agree with the below statement 

> I'd personally prefer to hide the visual glitch, but open to change it 👍

I also prefer to hide the visual glitch. 

> Testing on my (fairly low and mid range) device, the time the button is not visible is much shorter an in the simulator from the Gif.

This is a good point. When I use Gutenberg on my emulator it is very slow.

hypest:
I can verify that the delay of the Button appearance is there but not too much while testing on my physical Pixel 2XL Android 10. I think this is not a blocker for the time being.

etoledom:
- ✅ **TC001:** Upload completes successfully from the device library.
- ✅ **TC002:** Upload completes successfully from the WordPress Media library.
- ✅ **TC003:** Upload fails from the device library.
- ✅ **TC004:** Stop the file upload before it completes.
- ❌ **TC005:** Close the post with an ongoing file upload.
- ✅ **TC006:** Replace the selected files.
- ✅ **TC007:** Change the alignment in the File Block.
- ✅ **TC009:** Open link in new tab.
- ✅ **TC0010:** Show/hide download button.
- ✅ **TC0011:** Copy file URL.
- ❌ **TC0012: [(Not yet on the document)](https://github.com/wordpress-mobile/test-cases/pull/57/files#r531566870)** Close and reopen post with an ongoing file upload.

Issues:

**TC005:**
After the upload finishes and I open the post again. A red screen appears. [VIDEO HERE](https://cloudup.com/i5bd-hkV8fF)
Looks like something is not working with the Upload Processor.
This is the result HTML:

```html
<!-- wp:file {"id":99,"href":"file:///data/user/0/org.wordpress.android.prealpha/cache/bugreport-daisy-PKQ1-3.180917.001-2019-09-25-13-38-52.zip"} -->
<div class="wp-block-file"><a href="///data/user/0/org.wordpress.android.prealpha/cache/bugreport-daisy-PKQ1-3.180917.001-2019-09-25-13-38-52.zip" target="_blank" rel="noreferrer noopener">bugreport-daisy-PKQ1-3.180917.001-2019-09-25-13-38-52.zip</a><a href="///data/user/0/org.wordpress.android.prealpha/cache/bugreport-daisy-PKQ1-3.180917.001-2019-09-25-13-38-52.zip" class="wp-block-file__button" download>Download he</a></div>
<!-- /wp:file -->
```

**TC0012:**
After opening the post again while the upload is ongoing, the block doesn't seem to reconnect to the upload process.
[VIDEO HERE](https://cloudup.com/icV4AfWdm06)

This is how the expected behavior looks like:
ongoing-expected

Both flows are working on WPiOS, so this might be some issue with the native implementation on WPAndroid.

jd-alexander:
@etoledom I am taking a look at this.

etoledom:
✅  **TC005**: Close the post with an ongoing file upload.
✅  **TC0012**: Close and reopen post with an ongoing file upload.

Re-tested again and it's looking great not! 🎉 

I noticed that the error on TC12 that I experienced is a known issue. It happens when the editor loads while the file is on the *processing* stage. There's no more upload progress changes so the UI won't get updated.

We had this same effect long ago with the Video and Image blocks, and it's easier to see this happening on Android, since the editor load time is longer. Still, for this to happen, the processing time needs to be relatively long, and the user needs to re-open the editor at a precise time. It just happens to me that the file I was testing with takes some long-ish time to process and by re-opening the post just after closing it on WIFI connection (with fast upload times) was triggering this issue.

Just noting that it's not a problem specifically of the File block.
Testing with a bigger file/slower upload time, TC12 this works perfectly.

And I was able to reproduce the issue with the Video block.

loremattei:

iamthomasbishop:
> But I see he mentioned a mobile phone icon (which we don't really use there). All other blocks (including Video block 🤔 ) use the image icon (even though for the video block I'd guess not intended).

@etoledom @jd-alexander Ahh I wonder if we inherited a new icon with recent changes to the WP/G2 icon set 🤔 Do we have access to the `mobile` and `tablet` [icons](https://wordpress.github.io/gutenberg/?path=/story/icons-icon--library)? If so, we can probably just use that. (And if we can conditionally show the `mobile` icon on small screen vs `tablet` on larger, even better!)

Also, if we do end up going this route, it might make sense to apply the same icon to other instances of "choose from device".

jd-alexander:
> @etoledom @jd-alexander Ahh I wonder if we inherited a new icon with recent changes to the WP/G2 icon set 🤔 Do we have access to the mobile and tablet icons? If so, we can probably just use that. (And if we can conditionally show the mobile icon on small screen vs tablet on larger, even better!)

I will take a look and see if see that icon soon. 

> Also, if we do end up going this route, it might make sense to apply the same icon to other instances of "choose from device".

Understood!

etoledom:
Updated with the following changes:
https://github.com/WordPress/gutenberg/pull/27228#issuecomment-738300019

jd-alexander:
Updated with the following changes 
https://github.com/WordPress/gutenberg/pull/27228#issuecomment-738600342

jd-alexander:
I ran thorough tests using all the latest changes and all the test cases worked as expected. 

- ✅ **TC001:** Upload completes successfully from the device library.
- ✅ **TC002:** Insert completes successfully from the WordPress Media library.
- ✅ **TC003:** Upload fails from the device library.
- ✅ **TC004:** Stop the file upload before it completes.
- ✅ **TC005:** Close the post with an ongoing file upload.
- ✅ **TC006:** Replace the selected files.
- ✅ **TC007:** Change the alignment in the File Block.
- ✅ **TC009:** Open link in new tab.
- ✅ **TC0010:** Show/hide download button.
- ✅ **TC0011:** Copy file URL.
- ✅ **TC0012:** Close and reopen post with an ongoing file upload.

jd-alexander (approved):
All is well here. Awaiting final run-through on iOS.

etoledom:
✅ All good on the manual tests on my side too 🎉

PR #13486, opened 2020-11-27
Fixes issue preventing changes on a homepage from being previewed

antonis:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2832

## Description
Fixes issue preventing changes on a homepage from being previewed. This is achieved by appending the page slug to the preview url based on the observation [discussed here](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2832#issuecomment-732999563)

## To test:

1. Go to My Sites > + > Create a new WordPress.com site and complete the site creation flow.
1. Go to My Site > Pages > Home and select any layout from the "Choose a design" modal.
1. Edit your home page content.
1. Tap "Preview".
1. Try editing different pages created by headstart and try previewing changes to those pages.
1. Verify that the preview has the latest changes

PR submission checklist:


libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/UrlUtils.java
```diff
@@ -291,6 +291,26 @@ public static String appendUrlParameters(String url, Map<String, String> paramet
...
+    public static boolean isHomePage(@Nullable String urlString) {
+        if (urlString != null) {
+            Uri uri = Uri.parse(urlString);
+            String path = uri.getPath();
+            return path != null && (path.equals("") || path.equals("/"));
+        }
+        return false;
+    }
+
+    public static @Nullable String appendToPath(@Nullable String url, @Nullable String path) {
```
mkevins, line 309: One question that I have: can `post.getSlug()` be `null`, and if so, do we need to guard here, and perhaps return the passed `url` unchanged?

mkevins:
I've tested this and also ran all unit tests, and the workaround is working as described. I think this is a reasonable fallback if we cannot determine what the root cause is between the respective preview mechanisms on web and mobile in time for release.

WordPress/src/main/java/org/wordpress/android/ui/posts/PostUtils.java
```diff
@@ -545,6 +545,9 @@ public static String getPreviewUrlForPost(RemotePreviewType remotePreviewType, P
                 break;
             case REMOTE_PREVIEW_WITH_REMOTE_AUTO_SAVE:
                 previewUrl = post.getAutoSavePreviewUrl();
+                if (UrlUtils.isHomePage(previewUrl)) {
```
chipsnyder, line 548: @antonis I'm just now catching up on your progress here. Nice work tracking down this fix! I haven't tested this out at all but I'm wondering from your research, would there be any negatives effects if we include the slug in all previews, not just Home Page? 

I don't have any reason to make the change outside of I think that helps a bit with the "I feel that this is just a hack." feeling.

antonis: Tbh I didn't think of this. I tried it now (on the browser) and it seems that appending the slug to a non-home page url (that already has a slug) makes no difference since a redirection occurs. E.g.
`https://antonistest.wordpress.com/redhill/redhill/?preview_id=831&preview_nonce=d50d43d924` redirects to
`https://antonistest.wordpress.com/redhill/?preview_id=831&preview_nonce=d50d43d924`

>I don't have any reason to make the change outside of I think that helps a bit with the "I feel that this is just a hack." feeling.

I think that's the only downside :|

chipsnyder: Ah with having it to a redirect hop that might not be ideal then since we know we can avoid it by only doing it for the special case. Thanks for investigating this! I'm good with only appending the slug for Home pages then.

antonis:
A server side fix has been applied

PR #13491, opened 2020-11-27
Release/16.2

loremattei:
Merge Release/16.2 into `trunk`.

PR submission checklist:


.circleci/config.yml
```diff
@@ -320,7 +324,8 @@ jobs:
       - checkout-submodules
       - android/restore-gradle-cache
       - copy-gradle-properties
-      - update-gradle-memory
+      - update-gradle-memory:
+          jvmargs: "Xmx1024m"
```
AliSoftware, line 328: Could it be this new step that makes the Release Build job fail – as it updates the `gradle.properties` with a new value (1024 instead of the 2048 we used before) and later creates an error in the job when fastlane asks `"gradle.properties has changes that need to be merged. Would you like to see a diff?"` in interactive mode?

AliSoftware (approved):
Like for the PR merging in `develop`, I think the "Release Build" job failure is unrelated and not a blocker for this PR to be merged (but I'll let you decide to hit the green button depending on what you think about this).

At least maybe this diff offers a lead into the reason, see below

mkevins:
I was doing some testing via the apk listed in this PR, and just noticed the filename says that it is for "pr-13490", which seems to be [the PR merging this release back to `develop`](https://github.com/wordpress-mobile/WordPress-Android/pull/13490). Is this intentional / expected? My assumption is that this is only because both PRs are using the same hash (`741989c`), and the "merge back to develop" PR was opened first (so this generated apk is a cache hit).

.circleci/config.yml
loremattei: Yep! It's surely that one :-) We fixed it for beta builds, but missed it for the release build.

loremattei:
Hey @mkevins! Yes, exactly what you said! 👍

PR #13494, opened 2020-11-28
Create Quick Actions block item

### LINKED ISSUE #13345: Create a UI item for the list of quick actions
Quick actions contain stats, pages, posts and media. We need to create an UI item and a ViewHolder to draw this block. The actions should be handled in the ViewModel.

renanferrari:
Fixes #13345

This PR introduces the `QuickActionsBlock` item to the new My Site fragment, which contains the following buttons:

### To test

**Main scenario**

- Turn on the Improved My Site flag and restart the app
- Go to My Site fragment
- Notice the Quick Actions block is now visible
- Check that each button redirects to its respective screen (Stats button to Stats screen, Pages button to Pages screen, etc).

**Alternative scenarios**

- When the user role for the selected site doesn't have the capability to edit pages, the Pages button should not appear.
- The Stats button sometimes redirects to other screens, depending on the selected site type (Jetpack, Self-Hosted, etc). For more info on these conditions, please check the unit tests added in this PR.

---

PR submission checklist:


WordPress/src/main/res/layout/quick_actions_block.xml
```diff
@@ -0,0 +1,84 @@
...
+            style="@style/MySiteQuickActionButton"
+            android:src="@drawable/ic_stats_alt_white_24dp" />
+
+        <org.wordpress.android.widgets.WPTextView
+            style="@style/MySiteQuickActionButtonLabel"
+            android:layout_below="@+id/quick_action_stats_button"
+            android:text="@string/stats" />
+    </RelativeLayout>
+
+    <Space
```
planarvoid, line 27: TIL 👍

planarvoid (approved):
Thanks for the changes! It's looking great 👍

WordPress/src/main/res/layout/quick_actions_block.xml
renanferrari: Btw, this layout was just copied from the old fragment. Once we update the way we show the Quick Start indicators, I believe we can have a simpler structure here using `ConstraintLayout` or something like that.

renanferrari:
@planarvoid Thanks for reviewing this! I've merged the latest changes from `develop` and pushed some additional changes to fix conflicts, so you may want to take another look.

planarvoid (approved):
Thanks for the changes, it's looking good 👍 . I've tested all the flows and noticed one unrelated bug which I've fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/13529

PR #13499, opened 2020-11-30
Fixes NullPointerException on synthetic notifications_list RecyclerView

### LINKED ISSUE #12657: NullPointerException: Attempt to invoke virtual method 'void androidx.recyclerview.widget.RecyclerView.addOnScrollListe...
**Sentry Url**: https://sentry.io/share/issue/af9bf1418add4f589f5f91c794263e7a/
**User Count**: 388
**Count**: 557
**First Release**: org.wordpress.android@15.3-rc-1+891
**First Seen**: 2020-07-17T17:12:12.464000Z
**Last Seen**: 2020-08-12T09:58:23Z
**24 Hours**: 41
**30 Days**: 556

```
NullPointerException: Attempt to invoke virtual method 'void androidx.recyclerview.widget.RecyclerView.addOnScrollListener(androidx.recyclerview.widget.RecyclerView$OnScrollListener)' on a null object reference
     at org.wordpress.android.ui.notifications.NotificationsListFragmentPage$showNewUnseenNotificationsRunnable$1.run(NotificationsListFragmentPage.kt:71)
```

antonis:
Fixes https://github.com/wordpress-mobile/WordPress-Android/issues/12657

Description:
Adds safe call operator to prevent crash. 

*I was not able to reproduce this by posting notifications to the list.* From the crash reports it seems that synthetic notifications_list RecyclerView might be null in some corner case. The breadcrumbs indicate that the user has left the notifications screen before the `notifications update service` completes. My guess is that all this happen in the [1s of the `postDelayed` call](https://github.com/wordpress-mobile/WordPress-Android/blob/2ead79b7bf2e52f7dabec2edb6599b3d93c15abb/WordPress/src/main/java/org/wordpress/android/ui/notifications/NotificationsListFragmentPage.kt#L369). The side effect of the `mOnScrollListener` not being added (because the view is null) is that the `clearPendingNotificationsItemsOnUI` will not be called. In that case this call is irrelevant if the user has left the screen.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/notifications/NotificationsListFragmentPage.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/notifications/NotificationsListFragmentPage.kt
@@ -69,7 +69,7 @@ class NotificationsListFragmentPage : ViewPagerFragment(), OnScrollToTopListener
 
     private val showNewUnseenNotificationsRunnable = Runnable {
         if (isAdded) {
-            notifications_list.addOnScrollListener(mOnScrollListener)
+            notifications_list?.addOnScrollListener(mOnScrollListener)
         }
     }
 


mkevins (approved):
Nice work finding this Antonis! I was also unable to reproduce the issue, but I believe you are correct regarding the race condition you mentioned. I've confirmed that in all the events I analyzed, the last activity to be started was either My Site or Reader (i.e. _not_ Notifications), so indeed it seems an activity change occurs just prior to the crash.

<details>
<summary>Snippet to analyze last activity breadcrumb</summary>

```javascript
let events = await fetch('https://sentry.io/api/0/issues/1790223792/events/?full=1').then(raw => raw.json());

events.map(({entries: [,,{data: {values}}]}) => {
  const activityIds = values.filter(({message}) => message?.includes('activityId'));
  const {timestamp, message} = activityIds[activityIds.length - 1];
  return {timestamp, message};
});
```

</details>

I also observed a separate issue, in ~ 27% of the cases, which I don't think is related to the crash, but may be worth investigating. It seems that the GCM data is limited to 4096 bytes, and in many cases this results in the data being truncated, and JSON parsing exceptions.

<details>
<summary>Snippet to analyze truncated JSON</summary>

```javascript
// ~27% have truncated JSON (GCM limited to 4096 bytes)
events.filter(({entries: [,,{data: {values}}]}) => (
  values.some(({message}) => message?.includes('JSONException') && message?.includes('4096'))
).map(({entries: [,,{data: {values}}]}) => (
  values.map(({message}) => message)
)))
```

</details>

antonis:
Thank you for reviewing this @mkevins and your feedback (TIL how to filter Sentry data in the console 👍 )

> I also observed a separate issue, in ~ 27% of the cases, which I don't think is related to the crash, but may be worth investigating. It seems that the GCM data is limited to 4096 bytes, and in many cases this results in the data being truncated, and JSON parsing exceptions.

In my understanding this is an info level stack trace [coming from here](https://github.com/wordpress-mobile/WordPress-Android/blob/acf98a186373b4fa7e0d96577ec44fcfd58d87bd/WordPress/src/main/java/org/wordpress/android/models/Note.java#L545). I'm not sure how this can be improved further since the `null` `Note` returned seems to be handled correctly in the code 🤷 
I'll be happy to investigate this further in a separate issue if you have any suggestions.

mkevins:
> TIL how to filter Sentry data in the console

There might be better ways.. this was kind of a quick and dirty way I used, which seems to work ok :sweat_smile: 

> I'll be happy to investigate this further in a separate issue if you have any suggestions

I timeboxed this a bit investigating on the backend, but didn't get to the bottom of it yet.

PR #13500, opened 2020-11-30
Issue/13417 fix site creation restoration

### LINKED ISSUE #13417: IllegalStateException: Site already exists - seems like an issue with domain suggestions endpoint
**Sentry Url**: https://sentry.io/share/issue/0ea7aaaad4c24aa7aa4efe640a719494/
**User Count**: 290
**Count**: 1941
**First Release**: org.wordpress.android-12.6
**First Seen**: 2019-06-19T13:33:15Z
**Last Seen**: 2020-11-18T07:40:42Z
**24 Hours**: 6
**30 Days**: 119

```
IllegalStateException: Site already exists - seems like an issue with domain suggestions endpoint
     at org.wordpress.android.ui.sitecreation.services.SiteCreationServiceManager$createSite$1.invokeSuspend(SiteCreationServiceManager.kt:132)
```

malinajirka:
Fixes #13417 

Note: I think it'd be good if this PR was reviewed by 2 devs, since it touches a vital flow.

I've looked into #13417 more deeply and I noticed a few interesting things:
1. There is a note that the device is low on memory almost in all the stack-traces in Sentry
2. The app crashes within 2 seconds after it's re-opened
3. The app crashes when the site gets created on the server before the app is closed/killed

I was able to reproduce the crash
1. Go to developer settings
2. Turn on "Don't keep activities"
3. Set background process limit to 0
4. Open the app
5. Go through the SiteCreation flow
6. When you are on the final preview screen, switch to a different app
7. Switch back to WPAndroid and notice the app crashes (you won't see ANR or a crash dialog since it crashes in background. However you should see the crash in LogCat + you can notice that the Preview screen is not shown after the crash) 

I realized we don't handle state restoration in the SiteCreationPreviewFragment + VM. This PR fixes this issue.

Few notes about the existing logic to help you better understand the code:
1. SiteCreationPreview fragment is responsible for the following
    - starting a Service which communicates with the server and takes care of the actual site creation
    - listening to changes in the service and showing a corresponding state on the UI  -> Progress/Error/Content(preview)
    - when the service completes, the fragment keeps showing the progress screen and starts preloading the site url into a webview -> when the webview finishes loading the site url, it shows the content

To test:
1. Verify the above steps work after the changes
- test anything else you can think of

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/sitecreation/SiteCreationMainVM.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/sitecreation/SiteCreationMainVM.kt
@@ -28,6 +28,7 @@ import javax.inject.Inject
 
 const val TAG_WARNING_DIALOG = "back_pressed_warning_dialog"
 const val KEY_CURRENT_STEP = "key_current_step"
+const val KEY_SITE_CREATION_COMPLETED = "key_site_creation_completed"
 const val KEY_SITE_CREATION_STATE = "key_site_creation_state"
 
 @Parcelize
@@ -76,6 +77,7 @@ class SiteCreationMainVM @Inject constructor(
             tracker.trackSiteCreationAccessed()
             siteCreationState = SiteCreationState()
         } else {
+            siteCreationCompleted = savedInstanceState.getBoolean(KEY_SITE_CREATION_COMPLETED, false)
             siteCreationState = requireNotNull(savedInstanceState.getParcelable(KEY_SITE_CREATION_STATE))
             val currentStepIndex = savedInstanceState.getInt(KEY_CURRENT_STEP)
             wizardManager.setCurrentStepIndex(currentStepIndex)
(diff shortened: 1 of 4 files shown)


jd-alexander:
Thanks for these changes @malinajirka I glanced at the code and the overall approach looks good so far. I am going to be doing a thorough review and some tests tomorrow. I am just notifying you that I have seen this PR.

antonis (approved):
Great work @malinajirka 👍 
I run a few test and I confirm that the bug is reproduced on the dev branch and is fixed with this PR.
The code LGTM 🎉

develric:
Hi folks 👋 , I see already a couple of people on this so I will let @jd-alexander complete the review here. Just feel free to ping me in case 🙇

jd-alexander:
@develric I didn't get to this today. I can review tomorrow but if you come online before me you could definitely do the verification since @antonis already did a review. Thanks much 👍

develric (approved):
Great catch @malinajirka and super clean solution and PR description 🙇 ! I tested it on a OnePlus 7T API29 and confirm the issue and fix. Tested both creating a new site for an already existent user and adding during the signup flow. No issue found, good job 👍 . :shipit:

malinajirka:
Thank you all ;)!

PR #13506, opened 2020-12-01
Create my site list item

### LINKED ISSUE #13347: Create a UI item for standard menu item in the improved My Site
Create a UI item for the standard menu item - this includes layout, view holder and the action processed by the ViewModel.

planarvoid:
Fixes #13347 

This PR looks massive but the majority of it is unit tests. I wanted to carefully test all the list items logic. It's basically copy & paste of the current logic from MySiteFragment. Some parts might not make 100% sense but now we can look into some kind of consolidation

To test - use one device with develop and one with this PR
- Open the app, turn on my site flag and restart the app
- Notice there are now list items on the My Site screen
- Compare the items between both devices
- Try for various site types - self-hosted, Jetpack, WPCom, being an administrator and a contributor
- The items should be exactly the same in the same order
- Nothing is clickable yet

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mysite/SiteItemsBuilder.kt
```diff
@@ -3,20 +3,69 @@ package org.wordpress.android.ui.mysite
...
 
 class SiteItemsBuilder
 @Inject constructor(
-    private val siteCategoryItemBuilder: SiteCategoryItemBuilder
+    private val siteCategoryItemBuilder: SiteCategoryItemBuilder,
+    private val siteListItemBuilder: SiteListItemBuilder
 ) {
-    fun buildSiteItems(site: SiteModel): List<MySiteItem> {
+    fun buildSiteItems(site: SiteModel, onClick: ListItemInteraction): List<MySiteItem> {
         val siteItems = mutableListOf<MySiteItem>()
```
renanferrari: Maybe we can make this whole code block a little bit more readable if we use `mutableListOf<MySiteItem>().apply` here instead of calling `siteItems` multiple times. What do you think?

WordPress/src/main/java/org/wordpress/android/ui/mysite/SiteItemsBuilder.kt
```diff
@@ -3,20 +3,69 @@ package org.wordpress.android.ui.mysite
...
+        siteItems.addItemIfNotNull(siteListItemBuilder.buildPlanItemIfAvailable(site, onClick))
```
renanferrari: Maybe we could also rename `addItemIfNotNull` to `addIfNotNull` to match the original `add`.

renanferrari (commented):
Thanks for the changes! Code is looking pretty good, I just left a couple of minor suggestions below.

As for the overall behavior, I did some extensive testing and most things seems to be working as expected, although I did find two inconsistencies:

1. When selecting a Jetpack site, the "Jetpack Settings" item is not being shown: 
1. The "View Admin" item shouldn't be shown for WP.com sites, although I believe this bug is not related to the changes introduced here.

WordPress/src/main/java/org/wordpress/android/ui/mysite/SiteItemsBuilder.kt
planarvoid: that's a great idea 👍  I've looked into it and came up with a much better solution using `listOfNotNull` - 4e3cc6d

WordPress/src/main/java/org/wordpress/android/ui/mysite/SiteItemsBuilder.kt
planarvoid: this is gone now

planarvoid:
Thanks for the review @renanferrari . I've implemented the missing jetpack item and cleaned up SiteItemsBuilder

WordPress/src/main/java/org/wordpress/android/ui/mysite/SiteItemsBuilder.kt
renanferrari: Didn't know `listOfNotNull` was a thing, it seems perfect to this!

renanferrari (approved):
Thank you for the changes! I've tested the app again and everything seems to be working as expected.

LGTM :shipit:

PR #13507, opened 2020-12-01
Add jetpack package

zwarm:
Parent Issue #13329 

This PR introduces a `jetpack` package to support the Jetpack Mobile backup and restore features.

There are two packages within `jetpack`
- `backup` -  the process related to creating the downloadable backup file. 
- `rewind` - the existing rewind feature (in the latest design docs this is called *restore*.)  I am open to renaming this to `restore`, but for now I chose to stay consistent. 

This PR also moves the existing progress checkers and status service out of the `activitylog` package and into their associated package under `jetpack`.  I renamed the `download` classes to `backup` because it makes sense now that the feature is contained in the proper package. 

I had originally named the package `backup` (first commit), but as I got deeper into coding classes this afternoon something felt off with my naming. Same goes for using the word `download` on its own. I feel that using jetpack makes the structure easier to understand and makes for a clearer distinction for the features within it. Of course I am open to all other suggestions. 

To test:
- Fire up the app and make sure the rewind function operates as expected

PR submission checklist:


WordPress/src/test/java/org/wordpress/android/ui/jetpack/BackupStatusServiceTest.kt
```diff
@@ -46,7 +48,7 @@ class BackupDownloadStatusServiceTest {
     @Mock private lateinit var backupDownloadProgressChecker: BackupDownloadProgressChecker
     @Mock private lateinit var site: SiteModel
 
-    private lateinit var backupDownloadStatusService: BackupDownloadStatusService
+    private lateinit var mBackupStatusService: BackupStatusService
```
malinajirka: It seems your AS keeps adding "m" prefix even for kotlin files.

ParaskP7: Minor (🔍 ): I am not sure if we still want to keep using the Hungarian notation. If not, please consider removing the `m` suffix from this field. Wdyt?

WordPress/src/test/java/org/wordpress/android/ui/jetpack/BackupProgressCheckerTest.kt
```diff
@@ -20,18 +20,22 @@ import org.wordpress.android.fluxc.store.ActivityLogStore
...
+class BackupProgressCheckerTest {
     @Mock lateinit var activityLogStore: ActivityLogStore
     @Mock lateinit var site: SiteModel
     private val restoreId = 1L
     private lateinit var backupDownloadProgressChecker: BackupDownloadProgressChecker
 
     @Before
     fun setUp() {
-        backupDownloadProgressChecker = BackupDownloadProgressChecker(activityLogStore, TEST_SCOPE)
+        backupDownloadProgressChecker = BackupDownloadProgressChecker(
```
malinajirka: Did you intend to rename even the BackupDownloadProgressChecker?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupProgressChecker.kt
```diff
@@ -1,4 +1,4 @@
-package org.wordpress.android.ui.activitylog
+package org.wordpress.android.ui.jetpack.backup
```
ParaskP7: Warning (⚠️ ): I am seeing that although the file name was changed to `BackupProgressCheck` the class name is still `BackupDownloadProgressChecker`. Consider removing the `Download` word from the class name as well and have both the file and class name match.

ParaskP7 (commented):
Thanks for including me in this review @zwarm ! 🌟 

A have only a couple of comments on this PR.

Minor (🔍 ) In addition, I would suggest that the `test` and `main` folder structure of the `backup` and `rewind` feature match. For example, for the `main` source set the `jetpack` package structure includes the `backup` and `rewind` sub-packages, but the same is not happening on the `test` source set and the `jetpack` package within. Wdyt?

malinajirka (commented):
Thanks @zwarm!

> I feel that using jetpack makes the structure easier to understand and makes for a clearer distinction for the features within it. 

💯 

> This PR also moves the existing progress checkers and status service out of the activitylog package and into their associated package under jetpack. 

Good call ;)

> I renamed the download classes to backup because it makes sense now that the feature is contained in the proper package.

Note 1: It seems there are still some places (classes and field names + FluxC) which use the "backupDownload" convention. Is there some logic behind it or do you plan to rename everything iteratively?

Note 2: I don't want to suggest that we shouldn't rename it, I just want to share my point of view. When I read "BackupStatusService.start()" my first thought is that the service takes care of creating a backup (not that it prepares an already existing backup for download). Similar goes for `BackupProgressCheckerTest`. Wdyt?

WordPress/src/test/java/org/wordpress/android/ui/jetpack/BackupStatusServiceTest.kt
zwarm: Thanks @ParaskP7 - I didn't realize AS did that. Yeah, I'll be changing that

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupProgressChecker.kt
zwarm: @ParaskP7 - Sounds good

WordPress/src/test/java/org/wordpress/android/ui/jetpack/BackupStatusServiceTest.kt
zwarm: Yeah, @ParaskP7 called out the same thing above. I'l be changing that.

ParaskP7: Yea, this is indeed strange, you got to love AS! 😅 Many thanks

WordPress/src/test/java/org/wordpress/android/ui/jetpack/BackupProgressCheckerTest.kt
zwarm: @malinajirka 
> Note 1: It seems there are still some places (classes and field names + FluxC) which use the "backupDownload" convention. Is there some logic behind it or do you plan to rename everything iteratively?

I am on the fence about renaming everything in FluxC because having `download` there provides additional context since the logic sits outside of WP. 

> Note 2: I don't want to suggest that we shouldn't rename it, I just want to share my point of view. When I read "BackupStatusService.start()" my first thought is that the service takes care of creating a backup (not that it prepares an already existing backup for download). Similar goes for BackupProgressCheckerTest. Wdyt? 

This has been rattling my brain since I named these. *Backup* is the language that everyone has been using when referring to  *preparing the downloadable backup file*, but like you said the backup is already there. I have been torn using ubiquitous language or coming up with a different name that feels clear (and is not verbose). I _thought_ BackupDownload was it, but it just keeps rubbing me the wrong way and no way does "download" feel like it should stand on it's own (absolutely no context there).. 🤦  Names are hard. I can just bite the bullet and use BackupDownload - It was my first choice and perhaps that instinct was correct. 

Should the package name change to match 'BackupDownload' then?

WordPress/src/test/java/org/wordpress/android/ui/jetpack/BackupStatusServiceTest.kt
zwarm: Pushed with c3d265e

zwarm: Pushed in c3d265e

zwarm:
> but the same is not happening on the `test` source set and the `jetpack` package within. Wdyt?

@ParaskP7  I added the packages to test structure in c3d265e.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupProgressChecker.kt
zwarm: @ParaskP7  - I talked myself back into using BackupDownload, so all should be consistent again.

ParaskP7:
Many thanks for [c3d265e](https://github.com/wordpress-mobile/WordPress-Android/commit/c3d265efd8aaa0e8040e8ae78e267586271f323b) @zwarm ! 🌟

WordPress/src/test/java/org/wordpress/android/ui/jetpack/BackupProgressCheckerTest.kt
malinajirka: > I am on the fence about renaming everything in FluxC because having download there provides additional context since the logic sits outside of WP.

I see, make sense.

> This has been rattling my brain since I named these. Backup is the language that everyone has been using when referring to preparing the downloadable backup file, but like you said the backup is already there. I have been torn using ubiquitous language or coming up with a different name that feels clear (and is not verbose). I thought BackupDownload was it, but it just keeps rubbing me the wrong way and no way does "download" feel like it should stand on it's own (absolutely no context there).. 🤦 Names are hard. I can just bite the bullet and use BackupDownload - It was my first choice and perhaps that instinct was correct.

"Names are hard" - I completely agree :D. I think it might be easier to decide when we see all the classes/packages. So I'd personally postpone this decision for later and wouldn't worry about it now. For now, we can just consistently follow one of these approaches. Wdyt?

malinajirka (approved):
LGTM, thanks!

WordPress/src/test/java/org/wordpress/android/ui/jetpack/BackupProgressCheckerTest.kt
zwarm: > So I'd personally postpone this decision for later and wouldn't worry about it now. 

***YES***!! I changed the other classes back to BackupDownload in this PR and am moving forward with that. I'm leaving the package as Backup, but will use BackupDownload prefix for Activities, Frags, VM's and the rest. Going to merge and get back to my skeletons.

Issue #13501, opened 2020-11-30
Investigate the ability to remove some of stale code from Site Confirmation Previews


chipsnyder:
With the changes in https://github.com/wordpress-mobile/WordPress-Android/issues/13410 we might be able to clean up this chunk of code:

https://github.com/wordpress-mobile/WordPress-Android/blob/2a52dfae9ba2f934d5f17561948492440643fc4d/WordPress/src/main/java/org/wordpress/android/ui/sitecreation/previews/SiteCreationPreviewFragment.kt#L311-L319

Questions:
1. Is `wpadminbar` being successfully hidden by the backend?

This is what appears as the `wpadminbar` in the webinspector:

<kdb><a href="https://user-images.githubusercontent.com/3384451/100648770-69f32300-330f-11eb-83b6-0fb04d8c73c8.png"><img width="300" src="https://user-images.githubusercontent.com/3384451/100648770-69f32300-330f-11eb-83b6-0fb04d8c73c8.png"/></a></kdb>

chipsnyder:
I used the custom UserAgent `wp-android/16.2` in a webview to test out if this was enough to hide the admin bar as well and that worked nicely so I think this is method is safe to remove now.

PR #8382, opened 2018-10-01
Hide get started bar with javascript.

khaykov:
Fixes #8233 

Hide "Get started" bar using javascript. At least until we implement a proper server-side solution.
This PR uses same approach as [iOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/10040).

[Image from Gyazo](https://gyazo.com/e3d454b40d3f683d6151560dbb8a5198)

To test:

1. Create an account
2. Go through the site creation flow
3. Finish the process
4. The Site preview opens.
5. Make sure the "Get started" bar is not visible at the top of the preview screen.

PR #13512, opened 2020-12-01
Remove injected CSS/JS that is no longer needed

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/WordPress-Android/issues/13501

### To test:
1. Navigate to the Create WordPress.com site
    - Choose site > ➕ > Create WordPress.com site
1. Choose a Design or Select Skip
1. Search for and choose a domain
1. Select Create Site
1. Wait for your site to be created and for the preview to load
1. **Expect** 
    - To not see a header like the admin bar, the cookie notice or the create a page on WordPress.com notice
    - You also shouldn't be able to click any of the buttons or links on the preview webview. 
        - The menu will open but you shouldn't be able to navigate within there

📓 Just to tie the original PR to this one. This code was originally added here: https://github.com/wordpress-mobile/WordPress-Android/pull/8382

## PR submission checklist:

### DIFF
--- a/WordPress/src/main/res/layout/site_creation_preview_web_view_container.xml
+++ b/WordPress/src/main/res/layout/site_creation_preview_web_view_container.xml
@@ -42,7 +42,7 @@
             </FrameLayout>
         </com.google.android.material.card.MaterialCardView>
 
-        <org.wordpress.android.ui.WPWebView
+        <org.wordpress.android.widgets.NestedWebView
             android:id="@+id/sitePreviewWebView"
             android:layout_width="match_parent"
             android:layout_height="match_parent"
--- a/WordPress/src/main/java/org/wordpress/android/ui/sitecreation/previews/SiteCreationPreviewFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/sitecreation/previews/SiteCreationPreviewFragment.kt
@@ -4,7 +4,6 @@ import android.animation.Animator
 import android.animation.AnimatorListenerAdapter
 import android.animation.AnimatorSet
 import android.animation.ObjectAnimator
-import android.annotation.SuppressLint
 import android.content.Context
 import android.os.Bundle
 import android.text.Spannable
@@ -14,7 +13,6 @@ import android.view.View
 import android.view.View.OnLayoutChangeListener
 import android.view.ViewGroup
 import android.view.animation.DecelerateInterpolator
-import android.webkit.WebView
 import androidx.annotation.LayoutRes
 import androidx.appcompat.app.AppCompatActivity
 import androidx.core.content.ContextCompat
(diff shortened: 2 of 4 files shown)


malinajirka (commented):
Thanks @chipsnyder!

The app works as described ;). One thing we might want to do is remove the `hideGetStartedBar` logic from the VM and its tests, wdyt?

Probably unrelated: I noticed I can't even scroll within the webview. I believe it was supported when the site creation flow was introduced. It feels weird specially in the landscape mode as not even the title of the site fits the screen (and I'm not even using a small display). Wdyt?

chipsnyder:
Hey @malinajirka 👋 

> The app works as described ;). One thing we might want to do is remove the hideGetStartedBar logic from the VM and its tests, wdyt?

Oh gosh, I'm a dunce 🤦 Yup I'll take care of that too. I think I was on too much of an autopilot there thanks for pointing this out. 

>Probably unrelated: I noticed I can't even scroll within the webview. 

I noticed this as well, on iOS, you can scroll in that preview so I thought it was odd that there was a difference but each platform matched 16.2 so I wasn't sure what the intended experience was. I agree though I didn't call it out in the description because I felt like it was unrelated. It's a small fix though I'd be happy to explore it here if you don't mind me expanding the scope of this PR.

> It feels weird specially in the landscape mode as not even the title of the site fits the screen (and I'm not even using a small display). Wdyt?

I think the scrolling really depends on if we want users to really investigate their created site or just see "proof" that it did work. But ultimately I don't see a reason to prevent the scroll.

>I believe it was supported when the site creation flow was introduced. 

This helps a bunch. If this used to work this way then I say we enable it. I went ahead and added this to the PR.

WordPress/src/main/res/layout/site_creation_preview_web_view_container.xml
```diff
@@ -42,7 +42,7 @@
             </FrameLayout>
         </com.google.android.material.card.MaterialCardView>
 
-        <org.wordpress.android.ui.WPWebView
```
chipsnyder, line 45: @malinajirka Here's the change for the scrolling. I'll be honest I didn't do a deep dive into this, I just saw that `NestedWebView` is what we're using here: https://github.com/wordpress-mobile/WordPress-Android/blob/95b15937bb6a342f3cdfb80e8e5e744a0c0aff82/WordPress/src/main/res/layout/home_page_picker_preview_fragment.xml#L44 and it seems like this has the same nested view hierarchy problem mentioned here https://github.com/wordpress-mobile/WordPress-Android/blob/95b15937bb6a342f3cdfb80e8e5e744a0c0aff82/WordPress/src/main/java/org/wordpress/android/widgets/NestedWebView.kt#L14-L22

Basically just calling this out incase your more familiar with this class.

malinajirka: I see, nice catch! This bug was most probably introduced when we re-themed/designed AppBar on all screens of the app.

malinajirka:
LGTM, thanks! ;)

> Oh gosh, I'm a dunce 🤦 Yup I'll take care of that too. I think I was on too much of an autopilot there thanks for pointing this out.

No worries at all ;). It happens to all of us.

malinajirka (approved):

Issue #7954, opened 2018-06-28
Android: Add “Duplicate” or "Copy post" functionality


nozomimimi:
### Expected behavior
Currently on the mobile app, "Duplicate post" functionality is not available. (It exists on Calypso, see the image below)
We should bring this functionality to the mobile app.

[![https://gyazo.com/0eb5ef47b7f9c8c333ce096ae30f8895](https://i.gyazo.com/0eb5ef47b7f9c8c333ce096ae30f8895.png)](https://gyazo.com/0eb5ef47b7f9c8c333ce096ae30f8895)

osullivanchris:
Behaviour per Calypso:

- When a user chooses the 'copy' option on a post, they are launched straight into the editor
- All of the content from the existing post is copied over to the new post. However the duplicate post acts as a 'new' post in that it is a draft without any status selected yet, regardless of whether it was duplicated from a published or draft post
- The post has the exact same title name. So "Superman or Batman" creates another post also called "Superman or Batman", not appended by v2, copy, etc
- The Android design can be implemented with existing UI, just adding to the menu
- The iOS design is dependent on https://github.com/wordpress-mobile/WordPress-iOS/issues/11646 where we have the more flexible sheet menu

![Blueprint Android](https://user-images.githubusercontent.com/50576199/61964868-dc8b1a80-afc6-11e9-939e-0a0862a76601.png)
![Blueprint iOS](https://user-images.githubusercontent.com/50576199/61964869-dc8b1a80-afc6-11e9-9ef1-833f2b90b590.png)

osullivanchris:
iOS designs ready too. Should we create a separate iOS task or change the name / location of this task?

megsfulton:
🚢🚢🚢

antonis:
Fixed with https://github.com/wordpress-mobile/WordPress-Android/pull/13521

PR #13521, opened 2020-12-02
Adds Duplicate/Copy post functionality

antonis:
**Fixes** https://github.com/wordpress-mobile/WordPress-Android/issues/7954

Depends on: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1794

## Description
This PR implements the "Duplicate post" functionality (already exists on Calypso)

<details>
 <summary>Design</summary>
 
Blueprint Android
</details>

*Note: for post with conflicts or autosave versions I chose to redirect to the edit flow. It is debatable if this is the right approach but duplicating the edit flow conflict handling for the copy functionality seemed to add unnecessary complexity in my view*

### Duplicate published post
1. From *My Site* press the *Posts* button or scroll down and select *Blog Posts* from the list
1. Create a new post or open a post by tapping on it
1. Take a note of the tile and content
1. In *post cards* mode press the *More button* / In *post list* mode press the *dots* at the right of the post
1. **Verify** that the *Duplicate* button is visible as a last item
1. Press the *Duplicate* button
1. **Verify** that the editor opens with the title and content from step 3
1. Make a minor change
1. Publish or exit to save the post as draft
1. **Verify** that both the original and the duplicate post are listed

### Duplicate draft post
1. From *My Site* press the *Posts* button or scroll down and select *Blog Posts* from the list
1. Select the *Drafts* (second) tab
1. Create a new post or open a post by tapping on it
1. Take a note of the tile and content
1. In *post cards* mode press the *More button* / In *post list* mode press the *dots* at the right of the post
1. **Verify** that the *Duplicate* button is visible as a last item
1. Press the *Duplicate* button
1. **Verify** that the editor opens with the title and content from step 4
1. Make a minor change
1. Publish or exit to save the post as draft
1. **Verify** that both the original and the duplicate post are listed

*Note: For the steps below I'm not sure how to consistently create conflicts for testing. Any input on this is welcome.*

### Duplicate post with conflict
1. From *My Site* press the *Posts* button or scroll down and select *Blog Posts* from the list
1. Select a post with conflicts or a newer autosave version
1. In *post cards* mode press the *More button* / In *post list* mode press the *dots* at the right of the post
1. **Verify** that the *Duplicate* button is visible as a last item
1. Press the *Duplicate* button
1. **Verify** that the *Conflict dialog*(check screenshot below) appears

### Edit post with conflict
1. Follow the steps from **Duplicate post with conflict** above
1. Select *Edit the post first*
1. **Verify** that the conflict resolution dialogs appear and the post opens for editing

### Duplicate local post
1. Follow the steps from **Duplicate post with conflict** above
1. Select *Copy the version from this app*
1. **Verify** that the editor opens
1. Make a minor change
1. Publish or exit to save the post as draft
1. **Verify** that both the original and the duplicate post are listed

## Screenshots

|Published posts card list actions|Draft posts list actions|Conflict dialog|
|---|---|---|
|device-2020-12-03-132343|device-2020-12-03-132356|device-2020-12-03-151244|

PR submission checklist:


antonis:
Hey @osullivanchris 👋 
I've wrangled this issue during groundskeeping based on your [design](https://github.com/wordpress-mobile/WordPress-Android/issues/7954#issuecomment-515509809). Though the code is still draft the basic functionality has been implemented.
I'd appreciate any feedback on this 🙇

osullivanchris:
@antonis Thanks for looking at this, and sharing the build! I tried it out and everything works as expected 😄 

Just one note, per [the discussion here](https://github.com/wordpress-mobile/WordPress-iOS/issues/9662#issuecomment-729817942) we should change the label to 'duplicate' rather than 'copy' for consistency, as it sounds like we can't use 'copy' on iOS.

antonis:
Thank you for the prompt response @osullivanchris 👍  I really appreciate this 🙇 

> Just one note, per [the discussion here](https://github.com/wordpress-mobile/WordPress-iOS/issues/9662#issuecomment-729817942) we should change the label to 'duplicate' rather than 'copy' for consistency, as it sounds like we can't use 'copy' on iOS.

Oh, I missed this 🤷  I'll change the label in the next build.

malinajirka:
Thanks @antonis! I run out of time today and I haven't finish the review yet. But I think I found an issue and I have one question.

Idea: Have you considered createing a copy of the post outside of the EditPostActivity (EPA). The EPA has quite a lot of responsibilities already. I'm wondering if we could create the copy of the post before we actually open the post in the EPA. It feels like creating the copy of the post isn't related to the editor anyway. We could write tests for what we want to and what we don't want to copy from the post (how to handle PostStatus etc).  Wdyt?


Issue I encountered:
The app crashes with the following exception when I copy a post and I leave the editor without making any changes. It crashes only in debug builds, but we might still want to look into it. I think the issue is that we programatically set the title and content, which doesn't invoke autosave. When the user exits the editor we compare the state of the post when the EPA was created vs the current state (we are comparing an empty post with the post we programatically edited).
```
java.lang.RuntimeException: Debug build crash: Post had changes that needed to be saved when exiting the editor, and there was NOT an autosave pending. This means that the editor's autosave mechanism failed. This only crashes on debug builds.
        at org.wordpress.android.ui.posts.EditPostActivity.lambda$updateAndSavePostAsyncOnEditorExit$17$EditPostActivity(EditPostActivity.java:1784)
        at org.wordpress.android.ui.posts.-$$Lambda$EditPostActivity$9NVQt-kzRnzH2hHWsfH1Zj-h1kU.onPostUpdatedFromUI(Unknown Source:6)
        at org.wordpress.android.ui.posts.EditPostActivity.lambda$updateAndSavePostAsync$16(EditPostActivity.java:1749)
        at org.wordpress.android.ui.posts.-$$Lambda$EditPostActivity$qRWqo7UlqWexkHWD3rfBKjpicsg.invoke(Unknown Source:6)
        at org.wordpress.android.ui.posts.EditPostRepository$updateAsync$1.invokeSuspend(EditPostRepository.kt:128)
```

antonis:
Thank you for the prompt response on this @malinajirka 🙇 

> Idea: Have you considered createing a copy of the post outside of the EditPostActivity (EPA). The EPA has quite a lot of responsibilities already. I'm wondering if we could create the copy of the post before we actually open the post in the EPA. It feels like creating the copy of the post isn't related to the editor anyway. We could write tests for what we want to and what we don't want to copy from the post (how to handle PostStatus etc). Wdyt?

That's a good idea 👍 
I haven't actually looked this option. I'm seeing a `jetpack-copy` api call on the web. I will investigate this further since it will also resolve the reported Debug build crash.

osullivanchris:
> Oh, I missed this 🤷 I'll change the label in the next build.

Thanks @antonis !! 👍

antonis:
Hey @malinajirka 👋 

> Idea: Have you considered createing a copy of the post outside of the EditPostActivity (EPA). The EPA has quite a lot of responsibilities already. I'm wondering if we could create the copy of the post before we actually open the post in the EPA. It feels like creating the copy of the post isn't related to the editor anyway. We could write tests for what we want to and what we don't want to copy from the post (how to handle PostStatus etc). Wdyt?

I changed the implementation based on your suggestion. It avoids adding even more complexity in the EPA and doesn't trigger the debug crash 👍 
If/when you find some time please have a look 🙇

mkevins:
This is in good shape Antonis. I've tested the flows described, and it's mostly working as expected. I also ran the unit tests for `PostListMainViewModelCopyPostTest` and everything is passing. I was unable to run the tests for `PostListMainViewModelCopyPostTest` locally, because of a problem with AS 4.1 though.

> _Note: For the steps below I'm not sure how to consistently create conflicts for testing. Any input on this is welcome._

I've found that I could create conflicts by simultaneously editing a published post on both the web and the app, but the same "trick" did not work for a draft.

The one issue I noticed is that when a post is duplicated and closed without changes, the post list does not reflect the addition of the new post until the list is manually refreshed.

antonis:
Thank you for taking the time to review this @mkevins 🙇 

> I've found that I could create conflicts by simultaneously editing a published post on both the web and the app, but the same "trick" did not work for a draft.

That's really helpful for testing. Thank you :) 
 
> The one issue I noticed is that when a post is duplicated and closed without changes, the post list does not reflect the addition of the new post until the list is manually refreshed.

Nice catch 👍 
I've triggered a refresh of the posts list to fix this with f5ab73bb3a8e61520301315832041d4deb324eb0

mkevins (approved):
The new (unedited) duplicate post now shows up without manual refreshing. Nice work Antonis! LGTM!

malinajirka:
Thank you both! I have just quickly scanned through the changes and it looks great overall. 

One thing we might want to consider is invoking something like this:
` mDispatcher.dispatch(ListActionBuilder
                .newListRequiresRefreshAction(PostListDescriptor.calculateTypeIdentifier(postModel.getLocalSiteId())));`
in the `instantiatePostModel` instead of [manually refreshing the list](https://github.com/wordpress-mobile/WordPress-Android/pull/13521/commits/f5ab73bb3a8e61520301315832041d4deb324eb0) on the client. Feel free to ignore this comment if you think it's not worth it or not a good idea.

antonis:
Thank you both for taking the time to review this @malinajirka and @mkevins 🙇‍♂️ 

I agree with your suggestion @malinajirka and moved the list refresh to the `FluxC` project with 47d01d8d0a8d3a1d7382434bec827b8807056b75 👍

PR #13524, opened 2020-12-02
Issue/13333 add p2 site icon

### LINKED ISSUE #13333: Add a branding for P2 sites
As part of P2 improvements, we should add P2 branding where it makes sense.

This includes the P2 logo. The design is TBD.

khaykov:
Fixes #13333

This PR replaces a default globe site icon placeholder for P2 sites.

I replaced it in all the areas where it makes sense, including Reader.

@mattmiklic I'm not sure what was your intention regarding the colors of the P2 icon, so I just matched them to the current ones.

[Image from Gyazo](https://gyazo.com/2c846db37872b06b93787fd020258213)

[Image from Gyazo](https://gyazo.com/a32c4a34dd25ee3ccb2f716a7744bdd4)

[Image from Gyazo](https://gyazo.com/0c26b09e14c980625a017d4eea467c01)

To test:
- Make sure you have a P2 site without an icon and you also follow it in Reader.
- Check the icon of the P2 site in My Site, Site Picker, and Reader Blog Details and make sure it looks good (you might need to reinstall the app to refresh posts in reader since we added a P2 filed to the DB table)

PR submission checklist:


mattmiklic (approved):
Everything looks good here. My intent was to match the colors used for the existing default icon so this is perfect. 👍

develric (approved):
hey @khaykov 👋 , checked this one on my OnePlus 7T API29 and works great (also checked the reader db update script and it's ok 👍 ). Code LGTM. Sharing a thing I noticed but since it's not something introduced by this PR, I think we can eventually discuss/change later (or ignore if we think it's not worth 🙇 )

In the Reader Post Details the `ReaderPostDetailHeaderView.updateBlavatar` if the Url is null (as is the case when no icon is set) we do not show any icon.

| Site icon shown | Site icon not set so nothing is shown |
|---|---|
| image | image |

PR #13525, opened 2020-12-03
Stories: Fix StringIndexOutOfBoundsException when below another block

aforcier:
Fixes https://github.com/Automattic/stories-android/issues/615.

It turns out that the `StringIndexOutOfBoundsException` was happening when the Story block was not the first block in the post. The end of the first block in the post would be matched, causing the substring end index to be earlier than the start index.

Fixed by calculating the end of the Story block relative to its start.

### To test:
* Create a new post in Gutenberg
* Add a block, e.g. a paragraph block
* Below it, add a Story block, and add some slides
* Exit the story editor and try to save the post
* Confirm that the post is saved and there is no crash

PR submission checklist:


mzorz (approved):
Good catch @aforcier ! Excellent, verified it works and yep, it makes sense 👍 
LGTM :shipit:

PR #13529, opened 2020-12-03
Fix logic showing Jetpack header

planarvoid:
I've noticed the Jetpack header should be visible when you're logged in with a self-hosted Jetpack site but you don't have a WPCom account. In this case the site settings are not visible. This PR fixes this use case. 

Screenshot_1606991896

To test:
- Log in with a self-hosted Jetpack site
- Notice the "Jetpack" header is visible
- Notice the "Jetpack settings" item is gone
- Login to a WPCom account (for example through Stats)
- Notice the "Jetpack settings" item appears

PR submission checklist:


renanferrari (approved):
Good catch! Code is looking good and everything seems to be working as expected.

LGTM :shipit:

PR #13520, opened 2020-12-02
Issue/13329 backup download details skeleton

zwarm:
Parent #13329 

This PR contains skeleton classes for Jetpack Mobile Backup 
- The main BackupDownloadActivity and VM
- The BackupDownloadDetailsFragment and VM
- Dagger hookups
- Manifest entry
- Layout files, strings, and styles

It may be easier to look at this in commits
- The first has the layouts files
- The second has skeleton classes
- Final commit is to fix lint because I added my string resources too early. 

Note: This is the first PR in a series that will build out the BackupDownload feature. Styles are preliminary. 

To test:
Nothing to test at the moment. The next PR will focus on creating an entry point to view the backup feature itself.

PR submission checklist:

- [X] I have considered adding unit tests where possible.
- [X] I have considered adding accessibility improvements for my changes.
- [X] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13532, opened 2020-12-03
Issue/13329 expose backup feature

zwarm:
Parent #13329 

This PR adds exposes a temporary entry point for development purposes to the Backup Download feature. When the `BackupFeatureConfig` is enabled, the entry point can be accessed via the Activity Log action item click. The following dialog will be launched and the user can navigate to Backup download or continue using the rewind site feature.

Screenshot

In a future iteration the dialog box will be removed and a true More menu will be displayed. There is a line item task for this in the parent issue.

**To test:**
- Launch the App
- Navigate to Activity Log
- Notice that the action button icon still shows the rewind icon
- Head over to App Settings -> test feature configuration
- Enable the `BackupFeatureConfig` and restart the app
- Notice that the action button is now the 3 dot vertical more icon
- Tap on the icon to show the dialog box
- Tap the "Create downloadable backup" option (You can still select the normal rewind action from the dialog)
- The backup detail view is shown (The feature itself is not functional, this is just to test that the view shows)

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/EventItemViewHolder.kt
```diff
@@ -35,12 +35,12 @@ class EventItemViewHolder(
         summary.text = activity.title
         text.text = activity.description
 
+        ColorUtils.setImageResourceWithTint(
```
zwarm, line 38: I moved this outside of the visibility check to ensure that icon is not null when the diffUtil runs in the adapter. The icon used to be hardcoded in the xml file, but now it's created dynamically based on the Backup Feature config enabled flag.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListActivity.kt
```diff
@@ -5,13 +5,18 @@ import android.os.Bundle
...
 import org.wordpress.android.ui.posts.BasicFragmentDialog
+import org.wordpress.android.util.BackupFeatureConfig
 import org.wordpress.android.viewmodel.activitylog.ACTIVITY_LOG_REWIND_ID_KEY
+import javax.inject.Inject
 
-class ActivityLogListActivity : LocaleAwareActivity(), BasicFragmentDialog.BasicDialogPositiveClickInterface,
+class ActivityLogListActivity : LocaleAwareActivity(),
+        BasicFragmentDialog.BasicDialogPositiveClickInterface,
         BasicFragmentDialog.BasicDialogNegativeClickInterface {
+    @Inject lateinit var backupFeatureConfig: BackupFeatureConfig
```
ParaskP7, line 20: Question (❓ ): Is this injection required here? Did you maybe do is as a preparation for when it will be actually needed afterwards?

ParaskP7 (commented):

Just one question from my side.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListActivity.kt
zwarm: Nope, not needed. I was going to do something there and decided against it. I wonder why it didn't get kicked out when I ran checkstyle. 🤷  Thanks

ParaskP7: Yeap this is indeed tricky, but mainly because of the `@Inject` annotation, plus the `lateinit` thingy, which makes the compiler and tooling (`IDE` in this case) think that this is used anyway. Only when you remove both and make it nullable you get to see it is unused.

Also, `CheckStyle` will not do any good here since it only works for Java files and our current `KtLint` static analysis tool for Kotlin is not helping as well, being old (2 years now) and mainly good for formatting.

`Detekt` to the rescue... soon I hope! 🎉 

Unfortunately, until then, we are as good as the eye can see, authors and reviewer! 💦

zwarm: > `Detekt` to the rescue... soon I hope! 🎉

👍

ashiagr (approved):
Thanks @zwarm! I tested and all looks good to me 👍. I was able to see the 3 dot vertical more icon, the dialog with the option to download backup and the backup detail view on clicking it.

Feel free to merge it once #13520 has been merged.

ashiagr:
Just noticed if the `BackupFeatureConfig` is not enabled then on clicking cancel on the "Rewind Site" dialog opens "Download Backup" screen.

Let's fix it before the merge.

zwarm:
> Just noticed if the `BackupFeatureConfig` is not enabled then on clicking cancel on the "Rewind Site" dialog opens "Download Backup" screen.

Oh my, yes, I'll fix that a.s.a.p. - I don't remember if I even pressed the cancel button. Great find. Thank you.

@ashiagr In commit 7ba3df7.
Thanks again.

ashiagr:
Thanks @zwarm! Issue is fixed now 🎉.

PR #13533, opened 2020-12-03
Stories in the updated my site

### LINKED ISSUE #13531: Implement stories handling to the new my site fragment
We need to support stories handling in the new my site fragment

planarvoid:
Fixes #13531 

This PR is copying some of the Stories logic from MySiteFragment into ImprovedMySiteFragment. I've extracted some parts but unfortunately the code is not really testable (because of how the stories code is written - there are static objects, it uses Intent and the Android classes don't work in unit tests). 

To test:
- Turn on the flag and restart the app
- Click on the FAB on My Site and on "Add Story"
- Create a story with items from your device
- Notice you can finish the story flow and create a story
- Notice the snackbars are visible
- Repeat the flow with media from WP Media library

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
```diff
@@ -82,6 +87,7 @@ class ImprovedMySiteFragment : Fragment(),
     @Inject lateinit var meGravatarLoader: MeGravatarLoader
     @Inject lateinit var mediaPickerLauncher: MediaPickerLauncher
     @Inject lateinit var uploadUtilsWrapper: UploadUtilsWrapper
+    @Inject lateinit var storiesMediaPickerResultHandler: StoriesMediaPickerResultHandler
```
renanferrari: AFAICS, this is not being used in this class. Can we remove it? Let me know if I'm missing something.

WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteViewModel.kt
```diff
@@ -376,9 +379,16 @@ class MySiteViewModel
 
     override fun onCleared() {
         siteIconUploadHandler.clear()
+        siteStoriesHandler.clear()
         super.onCleared()
     }
 
+    fun onStoriesPhotoPicker(data: Intent) {
```
renanferrari: This is just a nitpick, but there are similar methods in this class that are named like `handleXResult`. Maybe we can rename this to something like `handleStoriesPhotoPickerResult` for consistency?

WordPress/src/main/java/org/wordpress/android/ui/stories/StoriesMediaPickerResultHandler.kt
```diff
@@ -2,43 +2,89 @@ package org.wordpress.android.ui.stories
...
+import org.wordpress.android.viewmodel.Event
 import javax.inject.Inject
 
 class StoriesMediaPickerResultHandler
 @Inject constructor() {
+    private val _onNavigation = MutableLiveData<Event<SiteNavigationAction>>()
+    val onNavigation = _onNavigation as LiveData<Event<SiteNavigationAction>>
+
     /* return true if MediaPickerResult was handled */
     fun handleMediaPickerResultForStories(
```
renanferrari, line 28: Since this is now being used only by the old `MySiteFragment` and `PostsListActivity`, couldn't we deprecate it in favor of the newer method of same name? (assuming we want newer classes to use the newer method)

WordPress/src/main/java/org/wordpress/android/ui/mysite/SiteStoriesHandler.kt
```diff
@@ -0,0 +1,105 @@
...
+        eventBusWrapper.register(this)
+    }
+
+    fun clear() {
+        eventBusWrapper.unregister(this)
+    }
+
+    @Subscribe(sticky = true, threadMode = ThreadMode.MAIN)
+    fun onEventMainThread(event: StorySaveResult) {
+        EventBus.getDefault().removeStickyEvent(event)
```
renanferrari: Shouldn't we use `eventBusWrapper` here instead?
```suggestion
        eventBusWrapper.removeStickyEvent(event)
```

WordPress/src/main/java/org/wordpress/android/ui/mysite/SiteStoriesHandler.kt
```diff
@@ -0,0 +1,105 @@
...
+                                )
+                            }
+                    )
+            )
+        }
+    }
+
+    @Subscribe(sticky = true, threadMode = ThreadMode.MAIN)
+    fun onStorySaveStart(event: StorySaveProcessStart) {
+        EventBus.getDefault().removeStickyEvent(event)
```
renanferrari: Same thing as above.
```suggestion
        eventBusWrapper.removeStickyEvent(event)
```

renanferrari (commented):
Thanks for the changes! 

I've tested the app and everything seems to be working as expected, the only issue I found is that the "Uploading" snackbar seems to appear behind the FAB (although I'm not sure this is related to changes introduced here):



For comparison, here's the "Post published" snackbar, which appears correctly:



The code is also looking pretty good, I just left a few minor comments below.

Since this is a feature in active development (AFAIU), I'm also wondering if we shouldn't get a second pair of eyes from someone currently working on it, just to make sure we're not missing anything. This might be unnecessary, but let me know what you think.

WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
planarvoid: you're right 👍  I've stopped using it in one of the refactoring rounds

WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteViewModel.kt
planarvoid: that's a better name for sure 👍

WordPress/src/main/java/org/wordpress/android/ui/stories/StoriesMediaPickerResultHandler.kt
planarvoid: Good idea 👍

planarvoid:
thanks for the review @renanferrari ! I've implemented all of your suggestions in 16032ce

renanferrari (approved):

PR #13544, opened 2020-12-04
Issue/13329 choose items to recyclerview

zwarm:
Parent #13329 

This PR swaps in a recyclerView for the inflexible hardcoded layout to support selecting available items to include in the downloadable backup. At the time of this writing, the available items are not dynamic, so I put them into `BackupAvialableItemsProvider` so it can be reused with the Restore details view. This PR also introduces base UiStates for the BackupDownloadDetails view. The UiStates will continue to be built out in future PRs.

Notes:
- Unit Test are following in PR #13545 
- Design has not been finalized
- Checkboxes default to checked and will not persist across device rotation yet (see todo's)

**Merge instructions**
- Make sure ~~#13520~~ & #13532 have been merged
- Remove Not Ready For Merge label
- Merge as normal

**To test:**
- Enable the BackupFeatureConfig and restart the app
- Navigate to the Activity Log
- Tap on the More menu icon
- Tap the "Create downloadable backup" option
- The backup detail view is shown 
- Tap/untap on the checkboxes

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/details/BackupDownloadDetailsAdapter.kt
```diff
@@ -0,0 +1,66 @@
...
+import androidx.recyclerview.widget.DiffUtil
+import androidx.recyclerview.widget.RecyclerView
+import org.wordpress.android.ui.jetpack.backup.details.BackupDownloadDetailsViewModel.ListItemUiState
+import org.wordpress.android.ui.utils.UiHelpers
+
+class BackupDownloadDetailsAdapter(private val uiHelpers: UiHelpers) :
+        RecyclerView.Adapter<BackupDownloadDetailsViewHolder>() {
+    private val items = mutableListOf<ListItemUiState>()
+
+    override fun onCreateViewHolder(
```
ParaskP7, line 14: 1) Minor (🔍 ): Consider preserving the order of the override functions, just like they appear on the parent class:
- `onCreateViewHolder(...)`
- `onBindViewHolder(...)`
- `getItemId(...)`
- `getItemCount()`
2) Minor (🔍 ) For all those functions consider converting to expression body and removing the explicit type specification (`: Long`). This will make them a one-liner (ish).

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/details/BackupDownloadDetailsFragment.kt
```diff
@@ -30,24 +39,39 @@ class BackupDownloadDetailsFragment : Fragment() {
...
+        recycler_view.layoutManager = LinearLayoutManager(context, RecyclerView.VERTICAL, false)
+        initAdapter()
+    }
+
+    private fun initAdapter() {
+        recycler_view.adapter = BackupDownloadDetailsAdapter(uiHelpers)
+    }
+
+    private fun initViewModel() {
         parentViewModel = ViewModelProviders.of(requireActivity(), viewModelFactory)
```
ParaskP7: Question (❓ ): How would the `parentViewModel` be used later on? Do we need it? I know this was a change from a previous PR, but since I am not seeing it being used yet, I thought I might ask now.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/details/BackupDownloadDetailsFragment.kt
```diff
@@ -30,24 +39,39 @@ class BackupDownloadDetailsFragment : Fragment() {
...
+            when (uiState) {
+                is Loading -> ToastUtils.showToast(requireContext(), "Implement loading")
+                is Content -> showContent(uiState)
+                is Error -> ToastUtils.showToast(requireContext(), "Implement Error")
+            }
+        })
 
-    private fun setupViews() {
-        // TODO: annmarie implement setupViews
+        viewModel.start()
```
ParaskP7, line 69: Warning (⚠️ ): Although the rotation should work, this line gets triggered twice on a configuration change. First ignores the `start()` (as expected) and the next one doesn't (new instance). As such, this defaults every checkbox to their default state from the provider.

This might have to do with the fact that the parent activity shows the fragment during the `setupViewModel()`. Can you check that, I know this is a temporary start, but testing the solution on configuration change seems to not work, which might make it harder to test any additional solution going forward. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/details/BackupDownloadDetailsViewHolder.kt
```diff
@@ -0,0 +1,32 @@
...
+import android.view.ViewGroup
+import androidx.annotation.LayoutRes
+import androidx.recyclerview.widget.RecyclerView
+import kotlinx.android.synthetic.main.backup_details_list_item.view.*
+import org.wordpress.android.R
+import org.wordpress.android.ui.jetpack.backup.details.BackupDownloadDetailsViewModel.ListItemUiState
+import org.wordpress.android.ui.utils.UiHelpers
+
+sealed class BackupDownloadDetailsViewHolder(
+    internal val parent: ViewGroup,
```
ParaskP7: 1) Question (❓ ): Why was the `internal` modifier used here?
2) Minor (🔍 ): The whole `internal val` can be removed, since it is just a past through. Wdyt?
3) Minor (🔍 ): Consider preserving the order in which the fields are defined, matching that of the parent. `layout` first and then `parent`.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/details/BackupDownloadDetailsViewModel.kt
```diff
@@ -1,15 +1,97 @@
...
+                        availableItemType = it.availableItemType,
+                        label = UiStringRes(it.labelResId),
+                        checked = true,
+                        onClick = { onItemClicked(it.availableItemType) }
+                )
+            }
+            // todo: annmarie - swap out the placeholder for date from record
+            Content(
+                    description = UiStringRes(R.string.backup_download_details_description),
+                    items = availableItemsListItems
```
ParaskP7, line 59: Question (❓ ): How would the `1%1s` be replaced in this context?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/BackupAvailableItemsProvider.kt
```diff
@@ -0,0 +1,61 @@
...
+import org.wordpress.android.ui.jetpack.BackupAvailableItemsProvider.BackupAvailableItemType.THEMES
+import org.wordpress.android.ui.jetpack.BackupAvailableItemsProvider.BackupAvailableItemType.PLUGINS
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * This class provides the available item choices for Jetpack Backup Download
+ */
+@Singleton
+class BackupAvailableItemsProvider @Inject constructor() {
```
ParaskP7, line 18: Idea (💡): What if we add a `Dummy` prefix on this provider's name. I recall @malinajirka also did that on one of his dummy providers. This can help us be consistent when using such providers. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/BackupAvailableItemsProvider.kt
```diff
@@ -0,0 +1,61 @@
...
+    fun getAvailableItems(): List<BackupAvailableItem> {
+        return listOf(
+                BackupAvailableItem(
+                        THEMES,
+                        string.backup_item_themes
```
ParaskP7: Warning (⚠️ ): This is probably a wrong import. Consider using `R.string.xyz` instead.

PS: This applies to all the string resources in this provider.

ParaskP7 (commented):

Everything works almost as expected. The only things that seems to not be working as expected is when rotating the screen. I have left an individual comment on that.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/details/BackupDownloadDetailsViewModel.kt
zwarm: Not sure how I am going to handle this - that's why I left the comment 😉

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/details/BackupDownloadDetailsFragment.kt
zwarm: @ParaskP7 - I call out in the notes that rotation isn't addressed in this PR. Thanks for the great detail though, I'll make sure to review when I start to dig into the details. 🙇

WordPress/src/main/java/org/wordpress/android/ui/jetpack/BackupAvailableItemsProvider.kt
zwarm: @ParaskP7  - It's not a dummy class - the data is not coming from the server, it is hardcoded as you see it. I added it to a provider because the potential exists for options to grow *AND* perhaps someday served by the API.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/details/BackupDownloadDetailsFragment.kt
zwarm: @ParaskP7 - this will get flushed out as I move further along in the development. Thanks.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/BackupAvailableItemsProvider.kt
zwarm: Changed in 244789c

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/details/BackupDownloadDetailsViewHolder.kt
zwarm: Changed in 3b2546f
Thanks for the tips

WordPress/src/main/java/org/wordpress/android/ui/jetpack/BackupAvailableItemsProvider.kt
ParaskP7: Oh, my bad, for some reason I thought that this data will become dynamic, please ignore me! 😊 Many thanks for explicating that to me. 🙏

ashiagr (approved):
Thanks both of you! All looks good to me 👍.

PR #13545, opened 2020-12-04
Issue/13329 backup details unit tests

zwarm:
Parent #13329 

This PR adds unit tests for `BackupDownloadDetailsViewModel` and `BackupAvailableItemsProvider`

To test:
Run tests

PR submission checklist:


WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadDetailsViewModelTest.kt
```diff
@@ -0,0 +1,73 @@
...
+        )
+    }
+
+    @Test
+    fun `when available items are fetched, the content view is shown`() = test {
+        val uiStates = initObservers().uiStates
+
+        viewModel.start()
+
+        assertThat(uiStates[0]).isInstanceOf(UiState.Content::class.java)
```
ParaskP7: Minor (🔍 ): The explicit `UiState` can be omitted, or the `UiState.Content` import can be removed and the `UiState` can be made explicit in places required.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadDetailsViewModelTest.kt
```diff
@@ -0,0 +1,73 @@
...
+
+    @Test
+    fun `item is unchecked, when it is clicked on`() = test {
+        val uiStates = initObservers().uiStates
+
+        viewModel.start()
+
+        ((uiStates.last() as Content).items[1]).onClick.invoke()
+
+        assertThat(((uiStates.last() as Content).items[1]).checked).isFalse()
```
ParaskP7, line 47: Minor (🔍 ): Consider a property access syntax as suggest here by the IDE.

PS: The same apply for the `.isTrue()` in the below test.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadDetailsViewModelTest.kt
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Test
+    fun `item is checked, when it is clicked on twice `() = test {
```
ParaskP7: Idea (💡 ): These kind of tests scenarios can benefit from a `given/when/then` pattern. Utilising this pattern, I would have named this test case as: `given item is unchecked, when item is clicked, then item gets checked`. Also I would have asserted the `unchecked` state in the setup, like with the `.isFalse()` above and then, trigger the next `.invoke()` action separately and check the `checked` state afterwards.

ParaskP7 (commented):
Thank you again @zwarm ! 🌟 

Only minors from my sides.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadDetailsViewModelTest.kt
zwarm: @ParaskP7 - interesting, but my IDE is not suggesting property access syntax. 🤷

ParaskP7: @zwarm - Can you check your Kotlin version? We can sit down together to find out why is that.

ashiagr:
Thanks @zwarm! Tests look fine for the current functionality. We can merge this PR once comments mentioned by @ParaskP7 are resolved.

ashiagr (approved):

PR #13548, opened 2020-12-04
WP Stories: changed flow for remote media from WP site to be copied locally

mzorz:
With this change, instead of passing a remote URL to StoryComposerActivity through `onActivityResult()`, we're first making a local copy of it and then passing it over to the composer.

It has proven to have better results and avoid metadata extractor not to be ready when needed in order to calculate the video size and rotation and so on.

I only made this test in the old PhotoPicker. We should follow up on this on the new media picker when it's also used in Stories. EDIT - tracked in https://github.com/wordpress-mobile/WordPress-Android/issues/13549

To test:
1. download [this video](https://cloudup.com/cbGwE3rsv1R) to a your emulator / handset. (this one has proven to show especial trouble on a Pixel 3 emulator, a Pixel 2 actual device, and a Samsung S5e tablet).
2. go to the Media section of the app and upload that video from your handset. Follow the next steps when ready.
3. from the Posts list or from the My Site screen on the app, tap on the FAB and tap on new Story
4. when the media picker appears, tap on the W option to see the site's media items
5. you should see the video you recently uploaded. Select that one and then tap on the ✔️ on the top-right corner of the screen
6. observe a dialog reading "Uploading..." appears briefly (for as long as it takes for the video to get copied), then the Story composer appears and the video starts playing in a loop.
7. go ahead and add some text to your slide
8. publish your story
9. verify it gets saved correctly and it gets published correctly.

Before this change, it would randomly err or crash when trying to save the video.

nice_video

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoPickerActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoPickerActivity.java
@@ -325,12 +325,7 @@ public class PhotoPickerActivity extends LocaleAwareActivity
         if (mediaIds != null && mediaIds.size() > 0) {
             if (mBrowserType == MediaBrowserType.WP_STORIES_MEDIA_PICKER) {
                 // TODO WPSTORIES add TRACKS (see how it's tracked below? maybe do along the same lines)
-                Intent data = new Intent()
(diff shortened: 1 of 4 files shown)


aforcier:
@mzorz looks like some unit tests need to be updated.

aforcier:
This is working well, I followed the test steps and tried various local/wp library image/video uploads for good measure.

:shipit:  once CI is green.

PR #13436, opened 2020-11-20
Stories: Show 'New' tag for Story entry in bottom sheets

aforcier:
Adds a 'New' tag next to the 'Story post' entry in the My site and Post List post creation bottom sheets.

bottom-sheet-after

### To test:

1. Make sure the Stories feature is enabled and you're on a site that supports them (WordPress.com or Jetpack 9.1+)
2. Tap the FAB in the My site, check that the 'New' tag appears next to the Story post entry
3. Do the same from the post list

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13551, opened 2020-12-05
Revert "Stories: Show 'New' tag for Story entry in bottom sheets"

aforcier:
Reverts wordpress-mobile/WordPress-Android#13436, removing the 'New' tag from the Story entry in the bottom sheets.

We only wanted to keep that for one release cycle (and in fact the current release cycle is a full month long), so with the original change set for release in `16.3`, we can remove it from `develop` so it's removed in `16.4`.

bottom-sheet-after

### To test:

1. Make sure the Stories feature is enabled and you're on a site that supports them (WordPress.com or Jetpack 9.1+)
2. Tap the FAB in the My site, check that the Story post entry appears in the bottom sheet
3. Check that there's no 'New' tag next to it
4. Do the same from the post list

PR submission checklist:


mzorz (approved):
Verified, works as expected, removes the tag as described.
LGTM :shipit:

PR #13554, opened 2020-12-07
Use function pointers in ListItemInteraction calls

planarvoid:
When we're using RecyclerView with data classes, we usually compare them by equality. The problem is that lambdas are never equal (even if they have the same body). This works if you're using function references instead (because they are constant). You can't normally use function references if you need to pass a parameter to the function. That's why I've created the `ListItemInteraction` which allows devs create an object with a parameter and a function reference and use it to pass the function to the view holder. In order for this to work we need to use functional references instead of lambdas. This PR replaces the usage of lambdas with the `ListItemInteraction`. 

In case of QuickActions there shouldn't be a big difference in functionality. There should be just some cases when the quick actions block doesn't get redrawn (when another part of the screen gets changed). We should see a difference when we implement the quick start block because at that point we'll be refreshing the item much more often.

To test:
- Check that the quick actions interactions work as expected
- Check the Stats/Insights management item interactions work as expected

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteViewModel.kt
@@ -136,10 +136,10 @@ class MySiteViewModel
             )
             siteItems.add(
                     QuickActionsBlock(
-                            ListItemInteraction.create { quickActionStatsClick(site) },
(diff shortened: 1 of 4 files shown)


renanferrari (approved):
TIL! This distinction went completely under my radar when I was working with this class. Thanks a lot for pointing this out and for introducing helpful comments and tests.

LGTM! :shipit:

PR #13555, opened 2020-12-07
Jetpack Mobile: Scan menu

ashiagr:
Parent #13326

It displays scan menu on My Site for sites with scan capability for a user.

**To test**

Prerequisite:  Make sure both Scan feature flag and MySiteImprovements flag are set to on in the App settings.
(Changes applied only to MySiteImprovements flow)

Test that scan menu is shown on My Site only when the user has scan feature unlocked.

|My Site Scan Menu|
|--------------------|
|screnshot|

Case | Site Type | Login Type | Sub Type  |  User Role | Show Scan |  Scan State API Response (Reason)
-- | -- | -- | -- | -- | -- | --
1a | **.com** |   | simple |  | ❌ | unavailable (wpcom_site)
1b |   |   | atomic | admin | ✅ | available - idle or scanning
1c |   |   |   | non admin | ❌  |  error (authorization_required)
2a | **Jetpack** | wp.com login | plan without scan feature |  | ❌ | unavailable (missing_scan_capability)
2b |   |   | plan with scan feature | admin | ✅ | available - idle or scanning
2c |   |   |   | non admin | ❌  |  error (authorization_required)
2d |   | self-hosted login | plan without scan feature  |  | ❌ |  error (authorization_required)
2e |   |   | plan with scan feature |   | ❌ |  error (authorization_required)

**Questions**

2d, 2e - Scan state rest API gives `authorization_required` error for self-hosted login for a Jetpack site. Should we always display scan menu via this flow? - Answered in https://github.com/wordpress-mobile/WordPress-Android/pull/13555#issuecomment-740625764

Not in scope
Scan menu icon fine-tuning
Click action

PR submission checklist:


malinajirka:
> Scan state rest API gives authorization_required error for self-hosted login for a Jetpack site. Should we always display scan menu via this flow?

Great question! I'd consider asking this question on slack and share the summary of our discussion here. I'd personally vote for making a consistent change across all the jetpack features within the app. We might want to always redirect the user to the .com login flow when they access a jetpack feature and they are using a self-hosted login (if we are sure the site has jetpack installed). If the site doesn't have jetpack installed it's a bit more tricky, but I think it's a different discussion.

ashiagr:
Summary of our discussion from Slack:

2d, 2f - Since scan is accessible only with a paid plan and we don’t know what plan users are on if they're are not using wp.com login, we should keep scan menu hidden on My Site for self-hosted login.

Updated the table in the description.

malinajirka:
Thanks @ashiagr for working on this PR. Sorry I didn't review it sooner. I marked the notification as read so it slipped through my filters.

I have a couple of points for discussion. I'm not sure which approach is better but I thought I'd share my thoughts.

💡 I'm wondering if we need a service for this task. 
- It's a simple one time request
- Stateful singletons are imo always a bit tricky to work with
- It seems the start() initiates the request, but when the request finishes the service doesn't do anything else, right? Until someone calls stop and start again?
- Do we need start/stop methods? Have you considered passing the scope from the caller?
To summarize: I'm a bit worried it might be an overkill, but I might be missing something crucial. Wdyt?

Update: This comment is related to [this discussion](https://github.com/wordpress-mobile/WordPress-Android/pull/13578/files#r540935052). Let's continue the discussion there and come back here when we decide on an approach.

malinajirka:
Thanks @ashiagr! As we discussed on the call you can ignore the above comment. I didn't realize the RewindStatusService uses this approach. I agree we should keep them consistent and refactor all of them later.

I tested all the cases described above and I noticed a possible issue.

When I 
1. select a site (plan with scan feature | admin)
2. Go to web in the browser and remove admin privileges to the user
3. Go back to the app and notice scan is still visible (expected)
4. Click on switch site
5. Select a different site
6. Click on switch site
8. Select site from step 1
9. Notice the scan is still visible even though the API response is "{"code":"authorization_required","message":"Can't query information about the current state.","data":{"status":401}}"


There is also a version conflict with `develop`.

ashiagr:
> I tested all the cases described above and I noticed a possible issue.

Great catch ! Fixed in 0603320. Also included unit test for it.
Since we don't update db on error, scan menu will be displayed briefly for scenarios like these (as it first displays last state in the db) before the request to get availability status fails. Is that ok?

malinajirka:
> Also included unit test for it.

❤️ 

> Since we don't update db on error, scan menu will be displayed briefly for scenarios like these (as it first displays last state in the db) before the request to get availability status fails. Is that ok?

I think it'd be fine if we showed the "Scan" item once after you switch to the site. However, I noticed I can see the animation every time I select the site. It's not a huge issue, but feels weird. Could we perhaps delete the item from the database on Authorization Error?

ashiagr:
>  Could we perhaps delete the item from the database on Authorization Error?

Yes I think we can do that. Added to backlog (#13326) as it will require fluxc changes: 
> Delete scan state from the database on Authorization Error

I'll link it here once it is done. 

Can we merge this one as we've other PRs ready to merged built upon this one?

malinajirka (approved):
Sounds good to me, thank you! ;)

Issue #11412, opened 2020-03-09
Stats aren't updating to the current day in the background after the date changes


designsimply:
> Since I updated to Android 10 last week [Stats - Daily view in] the app doesn't automatically go to the current day. It only goes as far as the last day I opened the app from scratch. I updated to the latest version of the app (Feb, 2020) but still have the same issue.

> You must close the app and reopen it in order for it to show the current day's site stats. Once you have reopened it, it does refresh the stats during the course of the day but will not automatically advance on the next day unless you close and reopen the app.

> If I drag my finger downwards to update it does not go to the current date. It only shows me the dates up to the yesterday's date.

* Reported in 2761834-zen (internal reference)
* WordPress - 14.2 - Version code: 831
* Android device name: Samsung SM-N960U
* Site Type: Jetpack Free Plan

Additional recommended troubleshooting steps:

1. Verify that your blog and your device are both set to the same timezone.
1. If possible, please get a copy of your app logs from Account > Application log > copy icon (top right) and then send them to us in the app using the Contact Us form in the Account > Help & Support section of the app _right after_ triggering the problem by doing a pull-to-refresh and seeing it fail for the Days Stats screen on a day you did not close and re-open the app since the previous day. It would be helpful to see if any relevant or helpful logs are generated **and** it needs to be just after triggering the problem and then copying the logs right afterward. 
1. Please try going to Android Settings > Apps & notifications > WordPress app > Storage & cache and tapping the "Clear cache" option. Does the problem persist the next day?

Note: sounds very similar to the iOS issue at https://github.com/wordpress-mobile/WordPress-iOS/issues/12823. 🤔

h/t @erricgunawan for the original report. Thank you!

rezzap:
The user has replied to confirm that clearing their cache options has fixed the issue. Shouldn't this be happening automatically on a re-install?

erricgunawan:
The user is back on 2828761-zen:

>  I upgraded to version 14.4.1 a few weeks ago but it did not help. I see an update posted does not address the issue. Clearing the cache only works temporarily. The issue returns a day or two later. 

I don't think they have sent a log yet, do we need to still ask it from them?

Is there anything else we can recommend, in the meantime?

designsimply:
> The user confirmed that clearing their device cache helped previously. Should this be happening automatically when reinstalling?

Yes, a complete uninstall and reinstall for an app should clear the cache for that specific app. 

[...]

PR #13556, opened 2020-12-07
Add tracking for stats inconsistency

planarvoid:
Helps with #11412 

With the previous attempts I've added logs to find out what's happening when the user gets inconsistent stats. I'm now suspicious of some API caching so I'm adding tracking for the cases when the app is showing old data to the user. The threshold is 2 days so let's see if we get any relevant information out of this. I've added a field for checking the age of visible stats so we can filter out users who haven't been online in a while. 

To test:
- Go to Stats/Days tab
- Check that the overview (first) block is showing you the correct dates and nothing is crashing there
- The tracked error should not be visible in the logs - filter out `stats_overview_error`

PR submission checklist:


designsimply:
Thank you for working on this! Apologies for the possibly obvious answer, but where exactly can I look up the stats for this? If I have a user reporting the problem in the future, what's the best way to figure out if they are included in this error logging?

renanferrari:
@designsimply I'm not sure if that's what you mean, but this is still going to be logged on Tracks under the `wpandroid_stats_overview_error` event, and if a user report this problem in the future and they're on version 16.4 or later, it's probably going to show up on their application log. Let me know if this actually answers your question or if I've misunderstood what you've asked.

renanferrari (approved):
Thanks for the changes! I've tested the app with the provided steps and also ran the unit tests that have been introduced, and everything seems to be working as expected. Code also looks good. 

LGTM :shipit:

designsimply:
That's exactly what I meant. Thank you!

PR #13561, opened 2020-12-08
Issue/13514 add p2 category to reader

### LINKED ISSUE #13514: Add P2 category to the reader
We should add a `Followed P2s` tab to the reader.

The WordPress.com Reader:

Screen Shot 2020-11-19 at 1 38 32 PM

[iOS issue.](https://github.com/wordpress-mobile/WordPress-iOS/issues/15343)

khaykov:
Fixes #13514

This PR adds a P2 category to the reader tab bar. We introduced all required logic in separate PR's so in this one we had to update the REST client version and fix an issue with site following.

P2 section will appear if you are following a P2 or a member of one.

If you are just following a P2 (meaning you don't have any special permission), the postcards from it will not have a special P2 header. This is a known issue, and we are working on addressing it by updating the endpoint and client-side logic.

If you follow an empty P2 (or one with default posts) the P2 category will appear, but it will not have any posts.

To get the following working I had to fix an issue that was preventing you from following the P2 site using an URL with `https` in it. Thanks to @mindgraffiti for helping me figure out how it's done on iOS 🙇 

To test P2 following:
- Make sure that you are not a member and don't follow any P2 before this test.
- Open reader settings and follow a public P2 by URL.
- Navigate back to Reader and make sure the P2 tab appears and contains content of the P2 you followed.
- Open reader settings and unfollow the P2 site.
- Navigate back to Reader and confirm that the P2 tab is gone.

Tag refresh in the Reader is performed periodically, so if you don't already have a P2 tab, and someone adds you as a member of their P2 you will have to wait until the tags are refreshed in the app. I decreased the current frequency of refresh requests from 1 hour to 10 minutes. With the current implementation of the Reader, I don't think we can do anything about it, unless we want to refresh tags on each pull-to-refresh. We would be able to make this problem less prominent by having an ability to accept P2 invite links in the app.

To test P2 category sync with remote:
- Make sure that you are not a member and don't follow any P2 before this test.
- Using Calypso invite yourself to a P2 in a role other than Viewer.
- Depending on when the tags were last refreshed, you might need to wait for up to 10 minutes (so just wait for 11 minutes before the next step). (for purposes outside of this test, you can open reader settings to force Tags update)
- Open Reader in the app and make sure the P2 tab is there.

If you are removed from the P2, you will still be following it unless the P2 owner removes you from the followers list manually.

To smoke test site following:
- Follow wpcom, atomic, jetpack sites by URL.
- Follow self-hosted site by URL.

PR submission checklist:


develric:
Hey Klym 👋 , thanks for this PR, it mostly look pretty good 👍 ; leaving here couple of things we discussed on Friday for tracking purposes:
- It seems I cannot follow feeds like [this](http://backr24.ilsole24ore.com/podcast/due-denari.xml) with the code from this PR. I'm able to do it in 16.3-rc-2 version.
- Re this `I decreased the current frequency of refresh requests from 1 hour to 10 minutes. With the current implementation of the Reader, I don't think we can do anything about it, unless we want to refresh tags on each pull-to-refresh.` . Probably can make sense to consider a refresh of tags in PTR as we do already in reader settings and in filter bottom sheet. Let's explore that direction 👍

khaykov:
@develric Thanks for the review and suggestions! I changed the way we follow sites to follow by feed ID when available, which solves the issue you mentioned. Also added a followed sites and tag refresh when refreshing posts.

I encountered some wonky behavior in the reader, but it also happens with `develop` branch. One of the issues is #13601

khaykov:
Removed changes to ReaderBlogActions, since it looks lie we can follow P2's normally again 🤷‍♂️

develric (approved):
Hey @khaykov 😄 , thanks for narrowing down and fixing especially the orphaned P2 posts issue (not trivial 👍 ). Made another test round and worked as described. LGTM :shipit:

PR #13565, opened 2020-12-08
Add accessibility test framework to project and tests

JavonDavis:
Introduces Accessibility test framework to our Connected/UI tests. 

To test: UI tests should still pass and logcat should contain the Accessibility suggestions. Filtering by "AccessibilityValidator" should make it easier to spot. 

PR submission checklist:

### DIFF
--- a/WordPress/build.gradle
+++ b/WordPress/build.gradle
@@ -363,6 +363,9 @@ dependencies {
 
     implementation "org.jsoup:jsoup:1.10.3"
     implementation 'androidx.emoji:emoji:1.0.0'
+
+    // Enables accessibility checks in Espresso
+    androidTestImplementation 'androidx.test.espresso:espresso-accessibility:3.3.0-alpha05'
 }
 
 configurations.all {
--- a/WordPress/src/androidTest/java/org/wordpress/android/e2e/BlockEditorTests.java
+++ b/WordPress/src/androidTest/java/org/wordpress/android/e2e/BlockEditorTests.java
@@ -2,6 +2,7 @@ package org.wordpress.android.e2e;
 
 import android.Manifest.permission;
 
+import androidx.test.espresso.accessibility.AccessibilityChecks;
 import androidx.test.rule.ActivityTestRule;
 import androidx.test.rule.GrantPermissionRule;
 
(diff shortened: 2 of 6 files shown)


WordPress/build.gradle
```diff
@@ -363,6 +363,9 @@ dependencies {
 
     implementation "org.jsoup:jsoup:1.10.3"
     implementation 'androidx.emoji:emoji:1.0.0'
+
+    // Enables accessibility checks in Espresso
+    androidTestImplementation 'androidx.test.espresso:espresso-accessibility:3.3.0-alpha05'
```
mokagio, line 368: Is there a particular reason to use `3.3.0-alpha05` instead of the latest stable?

I don't have anything against working with alphas. Just curious 😄

mokagio (approved):
Worked for me. I run it on a the login tests only (see below why): the tests passed and I got a few accessibility notes in Logcat

image

---

When I first launched the tests, they got stuck for 15m+, eventually failing.

Screen Shot 2020-12-10 at 3 42 32 pm

I tried the login subset and got the same failure. I then realized there was this save password prompt in my emulator:

Screen Shot 2020-12-10 at 3 51 55 pm

Once I dismissed that, the tests run fine. I doubt that's related with the change you implemented here, so ✅ .

The emulator I used was a Pixel 3 with Android R and Google Play. _Thinking about it, maybe that's why I got the error? I wonder if the behavior is the same on an emulator without Google Play_...

mokagio:
> The emulator I used was a Pixel 3 with Android R and Google Play. _Thinking about it, maybe that's why I got the error? I wonder if the behavior is the same on an emulator without Google Play_...

I tried on a Nexus 9 API 28 emulator without the Play Store and I didn't get that prompt. **TIL** 😄

JavonDavis:
> I tried on a Nexus 9 API 28 emulator without the Play Store and I didn't get that prompt. TIL 😄

@mokagio That's weird! That prompt actually has showed up for me but only after I recently created the emulator, doesn't usually show up anymore but there's also weird prompts like asking you to rate the app that will pop up and cause failures like this 😅 AFAIK this one of the reasons we enabled demo mode since I think it should disable some these 🤷🏽‍♂️

WordPress/build.gradle
JavonDavis: Nope! I think that was just the recommended version on their guide I was using

mokagio: 👌

mokagio:
> That prompt actually has showed up for me but only after I recently created the emulator

It is possible that I never run the app on that particular emulator, actually.

PR #13567, opened 2020-12-08
Issue/13329 activity more menu

zwarm:
Parent #13329 

This PR adds the more menu off of the activity log to support Restore and Creating the Downloadable Backup. This logic replaces the previous exposure methodology (dialog). 

Screenshot

**Merge instructions**
- Make sure ~~#13532~~ and #13545 have been merged
- Remove Not Ready For Merge label
- Merge as normal

**Notes**
- Text & icon color need to be finalized

**To test:**
- Launch the App
- Navigate to Activity Log
- Notice that the action button icon still shows the rewind icon
- Head over to App Settings -> test feature configuration
- Enable the BackupFeatureConfig and restart the app
- Notice that the action button is now the 3 dot vertical more icon
- Tap on the icon to show the menu options
- Tap the "Download backup" menu item
- Note the backup detail view is launched (The feature itself is not functional, this is just to test that the view shows)
- Go back to the Activity Log
- Tap on the icon to show the menu options
- Tap on the "Restore to this point" menu item
- Note the existing Rewind confirmation dialog is shown

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListActivity.kt
```diff
@@ -53,10 +52,6 @@ class ActivityLogListActivity : LocaleAwareActivity(),
     }
 
     override fun onNegativeClicked(instanceTag: String) {
```
ParaskP7, line 54: Minor (🔍 ): It is generally suggested to avoid method with empty body. When that is indeed unavoidable, the best practice is adding an explicit comment like `// Do nothing.`. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
```diff
@@ -230,10 +223,17 @@ class ActivityLogListFragment : Fragment() {
         viewModel.onActionButtonClicked(item)
     }
 
+    private fun onSecondaryActionClicked(
```
ParaskP7, line 226: 1) Idea (💡 ): Since this is going to be the only action later on, once the feature exclusively uses the more menu, what if this is renamed to `onMoreMenuClicked(...)` instead?
2) Idea (💡 ): If the above seems reasonable, then I would suggest renaming every occurrence of `secondary` with `moreMenu`, included in tests. For example, the `secondaryActionClickListener` can become the more explicit `moreMenuClickListener`.
3) Question (❓ ): Why does this click listener returns a `Boolean` value instead of `Unit`? Using `Unit` instead will simplify its usage everywhere. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListItem.kt
```diff
@@ -121,4 +123,8 @@ sealed class ActivityLogListItem(val type: ViewType) {
             fun fromValue(value: String?) = map[value] ?: DEFAULT
         }
     }
+
+    enum class SecondaryAction(val itemId: Long) {
+        RESTORE(0), DOWNLOAD_BACKUP(1)
```
ParaskP7: 1) Minor (🔍 ): Consider having the enum values in separate lines.
2) Minor (🔍 ): Consider adding the explicit `;` character to terminate the last enum. This way if more logic is added to this enum later on there would be no line change diffs there.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListItemMenuAdapter.kt
```diff
@@ -0,0 +1,78 @@
...
+    override fun getItemId(position: Int): Long {
+        return items[position].itemId
+    }
+
+    override fun getView(position: Int, convertView: View?, parent: ViewGroup): View {
+        var view = convertView
+        val holder: ActivityListItemMenuHolder
+        if (view == null) {
+            view = inflater.inflate(R.layout.activity_log_list_item_menu_item, parent, false)
+            holder = ActivityListItemMenuHolder(view!!)
```
ParaskP7: Warning (⚠️ ): This is not a warning per se, since it is guaranteed that the `!!` will never fail, but going forward and with more potential updates to it can create problems. Consider the following alternative:
```
val view: View
...
if (convertView == null) {
    view = inflater.inflate(R.layout.activity_log_list_item_menu_item, parent, false)
    ...
} else {
    view = convertView
    ...
}
```

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -182,15 +188,29 @@ class ActivityLogViewModel @Inject constructor(
...
+        }
+    }
+
+    fun onSecondaryActionClicked(
+        secondaryAction: ActivityLogListItem.SecondaryAction,
+        item: ActivityLogListItem
+    ): Boolean {
+        if (item is Event) {
+            val navigationEvent = when (secondaryAction) {
+                RESTORE -> {
```
ParaskP7, line 204: Minor (🔍 ): This and the below lines can become one-liners. Consider removing braces from the `when` entries.

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -182,15 +188,29 @@ class ActivityLogViewModel @Inject constructor(
...
+                    ShowRestore(item)
+                }
+                DOWNLOAD_BACKUP -> {
+                    ShowBackupDownload(item)
+                }
             }
+            _navigationEvents.value = org.wordpress.android.viewmodel.Event(navigationEvent)
```
ParaskP7, line 211: Minor (🔍 ): Since this `Event` import seems to be more generic, consider moving this import up and use `ActivityLogListItem.Event`everywhere instead. This way, the whole `org.wordpress.android.viewmodel.Event` will be avoided. Also, the usage of `ActivityLogListItem.Event` will not be confused with the generic view model `Event`. Wdyt?

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -288,7 +308,7 @@ class ActivityLogViewModel @Inject constructor(
 
     private fun requestEventsUpdate(isLoadingMore: Boolean) {
         if (canRequestEventsUpdate(isLoadingMore)) {
-            val newStatus = if (isLoadingMore) ActivityLogListStatus.LOADING_MORE else ActivityLogListStatus.FETCHING
+            val newStatus = if (isLoadingMore) LOADING_MORE else ActivityLogListStatus.FETCHING
```
ParaskP7, line 311: Minor (🔍 ): Consider using one alternative or the other, not both.

PS: For enums it is usually suggested to use `ActivityLogListStatus.LOADING_MORE` instead `LOADING_MORE` since this makes sure that imports collisions are avoided and might help the reader understand the import directly. If you agree, I suggest using the explicit `ActivityLogListStatus.XYZ` everywhere, which will also improve the imports. Wdyt?

WordPress/src/test/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModelTest.kt
```diff
@@ -380,6 +389,20 @@ class ActivityLogViewModelTest {
         assertEquals(selectedItems, viewModel.showActivityTypeFilterDialog.value!!.initialSelection)
     }
 
+    @Test
+    fun onSecondaryActionClickRestoreNavigationEventIsShowRestore() {
+        viewModel.onSecondaryActionClicked(RESTORE, event)
+
+        Assertions.assertThat(navigationEvents.last().peekContent()).isInstanceOf(ShowRestore::class.java)
```
ParaskP7, line 396: Minor (🔍 ): Consider moving `Assertions` to imports, just like it is done for `Assert` for the `org.junit.Assert` import.

ParaskP7 (commented):
Great work @zwarm and and again thanks for including me in this PR! 🌟 

I have tested the solution and it works as expected.

Only minor comments (🔍 ), ideas (💡 ) for improvement, questions (❓ ) and one minor language warning (⚠️ ) from my side.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
zwarm: > 1. Idea (💡 ): Since this is going to be the only action later on, once the feature exclusively uses the more menu, what if this is renamed to `onMoreMenuClicked(...)` instead?

This is for a menu item click not the menu click.

> 2. Idea (💡 ): If the above seems reasonable, then I would suggest renaming every occurrence of `secondary` with `moreMenu`, included in tests. For example, the `secondaryActionClickListener` can become the more explicit `moreMenuClickListener`.

See above

> 3. Question (❓ ): Why does this click listener returns a `Boolean` value instead of `Unit`? Using `Unit` instead will simplify its usage everywhere. Wdyt?
Because the construct for menu item click requires a Boolean

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
zwarm: @ParaskP7 - I was dreaming about this last night. I am so changing this. Updated in 4b8cab1

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
zwarm: I'm going to leave this as is because it's not part of this PR, but a result of a merge. There may be more coming.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListItemMenuAdapter.kt
zwarm: Good call out - in commit 2054d2b

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
ParaskP7: Such sweet dreams you have Annmarie! 🤣

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
ParaskP7: @zwarm Aaa ok then, now I understand how you thought about it, I was thinking it from another angle!

What about then renaming that to `onMenuItemClicked(...)` and a `menuItemClickLIstener` instead?

zwarm: @ParaskP7 - Staying consistent with other places in the app that call menu clicks `SecondaryActions`. :) This sounds like a good conversation to have with the rest of the Raven folks and pick a direction. Thanks for the thoughts.

ParaskP7: Thanks for letting me know of this Annmarie, didn't know that this is a convention already! So, definitely, staying consistent is the way to go here. 🌟

WordPress/src/main/res/layout/activity_log_list_item_menu_item.xml
```diff
@@ -0,0 +1,28 @@
...
+        android:id="@+id/image"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:contentDescription="@null"
+        android:tintMode="multiply"
+        android:layout_marginEnd="@dimen/margin_large"
+        tools:src="@drawable/ic_history_white_24dp"
+        tools:tint="@color/primary" />
+
+    <org.wordpress.android.widgets.WPTextView
```
ashiagr: We might want to use material text view in the new layout.

ashiagr (approved):
Thank you both of you! 

@zwarm I tested the described flow and everything worked as expected. 
Just left a minor comment. Feel free to consider it in the upcoming PRs and merge this one.

WordPress/src/main/res/layout/activity_log_list_item_menu_item.xml
zwarm: @ashiagr - thanks and yes, I'll add that to my task list. Actually it's changed in cd18c19

PR #13574, opened 2020-12-10
Issue/13573 add organization id to reader

### LINKED ISSUE #13573: Introduce organization_id in Reader for p2 and a8c content.
Instead of relying on `is_wp_forteams` flag, we should utilize `organization_id` to differentiate between different types of content.

khaykov:
Fixes #13573

This PR introduces the concept of Organization instead of "wp for teams" flag.
I replaced all the instances of flag usage with Organization check in the Reader.

I left the migration script that creates the `is_wp_for_teams` and `is_wpforteams_site` flags since I'm not sure if it's ok to change those?

To test:
- Test DB migration by installing this branch over develop.
- Check a8c tab in the reader and confirm that the special p2 header is visible at postcards from supported sites.

PS: Note that since it's a new field in DB table, you might need to refresh posts to populate it.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/datasets/ReaderDatabase.java
```diff
@@ -218,6 +219,10 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
             case 144:
                 db.execSQL("ALTER TABLE tbl_blog_info ADD is_wp_for_teams BOOLEAN;");
                 currentVersion++;
+            case 145:
+                db.execSQL("ALTER TABLE tbl_blog_info ADD organization_id INTEGER;");
+                db.execSQL("ALTER TABLE tbl_posts ADD organization_id INTEGER;");
+                currentVersion++;
         }
```
develric: wonder if here we should remove the `tbl_posts.is_wpforteams_site` and `tbl_blog_info. is_wp_for_teams `. wdyt?

WordPress/src/main/java/org/wordpress/android/ui/Organization.kt
```diff
@@ -0,0 +1,15 @@
...
+    UNKNOWN(-1),
+    NO_ORGANIZATION(0),
+    A8C(1),
+    P2(2);
+
+    companion object {
+        private val map = values().associateBy(Organization::orgId)
+
+        @JvmStatic
+        fun fromOrgId(type: Int) = map[type] ?: UNKNOWN
```
develric: np: I would consider to use `orgId` instead of `type`; wdyt?

develric (commented):
Hey @khaykov 👋 , looked into this one. Code looks good and it works well 👍 ! Also the db upgrade script is working fine. I left a couple of notes for your review. Let me know wdyt 🙇 .

WordPress/src/main/java/org/wordpress/android/datasets/ReaderDatabase.java
khaykov: I'm not sure if we can remove the columns in SQL Lite? At least without copying the content of the table into them one, dropping/recreating the original table, and copying content back. But maybe I'm missing something here :)

develric (approved):
Thanks for the changes and follow up @khaykov 🙇 . Agree on at least removing the creation of deprecated columns where possible in the upgrade path. LGTM :shipit:

PR #13577, opened 2020-12-10
Experiment with Gutenberg from Bintray

mokagio:
This PR is not meant for merging. Once the setup works, we'll create a clean new one.

Expect lots of notes to self in the comments. I'm also trying to document every experiment and failure in the commit messages.


mokagio:
Lint aside, this builds on CI and runs on device.

It's far from done, though. I went in the Gutenber submodule and checked out an old tag as a test see if the build worked and got a null pointer exception when running `./gradlew assembleDebug`.

Clearly, there are still settings pointing to the files in the repo rather than the AAR.

An open question I have is whether the Azetec library that WordPressEditor refers to comes from the fat AAR or somewhere else.

mokagio:
Keen for your feedback @mchowning 😄 🙏

libs/editor/WordPressEditor/build.gradle
```diff
@@ -79,10 +81,13 @@ dependencies {
     api ("com.github.wordpress-mobile.WordPress-Aztec-Android:wordpress-shortcodes:$aztecVersion")
     api ("com.github.wordpress-mobile.WordPress-Aztec-Android:wordpress-comments:$aztecVersion")
 
-    implementation (project(':@wordpress_react-native-bridge')) {
-        exclude group: "com.android.volley"
-        exclude group: 'com.google.code.findbugs'
-    }
+    // The version value here is the same as the one in the repo's root
+    // build.gradle
+    implementation "com.github.wordpress-mobile.gutenberg-mobile:react-native-gutenberg-bridge:20201210-144401"
```
mchowning: Can probably use `rootProject.ext.gutenbergMobileVersion` here for the version number.

settings.gradle
```diff
@@ -89,5 +90,17 @@ if (localBuilds.exists()) {
             }
         }
     }
+
+    // Not sure what to use in place of :WordPressUtils here. Leaving it for
+    // now.
```
mchowning: I agree that ignoring this for now makes sense. WPAndroid already has a working capability to build using gb-mobile source by using the `wp.BUILD_GUTENBERG_FROM_SOURCE` flag, so we can focus on getting the binary build working (without breaking the current build-from-source mode). If you get that working you're going to have a lot of new Android BFFs!

After that, we could work on updating our build-from-source to use the new (and better) approach in a separate follow-up PR. With that said, if you want to go ahead and knock them both out, don't let me stop you! 😆

mchowning:
> An open question I have is whether the Azetec library that WordPressEditor refers to comes from the fat AAR or somewhere else.

This is a little tricky since it is a dependency of both gutenberg-mobile and the WordPressEditor. If it could be included in the fat aar and have WordPressEditor "depend" on it via gutenberg-mobile, it would be an easy way to ensure the versions stay in sync. That might require changing some dependencies from `compile` or `implementation` to `api` though.

mchowning:
I only had time to do a quick test, but using c397b9e from this PR and 39bb2901be from [the gutenberg PR](https://github.com/WordPress/gutenberg/pull/27836), both the from-source and from-binary builds worked for me! 🎉  This is some great work @mokagio !

mokagio:
This was an experimental branch where to mess around with the setup but I needed a PR to share the experiments and run the tests in CI. This is done now. Closing.

PR #13578, opened 2020-12-10
Jetpack Mobile: Scan Introduce UiState

ashiagr:
Parent #13326

This is a small preparatory PR with a scope to just link things (Service, VM, Fragment, ViewHolder) together for the Scan screen UI. It 
- introduces recycler view + adapter to the `ScanFragment`
- introduces Content UiState to the `ScanViewModel` 
- populates basic elements (scan state icon, title) on the screen using the state received from the scan service. 

I plan to use (in the upcoming PRs) different view types - scan state (with icon, title, description, action buttons), progress, threat header, threat items e.t.c. inside a RecyclerView so that entire screen is scrollable .

**To test**

Prerequisite: Make sure both Scan feature flag and MySiteImprovements flag are set on in the App settings.

- Select a site with scan capability on the My Site tab
- Notice that scan menu is displayed
- Click scan menu item 
- Notice that basic elements (icon, title) are displayed on the Scan screen from the service. 

device-2020-12-10-170621

**Note**
- Used placeholder icons, texts. Ignore all stying.
- `ScanListItemState` has to be fine tuned. I'll pickup and fine tune different view types (with their states) in the upcoming PRs.
- I'll create a separate FluxC PR for storing/ retrieving threats into/ from db.

**Merge Instructions**

1. Make sure #13555 is merged
3. Merge this PR to develop

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
```diff
@@ -3,26 +3,68 @@ package org.wordpress.android.ui.jetpack.scan
...
-    }
-
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
-        initViewModel()
+
+        val nonNullActivity = requireActivity()
+        (nonNullActivity.application as WordPress).component()?.inject(this)
+
+        val site = if (savedInstanceState == null) {
```
ashiagr: I think we can update FluxC code to just use `siteId`. We can eliminate this passing and retrieval of the entire site model. Lmkwyt.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanStatusService.kt
```diff
@@ -31,6 +31,10 @@ class ScanStatusService
     private val _scanAvailable = MutableLiveData<Boolean>()
     val scanAvailable: LiveData<Boolean> = _scanAvailable
 
+    private val _scanState = MutableLiveData<ScanStateModel>()
```
ashiagr, line 34: Scan state just needs to observed from the VM. I'm using `observeForever`. I think we can use Flow   here. Lmk, I'll make the switch.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
ParaskP7: 👋 @ashiagr !

I think you are right and I agree. Since only `siteId` is the only property needed and considering that no other property from the entire site model will be used in the future, I believe the best course of action is to replace `site` with `siteId`.

IMHO, the main gain from doing something like that is to make it explicit to the reader that this is the only thing needed, so they know straight away and can score the feature down a bit. Obviously, there is a performance optimization as well, but to be honest this will not save us much.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanStatusService.kt
ParaskP7: 👋 @ashiagr !

I am not sure if using `Flow` here is going to help. Since, this is a one off trigger, `scanStore.fetchScanState(...)` and `scanStore.getScanStateForSite(...)`, just like most of the api calls within our view models. It wouldn't benefit if we where to use `Flow`, would it? Am I missing something here? It is just because there is an extra layer that needs to be observed, and passed through to another observer, that you are thinking of using `Flow`, right?

What I am also seeing, is the usage of `scanStatusService.start(...)` and `scanStatusService.stop()`, which might create an intention related confusion. Without me reading through `ScanStatusService`, I though that this is indeed a `start` function that needs to be `stopped` at some point and unless it is stopped, start will continue working, producing results, but this is not the case, is it? This is actually a `trigger` function instead, and `start/stop` are being used for cleaning things up, not stopping the started service? Maybe `start/stop` should be renamed. Maybe even the `Service` naming convention is not the right one. It seems to me more like a `UseCase`. Wdyt?

Having said all of the above, I might be missing something very basic here... not sure!

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/adapters/ScanAdapter.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    class ScanDiffCallback(
+        private val oldList: List<ScanListItemState>,
+        private val newList: List<ScanListItemState>
+    ) : DiffUtil.Callback() {
+        override fun areItemsTheSame(oldItemPosition: Int, newItemPosition: Int): Boolean {
+            val oldItem = oldList[oldItemPosition]
+            val newItem = newList[newItemPosition]
+            return when {
+                oldItem is ScanState && newItem is ScanState -> // TODO: ashiagr revisit
```
ParaskP7, line 57: Question (❓ ): What would you revisit here?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
```diff
@@ -3,26 +3,68 @@ package org.wordpress.android.ui.jetpack.scan
...
 
-    override fun onCreate(savedInstanceState: Bundle?) {
-        super.onCreate(savedInstanceState)
-        (requireActivity().application as WordPress).component().inject(this)
-    }
-
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
-        initViewModel()
+
```
ParaskP7: Idea (💡 ): This is just me proposing one of those readability things that I like advocating in favor (😊 ), but this is very opinionated, so please consider it as such. What if the `onViewCreated(...)` function looked like that, do you find the below more readable?

```
        initDagger()
        initAdapter()
        initViewModel(getSite(savedInstanceState))
    }
```

`initDagger()` can be extracted in its own function using the `requireActivity()` directly. Then, the site state related code can be extracted to another private function as well and used directly through `getSite(...)`. Here again, `requireActivity()` can be used directly to avoid creating an extra variable. Wdyt?

WordPress/src/main/res/layout/scan_fragment.xml
```diff
@@ -1,5 +1,20 @@
...
     android:layout_width="match_parent"
-    android:layout_height="match_parent" />
+    android:layout_height="match_parent"
+    android:padding="@dimen/margin_extra_large">
+
+    <org.wordpress.android.ui.prefs.EmptyViewRecyclerView
+        android:id="@+id/recycler_view"
+        android:layout_width="0dp"
+        android:layout_height="0dp"
+        app:layoutManager="androidx.recyclerview.widget.LinearLayoutManager"
```
ParaskP7, line 13: Question (❓ ): I think I asked @malinajirka about this and I got the understanding that in this codebase the preferred way is to do `layoutManager` programmatically instead. Please let me know your thoughts?

PS: I only ask these questions, actually all my questions so far, because when I start contributing as well, I want to be sure I using the right and agreed approach. Although, I too believe that having it here might be better. 😊

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
```diff
@@ -3,26 +3,68 @@ package org.wordpress.android.ui.jetpack.scan
...
+    private fun initViewModel(site: SiteModel) {
         viewModel = ViewModelProvider(this, viewModelFactory).get(ScanViewModel::class.java)
+        setupObservers()
+        viewModel.start(site)
+    }
+
+    private fun setupObservers() {
+        viewModel.uiState.observe(
+            viewLifecycleOwner,
+            Observer { uiState ->
```
ParaskP7, line 48: Question (❓ ): Does your IDE also suggests that this redundant SAM-constructor can be removed?
Minor (🔍 ): In the end I can make this code look like the below:
```
        viewModel.uiState.observe(viewLifecycleOwner, { uiState ->
            if (uiState is Content) refreshContentScreen(uiState)
        })
```

PS: Usually, in Kotlin, I tend to keep simple, not complex, if statements, a one-liner. Thus, the suggestion above. If I can recall correctly this is also something that is suggested in the official Android Kotlin style guide.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanListItemState.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.wordpress.android.ui.utils.UiString.UiStringRes
+
+sealed class ScanListItemState(val type: ViewType) {
+    open fun longId(): Long = hashCode().toLong()
+
+    // TODO: ashiagr fine-tune states, dynamic texts, add button states
+    sealed class ScanState : ScanListItemState(SCAN_STATE) {
+        @DrawableRes open val scanIcon: Int = R.drawable.ic_scan_scanning
+        open val scanTitle: UiString = UiStringRes(R.string.scan_scanning_title)
+        open val scanDescription: UiString = UiStringRes(R.string.scan_scanning_description)
```
ParaskP7: Question (❓ ): I am not sure if those parent fields will be used somehow in the future, if not, what if this is changed to being `abstract` instead?
```
        abstract val scanIcon: Int
        abstract val scanTitle: UiString
        abstract val scanDescription: UiString
```

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanListItemState.kt
```diff
@@ -0,0 +1,42 @@
+package org.wordpress.android.ui.jetpack.scan
+
+import androidx.annotation.DrawableRes
+import org.wordpress.android.R
+import org.wordpress.android.ui.jetpack.scan.ScanListItemState.ViewType.SCAN_STATE
+import org.wordpress.android.ui.utils.UiString
+import org.wordpress.android.ui.utils.UiString.UiStringRes
+
+sealed class ScanListItemState(val type: ViewType) {
+    open fun longId(): Long = hashCode().toLong()
```
ParaskP7: Question (❓ ): Is the `open` modifier required here?
Minor (🔍 ): Also, just as an FYI, the explicit `Long` type specification can be removed here.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanStatusService.kt
```diff
@@ -72,8 +76,9 @@ class ScanStatusService
         return false
     }
 
-    private fun updateScanState(scanState: ScanStateModel?) {
-        _scanAvailable.value = scanState?.state != UNAVAILABLE && scanState?.state != UNKNOWN
+    private fun updateScanState(scanStateModel: ScanStateModel) {
+        _scanAvailable.value = scanStateModel.state != UNAVAILABLE && scanStateModel.state != UNKNOWN
+        _scanState.value = scanStateModel
```
ParaskP7: Question (❓ ): While reviewing the code, and since I haven't seen this class before, trying to answer the `Flow` question that you posted, I have an extra question about the `reloadScanState`. So, when the `start(...)` function get triggered, wouldn't `reloadScanState()` be triggered twice? Once during `requestScanStateUpdate()` and once directly? Why do we want that?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanStatusService.kt
```diff
@@ -72,8 +76,9 @@ class ScanStatusService
...
     }
 
     private fun onScanStateFetched(scanStateError: ScanStateError?, isError: Boolean) {
```
ParaskP7: Minor (🔍 ): `isError` function parameter seems to be unused, is that needed for future?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
```diff
@@ -1,15 +1,65 @@
...
+import org.wordpress.android.ui.jetpack.scan.ScanListItemState.ScanState.ScanScanningState
+import org.wordpress.android.ui.jetpack.scan.ScanViewModel.UiState.Content
 import javax.inject.Inject
 
-class ScanViewModel @Inject constructor() : ViewModel() {
+class ScanViewModel @Inject constructor(
+    private val scanStatusService: ScanStatusService
+) : ViewModel() {
     private var isStarted = false
+    private var lastScanState: State? = null
```
ParaskP7: Idea (💡 ): It initially confused me a bit, thinking that the `State` here is the `UiState`. I am still not used to that I guess. Since those two are really close, readability-wise, what if the full qualifier is used instead, like `ScanStateModel.State`?

PS: The same applies everywhere within this view model. I understand that this is more verbose and could make the `when (model.state) { ... }` look funnier, but I have a suggestion there as well. 😊

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
```diff
@@ -1,15 +1,65 @@
...
+            _uiState.value = buildContentUiState(it)
+        }
+    }
+
+    lateinit var site: SiteModel
+
+    fun start(site: SiteModel) {
         if (isStarted) {
             return
         }
```
ParaskP7, line 31: Minor (🔍 ): Consider keeping the `if (isStarted)` close to the `isStated = true` logic. Otherwise this can have the potential to become a problem in the future.
Minor (🔍 ): Consider have the `if (isStarted) return` a one-liner. But, as I said, this is just what I would do.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
```diff
@@ -1,15 +1,65 @@
...
+    private fun init() {
+        scanStatusService.scanState.observeForever(scanStateObserver)
+        scanStatusService.start(site)
+    }
+
+    private fun buildContentUiState(model: ScanStateModel): Content {
+        val scanStateItem = when (model.state) {
+            State.IDLE -> model.threats?.let { ScanIdleState.ThreatsFound() } ?: ScanIdleState.ThreatsNotFound()
+            State.SCANNING -> ScanScanningState()
+            State.PROVISIONING, State.UNAVAILABLE, State.UNKNOWN -> ScanScanningState() // TODO: ashiagr filter out
```
ParaskP7: Idea (💡 ): I usually suggest every `when` case to stand by itself, even if the action is duplicated. I believe this improves the readability and makes sure that creepy bugs are avoided. Consider the below instead, but again, this is very subjective:
```
        val scanStateItem = when (model.state) {
            ScanStateModel.State.IDLE -> {
                model.threats?.let { ScanIdleState.ThreatsFound() } ?: ScanIdleState.ThreatsNotFound()
            }
            ScanStateModel.State.SCANNING -> ScanScanningState()
            ScanStateModel.State.PROVISIONING -> ScanScanningState()
            ScanStateModel.State.UNAVAILABLE -> ScanScanningState()
            ScanStateModel.State.UNKNOWN -> ScanScanningState()
        }
```

PS: I can see that `ScanScanningState()` is used on all four cases, but you split it in two, why is that?

WordPress/src/main/res/layout/scan_list_scan_state_item.xml
```diff
@@ -0,0 +1,39 @@
...
+    <com.google.android.material.textview.MaterialTextView
+        android:id="@+id/scan_state_title"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="@dimen/margin_extra_large"
+        android:textStyle="bold"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@id/scan_state_icon"
+        tools:text="Don't worry about a thing" />
```
ParaskP7: Question (❓ ): I am not sure what is the best or agreed approach here, but what if the actual string res was used for both `tools:text` attributes instead of a hard-coded string that can end up being out-dated?

```
tools:text="@string/scan_idle_no_threats_found_title"
...
tools:text="@string/scan_idle_no_threats_found_description"
```

ParaskP7 (commented):
Good job and thanks for including me in this PR @ashiagr ! 🌟 

I have few comments, but very minor ones and most of them are kind of question related.

PS: The testing part will be done on a separate commit or on another PR?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanStatusService.kt
ashiagr: Thank you for these thoughts 🙏.

> It is just because there is an extra layer that needs to be observed, and passed through to another observer, that you are thinking of using Flow, right?

This is right, also `observeForever` is not lifecycle aware and we need to remove observer explicitly in the VM's `onCleared` but I agree to use existing conventions. 

> What I am also seeing, is the usage of scanStatusService.start(...) and scanStatusService.stop(), which might create an intention related confusion. 

I had same thoughts but I don't want to deviate too much from the `RewindStatusService` logic which is used for similar purpose and has similar `start` and `stop` methods. We might want to discuss with @malinajirka, @zwarm about the naming convention.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
ashiagr: I wanted a second opinion if we should be doing it, thanks for sharing your thoughts 🙏. I'll not make any changes as of now, may be just add "nice to have" in the backlog.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
ashiagr: > Does your IDE also suggests that this redundant SAM-constructor can be removed?

My IDE highlights it as well. It is for [this reason](https://github.com/wordpress-mobile/WordPress-Android/pull/13469#discussion_r530843257) mentioned in another PR review that I included it:
> we're not removing them yet in the code for compatibility with older AS versions, right?

----
> I tend to keep simple, not complex, if statements, a one-liner.

I just followed existing code convention in the VMs which use braces around `if`:

```
if (isStarted) {
    return
}
```

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanStatusService.kt
ashiagr: I think this is intended. We want to load the screen with the last state and then reload it again after the fetch is complete.  Do you see any downside?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
ashiagr: > Consider keeping the if (isStarted) close to the isStated = true logic. Otherwise this can have the potential to become a problem in the future.

👍 in deb7a70

> Consider have the if (isStarted) return a one-liner. But, as I said, this is just what I would do.

Followed existing code convention with braces around if:

```
if (isStarted) {
    return
}
```

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
ashiagr: Full qualifier added in 0149797

> I can see that ScanScanningState() is used on all four cases, but you split it in two, why is that?

For `SCANNING` I was sure we can use `ScanScanningState`(). I clubbed the rest as I was undecided if we should use that state (ideally should not as those states don't support scanning).

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
ashiagr: In 01129a8

WordPress/src/main/res/layout/scan_fragment.xml
ashiagr: No preference as such, it could be set directly in the xml so I did that. If it had been a custom layout manager then of course it would need to be added in the code.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanStatusService.kt
ashiagr: Yep, it will be needed in the future.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanListItemState.kt
ashiagr: In a576c37

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanListItemState.kt
ashiagr: > Is the open modifier required here?

Not actually 

> the explicit Long type specification can be removed here.

Removed both in a576c37

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
ashiagr: Nice suggestion 👍 - full qualifier added in 0149797

WordPress/src/main/res/layout/scan_list_scan_state_item.xml
ashiagr: In f680209

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/adapters/ScanAdapter.kt
ashiagr: The following line that compares state title and description to find if state view items are the same. There might be more states in the future which may have same title and description.

ashiagr:
> I have few comments, but very minor ones and most of them are kind of question related.

Thanks so much for a thorough review @ParaskP7! 🌟. I've addressed your comments and answered questions. Let me know if any thing is not clear.

> PS: The testing part will be done on a separate commit or on another PR?

They will be covered in a separate PR once UiStates are fine-tuned.

Ready for another round 🙇‍♀️.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
ParaskP7: Yeap, definitely! But, on the other hand, it might be decided and for a good reason to keep using whole models instead, confronting my argument above, so I keep my mind open, as it indeed can help not having to make two PRs every time you want to add an extra property from the model or the whole model next time you needs more from it. 🤔

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanStatusService.kt
ParaskP7: I agree, there is a candidate for a small discussion here! 🤓

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/adapters/ScanAdapter.kt
ParaskP7: Got it, thanks for sharing! 🌟

WordPress/src/main/res/layout/scan_fragment.xml
ParaskP7: Definitely, this is my exact thinking as well! 🌟

It just that I had a similar discussion with Jirka (if I remember correctly) but the other around, thus I pinged him. 😊

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
ParaskP7: > My IDE highlights it as well. It is for this reason mentioned in another PR review that I included it:

Interesting, thanks for sharing Ashita! I have left the below comment there and quoting it here.

> Hmmm... Is this due to compatibility reason with AS version or Kotlin versions? I am thinking the later, right? And if that is the case, I suggest having every develop be forced to update the AS Kotlin plugin to at least the Gradle's Kotlin version, currently 1.4.10, as specified by our build files. This will make sure everyone is on the same boat. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanStatusService.kt
ParaskP7: Ah ok, that makes sense! 🙏 No, no major downside I can think of, I would have to think about it a bit more to find an alternative (if any).

ParaskP7:
> Thanks so much for a thorough review @ParaskP7! 🌟. I've addressed your comments and answered questions. Let me know if any thing is not clear.

Thank YOU Ashita for addressing all of them and so gracefully!

> They will be covered in a separate PR once UiStates are fine-tuned.

Perfect, I was sure of that, just trying to make sure I am not missing something.

> Ready for another round 🙇‍♀️.

All good from my side! I just kept one comment unresolved just to trigger a short discussion for anyone interested! 😊

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
```diff
@@ -3,26 +3,70 @@ package org.wordpress.android.ui.jetpack.scan
...
+    }
+
+    private fun initDagger() {
+        (requireActivity().application as WordPress).component()?.inject(this)
     }
 
-    private fun initViewModel() {
+    private fun initAdapter() {
+        val scanAdapter = ScanAdapter(uiHelpers)
+        scanAdapter.setHasStableIds(true)
```
malinajirka: Nitpick 🔍 : We usually set stable ids in the adapter (in the constructor or the init section).

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
```diff
@@ -3,26 +3,70 @@ package org.wordpress.android.ui.jetpack.scan
...
+                }
+            }
+        )
+    }
+
+    private fun refreshContentScreen(content: Content) {
+        ((recycler_view.adapter) as ScanAdapter).update(content.items)
+    }
+
+    private fun getSite(savedInstanceState: Bundle?): SiteModel {
```
malinajirka, line 60: ❓ Have you considered using a field instead of this method? It feels that the method isn't re-usable. Since if I want to invoke it eg. from "observe(..)" I don't have access to the "savedInstanceState" anymore. Wdyt?

Related note: I noticed we recently introduced `SelectedSiteRepository`. I'm not sure if it's intended even for this type of usecase, but I think it might be. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanStatusService.kt
malinajirka: Funny, I left a very [related comment](https://github.com/wordpress-mobile/WordPress-Android/pull/13555#issuecomment-745199040) on the previous PR. We had a call with Ashita and decided to stay consistent with the RewindStatusService for now. We'll refactor all Rewind/Backup/ScanStatusServices when they are all finished. I agree that the current state isconfusing, to say the least. Even renaming some of the methods/fields might help quite a bit - but we could improve the classes ever more.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
```diff
@@ -1,15 +1,67 @@
...
+import org.wordpress.android.ui.jetpack.scan.ScanListItemState.ScanState.ScanScanningState
+import org.wordpress.android.ui.jetpack.scan.ScanViewModel.UiState.Content
 import javax.inject.Inject
 
-class ScanViewModel @Inject constructor() : ViewModel() {
+class ScanViewModel @Inject constructor(
+    private val scanStatusService: ScanStatusService
+) : ViewModel() {
     private var isStarted = false
+    private var lastScanState: ScanStateModel.State? = null
```
malinajirka: 💡 Idea / 🔍 Nitpick: I'm wondering if it'd be cleaner to always access `scanStatusService.scanState` instead of introducing a variable. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/adapters/viewholders/ScanStateViewHolder.kt
```diff
@@ -0,0 +1,23 @@
...
+import org.wordpress.android.ui.utils.UiHelpers
+
+class ScanStateViewHolder(
+    private val uiHelpers: UiHelpers,
+    parent: ViewGroup
+) : ScanViewHolder(R.layout.scan_list_scan_state_item, parent) {
+    override fun onBind(itemUiState: ScanListItemState) {
+        val scanState = itemUiState as ScanState
+
+        scan_state_icon.setImageResource(scanState.scanIcon)
```
malinajirka: 💡 Idea: We might want to consider using an `ImageManager.load(imageView: ImageView, @DrawableRes resourceId: Int, scaleType: ScaleType = CENTER)` here. I think it'd be great if we used it for all image loading. Wdyt?

WordPress/src/main/res/drawable/ic_scan_idle_threats_found.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
malinajirka, line 1: ⚠️ We usually name the resources `ic_[NAME]_[WIDTH]dp.xml`. I think it'd be good to update the names to keep the consistency. Thanks!

WordPress/src/main/res/drawable/ic_scan_idle_threats_found.xml
```diff
@@ -0,0 +1,9 @@
...
+    android:width="138dp"
+    android:height="162dp"
+    android:viewportWidth="138"
+    android:viewportHeight="162">
+  <path
+      android:pathData="M69,8L12,33.046V70.614C12,105.364 36.32,137.861 69,145.75C101.68,137.861 126,105.364 126,70.614V33.046L69,8ZM62.667,45.568H75.333V83.136H62.667V45.568ZM62.667,95.659H75.333V108.182H62.667"
+      android:fillColor="#DC3232"/>
```
malinajirka, line 8: ⚠️  : We might want to use a resources from colors.xml. If it doesn't exist there, we might want ask Chris if there is any similar color or if we should really introduce a new color.

WordPress/src/main/res/layout/scan_fragment.xml
malinajirka: 👍 yeah, I usually set the manager in code and I think that's what most devs are doing. Having said that it's such a minor thing that I guess it doesn't matter and we can leave it on personal preference.

WordPress/src/main/res/values/strings.xml
```diff
@@ -1040,6 +1040,15 @@
     <!-- scan -->
     <string name="scan">Scan</string>
 
+    <string name="scan_state_icon">Scan state icon</string>
+    <string name="scan_scanning_title">Preparing to scan</string>
+    <string name="scan_idle_no_threats_found_title">Don\'t worry about a thing</string>
+    <string name="scan_idle_threats_found_title">Your site may be at risk</string>
+    <string name="scan_scanning_description">Welcome to Jetpack Scan, we are taking a first look at your site now and the result will be with you soon.
+        \n We will send you a notification once the scan is complete. In the meantime feel free to continue to use your site as normal. you can check the progress at anytime.</string>
+    <string name="scan_idle_no_threats_found_description">The last Jetpack scan ran 7 hours ago and everything looked great. Run a manual scan now or wait for Jetpack to scan your site later today.</string>
```
malinajirka, line 1053: ⚠️ Should we replace "7 hours" with a parameter?

malinajirka (commented):
Thank you @ashiagr, great job! ;) I've left some comments. Let me know what you think.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanStatusService.kt
ParaskP7: Thank you, both! 🌟

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
ParaskP7: Seems reasonable to me.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/adapters/viewholders/ScanStateViewHolder.kt
ParaskP7: Didn't know that this existed, thanks for sharing Jirka! Btw, since this is Kotlin, why not using this extension method pattern for loading images. Has this been discussed or implemented here yet?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
ashiagr: It came by referring to a similar check in the `ActivityLogModel` [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt#L125) showing  a rewind progress indicator. It might have become useful when we would've started showing a scan progress indicator. However, I'm removing it for now and re-add if needed.

In ca0c82a

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/adapters/viewholders/ScanStateViewHolder.kt
ashiagr: Noticed `imageView.setImageResource(activity.icon.drawable)` in [EventItemViewHolder](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/EventItemViewHolder.kt#L55
) and it seemed like icon drawables were directly set using `setImageResource`. But I liked your idea and changed to it in 6522dff.

ashiagr:

For this one:

⚠️ Should we replace "7 hours" with a parameter?

I've already covered some of the dynamic texts in the next PR https://github.com/wordpress-mobile/WordPress-Android/pull/13605. I mentioned in the PR description that I've used placeholder icons and texts but thanks  so much for noticing it 🌟.

Regarding below warnings, Android scan screen specs are not yet ready. I can add temporary dp width for now. Ideally I'd wait for the first set of specs to be shared with us before I start making these refinements. Lmkwyt.

⚠️ We usually name the resources ic_[NAME]_[WIDTH]dp.xml. I think it'd be good to update the names to keep the consistency. Thanks!

⚠️ : We might want to use a resources from colors.xml. If it doesn't exist there, we might want ask Chris if there is any similar color or if we should really introduce a new color.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
ashiagr: In a6a7faf

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
ashiagr: I haven't quite understood this comment. Let's discuss it tomm. when you're online.

malinajirka: This was not an important comment so I'll just close this convo to unblock this PR.

The first part `Have you considered using a field instead of this method? It feels that the method isn't re-usable. Since if I want to invoke it eg. from "observe(..)" I don't have access to the "savedInstanceState" anymore. Wdyt?` The current usage of the method is completely fine and I don't see any issues with it. However, if you imagine that we'd like to invoke `getSite(..)` from a different place than `onViewCreated()` we won't have access to `savedInstanceState`.  So we might want to consider introducing "site" field rather than using this method.

malinajirka:
Thanks @ashiagr! All LGTM ;)

> Regarding below warnings, Android scan screen specs are not yet ready. I can add temporary dp width for now. Ideally I'd wait for the first set of specs to be shared with us before I start making these refinements. Lmkwyt.

Sounds good to me. One thing we might want to consider is adding checkboxes to the parent issue (https://github.com/wordpress-mobile/WordPress-Android/issues/13326) so we don't forget about it. Feel free to merge this PR.

ashiagr:
> One thing we might want to consider is adding checkboxes to the parent issue (#13326) so we don't forget about it.

Checkbox added under Scan - Polish the UI: 
> Fine-tune icons: include dp width in the icon file names, color resources from colors.xml.

Merging is blocked after base change. Re-requesting approval.

malinajirka (approved):
LGTM, thanks!

PR #13586, opened 2020-12-11
Issue/13329 backup wizard skeleton

zwarm:
Parent #13329 

This PR supports using the `WizardManager` for the backup download process and includes the basics for getting this started: dagger support, defining steps and simple navigation. The remainder of the logic will be implemented in future PRs.
This PR also includes some cleanup based on Kotlin guidelines - removal of redundant qualifiers

~~NOTE: Lint is failing on an unrelated issue with Utils package. A PR is underway to suppress the error in lint-baseline.xml~~

To test:
- Launch the app and enable the BackupFeatureConfig (restart the app)
- Tap on more menu
- Tap the "Download backup" menu item
Note that the backup detail view is launched (The feature itself is not functional, this is just to test that the view shows)

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadActivity.kt
```diff
@@ -21,52 +34,94 @@ class BackupDownloadActivity : LocaleAwareActivity() {
...
+    private fun initViewModel(savedInstanceState: Bundle?) {
         viewModel = ViewModelProvider(this, viewModelFactory)
                 .get(BackupDownloadViewModel::class.java)
-        viewModel.start()
 
-        // todo: annmarie temporary start
-        val fragment = BackupDownloadDetailsFragment.newInstance()
-        showFragment(fragment, BackupDownloadDetailsFragment.TAG, slideIn = false, isRootFragment = true)
+        viewModel.navigationTargetObservable
+                .observe(this, { target ->
```
ParaskP7: Minor (🔍 ): Consider moving the `.observer(...)` to the line above, just like it is done for the rest of the observables, to keep this one consistent as well.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadActivity.kt
```diff
@@ -21,52 +34,94 @@ class BackupDownloadActivity : LocaleAwareActivity() {
...
+            // Change the activity title for accessibility purposes
+            this.title = getString(title)
+        })
+
+        // todo: annmarie - may there are more states?
+        // Canceled, Running, Complete -> (Running = kick off status)
+        viewModel.wizardFinishedObservable.observe(this, { backupDownloadWizardState ->
+            backupDownloadWizardState?.let {
+                val intent = Intent()
+                val (backupDownloadCreated, activityId) = when (backupDownloadWizardState) {
```
ParaskP7: Minor (🔍 ): The `activityId` is unused atm. I am not sure how/if it is going to be used in the future. Consider renaming it to `_` for now.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadActivity.kt
```diff
@@ -21,52 +34,94 @@ class BackupDownloadActivity : LocaleAwareActivity() {
...
+                    // teh request was canceled
+                    is BackupDownloadCanceled -> Pair(false, null)
+                    is BackupDownloadInProgress -> {
```
ParaskP7: Minor (🔍 ): Consider refactor this to a one-liner to be consistent with the rest. The comment can then be moved to the when value instead, something like:
```
// The request is in progress and user didn't want to wait around
is BackupDownloadInProgress -> Pair(true, backupDownloadWizardState.activityId)
                    
```

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadStep.kt
```diff
@@ -0,0 +1,31 @@
...
+
+import org.wordpress.android.util.wizard.WizardStep
+import javax.inject.Inject
+import javax.inject.Singleton
+
+enum class BackupDownloadStep : WizardStep {
+    DETAILS, PROGRESS, COMPLETE;
+
+    companion object {
+        fun fromString(input: String): BackupDownloadStep {
```
ParaskP7: Minor (🔍 ): Consider simplifying this function to something like: `fun fromString(input: String) = when (input) { ... }`

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadStep.kt
```diff
@@ -0,0 +1,31 @@
...
+                "backup_download_complete" -> COMPLETE
+                else -> throw IllegalArgumentException("SiteCreationStep not recognized: \$input")
+            }
+        }
+    }
+}
+
+@Singleton
+class BackupDownloadStepsProvider @Inject constructor() {
+    fun getSteps(): List<BackupDownloadStep> {
```
ParaskP7: 1) Minor (🔍 ): Consider simplifying the function to something like: `fun getSteps() = listOf(...)`
2) Question (❓ ): Does the list order matter?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
```diff
@@ -1,15 +1,130 @@
...
+import org.wordpress.android.viewmodel.SingleLiveEvent
 import javax.inject.Inject
 
-class BackupDownloadViewModel @Inject constructor() : ViewModel() {
-    private var isStarted: Boolean = false
-    fun start() {
-        if (isStarted) {
-            return
-        }
+const val KEY_BACKUP_DOWNLOAD_ACTIVITY_ID_KEY: String = "key_backup_download_activity_id_key"
```
ParaskP7: Minor (🔍 ): Consider removing explicit type specification for this constant to make it consistent with the rest.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
```diff
@@ -1,15 +1,130 @@
...
+            )
+            backupDownloadState = requireNotNull(
+                    savedInstanceState.getParcelable(
+                            KEY_BACKUP_DOWNLOAD_STATE
+                    )
+            )
+            val currentStepIndex = savedInstanceState.getInt(KEY_BACKUP_DOWNLOAD_CURRENT_STEP)
+            wizardManager.setCurrentStepIndex(currentStepIndex)
+        }
         isStarted = true
```
ParaskP7: Question (❓ ): Was this `isStarted` moved in between the if statement logic for a reason? If not, consider moving that close to the `if (isStarted) return` logic.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
```diff
@@ -1,15 +1,130 @@
...
+        if (savedInstanceState == null) {
```
ParaskP7: Question (❓ ): Why was the if logic split here? Why not doing this below? Does this have to do somehow with the `isStarted` state?
```
        if (savedInstanceState == null) {
            backupDownloadState = BackupDownloadState()
            // Show the next step only if it's a fresh activity so we can handle the navigation
            wizardManager.showNextStep()
        }
```
```

File: WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt [hunk 9]
```diff
@@ -1,15 +1,130 @@
...
+
+    private val _onBackPressedObservable = SingleLiveEvent<Unit>()
+    val onBackPressedObservable: LiveData<Unit> = _onBackPressedObservable
+
+    fun start(savedInstanceState: Bundle?) {
+        if (isStarted) return
+        if (savedInstanceState == null) {
+            backupDownloadState = BackupDownloadState()
+        } else {
+            backupDownloadCompleted = savedInstanceState.getBoolean(
```
ParaskP7: Minor (🔍 ): Consider making all the assignment a one-liner since they fit.
```
backupDownloadCompleted = savedInstanceState.getBoolean(KEY_BACKUP_DOWNLOAD_COMPLETED, false)
backupDownloadState = requireNotNull(savedInstanceState.getParcelable(KEY_BACKUP_DOWNLOAD_STATE))
            
```

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
```diff
@@ -1,15 +1,130 @@
...
+        }
+    }
+
+    fun writeToBundle(outState: Bundle) {
+        outState.putBoolean(KEY_BACKUP_DOWNLOAD_COMPLETED, backupDownloadCompleted)
+        outState.putInt(KEY_BACKUP_DOWNLOAD_CURRENT_STEP, wizardManager.currentStep)
+        outState.putParcelable(KEY_BACKUP_DOWNLOAD_STATE, backupDownloadState)
+    }
+
+    fun onBackPressed() {
```
ParaskP7, line 88: 1) Minor (🔍 ): Consider making this a one-liner and move the `TODO` comment like below.
```
    // todo: annmarie - what should happen on backPress - always exit, please revisit
    fun onBackPressed() = exitFlow()
```
2) Minor (🔍 ): Since `exitFlow()` is used only here, consider moving this function closer to usage. Also, why not using `_exitFlowObservable.call()` directly?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
```diff
@@ -1,15 +1,130 @@
...
+        return exitFlow()
+        // todo: annmarie - what should happen on backPress - always exit, please revisit
+    }
+
+    private fun clearOldBackupDownloadState(wizardStep: BackupDownloadStep) {
```
ParaskP7, line 93: 1) Minor (🔍 ): Consider moving this function close to usage, right below the `navigationTargetObservable`.
2) Idea (💡 ): Since `DETAILS` and `PROGRESS` have duplicate logic, consider extracting the logic in a private method and reusing it, then the function can transform to:
```
    private fun clearOldBackupDownloadState(wizardStep: BackupDownloadStep) = when (wizardStep) {
        DETAILS -> extractedLogic()
        PROGRESS -> extractedLogic()
        COMPLETE -> Unit // intentionally left empty
    }
```

ParaskP7 (commented):
👋 @zwarm !

I have tested the solution and left minors/questions. Let me know if you need my help.

Question (❓ ): Will the testing part of the view models be covered later on?

PS: I found it difficult to understand the whole wizard logic and how it will be clued together. Thus, I didn't do a thorough enough review. I will probably get all the information I need from the next PRs, so I'll just keep it on.
PSS: I usually do commit-by-commit PRs. It helps me when commits are small and structured in a way that I can clearly follow the author's development process and train on thought. On the other hand, if I get commits like `Add wizard support`, which IMHO does lots, I find it hard to review and more often than not, I don't do a good job in reviewing it.

zwarm:
> Question (❓ ): Will the testing part of the view models be covered later on?

Unit test will come later on, as a lot of this is prep work. This PR is actually a combination of a POC and skeletons. 
 
> PS: I found it difficult to understand the whole wizard logic and how it will be clued together. Thus, I didn't do a thorough enough review. I will probably get all the information I need from the next PRs, so I'll just keep it on.

I hear ya, it's not all wired together because I was only committing to the first step and it's working. I still haven't decided if I want to use a "parentViewModel" or "listener" based arch for communication back to the activity. Would love to hear your thoughts, but that is for outside of this PR.  Take a look at SiteCreation for a full implementation of wizard. I only have the starting points here

> PSS: I usually do commit-by-commit PRs. It helps me when commits are small and structured in a way that I can clearly follow the author's development process and train on thought. On the other hand, if I get commits like `Add wizard support`, which IMHO does lots, I find it hard to review and more often than not, I don't do a good job in reviewing it.

It was tough in this one, since I had started as a POC and felt it was a better use of time to continue with it than start over. My following PRs use that format.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadStep.kt
zwarm: > Question (❓ ): Does the list order matter?

It appears that they do.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
zwarm: > Was this isStarted moved in between the if statement logic for a reason? If not, consider moving that close to the if (isStarted) return logic.

Nope, I'll move it in a later branch

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
zwarm: > Question (❓ ): Why was the if logic split here? Why not doing this below? Does this have to do somehow with the isStarted state?

Because it was lifted from SiteCreationVM for my POC. It's changed a future branch

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
zwarm: > Since exitFlow() is used only here, consider moving this function closer to usage. Also, why not using _exitFlowObservable.call() directly?

Lots of tweaking coming to this class ... it may not be the only place that uses exitFlow(). I'll note this for later. thanks.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
zwarm: > Since DETAILS and PROGRESS have duplicate logic, consider extracting the logic in a private method and reusing it, then the function can transform to:

There is lots of duplicate logic right now. On the surface everything is the same, but the implementation for each part is yet to come. Thanks for pointing this out.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadStep.kt
ParaskP7: > It appears that they do.

Thanks for clarifying that for me.

ParaskP7:
> Unit test will come later on, as a lot of this is prep work. This PR is actually a combination of a POC and skeletons.

Thanks for letting me know! 🌟 

> I hear ya, it's not all wired together because I was only committing to the first step and it's working. I still haven't decided if I want to use a "parentViewModel" or "listener" based arch for communication back to the activity. Would love to hear your thoughts, but that is for outside of this PR. Take a look at SiteCreation for a full implementation of wizard. I only have the starting points here

Many thanks, I'll take a look at the `SiteCreation` to understand more about it. 🙏 

> It was tough in this one, since I had started as a POC and felt it was a better use of time to continue with it than start over. My following PRs use that format.

My future self thanks you! 😄

ParaskP7:
I am assigned to this PR, after our group chat confirmation, but it is not 100% clear to me whether I should merge this or whether you want to do any additional work on it. If it's okay, as a reviewing, I'll start merging the PRs from the next one I'll get assigned to?

I'll just approve this one for you to merge when you are ready.

PS: Also, I am taking this opportunity to reply to the below comment from your [4th PR down the line](https://github.com/wordpress-mobile/WordPress-Android/pull/13595), see below:

> Unfortunately this PR (like some of the others) is still valid, but old. I still have 2 additional branches locally, that change quite a bit what you see here. I have been moving faster than the PRs can get reviewed. I don't know how we can solve this issue, but an open to suggestions.

This is really important and I think we should work it out within our team on how to handle that going forward. That is one of the reasons that I am not comfortable merging a PR as a reviewer. I don't know how all those other branches and logic look like, I can only assume and trust the author. Also, PR comments that are already addressed or will be address on those branches make me less confidence.

What I used to do in my previous roles, when I was facing this situation, is to have only a depth level of 3 chained PRs. And, I was trying really hard to avoid even that, when possible. Thinking about it, I have the feeling that all these 4 PRs, that are chained with each other, could have been separated from one another and target `developer` directly, don't you think? Would that help?

ParaskP7 (approved):

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
```diff
@@ -1331,8 +1332,18 @@ public static void viewPagesInNewStack(Context context) {
         context.startActivity(intent);
     }
 
-    public static void showBackupDownload(Activity activity) {
+    public static void showBackupDownloadForResult(
```
malinajirka: Nitpick 🔍 : We usually don't put parameter-per-line in java classes. We might want to change this to:
```suggestion
    public static void showBackupDownloadForResult(Activity activity, SiteModel site, String activityId,
                                                   int resultCode) {
```
Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
```diff
@@ -1331,8 +1332,18 @@ public static void viewPagesInNewStack(Context context) {
...
+            Activity activity,
+            SiteModel site,
```
malinajirka: Nitpick 🔍 : We might want to consider using `@NonNull`. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
```diff
@@ -1,15 +1,130 @@
...
+    val navigationTargetObservable: SingleEventObservable<NavigationTarget> by lazy {
+        SingleEventObservable(
+                Transformations.map(wizardManager.navigatorLiveData) {
+                    clearOldBackupDownloadState(it)
+                    WizardNavigationTarget(it, backupDownloadState)
+                }
+        )
+    }
 
+    private val _wizardFinishedObservable = SingleLiveEvent<BackupDownloadWizardState>()
```
malinajirka: Warning ⚠️ : We don't use SingleLiveEvent in new classes. We might want to consider replacing it with MutableLiveData<Event< BackupDownloadWizardState>>(). Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
```diff
@@ -1,15 +1,130 @@
...
+    val wizardFinishedObservable: LiveData<BackupDownloadWizardState> = _wizardFinishedObservable
+
+    private val _screenTitleObservable = SingleLiveEvent<@StringRes Int>()
```
malinajirka, line 57: Question ❓: Do you plan to use a single UiState in the upcoming PRs or use an observable per ui item?

Both approaches are fine within the codebase. So feel free to use either. I just wanted to mention that the VM doesn't need to use all practices used in SiteCreation. Wdyt?

malinajirka (approved):
Thanks @zwarm for working on this ;) and thank you @ParaskP7 for the review!

I've left some comments. Most of them are minors. The one with the warning symbol isn't critical, but I think it'd be nice to update it (at least in one of the next PRs). 

>> Unfortunately this PR (like some of the others) is still valid, but old. I still have 2 additional branches locally, that change quite a bit what you see here. I have been moving faster than the PRs can get reviewed. I don't know how we can solve this issue, but an open to suggestions.


Regarding `This is really important and I think w....`. If any of you has proposals how to improve the process, let's change the way we do it. Here is my take on this. 

1. It might have been a bit easier when we used feature branches for the whole feature (we replaced them with feature flags). Both approaches have their pros and cons and this is one of the cons.
2. I usually fix the issue in a single commit. I reply to the suggestions with a link to the commit. If I'm afraid the commit would cause a lot of difficult conflicts it's ok to make the change on the following PRs and share a link to the commit on the PR under review.
3. When it's a nice-to-have thing and I don't want to work on the code right now, I usually create a checkbox on the related GitHub issue so the author on the comment knows it won't be forgotten.
 
> I don't know how all those other branches and logic look like, I can only assume and trust the author.

I usually verify it's behind a feature flag = it's not user facing change. If it's not behind a feature flag all merged code into develop should be final and ready for production. When I need to chain PRs and there isn't a feature flag I always create a temporary feature branch.
When a PR with removal of the feature flag is created, the reviewer shouldn't do a full review, but at the same time they should scan through all the code which gets enabled (eg. check there aren't any todos + thoroughly test the app). I agree it's not perfect but I don't know about any other approach how to mitigate this. Wdyt?

> is to have only a depth level of 3 chained PRs

We use this "unspoken" rule as well (there are rare exceptions).

> Thinking about it, I have the feeling that all these 4 PRs, that are chained with each other, could have been separated from one another and target developer directly, don't you think?

I didn't check the PRs but if that's the case I agree it might be easier to keep them separated.

ParaskP7:
> Regarding This is really important and I think w..... If any of you has proposals how to improve the process, let's change the way we do it. Here is my take on this. 1) It might have been a bit easier when we used feature branches for the whole feature (we replaced them with feature flags). Both approaches have their pros and cons and this is one of the cons. 2) I usually fix the issue in a single commit. I reply to the suggestions with a link to the commit. If I'm afraid the commit would cause a lot of difficult conflicts it's ok to make the change on the following PRs and share a link to the commit on the PR under review. 3) When it's a nice-to-have thing and I don't want to work on the code right now, I usually create a checkbox on the related GitHub issue so the author on the comment knows it won't be forgotten.

1) I am not sure whether this is a feature branch problem, or I am afraid I don't understand what you meant. If we were to use a feature branch, we would have the same problem in this instance, no? The problem is that there are 6 PRs (ish), one chained after the other and 4 of them pushed for review, all that happens within a specific time frame, with the code review happening later. Thus, this created this problem. I am not sure how would feature branches solve that?
2) Yes, I like that, I do the same, whenever there exist a comment that I addressed in the PR, I'll make sure to add the commit that fixes it in the response. Sharing the commit from the next PR is also nice and can help with keeping track of the to be resolved comments, thanks for sharing.
3) That is a nice idea, thanks for sharing!

Btw, @zwarm, @malinajirka and @ashiagr, what is your process for resolving code review comments. I resolve only my own code review comments when I am a reviewer, when I am the author I never resolve comments of the reviews. How do you deal with that? We might want to align there as well.

> I usually verify it's behind a feature flag = it's not user facing change. If it's not behind a feature flag all merged code into develop should be final and ready for production. When I need to chain PRs and there isn't a feature flag I always create a temporary feature branch.
When a PR with removal of the feature flag is created, the reviewer shouldn't do a full review, but at the same time they should scan through all the code which gets enabled (eg. check there aren't any todos + thoroughly test the app). I agree it's not perfect but I don't know about any other approach how to mitigate this. Wdyt?

Yeap, I am afraid of that too. For the parent PRs that might be merged to `develop` at a future time, usually less effort is put on reviewing (correctly so), assuming that the children PRs were thoroughly reviewed, and if possible tested.

> We use this "unspoken" rule as well (there are rare exceptions).

Good to know, thanks for sharing! 🌟

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
zwarm: > We usually don't put parameter-per-line in java classes

I did not know this. Thanks

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
zwarm: A good improvement - yep, I was following existing pattern

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
zwarm: > We don't use SingleLiveEvent in new classes

That's good to know. Thanks

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
zwarm: > I just wanted to mention that the VM doesn't need to use all practices used in SiteCreation

Sounds good to me. I'm still on the fence and letting this play out as I move along with implementation

zwarm:
@ParaskP7 
> I have the feeling that all these 4 PRs, that are chained with each other, could have been separated from one another and target developer directly, don't you think?

Not for this one, because each following PR make references code in the parent PR. 

I guess I should have waited for the review to be done before moving forward.

zwarm:
@ParaskP7 @malinajirka

ParaskP7:
> Not for this one, because each following PR make references code in the parent PR.

I was checking that and realised that each following PR didn't have to make reference in the parent PR, I might be missing something. If you want to pick my brain on the details, I would love to sit and further discuss that with you.

zwarm:
> > Not for this one, because each following PR make references code in the parent PR.
> 
> I was checking that and realised that each following PR didn't have to make reference in the parent PR, I might be missing something. If you want to pick my brain on the details, I would love to sit and further discuss that with you.

I created the branches off the previous branch.:(

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadActivity.kt
zwarm: In 211c480

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadActivity.kt
zwarm: in 211c480

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadActivity.kt
zwarm: in 211c480

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
zwarm: in 4fabf1b

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
zwarm: in 4fabf1b

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadStep.kt
zwarm: in 6448ff8

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadStep.kt
zwarm: in 6448ff8

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
zwarm: > Consider making all the assignment a one-liner since they fit.

Running a code reformat pushes these to separate lines, so if I move them to one line and then code reformat they'll move back.

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
malinajirka: Yeah, there are both patterns used within this class. But I think using @NonNull is clearer. Thanks for the change!

malinajirka:
Thanks for the changes @zwarm! There is one unresolved discussion but - feel free to make the change later and merge this PR or make the change here and merge this PR (it's not being used right now anyway).

malinajirka:
> I resolve only my own code review comments when I am a reviewer, when I am the author I never resolve comments of the reviews.

Yes, this is our practice. The author of the comment is responsible for resolving it. Unless another reviewer takes over the review, in that case they can resolve it. (cc @ParaskP7 )

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
zwarm: In ef12c38
With the exception of `_screenTitleObservable` which changes to UiToolbarState later on down the road

zwarm:
@malinajirka  - I made most recommended changes including the nit picks. :)

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
zwarm: in ef12c38

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
zwarm: in ef12c38

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
zwarm: in ef12c38

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
zwarm: in ef12c38 - but they'll change back, so 🤯

malinajirka:
Awesome, thank you so much! Looks great!

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadViewModel.kt
malinajirka: If that's the case you might want to double check setup of your formatter for kotlin files - AFAIK we use 120chars per line.

ParaskP7:
Thanks Annmarie, I also had a quick look, looks great! 🌟

PR #13587, opened 2020-12-11
Issue/13329 backup progress skeleton

zwarm:
parent #13329 

This PR creates skeleton fragment, vm, and test for the Backup Download Progress screen.

**Merge instructions**
- Make sure #13586 has been merged.
- Remove Not Ready for Merge label
- Merge as normal

To test:
There is nothing to test on this PR

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/progress/BackupDownloadProgressFragment.kt
```diff
@@ -0,0 +1,39 @@
...
+import javax.inject.Inject
+
+class BackupDownloadProgressFragment : Fragment(R.layout.backup_download_progress_fragment) {
+    @Inject lateinit var viewModelFactory: ViewModelProvider.Factory
+    private lateinit var viewModel: BackupDownloadProgressViewModel
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        val nonNullActivity = requireActivity()
+        (nonNullActivity.application as WordPress).component().inject(this)
```
ParaskP7: Idea (💡 ): These two lines can be extracted into a descriptive one-liner `initDagger` private function:
```
    private fun initDagger() {
        (requireActivity().application as WordPress).component().inject(this)
    }
```
Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/progress/BackupDownloadProgressFragment.kt
```diff
@@ -0,0 +1,39 @@
...
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        initViewModel()
+    }
+
+    private fun initViewModel() {
+        viewModel = ViewModelProvider(
+                this,
+                viewModelFactory
+        ).get(BackupDownloadProgressViewModel::class.java)
```
ParaskP7: Minor (🔍 ): Consider breaking it down into two lines and by the `.get(...)` instead, like it is used in the builder pattern:
```
        viewModel = ViewModelProvider(this, viewModelFactory)
                .get(BackupDownloadProgressViewModel::class.java)
```

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/progress/BackupDownloadProgressViewModel.kt
```diff
@@ -0,0 +1,15 @@
+package org.wordpress.android.ui.jetpack.backup.progress
+
+import androidx.lifecycle.ViewModel
+import javax.inject.Inject
+
+class BackupDownloadProgressViewModel @Inject constructor() : ViewModel() {
+    private var isStarted = false
+
+    fun start() {
+        if (isStarted) {
```
ParaskP7, line 10: Minor (🔍 ): Consider making this a one-liner. This way `if (started) return` and `isStarted = true` will be as close to each other as possible.

WordPress/src/main/res/layout/backup_download_progress_fragment.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
```
ParaskP7, line 1: Minor (🔍 ): Consider reformatting this xml file.

WordPress/src/main/res/layout/backup_download_progress_fragment.xml
```diff
@@ -0,0 +1,19 @@
...
+    xmlns:tools="http://schemas.android.com/tools">
+
+    <com.google.android.material.button.MaterialButton
+        android:id="@+id/backup_details_action_button"
+        style="@style/Backup.Details.PrimaryButton"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintBottom_toBottomOf="parent"
+        tools:text="Next step">
```
ParaskP7, line 16: Minor (🔍 ): Consider ending the tags with `/>` when possible.

ParaskP7 (commented):
👋 @zwarm !

I have reviewed the solution and left minors/ideas. Let me know if you need my help.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/progress/BackupDownloadProgressViewModel.kt
zwarm: This is another one of those instances in which the code base does both. 🤷

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/progress/BackupDownloadProgressFragment.kt
zwarm: It has already been changed in a future branch

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/progress/BackupDownloadProgressFragment.kt
zwarm: @ParaskP7 - I'm on the fence about this one. (A) because we mention Dagger and yes, although we use it exclusively today, we might not in the future so I don't want to litter the code with references to an external library (B) It seems like overkill to put this one line in its own fun. Another good candidate for the team to discuss in slack.

ParaskP7: Thanks @zwarm ! 🌟 

1) I agree on the logic that `Dagger` is too specific, I was also thinking about calling that `initDi()` or something like that, but I end up going with the `initDagger()` suggestion because actually this is a code line specific to `Dagger`, not an interface we can swap and use another framework instead. All `DI` frameworks I used so far have their own way of dealing with dependency injection.
2) It seems indeed that it is an overkill, I agree. How I think about it is that having `initDagger()` instead of this more complex inject dependency injection related line, will make it easier for me to read the code without needing to always read it again and again. By reading `initDagger()` I will not put an effort next time in reading the whole line trying to figure out if something is missing.

PS: This is just conversation for conversation sake, and it is very subjective, so please treat it as such. 🙏

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/progress/BackupDownloadProgressViewModel.kt
ParaskP7: Definitely, I agree on that, in my case, if I were to write this code I would always choose the one-liner alternative for such simple logic. This is also what the official Kotlin style guide suggests, thus it makes it easier on me deciding on it.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/progress/BackupDownloadProgressFragment.kt
zwarm: > I will not put an effort next time in reading the whole line trying to figure out if something is missing.

I can go along with this reasoning - especially if we all agree to move it. See commit efa3a68

ParaskP7 (approved):

ParaskP7:
👋 @zwarm !

Since this PR is pretty similar to [the next one](https://github.com/wordpress-mobile/WordPress-Android/pull/13588) that I already approved, I assigned myself to this as well and approved it to unblock you, feel free to merge it when you feel ready.

PR #13592, opened 2020-12-11
Suppress a lint false positive related to WordPress-Android-Utils

loremattei:
After having merged https://github.com/wordpress-mobile/WordPress-Android/pull/13466, we have started to experience random lint failures on CI.

After some more analysis, it still seems a false positive to me: 
- It looks like the failure is raised the first time a new commit is built and gets resolved by building a second time. 
- the error itself says that `SwipeToRefreshHelper` expects a resource id, but [looking at the code](https://github.com/wordpress-mobile/WordPress-Utils-Android/blob/a15474122f61bce40117ad1fc9cd90458232b75a/WordPressUtils/src/main/java/org/wordpress/android/util/helpers/SwipeToRefreshHelper.java#L53), it doesn't seem to be the case. 

**To Test:**
1. Double check the code to make sure I'm not blind and missing an actual error. 
2. Verify that CI is green. 

PR submission checklist:

### DIFF
--- a/WordPress/lint-baseline.xml
+++ b/WordPress/lint-baseline.xml
@@ -4060,4 +4060,20 @@
             column="28"/>
     </issue>
 
+    <issue
+        id="ResourceType"
+        severity="Error"
+        message="Expected a color resource id (`R.color.`) but received an RGB integer"
+        category="Correctness"
+        priority="7"
+        summary="Wrong Resource Type"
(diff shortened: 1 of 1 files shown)


zwarm (approved):
@loremattei - Everything looks good to my untrained baseline rule writing eyes. I'd feel better if we had someone else jump in with an approval too. I tried to do a before/after test locally, but I am getting 100+ errors when `lintVanillaRelease` just on the before test. I don't trust my local setup.

malinajirka (approved):
Thank you both! ;) 

LGTM 🚢

PR #13593, opened 2020-12-11
Remove Gutenberg Editor Popup

mchowning:
We will be removing the ability to change the default editor to be anything but Gutenberg shortly. Before that though, we wanted to remove the dialog advising of the "new" Gutenberg editor.

#### All SItes Defaulted To Gutenberg

1. Clear all the app's data
1. Log into an account where _all_ the sites on the account are defaulted to Gutenberg
1. **Background the app and bring it back to the foreground**
1. Open the Editor
1. Verify that the Gutenberg editor is presented and there is no popup

#### Site Defaulting to Aztec

1. After completing the "All SItes Defaulted To Gutenberg" test, change the default for one of the sites to Aztec. Note that this is remarkably difficult to do reliably because the default editor toggle [is barely functional](https://github.com/wordpress-mobile/gutenberg-mobile/issues/1632). The most reliable method is to ignore the visual state of the toggle and instead when you use the toggle check the actual API call to confirm that the response from the backend states that the site is set to use Aztec on mobile.
1. Clear the app's data
1. Log into the same account
1. **Background the app and bring it back to the foreground**
1. Open the Editor on the site that you changed to use Aztec in Step 1
1. Verify that the Gutenberg editor is presented and there is no popup
1. Exit the Editor 
1. Open the Editor **again**
1. Verify that the Gutenberg editor is presented and there is no popup

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -1625,27 +1623,18 @@ private void performPrimaryAction() {
...
-                mEditPostRepository.isPage() ? getString(R.string.dialog_gutenberg_informative_description_page)
-                        : getString(R.string.dialog_gutenberg_informative_description_post),
-                getString(org.wordpress.android.editor.R.string.dialog_button_ok));
-
-        gbInformativeDialog.show(getSupportFragmentManager(), TAG_GB_INFORMATIVE_DIALOG);
+        // We are no longer showing the dialog, but we are leaving all the surrounding logic because
+        // this is going in shortly before release, and we're going to remove all this logic in the
+        // very near future.
+
         AppPrefs.setGutenbergInfoPopupDisplayed(mSite.getUrl(), true);
```
mchowning, line 1652: I left this preference because that preference is [used as an early exit in the setGutenbergEnabledIfNeeded function](https://github.com/wordpress-mobile/WordPress-Android/blob/76a49558062397c35be6c6a19e671fd3aba7c479/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L1642-L1644), and I wanted to avoid any changes that would affect the control flow.

cameronvoell: I don't think `showGutenbergInformativeDialog` is called anywhere anymore, so this code shouldnt be relevant. But I don't want to be too picky since we want to err on the side of being safe and not perfect in stripping out this functionality. We can remove this after ongoing releases to be safer if we prefer.

cameronvoell (approved):
Seems to be working for me. Left one comment but I think t can be dealt with after ongoing releases in a follow up PR.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
mchowning: Good catch @cameronvoell ! This means that I actually was changing the control-flow in a small way--exactly the thing I wanted to avoid. For that reason, I have updated this to make fewer changes and leave the control-flow intact. Sorry to ask you to retest this, but I think that's just the safer approach at this point.

In addition, there is an additional step of backgrounding the app that I added to the testing instructions (in bold). It is important to background the app after logging in because it is when the app is backgrounded that this code gets run.

I also wanted to alert you that when I had a site with a default of Aztec and as part of the gutenberg-default-migration, the POST call is made to `wpcom/v2/me/gutenberg` to update the mobile editor default to be gutenberg. The response to that call would indicate that the site was now defaulting to gutenberg on mobile, **but on some occasions it would not update and the response would say that the site was still Aztec**. I need to investigate this further, but I wanted to alert you to it in case you see anything like that. I believe this issue is unrelated to any app-side logic, so I don't think it's anything that should block this PR. I'll be investigating it further when I remove the gutenberg toggle.

mchowning:
I'm a little stumped as to how my last commit led to [these lint errors](https://app.circleci.com/pipelines/github/wordpress-mobile/WordPress-Android/19288/workflows/0bd9f898-d079-4e8d-bb0d-43155bdb7f72/jobs/87238). 🤔

loremattei:
Hey! I've moved this to 16.5 because 16.4 has been cut. If you want this to make it to 16.4, please feel free to ping me.

mchowning:
> Hey! I've moved this to 16.5 because 16.4 has been cut. If you want this to make it to 16.4, please feel free to ping me.

:wave: @loremattei ! This does need to go into 16.4, so I updated the milestone and retargeted the PR to `release/16.4`. I expect that this will get merged today.

guarani (approved):
Tested using the apk build and I see no traces of the popup. Approving now from a testing perspective.

PR #13588, opened 2020-12-11
Issue/13329 backup complete skeleton

zwarm:
Parent #13329 

This PR creates skeleton fragment, vm, and test for the Backup Download Progress screen.

Merge instructions
- Make sure #13586 & #13587 have been merged.
- Remove Not Ready for Merge label
- Merge as normal

To test:
There is nothing to test on this PR

PR submission checklist:

- [X] I have considered adding unit tests where possible.
- [X] I have considered adding accessibility improvements for my changes.
- [X] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13595, opened 2020-12-12
Issue/13329 backup download list states

zwarm:
Parent #13329 

The PR contains prep work for switching BackupDownload detail view to a full page recyclerview.. It introduces `BackupDownloadListState` which will be used when displaying backupDownload Details. 

The PR contains
- New layout files
- The initial states (these will be enhanced as the feature continues to be developed & names changes come in next PR too)
- A string addition

I am most interested in opinions on the listState usage ... each item on the screen is broken down into its own state. Thoughts? 

Note: The next PR includes setting up the view holders. There are also some name changes.

Merge Instructions
Ensure that #13588  has been merged.

To test:
There is nothing new to test in this PR

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/details/BackupDownloadDetailsFragment.kt
```diff
@@ -74,6 +74,17 @@ class BackupDownloadDetailsFragment : Fragment() {
         ((recycler_view.adapter) as BackupDownloadDetailsAdapter).update(content.items)
     }
 
+    // todo: annmarie -(REMOVE) this dummy method references layout files that lint says aren't
+    // used, but they will be in the next PR because there were too many changes for 1 PR
+    private fun dummyAccess() {
```
ParaskP7, line 79: Minor (🔍 ): Since this function is not used, and it is a dummy function, I wouldn't have added it to this PR and that would make me avoid any static analysis warnings as well. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadListItemState.kt
```diff
@@ -0,0 +1,88 @@
...
+import org.wordpress.android.ui.jetpack.backup.BackupDownloadListItemState.ViewType.SUBHEADER
+import org.wordpress.android.ui.utils.UiString
+import org.wordpress.android.ui.utils.UiString.UiStringRes
+import org.wordpress.android.ui.utils.UiString.UiStringResWithParams
+
+// todo: annmarie - tighten this up as the remainder of backup download is implemented
+sealed class BackupDownloadListItemState(val type: ViewType) {
+    open fun longId(): Long = hashCode().toLong()
+
+    sealed class HeaderState : BackupDownloadListItemState(HEADER) {
```
ParaskP7: Minor (🔍 ): Consider adding the `ViewType` explicitly and removing all the imports. This way it might become a bit more evident what this enum does and the need for any import will not be required.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadListItemState.kt
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    sealed class DescriptionState : BackupDownloadListItemState(DESCRIPTION) {
+        abstract val text: UiString
+
+        data class DetailsSubHeaderState(
+            override val text: UiString
+        ) : DescriptionState() {
+            constructor(params: List<UiString>) : this(
+                    text = UiStringResWithParams(
```
ParaskP7, line 37: Minor (🔍 ): Consider this one-liner alternative:
```
constructor(params: List<UiString>) : this(
        UiStringResWithParams(R.string.backup_download_details_choose_items_header, params)
)
```

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadListItemState.kt
```diff
@@ -0,0 +1,88 @@
...
+                            R.string.backup_download_details_choose_items_header,
+                            params
+                    )
+            )
+        }
+    }
+
+    sealed class IconState : BackupDownloadListItemState(DESCRIPTION) {
```
ParaskP7: Warning (⚠️): The `DESCRIPTION` enum usage is duplicated here. Probably this was suppose to be an `IMAGE` instead, right?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadListItemState.kt
```diff
@@ -0,0 +1,88 @@
...
+        abstract val contentDescription: UiString
+
+        data class DetailsIconState(
+            @DrawableRes override val icon: Int = R.drawable.ic_get_app_24dp,
+            override val contentDescription: UiString =
+                    UiStringRes(R.string.backup_download_details_icon_content_description)
+        ) : IconState()
+    }
+
+    sealed class ActionButtonState : BackupDownloadListItemState(ACTION_BUTTON) {
```
ParaskP7: Question (❓): How come this data class doesn't have an associated `DetailsCheckboxState` like the rest of them?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadListItemState.kt
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    data class CheckboxState(
+        val availableItemType: BackupAvailableItemType,
+        val label: UiString,
+        val checked: Boolean = false,
+        val onClick: (() -> Unit)
+    ) : BackupDownloadListItemState(CHECKBOX)
+
+    enum class ViewType(val id: Int) {
```
ParaskP7, line 75: 1) Question (❓): Are the enums ordered based on appearance on UI?
2) Minor (🔍 ): Consider matching the order of enums with the sealed/data class oder, as defined in this class.

WordPress/src/main/res/layout/backup_download_list_checkbox_item.xml
```diff
@@ -0,0 +1,38 @@
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
ParaskP7, line 1: 1) Question (❓): This is identical to `backup_details_list_item.xml` file, isn't it? Maybe this duplication should be dealt with now. Wdyt?
2) Minor (🔍 ): Consider reformatting this layout file, I believe they are not in the right format. PS: I think this apply to all the new layouts.

WordPress/src/main/res/layout/backup_download_details_fragment.xml
```diff
@@ -1,70 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
-<ScrollView xmlns:app="http://schemas.android.com/apk/res-auto"
-    xmlns:tools="http://schemas.android.com/tools"
-    xmlns:android="http://schemas.android.com/apk/res/android"
-    android:id="@+id/backup_details_scroll_view"
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
ParaskP7, line 2: Question (❓): The commit for this file change said `Rename styles to backup download for reuse`, but I saw that there was a bit of deletion happened as part of this commit for this xml file, what was the reason for that? Since the new functionality is not hooked up yet, I suggest that this UI stay as is, until it gets replaced. Otherwise, on the next PR it might look as if it was created from scratch. Wdyt?

ParaskP7 (commented):
👋 @zwarm !

I have reviewed the solution and left minors/questions (🔍 /❓), with just one warning (⚠️). Let me know if you need my help.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/details/BackupDownloadDetailsFragment.kt
zwarm: @ParaskP7  - I created and committed all the layout files and submitted with the PR. The PR fails CI without this fun and it is gone in the very next PR. The dummy fun passes CI - believe you me, I wasn't happy when I realized I couldn't commit all the layouts at once (seemed logical to me) and by the time I noticed I was two branches past this one.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadListItemState.kt
zwarm: Good suggestion. I'll use it in a future branch

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadListItemState.kt
zwarm: > The DESCRIPTION enum usage is duplicated here. Probably this was suppose to be an IMAGE instead, right?

Yes, already fixed in a future PR

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadListItemState.kt
zwarm: > How come this data class doesn't have an associated DetailsCheckboxState like the rest of them?

It is standalone. I have to add multiple checkbox rows and use a provider to create them, so there are no hardcoded values here. Plus it can be reused.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadListItemState.kt
zwarm: > Are the enums ordered based on appearance on UI?

Yep, because I created them while I was looking at the design mocks. :) I'm not married to them either way, I'll switch them to match in a future PR.

WordPress/src/main/res/layout/backup_download_list_checkbox_item.xml
zwarm: > This is identical to backup_details_list_item.xml file, isn't it? Maybe this duplication should be dealt with now. Wdyt?

This file no longer exists in a future branch. I am 3 branches ahead of this one locally.

WordPress/src/main/res/layout/backup_download_details_fragment.xml
zwarm: > Question (❓): 

Ah, the commit first changed the styles and then the next one changed to a recyclerView. Either way, this file is just a recyclerView and the parts that were deleted are truly gone.

zwarm:
@ParaskP7 - Unfortunately this PR (like some of the others) is still valid, but old. I still have 2 additional branches locally, that change quite a bit what you see here. I have been moving faster than the PRs can get reviewed. I don't know how we can solve this issue, but an open to suggestions.

BTW - All your one-line comments and suggestions I am tracking in my punch list. They are not entered in vain. Thanks!

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/details/BackupDownloadDetailsFragment.kt
ParaskP7: Yeap, it is always tricky to work with chained branches and PRs. 😞 

I have left an elaborate comment on your [1st PR in this group](https://github.com/wordpress-mobile/WordPress-Android/pull/13586). Let me know your thoughts.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadListItemState.kt
ParaskP7: Thanks for letting me know of that Annmarie! 🤗

ParaskP7:
> Unfortunately this PR (like some of the others) is still valid, but old. I still have 2 additional branches locally, that change quite a bit what you see here. I have been moving faster than the PRs can get reviewed. I don't know how we can solve this issue, but an open to suggestions.

I have left an elaborate comment on your [1st PR in this group](https://github.com/wordpress-mobile/WordPress-Android/pull/13586). Let me know your thoughts.

> BTW - All your one-line comments and suggestions I am tracking in my punch list. They are not entered in vain. Thanks!

No worries, please, I am not attached to them. Indeed treat them as `SUPER MINOR`, although I don't add the `SUPER` to not create another layer of commenting structure. They can be considered as food for thought (to both parties) or as something to team discuss for next time. Thank YOU! ❤️

ashiagr (approved):
Thank you both of you!

@zwarm , I've reviewed the changes and have only few thoughts:

1. The changes include `CheckboxState` for `BackupDownload` screen but `BackupDownload` UI doesn't seem to include any checkbox in the latest design iteration. I do see checkboxes for the `Restore` details screen. Is it intended (and included so that we can use the `ListStates` on both the screens)?

2.
> I am most interested in opinions on the listState usage ... each item on the screen is broken down into its own state.

I'm not sure about the best approach but in the `Scan` feature I've clubbed together elements (icon, text, description, buttons) into one sealed class corresponding to a scan state and threats as separate items.
This approach with separate VH for individual elements looks fine as well. We'll see how things proceed, hopefully both approaches should work fine, it's just the way we're grouping the elements. 

🟢 from my side ... nice work!

Feel free to merge once you're addressed comments from Petros.

zwarm:
> 1. The changes include `CheckboxState` for `BackupDownload` screen but `BackupDownload` UI doesn't seem to include any checkbox in the latest design iteration. I do see checkboxes for the `Restore` details screen. Is it intended (and included so that we can use the `ListStates` on both the screens)?

The i5 designs are for the standalone backup download process only. After tapping on the restore or create downloadable backup file buttons are pressed, the user is sent to this wizard to actually select the items. There are two entry points in (1) From the My Site menu and (2) the activity log. My PRs are entry from the activityLog list.

>  We'll see how things proceed, hopefully both approaches should work fine, it's just the way we're grouping the elements.

Let me take a look at scan to see what's going on there. Maybe we should align to keep things simple

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/BackupDownloadListItemState.kt
zwarm: Pushed in commit 3d9cb8c

ashiagr:
> The i5 designs are for the standalone backup download process only. After tapping on the restore or create downloadable backup file buttons are pressed, the user is sent to this wizard to actually select the items. There are two entry points in (1) From the My Site menu and (2) the activity log. My PRs are entry from the activityLog list.

Got it, thanks for the clarification @zwarm!

> Let me take a look at scan to see what's going on there. Maybe we should align to keep things simple

Yep, let's discuss it first before making any changes.

PR #13597, opened 2020-12-14
Exercise - WordPress Application class: move part of the initialization logic out from onCreate() to a thread

mzorz:
Aims at alleviating https://github.com/wordpress-mobile/WordPress-Android/issues/10494

This PR left intentionally as a draft for now, but pinging people to take a look and discuss. 

I have a (perhaps far-fetched) hypothesis. Having read the thread linked to the actual Android Chromium bug that makes the app crash, I wonder if making the time spent in `onCreate()` as less as possible can at any point make things easier for Android to create a new instance of the app on its own new process. Having read that onCreate() is blocking and Android will not create any Activity before onCreate has returned, I thought it might make sense to try something along these lines as an exercise.

From the [docs:](https://developer.android.com/reference/android/app/Application#onCreate()):
> Implementations should be as quick as possible (for example using lazy initialization of state) since the time spent in this function directly impacts the performance of starting the first activity, service, or receiver in a process.

> If you override this method, be sure to call super.onCreate().

This PR then, moves whatever can be moved from the Application's `onCreate()`  method out to a background thread, so to make the execution of `onCreate()` not block anything.

Unfortunately, we couldn't find a way to test the actual conditions in which the crash happens, so I'm opening this as a draft PR to discuss with others what they think. Again, I don't have a very strong feeling about the success with which this may help alleviate the case, but in any case would like to hear any thoughts you may have in going with an approach like this (basically, 
to return from onCreate as soon as possible).

To test:
N/A, just install and run the app.
- I've tested a fresh install
- upgrades (keeping data)
on a couple emulators (API 29).

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/WordPress.java
+++ b/WordPress/src/main/java/org/wordpress/android/WordPress.java
@@ -242,6 +242,89 @@ public class WordPress extends MultiDexApplication implements HasAndroidInjector
         return mBitmapCache;
     }
 
+    private void asyncInit(long startDate) {
+        new Thread(new Runnable() {
+            @Override
+            public void run() {
+                AppLog.i(T.UTILS, "init-measure STARTING ASYNC INIT()" + (SystemClock.elapsedRealtime() - startDate));
+                initWpDb();
+                enableHttpResponseCache(mContext);
+
+                AppRatingDialog.INSTANCE.init(mContext);
+
+                // EventBus setup
+                EventBus.TAG = "WordPress-EVENT";
+                EventBus.builder()
+                        .logNoSubscriberMessages(false)
+                        .sendNoSubscriberEvent(false)
+                        .throwSubscriberException(true)
+                        .installDefaultEventBus();
+
+
+                RestClientUtils.setUserAgent(getUserAgent());
+
+                mZendeskHelper.setupZendesk(mContext, BuildConfig.ZENDESK_DOMAIN, BuildConfig.ZENDESK_APP_ID,
+                        BuildConfig.ZENDESK_OAUTH_CLIENT_ID);
+
+                // initialize our ApplicationLifecycleMonitor, which is the App's LifecycleObserver implementation
+                mApplicationLifecycleMonitor = new ApplicationLifecycleMonitor();
+                ProcessLifecycleOwner.get().getLifecycle().addObserver(WordPress.this);
+
+                initAnalytics(SystemClock.elapsedRealtime() - startDate);
+
+                createNotificationChannelsOnSdk26();
+
+                // Allows vector drawable from resources (in selectors for instance) on Android < 21 (can cause issues
+                // with memory usage and the use of Configuration). More information: http://bit.ly/2H1KTQo
+                // Note: if removed, this will cause crashes on Android < 21
+                AppCompatDelegate.setCompatVectorFromResourcesEnabled(true);
+                AppThemeUtils.Companion.setAppTheme(mContext);
+
+                // verify media is sanitized
+                sanitizeMediaUploadStateForSite();
+
+                // remove expired lists
+                mDispatcher.dispatch(ListActionBuilder.newRemoveExpiredListsAction(new RemoveExpiredListsPayload()));
+
+                // setup the Credentials Client so we can clean it up on wpcom logout
+                mCredentialsClient = new GoogleApiClient.Builder(mContext)
+                        .addConnectionCallbacks(new GoogleApiClient.ConnectionCallbacks() {
+                            @Override
+                            public void onConnected(@Nullable Bundle bundle) {
+                            }
+
+                            @Override
+                            public void onConnectionSuspended(int i) {
+                            }
+                        })
+                        .addApi(Auth.CREDENTIALS_API)
+                        .build();
+                mCredentialsClient.connect();
+
+                initWorkManager();
+
+                // Enqueue our periodic upload work request. The UploadWorkRequest will be called even if the app is closed.
+                // It will upload local draft or published posts with local changes to the server.
+                UploadWorkerKt.enqueuePeriodicUploadWorkRequestForAllSites();
+
+                mSystemNotificationsTracker.checkSystemNotificationsState();
+                ImageEditorInitializer.Companion.init(
+                        mImageManager,
+                        mImageEditorTracker,
+                        mImageEditorFileUtils,
+                        mDefaultScope
+                );
+
+                initEmojiCompat();
+                mStoryNotificationTrackerProvider = new StoryNotificationTrackerProvider();
+                mStoryMediaSaveUploadBridge.init(mContext);
+                ProcessLifecycleOwner.get().getLifecycle().addObserver(mStoryMediaSaveUploadBridge);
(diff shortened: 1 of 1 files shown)


jd-alexander:
@mzorz thanks for looking into this 😄  Are there any caveats to or gotchas we have to look out for when moving logic to a background thread. Eg, I am seeing where UI-facing components are being initialized on a background thread. Is it that it's fine to do the initialization of components on a background thread and when ready to perform actual modification of the UI thread do we switch to it? I am just trying to understand how I should approach verifying that the logic is in full compliance based on the type of context and thread switching we are doing. Let me know if this helps. If it's not clear that's also fine. I am just trying to fill any gaps in my knowledge so I can contribute to a solution here.  

Also, would we want to introduce any form of performance measurements here to see if we notice any changes in execution time based on the approach we are trying in this PR?

mzorz:
Thanks for your comments @jd-alexander ! 
>  Is it that it's fine to do the initialization of components on a background thread and when ready to perform actual modification of the UI thread do we switch to it? I am just trying to understand how I should approach verifying that the logic is in full compliance based on the type of context and thread switching we are doing.

Exactly - the idea here is we process first then update the UI. We could [just do the minimum needed there](https://developer.android.com/topic/performance/vitals/launch-time#heavy-app) like initializing DI, registering for events, WellSql, etc. and use a new loading or [launch activity screen](https://material.io/design/communication/launch-screen.html#usage) while the rest of things are being initialized, even if it only takes 1-2 secs in total.

Measurements as you mentioned, tracking, correct handling of UI and such would need to be addressed for sure - I just drafted this very quickly to show on the whiteboard what I had in mind and get the discussion going; if we think it may be worth a try, we could actually implement the launch screen and roll out a version to a percentage of users and see how it goes. I was thinking if we think the approach might make sense, then  perhaps using [some A/B testing mechanism](https://developer.android.com/distribute/best-practices/develop/in-app-a-b-testing) for this works so to get us more data.

So the key question I'm asking devs here would be, apart from improving initial cold load times: **do you think the particular issue could be alleviated by making time spent in `onCreate()` shorter?** If yes then it's worth spending more time coming up with a solution that refactors initialization carefully (a lot is going on in there so, we need to make sure we're not breaking anything as well). Wdyt @jd-alexander ?

malinajirka:
Hi @mzorz, thanks for working on this issue ;). It's a very tricky exception.

My comment is just my opinion = I don't have any hard proofs.

I'm a bit worried it'll be quite difficult to decide what can be put into the thread and what needs to remain in "onCreate". For example, javadoc comment on WorkManager.initialize() says ```Invoke this method in {@code Application#onCreate} or a {@code ContentProvider}. Note that this method <b>must</b> be invoked in one of these two places or you risk getting a {@code NullPointerException} in {@link #getInstance(Context)}.``` I expect similar comments on the other places.

I think the time spent in the onCreate affects only the startup time of the application. I think moving it to a background thread might help with ANRs (if we have any), but I'm personally doubtful that it'll help with this particular issue. 

What we might want to consider is measure the current state (as @jd-alexander  suggested) - ideally method by method. If we find any methods which take particularly long time, we might try to put just those into a bg thread - after we carefully read their documentation and verify that it's fine to initialize them after "onCreate".

mzorz:
> I'm a bit worried it'll be quite difficult to decide what can be put into the thread and what needs to remain in "onCreate". 

Yes I hear you @malinajirka and I have the same assessment about the need to analyze the context of each method call carefully to make sure everything is called in the right moment and nothing is impacted negatively by a change like this.

> I think the time spent in the onCreate affects only the startup time of the application. I think moving it to a background thread might help with ANRs (if we have any), but I'm personally doubtful that it'll help with this particular issue.

Thanks for sharing this 👍 🙇 I'm slightly inclined to the same side, that's why I wanted to poll on what other's opinions are in this regard and see if it sounds like this could be a shot we'd want to try, and A/B test whether this change has any impact (statistically) in helping drop the amount of occurrences seen for the issue. What's your thoughts @jkmassel @khaykov?

antonis:
Hello @mzorz 👋 
Thank you for bringing this up and working on this issue 🙇 

> I'm a bit worried it'll be quite difficult to decide what can be put into the thread and what needs to remain in "onCreate". For example, javadoc comment on WorkManager.initialize()

Regarding cases like the one @malinajirka mentioned I think we may also check if we can use the [App Startup library](https://developer.android.com/topic/libraries/app-startup). I remembered this because the WorkManager is used as an [example](https://developer.android.com/topic/libraries/app-startup#implement-initializers).

mzorz:
Nice @antonis ! TIL 👍

ParaskP7:
👋  @mzorz !

Thank you for thinking about these kind of problems. 🙏 

To attempt and solve this #10494 issue might get tricky indeed.

What I am seeing here is an opportunity to do that while improving our app startup logic and time. I would have done the below steps, with the hope that while doing so, the solution with bubble up itself:
- First, add [Firebase Performance Monitoring](https://firebase.google.com/docs/perf-mon) and start monitoring the app startup time. 
- Then, start refactoring the `WordPress` application class in such a way that it is evident what is what. Looking at it now, it seems that a lot can be done to separate things, group initialization logic and create a sense of layering.
- Later on, use the suggested by @antonis new [App Startup](https://developer.android.com/topic/libraries/app-startup) Jetpack component to extract groups of initializing logic and potentially speed up the overall app startup.
- While doing so, continuously monitor and verify with [Firebase Performance Monitoring](https://firebase.google.com/docs/perf-mon) that the app startup time is indeed improving.

mzorz:
Sounds like a sound approach @ParaskP7 ! Also making things more granular and separated like this not only would make things more readable but also (hopefully) make any further problems more contained and easier to spot. 👍 taking note of it!

ParaskP7:
> Also making things more granular and separated like this not only would make things more readable but also (hopefully) make any further problems more contained and easier to spot.

💯 x 💯 !

jkmassel:
FWIW the crash logging stuff should be fine with this approach, though some stack traces _may_ change for existing crashes, which would cause them to be replaced in Sentry. Not a huge deal, but figured I'd note it.

mzorz:
I won't be conducting this experiment further so I'm closing this draft PR - will update the issue's comments directing to this interesting conversation, I think all the comments here are of great value to actually lay down a proper solution, with the steps depicted by @ParaskP7 being of particular help as a blueprint to it 👍 

Thank you everyone! ❤️

ParaskP7:
👋 @mzorz !

I recently read this article on how Dropbox managed to speed up app startup by 30%, wanted to include the link there for future reference.

How we sped up Dropbox Android app startup by 30%: https://dropbox.tech/mobile/how-we-sped-up-dropbox-android-app-startup-by-30-

Spoiler Alert:
```
Firebase Performance Library

Unfortunately, Firebase Performance library also comes with some hidden costs. 
Among them are an expensive initialization process and as a result a significant 
increase in the build times. In our debugging, we discovered that Firebase suite 
initialization was seven times longer when Firebase Performance tool was enabled. 
To fix the performance issue, we chose to remove the Firebase Performance tool 
from the Android Dropbox application.
```

PR #13612, opened 2020-12-16
[Gutenberg] Audio block -  Consolidated & Normal Media Picker Integration

jd-alexander:
`gutenberg`: https://github.com/WordPress/gutenberg/pull/27401
`gutenberg-mobile` PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/2854
`WPiOS` PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/15420
`FluxCAndroid` https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1823

## Solution
This solution seeks to address the ability to add an audio file to the Audio Block from the device or WordPress media library utilizing the media picker components that are present within the codebase.[ A similar implementation was created for the File Block.](https://github.com/wordpress-mobile/WordPress-Android/pull/13413) This PR integrates the menu options from the Audio block with the consolidated media picker.
The primary changes are:

1.  introducing a `SINGLE_AUDIO_FILE_PICKER` type in the `MediaBrowserType` so the behavior that's specific to the Audio Block can be constrained by this picker type. 

2. The GutenbergEditor's WP Media Library `AudioButtonClicked` behavior was wired to the WP Media Library and local file source behavior that opens the pickers.

3. `showFilePicker` was duplicated and a  `showAudioFilePicker` variant was created to manage the device file picker for audio. 

### Prerequisite For All Testing

Ensure you are utilizing a site that is on a premium plan as free plans aren't able to upload audio files. 

To test:

####   Picking local files. 

1. Create a new post. 
2. Add an Audio Block. 
3. Press Choose Audio. 
4. Select “Choose from device”. 
5. The File Picker should be visible. 
6. You should be able to pick an audio file. ( No other file type should be present.) 
7. You should see the "Audio Player goes here" comment with upload progress.
8. You can preview the post and download the file for further validation.

Select file | See Picker 
--------           |-------
  <kbd></kbd>  | <kbd></kbd>
    
  ### Normal Media Picker
  
  ####   Picking local files. 
  
1. Create a new post. 
2. Add an Audio Block. 
3. Press Choose Audio. 
4. Select “Choose from device”. 
5. The local file picker should be visible. 
6. You should be able to pick an audio file. ( No other file type should be present.) 
7. You should see the "Audio Player goes here" comment with upload progress.
8. You can preview the post and download the file for further validation.

Select file | See Picker 
--------           |-------
<kbd></kbd>  | <kbd></kbd>
  
 ----------------------

#### Picking WP media picker files. 

1. Create a new post. 
2. Add an Audio Block. 
3. Press Choose Audio. 
4. Select “WordPress Media Library”
5. Select an audio file. ( No other file type should be present.) 
6. You should see the "Audio Player goes here" comment. 
7. You can preview the post and download the file for further validation.

Select Option | See Picker 
--------           |-------
  <kbd></kbd>  | <kbd></kbd>
  
   

#### Picking WP media picker files. 

1. Create a new post. 
2. Add an Audio Block. 
3. Press Choose Audio. 
4. Select “WordPress Media Library”
5. Select an audio file. ( No other file type should be present.) 
6. You should see the "Audio Player goes here" comment. 
7. You can preview the post and download the file for further validation.

Select Option | See Picker 
--------           |-------
  <kbd></kbd>  | <kbd></kbd>

-----------

### File Block - Consolidated and Normal Media Picker 

Verify that the File Block's behavior of being able to select any type of file either locally or from the WordPress Media Library is intact and no regressions have occurred since most of its behavior is shared with the Audio block. 

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/photopicker/MediaPickerLauncher.kt
```diff
@@ -212,6 +212,32 @@ class MediaPickerLauncher @Inject constructor(
         }
     }
 
+    fun showAudioFilePicker(activity: Activity, canMultiselect: Boolean = false) {
+        if (consolidatedMediaPickerFeatureConfig.isEnabled()) {
```
jd-alexander, line 214: I am not sure if this approach is fine. What I did here was I duplicated the `showFilePicker` method because I need the `allowTypes` to be different for the file and audio pickers. At first, I thought of making this property a default one in the method params but this doesn't work with Java. Is this behavior okay? Or should I do some form of method override? Let me know.

jd-alexander: Also, I didn't bother to bring the Normal Picker behavior up to speed here because it seems like the consolidated media picker will be the only one in use for the near future so I thought I would waste time building it out and there's a high possibility we won't do a rollback. Was that the right judgment call? ? I did make modifications in the past to support the File Block's file selection behavior on the normal and consolidated picker hence my question.

develric: What I would consider doing here is creating a private fun that incorporates the common code and that accepts `(activity, canMultiselect, allowedTypes)` then the publicly exposed methods (showFilePicker, showAudioFilePicker) can call this private fun passing the desired allowedTypes (I would not expose the allowedTypes as public fun params though so the calling function should not worry about it). This in a way also ease the second item:

>Also, I didn't bother to bring the Normal Picker behavior up to speed here because it seems like the consolidated media picker will be the only one in use for the near future so I thought I would waste time building it out and there's a high possibility we won't do a rollback.

I think the rationale here was as you mention to allow an easy rollback (doing it by remote) in case of any issues with the consolidated media picker. To allow this we need to have the standard media picker fallback in place. I agree it's pretty unlikely we should rollback since the media picker has been released in 16.2 already, but until we do not remove this `duplicated` approach (probably in other couple of release or so) I would keep this consistent. wdyt?

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
```diff
@@ -544,6 +557,26 @@ public void onActivityResult(int requestCode, int resultCode, @Nullable Intent d
...
+                    "Failed to initialize other media options because the activity is null");
+            return otherMediaOptions;
+        }
+
+        String packageName = activity.getApplication().getPackageName();
+
+        int chooseAudioFileResourceId =
+                getResources().getIdentifier("photo_picker_choose_file", "string", packageName);
+
+        otherMediaOptions.add(new MediaOption(MEDIA_SOURCE_AUDIO_FILE, getString(chooseAudioFileResourceId)));
```
develric: When it comes to GB, GlueCode etc I'm still on the learning path 😄  , so not sure the following is worth or can have any constraint, but was wondering given the only difference seems to be the `MEDIA_SOURCE_FILE` vs `MEDIA_SOURCE_AUDIO_FILE` parameter, should we try to extract here also the common code with `initOtherMediaFileOptions`? wdyt?

develric (commented):
Hey @jd-alexander 👋 ! I checked the code and it looks good overall. I left a couple of comments for you to review. Let me know wdyt 🙇 .

This said, if I'm not wrong I see the Audio block is not in the APK from this PR; I tried to build the GB from sources adding `wp.BUILD_GUTENBERG_FROM_SOURCE=true` to my gradle.properties but I think I have some missing piece in my local setup to build GB locally (getting every kind of errors 😄 ). I think I need to spend some time to get my setup good for local GB building or maybe there is another way we can have the CI built APK have audio block enabled? Let's sync when you come online on this in case I'm missing anything, thanks 🙇‍♂️..

guarani:
Bumping the milestone in light of https://github.com/wordpress-mobile/gutenberg-mobile/pull/2854#issuecomment-754165653

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
jd-alexander: That sounds like a good idea @develric Thanks! I am going to look into doing this.

WordPress/src/main/java/org/wordpress/android/ui/photopicker/MediaPickerLauncher.kt
jd-alexander: This is great insight @develric I am going to do this implementation and then share my findings with you.

jd-alexander:
I created a FluxC PR that supports this PR with the ability to fetch the [audio-based MIME types](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1823).

WordPress/src/main/java/org/wordpress/android/ui/photopicker/MediaPickerLauncher.kt
jd-alexander: Done ff71990 3ca76e1

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
jd-alexander: Done e4039c0 c724657

jd-alexander:
> Hey @jd-alexander 👋 ! I checked the code and it looks good overall. I left a couple of comments for you to review. Let me know wdyt 🙇 .

Thanks for such a thorough review @develric 🙇 I implemented all the changes that you requested and did minor refactors in the same code paths to simplify code and increase code reusability.

> This said, if I'm not wrong I see the Audio block is not in the APK from this PR; I tried to build the GB from sources adding wp.BUILD_GUTENBERG_FROM_SOURCE=true to my gradle.properties but I think I have some missing piece in my local setup to build GB locally (getting every kind of errors 😄 ). I think I need to spend some time to get my setup good for local GB building or maybe there is another way we can have the CI built APK have audio block enabled? Let's sync when you come online on this in case I'm missing anything, thanks 🙇‍♂️..

Good question! So when you set your project to build from source you have to go to the `libs` directory and then to the `gutenberg-mobile` directory and run `npm install` and then run `npm run start` to start the metro server. 

When you go to the editor you will see that it loaded the Javascript files from the metro server and then you will see the Audio Block. 

Another approach is to [use the APK that's generated on this PR. ](https://github.com/wordpress-mobile/WordPress-Android/pull/13612#issuecomment-745829501)

The only caveat of this approach is you won't be able to switch between the Normal and Consolidated Media Picker. However, if you want to do that with this PR, you can make a commit turning off the consolidated media picker as the default and the APK will be generated.

[@Feature(CONSOLIDATED_MEDIA_PICKER_REMOTE_FIELD, false)](https://github.com/wordpress-mobile/WordPress-Android/blob/784496e7cad772385aecc3a7578de03d2bd199ba/WordPress/src/main/java/org/wordpress/android/util/config/ConsolidatedMediaPickerFeatureConfig.kt)

 Once you have done testing you could reverse this commit.

WordPress/src/main/java/org/wordpress/android/util/WPMediaUtils.java
```diff
@@ -256,34 +263,40 @@ public static void launchChooserWithContext(
...
+        return prepareIntent(context, multiSelect, Intent.ACTION_GET_CONTENT, "image/*",
+                new MimeTypes().getImageTypesOnly(), R.string.pick_photo);
+    }
+
     private static Intent prepareVideoLibraryIntent(Context context, boolean multiSelect) {
-        Intent intent = new Intent(Intent.ACTION_GET_CONTENT);
-        intent.setType("video/*");
-        intent.putExtra(Intent.EXTRA_MIME_TYPES, new MimeTypes().getVideoTypesOnly());
-        if (multiSelect) {
-            intent.putExtra(Intent.EXTRA_ALLOW_MULTIPLE, true);
```
jd-alexander, line 264: This line in these intent methods was not getting the `multiSelect` param from the function so I modified it in the refactor. I checked and the behavior is the same. Let me know what you think as well.

WordPress/src/main/java/org/wordpress/android/ui/media/MediaBrowserActivity.java
```diff
@@ -710,7 +713,8 @@ public void onMediaItemSelected(int localMediaId, boolean isLongClick) {
                 || mBrowserType.isPicker() && isLongClick) {
             showMediaSettings(media);
         } else if ((mBrowserType.isSingleImagePicker() || mBrowserType.isSingleMediaPicker() || mBrowserType
-                .isSingleFilePicker()) && !isLongClick) {
+                .isSingleFilePicker()) || mBrowserType
+                .isSingleAudioFilePicker() && !isLongClick) {
```
develric: I suspect the parenthesis here are not what we want? I would put them to group all the `||` before the `&& !isLongClick` . wdyt?

WordPress/src/main/java/org/wordpress/android/util/WPMediaUtils.java
```diff
@@ -256,34 +263,40 @@ public static void launchChooserWithContext(
...
+    }
+
+    private static Intent prepareIntent(Context context, boolean multiSelect, String action, String intentType,
+                                        String[] mimeTypes, @StringRes int title) {
+        Intent intent = new Intent(action);
+        intent.setType(intentType);
+        intent.putExtra(Intent.EXTRA_MIME_TYPES, mimeTypes);
         if (multiSelect) {
-            intent.putExtra(Intent.EXTRA_ALLOW_MULTIPLE, true);
+            intent.putExtra(Intent.EXTRA_ALLOW_MULTIPLE, multiSelect);
```
develric: tbh I would rather keep the `true`, it's like more obvious we actually want the only possible value to be true (in fact we check for `if (multiselect)` and we don't want to have a false value there (but rather no value at all) when we don't want to multiselect. wdyt?

develric (commented):
Thanks for providing an APK since the ci built APK seems to not contain the audio block for me (maybe some flag we should take care of, ci built APK are really handy when reviewing 😄  ). 

Code looks good overall, good job 👍 . I left a couple of comment inline for your review 🙇 . I also tested using your APK. Choosing from WP Media lib works as expected both with new and old media picker, but I found the following. Let me know 😊

- unless the apk did not rebuild correctly or I'm missing something, I still cannot pick mp3 and wav audio files from the device internal storage (like mp3 and wav are grayed out); ogg seems to work. This is true in GB and media picker also. It can be a different task from this PR, but actually for the Media Lib I don't remember which was the expected behaviour (cc @planarvoid ), should we have a dedicated `Choose audio from device` menu or leverage `Choose file from device`? @jd-alexander let me know if you can reproduce or you need me to make some additional check for that 🙇‍♂️. (I was using a Business plan test site if this matters)
- don't think it's related to this PR but I'm mentioning to cross check with you. I think currently I cannot play the uploaded audio file from inside the editor right? So I tried to open a preview of the post I was editing and I get the audio player with the file I uploaded. If I then try to play that file and exit the preview, the audio is still playing and I need to kill the app to stop it. If this happens to you also maybe we should open a GH issue for it.

PS: approved the fluxc companion PR if you want to tag and relink it here 👍 .

jd-alexander:
> Thanks for providing an APK since the ci built APK seems to not contain the audio block for me (maybe some flag we should take care of, ci built APK are really handy when reviewing 😄 ).

No problem 🙇 

> Code looks good overall, good job 👍 . I left a couple of comment inline for your review 🙇 . I also tested using your APK. Choosing from WP Media lib works as expected both with new and old media picker, but I found the following. Let me know 😊

I appreciate it :) 

> unless the apk did not rebuild correctly or I'm missing something, I still cannot pick mp3 and wav audio files from the device internal storage (like mp3 and wav are grayed out); ogg seems to work. This is true in GB and media picker also. It can be a different task from this PR,
This is interesting because I am not experiencing this issue on Google Pixel 3 XL. Below is a gif utilizing all these file types. 

Consolidated Media Picker | Normal Picker 
--------|-------
<kbd></kbd>        |       <kbd></kbd>

Which file app are you utilizing? Could you create a video recording of your screen so I can see exactly what is happening?
The File Manager apps out there have their own means of interpreting the intent and the MIME types supported so there's a possibility that some may not do this as well as we would like. This hypothesis would have to be tested. I am working on a way to test a popular file app to see what the results are.  

>  but actually for the Media Lib I don't remember which was the expected behaviour (cc @planarvoid ), should we have a dedicated Choose audio from device menu or leverage Choose file from device? @jd-alexander let me know if you can reproduce or you need me to make some additional check for that 🙇‍♂️. (I was using a Business plan test site if this matters)

Yes, if you could reproduce this it would be great as I am learning about the consolidated media picker project as I go through the code. 

> don't think it's related to this PR but I'm mentioning to cross check with you. I think currently I cannot play the uploaded audio file from inside the editor right? 

Good question! This behavior will probably be in an updated version of the Audio block. 

> So I tried to open a preview of the post I was editing and I get the audio player with the file I uploaded. If I then try to play that file and exit the preview, the audio is still playing and I need to kill the app to stop it. If this happens to you also maybe we should open a GH issue for it.

I am going to check this out. Thanks for reporting! 

> PS: approved the fluxc companion PR if you want to tag and relink it here 👍 .

Thanks much!

WordPress/src/main/java/org/wordpress/android/util/WPMediaUtils.java
jd-alexander: Oh yes, good spotting that!

WordPress/src/main/java/org/wordpress/android/ui/media/MediaBrowserActivity.java
jd-alexander: Sounds about right. Thanks 3309187

WordPress/src/main/java/org/wordpress/android/util/WPMediaUtils.java
jd-alexander: Good catch. Resolved here 1e5aa6d

develric:
Hey @jd-alexander 👋 . I tried to dig a bit forcing some behaviour with the debugger, what I can say is:
- Using the Pixel 3 Emu API 30 the system picker seems to work as expected for me
- Using the default system picker in my OnePlus 7T I'm getting the issues I reported above. So, as you were suggesting, it seems something related to specific sistem picker (but I'm using the OnePlus stock 🤷 ). 

Maybe we should further dig the ways we are opening the system picker (at least for audio) also on other devices. I see we are setting the following:

- type: \*/\*
- EXTRA_MIME_TYPES: our audio supported mime types (like audio/mpeg, audio/ogg etc)

so basically relaying on the EXTRA_MIME_TYPES to filter out the unwanted mime types.

In my case with the above I get some audio files grayed out but cannot understand with which logic (maybe something in the files but they are basically the test files we are using in the Pixel Emu also)

<p align=center><image src="https://user-images.githubusercontent.com/47797566/104661731-74a02680-56c9-11eb-9be8-fbeec60a3159.png" width=300/></p>

I get the same results if I keep the EXTRA_MIME_TYPES but set type as `audio/*`. If I set type as `audio/*` and do not set the EXTRA_MIME_TYPES I actually get something that seems more correct

<p align=center><image src="https://user-images.githubusercontent.com/47797566/104661912-e9736080-56c9-11eb-8abd-cf1cb62f087c.png" width=300/></p>

but in this case I guess we are potentially not filtering out not supported audio mime types that's why we set the EXTRA_MIME_TYPES.

Googling quickly I see that there can be also the possibility to put multiple sets in the type like `audio/mpeg|audio/ogg` but running out of time and couldn't try it. 

Unfortunately it seems I still cannot build GB with my setup, so not easy to check things, but since we have still some time before the end of the sprint maybe we should spend a bit more time to check how we can broaden device support when coming to picking from system picker (at least for audio). For example in this PR we are using type \*/\* but thinking to it maybe an audio/* would be more correct (at least in the editor)? Are we releasing the audio block with this PR or this is an interstitial PR? We can probably address this in a separate PR but let's keep in touch on slack to decide how to proceed 🙇

jd-alexander:
Thanks for digging into this @develric and I agree that we will have to play with the params below that are being passed to the intent to achieve behavior that's compatible with all file pickers. 

```
intent.setType(intentType);
intent.putExtra(Intent.EXTRA_MIME_TYPES, mimeTypes);
```

> Are we releasing the audio block with this PR or this is an interstitial PR? We can probably address this in a separate PR but let's keep in touch on slack to decide how to proceed 🙇

Yes, this is an intermediary PR, so there are several other PRs depending on it that need to be merged in so it would be good to get these PRs merged in, and then we can create an issue and address the improvements we need separately. Thanks much for the thorough review and actually spotting this on a different device configuration.

develric (approved):
Thanks for all the work and improvements here @jd-alexander 🙇 . Let's make sure to open a ticket to further dig those intent parameters for system pickers support before final release of audio block (let me know if you want me to open it). This said LGTM and good job 👍 . :shipit:

jd-alexander:
> Thanks for all the work and improvements here @jd-alexander 🙇 . Let's make sure to open a ticket to further dig those intent parameters for system pickers support before final release of audio block (let me know if you want me to open it). This said LGTM and good job 👍 . :shipit:

Thanks again @develric I created an issue to track [the intent params issue here](https://github.com/wordpress-mobile/WordPress-Android/issues/13792) so we can continue looking into it.

jd-alexander:
This PR was merged before the gutenberg and gutenberg mobile PRs were approved, so a [revert PR](https://github.com/wordpress-mobile/WordPress-Android/pull/13805) was created to remove the changes from `develop` and a [new PR](https://github.com/wordpress-mobile/WordPress-Android/pull/13808) was reopened with the changes pending another review.

PR #13613, opened 2020-12-16
Track Reader Discover Content Presented

### LINKED ISSUE #13609: Reader: Track Cards API response
### Goal

The goal is to track when the Reader Cards API response has arrived. We want to make a funnel to check how many users are leaving the Reader _before_ the response returns.

In WPiOS [this was implemented](https://github.com/wordpress-mobile/WordPress-iOS/pull/15183) by triggering an event when the cards appear: `reader_discover_content_presented`.

#### Appearing

1. Open Discover
2. Wait for the cards to appear
3. Check that `reader_discover_content_presented` is fired

#### Leaving the screen before it appears

1. Open Discover
2. Immediately switch to other section (like Following)
3. Check that `reader_discover_content_presented` is *not* fired

ParaskP7:
Fixes #13609
iOS PR: [WordPress-iOS#15183](https://github.com/wordpress-mobile/WordPress-iOS/pull/15183)

This PR tracks if:

1) The API response has arrived.
2) The user is still waiting for the content/error to be presented (aka didn't left the `Discover` screen).

The event fired is intended to be used in a funnel, to check how many users are leaving the `Discover` screen due to the API taking too long to respond.

### To test

Launch the app:

1) Open `Discover` screen.
1) Wait for the content/error to be presented.
1) Check that `reader_discover_content_presented` is fired.

Launch the app again (to clear cache):

1) Open `Discover` screen.
1) Immediately switch to another section (like the `Following` screen).
1) Check that `reader_discover_content_presented` is not fired.

PR submission checklist:


malinajirka:
Thanks @ParaskP7!

I'll delay the review until [this discussion](https://github.com/wordpress-mobile/WordPress-iOS/pull/15183) is resolved. The implementation might change significantly as a result of the discussion. Please let me know when it's ready for review. Thanks!

leandroalonso:
Hi @ParaskP7, I'll answer the questions you made in https://github.com/wordpress-mobile/WordPress-iOS/pull/15183 here:

> What exactly do we want to track?

The % of users that leave the Reader screen before the content appears.

> What conclusions do we want to derive from the data and how?

We want to check if the API being slow is affecting Reader usage in the Discover section. The how is by using two events in a funnel.

> Are we only interested in the first response or even load more? PS: In my solution, I am assuming that we are only interested in the first response.

Only in the first. And this applies only to Discover.

> If we show data from cache and send a request for more up to date data, should we track the event or not? PS: In my solution I am assuming yes.

Does the Discover section in Android show cached data? Basically, in iOS this cache will be there during the current session. If the users closes the app and reopen again, we will not display data from any cache and request it again.

Anyway, to summarize: we're interested to see if the user is waiting for the content to appear when nothing is being presented.

> If the user leaves before the request finishes and comes back later, should we track the event or not? PS: In my solution I am assuming yes, but this is a tricky question. For example, if the user leaves the screen and comes back again, quickly, before the data is fetched then the tracking will still occur.

I guess this is not a problem. :)

ParaskP7:
> I'll delay the review until this discussion is resolved. The implementation might change significantly as a result of the discussion. Please let me know when it's ready for review. Thanks!

Thank you @malinajirka ! 🌟 

You are right. Also, as you mentioned this solution is currently incomplete as it will track all events, whenever the data get updates, post upload finishes, user clicks on bookmark action, like action, etc. In order for this solution to become complete, even naively, more logic (state) needs to be added.

ParaskP7:
Thanks for the quick response @leandroalonso !

> The % of users that leave the Reader screen before the content appears.

👍 

> We want to check if the API being slow is affecting Reader usage in the Discover section. The how is by using two events in a funnel.

👍 

> Only in the first. And this applies only to Discover.

👍 

> Does the Discover section in Android show cached data? Basically, in iOS this cache will be there during the current session. If the users closes the app and reopen again, we will not display data from any cache and request it again.

Maybe @malinajirka can jump in and help here?

> Anyway, to summarize: we're interested to see if the user is waiting for the content to appear when nothing is being presented.

👍 

> I guess this is not a problem. :)

If we have decided to not have it as a problem, then I am okay with it as well. ☺️

ParaskP7:
> You are right. Also, as you mentioned this solution is currently incomplete as it will track all events, whenever the data get updates, post upload finishes, user clicks on bookmark action, like action, etc. In order for this solution to become complete, even naively, more logic (state) needs to be added.

@malinajirka , I updated the solution with a **first naive approach** to solving this tracking problem. Let me know your thoughts on it. I don't quite like it as much but since I am not sure how much time to invest in it, I am not sure if I should keep going. As such, I'll pause until further notice.

PS: This solution is a bit different to iOS, since they are tracking again the `contented presented` on a pull-to-refresh action. After discussing it with @leandroalonso , he thinks this should not affect the result much, as such it is negligible.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
```diff
@@ -426,6 +427,19 @@ class ReaderDiscoverViewModel @Inject constructor(
         }
     }
 
+    fun updateScreenVisibility(isScreenVisible: Boolean) {
+        this.isScreenVisible = isScreenVisible
+    }
+
+    private fun trackContentPresentedOnce() {
+        if (!isContentPresentedTrackedOnce) {
+            isContentPresentedTrackedOnce = true
```
malinajirka, line 436: ❓ Do we intentionally update the `isContentPresentedTrackedOnce` even when the screen is not visible?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
```diff
@@ -145,6 +144,7 @@ class ReaderDiscoverViewModel @Inject constructor(
                         }
                     }
                 }
+                trackContentPresentedOnce()
```
malinajirka, line 147: ❓ This will be tracked even when we load data from cache. I believe it won't help us answer `We want to check if the API being slow is affecting Reader usage in the Discover section. The how is by using two events in a funnel.`. Thoughts @ParaskP7 @leandroalonso ?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
```diff
@@ -258,12 +258,13 @@ class ReaderDiscoverViewModel @Inject constructor(
                     )
                 }
             }
+            trackContentPresentedOnce()
```
malinajirka, line 261: ❓ Do we intentionally track even errors? AFAIU it won't help us answer `The % of users that leave the Reader screen before the content appears.` -> assuming error isn't considered as content.

malinajirka (commented):
Thanks @ParaskP7!

Quite tricky task :P.  I have left some comments, but it looks good overall.

malinajirka:
The more I think about it the more I'm unsure if the funnel make sense on Android. Since the only time we don't show local content is right after installation. In all the other cases, data from cache are shown. Wdyt @leandroalonso ?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
ParaskP7: Yes, I am doing that intentionally. What I try to achieve is that when the user enters the screen and then leaves right after, the event gets triggered and the content arrives, even if they are not looking, so I count it as track and miss, since I don't want to count it on any other occurrence afterwards.

Does this makes sense or I am mumbling nonsense here? 😊

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
ParaskP7: I am not sure about this, maybe in iOS, because they have no cache data.

But, in our case, what I want to achieve is based on the content shown, not the data arriving. So, if the content is shown, either through the API or cache, I am okay with that, as the result will be the same for the user, their content being shown. On the other hand, if the content is not shown, because of the API being slow, or maybe because of cache problems (doubt this will happen), then I want to track that. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
ParaskP7: I tracked errors intentionally for two reasons:
1) iOS was tracking errors as well.
2) I was trying to track all state changes, except for `LoadingState`. So, both, `ContentUiState` and `EmptyUiState`.

ParaskP7:
> Quite tricky task :P. I have left some comments.

Indeed, thanks for the comments and for thinking about the problem that hard!

> ..., but it looks good overall

I am not sure how good it looks, but thanks for being kind. 😞 

> The more I think about it the more I'm unsure if the funnel make sense on Android. Since the only time we don't show local content is right after installation. In all the other cases, data from cache are shown. Wdyt @leandroalonso ?

This is a very interesting question. Indeed showing data from cache seems fast. Maybe on less performant devices this will be a problem as well, but we are not actually trying to tackle this anyway.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
leandroalonso: @malinajirka we track that because even if the API fails, the user is still there. Perhaps the name of the event is kinda misleading, if you both think that we should change the name, let's do it. :)

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
leandroalonso: @ParaskP7 I'll only trigger that if a request has been made. Otherwise, as @malinajirka mentioned we won't be able to answer the main question.

So:

1. Request triggered -> Content rendered: fire the event
2. Content rendered: don't fire

Again, just like mentioned below, if you both think the event name is misleading, we can change it. :)

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
malinajirka: Ahh, I see. In that case, I'd consider renaming the variable. It feels a bit confusing that `isContentPresentedTrackedOnce == true` but the event was never actually tracked. Wdyt?

Perhaps something like "isFirstTimeContentPresented"?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
malinajirka: What you are proposing @leandroalonso is quite difficult to do in our implementation and I'm not sure it's worth it.

- ViewModel has no idea where the data come from (cache or server)
- DataProvider and API layer have no idea what is on the UI

Moreover, this will come down to -> track this event only after the first installation if the user didn't leave before the content was shown. Is this really what we want to track?

malinajirka:
See [this comment first](https://github.com/wordpress-mobile/WordPress-Android/pull/13613#discussion_r544855346). I personally don't think this event is worth implementing on Android. I don't think we'll be able to derive any usable data from it. Unless I'm missing something it'll break our layer separation. This wouldn't be a big deal per se, but I'm a bit worried it'll introduce a precedence and future devs might feel tempted to follow the same pattern for other things than this track event.

Moreover, the question we are trying to answer feels platform independent. We can come to two conclusions on iOS
1. The slow API is not causing any issues
2. Users tend to leave because of the slow API

(sidenote: I still don't fully understand how are we going to separate users who leave because they are not interested in Discover and those who leave because the API is too slow)
If the conclusion is `1` we don't need to make any changes on either platform. If the conclusion is `2` the only affected flow is first usage after installation on Android and we might want to look into it. To summarize, I think we can come to the same conclusions even without data from Android. Thoughts @leandroalonso @ParaskP7 ?

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
ParaskP7: Yeap, I was thinking about it as well, I like your suggestion, will do! 🌟 

PS: The way I was thinking about it was since this variable is used within the `trackContentPresentedOnce()` function, which means that the caller wanted to track the content presented, then I count it as such, as once, whether that actually did the tracking based on the screen being visible is an implementation detail down the line for the caller code to not worry about it. I agree it is tricky...

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
ParaskP7: I agree with Jirka @leandroalonso , in the Android case, the way it is implemented, this layer is agnostic of where the data came from, and for a good reason. To me, since what we want to understand, is when the user leaves the screen, before the content gets presented to them, we shouldn't care how this is done on the data layer, we should only care about the end result, content being presented or not. Thus, I am leaving the API out of it, being an implementation detail. Wdyt?

ParaskP7:
> I still don't fully understand how are we going to separate users who leave because they are not interested in Discover and those who leave because the API is too slow

This is a really good point! I don't think that with the current implementation, both in iOS and Android, we are able to separate users who leave because they are not interested in Discover.

> I think we can come to the same conclusions even without data from Android.

Indeed, I will agree on that too. If the API is slow, then we don't need both platforms to verify that, the user base is large enough to derive that from one platform. I am only skeptical on the fact that Android users, which have access to cheaper devices, and might be living in areas where the connectivity is worst, due to the economics of the world, might showcase a more significant API slowness than when running this experiment on iOS users. Wdyt?

Having said the above and considering the fact that Android can only contribute to the result by verifying API slowness during first install, I also believe it might not be worth it. As a matter of fact, maybe those results might contribute to false positive!

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverViewModel.kt
ParaskP7: PS: I am not going to do any changes until we decide to move forward with this tracking on Android. If we do, I'll make the change, else I'll just close the PR and delete the branch.

leandroalonso:
@malinajirka @ParaskP7 

> Moreover, the question we are trying to answer feels platform independent.

I won't assume that. We've seen very different user behaviors in Android and iOS, the same might apply here. @ParaskP7 summarized very well on his comment, the points he made are more than enough to create differencies between iOS and Android.

> Unless I'm missing something it'll break our layer separation.

I can see, that makes sense.

@malinajirka about your conclusions, just to make clear: this data is not for us. It's for the API team. If they tackle a project to improve the speed of the API, a funnel that shows if the user is waiting or no it's going to be useful for them (to measure the impact). It's more for them than for us. Unfortunately, there's no way this can be tracked on the backend side.

> Having said the above and considering the fact that Android can only contribute to the result by verifying API slowness during first install, I also believe it might not be worth it. 

This is a really good point. Given Android uses a cache, most of the time returning users will see content already there. And this affects this data, it's completely different from iOS (every new session in Discover will use fresh API data).

Let's pause this for now. I'll try to think something and if I come up with anything I'll let you know. :)

ParaskP7:
@leandroalonso thanks for your conclusion, I agree, I'll now close the PR until further notice.

@malinajirka many thanks for the review and for your so valuable input. Kudos for not letting it slip to `develop`.

This was a hard problem to solve, involving major platform differences and non-trivial tracking logic. Thank you both for putting on your analytical thinking hat and deriving to the most effective result!

❤️

PR #13615, opened 2020-12-16
Issue/13268 filters 2

malinajirka:
Parent #13268

This PR
- implements cancel actions for both activity log filters
- updates filter labels based on selected filters
      - "Activity Type" when the filter is empty
      - "Activity Type (N)" when there is more than 1 activity selected
      - "[ActivityName]" when there is exactly 1 activity selected (this is implemented just partially, activity type id is used as name)
      -  "Date Range" when filter is empty
      - "Nov 14 - 17" when both selected dates are within one month
      - "Nov 14 - Dec 18" when dates are not within one month
      - "Nov 15, 2020 - Jan 1, 2021" when dates are not within on year
      

Screenshot 2020-12-16 at 14 12 30

Notes:
- The actual filtering is not implemented yet
- @osullivanchris All design feedback related to the filter bar is welcome (including scrolling behavior)

To test:
- Verify that all the states mentioned above work as expected
- Verify both clear actions work as expected

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
@@ -208,15 +208,15 @@ class ActivityLogListFragment : Fragment() {
 
     private fun updateFilters(uiState: FiltersShown) {
         with(requireActivity().date_range_picker) {
-            this.text = uiHelpers.getTextOfUiString(requireContext(), uiState.dateRangeLabel)
-            this.isCloseIconVisible = uiState.clearDateRangeFilterClicked != null
-            this.setOnCloseIconClickListener { uiState.clearDateRangeFilterClicked?.invoke() }
+            text = uiHelpers.getTextOfUiString(requireContext(), uiState.dateRangeLabel)
+            isCloseIconVisible = uiState.clearDateRangeFilterClicked != null
+            setOnCloseIconClickListener { uiState.clearDateRangeFilterClicked?.invoke() }
         }
 
         with(requireActivity().activity_type_filter) {
-            this.text = uiHelpers.getTextOfUiString(requireContext(), uiState.activityTypeLabel)
-            this.isCloseIconVisible = uiState.clearActivityTypeFilterClicked != null
-            this.setOnCloseIconClickListener { uiState.clearActivityTypeFilterClicked?.invoke() }
+            text = uiHelpers.getTextOfUiString(requireContext(), uiState.activityTypeLabel)
+            isCloseIconVisible = uiState.clearActivityTypeFilterClicked != null
+            setOnCloseIconClickListener { uiState.clearActivityTypeFilterClicked?.invoke() }
         }
     }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
@@ -209,14 +209,14 @@ class ActivityLogListFragment : Fragment() {
     private fun updateFilters(uiState: FiltersShown) {
         with(requireActivity().date_range_picker) {
             text = uiHelpers.getTextOfUiString(requireContext(), uiState.dateRangeLabel)
-            isCloseIconVisible = uiState.clearDateRangeFilterClicked != null
-            setOnCloseIconClickListener { uiState.clearDateRangeFilterClicked?.invoke() }
+            isCloseIconVisible = uiState.onClearDateRangeFilterClicked != null
+            setOnCloseIconClickListener { uiState.onClearDateRangeFilterClicked?.invoke() }
         }
 
         with(requireActivity().activity_type_filter) {
             text = uiHelpers.getTextOfUiString(requireContext(), uiState.activityTypeLabel)
-            isCloseIconVisible = uiState.clearActivityTypeFilterClicked != null
-            setOnCloseIconClickListener { uiState.clearActivityTypeFilterClicked?.invoke() }
+            isCloseIconVisible = uiState.onClearActivityTypeFilterClicked != null
+            setOnCloseIconClickListener { uiState.onClearActivityTypeFilterClicked?.invoke() }
         }
     }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
@@ -137,6 +137,14 @@ class ActivityLogListFragment : Fragment() {
                         requireContext(),
                         uiState.activityTypeLabel
                 )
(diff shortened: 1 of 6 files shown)


WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
```diff
@@ -215,6 +206,20 @@ class ActivityLogListFragment : Fragment() {
...
+            this.isCloseIconVisible = uiState.clearDateRangeFilterClicked != null
+            this.setOnCloseIconClickListener { uiState.clearDateRangeFilterClicked?.invoke() }
+        }
+
+        with(requireActivity().activity_type_filter) {
+            this.text = uiHelpers.getTextOfUiString(requireContext(), uiState.activityTypeLabel)
+            this.isCloseIconVisible = uiState.clearActivityTypeFilterClicked != null
+            this.setOnCloseIconClickListener { uiState.clearActivityTypeFilterClicked?.invoke() }
+        }
+    }
```
ashiagr: Nitpick: we can omit `this` keyword.

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -384,7 +433,9 @@ class ActivityLogViewModel @Inject constructor(
 
         data class FiltersShown(
             val dateRangeLabel: UiString,
-            val activityTypeLabel: UiString
+            val activityTypeLabel: UiString,
+            val clearDateRangeFilterClicked: (() -> Unit)?,
+            val clearActivityTypeFilterClicked: (() -> Unit)?
```
ashiagr: ❓: I see all action methods prefixed with "on" in this `ViewModel` and in general within the codebase. Should we prefix `on` before these click method names?

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/utils/DateUtils.kt
```diff
@@ -49,6 +49,10 @@ class DateUtils @Inject constructor(
         )
     }
 
+    fun formatDateRange(from: Long, to: Long): String {
+        return DateUtils.formatDateRange(contextProvider.getContext(), from, to, DateUtils.FORMAT_ABBREV_MONTH)
+    }
```
ashiagr: Nitpick: Prefer using an expression body for functions with the body consisting of a single expression.
(Source: [Kotlin Coding Conventions: Function formatting](https://kotlinlang.org/docs/reference/coding-conventions.html#function-formatting))

WordPress/src/main/res/values/strings.xml
```diff
@@ -1031,6 +1031,7 @@
     <string name="activity_log_activity_type_filter_clear">Clear</string>
     <string name="activity_log_date_range_filter_label">Date range</string>
     <string name="activity_log_activity_type_filter_label">Activity type</string>
+    <string name="activity_log_activity_type_filter_active_label">Activity type (%s)</string>
```
ashiagr: Nitpick: We might want to be consistent with the "t" in "Activity type" here and in the details screen where it shows it as capitalised.




In case we decide to capitalise "t" here, we might want to capitalise "R" in "Date range".

ashiagr (commented):
Great job! Clear action and labels display work as described. Just left very minor nitpicks in the code. 

Also noticed that when we tap date range in the light mode, status bar changes to black but on tapping activity type, status bar remains white in color. Is there anything we can do about it?

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
malinajirka: 👍 Fixed in 07479bb4a4b3104b5cd1e286a6f72b124158d7d1

Note: I usually prefer to keep it to emphasise which variables/methods are related to the `with`. However, in this case it doesn't make much sense, since we use it just for value assignment.

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
malinajirka: Good call, fixed in a0eb3425bb753fbbdc79eb7c1230be7edfafb470

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/utils/DateUtils.kt
malinajirka: 👍 Fixed in 0711f4e65021594a5b500b24c709aa946528975a

WordPress/src/main/res/values/strings.xml
malinajirka: Nice catch, fixed in f3b9f2b13fb309baadea0e4816c08a02fe7efa04

malinajirka:
Thank you for the review @ashiagr! I've fixed all the issues you mentioned. It's ready for another round 🙇 

> Also noticed that when we tap date range in the light mode, status bar changes to black but on tapping activity type, status bar remains white in color. Is there anything we can do about it?

This is something I was fighting in [this PR](https://github.com/wordpress-mobile/WordPress-Android/pull/13579). I tried several different approaches and none of them worked. I created a sub-task on the parent issue. I plan to look into it later. Having said that if you have ideas how to achieve the desired behavior, please let me know.

ashiagr:
Thanks for the changes, LGTM 👍.

> I created a sub-task on the parent issue. I plan to look into it later. Having said that if you have ideas how to achieve the desired behavior, please let me know.

Adding to backlog is fine. We can look into it later, it's not a blocking issue.

I'm not merging it yet so that Chris can add his feedback.

ashiagr (approved):

osullivanchris:
Reviewed on landscape, portrait, dark mode, light mode.

Positive:
- Visually the chips look great.
- Selected chips look good too
- Clearing them works fine. 
- Shadow on scroll is what we've been doing elsewhere. This looks good. I worked with Klymentiy to hide the divider on scroll, and only have shadow. But it was quite tricky for him I think. The divider and shadow actually looks ok here.\
- Dark mode: lift on scroll makes the header grey. Per other pages. Looks good too.
- Great work on the logic for the dates (E.g. showing the year only when overlapping 2 years).

 Suggested changes/enhancements:
- divider under the app bar looks too strong on dark mode - can we use a more subtle color here?
- When there's a long string in the chips (e.g. select a date over two years and an activity type) you can scroll horizontally to see. But I think we can hide the scroll bar, it doesn't look great. WDYT?

malinajirka:
Thank you both! 🙇 

I've updated divider color and hid the horizontal scrollbar. It's ready for another round.

Screenshot 2020-12-18 at 08 40 11

osullivanchris:
> I've updated divider color and hid the horizontal scrollbar. It's ready for another round.

Looks good 👍

ashiagr (approved):
Looks good 👍.

PR #13617, opened 2020-12-16
Rename to support reuse

zwarm:
Parent #13329 

This PR renames backup_styles.xml to jetpack_styles.xml to support reuse amongst the jetpack features (backup, restore, and scan)

To test:
There is nothing to test in this PR

PR submission checklist:


ashiagr (commented):
Looks good @zwarm! We might want to rename `Backup.Download` with something generic like `Jetpack` to allow reusability. Should we do it now or later when all styles are included?

zwarm:
> Looks good @zwarm! We might want to rename `Backup.Download` with something generic like `Jetpack` to allow reusability. Should we do it now or later when all styles are included?

@ashiagr I was on the fence about doing that now or later ... I can do it right now. Hold on.
In commit 9183961 . Thanks!

ashiagr (approved):
LGTM 👍.

PR #13546, opened 2020-12-04
Fix position of Quick Start focus point in bottom sheet

khaykov:
Fixed the vertical offset of QS focus point in the bottom sheet. It will now appear right next to the title (in any locale), and not in the center, where it potentially can cover it.

[Image from Gyazo](https://gyazo.com/f325c3af2ff245a2d8c67a5198ea89b9)

To test:
- Create a new site and start Quick Start flow.
- Navigate to `Grow your audience` category and select `Publish a post` task.
- Tap on the FAB and make sure that the quick start indicator in the bottom sheet is visible next to the `Blog post` title on the first row of bottom sheet.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13624, opened 2020-12-17
Revert "Fix position of Quick Start focus point in bottom sheet"

khaykov:
Reverts wordpress-mobile/WordPress-Android#13546

Reversion in #13551 broke the layout for the Quick Start focus point, so I need to revert the fix to make it work again.

To test:
- Create a new site.
- Accept Quick Start promptly.
- Navigate to Grow Your Audience QS option.
- Select "Publish a post" task.
- Tap FAB so the bottom sheet would appear.
- Confirm that the QS focus point is correctly placed on "Blog Post" option.

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/ActionListItemViewHolder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/ActionListItemViewHolder.kt
@@ -61,7 +61,7 @@ class ActionListItemViewHolder(
                     dimen.quick_start_focus_point_size
             )
             actionRowContainer.post {
-                val verticalOffset = -focusPointSize / 2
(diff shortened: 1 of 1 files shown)


aforcier (approved):
Ah - sorry @khaykov , I hadn't realized the 'new' tag was the cause of the issue #13546 was fixing or I would have warned you when removing the 'new' tag in the next version 😬 

Confirmed the issue and that this PR returns the quick start indicator to its previous place. :shipit:

aforcier:
cc @wordpress-mobile/release-managers that it might be worth including this change if we release another RC for 16.4.

loremattei:
Thank you @aforcier! I opened a PR to bring this fix to the release branch: https://github.com/wordpress-mobile/WordPress-Android/pull/13627

PR #13625, opened 2020-12-17
Adding unread posts count to Reader bottom sheet

### LINKED ISSUE #13454: Add unread post count to Reader
We should add the count of unread posts to the Reader, similiar to how it's done in calypso:
followed

Design explorations for this are currently underway.

develric:
Fixes #13454

This adds unread posts count to the reader filter bottom sheet. Note that to align the reader subfilter to the iOS implementation we removed the selected item checkmark in this #13648 

The information about the unread posts is contained in the endpoint `read/following/mine` from v1.2 API, so we needed to upgrade the rest client from v1.1 for this call.

The following images show the implemented result both in light and dark mode.

| Light mode | Dark mode
|---|---|
| image | image |

**Note 1:** please note that the update of the counters is async so sometimes is possible you need to Pull To Refresh before to get the updated numbers in the bottom sheet

**Note 2:** to see the counters in Calypso, it is possible you need to use a a8c account

**Note 3:** currently when a post is read from the app is not marked as seen so the counters are not updated by this action on the app. There is an task that will land in a separate PR to add this enhancement in the app reader.

**Note 4:** a small deviation from design is that I kept 16dp right edge as the left side in the bottom sheet instead of the 14dp (pinging @mattmiklic for a quick check on this and other things he may notice eventually 🙇 )

<p align=center><image src="https://user-images.githubusercontent.com/47797566/103519572-0ca24280-4e76-11eb-82df-6df8ee4a8102.png" width=300/></p>

## To test
The feature is behind the `UNREAD_POSTS_COUNT` feature flag. You need to build the wasabiDebug flavor from sources or use the APK from this PR CI build remembering to activate the feature in `App Settings -> Test Feature Configuration` (IMPORTANT: remember to restart the app after changing these flags configuration)

<p align=center><image src="https://user-images.githubusercontent.com/47797566/102915052-73d8f380-4481-11eb-9fbb-8ff66def1b75.png" width=300/></p>

### Checking the feature
- Login into the app with an a8c account (in principle you can also login with a not a8c account but this step is needed to be able to compare the counters with Calypso; see Note 2 above)
- Go to the reader and check the counters are shown
- Using the same account as for the app, go to `https://wordpress.com/read` and check the counters are coherent

| WP Android app | Web reader |
|---|---|
| image | image |

- Try to publish one or more post, refresh the app and the web and check the numbers stay consistent

### Checking big numbers abbreviation
- Numbers above 1000 are abbreviated with multipliers like `k`, `M` etc...
- In AS place an `Evaluate and Log` breakpoint like below in `SiteViewHolder` Line 35 with this config `this.itemUnseenCount.text = statsUtils.toFormattedString(123456L, ONE_THOUSAND)`

<p align=center><image src="https://user-images.githubusercontent.com/47797566/102917943-5c503980-4486-11eb-8b9f-924e2c09f921.png" width=400/></p>

- Connect to the app with the debugger and check you get expected abbreviation (you can also change the number 123456L in the breakpoint to test other number ranges)

<p align=center><image src="https://user-images.githubusercontent.com/47797566/102918225-cec11980-4486-11eb-97a9-f703284c7b2f.png" width=400/></p>

### Smoke testing without the feature enabled
- Disable the feature flag in `App Settings -> Test Feature Configuration` and restart the app
- smoke test the reader and the filter bottom sheet

PR submission checklist:


mattmiklic:
This looks good to me, thanks @develric!

renanferrari:
Hey @develric, sorry for the delay on this one. I'll make sure to take a look at it by tomorrow, if no one gets to it before me :)

WordPress/src/main/java/org/wordpress/android/models/ReaderBlog.java
```diff
@@ -81,6 +82,10 @@ public static ReaderBlog fromJson(JSONObject json) {
             blog.numSubscribers = json.optInt("subscribers_count");
         }
 
+        if (json.has("unseen_count")) {
+            blog.numUnseenPosts = json.optInt("unseen_count");
```
renanferrari, line 86: Checking that the parameter exists with `has()` doesn't seem to be necessary here, as we're using `optInt()` which fallbacks to `0` by default.

renanferrari (approved):
Thanks for the changes @develric! I've tested the app and everything seems to be working as expected. Code also looks pretty good – I just left one comment below, but it's pretty minor.

develric:
Hey @renanferrari 👋 , thanks so much for the review 🙇 ! I agree with the np, since you already approved this one and I need to create a dedicated PR to change the multiplier from lower case k to upper case K I'm going to merge this one and will manage it in the new PR. Thanks again 😄

PR #13626, opened 2020-12-17
Fix a typo in a new string

loremattei:
This PR fixes a typo in a new string.
Since we're targeting the release branch, I've also updated the shadow `strings.xml` which is sent to translators. 

PR submission checklist:


zwarm (approved):
@loremattei - The design mocks don't show a period, but that definitely points to the wp-config.php file which absolutely needs a period. Thanks!!!

PR #13630, opened 2020-12-17
Add download package to jetpack.backup

zwarm:
Parent #13329 
The PR adds the `download` package to the `Jetpack.Backup` package to separate the download process from the daily backup views.

To test:
- Launch the app 
- Tap on Activity Log
- Tap on the more menu
- Tap the download backup option
- Ensure the backup download view is shown

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadListItemState.kt
```diff
@@ -1,8 +1,13 @@
...
+package org.wordpress.android.ui.jetpack.backup.download
 
 import androidx.annotation.DrawableRes
 import org.wordpress.android.R
 import org.wordpress.android.ui.jetpack.BackupAvailableItemsProvider.BackupAvailableItemType
+import org.wordpress.android.ui.jetpack.backup.download.BackupDownloadListItemState.ViewType.ACTION_BUTTON
+import org.wordpress.android.ui.jetpack.backup.download.BackupDownloadListItemState.ViewType.CHECKBOX
+import org.wordpress.android.ui.jetpack.backup.download.BackupDownloadListItemState.ViewType.DESCRIPTION
+import org.wordpress.android.ui.jetpack.backup.download.BackupDownloadListItemState.ViewType.HEADER
+import org.wordpress.android.ui.jetpack.backup.download.BackupDownloadListItemState.ViewType.SUBHEADER
```
ParaskP7, line 10: Minor (🔍 ): Probably this was automatically done due to the way your `Optimize imports` works, not sure, we can check it out if you like. In the meanwhile, can you revert this change if that wasn't intentional?

ParaskP7 (approved):
Good job @zwarm ! 🌟 

I have tested the change and checked on the updated `backup` package structure, it looks great!

I have left one minor import related comment, but approving the PR so that I don't block you.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadListItemState.kt
zwarm: This is changed in a future PR. so all is good.

PR #13631, opened 2020-12-17
Use real endpoint for activity type filter

malinajirka:
Parent issue #13268

Merge instructions
1. Update FluxC version when https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1807 gets merged
2. Merge this PR

This PR replace DummyActivityTypeProvider with the actual implementation from FluxC. Corresponding FluxC PR is [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1807). It also starts passing current DateRange filter to the fetchActivityTypes endpoint.

Note:
- It seems the server currently ignores the "after" and "before" dates. We use a different format than Calypso. However, we use this format on all the APIs. I plan to fix this in a separate PR when we decide on the best approach.

To test:
1. Make sure ActivityLogFiltersFeatureConfig is enabled
2. Open Activity Log
3. Click on Activity Type filter
4. Notice, that actual activities get fetched and shown on the UI
5. Notice each row contains number of activity log entries (eg. User (23))
6. Change language of the app to German
7. Open Activity Log
8. Click on Activity Type filter
9. Notice the Activity Types are translated
----------------------------------------
1. Make sure ActivityLogFiltersFeatureConfig is enabled
2. Make sure you are using atomic site or a jetpack site with a paid plan
3. Open Activity Log
4. Click on Activity Type filter
5. Remember what activities and numbers you see
6. Go back and open date range filter
7. Select 17-18th of December
8. Open Activity Type filter and notice the result differs from step 5

PR submission checklist:


ParaskP7:
👋 @malinajirka !

Could you please check the build on this PR, it seems that the CI is failing on it and I am not being able to build/install this branch locally as well (same failure). It has probably something to do with the `FluxC` hash version.

```
* What went wrong:
Could not determine the dependencies of task ':WordPress:compileJalapenoDebugJavaWithJavac'.
> Could not resolve all task dependencies for configuration ':WordPress:jalapenoDebugCompileClasspath'.
   > Could not find com.github.wordpress-mobile.WordPress-FluxC-Android:fluxc:d52f68e147ca4c8dd957ac6f778e442b9e112d36.
     Required by:
         project :WordPress
```

Let me know when you fix that and I'll resume my review! 🙏

malinajirka:
@ParaskP7 The PR is ready for review. Feel free to review it next week.

I added a missing test in https://github.com/wordpress-mobile/WordPress-Android/pull/13631/commits/c0ecdab3fbc05ed9f3fd9a9e1fc927f9f56ab813

ParaskP7:
> Feel free to review it next week.

👋 @malinajirka !

Yeap, I will have to postpone this for Monday morning if that's okay with you. Lots of investigation and zoom calls today, but good progress nevertheless. 😅 

PS: I see some conflicting files. Maybe you want to check on that to have this PR ready for merging after I am done with my review.

malinajirka:
> Yeap, I will have to postpone this for Monday morning if that's okay with you.

No problem at all ;).

The conflicts are fine, I'll to need replace FluxC hash with a tag anyway so I'll do that together.

ParaskP7:
👋 @malinajirka !

Apologies for doing that again, but could you please check the build on this PR, it seems that the CI is failing on it and I am not being able to build/install this branch locally as well (same failure). It has probably something to do with the `FluxC` hash version.

```
* What went wrong:
Could not determine the dependencies of task ':WordPress:compileWasabiDebugJavaWithJavac'.
> Could not resolve all task dependencies for configuration ':WordPress:wasabiDebugCompileClasspath'.
   > Could not find com.github.wordpress-mobile.WordPress-FluxC-Android:fluxc:1.8.0-beta-3.
     Required by:
         project :WordPress
```

Let me know when you fix that and I'll resume my review again! 🙏

malinajirka:
Sorry, my bad. I was replacing FluxC hash with the correct tag name at the same time as you started the review. I should have pinged you first. All should work as expected now.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
```diff
@@ -99,9 +103,19 @@ class ActivityLogTypeFilterFragment : DialogFragment() {
         viewModel.dismissDialog.observe(viewLifecycleOwner, Observer {
             it.applyIfNotHandled { dismiss() }
         })
+
+        val dateRange: DateRange? = if (requireNotNull(arguments).containsKey(ARG_DATE_RANGE_AFTER)) {
```
ParaskP7: Question (❓): What if we make this more concrete by checking the `ARG_DATE_RANGE_BEFORE` as well? I understand that the `ARG_DATE_RANGE_AFTER` is assuming that the `ARG_DATE_RANGE_BEFORE` is available, but what if a developer in the future makes a change that somehow alters this assumption? Is it better to make this more concrete now that have it potentially fail in the future. Wdyt?

PS: Also, maybe by having that more concrete improves the intention readability, don't you think?

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
```diff
@@ -41,46 +45,51 @@ class ActivityLogTypeFilterViewModel @Inject constructor(
...
+        initialSelection: List<String>
     ) {
         if (isStarted) return
         isStarted = true
         this.remoteSiteId = remoteSiteId
         this.parentViewModel = parentViewModel
         this.initialSelection = initialSelection
-
+        this.dateRange = dateRange
         fetchAvailableActivityTypes()
```
ParaskP7, line 57: Idea (💡): What if this function is transformed to `fetchAvailableActivityTypes(remoteSiteId, dateRange)`, accepting both parameters in its signature instead of depending on the view model state. Wdyt?

PS: By having the suggestion, the function becomes more pure, don't you think?

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
```diff
@@ -41,46 +45,51 @@ class ActivityLogTypeFilterViewModel @Inject constructor(
...
-            // TODO malinjir replace "it.toString()" with activity type name
             val activityTypeListItems: List<ListItemUiState.ActivityType> = activityTypes
                     .map {
                         ListItemUiState.ActivityType(
-                                id = it.id,
-                                title = UiStringText(it.toString()),
-                                onClick = { onItemClicked(it.id) },
-                                checked = initialSelection.contains(it.id)
+                                id = it.key,
+                                title = UiStringText("${it.name} (${it.count})"),
```
ParaskP7, line 90: Idea (💡 ): Consider removing the corresponding to the change TODO comment with the commit that actually removes the TODO. For example, the `Use real getActivityTypes endpoint` commit removed the TODO, but the actual change happened few commits later, this confused me as I was trying to figure out why this comment was removed for a couple of commits as I was going through review.

PS: By making every commit do exactly one think, this kind of problems are avoided and it helps the reviewer read the PR story as it was being written by the author, thus helping both sides. This way, the reviewer can review better and the author benefit from that ,more concrete, review.

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
```diff
@@ -11,25 +14,31 @@ import org.junit.Test
...
 import org.wordpress.android.ui.activitylog.list.filter.ActivityLogTypeFilterViewModel.ListItemUiState.ActivityType
 import org.wordpress.android.ui.activitylog.list.filter.ActivityLogTypeFilterViewModel.UiState
 import org.wordpress.android.ui.activitylog.list.filter.ActivityLogTypeFilterViewModel.UiState.Content
-import org.wordpress.android.ui.activitylog.list.filter.DummyActivityTypesProvider.DummyActivityType
-import org.wordpress.android.ui.activitylog.list.filter.DummyActivityTypesProvider.DummyAvailableActivityTypesResponse
 import org.wordpress.android.viewmodel.activitylog.ActivityLogViewModel
+import org.wordpress.android.viewmodel.activitylog.DateRange
 
 @InternalCoroutinesApi
 class ActivityLogTypeFilterViewModelTest : BaseUnitTest() {
```
ParaskP7, line 36: Idea (💡): Consider updating the tests with every commit, instead updating them once after all the previous main source set commits are done.

PS: This way the reviewer is able to navigate through the change and review it better, without needing to go back few commits and understand what happened there that cause the tests to be updated.
PSS: I personally have the rabbit to run the tests associated to my main source set commit change before committing, making sure that I didn't break anything and the test still pass, kind of TDD thinking reverse.

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
```diff
@@ -172,15 +181,27 @@ class ActivityLogTypeFilterViewModelTest : BaseUnitTest() {
...
+                .filter { it.checked }.map { it.id }
         ).containsExactlyElementsOf(initialSelection)
     }
 
+    @Test
+    fun `date range passed to the api, when provided on view model start`() = test {
+        init().uiStates
+        val captor: KArgumentCaptor<FetchActivityTypesPayload> = argumentCaptor()
+
+        startVM(dateRange = Pair(1L, 2L))
```
ParaskP7: Minnor (🔍): Consider using a variable instead of hardcoded duplicate values for when creating the action and asserting afterwards that the parameter put to the actions is the one expected.
```
val after = 1L
val before = 2L

startVM(dateRange = Pair(after, before))

assertThat(captor.firstValue.after?.time).isEqualTo(after)
assertThat(captor.firstValue.before?.time).isEqualTo(before)
```

PS: This way the test because more concrete. Wdyt?

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
```diff
@@ -172,15 +181,27 @@ class ActivityLogTypeFilterViewModelTest : BaseUnitTest() {
...
+
+        verify(activityLogStore).fetchActivityTypes(captor.capture())
+        assertThat(captor.firstValue.after!!.time).isEqualTo(1L)
+        assertThat(captor.firstValue.before!!.time).isEqualTo(2L)
```
ParaskP7: Question (❓): Maybe I asked that before, but since I would have used `?` instead of `!!`, trying to make sure we can be as consistent as possible next time, what was your reasoning behind using `!!` instead of `?`?

ParaskP7 (approved):
Good job @malinajirka ! 🌟 

I particularly liked how to split each commit and make sure that the reviewer can follow your train of thoughts while working on this solution. I enjoyed that. 🙏 

I have tested the solution and it works as expected. 🟢 

I have mostly left questions and ideas for improving on our commit structure and other code/testing practices, food for thought mainly. Thus, I am approving this PR. Feel free to merge it when/if you deal with my comments.

PS: I have noticed that when switching to the `Greek` language the API is not returning the correct translation for the `Backups and Restores` title, but when I did the same for `Spanish` everything seemed okay, should we report that? See JSON response below:
```
{
  "groups": {
    "post": {
      "name": "\u0386\u03c1\u03b8\u03c1\u03b1 \u03ba\u03b1\u03b9 \u03a3\u03b5\u03bb\u03af\u03b4\u03b5\u03c2",
      "count": 71
    },
    "attachment": {
      "name": "\u03a0\u03bf\u03bb\u03c5\u03bc\u03ad\u03c3\u03b1",
      "count": 64
    },
    "rewind": {
      "name": "Backups and Restores",
      "count": 52
    },
    "plugin": {
      "name": "\u03a0\u03c1\u03cc\u03c3\u03b8\u03b5\u03c4\u03b1",
      "count": 50
    },
    "theme": {
      "name": "\u0398\u03ad\u03bc\u03b1\u03c4\u03b1",
      "count": 9
    },
    "setting": {
      "name": "\u03a1\u03c5\u03b8\u03bc\u03af\u03c3\u03b5\u03b9\u03c2 \u03b9\u03c3\u03c4\u03cc\u03c4\u03bf\u03c0\u03bf\u03c5",
      "count": 4
    },
    "user": {
      "name": "\u0386\u03bd\u03b8\u03c1\u03c9\u03c0\u03bf\u03b9",
      "count": 4
    },
    "comment": {
      "name": "\u03a3\u03c7\u03cc\u03bb\u03b9\u03b1",
      "count": 2
    }
  },
  "totalItems": 256,
  "took": 0.08943999999999996
}
```

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
malinajirka: I wanted to follow the fail fast principle. However, I agree that the code isn't easy to follow. I made some changes in 94c937b344b326087bac64989631f4d7e6158b6a, let me know what you think.

ParaskP7: @malinajirka I can't find the changes following this hash, can you maybe link it? Probably you still haven't pushed the changes, I'll wait. 😄

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
malinajirka: Thanks for the comment ;). Hmm, I was thinking about this (I even tried to implement it), but I'm still not sure.

If the method wasn't updating an internal state of the VM (or it was public), I'd completely agree. However, since the method is updating `_uiState` + it's updating the db, I feel like this improvement loses its main benefits. Moreover, imo the code becomes a bit more difficult to follow since there are more fields/parameters.

In general, I'm a fan of passing whatever the method needs (and nothing more) into a method. But in this case it feels a bit weird. Having said all that, please do follow up if you think the code would be better. Perhaps, I'm missing something or I'm too used to a certain approach even though it's not good.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
malinajirka: 👍 makes sense, I'll try to follow this ;). Thanks!

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
malinajirka: Interesting, I'll give this a try. I usually follow this approach while refactoring - since I run the tests after each change. However, I never considered it during a regular development. Thanks for the tip.

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
malinajirka: :D Funny, I do this in 95% cases. I considered it even in this method, but it felt like it's not worth it. Since you pointed it out, it's probably worth it. Fixed in 28163726184bf8f142f1108ca6fff8f41fdf9764

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
ParaskP7: Thanks for the response Jirka! 🤗 

I might be able to help you understand why I comment it in the first place and how I ended up to asking for this alternative. I saw the new `dateRange` field being added, then I saw that the `fetchAvailableActivityTypes()` function was the only one using it. Then I asked myself, why not remove the field and pass it to `fetchAvailableActivityTypes()`, only to realise that this function is used by another entry point. I then realised whey this state is needed.

Then, I thought, what if we make this more concrete and straightforward. In general having stateful class is something to be avoided, thus by making the usage of this state more evident it feels that it helps visualise the statefulness instead of hiding it within the internal of a function.

All the above is in addition to having a function as pure as possible. But, as I said, this is food for thought, just a conversation to have until next time.

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
malinajirka: I feel like that overusing "?" just leads to hidden bugs which are harder to track down. I'm trying to follow the fail fast principle even in these cases. 

In this particular test it doesn't really matter, for two reasons
1. "time" is not nullable, but imagine that "time" would be a nullable field. If it failed with "null but 2L was expected", you might be trying to figure out why "time" was null. However, when you use "!!" you can see right away, that "before" was null not "time". 

2. These are the last two lines of the test, but imagine it was used in the middle of the method. 


In general, I think that "?" should be used for cases when
1. null value is expected and it is a valid value
2. working with legacy APIs

If the state is not expected, I think the app/test should crash as soon as possible (fail fast principle).

malinajirka:
Thanks for the review and tips @ParaskP7!

> PS: I have noticed that when switching to the Greek language the API is not returning the correct translation for the Backups and Restores title, but when I did the same for Spanish everything seemed okay, should we report that? See JSON response below:

I think that no-one translated it yet. Tbh I'm not sure if these strings get translated by the community or a professional. Either way, it might be worth pinging someone from the API just to be sure. 

It's ready for another round ;).

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
malinajirka: I usually reply to comments as I go through the review and push the changes in the end. In general, it's better to wait for my "request re-review" request before you as reviewer start replying. Either way, F5/cmd+R, should solve this issue 
 now ;).

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
ParaskP7: Thanks for sharing, Jirka again! 🤗 

I am still not sure, but I urge you to have this mindset on my future PRs and let me know when I do use `?` instead of `!!`. I would love to know more of how you think about it, since I am still unsure about it. For me atm, `tests` are test, they either pass or fail. Adding `crash` logic and thinking into test is another level that I might need to explore more. So far, I only had the tests `crash` when I was trying to verify the `exception`.

Again, thanks for sharing and please keep helping me in understanding this approach more. 🤗

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
malinajirka: >  In general having stateful class is something to be avoided, 

💯 

> thus by making the usage of this state more evident it feels that it helps visualise the statefulness instead of hiding it within the internal of a function.

Interesting, I personally find it more evident when it's used directly. But I guess I see your point, if you find usages of the field, it's used just at a single place. However, if I was passing it as an argument, it'd be used at two places.

I'll need to think about this. I can see pros and cons of both approaches and I'm not sure yet, which one is better. Having said that, thank you for starting this discussion ;)!

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
```diff
@@ -36,6 +36,8 @@ class ActivityLogTypeFilterViewModelTest : BaseUnitTest() {
     @Mock private lateinit var parentViewModel: ActivityLogViewModel
     @Mock private lateinit var activityLogStore: ActivityLogStore
 
+    private val remoteSiteId = 0L
```
ParaskP7: Minor (🔍): Consider making this a constant.

malinajirka: Fixed in d1d05d26adf4ccee3b6c4ceb3dc0685e2cc34be0

ParaskP7:
Thanks for the updates @malinajirka ! 🌟 

I did review it and left a minor comment so far, approved by me.

PS: The commit links are not working for me, are they working for you?

ParaskP7:
PS: Thanks for the constant change, feel free to merge the PR when the CI is done! 🟢

malinajirka:
> PS: The commit links are not working for me, are they working for you?

Yes, have you tried refreshing the page after I pushed the commits into the remote?

ParaskP7:
> Yes, have you tried refreshing the page after I pushed the commits into the remote?

Yes, I did, but actually they just started to work now! Weird, maybe it takes some time until those commits are being transformed to linkable by GitHub itself! 🤷

WordPress/src/test/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModelTest.kt
malinajirka: I think that using this approach in tests might be a bit too much. At the same time valid usage of `!!` is way more common in tests than in regular code. Therefore, I like to use the same approach even in tests so people get use to not overusing `?`.

> Adding crash logic and thinking into test is another level that I might need to explore more.

Actually, you are not adding crash logic, you are just not adding a safe-guard against null.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
ParaskP7: Thank YOU for the discussion, I love having those! ❤️

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
ParaskP7: Thank YOU! Yeap, I was trying to be fast, next time I'll just wait... 😅

malinajirka: Thank you for trying to be fast;)!! I appreciate it.

I wish GitHub had the "code review (submit at the end)" mode even when working on the PR feedback.

Issue #13583, opened 2020-12-10
Create new Quick Start card item


renanferrari:
We need to create a new Quick Start card based on the latest design spec:

![quick-start-card](https://user-images.githubusercontent.com/830056/101840390-c5e55500-3b22-11eb-8afb-043d31ec610e.png)

This task includes:
- The progress bar at the top left corner and the ability to set a value to it.
- The overflow menu at the top right corner.
- The ability to show nested items and horizontally scroll through them.

This task does not include:
- The UI for the nested card (Task card), which should be handled in #13584.
- The necessary logic to show the card and its children, which should be handled in #13349.

##### Internal ref: pbArwn-1mk-p2

PR #13634, opened 2020-12-18
Update Material Components and Constraint Layout libraries

renanferrari:
This PR updates the following libraries:

- Material Components: `1.2.1` → `1.3.0-beta01`
- Constraint Layout: `1.1.3` → `2.0.4`

The Material Components update is motivated by the new [Progress Indicator components](https://material.io/components/progress-indicators), as they would come in very handy for the new Quick Start card designs (#13583). The Constraint Layout update is a dependency of the first one.

According to the following changelogs, none of these updates introduce breaking changes:

- [Material Components](https://github.com/material-components/material-components-android/releases)
- [Constraint Layout](https://developer.android.com/jetpack/androidx/releases/constraintlayout)

Other than bugfixes, Material Components 1.3 also introduces an interesting new [Time Picker component](https://material.io/components/time-pickers), while Constraint Layout 2.0 introduces the [`MotionLayout`](https://developer.android.com/reference/androidx/constraintlayout/motion/widget/MotionLayout), which is a subclass of `ConstraintLayout` that makes it much easier to animate between layout states and also [`Flow`](https://developer.android.com/reference/androidx/constraintlayout/helper/widget/Flow), which is a new layout mechanism that I believe would be very useful for us (I personally have been wanting to use it for quite a while now).

## To test

Smoke test the app, with special attention for high-priority areas such as login, editor, stats, notifications, etc. Make sure nothing is broken or looks out of place. 

Feel free to double-check components and layouts currently using any of the affected libraries.

---

PR submission checklist:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -141,7 +141,7 @@ ext {
 
     appCompatVersion = '1.0.2'
     coreVersion = '1.2.0'
-    constraintLayoutVersion = '1.1.3'
+    constraintLayoutVersion = '2.0.4'
     materialVersion = '1.3.0-beta01'
     preferenceVersion = '1.1.0'
     swipeToRefresh = '1.1.0'
--- a/build.gradle
+++ b/build.gradle
@@ -142,7 +142,7 @@ ext {
     appCompatVersion = '1.0.2'
     coreVersion = '1.2.0'
     constraintLayoutVersion = '1.1.3'
-    materialVersion = '1.2.1'
+    materialVersion = '1.3.0-beta01'
     preferenceVersion = '1.1.0'
     swipeToRefresh = '1.1.0'
     uCropVersion = '2.2.4'


malinajirka:
I'm not reviewing this PR until we decide if it's ok to use a beta version. 

@renanferrari Could you please elaborate a bit more on what exactly the new beta version brings that we'd like to use and is not available in the previous versions of the library? I checked the designs and it seems that they contain a regular circular progress indicator which has been available for years 😕. Thanks!

renanferrari:
Hey @malinajirka 👋 Thanks for asking!

So, there are basically 3 features that the new `ProgressIndicator` component supports out of the box that would be very relevant for what we're trying to achieve:
1. The ability to enable determinate mode on a circular progress bar.
2. The ability to easily customize both the indicator and background colors on a circular progress bar.
3. The ability to customize the corner radius of the indicator on a circular progress bar.

The regular `ProgressBar` component doesn't support any of these things out of the box and I had initially reached a dead end when trying to work around those limitations. That said, your question prompted me to take a second look at it, just to make sure I wasn't missing anything (which is why it took me a while to get back to this, sorry about that!).

What I eventually found out was that it is possible to achieve number 1 and 2 with a `ProgressBar` by using a custom drawable. It's not as simple as using the `ProgressIndicator` component, but it seems to work fine (except for a bug on Lollipop, which took me a while to figure out). As for number 3, I couldn't find a reasonable solution for it, but I'm assuming it's not as important as the other ones and can probably wait a bit longer.

In conclusion, I have decided to keep the `ProgressBar` for the time being and if we don't find any issues with it in the next few weeks, I'm going to go ahead and close this PR.

PR #13635, opened 2020-12-18
[Gutenberg] File Block upload failure handling

### LINKED ISSUE #2910: File Block: file uploads fail with a generic error for file types I don't have permissions to upload and then the post cannot be published
Steps to reproduce:

1. Open a WordPress.com Free site (no plan) in the app.
2. Go to My Site > Blog Posts > + > Blog post.
3. Add a file block and choose an mp3 file from the device ([or any file type that requires a plan](https://wordpress.com/support/accepted-filetypes/)).
4. Observe that the block now shows "( ! ) Error" in red text.
5. Observe that if you leave and re-open the post the error goes away (but the upload is still in a failed state).
6. Observe that the file is labeled "Upload Failed" in the Media screen.
7. Re-open the post.
8. Try to preview the post.
9. Try to publish the post

**Result:** if I try to upload a file type that is not allowed for my plan (i.e. an mp3 file on a free site) to a file block, the upload will fail with a generic error and then I cannot preview or publish the post. ([2m59s](https://cloudup.com/cQ5KQoUxRb2))

"( ! ) Error" in red text|No error after close/reopen|"Upload Failed" in Media
---|---|---
Screen Shot 2020-12-16 at Dec 16 4 05 46 PM|Screen Shot 2020-12-16 at Dec 16 4 06 00 PM|Screen Shot 2020-12-16 at Dec 16 4 06 09 PM
Saving is stuck on preview or publish|"We couldn't upload this media, and didn't publish the post." shows in Posts List
Screen Shot 2020-12-16 at Dec 16 4 06 40 PM|Screen Shot 2020-12-16 at Dec 16 4 07 42 PM

<sup>Tested with WPAndroid 16.4-rc-1 on Pixel 3 Android 11 on WordPress.com Free site flibbitygibbit.home.blog logged in as user907.</sup>

jd-alexander:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/2910

`gutenberg PR` https://github.com/WordPress/gutenberg/pull/27795
`gutenberg mobile PR` https://github.com/wordpress-mobile/gutenberg-mobile/pull/2917

### Description
We are implementing this behavior because currently in WP Android we have no way to know which MIME types are supported by a site. This functionality is available on iOS already so the WP iOS app isn't allowed to get into this state as the user isn't able to upload an unsupported file type. Modifications will have to be done to the Consolidated Media Picker in the future to rectify this issue, but until then an interim solution will be utilized. 

The solution involves mechanisms for clearer error propagation to the File Block from the `GlueCode` that communicates with the editor. Currently, the user just sees the word "Error" when a file upload error occurs. Errors can vary in nature so now the specific native platforms can pass a more detailed message along if necessary. In this case,the logic here is passing an error message pertaining to unsupported file types being uploaded to a site. So when the error occurs the File Block will now show the custom message within the error UI container of the File Block.

To test:

1. Select a site that is on a free plan. 
2. Add a File Block. 
3. Add an audio file. 
4. Notice the error below. 

Before | After 
--------|-------
<kbd></kbd>  |  <kbd></kbd>

PR submission checklist:


guarani:
👋 @jd-alexander the 1.44.0 editor release is being cut now (apologies for the late heads up), so I'm bumping this to 16.6 (which corresponds to editor release 1.45.0).
If this needs to make it into 16.5, and depends on editor PRs, please let me know and we can merge it in as soon as possible after the editor release is cut.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
```diff
@@ -1189,8 +1192,14 @@ public void onMediaUploadProgress(final String localMediaId, final float progres
...
     @Override
-    public void onMediaUploadFailed(final String localMediaId) {
-        getGutenbergContainerFragment().mediaFileUploadFailed(Integer.valueOf(localMediaId));
+    public void onMediaUploadFailed(final String localMediaId, final String errorMessage) {
+        if (errorMessage.equals(FILE_TYPE_NOT_PERMITTED_ERROR)) {
+            getGutenbergContainerFragment().mediaFileUploadFailed(Integer.valueOf(localMediaId),
+                    getString(R.string.error_file_type_not_supported));
+        } else {
+            getGutenbergContainerFragment().mediaFileUploadFailed(Integer.valueOf(localMediaId), null);
+        }
```
jd-alexander: I need some feedback on the direction I decided to take here. My thoughts were as follows. Currently, when the we try to upload an audio file on a free site we get the error "Sorry, this file type is not permitted for security reasons." So my idea was that once this occurs then we should tell the user that the file type is not supported. 
However, when I did a search online I realized that [this error occurs for scenarios](https://kinsta.com/knowledgebase/sorry-this-file-type-is-not-permitted-for-security-reasons/) other than the site being free or not. A primary example ( contained in the mentioned link) is when a mime type is not supported. So is it okay to keep this behavior for now until we implemented the ability to detect which mime types are supported by the site's plan?

ceyhun: Could there be any more clues inside the error we receive that's the same between other types of sites for this kind of error? Like HTTP status code or header fields? Would that be possible to pass those in here somehow?

jd-alexander: Good question. I am going to investigate if there exists anything specific for this kind of error.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
```diff
@@ -105,6 +105,9 @@
 
     private static final String USER_EVENT_KEY_TEMPLATE = "template";
 
+    private static final String FILE_TYPE_NOT_PERMITTED_ERROR =
```
cameronvoell, line 108: Where does this error message  "_Sorry, this file type is not permitted for security reasons._" come from? I noticed below `onMediaUploadFailed`  method is expecting a `errorMessage` to have this value, but it wasnt clear to me here or the other PRs where that String sentence comes from. If it is server code or elsewhere in Gutenberg/WPAndroid, perhaps we should be using an error ID or something? Not a blocker but seems strange to matching against this long String. Maybe at least a comment explaining its origin would be helpful.

cameronvoell: Ah I see in your other comment now https://github.com/wordpress-mobile/WordPress-Android/pull/13635/files#r552846333

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
cameronvoell: I agree with Ceyhun that it might be better to check for another way to detect the error other than matching the error.message against that specific string. If after more digging this is the only way we can find, some comments explaining the Strings origin would be helpful.

cameronvoell:
Hey @jd-alexander left a comment about perhaps detecting the error another way or adding a comment explaining the origin of the String error.message you are matching against. Otherwise the rest of the code looks good.

jd-alexander:
Hi @cameronvoell thanks for the feedback and reviews thus far. I added a comment about the origins of the error message f999993 and now only if the site is a free one will this error be shown. 96ebb02

I also did a screenshot of the results associated with the file upload error. 

Screenshot 2021-01-18 at 8 14 48 PM

As seen above it doesn't currently return anything specific about the site's plan. I am going to have a check with someone who can look at the API just to verify that this isn't a property that can be easily added. 

Although, we are going to be creating a mechanism similar to iOS where the file picker doesn't allow the picking of files not supported by a site's plan in the first place so if the API changes aren't simple then it would be best to ship this fix until the file picker changes are made.

jd-alexander:
Closed via https://github.com/WordPress/gutenberg/pull/27795#issuecomment-782212026

PR #13636, opened 2020-12-18
Refactor ScanListItemState to support reuse

ashiagr:
This PR splits `ScanStateViewHolder` and `ScanListItemState.ScanState` into separate ones for icon, header, description, buttons to support reuse. (Relevant discussion: https://github.com/wordpress-mobile/WordPress-Android/pull/13595#pullrequestreview-552313838)

`JetpackListItemState` is extracted out of `ScanListItemState` to contain common list item states and each state is added as an individual list item using the `ScanStateListItemBuilder`.

I've commented out tests for scan screen VM and content builder, will tweek them in another commit after early feedback for this PR as it might need few adjustments.

Note: 
1. Ignore all UI styles
2. layout in VHs need to be renamed

CC: @malinajirka, @ParaskP7 

PR submission checklist:

- [ ] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13638, opened 2020-12-18
Rename common components to jetpack

zwarm:
Parent #13329 

This PR is a follow-up to #13636. It renames layout files and resource ids to support reuse between the Jetpack scan, restore, and backup download features

To test:
Nothing to test in this PR

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/details/BackupDownloadDetailsFragment.kt
```diff
@@ -78,12 +78,12 @@ class BackupDownloadDetailsFragment : Fragment() {
...
-        val buttonLayoutId = R.layout.backup_download_list_button_item
+        val buttonLayoutId = R.layout.jetpack_list_button_item
         val checkboxLayoutId = R.layout.backup_download_list_checkbox_item
-        val descriptionLayoutId = R.layout.backup_download_list_description_item
-        val headerLayoutId = R.layout.backup_download_list_header_item
+        val descriptionLayoutId = R.layout.jetpack_list_description_item
+        val headerLayoutId = R.layout.jetpack_list_header_item
         val subheaderLayoutId = R.layout.backup_download_list_subheader_item
-        val imageLayoutId = R.layout.backup_download_list_image_item
+        val imageLayoutId = R.layout.jetpack_list_icon_item
```
ParaskP7, line 86: Minor (🔍): Maybe you want to update the `image` prefix on the variable with `icon` as well, not sure?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/common/viewholders/IconViewHolder.kt
```diff
@@ -10,9 +10,9 @@ import org.wordpress.android.util.image.ImageManager
 class IconViewHolder(
```
ParaskP7: Idea (💡): I know this is not this PRs change, but since this renaming is done now, maybe consider renaming the `JetpackViewHolder` child classes to hold the `Jetpack` name in them and avoid having them being generic. Same logic that we discussed on the `Backup` naming conventions, for searchability purposes. So, I propose:
- from `IconViewHolder` -> to `JetpackIconViewHolder` or `IconJetpackViewHolder`
- from `HeaderViewHolder` -> to `JetpackHeaderViewHolder` or `HeaderJetpackViewHolder`
- from `DescriptionViewHolder` -> to `JetpackDescriptionViewHolder` or `DescriptionJetpackViewHolder`
- from `ActionButtonViewHolder` -> to `JetpackButtonViewHolder` or `ButtonJetpackViewHolder`

Note that I removed the `Action` from the `ActionButtonViewHolder` to match that with the `jetpack_list_button_item` layout file, which doesn't contains it. Else you can add the `action` to the layout file, either way works depending how much you want the `Action` word in them.

WordPress/src/main/res/values/strings.xml
```diff
@@ -3345,4 +3345,7 @@ translators: sample content for "Services" page template -->
     <string name="backup_item_roots">WordPress root (includes wp-config.php and any non WordPress files)</string>
     <string name="backup_item_contents">WP-content directory (excludes themes, plugins, and uploads)</string>
     <string name="backup_item_sqls">Site database (SQL)</string>
+
+    <string name="jetpack_icon_content_description">icon</string>
```
ParaskP7, line 3349: Question (❓): Is that string all lowercase on purpose, maybe because it will not be going to be used anyway? Maybe `Jetpack icon` will be better?

ParaskP7 (approved):
Good job @zwarm ! 🌟 

I have left just minors/questions/ideas for you to consider, but approving the PR anyway.

WordPress/src/main/res/values/strings.xml
zwarm: >Is that string all lowercase on purpose, maybe because it will not be going to be used anyway? Maybe Jetpack icon will be better?

It will be overwritten in the classes, but need in the XML as a placeholder. I thought of Jetpack icon - but that really doesn't describe the icon well either.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/common/viewholders/IconViewHolder.kt
zwarm: Ahh, that is a GOOD idea!! Going to do that now! @ashiagr FYI

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/details/BackupDownloadDetailsFragment.kt
zwarm: Nah, not worried about this. I had another PR ready to go this morning and then Ashita & I switched up the states. So this will be gone in a few hours. Thanks.

ParaskP7:
> Commit `Add jetpack prefix to commong viewholder for context`

Reviewed, thanks for implementing the idea!

🟢  light from me, feel free to merge when the CI is done!

WordPress/src/main/res/values/strings.xml
ParaskP7: True, I only suggested the `Jetpack icon` because of the resources being named as such! 😅 

But, as you said, it doesn't matter that much.

PR #13640, opened 2020-12-18
MediaGridAdapter: use Glide to load video thumbnails

### LINKED ISSUE #9999: TimeoutException: android.media.MediaMetadataRetriever.finalize() timed out after 60 seconds
Sentry Issue: https://sentry.io/share/issue/536eac3cce544a968c6df9e044c8425a/

```
TimeoutException: android.media.MediaMetadataRetriever.finalize() timed out after 60 seconds
    at android.media.MediaMetadataRetriever.native_finalize(MediaMetadataRetriever.java)
    at android.media.MediaMetadataRetriever.finalize(MediaMetadataRetriever.java:500)
    at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:250)
    at java.lang.Daemons$FinalizerDaemon.runInternal(Daemons.java:237)
    at java.lang.Daemons$Daemon.run(Daemons.java:103)
...
(1 additional frame(s) were not displayed)

android.media.MediaMetadataRetriever.finalize() timed out after 60 seconds
```

mzorz:
Hopefully fixes #9999 

Foreword: these two comments ([@jd-alexander comment](https://github.com/wordpress-mobile/WordPress-Android/issues/9999#issuecomment-569328174) and [@develric comment](https://github.com/wordpress-mobile/WordPress-Android/issues/9999#issuecomment-667255683)) offer good insight about why this issue exists. 
Go through these first and then come back to this PR's description 👍 

#### First thoughts
I was wondering whether we could alleviate the occurrences of this crash, by avoiding `MediaMetaDataRetriever` having to perform a network call itself. My thought was something like, if the MediaMetaDataRetriever doesn't need to go out and perform the network call itself, then we can perform this in a shorter period of time, and then we would lessen the possiblities of the app going to the background / device going to sleep while all of this is happening. I have run into other various problems with `setDataSource(, headers)` method before, while it proved to be pretty stable when using the MetaDataRetriever on **_local_** files.

That said, the problem with this is that video files are generally large, and we only need a thumbnail 🙃 - it is less desirable to download a video in full just to obtain a thumbnail.

So, if we're showing a WP Media section video item, we should be able to retrieve a preprocessed thumbnail from the WP Media section (or at least that's what I thought).

Unfortunately this is not always the case, and many times there are no thumbnails available on the MediaModel (populated after what comes from the site's API). And in any case, we're already covering the case in [`loadVideoThumbnail()` in `MediaGridAdapter`](https://github.com/wordpress-mobile/WordPress-Android/blob/a9eeed79f6d85fd3a3a55157ff9b2932248c00cc/WordPress/src/main/java/org/wordpress/android/ui/media/MediaGridAdapter.java#L455-L458)  - we first try that if the thumbnail exist, and then only start figuring out some fallbacks until there's no other way around it than trying to extract the thumbnail from an actual video frame.

#### Digging further

Looking further I wondered how Glide does this. Both our implementation and Glide's end up correctly calling `release()` (see [Glide's source code here](https://github.com/bumptech/glide/blob/e464e01e0511e96ce78d4049e4110b79da8f5876/library/src/main/java/com/bumptech/glide/load/resource/bitmap/VideoDecoder.java#L183) and  see [this snippet ](https://github.com/wordpress-mobile/WordPress-Android/blob/8e55b05faff828e3d91b8c2c2be6124acabfd31e/libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/ImageUtils.java#L640)in our case)

But a difference I found here is that we were making this run on an anonymous Thread, while Glide will be run as part of our `ImageManager` singleton, which should exist throughout the Application lifecycle timespan. I think this may help avoid this problem given the references  for garbage collector to collect are still managed. 

See the docs for `finalize()` https://developer.android.com/reference/android/media/MediaMetadataRetriever#finalize()

> For example, the finalize method for an object that represents an input/output connection might perform explicit I/O transactions to break the connection before the object is permanently discarded.

One thing that caught my eye is this:
> The Java programming language does not guarantee which thread will invoke the finalize method for any given object. It is guaranteed, however, that the thread that invokes finalize will not be holding any user-visible synchronization locks when finalize is invoked. If an uncaught exception is thrown by the finalize method, the exception is ignored and finalization of that object terminates.
[...]
Any exception thrown by the finalize method causes the finalization of this object to be halted, but is otherwise ignored.

So, according to this, if an exception occurs during the execution of `finalize()` , then such an exception should be ignored. BUT, we have evidence in Sentry that this is not the case, and the app is crashing nevertheless.

That's because the crash comes from the Timeout  exception thrown, as explained  in the [SO question](https://stackoverflow.com/questions/24021609/how-to-handle-java-util-concurrent-timeoutexception-android-os-binderproxy-fin) linked in this [comment](https://github.com/wordpress-mobile/WordPress-Android/issues/9999#issuecomment-569328174) by @jd-alexander 

So I've been investigating and wondered, maybe Glide will kill / suspend its threads more intelligently (as opposed to our plain old Java Thread we launch to fetch each video thumbnail), and as such we may never see this crash (given there's no sleeping threads that will then get their MediaMetadataRetriever objects called on their `release()` method).

However this was not true, although code should be generally more legible by using Glide this way, so I left that improvement in.

But then, how can we solve the problem? The problem can't be really solved at the app's level, but we can at least try to avoid it.

### Hypothesis for this PR
The hypothetical situation described in the SO question might be hard to reproduce, and involves a GC call starting, then the device going to sleep, then resuming. However the hypothesis I used to try the solution in this PR is: when a device goes to sleep the Activity might get killed, but first of all the Activity will be sent a `onStop()` call as it's going to the background. Hence, if we knew which ImageViews are being loaded with the MediaMetadataRetriever extractor, we can cancel those requests and resume them when the Activity comes back to the foreground. Given those calls would be cancelled when we're still in  control of them, we would be avoiding the problem.

This PR then adds some logic to the adapter in methods:
- `cancelPendingRequestsForVisibleItems()` which is called at `onStop()`
- `refreshCurrentItems()` which is called at `onStart()`. This method calls `notifyItemChanged` on the adapter causing the `onBind`  method to get run and as such, the imageManager loader method to get executed.

While I have seen some lag sometimes,  I have tested the app with the memory profiler and haven't found anything unusual (it does seem to use more memory with Glide but it's able to free it up when the GC runs without problem), and I generally can't see any blockers.

**To test:**
Smoke test the app:
- open the Media section of the app, go to the videos tab
- scroll up and down, delete the app's cache
- observe it works as usual

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/media/MediaGridAdapter.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/media/MediaGridAdapter.java
@@ -19,6 +19,7 @@ import android.widget.TextView;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.recyclerview.widget.GridLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
 
 import org.wordpress.android.R;
@@ -65,6 +66,7 @@ public class MediaGridAdapter extends RecyclerView.Adapter<MediaGridAdapter.Grid
 
     private final Handler mHandler;
     private final LayoutInflater mInflater;
+    private GridLayoutManager mLayoutManager;
 
     private final Context mContext;
     private final SiteModel mSite;
@@ -77,6 +79,8 @@ public class MediaGridAdapter extends RecyclerView.Adapter<MediaGridAdapter.Grid
 
     private static final float SCALE_NORMAL = 1.0f;
     private static final float SCALE_SELECTED = .8f;
+
+    private static final String VIEW_TAG_EXTRACT_FROM_REMOTE_VIDEO_URL = "view_tag_extract_from_remote_video_url";
 
     @Inject ImageManager mImageManager;
     @Inject AuthenticationUtils mAuthenticationUtils;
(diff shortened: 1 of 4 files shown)


antonis (approved):
Great work tracking this down @mzorz 👍 
Your hypothesis seems valid and the code changes are consistent 🎉 
I performed a few tests (Pixel 2 XL/Android 11) with `Don't keep activities` developer option enabled and the app performs as expected.

I was not able to reproduce the reported crash but the thumbnail generation seems to be more consistent with your fix.

The first video is 16.4-rc1. It is a few seconds faster but the thumbnails seems to change randomly till it loads.

https://user-images.githubusercontent.com/304044/102991681-61fb5d00-4522-11eb-8e8c-9f0683aa854c.mp4

The build from this PR using Glide is a bit slower in generating the thumbnails but seems to work more consistently.

https://user-images.githubusercontent.com/304044/102991673-5d36a900-4522-11eb-93da-e0660489abdc.mp4

I did not notice any crashes or problematic behaviour and overall I consider this an improvement over the previous thumbnail generation.

mzorz:
Thanks a lot for your review @antonis! Will proceed and merge, we can reopen if we still get those crashes in Sentry 👍 Thanks once more! 🎉

PR #13639, opened 2020-12-18
1ssue/13329 details add extras

zwarm:
Parent #13329 

This PR 
- Passes needed intent extras to `BackupDownloadDetailsViewModel`
- Moves setting toolbar state into `BackupDownloadDetailsViewModel` and out of `BackupDownloadDetailsFragment`

To test:
- Launch the app with BackupDownloadConfig flag on
- Navigate to ActivityLog
- Tap the More menu
- Tap Donwload Backup menu item
- Note that the view is shown as expected

PR submission checklist:

- [X] I have considered adding unit tests where possible.
- [X] I have considered adding accessibility improvements for my changes.
- [X] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13641, opened 2020-12-20
Issue/13329 backup details viewholder

zwarm:
Parent #13329 

This PR adds support for the common `JetpackListItemState` and ViewHolders into the Backup Download Details area.

`CheckboxState` was added to the common states because it is used by both download & restore and it didn't require hardcoded values. I can add a "Backup" prefix like I did for the BACKUP_SUB_HEADER (which is also shared between backup & restore), but wanted to solicit opinions first. It shouldn't hold up the PR, because I can quickly follow this one with a prefix one.

Added some new packages for adapters & viewholders under jetpack -> backup -> details. I think the opportunity exists to pull these out into a common package to share between backup & restore and I will do during the restore implementation. Trying to keep consistent with packaging across all the features in jetpack.

I renamed BackupAvailableItemsProvider to JetpackAvailableItemsProvider and added it to the common -> provider package under common. It will be shared between restore and backup download. Again, this can be moved to common package under backup. I'm not married to the `JetpackAvailableItemsProvider` name or location. Wanted to get the concepts down and let the location fall about naturally. Again, would love opinions.

I added a ColorResId property to IconState for those instances in which we use a Material Design icon that needs coloring. Take a look at the implementation in `JetpackIconViewHolder`

I tried my best to keep the PR small, but just changing out ListItemUiState with the new one was large. The commits should be helpful. This PR builds upon another #13639 (details in the commit on dec 18) that sets up the VM to request data - it was prep work for this PR to work properly.

**Merge Instructions**
- Make sure #13639 has been merged
- Remove Not Ready for Merge Label
- Merge as normal

**Not addressed in this PR:**
(1) Any design work
(2) The action button is not hooked up
(3) Additional tests 

**To test:**
- Navigate to ActivityLog
- Tap the More menu
- Tap _Download Backup_ menu item
- Note that the view is shown as expected
- Tap on a checkbox and it should ucheck

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadListItemState.kt
```diff
@@ -1,88 +1,12 @@
...
+import org.wordpress.android.ui.jetpack.common.ViewType
 import org.wordpress.android.ui.utils.UiString
-import org.wordpress.android.ui.utils.UiString.UiStringRes
-import org.wordpress.android.ui.utils.UiString.UiStringResWithParams
 
-// todo: annmarie - tighten this up as the remainder of backup download is implemented
-sealed class BackupDownloadListItemState(val type: ViewType) {
-    open fun longId(): Long = hashCode().toLong()
+sealed class BackupDownloadListItemState(override val type: ViewType) : JetpackListItemState(type) {
+    override fun longId(): Long = hashCode().toLong()
```
ashiagr: I think we don't need to override `longId` here. I'll remove from `ScanListItemState` as well.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/common/providers/JetpackAvailableItemsProvider.kt
```diff
@@ -0,0 +1,61 @@
+package org.wordpress.android.ui.jetpack.common.providers
+
+import androidx.annotation.StringRes
+import org.wordpress.android.R.string
```
ashiagr: Np: Can we remove this import and use `R.string.xxx`? Feel free to ignore it if it is just a temporary data provider.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/details/BackupDownloadDetailsViewModel.kt
```diff
@@ -49,38 +48,40 @@ class BackupDownloadDetailsViewModel @Inject constructor(
...
+                if (contentState.type == CHECKBOX) {
+                    contentState as CheckboxState
+                    if (contentState.availableItemType == itemType) {
+                        contentState.copy(checked = !contentState.checked)
+                    } else {
+                        contentState
+                    }
                 } else {
-                    itemUiState
+                    contentState
```
ashiagr, line 84: Np: How about we use filters instead of nested ifs 😊?

```
val updatedList = content.items.filterIsInstance(CheckboxState::class.java)
                .filter { it.availableItemType == itemType }
                .map { it.copy(checked = !it.checked) }
```

WordPress/src/main/res/layout/backup_list_subheader_item.xml
```diff
@@ -7,7 +7,7 @@
     android:layout_height="?attr/listPreferredItemHeight">
 
     <com.google.android.material.textview.MaterialTextView
-        android:id="@+id/backup_download_secondary_header"
+        android:id="@+id/text"
```
ashiagr: Was it renamed to `text` to make it more generic and usable for the `restore` functionality? I know we're still finalising the names 😅,  just wanted to mention `text` looks too generic (if it is not already in your todo tasks list).

ashiagr (approved):
👋 @zwarm, changes look good.

Regarding:

> Added some new packages for adapters & viewholders under jetpack -> backup -> details. I think the opportunity exists to pull these out into a common package to share between backup & restore ...

> I renamed BackupAvailableItemsProvider to JetpackAvailableItemsProvider and added it to the common -> provider package under common. It will be shared between restore and backup download. Again, this can be moved to common package under backup.

I think it makes sense to have common package under `backup` for sharing classes between `restore` and `backup download` rather than keeping them in the `common` package shared also with the `scan` feature. Feel free to restructure it whenever you feel the right time is. 

Besides I left just minor nitpicks, functionality works as described 👍. Approving this one so it doesn't block you.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadListItemState.kt
zwarm: Sounds good. In commit beb1663

WordPress/src/main/java/org/wordpress/android/ui/jetpack/common/providers/JetpackAvailableItemsProvider.kt
zwarm: Dang it, I thought I caught all of those. Thanks. I don't know why my IDE does that. In commit 
bbaba0d

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/details/BackupDownloadDetailsViewModel.kt
zwarm: Great suggestion, I'll rewrite in the next PR because filter can be tricky. The above example snippet will filter out all values and I'll be left with a single checkbox on the screen , but I see where you were going. Thanks 👍

WordPress/src/main/res/layout/backup_list_subheader_item.xml
zwarm: It should be subheader 🤦 . I changed in commit 1009f5b

zwarm:
@ashiagr - With the exception of the filter one, I updated to your suggestions. 

> I think it makes sense to have common package under `backup` for sharing classes between `restore` and `backup download` rather than keeping them in the `common` package shared also with the `scan` feature. Feel free to restructure it whenever you feel the right time is.

Yep - time will be right shortly. :) thanks

ashiagr:
Thanks for quick fixes 🙌. All looks good to me. Feel free to merge when CI is done.

PR #13651, opened 2020-12-21
Add use case

zwarm:
Parent #13329 

This PR adds the `GetActivityLogItemUseCase` and replaces the direct call to ActivityLogStore in the VM.

**To test:**
- Navigate to ActivityLog
- Tap the More menu
- Note that the view is shown as expected and the date in the description matches the activity log item date

PR submission checklist:


ashiagr (approved):
@zwarm, @malinajirka 

I went ahead and reviewed this small PR. I've tested and everything works as before.  Just that we can create `usecases` package under `jetpack` (just like `reader` -> `usecases` ) to add this one and any future use cases.

zwarm:
> I went ahead and reviewed this small PR. I've tested and everything works as before. Just that we can create `usecases` package under `jetpack` (just like `reader` -> `usecases` ) to add this one and any future use cases.

Thanks @ashiagr . Added the package in commit e005b44.

PR #13652, opened 2020-12-21
Remove checkmark in reader subfilter

### LINKED ISSUE #13648: Removing checkmark in reader subfilter bottomsheet
In the original implementation of the Reader subfilter menu we used a checkmark to highlight the item that was selected (if any).

<p align=center><image src="https://user-images.githubusercontent.com/47797566/102787444-62201f00-43a1-11eb-942d-8e800c15f9a3.png"  width=300/></p>

In the subsequent implementation of the feature in iOS it was decided in a design review to remove that checkmark. We should align to this. 

Additionally, since we are going to add the [unread posts counts](https://github.com/wordpress-mobile/WordPress-Android/issues/13454) in the filter items for sites, it comes convenient to remove the checkmark.

develric:
Fixes #13648

This PR removes the checkmark that was visualized on the selected item in the reader bottom sheet. This is done to pair with iOS subsequent implementation. Also it is in preparation to unread posts count modification as per issue #13454

| Before | After |
|---|---|
| image | image |

While looking into this I noticed that the color of the not selected tab text was wrong in light mode (I would expect the same color scheme of the main tabs reader for example). I think the issue is setting explicitly the `app:tabTextColor="@color/primary_on_surface_selector"` in the `subfilter_bottom_sheet`. Removing it allows to use the styles set for the for the tabs in the app theme.

| Before | After |
|---|---|
| image | image |

Also Dark mode aligns better (looking to the unselected tab text) with the main reader tabs color scheme.

| Before | After |
|---|---|
| image | image |

### Removing checkmark
- Open the reader subfilter
- Select a site 
- Open the filter again and notice the checkmark is not present
- Check the same with Topics
- Smoke test the subfilter

### Align Tabs text color to theme schema
- Set the light app theme
- In the reader Following tab open the subfilter
- Check the tabs color for selected and unselected reflect that in the main reader tabs
- Check the same with dark app theme

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/viewholders/SubfilterListItemViewHolder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/viewholders/SubfilterListItemViewHolder.kt
@@ -17,10 +17,6 @@ open class SubfilterListItemViewHolder(
     @LayoutRes layout: Int
 ) : ViewHolder(LayoutInflater.from(parent.context).inflate(layout, parent, false)) {
     open fun bind(filter: SubfilterListItem, uiHelpers: UiHelpers) {
-        val selectedTick: ImageView? = this.itemView.findViewById(R.id.item_selected)
-        selectedTick?.let {
-            it.visibility = if (filter.isSelected) View.VISIBLE else View.GONE
-        }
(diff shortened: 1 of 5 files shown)


ashiagr (approved):
Both tests passed:

Removing checkmark ✅
Align Tabs text color to theme schema ✅

Thanks for noticing and fixing the tabs theme schema as well along with the linked issue 🌟.

PR #13673, opened 2020-12-23
Home Page Picker: Adds tablet and desktop preview

antonis:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2977

Depends on https://github.com/wordpress-mobile/WordPress-Android/pull/13674

## Description
This PR enhance the [HPP Preview screen](https://github.com/wordpress-mobile/WordPress-Android/pull/13399) adding a mobile, tablet, desktop preview mode selector similar to the site/page preview
* The HPP preview screen displays a preview mode icon
* When the icon is pressed a popup appears with Mobile, Tablet and Desktop options
* On load the setting from the [design selection screen is used](desktop)
* A small indication in the bottom/right corner shows the current mode

## To test:

<details>
<summary><strong>To start the Home Page Picker screen</strong></summary>

1. Start the site creation flow (e.g. *Choose site > Add button*)
1. Select the *Create WordPress.com site* option
</details>

### Preview options availability
1. Start the Home Page Picker
1. Select a design
1. Press the **Preview** button
1. **Verify** that there is a preview mode button at the top right of the screen
1. Press the **preview mode button**
1. Verify that three options are displayed: Mobile, Tablet, Desktop

### Mobile preview
1. Start the Home Page Picker
1. Select a design
1. Press the **Preview** button
1. Press the **preview mode button**
1. Select **Mobile**
1. **Verify** that a mobile preview is displayed

### Tablet preview
1. Start the Home Page Picker
1. Select a design
1. Press the **Preview** button
1. Press the **preview mode button**
1. Select **Tablet**
1. **Verify** that a tablet preview is displayed

### Desktop preview
1. Start the Home Page Picker
1. Select a design
1. Press the **Preview** button
1. Press the **preview mode button**
1. Select **Desktop**
1. **Verify** that a desktop preview is displayed

### Change mode on the preview screen
1. Start the Home Page Picker on a phone
1. Select a design
1. Press the **Preview** button
1. Press the preview mode button
1. Select **Desktop**
1. **Verify** that desktop mode is displayed
1. Close the preview screen (back or swipe down)
1. **Verify** that desktop thumbnails are displayed

### Change mode from the design screen
1. Start the Home Page Picker on a phone
1. Press the preview mode button
1. Select **Desktop**
1. Select a design
1. Press the **Preview** button
1. **Verify** that a desktop preview is displayed

## Screenshots
|Mobile|Tablet (visible selector)|Desktop|
|---|---|---|
|mobile|tablet|desktop|

[...]

PR #13674, opened 2020-12-23
Home Page Picker: Adds mobile/tablet/desktop thumbnails mode

antonis:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2978

Depends on https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1837

## Description
This PR enhance the HPP design selection screen adding the ability to select between mobile, tablet and desktop thumbnails

## To test:

<details>
<summary><strong>To start the Home Page Picker screen</strong></summary>

</details>

### Initial Mode on a Phone
1. Start the Home Page Picker on a phone
1. **Verify** that **mobile** thumbnails are displayed

### Initial Mode on a Tablet
1. Start the Home Page Picker on a phone
1.  **Verify** that **tablet** thumbnails are displayed

### Thumbnail options availability
1. Start the Home Page Picker on a phone
1.  **Verify** that there is a **thumbnail button** at the left of skip
1. Press the thumbnail button
1.  **Verify** that three options are displayed: **Mobile, Tablet, Desktop**

### Change to Tablet
1. Start the Home Page Picker on a phone
1. Press the thumbnail button (at the left of skip)
1. Select **Tablet**
1.  **Verify** that tablet thumbnails are displayed

### Change to Desktop
1. Start the Home Page Picker on a phone
1. Press the thumbnail button (at the left of skip)
1. Select **Desktop**
1.  **Verify** that desktop thumbnails are displayed

### Change to Mobile
1. Start the Home Page Picker on a phone
1. **Change to Desktop** as above
1. Press the thumbnail button (at the left of skip)
1. Select **Mobile**
1.  **Verify** that mobile thumbnails are displayed

## Screenshots
|Mobile Mode (default)|Tablet Mode|Desktop Mode|
|---|---|---|
|mobile|mobile-tablet|desktop|

|Tablet|Tablet Portrait|
|---|---|
|tablet1|tablet2|

PR submission checklist:


antonis:
Hello @kyleaparker @iamthomasbishop @SylvesterWilmott 👋 ,

This PR is a quick experiment on adding a mobile/desktop thumbnail mode in the design selection screen of the Home Page Picker after a discussion with @elibud.
Do you think this would be useful? If yes, do you have any design suggestions on improving this?

Any feedback on this is welcome :)

kyleaparker:
How does it look to the left of the skip button? I think we should always have the skip button in the same place, since we use it on multiple screens

enejb:
This is really cool. I created a PR that returns the mobile screenshot URL in the API so that we don't have to do the string replace part. See D54903-code. 

I set the mobile with to 480px would that work? Should we maybe let that part be dynamic?

iamthomasbishop:
This looks great, @antonis ! I agree with @kyleaparker that we might want to place the `Skip` button on the right for continuity.

@enejb that's great! Is that something that can be generated on the fly based on the device size? Or do we choose a single image size for all scenarios? If the latter, I think we should generate something closer to the "standard" phone screen width, something like ~360-400?

antonis:
Thank you @kyleaparker and @iamthomasbishop for the feedback 🙇 

>I agree with @kyleaparker that we might want to place the Skip button on the right for continuity.

I moved the skip button on the right as suggested. Please find a couple of screenshots below.

|Collapsed header|Initial state|
|---|---|
|device-2021-01-06-175130|device-2021-01-06-175107|


> This is really cool. I created a PR that returns the mobile screenshot URL in the API so that we don't have to do the string replace part. See D54903-code.

Thank you for the addition @enejb 👍 

> I set the mobile with to 480px would that work? Should we maybe let that part be dynamic?

I think it is better to have standard sizes for this returned from the server. Probably we need to add one more thumbnail for tablet as @elibud suggests (ref `pb3aDo-FV`).

>I would love it if we aligned with how previews are done in the web (Desktop, Tablet, Phone), and eventually, we should add Social too. The Default label is not very descriptive.

Can we use the same sizes we use on the web?

cc @chipsnyder

enejb:
Sounds great. 

I will update/set the width at which the screenshots are taken for the mobile case to be 400px. 
Should I add the tablet case as well to the endpoint? What width should we use? 800px? 
I think the current desktop is set to 1200px.

antonis:
> Sounds great.
> 
> I will update/set the width at which the screenshots are taken for the mobile case to be 400px.
> Should I add the tablet case as well to the endpoint? What width should we use? 800px?
> I think the current desktop is set to 1200px.

Thank you @enejb. I think the web uses 780px for tablet previews but 800px sound good to me.
I guess we can change this if needed cc @iamthomasbishop

iamthomasbishop:
@enejb @antonis I think 780px is probably a safe value as an equivalent to portrait-tablet, so I'd be fine with that. I also wouldn't expect there to be any substantial differences in layouts as a result of breakpoints between 780-800, but if we try 780px and find that some themes' breakpoints give us weird results in that range, we can always adjust.

mkevins (approved):
The code looks good and I've tested all the flows via Pixel 3a (physical device), and things are working as expected. I tested this via https://github.com/wordpress-mobile/WordPress-Android/pull/13673, so lets merge when the remaining issue there is resolved. Nice work!

PR #13660, opened 2020-12-22
Fix getMediaWithLocalId receiver to accept null value

mzorz:
Fixes #13658 

Title has it - the `val` receiving the result from `mediaStore.getMediaWithLocalId()` should accept nulls (we even considered a specific flow for that case a few lines below), but was being declared as `val media: MediaModel` (making Kotlin strict about non-nullability here).

This PR fixes that (df47bbb) and also removes an unneeded `Yes/No` positive/negative option in the informative dialog, making it just `Ok` now (change in 3005c04).

dialog_null

To test:
1. create a story with two images (slides) and added text
2. when publishing, tap publish and then immediately turn airplane mode ON
3. wait for the error notification to appear
4. go to the media section of the app, and delete one of the flattened Story images, corresponding to the image you selected plus the text you added (this is the new composed image which was going to be uploaded to the site, as part of your Story post)
5. now go to the Posts list, find your failed post, open it
6. turn airplane mode OFF
7. tap on the Story block which should have the failed overlay and the retry dialog appears
8. tap on the `Retry` option
9. observe a new error dialog appears, stating the image is gone now (as opposed to having the app crash)

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13683, opened 2020-12-28
WP Stories event listener: allow receiver of getMediaWithLocalId() to accept null value, and log error

### LINKED ISSUE #13680: NullPointerException: mediaStore.getSiteMediaW…e, localMediaId.toLong()) must not be null
Sentry Issue: [WORDPRESS-ANDROID-11ZA](https://sentry.io/share/issue/d9fe35a625fe414b903d198ed2f02e63/)

```
NullPointerException: mediaStore.getSiteMediaW…e, localMediaId.toLong()) must not be null
    at org.wordpress.android.ui.posts.editor.StoriesEventListener.onStoryFrameSaveCompleted(StoriesEventListener.kt:144)
...
(6 additional frame(s) were not displayed)
```

mzorz:
Fixes #13680

A few things happening here:
- This case is similar to the one fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/13660/ where the FLuxC method can return null but we were assigning the return value to a val that can't accept  null.
- Also we were using the wrong method to retrieve the site's media, which is why I think we got a `null` value in the first place: the Media item we're saving right now simply still does not exist on the site yet, but we were using the wrong FluxC MediaStore method call `getSiteMediaWithId()` (which needs the remote mediaId, but we only have a local Media Id at this point)
- if the result was indeed null, we would just silently do nothing so, not good either.

This PR then, accounts for that and is now using the right FluxC method call and, when it is the case that the media item does not exist (which would be strange), we do log an error. Note that we are  not notifying the user in this case either because I think we should never reach such a point anyway after using the right FluxC method.

To test:
N/A - I haven't been able to reproduce and I think it definitely shouldn't be possible to get a null for a file we just saved,  after the change to use the correct FluxC API (`getMediaWithLocalId()`)

PR submission checklist:


mzorz:
Closing in favor of https://github.com/wordpress-mobile/WordPress-Android/pull/13687

PR #13684, opened 2020-12-28
WP Stories: assigning temporary Id in FrameSaveStart event listener

### LINKED ISSUE #13662: IllegalArgumentException: Required value was null
Sentry Url: https://sentry.io/share/issue/a32c45528c414af8973965f6c9edd293/
Senry Url 2: https://sentry.io/share/issue/fa68ef7d97294b979c94299bb1d78f0d/

```
java.lang.IllegalArgumentException: Required value was null.
    at org.wordpress.android.ui.posts.editor.StoriesEventListener.onStoryFrameSaveCompleted(StoriesEventListener.kt:123)
    at java.lang.reflect.Method.invoke(Method.java)
    at java.lang.reflect.Method.invoke(Method.java:372)
    at org.greenrobot.eventbus.HandlerPoster.handleMessage(HandlerPoster.java:67)
    at android.os.Handler.dispatchMessage(Handler.java:111)
    at android.os.Looper.loop(Looper.java:194)
    at android.app.ActivityThread.main(ActivityThread.java:5648)
    at java.lang.reflect.Method.invoke(Method.java)
    at java.lang.reflect.Method.invoke(Method.java:372)
    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:960)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:755)
```

Logs for the user show a low memory situation, let's keep an eye on this and see if it has more occurences:

```
state: saveInstanceState
screen: PagesActivity
info
19:48:43
ui.lifecycle
state: stopped
screen: PagesActivity
info
19:48:43
generic
d/WordPress-API: ReactNativeStore: executeRequest
info
19:48:44
generic
d/WordPress-API: Dispatching action: PostAction-UPDATE_POST
info
19:48:46
generic
d/WordPress-API: Dispatching action: ListAction-LIST_DATA_INVALIDATED
info
19:48:46
generic
d/WordPress-POSTS: PageStore: getPagesFromDb
info
19:48:46
device.event
Low memory
level: 10
action: LOW_MEMORY
warning
19:49:03
exception
IllegalArgumentException: Required value was null.
```

mzorz:
Fixes #13662 

The crash observed in the related issue is due to a StoryFrameItem not having an id associated with it by the time it is **finished** being saved.

This works well when a Story is started from scratch, but the problem shows up when editing an existing Story block and adding a new slide to it.

The `StorySaveListener` is a component of the Gutenberg Editor (and hence is not present when writing a Story from scratch), and it basically handles updates for saving progress from the `FrameSaveService` and the Gutenberg Editor bridge. As such, it must have been assigned an id beforehand (like, at the start of being saved so we can correctly match save progress update events to the right Story Frame item and containing Story block)

This situation went like this:
- in `onStoryFrameSaveStart()` we're doing a `val localMediaId = event.frameId.toString()` which, if `event.frameId` is null (and it was), then `toString()` just converted it to a `"null"` String, and this is what was then passed on to the StoryMediaSaveListener which connected to the Gutenberg bridge. Nothing went wrong then afterwards, there simply wasn't a match for an id such as `"null"`.
- same thing in progress updates in `onStoryFrameSaveProgress()`, so nothing crashed there (and the event info hit the bridge only to have Gutenberg find no matches for id `"null"`
- finally, when we reached the `onStoryFrameSaveCompleted()` method, the non-null requirement is enforced and hence we crashed.

We do still want to crash if that is the case, what we need to do is to actually assign a temporary id at the beginning of the frame's save process, the same way we do when we start an empty Story block from the block editor from scratch.

To test:
1. create a story from scratch and publish it
2. go to the published Posts list
3. open the Story post you just published
4. when the Story composer appears, tap on the + control to add a new slide (that's on the right-bottom corner of the screen)
5. pick an image, add some text on it
6. tap on the ✅ icon on the top-right of the screen to save your changes
7. observe the crash doesn't happen anymore, and the Story gets correctly updated on your site (that is, you can see the new slide is there on your story on your site)

PR submission checklist:


mzorz:
Assigning a temporary id in an event listener doesn't seem right - these events should only be used to reflect and represent state change on the UI but not to make state changes themselves. Abandoning this approach.

PR #13702, opened 2021-01-03
Issue/13329 backup handlers

zwarm:
Parent #13329 

This PR adds:
- `BackupDownloadHandler` to handle the backup download request via `PostBackupDownloadUseCase`.
- `BackupDownloadStatusHandler` to  handle the backup download request via `BackupDownloadStatusUseCase `.
The above introduces `flow` and `emit` to handle progress updates.
- Starts to build out the wizard `BackupDownloadState` which gets passed to Progress & Complete views.
- Starts to build out `BackupDownloadProgressViewModel` and `BackupDownloadProgressFragment`
 
**Merge Notes**
- There is an existing[ FluxC PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1821) that needs to be merged into develop first.  

**Note:**
- There is no view in Progress or Complete. The snackbar is used to show progress for the time being. My next task will build out the progress pieces.

**To Test**
(To test locally before the FluxC PR is merged, pull down the [branch](https://github.com/wordpress-mobile/WordPress-FluxC-Android/tree/issue/13329-adjust-backup-download-gseon-type) or point to FluxC Commit 7590938)
- Navigate to ActivityLog
- Tap the More menu
- Select items to backup and then tap the Create Downloadable File Button
- Note that the progress view slides in and the snackbar is updated with progress. When download is ready, the view will slide out to complete.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/handlers/BackupDownloadHandler.kt
```diff
@@ -0,0 +1,55 @@
...
+        types: BackupDownloadRequestTypes
+    ) {
+        postBackupDownloadUseCase.postBackupDownloadRequest(rewindId, site, types)
+                .flowOn(bgDispatcher).collect { state -> handleState(state) }
+    }
+
+    private fun handleState(state: BackupDownloadHandlerStatus) {
+        when (state) {
+            is Failure -> {
+                _snackbarEvents.postValue(Event(SnackbarMessageHolder(state.message)))
```
malinajirka: ❓ Should we emit failure here (_statusUpdate)?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/handlers/BackupDownloadStatusHandler.kt
```diff
@@ -0,0 +1,53 @@
...
+    suspend fun handleBackupDownloadStatus(site: SiteModel, downloadId: Long) {
+        useCase.getBackupDownloadStatus(site, downloadId).flowOn(bgDispatcher).collect { state ->
+            handleState(state)
+        }
+    }
+
+    private fun handleState(state: BackupDownloadStatusHandlerState) {
+        when (state) {
+            is BackupDownloadStatusHandlerState.Error -> {
+                _snackbarEvents.postValue(Event(SnackbarMessageHolder(state.message)))
```
malinajirka: ❓ Should we emit failure/error here?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/handlers/BackupDownloadHandler.kt
```diff
@@ -0,0 +1,55 @@
...
+import org.wordpress.android.ui.jetpack.backup.download.handlers.BackupDownloadHandler.BackupDownloadHandlerStatus.Failure
+import org.wordpress.android.ui.jetpack.backup.download.handlers.BackupDownloadHandler.BackupDownloadHandlerStatus.Success
+import org.wordpress.android.ui.jetpack.backup.download.usecases.PostBackupDownloadUseCase
+import org.wordpress.android.ui.pages.SnackbarMessageHolder
+import org.wordpress.android.ui.utils.UiString
+import org.wordpress.android.viewmodel.Event
+import javax.inject.Inject
+import javax.inject.Named
+
+class BackupDownloadHandler @Inject constructor(
```
malinajirka: ❓ Do we plan to add more logic into this class in the future? It feels like it doesn't do anything and it adds a lot of complexity (eg. we need to observerForever in the VM). Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/handlers/BackupDownloadStatusHandler.kt
```diff
@@ -0,0 +1,53 @@
...
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.modules.BG_THREAD
+import org.wordpress.android.ui.jetpack.backup.download.usecases.BackupDownloadStatusUseCase
+import org.wordpress.android.ui.pages.SnackbarMessageHolder
+import org.wordpress.android.ui.utils.UiString
+import org.wordpress.android.viewmodel.Event
+import javax.inject.Inject
+import javax.inject.Named
+
+class BackupDownloadStatusHandler @Inject constructor(
```
malinajirka: ❓ Same as above - Do we plan to add more logic into this class?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/BackupDownloadStatusUseCase.kt
```diff
@@ -0,0 +1,53 @@
...
+    private val networkUtilsWrapper: NetworkUtilsWrapper,
+    private val activityLogStore: ActivityLogStore
+) {
+    suspend fun getBackupDownloadStatus(site: SiteModel, downloadId: Long) = flow {
+        if (!networkUtilsWrapper.isNetworkAvailable()) {
+            emit(Error(UiStringRes(R.string.error_network_connection)))
+            return@flow
+        }
+        var result: OnBackupDownloadStatusFetched?
+        while (true) {
```
malinajirka: 🚫 Do we need to add some delay here so we don't keep fetching the state every few ms or it's being handled somewhere else?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/PostBackupDownloadUseCase.kt
```diff
@@ -0,0 +1,54 @@
...
+import org.wordpress.android.ui.utils.UiString.UiStringRes
+import org.wordpress.android.ui.utils.UiString.UiStringText
+import org.wordpress.android.util.NetworkUtilsWrapper
+import javax.inject.Inject
+
+class PostBackupDownloadUseCase @Inject constructor(
+    private val networkUtilsWrapper: NetworkUtilsWrapper,
+    private val activityLogStore: ActivityLogStore
+) {
+    suspend fun postBackupDownloadRequest(
```
malinajirka, line 24: ❓ It seems this method always returns a single result and finishes. I'm wondering if using flow brings us any benefits. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/BackupDownloadStatusUseCase.kt
```diff
@@ -0,0 +1,53 @@
...
+                            downloadStatusForSite.downloadId,
+                            downloadStatusForSite.url))
+                    return@flow
+                } else {
+                    emit(Progress(downloadStatusForSite.rewindId, downloadStatusForSite.progress))
+                }
+            }
+            result = activityLogStore.fetchBackupDownloadState(FetchBackupDownloadStatePayload(site))
+            if (result.isError) {
+                val message: UiString = if (result.error.message != null) {
```
malinajirka: ⚠️ If I recall correctly, we decided that the use case layer should inform the caller about the state but it shouldn't make any decision on how to act upon them. We might want to consider moving this to the handler/vm. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/PostBackupDownloadUseCase.kt
```diff
@@ -0,0 +1,54 @@
...
+        rewindId: String,
+        site: SiteModel,
+        types: BackupDownloadRequestTypes
+    ) = flow {
```
malinajirka: Same as above - Should we move the ui related things (UiStrings) to the above layer. Wdyt?

malinajirka (commented):
Thanks @zwarm! I'm not sure if the app behaves as expected - the snackbar is always empty.

zwarm:
> Thanks @zwarm! I'm not sure if the app behaves as expected - the snackbar is always empty.

@malinajirka -Thanks so much for taking a look. I'm grateful for the feedback. Can you let me know what version of FluxC you are using, as that PR hasn't been merged into develop yet. 

That fluxC PR was just merged into develop, so if you are using composite builds please do a pull on develop. Thanks.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/handlers/BackupDownloadHandler.kt
zwarm: @malinajirka - I was on the fence about that. I was going back and forth on whether to eliminate the handlers at all and just drop the code directly in the VM. in which case I would just handle the failure directly. I did this work over the vacation and didn't have anyone to share ideas with.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/handlers/BackupDownloadStatusHandler.kt
zwarm: See my reply to previous comment

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/handlers/BackupDownloadHandler.kt
zwarm: @malinajirka - Guess I should have read all your comments before posting my first response. Yes, I can eliminate the handlers all together and move this logic directly to the view models. :)

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/handlers/BackupDownloadStatusHandler.kt
zwarm: See above comment. :) thanks

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/BackupDownloadStatusUseCase.kt
zwarm: Yes, I'll add it here

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/PostBackupDownloadUseCase.kt
zwarm: I was considering passing a "Starting" value, added it, then took it out at the last moment. I wanted to keep any "request errors" in the details fragment rather than the pass them along to the ProgressFragment. Honestly, I'm not sure the wizard was the way to go, but it's the course I am on so it stays. A case can be made that these are just different views in the same fragment, but I digress. Yes, I could eliminate flow from this use case

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/BackupDownloadStatusUseCase.kt
zwarm: Sounds right .. no UiStrings. I'll remove them

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/PostBackupDownloadUseCase.kt
zwarm: Yes, moving them

zwarm:
@malinajirka  - Update
- I removed the handlers, as they were single purpose. If they are needed in the future, they can easily be added back.
- The handler logic is now executed in the VM's and it made sense to include a single state for both cases to use
- Use cases are updated and `PostBackupDownloadUseCase` no longer uses flow. 
- All errors are shown in snackbars, but I wonder what should really happen? Other than a network connection error (which could have a RETRY action in the snackbar), I am planning on sending the user back to the activityLog with an extras bundle with a resid message - I will log the message. Would love thoughts on how to move forward if a value like downloadId or rewindId come back null from the API request. I don't think we can recover from that, but I don't want to crash the app. Thoughts?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/details/BackupDownloadDetailsViewModel.kt
```diff
@@ -171,6 +149,42 @@ class BackupDownloadDetailsViewModel @Inject constructor(
...
+                                result.downloadId,
+                                extractPublishedDate())
+                    }
+                } else {
+                    _snackbarEvents.postValue(Event(OtherRequestRunningMsg))
+                }
+            }
+            is Progress -> {
+            } // no op
+            is Complete -> {
```
malinajirka: ❓ Is there any difference between "Success" and "Complete"?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/details/BackupDownloadDetailsViewModel.kt
```diff
@@ -100,13 +119,78 @@ class BackupDownloadDetailsViewModel @Inject constructor(
...
+    private fun handleBackupDownloadRequestResult(result: BackupDownloadRequestState) {
+        when (result) {
+            is NetworkUnavailable -> {
+                _snackbarEvents.postValue(Event(NetworkUnavailableMsg))
+            }
+            is RemoteRequestFailure -> {
+                _snackbarEvents.postValue(Event(GenericFailureMsg))
+            }
+            is Success -> {
+                if (result.requestRewindId == result.rewindId) {
```
malinajirka: 🔍 Nitpick: It feels a bit weird that we show "GenericFailureMsg/OtherRequestRunningMsg" on "Success". It technically means it wasn't a success, right?

I'm wondering if the usecase should interpret the results and the vm should only act upon them. For example, the use case would return "GenericFailure/OtherRequestRunning/NetworkFailure/Success/(Progress)" and the VM would only show a snackbar or redirect the user. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCase.kt
```diff
@@ -1,53 +1,52 @@
...
-                val message: UiString = if (result.error.message != null) {
-                    UiStringText(result.error.message.toString())
-                } else {
-                    UiStringRes(R.string.backup_download_generic_failure)
-                }
-                emit(Error(message))
+                emit(RemoteRequestFailure)
                 return@flow
             }
+            kotlinx.coroutines.delay(DELAY_MILLIS)
```
malinajirka: 🔍 Nitpick: We might want to consider dropping the "kotlinx.coroutines" prefix.

malinajirka (commented):
> Can you let me know what version of FluxC you are using, as that PR hasn't been merged into develop yet.

Ohh I missed this "To test locally before the FluxC PR is merged, pull down the branch or point to FluxC Commit 7590938" sentence. I usually use the CI build to test PRs. Is there any reason why the PR isn't targeting that commit in FluxC?

I build it locally targeting "7590938" and I can't even start the download backup flow :( :




> All errors are shown in snackbars, but I wonder what should really happen? Other than a network connection error (which could have a RETRY action in the snackbar).

We might want to consult this with Chris. We usually show a fullscreen error view instead of a SnackBar, having said that I kind of like the idea of using SnackBar in this flow. The only downside I can foresee is that it's more difficult to hit the "Retry" button when using TalkBack, but I don't think it's a blocker.

Btw the snackbar with progress is for testing purposes or we plan to keep it?

> I don't think we can recover from that, but I don't want to crash the app. Thoughts?

Message like "Unexpected error" or similar + logging a message seems appropriate for such rare cases.

Thank you for all the changes @zwarm! They look good overall, however I haven't tested the app yet. 

Note: I usually prefer to split PR feedback changes into several commits. It's easier to follow and check what has changed.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/details/BackupDownloadDetailsViewModel.kt
zwarm: @malinajirka - The difference is that the "post" request returns Success & Complete is returned during a status query  (when progress == null). Two different scenarios. I can split the states if you think that is better (which is how I originally had them)

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/details/BackupDownloadDetailsViewModel.kt
zwarm: @malinajirka - I had the usecase do that logic and return the error type originally. One of your earlier comments said we shouldn't be doing that logic in the use case, so I moved it out to the VM in this last update. I can move it back to the use case. Wdyt?

zwarm:
> Is there any reason why the PR isn't targeting that commit in FluxC?

Because this PR came 4 days after my FluxC PR and I was overly optimistic that it would have been merged already. It was not an app breaking change when I submitted the FluxC PR and hadn't planned on updating the version code in the app. The FluxC PR has since been merged to develop, I'll go ahead and tag it then create a PR to update build.gradle. (after I respond to these comments)

> We might want to consult this with Chris. We usually show a fullscreen error view instead of a SnackBar, having said that I kind of like the idea of using SnackBar in this flow. The only downside I can foresee is that it's more difficult to hit the "Retry" button when using TalkBack, but I don't think it's a blocker. 

I'll reach out to Chris on the channel. I was leaning towards Full Page retry for network connection errors, but all the "null" possibilities from the API still have to be handled (probably edge cases).

> Btw the snackbar with progress is for testing purposes or we plan to keep it?

Nope, I have another PR waiting in the wings that shows the progress bar. This is just for this PR, otherwise I would have had a bazillion more changes to submit in this one PR.
 
> Message like "Unexpected error" or similar + logging a message seems appropriate for such rare cases.

Yep, almost all the required fields are null from the API level and there really is no way to recover ... not even sure a "Retry" would fix this stuff.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/details/BackupDownloadDetailsViewModel.kt
malinajirka: Could you send me a link to the comment you are referring to?

The only comment which seems related is [this one](https://github.com/wordpress-mobile/WordPress-Android/pull/13702#discussion_r551260858).

There is a difference (at least in my mind:P), between returning an interpreted state from the usecase vs returning an action (acting upon the result). For example
- returning NetworkFailureState -> is interpreted result
- returning UiStringText("Check your internet connection and try again") or updating snackbar livedata -> is acting upon the result

So I think it'd be nice if the usecase would return interpreted result, but the VM would decide what to do (how to act upon it). Does that make sense?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/details/BackupDownloadDetailsViewModel.kt
malinajirka: I see, thanks for the clarification

malinajirka:
> Because this PR came 4 days after my FluxC PR and I was overly optimistic that it would have been merged already. 

Even if it was merged already we'd still need to update this PR to refer to the new version of FluxC right? I think it's always better to commit the hash so the reviewer can use the CI build. Committing the hash shouldn't take more than a minute and it saves 15 minutes to the reviewer as they don't need to build the app locally.

> but all the "null" possibilities from the API still have to be handled (probably edge cases).

The null possibilities should be handled in the easiest way possible - we just don't want to crash the app. So even if we didn't show any error it'd be fine. Showing a generic SnackBar message is more than enough.

> Nope, I have another PR waiting in the wings that shows the progress bar. This is just for this PR, otherwise I would have had a bazillion more changes to submit in this one PR.

I completely agree that submitting more changes in this PR wouldn't be a good call. Tbh I'd have split even this into 2 PRs. 

Showing at least something is great and helps the reviewer verify that it works as expected. What I usually do in these cases is that I add a "todo" above the line which is there just for testing purposes. Just to make sure I don't forget to update it before releasing the feature.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/details/BackupDownloadDetailsViewModel.kt
zwarm: > So I think it'd be nice if the usecase would return interpreted result

Absolutely, which is why I had them there to begin with. Guess I misinterpreted your comment (Yes, the same one you listed). I'll move that out of the VM and back into the use case. 🤦

zwarm:
> Tbh I'd have split even this into 2 PRs.

Yes, I tried to keep it light, but in order to show things worked, I had to move on to progress. It was a vicious cycle.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCase.kt
zwarm: In commit 98e238c

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/details/BackupDownloadDetailsViewModel.kt
zwarm: Moved back to use case in commit 18c8aa6

zwarm:
> Even if it was merged already we'd still need to update this PR to refer to the new version of FluxC right? I think it's always better to commit the hash so the reviewer can use the CI build. Committing the hash shouldn't take more than a minute and it saves 15 minutes to the reviewer as they don't need to build the app locally.

Merged develop in to pick up new fluxC version in 6e876e1.
The snackbar will show - with the 10000L delay, it just takes a while and sometimes the process finishes too quickly. Much easier to see in the next PR, but that delay needs tweaking otherwise you won't see anything.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/details/BackupDownloadDetailsViewModel.kt
malinajirka: Perfect, it looks great now ;). Thanks for the changes and sorry for the confusion.

malinajirka (approved):
Thanks for all the changes, LGTM ;)

PR #13707, opened 2021-01-05
Issue/13268 activity type empty view

malinajirka:
Parent issue #13268

- Introduces "NoActivities" error state on Activity Log Type filter screen
- Updates message on Activity Log empty screen so it's more generic
- Replaces "rewind" with "restore" in user facing strings

To test:
1. Enable "ActivityLogFiltersFeatureConfig"
2. Turn on airplane mode
3. Open Activity Log
4. Click on "Activity Type" filter
5. Notice "No connection", "Please check your internet connection and retry."
6. Turn off airplane mode
7. Click on "Retry" and notice content is shown
8. Go back to activity log
9. Select a date range for which there are no activities (eg. 1-2 October 2020)
10. Notice "No matching events found", "Try adjusting your date range or activity type filters."
11.  Click on "Activity Type" filter
12. Notice "No activities available", "No activities recorded in the selected date range."

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
```diff
@@ -129,10 +129,11 @@ class ActivityLogTypeFilterFragment : DialogFragment() {
     }
 
     private fun refreshErrorScreen(uiState: Error) {
-        uiHelpers.setTextOrHide(actionable_empty_view.title, uiState.errorTitle)
-        uiHelpers.setTextOrHide(actionable_empty_view.subtitle, uiState.errorSubtitle)
-        uiHelpers.setTextOrHide(actionable_empty_view.button, uiState.errorButtonText)
-        actionable_empty_view.button.setOnClickListener { uiState.retryAction.action.invoke() }
+        actionable_empty_view.image.setImageResource(uiState.image)
```
malinajirka, line 132: I didn't use ImageManager here since it was resizing the image to fullscreen and I didn't want to spent time on it. Any tips are welcome.

zwarm: I will be facing a similar issue with using image manager, but haven't given it any thought yet - probably next week.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
```diff
@@ -156,13 +159,26 @@ class ActivityLogTypeFilterViewModel @Inject constructor(
             val loadingText: UiString = UiStringRes(R.string.loading)
         }
 
-        data class Error(val retryAction: Action) : UiState() {
+        sealed class Error : UiState() {
             override val errorVisibility = true
+            abstract val image: Int
```
zwarm, line 167: Small nit-pick - could this be marked as "@DrawableRes" ?

zwarm (approved):
All looks good - error pages are shown as expected. 
I left one tiny nit-pick, so go ahead and merge after your read it. 
👍

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
malinajirka: AFAIK since it's an abstract field it can't be annotated with @DrawableRes :(. Java doesn't know abstract fields so they are not implemented as actual fields -> the @DrawableRes annotation can be used only on `METHOD, PARAMETER, FIELD, LOCAL_VARIABLE`. 
I added the annotation at least to the subclasses in https://github.com/wordpress-mobile/WordPress-Android/pull/13707/commits/931c81b2110adf77d81663b1f88efdf508b65dc6. I could also make the abstract field open and assign `0` as default value. The annotation would work in that case. Let me know if you prefer that.

malinajirka:
Thanks for the review @zwarm! I've fixed the merge conflict and replied to your comment. Let me know which solution you prefer.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterViewModel.kt
zwarm: > Java doesn't know abstract fields so they are not implemented as actual fields -> the @DrawableRes annotation can be used only on `METHOD, PARAMETER, FIELD, LOCAL_VARIABLE`.

 <shakes fist at Java> - You are correct. 

> I added the annotation at least to the subclasses

Makes sense here. Thanks 👍

Issue #13629, opened 2020-12-17
Jetpack Section: Backup Screen


ParaskP7:
Parent #13267 

⚠️ The designs below are just an FYI of the latest exploration that @osullivanchris did on this screen. In the end, the team decided **NOT** to go with them and instead reuse the `Activity Log` design and screen to build this `Backup` screen.

⚠️ When it comes to design you will notice the below two differences on the `Backup` screen, compared to the `Activity Log` screen, those are:
1) The `Backup` screen doesn't have the `Activity Type` filter chip, and
1) The `Backup` screen doesn't have items in the list that are not rewindable (those without the ellipsis menu item on their right).

⚠️ As far as the `Back Details` screen in concerned, which actually will end up being the redesign `Activity Log Details` screen, the team decided **NOT** to go with the redesign on this screen for this iteration. The only change that occurred on this details screen is with the `RESTORE` button flow, which now goes through the new flow.

<img width="1779" alt="Screen Shot" src="https://user-images.githubusercontent.com/9729923/102493503-f1fd5a80-407b-11eb-928b-44985c42f600.png">

## Design

TBD

#### My Site Tab

- [x] Add feature flag for the `Backups` screen. #13709
- [x] Add `Backup` menu item as an entry point that displays the `Backups` screen. #13709
- [x]  Fine-tune the `Backup` menu icon. #13969 #13982

#### FluxC (Network + Persistence)
- [x] Add rewindable only flag for the `Backups` screen. [WordPress-Flux-Android#1834](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1834)
- [x] Option.1: (⚠️ - Scoped out for now!)
    - [x] Update `/sites/$site/activity/` API with a new `isRewindable` parameter. (⚙️ - Internal P2 link: pcdRpT-cB)
- [x] Option.2: (ℹ️ - If Option.1 is not ready)
    - [x] Update `ACTIVITY_LOG_PAGE_SIZE` to 100. [WordPress-Flux-Android#1863](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1863) #13975

#### Backups Screen

- [x] Option.1:
    - [x] Investigate requirements, concerns (⚠️) and suggestions (✅  + ❌ ) for reusing `Activity Log` screen. https://github.com/wordpress-mobile/WordPress-Android/issues/13629#issuecomment-758531526
[...]

PR #13709, opened 2021-01-05
Jetpack Section: Backup Screen - Introduce Feature Flag

ParaskP7:
Parent #13629 

This PR introduces a local feature flag for the `Backups` feature. When the feature flag is enabled a `Backup` menu is shown on the `My Site` tab (currently irrespective of the site plan).

To test:

- Launch app and go to `My Site` tab.
- Notice that `Backup` menu is hidden.
- Go to `My Site` -> `Toolbar Avatar` -> `App Settings` -> `Test feature configuration`.
- Enable `BackupsFeatureConfig`, scroll down and click the `RESTART THE APP` button. **DO NOT** confuse this feature flag with the `BackupFeatureConfig` (singular). Make sure to enable the right feature flag.
- Go back to `My Site` tab and notice `Backup` menu is shown.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/res/values/strings.xml
+++ b/WordPress/src/main/res/values/strings.xml
@@ -1037,6 +1037,7 @@
     <string name="activity_log_item_menu_download_backup_label">Download backup</string>
 
     <!-- scan -->
+    <string name="backup">Backup</string>
     <string name="scan">Scan</string>
 
     <string name="scan_state_icon">Scan state icon</string>
--- a/WordPress/src/main/res/values/strings.xml
+++ b/WordPress/src/main/res/values/strings.xml
@@ -1036,8 +1036,10 @@
     <string name="activity_log_item_menu_restore_label">Restore to this point</string>
     <string name="activity_log_item_menu_download_backup_label">Download backup</string>
 
-    <!-- scan -->
+    <!-- backup -->
     <string name="backup">Backup</string>
+
+    <!-- scan -->
     <string name="scan">Scan</string>
 
     <string name="scan_state_icon">Scan state icon</string>
--- a/WordPress/build.gradle
+++ b/WordPress/build.gradle
@@ -77,6 +77,7 @@ android {
         buildConfigField "boolean", "ANY_FILE_UPLOAD", "true"
         buildConfigField "boolean", "CONSOLIDATED_MEDIA_PICKER", "false"
         buildConfigField "boolean", "ACTIVITY_LOG_FILTERS", "false"
+        buildConfigField "boolean", "BACKUPS_AVAILABLE", "false"
         buildConfigField "boolean", "SCAN_AVAILABLE", "false"
         buildConfigField "boolean", "ENABLE_FEATURE_CONFIGURATION", "true"
         buildConfigField "boolean", "MY_SITE_IMPROVEMENTS", "false"
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
@@ -407,6 +407,9 @@ class MySiteFragment : Fragment(),
                     selectedSite
             )
         }
+        row_backup.setOnClickListener {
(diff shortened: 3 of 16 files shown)


WordPress/src/main/res/values/strings.xml
```diff
@@ -1037,6 +1037,7 @@
     <string name="activity_log_item_menu_download_backup_label">Download backup</string>
 
     <!-- scan -->
+    <string name="backup">Backup</string>
```
ashiagr: Minor 🔍: We might want to put `Backup` related strings in a separate section.

ashiagr (commented):
LGTM @ParaskP7! Also thank you for fixing android `R` imports 🙏.

Just a note for the future implementation: 

(I understand this PR only targets the "old My Site" section.)

We have a `ImprovedMySiteFragment` which will replace the old `MySiteFragment` soon. It gets enabled when `MySiteImprovementsFeatureConfig` is enabled in App Settings. 

"Scan" menu is functional only from the improved my site flow right now. We might want to sync in the next step to decide if we want to support new menus on the old My Site as I am not sure when `MySiteImprovements` is getting released.

ParaskP7:
👋 @ashiagr !

Thank you so much for this review! 🌟 

> We have a ImprovedMySiteFragment which will replace the old MySiteFragment soon. It gets enabled when MySiteImprovementsFeatureConfig is enabled in App Settings.

You are totally right, I wasn't aware of that and missed that completely. I am currently updating this PR to add the `Backup` menu there as well. ⚙️ 

> "Scan" menu is functional only from the improved my site flow right now. We might want to sync in the next step to decide if we want to support new menus on the old My Site as I am not sure when MySiteImprovements is getting released.

- I am seeing `Scan` being functional from the `MySiteFragment` as well, what do you mean by it being functional only from the `ImprovedMySiteFragment`, am I missing something?
- I agree, we should definitely sync on whether we need to support new menus on the old `MySiteFragment` and when the new `ImprovedMySiteFragment` is going live.

ashiagr:
> I am seeing Scan being functional from the MySiteFragment as well, what do you mean by it being functional only from the ImprovedMySiteFragment, am I missing something?

Oh I thought Scan menu  is not connected to the Scan screen if accessed from the `MySiteFragment`; it is very-much connected, so we're good here. It's the display logic based on scan availability which is present only in the `ImprovedMySiteFragment's` VM. This logic is being replaced by the site capabilities API that we recently figured out ( @malinajirka is working on it). Let's sync on Slack so we're all on the same page.

WordPress/src/main/res/values/strings.xml
ParaskP7: This is now done: 87590d6338a2c676f8860eafe5e818a595e6f9af

Thanks for point it out Ashita! ❤️

ParaskP7:
👋 @ashiagr !

I have uploaded the solution for `ImprovedMySiteFragment` as well. I just re-requested your review.

Many thanks in advance! 🙏

ashiagr (approved):
Thanks for the changes @ParaskP7! I have verified that backup menu is present on both old and new My Site fragment based on the feature flag & currently irrespective of the site capabilities.

Also thanks for splitting code changes into separate commits and including appropriate information in the commit messages. It made reviewing very easy, good job 🌟!

ParaskP7:
Thanks for approving and merging the PR @ashiagr ! 🌟 

I am really glad that you liked the way I structured my commits. I have been using this approach for few years now and been advocating in favor of it. It is really nice to see you found it valuable. In my opinion a reviewer should be aware of every step of the solution, small descriptive commits along with doing commit-by-commit review help navigating to the final outcome and in a sense doing an async pair programming with the author. I feel that this works really good. Also, doing that it is very beneficial for the author itself since they don't get overwhelmed by a major commit, which might include lots of feature additions, along with refactors, maybe tests, etc. With small commits, the author can quickly review their own commit much easier before committing it.

PS: Apologies for my mumbling above 😄 I am just really glad you liked it, and thanks again! 🤗

PR #13723, opened 2021-01-07
Issue/uppercase thousands suffix

khaykov:
In this PR we are uppercasing an English suffix for 1000 (k > K) to align with iOS, as mentioned [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/15581#discussion_r552209976). That string is translatable, so other locales could use an appropriate suffix.

The suffix will be changed in the stats and unread posts counter.

[Image from Gyazo](https://gyazo.com/36c8114d0486aecd284d06c9976d1a6d)

In addition, I removed the unneeded `has()` check, that was mentioned [here](https://github.com/wordpress-mobile/WordPress-Android/pull/13625#discussion_r552176448).

To test:
- You can check the new uppercase suffix in Stats (for values > 9999) or in unread posts count (if you have enough posts).

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/models/ReaderBlog.java
+++ b/WordPress/src/main/java/org/wordpress/android/models/ReaderBlog.java
@@ -82,9 +82,7 @@ public class ReaderBlog {
             blog.numSubscribers = json.optInt("subscribers_count");
         }
 
-        if (json.has("unseen_count")) {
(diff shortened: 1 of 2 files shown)


develric (approved):
Thanks for the PR @khaykov , works as described 👍 . :shipit:

Issue #13584, opened 2020-12-10
Create new Task card item


renanferrari:
We need to create a new Task card based on the latest design spec:

![task-card](https://user-images.githubusercontent.com/830056/101840966-ce8a5b00-3b23-11eb-9cd0-50c27e46e237.png)

This task includes:
- The card's illustration and description.
- The ability to click the card.
- The ability to set a card as "done".

This task does not include:
- The UI for the parent card (Quick Start card), which should be handled in #13583.
- The necessary logic to show the card and its children, which should be handled in #13349.

##### Internal ref: pbArwn-1mk-p2

PR #13724, opened 2021-01-08
Add initial Quick Start card item implementation to the improved My Site fragment

### LINKED ISSUE #13583: Create new Quick Start card item

renanferrari:
Fixes #13583

This PR introduces a basic implementation of the Quick Start card item. This basic implementation can be divided into 3 parts:

1. The Quick Start card's title bar, containing the progress bar on the left side, the overflow button on the right side, and the title in the middle.
1. The inner list of task cards that can be scrolled horizontally.
1. The bottom sheet menu that appears when the overflow button is clicked.

Items 1 and 2 are handled by this PR, while item 3 is handled by #13725. The UI of the nested task card and the necessary logic to show each task will be handled in future PRs. For testing purposes, this PR includes two static "sample" cards based on the design spec (more details on the linked issue):

https://user-images.githubusercontent.com/830056/104061781-01357b00-51d8-11eb-999f-bb264563dc80.mp4

Note:

In order to achieve the exact same design of the progress component that is depicted in the design spec, I initially considered [updating our Material Components library](https://github.com/wordpress-mobile/WordPress-Android/pull/13634) so we could use the new `ProgressIndicator` component. Later on, I decided to use a regular `ProgressBar` with a few tweaks, instead. I gave more details on this decision [here](https://github.com/wordpress-mobile/WordPress-Android/pull/13634#issuecomment-756477352). The most noticeable difference that comes with this is that the progress bar does not contain rounded corners. Hopefully we'll be able to replace it with a `ProgressIndicator` once it hits a stable version of the Material Components library.

### To test

Make sure all unit tests in `MySiteViewModelTest` pass. 

Additionally:

1. Go to the My Site screen.
1. Notice the 2 sample Quick Start cards.
1. Make sure nothing looks out of place.
1. Close the app and open it again.
1. Notice how the progress bar animates from 0 to its actual value (this happens pretty fast, so make sure not to miss it).

---

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteAdapter.kt
```diff
@@ -3,21 +3,26 @@ package org.wordpress.android.ui.mysite
...
 import org.wordpress.android.ui.mysite.MySiteItem.Type.LIST_ITEM
 import org.wordpress.android.ui.mysite.MySiteItem.Type.QUICK_ACTIONS_BLOCK
+import org.wordpress.android.ui.mysite.MySiteItem.Type.QUICK_START_CARD
 import org.wordpress.android.ui.mysite.MySiteItem.Type.SITE_INFO_BLOCK
 import org.wordpress.android.ui.utils.UiHelpers
 import org.wordpress.android.util.image.ImageManager
 
 class MySiteAdapter(val imageManager: ImageManager, val uiHelpers: UiHelpers) : Adapter<MySiteItemViewHolder>() {
     private var items = listOf<MySiteItem>()
+    private val quickStartViewPool = RecycledViewPool()
```
planarvoid, line 25: TIL 👍

WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteItem.kt
```diff
@@ -43,6 +48,29 @@ sealed class MySiteItem(val type: Type) {
...
+    ) : MySiteItem(QUICK_START_CARD) {
+        val doneTasks = tasks.filter { it.done }
+        val progress = if (tasks.isNotEmpty()) ((doneTasks.size / tasks.size.toFloat()) * 100).roundToInt() else 0
+
+        enum class ProgressColor(@ColorRes val indicatorColor: Int, @ColorRes val trackColor: Int) {
+            GREEN(R.color.green_20, R.color.green_0),
+            ORANGE(R.color.orange_40, R.color.orange_0)
+        }
+
+        data class DummyTask(
```
planarvoid, line 60: I think we can replace the "Dummy" names for the object and view holder with something we'll use later, right? They can easily be called something like "QuickStartTask" or something similar. Even if they are just dummy objects now, WDYT?

WordPress/src/main/java/org/wordpress/android/ui/mysite/QuickStartCardViewHolder.kt
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    override fun getItemCount() = items.size
+
+    inner class DummyTaskViewHolder(itemView: View) : ViewHolder(itemView) {
+        fun bind(task: DummyTask) = itemView.apply {
+            dummy_task_title.text = task.title
+            dummy_task_description.text = task.description
+
+            val alpha = if (task.done) 0.2f else 1.0f
```
planarvoid, line 133: Are these temporary or should we move them to constants now?

planarvoid (commented):
Thanks for the changes, it's looking great. I have two minor comments and one tiny concern which might not even be necessary. Do you think it makes sense to keep the horizontal scroll position on device rotation? It's possible but a little bit tricky. I'm not sure it makes sense in our case though.

mbshakti:
Looks so great ❤️ thank you for bringing this to life @renanferrari! I jotted down some notes:

- What's the type style for the card titles? In the specs I used headline 6, the one in this build seems to be smaller and bold. 
- For dark mode, I noticed that the color of the progress bar is still white. Could we introduce some opacity so it's not too bright in dark mode? Perhaps we could use `low emphasis` alpha. 
Screen Shot 2021-01-12 at 11 27 18 AM
- What is the overall padding for the screen? I think the padding for the quick start card seems to be 16px, but the padding the my site links list item seems to be 20px. 
Screen Shot 2021-01-12 at 11 41 21 AM


Questions:
- The ellipsis for the body text, I see that the placeholder text can run long so its nice to have the ellipsis. Do we plan to keep it when we have the actual copy?

WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteItem.kt
renanferrari: That's right. My intention was always to replace those dummy objects with real implementations later on, but I decided to do that while handling #13584 since it seemed to be more related to that part of the work and also because having dummy objects here would give me more space to focus on the parent card UI without committing to a specific approach for the children cards UI. Ultimately, I'll end up reutilizing more things than I expected, so I could get rid of the dummy objects here, but I still prefer to keep them for now, just because I'm about to submit another PR with these changes anyway.

WordPress/src/main/java/org/wordpress/android/ui/mysite/QuickStartCardViewHolder.kt
renanferrari: We should probably move these to constants, well noticed! 

Edit: On second thought, this would probably fit better as part of #13584 as well.

renanferrari:
Hey @planarvoid and @mbshakti, thank you for the reviews!

>Do you think it makes sense to keep the horizontal scroll position on device rotation?

Yes, I'll look into this!

>What's the type style for the card titles? In the specs I used headline 6, the one in this build seems to be smaller and bold.

That was my mistake, sorry! I temporarily reused another style we had and forgot to change it back to what it was supposed to be. I'll fix it right away.

>For dark mode, I noticed that the color of the progress bar is still white. Could we introduce some opacity so it's not too bright in dark mode? Perhaps we could use `low emphasis` alpha. 

I'm actually using solid colors for the progress bar background here, but I agree using an alpha value would be ideal. I'm not sure how doable that would be, but I'll look into it.

>What is the overall padding for the screen? I think the padding for the quick start card seems to be 16px, but the padding the my site links list item seems to be 20px.

Good question! I was trying to follow the default Material guidelines for this but I didn't realize the other items had a `20dp` padding. We should probably aim for consistency here. Should we also adjust the Quick Start card title?

Edit: I was supposed to add the following image here, but forgot.



>The ellipsis for the body text, I see that the placeholder text can run long so its nice to have the ellipsis. Do we plan to keep it when we have the actual copy?

I'm limiting the card title to one line and the description text to three lines and adding an ellipsis to the end when these limits are reached. We could definitely experiment with different values here, though. Up to you! 🙂

mbshakti:
>Should we also adjust the Quick Start card title?

Yup, I think we can adjust both. 

>Im limiting the card title to one line and the description text to three lines and adding an ellipsis to the end when these limits are reached. We could definitely experiment with different values here, though. Up to you! 

Ah makes sense. I think it would be most helpful to make this decision in context of the actual copy in here, so for that reason, I'd like to revisit this once we have the actual tasks in here. Hope that's okay!

renanferrari:
@planarvoid @mbshakti I implemented most of your suggestions and this is now ready for another round of reviews 🙂

planarvoid (approved):
Thanks for the fix, it's looking great 👍 . Feel free to merge this PR after @mbshakti  approves it

mbshakti (approved):
Looks great! Thanks @renanferrari

PR #13725, opened 2021-01-08
Add initial Quick Start card bottom sheet menu implementation

### LINKED ISSUE #13583: Create new Quick Start card item

renanferrari:
Fixes #13583

This is a follow-up to #13724, so make sure check that one out first.

This PR introduces a basic bottom sheet implementation for when the Quick Start card overflow button is clicked. It contains three options:

- Pin
- Hide
- Don't show again

The logic for clicking any of those options should be implemented in future PRs, so for now we're only interested to correctly display the bottom sheet and to correctly map each click to its correct Quick Start card:

https://user-images.githubusercontent.com/830056/104062403-2ecef400-51d9-11eb-930c-8841ba217f02.mp4

### To test

1. Go to the My Site screen.
1. Notice the 2 sample Quick Start cards.
1. Tap on the overflow button in the top right corner of either of the cards.
1. Notice the bottom sheet menu.
1. Tap any of the options.
1. Notice a toast message appears with the correct information about the Quick Start card and option clicked.

---

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mysite/QuickStartMenuViewModel.kt
```diff
@@ -0,0 +1,36 @@
...
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+import org.wordpress.android.ui.mysite.QuickStartMenuViewModel.QuickStartMenuInteraction.Hide
+import org.wordpress.android.ui.mysite.QuickStartMenuViewModel.QuickStartMenuInteraction.Pin
+import org.wordpress.android.ui.mysite.QuickStartMenuViewModel.QuickStartMenuInteraction.Remove
+import org.wordpress.android.viewmodel.Event
+import javax.inject.Inject
+
+class QuickStartMenuViewModel
```
planarvoid, line 12: Could you please add a test for this class? Even if they are quite dumb 👍

planarvoid (commented):
It's looking great 👍  Thanks for keeping the PRs small, it was very easy to review.

mbshakti:
Thank you @renanferrari -- overflow menu functionality looks and feels awesome!

Some notes:
- Could we keep the spacing between the icons and labels to 16px?
Screen Shot 2021-01-12 at 11 51 07 AM

- I imagine we could do this later when the item functionality is implemented, but when the "pinned" item is tapped, could the copy say "Unpin" instead of "Pin this", and it could be toggled as necessary. Just so it's clear to the user that they pinned a particular card.

WordPress/src/main/java/org/wordpress/android/ui/mysite/QuickStartMenuViewModel.kt
renanferrari: Will do! 👍

renanferrari:
Thanks again for the reviews, @planarvoid and @mbshakti!

>Could we keep the spacing between the icons and labels to 16px?

I'm reusing the same layout we use for the main FAB menu here. I guess we should probably keep them consistent with each other? What do you say?

|Main FAB menu|Quick Start card menu|
|---|---|
|||

>I imagine we could do this later when the item functionality is implemented, but when the "pinned" item is tapped, could the copy say "Unpin" instead of "Pin this", and it could be toggled as necessary. Just so it's clear to the user that they pinned a particular card.

Yes, this is totally doable and we should get to it when implementing the rest of the functionality. Speaking of it, since I was focusing on the UI here, I haven't dived into what each of these actions would actually entail but I still have some questions about them. I think the "Don't show again" action is pretty clear to me, but the others not so much. For example, how would the "Pin" action actually behave? I imagine it would it push that card to the top of the list? If so, then I'd guess just one card could be pinned at a time? Also, how would the "Hide" action work? Would it hide the entire card for a specific amount of time before appearing again or would it work more like a collapse/expand feature and just hide the nested Task cards?

mbshakti:
>I'm reusing the same layout we use for the main FAB menu here. I guess we should probably keep them consistent with each other? What do you say?

Wasn't aware of that, and yes makes total sense, let's keep it consistent with that :) 

Regarding the logic for the each of the items in the menu, great question. Let me document this on a google doc folder and share it with you so we are on the same page!

renanferrari:
@planarvoid Just realized I forgot to ping you to let you know this is ready for another round of reviews.

planarvoid (approved):
Thanks for the tests 👍

mbshakti (approved):
💯 thanks @renanferrari!

PR #13726, opened 2021-01-08
Issue/13329 wizard cleanup

zwarm:
Parent #13329 

This PR cleans up the Backup Download wizard; focusing on removing unused code, handling onBackPressed situations and setting up intent extras for result. This is prep work for handling unhappy paths.

**Notes:**
- Styling has not been addressed
- The buttons are not hooked up, but you can tap on them and see a snackbar
- Testing for button clicks in progress/complete views will be added after the actions are hooked up.

**To Test**
CI Tests should pass

Do a quick run through the process and make sure the pages move through the process as expected
- Navigate to ActivityLog
- Tap the More menu
- Tap the Create downloadable file button
- The view changes to progress
- Note that the view changes to complete upon conclusion of process

PR submission checklist:

### DIFF
--- a/WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModelTest.kt
+++ b/WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModelTest.kt
@@ -1,28 +1,158 @@
 package org.wordpress.android.ui.jetpack.backup.download
 
-import androidx.arch.core.executor.testing.InstantTaskExecutorRule
+import android.os.Bundle
+import com.nhaarman.mockitokotlin2.any
+import com.nhaarman.mockitokotlin2.argThat
+import com.nhaarman.mockitokotlin2.whenever
+import kotlinx.coroutines.InternalCoroutinesApi
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Before
-import org.junit.Rule
 import org.junit.Test
-import org.junit.runner.RunWith
 import org.mockito.Mock
-import org.mockito.junit.MockitoJUnitRunner
+import org.mockito.Mockito.clearInvocations
+import org.mockito.Mockito.verify
+import org.wordpress.android.BaseUnitTest
+import org.wordpress.android.ui.jetpack.backup.download.BackupDownloadViewModel.BackupDownloadWizardState
+import org.wordpress.android.ui.jetpack.backup.download.BackupDownloadViewModel.BackupDownloadWizardState.BackupDownloadCanceled
+import org.wordpress.android.ui.jetpack.backup.download.BackupDownloadViewModel.BackupDownloadWizardState.BackupDownloadCompleted
+import org.wordpress.android.ui.jetpack.backup.download.BackupDownloadViewModel.BackupDownloadWizardState.BackupDownloadInProgress
 import org.wordpress.android.ui.jetpack.backup.download.BackupDownloadViewModel.ToolbarState
+import org.wordpress.android.ui.jetpack.backup.download.BackupDownloadViewModel.ToolbarState.CompleteToolbarState
+import org.wordpress.android.ui.jetpack.backup.download.BackupDownloadViewModel.ToolbarState.DetailsToolbarState
+import org.wordpress.android.ui.jetpack.backup.download.BackupDownloadViewModel.ToolbarState.ProgressToolbarState
 import org.wordpress.android.util.wizard.WizardManager
+import org.wordpress.android.viewmodel.SingleLiveEvent
+import java.util.Date
 
-@RunWith(MockitoJUnitRunner::class)
-class BackupDownloadViewModelTest {
-    @Rule
-    @JvmField val rule = InstantTaskExecutorRule()
-
+@InternalCoroutinesApi
+class BackupDownloadViewModelTest : BaseUnitTest() {
     @Mock lateinit var wizardManager: WizardManager<BackupDownloadStep>
+    @Mock lateinit var backupDownloadStep: BackupDownloadStep
+    @Mock lateinit var savedInstanceState: Bundle
+    private val wizardManagerNavigatorLiveData = SingleLiveEvent<BackupDownloadStep>()
 
     private lateinit var viewModel: BackupDownloadViewModel
 
+    private val rewindId = "rewindId"
+    private val downloadId = 100L
+    private val published = Date(1609690147756)
+    private val url = "url"
+
+    private val backupDownloadState = BackupDownloadState(
+            activityId = "activityId",
+            rewindId = "rewindId",
+            downloadId = 100L,
+            siteId = 200L,
+            url = null,
+            published = Date(1609690147756)
+    )
+
     @Before
     fun setUp() {
+        whenever(wizardManager.navigatorLiveData).thenReturn(wizardManagerNavigatorLiveData)
+        whenever(wizardManager.showNextStep()).then {
+            wizardManagerNavigatorLiveData.value = backupDownloadStep
+            Unit
+        }
         viewModel = BackupDownloadViewModel(wizardManager)
+    }
+
+    @Test
+    fun `when view model is started, process moves to next step`() {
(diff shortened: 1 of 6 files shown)


WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModelTest.kt
```diff
@@ -1,30 +1,160 @@
...
+        viewModel.start(null)
+        clearInvocations(wizardManager)
+        whenever(wizardManager.showNextStep()).then {
+            wizardManagerNavigatorLiveData.value = BackupDownloadStep.PROGRESS
+            Unit
+        }
+
+        viewModel.onBackupDownloadDetailsFinished(rewindId, downloadId, published)
+
+        assertThat(navigationTargets.last().wizardState.rewindId).isEqualTo(rewindId)
```
ParaskP7: Minor (🔍): What do you think of the below alternative that utilised one `assertion` to achieve the same result?

```
assertThat(navigationTargets.last().wizardState)
                .isEqualTo(BackupDownloadState(rewindId = rewindId, downloadId = downloadId, published = published))
```

PS: Same applies to the below test. Although it already uses one `assertion` because it tried to verify only the `url` maybe it would be beneficial to keep them consistent. Wdyt?
PSS: Also, with this approach you are making sure that the `wizardState` is not being altered in any other way.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModelTest.kt
```diff
@@ -1,30 +1,160 @@
...
+            Unit
+        }
+
+        viewModel.onBackupDownloadProgressFinished(url)
+
+        assertThat(navigationTargets.last().wizardState.url).isEqualTo(url)
+    }
+
+    @Test
+    fun `when onBackPressed while in details step, invokes wizard finished with cancel`() {
```
ParaskP7: 1) Question (❓): I am not sure if you want to utilise the `given, when, then` approach, but if you do want that, this test scenario could have been written as such: `given in detail step, when onBackPressed, then invokes wizard finished with cancel`

PS: The same apply to all tests. For example, the `then` wording is missing on all of them. Again, not sure if that is what you want. If not, please ignore this comment.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModelTest.kt
```diff
@@ -1,30 +1,160 @@
...
+        val wizardFinishedObserver = initObservers().wizardFinishedObserver
+
+        viewModel.start(null)
+        clearInvocations(wizardManager)
+
+        whenever(wizardManager.currentStep).thenReturn(0)
```
ParaskP7: Minor (🔍): Consider using the `BackupDownloadStep.DETAILS.id` instead of using zero directly. `0`, `1` and `2` can be considered magic numbers in this case.

PS: The same applies to the below two `Progress` and `Complete` related tests.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModelTest.kt
```diff
@@ -34,13 +164,56 @@ class BackupDownloadViewModelTest {
         assertThat(toolbarStates.size).isEqualTo(0)
     }
 
+    @Test
+    fun `backupDownloadState is writtenToBundle`() {
+        viewModel.start(savedInstanceState = null)
+
+        viewModel.writeToBundle(savedInstanceState)
```
ParaskP7, line 170: Minor (🔍): Consider adding an empty line here to split the `Act` from the `Assert`.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModelTest.kt
```diff
@@ -34,13 +164,56 @@ class BackupDownloadViewModelTest {
...
+        verify(savedInstanceState)
+                .putParcelable(any(), argThat { this is BackupDownloadState })
+    }
+
+    @Test
+    fun `when setToolbarState is invoked, toolbar state is updated`() {
```
ParaskP7: Minor (🔍): Consider splitting this to 3 different tests.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModelTest.kt
```diff
@@ -34,13 +164,56 @@ class BackupDownloadViewModelTest {
...
+
+        viewModel.setToolbarState(ProgressToolbarState())
+        assertThat(toolbarStates.last()).isInstanceOf(ProgressToolbarState::class.java)
+
+        viewModel.setToolbarState(CompleteToolbarState())
+        assertThat(toolbarStates.last()).isInstanceOf(CompleteToolbarState::class.java)
+    }
+
+    @Test
+    fun `step index is restored`() {
```
ParaskP7: Minor (🔍): Consider adding empty lines to split the `Arrange`, `Act` and `Assert`.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModelTest.kt
```diff
@@ -34,13 +164,56 @@ class BackupDownloadViewModelTest {
...
+        val index = 2
+        whenever(savedInstanceState.getInt(KEY_BACKUP_DOWNLOAD_CURRENT_STEP)).thenReturn(index)
+        whenever(savedInstanceState.getParcelable<BackupDownloadState>(KEY_BACKUP_DOWNLOAD_STATE))
+                .thenReturn(backupDownloadState)
+        viewModel.start(savedInstanceState = savedInstanceState)
```
ParaskP7, line 217: Minor (🔍): Consider removing the explicit `savedInstanceState` named argument. Within this whole test suite, I am seeing both, this to be and not being used, most of the tests are not using it. This way all test will end up being more consistent. Also, due to the fact that this function accepts only one parameter it makes it less necessary.

ParaskP7 (approved):
👋 @zwarm !

I have tested the solution and it works as expected and described.

I am approving this PR since all my comments are test related. I will not merge this PR to give you some space to address any of the comments you feel addressing. Feel free to merge the PR afterwards or re-requesting my review to have another look.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModelTest.kt
zwarm: Sounds good, thanks for the suggestion.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModelTest.kt
zwarm: I split these out based on your earlier suggestion. :)

zwarm:
@ParaskP7 - I incorporated your suggestions, want to take a second look and approve if all is ok? Thanks. :)

ParaskP7:
> @ParaskP7 - I incorporated your suggestions, want to take a second look and approve if all is ok? Thanks. :)

Awesome @zwarm ! 🌟 I check and resolved of all them, thanks for incorporating these suggestions! ❤️ I am now going to merge this PR.

PS: Just a quick clarification on the `given` part of a test sentence. The `given` is optional and only required when there need to be an explicit statement to showcase the requirement or when there are multiple test that need to be differentiated by it. Only the `when` and `then` are always mandatory. For example:
- `given view model` is not needed, however
- `given in progress step` is indeed totally valid.

PR #13728, opened 2021-01-08
Adding filtering to a8c p2s tabs

develric:
A few temporary notes:

- Adding subfilter to P2 and Automattic tabs. The backend provides this mapping to Organization based on the stream (we added UNKNOWN as default value).

```
    UNKNOWN(-1),
    NO_ORGANIZATION(0),
    A8C(1),
    P2(2);
```
   
- Added Organization to each reader tab. For tabs that should present the Filter that is meaningful, otherwise it defaults to UKNOWN
- Introduced the concept of isFilterable vs isFollowing and the functions isFilterableScreen vs isFollowingScreen (this isFollowingScreen is still used in a few places where I thought it makes sense)
- Introducing SubFilterSharedViewModel to host Event bus for subfilters and mainly bottom sheet visibility for all the reader tabs. SubFilterViewModel will be dedicated per each tab instantiated based on the Organization key when the tab is Filterable.
- SubfilterBottomSheetFragment is kept as a unique instance regardless the tab but populated based on Organization parameter. Both SubfilterBottomSheetFragment and SubfilterPageFragment make use of shared VM and Organization parameter.
- It was decided with Design to keep current filter per each tab in memory but no need to keep it when restarting the app, so SubfilterListItemMap was removed.

Fixes #

To test:

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -373,6 +384,10 @@ public void onCreate(Bundle savedInstanceState) {
             if (savedInstanceState.containsKey(ReaderConstants.ARG_TAG)) {
                 mCurrentTag = (ReaderTag) savedInstanceState.getSerializable(ReaderConstants.ARG_TAG);
             }
+            if (savedInstanceState.containsKey(ReaderConstants.ARG_ORIGINAL_TAG)) {
+                mTagFragmentStartedWith =
```
malinajirka, line 388: Not technically related to this PR, but I believe `mTagFragmentStartedWith` is legacy -> ReaderPostListFragment instance shouldn't change tags during its life. We used to have a single instance of ReaderPostListFragment which would change the tag, but we refactored this and now there should be an instance per tab = per tag. Wdyt?

Could you please elaborate why you need to use this field?

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -477,7 +496,7 @@ public void onCreate(Bundle savedInstanceState) {
         mViewModel.start(mReaderViewModel);
 
         if (isFollowingScreen()) {
-            mSubFilterViewModel.onUserComesToReader();
+            mSubFilterSharedViewModel.onUserComesToReader(mOrganization);
```
malinajirka, line 499: Why we want to invoke this method only for the following tab and not for all filterable tabs?

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -556,25 +579,29 @@ private void initSubFilterViewModel() {
...
+        mSubFilterSharedViewModel.getChangeBottomSheetVisibility().observe(getViewLifecycleOwner(), event -> {
+            event.applyIfNotHandled(subfilterVisibility -> {
                 FragmentManager fm = getFragmentManager();
                 if (fm != null) {
                     SubfilterBottomSheetFragment bottomSheet =
                             (SubfilterBottomSheetFragment) fm.findFragmentByTag(SUBFILTER_BOTTOM_SHEET_TAG);
-                    if (isShowing && bottomSheet == null) {
-                        mSubFilterViewModel.loadSubFilters();
-                        bottomSheet = new SubfilterBottomSheetFragment();
+                    if (subfilterVisibility.getShow() && bottomSheet == null) {
```
malinajirka, line 588: What if we kept the state whether the bottom sheet is shown in the VM so we don't need to have this logic in the view layer? Especially the "mSubFilterSharedViewModel.loadSubFilters();" part feels a bit weird.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -556,25 +579,29 @@ private void initSubFilterViewModel() {
             }
         });
 
-        mSubFilterViewModel.getChangeBottomSheetVisibility().observe(getViewLifecycleOwner(), event -> {
-            event.applyIfNotHandled(isShowing -> {
+        mSubFilterSharedViewModel.getChangeBottomSheetVisibility().observe(getViewLifecycleOwner(), event -> {
```
malinajirka, line 582: I'm wondering if this will always work as expected. There can be multiple instances of ReaderPostListFragment -> if more instances are observing the behavior won't be determinate -> whichever instance invokes `applyIfNotHandled` first, wins. 

Tbh I still don't fully undestand why we want to keep a single instance of the bottom sheet. It seems each Reader tab displays different data/state in the bottom sheet. What are the benefits of using a single instance?

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubFilterSharedViewModel.kt
```diff
@@ -0,0 +1,225 @@
...
+
+        isStarted = true
+
+        _filtersMatchCount.value = hashMapOf()
+        _onSubfilterClicked.value = hashMapOf()
+
+        eventBusWrapper.register(this)
+    }
+
+    fun loadSubFilters() {
```
malinajirka, line 78: Is the only reason we are using a shared VM that we don't want to invoke this method multiple times when "FollowedTagsChanged" or "FollowedBlogsChanged" is emitted. 

If it's not the case, ignore the rest of this comment:). If it is the case I think that it might be an over-optimization. Loading data from the local database is quite cheap and this method performs it on a background thread. So it shouldn't affect the user at all.

malinajirka (commented):
Thank you @develric! Really nice work ;)

I haven't done a full review - I left comments/questions on things I noticed while scanning the PR. Let me know what you think ;).

develric:
Closing this in favor of #13734

PR #13734, opened 2021-01-08
Adding filtering to a8c and p2s tabs

develric:
Fixes #

To test:

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubfilterVisibilityUiModel.kt
```diff
@@ -0,0 +1,8 @@
+package org.wordpress.android.ui.reader.subfilter
+
+import org.wordpress.android.ui.Organization
+
+sealed class SubfilterVisibilityUiModel(val show: Boolean) {
+    data class SubfilterShow(val organization: Organization) : SubfilterVisibilityUiModel(true)
```
khaykov, line 6: What do you think about renaming `SubfilterShow` and `SubfilterHide` to something like `SubfilterVisible` and `SubfilterHidden`?  `SubfilterShow` and `SubfilterHide` sound too much like function names, in my optionion.

This also could be said about the `show` parameter. Maybe we could name it `isVisible`? Then the getter method will sound will be a bit more descriptive (currently it's `getShow`).

WordPress/src/test/java/org/wordpress/android/ui/reader/subfilter/SubFilterSharedViewModelTest.kt
```diff
@@ -0,0 +1,277 @@
...
+
+        for (testStep in data.keys) {
+            viewModel.onSubfilterPageUpdated(testStep, data.getOrDefault(testStep, 0))
+        }
+
+        assertThat(viewModel.filtersMatchCount.value).isEqualTo(data)
+    }
+
+    @Test
+    fun `when WPCOM user selects empty bottom sheet SITES cta the subs is opened on followed blogs page`() {
```
khaykov, line 123: Should the be `taps`  instead of `selects`? I would also change `subs` to `subs activity` (or something like this) for clarity.

The same goes for similar tests below.

khaykov (commented):
Looks good apart from the issue when adding the site by the URL everything worked great for me. You mentioned refactoring of VM, so I'm going to leave out comments related to that :)

Apart from that, I have only minor semantic comments.

develric:
This iteration is superseded by #13766

PR #13743, opened 2021-01-11
Issue/13329 error handling

zwarm:
Parent #13329 

This PR adds logic to use the last step of the wizard "Complete" in both complete and error capacities.
- Removed "Error" UiState and refactored"Content" UiState in the viewmodels, as only one state is needed since everything is a list of items
- Updated `BackupDownloadCompleteStateListItemBuilder` to include building out a JetpackItemList for error
- Updated `GetBackupDownloadStatusUseCase` to check for network connectivity after each delay to fix a potential never ending loop issue.
- Send errors to Complete view as outlined in the following image

Screenshot

**Notes**
- Please pay no attention to styles
- Complete success buttons are not hooked up yet

**To test:**
- Navigate to ActivityLog
- Tap the More menu
- Tap the Create downloadable file button
- When the view changes to progress, quickly turn on airplane mode
- Note that the view changes to "Download Failed"
- Click the Done button to close the wizard
- Don't forget to turn airplane mode back on after the test is over

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
@@ -39,7 +39,8 @@ data class BackupDownloadState(
     val rewindId: String? = null,
     val downloadId: Long? = null,
     val published: Date? = null,
-    val url: String? = null
+    val url: String? = null,
+    val isError: Boolean = false
 ) : WizardState, Parcelable
 
 typealias NavigationTarget = WizardNavigationTarget<BackupDownloadStep, BackupDownloadState>
@@ -69,6 +70,9 @@ class BackupDownloadViewModel @Inject constructor(
     private val _snackbarEvents = MediatorLiveData<Event<SnackbarMessageHolder>>()
     val snackbarEvents: LiveData<Event<SnackbarMessageHolder>> = _snackbarEvents
 
+    private val _errorEvents = MediatorLiveData<Event<Boolean>>()
+    val errorEvents: LiveData<Event<Boolean>> = _errorEvents
+
     fun start(savedInstanceState: Bundle?) {
         if (isStarted) return
         isStarted = true
@@ -86,6 +90,12 @@ class BackupDownloadViewModel @Inject constructor(
     fun addSnackbarMessageSource(snackbarEvents: LiveData<Event<SnackbarMessageHolder>>) {
         _snackbarEvents.addSource(snackbarEvents) { event ->
             _snackbarEvents.value = event
+        }
+    }
+
+    fun addErrorMessageSource(errorEvents: LiveData<Event<Boolean>>) {
+        _errorEvents.addSource(errorEvents) { event ->
+            _errorEvents.value = event
         }
     }
 
@@ -117,7 +127,8 @@ class BackupDownloadViewModel @Inject constructor(
             backupDownloadState = backupDownloadState.copy(
                     rewindId = null,
                     downloadId = null,
-                    url = null
+                    url = null,
+                    isError = false
             )
         }
     }
@@ -141,6 +152,12 @@ class BackupDownloadViewModel @Inject constructor(
 
     fun setToolbarState(toolbarState: ToolbarState) {
         _toolbarStateObservable.value = toolbarState
+    }
+
+    fun transitionToError() {
+        backupDownloadState = backupDownloadState.copy(isError = true)
+        wizardManager.setCurrentStepIndex(BackupDownloadStep.indexForErrorTransition())
+        wizardManager.showNextStep()
     }
 
     sealed class BackupDownloadWizardState : Parcelable {
@@ -170,6 +187,11 @@ class BackupDownloadViewModel @Inject constructor(
 
         data class CompleteToolbarState(
             @StringRes override val title: Int = R.string.backup_download_complete_page_title,
+            @DrawableRes override val icon: Int = R.drawable.ic_close_24px
+        ) : ToolbarState()
+
+        data class ErrorToolbarState(
+            @StringRes override val title: Int = R.string.backup_download_complete_failed_title,
             @DrawableRes override val icon: Int = R.drawable.ic_close_24px
         ) : ToolbarState()
     }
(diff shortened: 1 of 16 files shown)


WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/complete/BackupDownloadCompleteStateListItemBuilder.kt
```diff
@@ -70,4 +70,27 @@ class BackupDownloadCompleteStateListItemBuilder @Inject constructor() {
     private fun buildAdditionalInformationState() = AdditionalInformationState(
             UiStringRes(R.string.backup_download_complete_info)
     )
+
+    fun buildCompleteListStateErrorItems(onDoneClick: () -> Unit) = mutableListOf(
+                buildErrorIconState(),
+                buildErrorDescriptionState(),
+            buildErrorDoneActionState(onDoneClick)
+    )
```
ashiagr: Suggestion: Consider replacing `mutableListOf` with immutable `listOf` for returning error items from the builder🙏.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCase.kt
```diff
@@ -20,13 +20,13 @@ class GetBackupDownloadStatusUseCase @Inject constructor(
     private val activityLogStore: ActivityLogStore
 ) {
     suspend fun getBackupDownloadStatus(site: SiteModel, downloadId: Long) = flow {
-        if (!networkUtilsWrapper.isNetworkAvailable()) {
-            emit(NetworkUnavailable)
-            return@flow
-        }
-
         var result: OnBackupDownloadStatusFetched?
         while (true) {
```
ashiagr: Minor 🔍: `result` can be initialised within the while loop at line 95 where it is actually used.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/progress/BackupDownloadProgressViewModel.kt
```diff
@@ -100,13 +97,13 @@ class BackupDownloadProgressViewModel @Inject constructor(
     private fun handleState(state: BackupDownloadRequestState) {
         when (state) {
             is NetworkUnavailable -> {
-                _snackbarEvents.postValue(Event(NetworkUnavailableMsg))
+                _errorEvents.postValue(Event(true))
             }
             is RemoteRequestFailure -> {
-                _snackbarEvents.postValue(Event(GenericFailureMsg))
+                _errorEvents.postValue(Event(true))
```
ashiagr: ❓: Do we plan to show different failure reasons (n/w or remote request error) to the user? Currently we always return `Event(true)`, we might want to return actual error or `Unit` instead of `Boolean` if it is never set to `false`.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
```diff
@@ -143,6 +154,12 @@ class BackupDownloadViewModel @Inject constructor(
         _toolbarStateObservable.value = toolbarState
     }
 
+    fun transitionToError() {
+        backupDownloadState = backupDownloadState.copy(isError = true)
+        wizardManager.setCurrentStepIndex(BackupDownloadStep.indexForErrorTransition())
+        wizardManager.showNextStep()
+    }
```
ashiagr: Can we add a unit test for this scenario when (given in progress step) wizard invokes next step on error?

ashiagr (commented):
Thanks @zwarm! Changes look good overall 👍.

I tested below error flows and just have one suggestion and minor code-level comments.

1. Initial error (backup download not started) - error shown in snackbar ✅
2. Error while backup download in progress - error shown on wizard next step ✅
3. Error when backup is downloaded for an activity log while another one is in progress: Currently "There is another download running" message is shown:

device-other-request-running

How about we disable "Download backup" popup option similar to web and restrict user to go to the next screen as user won't be able to do perform any action till the previous download completes:

web-download-backup

Lmkwyt.

zwarm:
> How about we disable "Download backup" popup option similar to web and restrict user to go to the next screen as user won't be able to do perform any action till the previous download completes:

Yes, I understand what you are saying, but this check is for an edge case situation, let me explain my thinking: The app can be open on this view, user can go to the web and start the download there, then come back to the app and press the button on the screen. In that case I need to show the message. The happy path would be the user doesn't use the web/app at the same time, and the activity log list will prevent the user from getting to this view to start with. I haven't implemented the logic in the activity log list yet, it's coming. In other words, this is a "fail safe" option.  

Make a little bit more sense to you or I am insane thinking about these edge cases?

ashiagr:
> but this check is for an edge case situation, let me explain my thinking: The app can be open on this view, user can go to the web and start the download there ...
 In other words, this is a "fail safe" option.

Oh perfect! Thank you for the clarification, let's keep it here 👍. You're right on the edge case and not insane at all :).

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCase.kt
zwarm: In commit 536e0f7

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/complete/BackupDownloadCompleteStateListItemBuilder.kt
zwarm: in 536e0f7

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/progress/BackupDownloadProgressViewModel.kt
zwarm: > ❓: Do we plan to show different failure reasons (n/w or remote request error) to the user? Currently we always return `Event(true)`, we might want to return actual error or `Unit` instead of `Boolean` if it is never set to `false`.

I've swapped out the boolean for an error type. In commit 82d589e

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
zwarm: I'm going to add tests in a separate task, as I have a few others to add for the new error types now

ashiagr (approved):
LGTM 👍.

PR #13744, opened 2021-01-11
Update visibility of scan menu item

malinajirka:
Parent issue #13326 

This PR uses "JetpackCapabilityUseCase" instead of "ScanStatusService" to find out if the user has permissions to use Scan feature. It implements the logic for both legacy and improved MySiteFragment.

The result of /capabilities endpoint is stored into shared preferences and is kept as valid for 15 minutes.

Merge instructions:
1. Merge [FluxC PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1831)
2. Replace fluxc hash with tag
4. Merge this PR

To test:
1. Disable "Scan" feature flag
2. Open My Site (make sure a jetpack site with scan feature is selected)
3. Verify "Scan" menu item is hidden
-------
1. Enable "Scan" feature flag
2. Open My site (make sure a jetpack site with scan feature is selected)
3. Verify "Scan" menu item is visible
4. Switch to a site which doesn't have Scan capability
5. Make sure the Scan menu item is hidden
------------
1. Enable "Scan" feature flag
2. Enable "ImproveMySite" feature flag
3. Open My Site (make sure a jetpack site with scan feature is selected)
4. Verify "Scan" menu item is visible
5. Switch to a site which doesn't have Scan capability
6. Make sure Scan menu item is hidden
7. Switch to a site with Backup feature
8. Open Activity Log
9. Make sure the filters are shown

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/JetpackCapabilitiesUseCase.kt
```diff
@@ -0,0 +1,90 @@
+package org.wordpress.android.ui.jetpack
```
malinajirka, line 1: I renamed the file - I'm not sure why GitHub is treating it as remove and add 🤷 .

WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
```diff
@@ -211,6 +226,7 @@ class MySiteFragment : Fragment(),
         super.onCreate(savedInstanceState)
         setHasOptionsMenu(true)
         (requireActivity().application as WordPress).component().inject(this)
+        uiScope = CoroutineScope(uiDispatcher + job)
```
malinajirka, line 229: I used this a bit hacky solution here, since MySiteFragment is going away soon so I didn't want to spent more time on it. Wdyt, is it ok?

WordPress/src/test/java/org/wordpress/android/ui/jetpack/JetpackCapabilitiesUseCaseTest.kt
```diff
@@ -0,0 +1,112 @@
+package org.wordpress.android.ui.jetpack
+
```
malinajirka, line 2: Same as above - I renamed the file and I'm not sure why github is treating it as remove and add.

WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
ashiagr: I'm fine with this hack. Let's keep it as you said MySiteFragment is going away soon.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/JetpackCapabilitiesUseCase.kt
ashiagr: I usually try to rename single file and commit at a time. I see `JetpackCapabilitiesUseCaseTest` is renamed as well. I'm not sure if this's the issue here though.

ashiagr (approved):
Thanks @malinajirka! Changes look good 👍. 
I tested jetpack sites with/ without scan capability, atomic and free sites and scan menu was appropriately shown/ hidden.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/JetpackCapabilitiesUseCase.kt
malinajirka: Hmm I used AS wizard for the rename - and the wizard renamed even the related tests. I've used the same approach in the past and it worked. I'm not sure what happened here 🤷 .

ashiagr: 🤷‍♀️

PR #13752, opened 2021-01-12
Improve quota snackbar

### LINKED ISSUE #7103: Notify the user clearly when they are out of storage to upload media
Currently when the user has run out of storage in their plan, we simply show an error message.

Suggestion from @sarahblackstock:

> The error does not directly tell them they’re out of space. Would it be possible to make it do that? Something that tells them they’re at 3 out of 3GB or directs them to buy a plan to add more could help them understand what the problem is right away.

cc @elibud

planarvoid:
Fixes #7103 

This PR changes the snackbar message when the quota is exceeded and adds a light snackbar in the dark mode. I've removed the `WPSnackbar` because it should no longer be necessary. We can use the standard component which already uses material design. 

I've added a snackbar to the editor which is shown when the media upload fails. The message is now:
"Media upload failed.
Failure reasion". 
It looks like this: 

Screenshot_1610983160

Could you check the light snackbar designs to see if they look alright @SylvesterWilmott ? I've used the standard styles with the action text color being the secondary color (pink) with no alpha instead of the standard primary (in our case blue) with 0.5 alpha. 

Screenshot_1610981211
Screenshot 2021-01-18 at 16 25 38
Screenshot 2021-01-18 at 16 27 35
Screenshot 2021-01-18 at 16 27 14

To test (both in light and dark mode):
- Trigger a snackbar without an action (for example by updating a post)
- Check it looks as expected
- Trigger a snackbar with an action
- Check it looks as expected
- Click on the action
- Check it works as expected

To test: 
- Try to add an image from the editor on a site with full quota
- Notice the snackbar is visible
- Notice the message is updated

PR submission checklist:


planarvoid:
I've updated the styles to remove the additional padding @develric

osullivanchris (approved):
Checked build in light and dark mode. Looks good to me.

Nice to have the reverse colour in dark mode. And if we are using more of the standard component that's great.

develric (approved):
Hi @planarvoid, both code and behaviour look really good! Good job 👍 .

I think this PR overall address the #7103; I only spawned this dedicated issue (#13815) in case we will want to further reconsider the convenience of changing the media upload fail message in the editor overlay.

:shipit:

PR #13753, opened 2021-01-12
Save Stories composer from crashing when Post cannot be obtained

### LINKED ISSUE #13681: RuntimeException: Unable to start activity ComponentInfo{org.wordpress.android/org.wordpress.android.ui.stories.Sto...
Sentry Issue: [WORDPRESS-ANDROID-123K](https://sentry.io/organizations/a8c/issues/2112632209/?referrer=github_integration)

```
NullPointerException: Attempt to invoke interface method 'boolean org.wordpress.android.fluxc.model.PostImmutableModel.isPage()' on a null object reference
    at org.wordpress.android.ui.posts.PostEditorAnalyticsSession.<init>(PostEditorAnalyticsSession.java:65)
    at org.wordpress.android.ui.posts.PostEditorAnalyticsSession.getNewPostEditorAnalyticsSession(PostEditorAnalyticsSession.java:95)
    at org.wordpress.android.ui.posts.PostEditorAnalyticsSessionWrapper.getNewPostEditorAnalyticsSession(PostEditorAnalyticsSessionWrapper.kt:13)
    at org.wordpress.android.ui.stories.StoryComposerViewModel.createPostEditorAnalyticsSessionTracker(StoryComposerViewModel.kt:107)
    at org.wordpress.android.ui.stories.StoryComposerViewModel.setupPostEditorAnalyticsSession(StoryComposerViewModel.kt:96)
...

RuntimeException: Unable to start activity ComponentInfo{org.wordpress.android/org.wordpress.android.ui.stories.StoryComposerActivity}: java.lang.NullPointerException: Attempt to invoke interface method 'boolean org.wordpress.android.fluxc.model.PostImmutableModel.isPage()' on a null object reference
    at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2994)
    at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3072)
    at android.app.ActivityThread.-wrap11
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1766)
    at android.os.Handler.dispatchMessage(Handler.java:106)
...
(5 additional frame(s) were not displayed)
```

mzorz:
Fixes #13681 

As said in the comment [here](https://github.com/wordpress-mobile/WordPress-Android/issues/13681#issuecomment-758168432), I wanted to understand the root cause of this.

Been trying a number of things or cases where this would be possible, but I couldn't find a way to reproduce. There are 2 ways to access the StoryComposer from the Editor, and these are:
1. opening a Post that contains a Story block, then tapping on the Story block's tool icon to edit
2. opening a Post, adding a Story block, adding media to it to open the Story block

In both cases, a Post needs to exist (because we're on the Editor in the first place anyway!) so the only possibilities left that I can think of are that the actual Post Id is indeed being passed, but somehow the Post can't be loaded by FluxC. I have seen low memory indicators in one of the Sentry issues reported there, so I'm thinking this could be a possibility, but the other report doesn't show low memory signals. I still can't find a way to reproduce.

Another (more awkward) scenario I thought of was what if the Post could be loaded in the Editor, but then we trashed it away? I tried combinations of tapping on the "trash" option in the Posts list and rapidly tapping the post item to open it in the editor (it actually worked a few times for me!) but the Post is not really deleted (only marked as Trashed) so, it is always the case that the actual Post object can be found and should be loaded.

Given all of this, I've decided to follow along the way things are handled in [EditPostActivity in this regard](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L612-L616) where we just validate we have an actual loaded valid Post object before continuing, and therefore I've written this small PR that prevents the app from crashing for this specific cause of not having been able to load a Post in the internal EditPostRepository instance used in the StoryComposer - hoping that letting the user retry once more will fix it, and logging it so we can monitor how things go.

To test:
Given I haven't found a way to reproduce, you should be able to test the code flow using the following steps:
1. modify [this line here](https://github.com/wordpress-mobile/WordPress-Android/compare/issue/13681-stories-post-null?expand=1#diff-a7bfc0b68b2b746ebe42c6aff3aa0b3eab23492eca2d6b606fa0651a4d526873R84) with some random, inexistent postId you know doesn't exist in the local database. For example I used: `editPostRepository.loadPostByLocalPostId(1231231213)`
2. run WPAndroid
3. create a Story post and publish it
4. go to your published Posts list
5. tap on the recently created Post to open it in the editor
6. once the content is rendered, tap on the Story block to select it, then tap on the tools icon to open the Story composer
7. observe the Toast appears on the device screen, and the following entry shows a line like the following: `2021-01-12 13:36:17.592 21497-21497/org.wordpress.android E/WordPress-EDITOR: StoryComposerViewModel's EditPostRepository has no Post loaded: 43`

toast_composer_no_post

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
@@ -26,6 +26,8 @@ import org.wordpress.android.ui.posts.PostEditorAnalyticsSession.Outcome.SAVE
 import org.wordpress.android.ui.posts.PostEditorAnalyticsSessionWrapper
 import org.wordpress.android.ui.posts.SavePostToDbUseCase
 import org.wordpress.android.ui.stories.usecase.SetUntitledStoryTitleIfTitleEmptyUseCase
+import org.wordpress.android.util.AppLog
+import org.wordpress.android.util.AppLog.T
 import org.wordpress.android.util.helpers.MediaFile
 import org.wordpress.android.viewmodel.Event
 import javax.inject.Inject
@@ -84,6 +86,7 @@ class StoryComposerViewModel @Inject constructor(
 
         // Ensure we have a valid post
         if (!editPostRepository.hasPost()) {
+            AppLog.e(T.EDITOR, "StoryComposerViewModel's EditPostRepository has no Post loaded: " + postId.value)
             return false
         }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
@@ -69,7 +69,7 @@ class StoryComposerViewModel @Inject constructor(
         postId: LocalId,
         postEditorAnalyticsSession: PostEditorAnalyticsSession?,
         notificationType: NotificationType?
-    ) {
+    ): Boolean {
         this.editPostRepository = editPostRepository
         this.site = site
 
@@ -82,6 +82,11 @@ class StoryComposerViewModel @Inject constructor(
             editPostRepository.loadPostByLocalPostId(postId.value)
         }
 
+        // Ensure we have a valid post
+        if (!editPostRepository.hasPost()) {
+            return false
+        }
+
         setupPostEditorAnalyticsSession(postEditorAnalyticsSession)
 
         notificationType?.let {
@@ -90,6 +95,8 @@ class StoryComposerViewModel @Inject constructor(
 
         lifecycleOwner.lifecycleRegistry.currentState = Lifecycle.State.STARTED
         updateStoryPostWithChanges()
+
+        return true
     }
 
     private fun setupPostEditorAnalyticsSession(postEditorAnalyticsSession: PostEditorAnalyticsSession?) {
--- a/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
@@ -75,6 +75,10 @@ class StoryComposerViewModel @Inject constructor(
         this.editPostRepository = editPostRepository
         this.site = site
 
+        notificationType?.let {
+            systemNotificationsTracker.trackTappedNotification(it)
+        }
+
         if (postId.value == 0) {
             // Create a new post
             saveInitialPostUseCase.saveInitialPost(editPostRepository, site)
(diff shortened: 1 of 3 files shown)


WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
```diff
@@ -82,22 +88,26 @@ class StoryComposerViewModel @Inject constructor(
             editPostRepository.loadPostByLocalPostId(postId.value)
         }
 
-        setupPostEditorAnalyticsSession(postEditorAnalyticsSession)
-
-        notificationType?.let {
-            systemNotificationsTracker.trackTappedNotification(it)
+        // Ensure we have a valid post
+        if (!editPostRepository.hasPost()) {
```
ParaskP7, line 92: Question (❓): What if this logic is moved a bit further up, in between the `notificationType?.let { ... }` and the `if (postId.value == 0) { ... } else { ... }` logic? It seems that the `editPostRepository` is doing some work before checking whether the post is available, like `editPostRepository.loadPostByLocalPostId(...)`, which we might want to avoid. Wdyt?

I am not familiar with the logic and most probably I am wrong, just wanted to mentioned that in case it is something we need to consider.

ParaskP7 (commented):
👋 @mzorz !

Thanks for this PR. I run by it quickly and it looks good to me. I have just left a question (❓) for you. I will only comment on this PR and let Alex do a more thorough review to approve since I am not that familiar with the stories component.

PS: Kudos for adding more tests! 🙏

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
mzorz: oh! The `EditPostRepository` is ether initialized or loaded with an actual post right before, so it only makes sense to check whether `editPostRepository.hasPost()` after having initialized it :) -  this is actually the earliest spot where we can validate a Post actually exists in the editPostRepository 👍

mzorz:
> Thanks for this PR. I run by it quickly and it looks good to me. I have just left a question (❓) for you. I will only comment on this PR and let Alex do a more thorough review to approve since I am not that familiar with the stories component.

Thank you for taking the time to test and review this one @ParaskP7 ! 🙇 Will let @aforcier chime in later then 🙇

mzorz:
> PS: Kudos for adding more tests! 🙏

:D thank you!

WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerViewModel.kt
ParaskP7: Cool, thanks for letting me know of this Mario! 🌟

aforcier (approved):
Looks good! Confirmed the toast appears as expected. :shipit:

PR #13757, opened 2021-01-13
Issue/13326 init history view pager

malinajirka:
Parent issue #13326

This PR
- Adds ScanHistoryListFragment: An instance per view pager tab will be used, ScanHistoryFragment will serve as parent of the ViewPager. I decided not to put the ViewPager into the Activity in order to keep the activity as simple as possible in case we decide to move to a single activity structure.
- Moves Toolbar from ScanHistory Activity to ScanHistoryFragment: Same reasons as above
- Adds ViewPager2 to ScanHistoryFragment
- Adds "lift on scroll" (Z-axis) behavior. I tried adding show/hide on scroll behavior but it didn't work, I decided not to spent more time on it right now and add it later. 

To test:
2. Make sure a site with Scan feature is selected
3. Open Scan
4. Click on "History" action in the toolbar
5. Notice a viewpager with 3 tabs (3 ScanHistoryListFragments) is shown.
6. Scroll down and notice the toolbar is lifted as you start scrolling (z-axis)

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryFragment.kt
```diff
@@ -1,34 +1,44 @@
...
+import android.view.MenuItem
 import android.view.View
+import androidx.appcompat.app.AppCompatActivity
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.ViewModelProvider
+import androidx.viewpager2.adapter.FragmentStateAdapter
+import com.google.android.material.tabs.TabLayoutMediator
+import kotlinx.android.synthetic.main.scan_history_fragment.*
 import org.wordpress.android.R
+import org.wordpress.android.R.string
```
ashiagr: Nitpick 🔍: We can remove this import.

WordPress/src/main/res/layout/scan_history_activity.xml
```diff
@@ -10,20 +10,6 @@
...
-        android:layout_height="wrap_content">
-
-        <com.google.android.material.appbar.MaterialToolbar
-            android:id="@+id/toolbar_main"
-            android:layout_width="match_parent"
-            android:layout_height="wrap_content"
-            app:theme="@style/WordPress.ActionBar" />
-
-    </com.google.android.material.appbar.AppBarLayout>
+        android:layout_height="match_parent" />
```
ashiagr: Nitpick 🔍: 1. Unused app name space can be removed
`xmlns:app="http://schemas.android.com/apk/res-auto"`

2. Do we need `CoordinatorLayout` in the activity xml now that it has been moved to the fragment?

WordPress/src/main/res/layout/scan_history_list_fragment.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <androidx.core.widget.NestedScrollView
```
ashiagr, line 8: ❓: I guess `NestedScrollView` is added just for demonstrating `liftOnScroll` behaviour, right?

ashiagr (approved):
LGTM @malinajirka! Thank you for keeping activity as simple as possible 🙏.
I just left two minor comments and a question. Feel free to merge after considering them.

WordPress/src/main/res/layout/scan_history_list_fragment.xml
malinajirka: Yes, I plan to use recyclerView for the real content.

WordPress/src/main/res/layout/scan_history_activity.xml
malinajirka: Ohh, good point... Actually, this might be the reason why the show/hide behavior wasn't working.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryFragment.kt
malinajirka: Fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/13757/commits/5dfb89cf2de980af3dace30bc8d4a96994322363

WordPress/src/main/res/layout/scan_history_activity.xml
malinajirka: Fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/13757/commits/be7f344315a44b4e2ee394d54146d98ab564ec60

malinajirka:
Thanks for the review @ashiagr! I've implemented your suggestions - it's ready for another round 🙇 .

ashiagr (approved):
Thanks for the changes, all looks good to me 👍.

PR #13762, opened 2021-01-13
Jetpack Section: Backup Screen - Introduce Backup Screen

ParaskP7:
Parent #13629 

This PR introduces the `Backups` screen, which is actually an `Activity Log` screen in disguise. As such, this PR reused the `Activity Log` screen to show the `Backups` screen. The below list of items change when launching an the `Activity Log` for the `Backups` screen:
* The screen title changes to `Backup`.
* The `Activity Type` chip is being hidden.
* The list of activities only contain `Rewindable` items.
* The empty screen state without filter changes to `Your first backup will be ready soon`.
* The empty screen state with filter changes to `No matching backups found`.

screen-20210113-175331 (3)

To test:
* Launch app and go to `My Site` tab.
* Enable `BackupsFeatureConfig`, scroll down and click the `RESTART THE APP` button. **DO NOT** confuse this feature flag with the `BackupFeatureConfig` (singular). Make sure to enable the right feature flag.
* Go back to `My Site` tab and notice `Backup` menu is shown.
* Select the `Activity` menu and verify UI correctness (see above) and that all features work as expected. This is includes but is not limited to:
  * Scrolling infinitely, from top to bottom and vice-versa.
  * Triggering the `Date Range` and `Activity Type` filters.
  * Selecting an activity and navigating to its detail page.
  * Triggering `Restore to this point` and `Download backup` from the menu item.
  * While restoring to a point observe the progress item.
  * Offline support.
  * Configuration changes (rotation, language, etc).
* Select the `Backup` menu and verify UI correctness (see above) and that all features work as expected. This is includes but is not limited to:
  * Scrolling infinitely, from top to bottom and vice-versa. ⚠️ - Scrolling is not optimized yet, try an break it... more on that later
  * Triggering the `Date Range` filter.
  * Selecting an activity and navigating to its detail page.
  * Triggering `Restore to this point` and `Download backup` from the menu item.
  * While restoring to a point observe the progress item.
  * Offline support.
  * Configuration changes (rotation, language, etc).

This solution is based on the investigation notes [here](https://github.com/wordpress-mobile/WordPress-Android/issues/13629#issuecomment-758531526). Below is a summary of the decisions that was made based on that:
* Filter the list of activities after storing them in db. The necessary API enhancement is not available yet and as such filtering it before would create unnecessary overhead and problems later on.
* Do not enhance the scrolling behavior since there will be anyway cases where `filtering after storing in db` approach chosen above would decrease user experience (ie. 1 rewindable item per 20 items per page) or even worse break it altogether (ie. 0 rewindable items per 20 items per page). As such, the API enhancement is unavoidable and it is already being discussed.
* Hide the `Activity Type` filter since the API for this filter is only available for all items. An intermediate solution is still viable, but it has been decided to just hide this filter for now.
* Reuse `Activity Log` screen since by not doing so this will create lots of duplicate code.

Future:
* Add tracking.
* Polish UI.
* Accessibility.
* Design review.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
```diff
@@ -76,6 +87,12 @@ class ActivityLogListFragment : Fragment() {
         } else {
             savedInstanceState.getSerializable(WordPress.SITE) as SiteModel
         }
+        val rewindableOnly = if (savedInstanceState == null) {
```
malinajirka: 🔍 Nitpick: I'm wondering if this `if` statement (and saving the flag in onSaveInstanceState) is needed. If the system re-creates the activity, the extras will still be available in "activity.intent", right? Could we perhaps write something like `nonNullActivity.intent?.getBooleanExtra(ACTIVITY_LOG_REWINDABLE_ONLY_KEY, false) ?: false", wdyt? 
Note: The SiteModel feels a bit different since it can change (ideally we'd store only the siteId, but that's a completely different story). However, the "rewindableOnly" key should be stable, right?

WordPress/src/main/res/values/strings.xml
```diff
@@ -1049,6 +1049,10 @@
 
     <!-- backup -->
     <string name="backup">Backup</string>
+    <string name="backup_empty_title">Your first backup will be ready soon</string>
+    <string name="backup_empty_subtitle">Your first backup will appear here within 24 hours and you will receive a notification once the backup has been completed</string>
```
malinajirka, line 1053: > Your first backup will appear here within 24 hours 

❓ I believe that's the case for "BackupDaily", are we sure it's the case even for other products/plans?

malinajirka (approved):
Really great job @ParaskP7! The differences between Activity Log and Backup Screen are clearly separated ❤️.

> Enable BackupsFeatureConfig, scroll down and click the RESTART THE APP button. DO NOT confuse this feature flag with the BackupFeatureConfig (singular). Make sure to enable the right feature flag.

Tbh I didn't realize these two flags are so similar. I think it might be worth renaming them (eg. BackupScreenFeatureFlag and BackupDownloadFeatureFlag). It'd be quite unfortunate if there was a bug in production and we disabled the wrong flag. Wdyt?

ParaskP7:
> Tbh I didn't realize these two flags are so similar. I think it might be worth renaming them (eg. BackupScreenFeatureFlag and BackupDownloadFeatureFlag). It'd be quite unfortunate if there was a bug in production and we disabled the wrong flag. Wdyt?

I agree! 🌟 

I'll create a PR after this PR is merged which will do exactly that. @zwarm are you okay Jirka's suggestion?

WordPress/src/main/res/values/strings.xml
ParaskP7: Hmmm... I think you are right, nice catch Jirka! 🌟

Adding @osullivanchris and @leandroalonso to this thread to increase transparency and maybe reach to a conclusion faster.

Your thoughts folks?

PS: In the meanwhile, I'll try and find what Calypso is displaying in such cases.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
ParaskP7: You are totally right @malinajirka ! 🌟 

I was trying to be consistent with the `SiteModel` as you figured out yourself, but indeed this is different since the `site` instance can change.

Please expect an update soon.

PS: Just as an FYI, for me this would have been either an Idea (💡) or a Warning (⚠️), most probably the former, depending on how strict we want to be with such cases. Thank you for being kind... 🙏 
PSS: Feel free to be as strict as you need with me. ❤️

ParaskP7: This is now done: 67dfde26c01b334562345df27f9684a8e8d742d6

PS: Also I have update the way this happens on the activity level to reduce boilerplate code: 4594b313da84887864168de0cf37d4b1fb77d7bc

WordPress/src/main/res/values/strings.xml
osullivanchris: Is the question whether that 'first backup will appear here in 24 hours' applies in other cases? If so, to be honest, I don't really know the differences in enough detail to know the answer here! For real-time backups, it could appear even sooner, but within 24 hours is still true. Are there some customers who will never get backups, or get them later than 24 hours?

leandroalonso: > Are there some customers who will never get backups, or get them later than 24 hours?

If we're showing this screen, the blog has the backups feature so the user will either have a backup in 24h or soon.

We can check the capabilities and change this message based on it (that way you can tell if it's `realtime` or `daily`).

I just copied the logic/message from the web, I'm really not sure if it worths it to add that logic just to change a message. If the user just bought the real-time backup it will probably start working right away, so the results will be there.

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
malinajirka: Awesome thanks! ;)

> Just as an FYI, for me this would have been either an Idea (💡) or a Warning (⚠️)

Thanks for the info! I shared it as nitpick since it's quite common to use it the way you did in the codebase. However, I consider it as a one of many small improvements we can make when working on new code.

WordPress/src/main/res/values/strings.xml
malinajirka: If it can't happen at a later time than in 24hours, I'm inclined towards not changing it ;)! I thought we support even one-time backups (BACKUP capability), but perhaps it means something else.

ParaskP7: > For real-time backups, it could appear even sooner, but within 24 hours is still true.

> If the user just bought the real-time backup it will probably start working right away, so the results will be there.

This actually makes sense, considering the above I would let is be as it is now.

@malinajirka your thoughts?

PS: Thanks for jumping in @osullivanchris @leandroalonso ! ❤️

malinajirka (approved):

malinajirka:
Feel free to merge this when the conflict is resolved ;)!

ParaskP7:
> I'll create a PR after this PR is merged which will do exactly that. @zwarm are you okay Jirka's suggestion?

FYI: This PR is now ready @malinajirka : #13785

PR #13759, opened 2021-01-13
Add restore feature flag

zwarm:
Fixes #13328 

This PR add a local feature flag for the Jetpack Restore feature. This is the precursor step to using the feature flag during development.

There is nothing to test in this PR.

PR submission checklist:

- [X] I have considered adding unit tests where possible.
- [X] I have considered adding accessibility improvements for my changes.
- [X] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #13328, opened 2020-11-09
Jetpack Section: Restore


zwarm:
Parent #13267 

<img width="1779" alt="Screen Shot" src="https://user-images.githubusercontent.com/506707/104459246-7c04ea00-557a-11eb-96eb-a830d82fe339.png">

**Design**
TBD

**Tasks**

- [x] Introduce restore request body to endpoint ([FluxC PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1761))
- [x] FluxC - Update endpoints to include message and current_entry for restore 
- [x] Replace rewind status service with restore use cases #13902
- [x] Hook wizard back into Activity Log List #13914
- [x] Fix dark theme issue with the progress row #13874
- [x] Add feature flag check to Activity Log Detail view to launch wizard #13979
- [x] Add error logging anywhere an exception is deliberately thrown #13910 
- [x] Implement error 3X retry on fetch within GetRestoreUseCase #13930 
- [x] Change "additional information" names to `footnote` to better describe the item #13912
- [x] On unchecking all checkbox items on the "Details" screen, we might want to disable the action button.  #13957
- [x] Clear DB on null Fetch from API to support edge case where we are caught between running and complete. FluxC [PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1860) and WPAndroid [PR](https://github.com/wordpress-mobile/WordPress-Android/pull/13944)
- [x] Remove the "Notify me" button from progress view Issue #14088 

#### Steps
- [x]  Restore Details (choose items to restore) [PR](https://github.com/wordpress-mobile/WordPress-Android/pull/13773) 
- [x]  Restore Warning PR (https://github.com/wordpress-mobile/WordPress-Android/pull/13775)
- [x]  Restore In Progress [PR](https://github.com/wordpress-mobile/WordPress-Android/pull/13778) 
- [x]  Restore Complete/Done [PR](https://github.com/wordpress-mobile/WordPress-Android/pull/13779)

#### Design
- [x]  Restore Details (choose items to backup)
- [x]  Restore Warning
- [x]  Restore In Progress
- [x]  Restore Complete/Done
- [x]  Update copy on all screens #13919 
- [x] Error view - Status vs. general layout (see iOS) #13919 

#### Fixes 
[...]

PR #13764, opened 2021-01-13
Issue/13328 restore skeleton

zwarm:
Parent #13328 

This PR introduces the "Restore" feature as accessed via the activity log list. This is prep PR.
 
`RestoreActivity`, `RestoreViewModel` are copies of the `BackupDownloadActivity` and `BackupDownloadViewModel` and functionality for observables and the wizard remain the same. Differences include supporting class renames, as well as function name changes. I have left //todo's to fill in particulars for each step as I include them.

`RestoreDetailsFragment` and `RestoreDetailsViewModel` are mere skeletons because it would have been too much to review. 

I did not use the `Rewind` package, but created a new one for `Restore`. I didn't rename because the current classes in `Rewind` are used by the ActivityLog. These `rewind` classes may be removed in the future and if so, this package will be deleted in its entirety.

This PR is a bit larger than normal, but to show that it works I had to go all the way to presenting the details view. It's best reviewed by commits which I have broken down as follows:

- Dagger & App Housekeeping: 9e4ad1b 
- Feature Flag in ActivityLog: ad0cc55
-- Adds support for showing the new restore wizard when the `RestoreFeatureConfig` is enabled. This will work with or without the BackupFeatureConfig flag.
- Restore Activity/Details: 723852c
-- This includes the cloned classes from BackupDownload feature
- Skeleton Unit Tests: 4f9d98b

**Merge Instructions**
- <del>Make sure #13759 has been merged<del> ✅
- <del>Remove Not Ready For Merge Label<del> ✅
- Merge as normal

**To test:**
- Launch the app with RestoreFeatureConfig flag on and the BackupFeatureFlag off
- Navigate to ActivityLog
- Tap the rewind icon for an item
- Note that an empty screen with the "Restore" title is shown

- Navigate to ActivityLog
- Tap the More menu
- Tap Restore to this point menu item
- Note that an empty screen with the "Restore" title is shown

- Launch the app with RestoreFeatureConfig flag off and the BackupFeatureFlag off
- Navigate to ActivityLog
- Tap the rewind icon for an item
- Note that the restore dialog box is shown

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
@@ -40,6 +40,7 @@ import javax.inject.Inject
 private const val ACTIVITY_TYPE_FILTER_TAG = "activity_log_type_filter_tag"
 private const val DATE_PICKER_TAG = "activity_log_date_picker_tag"
 private const val BACKUP_DOWNLOAD_REQUEST_CODE = 1710
+private const val RESTORE_REQUEST_CODE = 1720
 
 class ActivityLogListFragment : Fragment() {
     @Inject lateinit var viewModelFactory: ViewModelProvider.Factory
@@ -160,7 +161,16 @@ class ActivityLogListFragment : Fragment() {
 
         viewModel.showRewindDialog.observe(viewLifecycleOwner, {
             if (it is ActivityLogListItem.Event) {
-                displayRewindDialog(it)
+                if (it.launchRestoreWizard) {
+                    ActivityLauncher.showRestoreForResult(
+                            requireActivity(),
+                            viewModel.site,
+                            it.activityId,
+                            RESTORE_REQUEST_CODE
+                    )
+                } else {
+                    displayRewindDialog(it)
+                }
             }
         })
 
@@ -183,10 +193,21 @@ class ActivityLogListFragment : Fragment() {
                             viewModel.site,
                             event.activityId,
                             BACKUP_DOWNLOAD_REQUEST_CODE)
-                    // todo: annmarie replace with the ActivityLauncher for showing restore details
-                    is ShowRestore -> displayRewindDialog(event) }
+                    is ShowRestore -> {
+                        if (event.launchRestoreWizard) {
+                            ActivityLauncher.showRestoreForResult(
+                                    requireActivity(),
+                                    viewModel.site,
+                                    event.activityId,
+                                    RESTORE_REQUEST_CODE
+                            )
+                        } else {
+                            displayRewindDialog(event)
+                        }
+                    }
                 }
-            })
+            }
+        })
     }
 
     private fun displayRewindDialog(item: ActivityLogListItem.Event) {
--- a/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
@@ -23,6 +23,7 @@ import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.ui.ActivityLauncher
 import org.wordpress.android.ui.activitylog.ActivityLogNavigationEvents.ShowBackupDownload
 import org.wordpress.android.ui.activitylog.ActivityLogNavigationEvents.ShowRestore
+import org.wordpress.android.ui.activitylog.ActivityLogNavigationEvents.ShowRewindDialog
 import org.wordpress.android.ui.activitylog.list.filter.ActivityLogTypeFilterFragment
 import org.wordpress.android.ui.posts.BasicFragmentDialog
 import org.wordpress.android.ui.utils.UiHelpers
@@ -159,21 +160,6 @@ class ActivityLogListFragment : Fragment() {
             }
         })
 
-        viewModel.showRewindDialog.observe(viewLifecycleOwner, {
-            if (it is ActivityLogListItem.Event) {
-                if (it.launchRestoreWizard) {
-                    ActivityLauncher.showRestoreForResult(
-                            requireActivity(),
-                            viewModel.site,
-                            it.activityId,
-                            RESTORE_REQUEST_CODE
-                    )
-                } else {
-                    displayRewindDialog(it)
-                }
-            }
-        })
-
         viewModel.showSnackbarMessage.observe(viewLifecycleOwner, { message ->
             val parent: View? = activity?.findViewById(android.R.id.content)
             if (message != null && parent != null) {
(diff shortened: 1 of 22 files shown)


WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
```diff
@@ -183,10 +193,21 @@ class ActivityLogListFragment : Fragment() {
                             viewModel.site,
                             event.activityId,
                             BACKUP_DOWNLOAD_REQUEST_CODE)
-                    // todo: annmarie replace with the ActivityLauncher for showing restore details
-                    is ShowRestore -> displayRewindDialog(event) }
+                    is ShowRestore -> {
```
ParaskP7: Idea (💡): I wonder if moving the `if/else` logic to the view model would be better? This will help reduce the complexity on the activity level (keep it as dump as possible) and help with testing this logic as well. Wdyt? For example, having `ShowRestoreScreen` and `ShowRestoreDialog` out of `ShowRestore`, which will be then driven by the view model.

PS: Same applies for the `if/else` logic further above on the `showRewindDialog` live data object.
PSS: I know that this logic would be eventually removed at some point, but until them and since this will land on `develop`, maybe it is worth the extra effort.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreViewModel.kt
```diff
@@ -0,0 +1,172 @@
...
+    fun writeToBundle(outState: Bundle) {
+        outState.putInt(KEY_RESTORE_CURRENT_STEP, wizardManager.currentStep)
+        outState.putParcelable(KEY_RESTORE_STATE, restoreState)
+    }
+
+    fun onBackPressed() {
+        when (wizardManager.currentStep) {
+            DETAILS.id -> {
+                _wizardFinishedObservable.value = Event(RestoreCanceled)
+            }
```
ParaskP7, line 110: Question (❓): Why not adding the `PROGRESS.id` and `COMPLETE.id` cases to match that with the `BackupDownload` equivalent view model logic? Especially since the `RestoreInProgress` and `RestoreCompleted` objects are already there? Will that logic change for the `Restore` screen?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreViewModel.kt
```diff
@@ -0,0 +1,172 @@
...
+
+        @Parcelize
+        object RestoreCompleted : RestoreWizardState()
+    }
+
+    sealed class ToolbarState {
+        abstract val title: Int
+        abstract val icon: Int
+
+        data class DetailsToolbarState(
```
ParaskP7, line 167: 1) Question (❓): This is currently unused. What if this is being used by the `RestoreDetailsViewModel`, just like it happens for the `BackupDownload` equivalent code. Will it be used as such in the future?
2) Question (❓): Since the `DetailsToolbarState` is added, I am guessing that more of that would become available later on in the future. If my assumption is correct, why not add all of them now even if the strings/icons are not known yet?

WordPress/src/main/res/layout/restore_details_fragment.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
```
ParaskP7, line 6: 1) Minor (🔍): This layout file can be reformatted.
2) Question (❓): On the `BackupDownload` equivalent layout file, there is an unused `android:id` and an extra `android:padding`. I am guessing we can remove the unused `android:id` from there to increase consistency between the two layout files. However, I am not sure about the `android:padding` and whether this is required for the `Restore` case.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/restore/RestoreDetailsViewModelTest.kt
```diff
@@ -0,0 +1,26 @@
+package org.wordpress.android.ui.jetpack.restore
```
ParaskP7, line 1: Warning (⚠️): This test resides in the wrong package. Consider moving that to an inner `details` package to match that with the `main` source set structure.

ParaskP7 (commented):
👋 @zwarm !

I have review and tested this PR as per the instructions. Everything works as expected, great work here! 🌟 

1) Question (❓): Should there be a third entry point in addition to the two that got updated based on the feature flag. This third entry point would be from within the `ActivityLogDetailsFragment`, see `onRewindButtonClicked(...)`. Should this be changed to navigate to the newly added `Restore` screen as well? That is, instead of showing the `Restore Site` dialog?
2) Info (ℹ️): Most of my comments below was based on my diffing between the existing `BackupDownload` and new `Restore` files, this is how I reviewed this PR. As such, the questions (❓) I raised are based on that fact. Bear that in mind when reviewing them further below.

PS: Let me know how to process and I will approve the PR afterwards, there are no breaking changes, mainly minors that I want to give you a change to fix before merging this PR to `develop`.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/restore/RestoreDetailsViewModelTest.kt
zwarm: I didn't put any of the tests into packages. I shall task this separately because it would need to be done for backup download as well. Thanks.

WordPress/src/main/res/layout/restore_details_fragment.xml
zwarm: Great comment - I'll be focusing on styling next week. I am aware of the padding already. :)

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreViewModel.kt
zwarm: 1. It will be .. just the prep PR.
2. Trying to keep the pr's code relevant to specific topics. The other toolbar states will be added with each step implementation.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreViewModel.kt
zwarm: They will be added as I move onto those steps - tried to keep this PR focused to just enough so that restore is shown

WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/ActivityLogListFragment.kt
zwarm: > I know that this logic would be eventually removed at some point

Lol, but great idea. Let me implement and push the commit back. 👍

zwarm: @ParaskP7 in commit e081fe2. Can you give it a whirl

ParaskP7: Love it @zwarm ! 🌟 

PS: Since now we have additional flows on the view model, mainly creating a couple of test for those flows would be beneficial, wdyt? Just copy-paste the existing tests and add the differentiator factor. This way we would have full test coverage for this feature.

ParaskP7:
> Question (❓): Should there be a third entry point in addition to the two that got updated based on the feature flag. This third entry point would be from within the ActivityLogDetailsFragment, see onRewindButtonClicked(...). Should this be changed to navigate to the newly added Restore screen as well? That is, instead of showing the Restore Site dialog?

How about this question @zwarm ?

zwarm:
> How about this question @zwarm ?

It's on my task list already as a separate item. :)

ParaskP7 (approved):
Approving now, thanks for the updates and for making sure that those changes will happen on a subsequent PR @zwarm ! 🌟

PR #13765, opened 2021-01-13
Merge/release 16.5 into develop

loremattei:
This PR merges some fixes and the new release notes for 16.5 into `develop`.

@malinajirka this PR adds a fix which landed the release branch and that requires a new beta of FluxC. I built `1.9.0-beta-4` on top of your `beta-3`, but I see that it was not referenced in `develop` yet.
Is it ok to include the changes in `beta-3` now? 
Otherwise I can revert this PR to use `1.9.0-beta-2` and we can bring in `beta-4` later.

PR submission checklist:


mokagio (approved):
This looks good from a release point of view. 

Holding off merging to hear what @malinajirka has to say.

malinajirka:
The version was actually created by @ParaskP7, I'm written on the merge commit since I hit the PR merge button. Either way, I believe it's safe to merge the version as it doesn't change the current behavior - it just adds support for some additional parameters which will be used on the new Jetpack Backup screen. Thanks for double-checking!

PR #13766, opened 2021-01-14
Adding filtering feature to filterables tabs in reader

### LINKED ISSUE #13669: Adding filtering capability to a8c and p2s reader tabs
As part of the #13514 we should add the subfiltering capability to `Automattic` and `P2s` tabs category in reader; this should behave in a similar way like the filter in `Following` 

| image | image |
|---|---|

Only difference should be:
- The bottom sheet should contain only `SITES` and not `TOPICS` page
- The sites listes should be those relevant to the `AUTOMATTIC` and `P2S (not a8c p2s)` streams only

develric:
Fixes #13669 

This PR supersedes the #13728 and #13734 previous iterations. cc @malinajirka since you knows the previous iterations and we had also a pair programming session on them 🙇 (it's more like a FYI if you do not have time to have a look here, thanks 😄)

This PR introduces a filterable and organization fields to the `ReaderTag`. Currently the below are filterable:

| Tab page | Organization | Subfilter available |
|---|---|---|
| FOLLOWING | NO_ORGANIZATION | image |
| FOLLOWED P2S | P2 | image |
| AUTOMATTIC | A8C| image |

Note that the `TOPICS` is only available in the `FOLLOWING` bottom sheet since it makes sense only in that scope.

Moreover:
- Introduced the concept of isFilterable vs isFollowing and the functions isFilterableScreen vs isFollowingScreen (this isFollowingScreen is still used in a few places where I thought it makes sense)
- The SubfilterViewModel and the Bottom Sheet is now not shared but instantiated per ReaderPostListFragment tab page. Also the Bottom Sheet is hooked to the ReaderPostListFragment page.
- It was decided with Design to keep current filter per each tab in memory but no need to keep it when restarting the app. I reused the SubfilterListItemMap (that was used to save current filtering information into the shared prefs); in this case that is used to save into the fragment instance state. This is necessary since the fragment (and VM that is owned by it) can eventually be reinstantiated when swiping through the Reader tabs.

## To test
Please note that about unit testing, in this PR I basically fixed them to account for the changes. I will add some more coverage for the `SubFilterViewModel` in a dedicated PR.

Testing goal should be to check that everything works as before but you can additionally filter correctly not only in Following but also in P2s and A8c. Some scenarios to take care I can think of:

- User has only self hosted sites -> smoke test reader (the bottom sheet in following should offer a CTA to login)
- Fresh install (nothing in cache), create a new wp.com profile, without creating a site go to the reader and smoke test
- From the step before, create a new site and smoke test the reader
- Smoke test both from non A8C and from A8C account
- When swiping tabs far from left/rigth and vice versa check the filter selection in the 3 filterable tabs are kept. This should apply also if you place the app in background and come back to it. Restarting the app should reset the filters.
- Adding site by url (related to each of the 3 filterable tabs) you should get that new added site in the relevant subfilter when opening it

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
```diff
@@ -251,22 +253,19 @@ void saveInstance(Bundle bundle) {
...
-        return newInstanceForTag(tag, listType, false);
+        return newInstanceForTag(tag, listType, false, -1);
     }
 
-    static ReaderPostListFragment newInstanceForTag(ReaderTag tag, ReaderPostListType listType, boolean isTopLevel) {
+    static ReaderPostListFragment newInstanceForTag(
+            ReaderTag tag,
+            ReaderPostListType listType,
+            boolean isTopLevel,
+            int position
```
khaykov: I don't have all the context, but It looks like we are using position only to specify a unique a VM name. Do you think we can use something else for this (eg. `ReaderTag`?). Ideally the Fragment should not be aware of it's position in `ViewPager`.

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubFilterViewModel.kt
```diff
@@ -72,46 +78,64 @@ class SubFilterViewModel @Inject constructor(
...
     private var isStarted = false
     private var isFirstLoad = true
+    private var shouldRequestNewerPosts = true
+    private var mainStreamTag: ReaderTag? = null
 
     /**
      * Tag may be null for Blog previews for instance.
      */
-    fun start(tag: ReaderTag?) {
+    fun start(mainStreamTag: ReaderTag?, tag: ReaderTag?, subfilterJson: String?) {
```
khaykov: Not sure what is the `mainStreamTag` in this context :)

khaykov (commented):
Thanks for the refactoring, Riccardo! Way less code right now :)
I can also confirm that the issue with progress indicator and followed sites not being reflected in bottom sheet are fixed.

I also just realized that when you filter by a site, it's post cards do not have a dropdown menu button on the top right? I think this is a preexisting behavior, but I wonder what is the reason for this?

Another things is Discover. I'm wan't following updates to reader too much, but It looks like The Discover tab now shows tags you are following, and not the Discover feed. You can still follow Discover feed, and it will appear in Following screen, but it will not appear in the filter. Not sure this is an expected behavior.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
develric: Makes sense 👍 . Removed the mPosition field and using the primary db key of the main reader tag of the tab 28fe95a349766b818ac617a213eba21cca9f4435

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubFilterViewModel.kt
develric: 😄 . Maybe not clear sorry. In my intention it is the tag that is used to initially open the tab (the `mTagFragmentStartedWith` in ReaderPostListFragment; note that for how this fragment works, the `mTagFragmentStartedWith`is not always the same as `mCurrentTag` but it is nonetheless the tag used to recover the main full stream of posts. Not sure it clarifies what was my intention, let me know 😄 .

develric:
>Thanks for the refactoring, Riccardo! Way less code right now :)

🙇 

>I also just realized that when you filter by a site, it's post cards do not have a dropdown menu button on the top right? I think this is a preexisting behavior, but I wonder what is the reason for this?

Yep it seems something that was already there; not sure of the reason; if you agree I can open a GH issue for this and we can investigate with other folks if there is a reason or it was just being overlooked; let me know 😄 

>Another things is Discover.

Good catch 👍 ! I think we were filtering out the Discover feed `https://discover.wordpress.com` because it was the one we were visualizing in the Discover tab, but as you spotted does not make sense anymore and we should allow to list it in the following filters. Changed here f6d6d2905216f6923f442953bfaeb00f2d291133

Also added an extra commit here 1e53d993a5bbe0686289a0436cc16152ec0a3145 to fix the possible reset of the recycler adapter of the list of posts when ReaderPostListFragment is reinstantiated (swipe from far left tab to far right tab and vice versa). Basically we were not saving the VM `isFirstLoad` state in the fragment instance state. Doing it is also enough to get the progress indicator issue fixed, so I removed the `shouldRequestNewerPosts` in the SubfilterViewModel. 

Hope the above is clear enough 🙃, let me know wdyt 🙇

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubFilterViewModel.kt
khaykov: Got it! `mTagFragmentStartedWith` is pretty self explanatory, so I wonder if we could use the same or similar name here?

khaykov (approved):
Thanks for all the fixes and refactoring, Riccardo! I wasn't able to find any particular issues during testing, and code looks good to me 👍 

I left a comment about naming of `mainStreamTag`, but feel free to ignore it :)

WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubFilterViewModel.kt
develric: Yep, get your point 👍 ! Kept same name here 3d4c2ce

PR #13767, opened 2021-01-14
Issue/13326 ignore threat usecase

ashiagr:
Parent #13768 

This PR adds ignore threat use case.

To test:
That `IgnoreThreatUseCaseTest` tests run fine and cover action results.

PR submission checklist:

### DIFF
--- a/WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/details/usecases/IgnoreThreatUseCaseTest.kt
+++ b/WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/details/usecases/IgnoreThreatUseCaseTest.kt
@@ -32,7 +32,6 @@ class IgnoreThreatUseCaseTest : BaseUnitTest() {
     fun setup() = test {
         useCase = IgnoreThreatUseCase(networkUtilsWrapper, scanStore, TEST_DISPATCHER)
         whenever(networkUtilsWrapper.isNetworkAvailable()).thenReturn(true)
-        whenever(scanStore.ignoreThreat(any())).thenReturn(OnIgnoreThreatStarted(IGNORE_THREAT))
     }
 
     @Test
@@ -57,6 +56,8 @@ class IgnoreThreatUseCaseTest : BaseUnitTest() {
 
     @Test
     fun `when ignore threat is triggered, then Success is returned on success`() = test {
+        whenever(scanStore.ignoreThreat(any())).thenReturn(OnIgnoreThreatStarted(IGNORE_THREAT))
+
         val result = useCase.ignoreThreat(fakeSiteId, fakeThreatId)
 
         assertThat(result).isEqualTo(Success)
--- a/WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/details/usecases/IgnoreThreatUseCaseTest.kt
+++ b/WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/details/usecases/IgnoreThreatUseCaseTest.kt
@@ -1,0 +1,64 @@
+package org.wordpress.android.ui.jetpack.scan.details.usecases
+
+import com.nhaarman.mockitokotlin2.any
+import com.nhaarman.mockitokotlin2.whenever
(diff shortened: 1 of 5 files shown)


WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/details/usecases/IgnoreThreatUseCaseTest.kt
```diff
@@ -0,0 +1,64 @@
...
+    @Mock lateinit var scanStore: ScanStore
+
+    private val fakeSiteId = 1L
+    private val fakeThreatId = 11L
+
+    @Before
+    fun setup() = test {
+        useCase = IgnoreThreatUseCase(networkUtilsWrapper, scanStore, TEST_DISPATCHER)
+        whenever(networkUtilsWrapper.isNetworkAvailable()).thenReturn(true)
+        whenever(scanStore.ignoreThreat(any())).thenReturn(OnIgnoreThreatStarted(IGNORE_THREAT))
```
ParaskP7: Minor (🔍): Consider moving this `scanStore` mock setup to the last test and have it there explicitly. Below my reasoning:
1) This mock setup and the `networkUtilsWrapper` mock setup are not the same. In the network care we might want to apply the default to be `true`, which will be applicable to all tests, and have it only being overitten by a small number of tests (in that case just one). This is very valid in order to avoid duplication and increase readability.
2) However, for the `scanStore` case, by adding the default behavior to the `success` path, it also seems that we want to apply that to all test, but this is not true as this mock setup is only required for one test, thus it can make things ambiguous.
3) By having the `scanStore` mock setup explicitly in the last test, it can be compared easily with the above `failure` path and thus increase test coherence and consistency.

PS: Please feel free to ignore me. I am just writing down my thoughts here as an FYI since this subject is indeed very difficult and subjective. I just wanted to share my thoughts so that you know what I would have done if I where to implement these tests.

ParaskP7 (approved):
👋 @ashiagr !

I have reviewed and tested this PR as per the instructions. All looks good to me, nice work, especially with the tests! 🌟 

I have only minor test related comment(s). I am going to approve this PR but give you the change to apply any comments before merging this to `develop`. Feel free to merge it as is if you like.

PS: What would you say if we added a test for the `GetThreadModelUseCase`? I know this use case is very simple, as it will only verify that the `scanStore` is triggered, but even that is a valid test IMHO and can help setting up the testing structure so that when in the future this become more complicated the test a) might fail and b) be easier to enhance (since the test suite is already available). Wdyt?

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/details/usecases/IgnoreThreatUseCaseTest.kt
ashiagr: Thanks for sharing your thoughts 🙏. Completely valid arguments 🌟.
Moved in 0d652fd

ParaskP7: Double approving this PR then @ashiagr ! 😛 

Feel free to merge it anytime you are ready. ✅

ashiagr:
Thanks for the review @ParaskP7! You always come up with thoughtful suggestions!

> PS: What would you say if we added a test for the GetThreadModelUseCase? I know this use case is very simple, as it will only verify that the scanStore is triggered, but even that is a valid test IMHO and can help setting up the testing structure so that when in the future this become more complicated the test a) might fail and b) be easier to enhance (since the test suite is already available). Wdyt?

Added to the backlog, will surely consider adding it, thank you!

ParaskP7:
> Thanks for the review @ParaskP7! You always come up with thoughtful suggestions!

You are very kind @ashiagr ! Thank you for considering my suggestions. 🤗 PS: When you find yourself under the clock, feel free to skip any of those. My intention is mainly discussing the subject, whether that is applied or not afterwards is secondary to me.

> Added to the backlog, will surely consider adding it, thank you!

Thank YOU! ❤️

PR #13768, opened 2021-01-14
Issue/13326 ignore threat

ashiagr:
Parent #13326 

This PR handles ignore threat action on the threat details screen.
Child PR for `IgnoreThreatUseCase`: #13767

It mimics behaviour on the web:
1. Displays a confirmation dialog 
2. Disables buttons while action is taking place
3. Displays success/ error messages when action completes
4. Refreshes scan state when ignore threat action is successful

https://user-images.githubusercontent.com/1405144/104564678-37a84580-5671-11eb-8893-266e40bf6554.mov

https://user-images.githubusercontent.com/1405144/104564816-60c8d600-5671-11eb-8c6a-d4b9084eb60e.mov

To test:
Prerequisite: 

- You have access to WP.com account with a site having scan capability and threats (e.g. https://pressable-jetpack-daily-scan.mystagingwebsite.com). Server credentials should be set on the site.
- I've added a threat to the site. In case you run out of threats, re-install bad calendar plugin from [here](https://jetpackp2.files.wordpress.com/2020/05/calendar-1.zip) and re-run scan.

1. Login to the app using above WP.com account and select the site on the My Site tab
2. Click scan menu item on My Site
3. Notice that scan state is fetched and shown with threats
4. Click a threat 
5. Click ignore threat button on the threat details screen
6. Notice that a _confirmation dialog is shown_ (see video)
7. Click ok in the confirmation dialog
8. Notice that _action buttons are disabled_
8. Wait for some time till ignore threat action completes
9. Notice that _success message "Threat ignored." is shown_  
10. Notice that _update to scan state is triggered after ~2 secs

Repeat above steps and enable airplane mode after step 6. Notice that an error message is displayed and actions buttons are re-enabled.

(Tests included in `ThreatDetailsViewModelTest` )

**Notes**
1. Ignore all styling
2. Server credentials check is not handled
3. Loading state while action is taking place - yet to be added

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/details/ThreatDetailsViewModel.kt
```diff
@@ -1,44 +1,143 @@
...
+
+                    withContext(bgDispatcher) { delay(DELAY_MILLIS) }
+                    updateNavigationEvent(ShowUpdatedScanState)
+                }
+
+                is IgnoreThreatState.Failure -> {
+                    disableThreatActionButtons(false)
+                    val message = UiStringRes(R.string.threat_ignore_error_message)
+                    updateSnackbarMessageEvent(SnackbarMessageHolder(message))
+                }
```
ashiagr, line 91: I have to break this down into network and remote request errors.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
```diff
@@ -81,6 +84,15 @@ class ScanFragment : Fragment(R.layout.scan_fragment) {
         }
     }
 
+    override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
+        super.onActivityResult(requestCode, resultCode, data)
+        if (requestCode == RequestCodes.SHOW_THREAT_DETAILS &&
+            data?.getBooleanExtra(ThreatDetailsFragment.REQUEST_SCAN_STATE, false) == true
+        ) {
+            viewModel.fetchScanState()
```
malinajirka: 🔍 Nitpick: We might want to consider adding "onScanStateRequested" method into the VM instead of calling an action (fetch) from the view layer. It's not a big deal in this case, but I think it's good practice so other devs won't feel tempted to start calling actions from the view layer even in other cases.

```
if (requestCode == RequestCodes.SHOW_THREAT_DETAILS &&
            data?.getBooleanExtra(ThreatDetailsFragment.REQUEST_SCAN_STATE, false) == true
        ) {
            viewModel.onScanStateRequested()
}
```

```
fun onScanStateRequested() {
     fetchScanState()
}
```

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/details/ThreatDetailsViewModel.kt
```diff
@@ -1,44 +1,147 @@
...
+    }
+
+    private fun ignoreThreat() {
+        viewModelScope.launch {
+            disableThreatActionButtons(true)
+            when (ignoreThreatUseCase.ignoreThreat(site.siteId, threatId)) {
+                is IgnoreThreatState.Success -> {
+                    updateSnackbarMessageEvent(UiStringRes(R.string.threat_ignore_success_message))
+
+                    withContext(bgDispatcher) { delay(DELAY_MILLIS) }
```
malinajirka, line 84: ❓ Are we adding this delay here, so the Snackbar message doesn't disappear right after we show it?

⚠️ If that's the case, this won't work as expected with TalkBack - I believe that delay for snackbar messages when TalkBack is on is set to 10s. Have you considered showing the message on the Scan screen instead of the ThreatDetail screen? I think it'd feel a bit better even for users who don't use TalkBack, wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/details/ThreatDetailsViewModel.kt
```diff
@@ -1,44 +1,147 @@
...
+                ),
+                okButtonAction = this@ThreatDetailsViewModel::ignoreThreat
+            )
+        )
+    }
+
+    private fun onGetFreeEstimateButtonClicked() { // TODO ashiagr to be implemented
+    }
+
+    private fun disableThreatActionButtons(disable: Boolean) {
```
malinajirka: 💡 Idea: We might want to consider renaming this to "updateThreatActionButtons" or similar. It feels a bit weird that the method is called "disable" but it actually enables the items when you pass "false". Wdyt?

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/details/ThreatDetailsViewModelTest.kt
```diff
@@ -11,51 +11,213 @@ import org.junit.Test
...
+                    )
+            )
+            val expectedPositiveButtonLabel = R.string.dialog_button_ok
+            val expectedNegativeButtonLabel = R.string.dialog_button_cancel
+            val observers = init()
+
+            (observers.uiStates.last() as Content).items.filterIsInstance<ActionButtonState>()
+                .first().onClick.invoke()
+
+            val confirmationDialogAction = observers.navigation.last().peekContent() as OpenThreatActionDialog
```
malinajirka: 💡 Idea: We might want to consider splitting this method into 2 methods. 

First method would assert that ` observers.navigation.last().peekContent()` is instance of `OpenThreatActionDialog` == `when ignore threat button is clicked, then action to open ignore action confirmation dialog is triggered`.

Second method would assert that the dialog shows the expected content.

I agree that splitting it into 5 methods (assert title, message, positive btn, .. ) is an overkill, but it feels like the current method tests two different things - correct navigation + correct content. Wdyt?

malinajirka (commented):
Thanks @ashiagr!

I've left a few minor comments. One thing I noticed is that the confirmation dialog disappears after rotation - it's not a big deal, but we might want to fix it (feel free to add it into the backlog). Let me know what you think ;).

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
ashiagr: Good point. In 2d0991d.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/details/ThreatDetailsViewModel.kt
ashiagr: Sounds good, updated in 8429c5a

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/details/ThreatDetailsViewModelTest.kt
ashiagr: Nice idea 👍. Split done in 03c1f74.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/details/ThreatDetailsViewModel.kt
ashiagr: > Are we adding this delay here, so the Snackbar message doesn't disappear right after we show it?

Correct, delay was added with this 👆intention.

> If that's the case, this won't work as expected with TalkBack - I believe that delay for snackbar messages when TalkBack is on is set to 10s. 

TIL. Thanks for sharing it with me 🙏.

> Have you considered showing the message on the Scan screen instead of the ThreatDetail screen? 

Makes sense, I've added a TODO in 6ee3863. I'm thinking to refactor it with fix threat action which will re-use fix all threats implementation in the scan state screen.

Let me know if it's fine to handle it in a PR separate to this one.

ashiagr:
Thanks so much for the review @malinajirka! There were nice ideas that improved code readability. Thank you!

>  One thing I noticed is that the confirmation dialog disappears after rotation - it's not a big deal, but we might want to fix it (feel free to add it into the backlog).

Sounds good, added it to the backlog.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/details/ThreatDetailsViewModel.kt
malinajirka: Sounds good to me, thanks!

malinajirka (approved):
Thanks for the changes! I've checked the code and smoke tested the app and it LGTM. Feel free to merge this when the CI finishes.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/details/ThreatDetailsViewModel.kt
ashiagr: Resolved in https://github.com/wordpress-mobile/WordPress-Android/pull/13825

PR #13770, opened 2021-01-14
Update release-toolkit to fix build errors when checking new strings

loremattei:
# Background
Since we started to download and check translations as a part of the beta release, we have had an issue in the `new_beta_release` lane where the build step failed because of the weird behaviour of the Gutenberg submodule.
The lane checks for the repository to be clean at start, but for reasons the Gutenberg submodule is marked as dirty as soon as `git submodule update --init --recursive` is run. For this reason, we have to reset the submodule before starting the lane. But without the module initialized, the lint fails later when trying to check the new translations.

# Changes in this PR
In order to fix the issue, this PR updates `release-toolkit` to a new version which runs `git submodule update --init --recursive` before invoking the lint task. 

Moreover, when looking into this issue, I found that the lint task is currently run twice: the first time by the `update_metadata` helper in `relslease_toolkit` and the second by a call to `release-checks.sh`. 
Looking into `release-checks.sh`, I found that the only test which it still runs is actually the string check (this is true in all the Android repos), so I removed the call to `release-checks.sh` in `release-toolkit` and the script itself in this PR.

The related `release-toolkit` PR is https://github.com/wordpress-mobile/release-toolkit/pull/186.

# To test:
Testing the `new_beta_release` lane can be tricky because it pushes the changes to the release branch at every step and, as part of the prechecks, it makes sure to be on the current release branch. 
Since the changes in this PR only affects the `android_update_metadata` action, the easiest and safest way is probably to comment out all the other actions in the lane: 

- Check out a new branch from this one. 
- Comment all the actions in `new_beta_release` other than `android_update_metadata`.
- Commit the changes above and make sure the repository is clean (reset the `mobile-gutenberg` submodule if required).
- Run `bundle install`.
- Run `bundle exec fastlane new_beta_release`.
- Verify that the new translations are downloaded, the `mobile-gutenberg` submodule initialized and the lint run. 

_Note_: A new version of the `release-toolkit` has to be released before merging this PR.

PR submission checklist:


jkmassel (approved):
✅ Verify that the new translations are downloaded
✅ Verify that the mobile-gutenberg submodule initialized and the lint run.

PR #13771, opened 2021-01-14
Use single and partial states in order to be able to add more sources

planarvoid:
In order to implement quick start logic I've found out I need to add more sources to the generated UiModel. Since I fear we'd have to do this more than once for the future logic, I'm introducing a `scan` function and a single `UiState`. All the sources are now mapped to `PartialState` and the scan function updates the `UiState`. This `UiState` is transformed into the `UiModel`. I'm not 100% sure about the `scan available` functionality but it should be ok. Let me know what you think about this approach.

To test:
- The tests are passing so I'd smoke test the my site screen.

PR submission checklist:


renanferrari (approved):
Hey @planarvoid, thanks for the changes! I really like this approach, it's much more elegant than what we had. 

I've smoke tested the app and as far as I can see everything seems to be working as expected.

LGTM! :shipit:

PR #13773, opened 2021-01-14
Issue/13328 restore details

zwarm:
Parent #13328

This PR implements the "details" view (a.k.a. the select options) for the restore request.

Although RestoreDetailsViewModel was cloned off of BackupDownloadDetailsViewModel, the logic deviates because of the introduction of the Warning step in restore. 

**Cloned Classes**
RestoreListItemState
RestoreDetailsAdapter
RestoreDetailsFragment
RestoreSubHeaderViewHolder
RestoreDetailsStateListItemBuilder
restore_list_subheader_item.xml 

Notes:
- Ignore styling - this is being handled separately 
- Ignore unit test placement in packages - this is being handled separately 
- RestoreActivity & RestoreViewModel (+unit tests) will continue to get updated as I move throughout the steps of the wizard

To test:
- Launch the app with RestoreFeatureConfig flag on and the BackupFeatureFlag on
- Navigate to ActivityLog
- Tap the more menu
- Select the restore option
- Note (1) that the detail view is shown (2) the checkboxes can be selected/unselected, tapping on X or back button closes the view

PR submission checklist:

- [X] I have considered adding unit tests where possible.
- [X] I have considered adding accessibility improvements for my changes.
- [X] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13775, opened 2021-01-14
Issue/13328 restore warning

zwarm:
Parent #13328

This PR implements the "warning" wizard step/view for the restore request.

The basic architecture is a clone of BackupDownloadDetailsViewModel with regards to making the request. What differs is the handling of the onBackPressed and error events. I'm on the fence as to whether this should be an actual wizard step or just a different view in the Details step. I am going to leave it as it for this PR, as I can adjust how the errors are handled separately. 

**Merge Instructions**
- Make sure #13773 has been merged 
- Remove Not Ready For Merge Label 
- Merge as normal

**Cloned Classes**
RestoreWarningAdapter
RestoreWarningFragment
RestoreRequestState
RestoreWarningStateListItemBuilder
restore_warning_fragment.xml

**Notes:**
- Ignore styling - this is being handled separately
- Ignore unit test placement in packages - this is being handled separately
- RestoreActivity & RestoreViewModel (+unit tests) will continue to get updated as I move throughout the steps of the wizard

**To test:**
**Warning**: The Restore site button is hooked up, so if you press it your site will be backed up!
- Navigate to ActivityLog
- Tap the more menu
- Select the restore option
- Tap the Restore Site button on the Details View
- Verify 
- (1) The warning view is shown
- (2) Tapping on X or back button returns you to the details view
- (3) Tapping the restore site button will kick off a restore. You'll see a snackbar message, but the next page won't show because it hasn't been implemented

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/PostRestoreUseCase.kt
```diff
@@ -0,0 +1,48 @@
...
+    suspend fun postRestoreRequest(
+        rewindId: String,
+        site: SiteModel,
+        types: RewindRequestTypes
+    ): RestoreRequestState = withContext(ioDispatcher) {
+        if (!networkUtilsWrapper.isNetworkAvailable()) {
+            NetworkUnavailable
+        }
+
+        val result = activityLogStore.rewind(RewindPayload(site, rewindId, types))
```
ashiagr, line 33: ❓: After `NetworkUnavailable` is fired, control will go to the next line and `RemoteRequestFailure` will be fired and only this last state will be sent to the VM ignoring `NetworkUnavailable`. Can we re-confirm this behaviour? We might want to add unit tests for use case to check this.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/restore/RestoreWarningViewModelTest.kt
```diff
@@ -0,0 +1,126 @@
...
+    fun `given a network connection issue, then an error event is posted`() = test {
+        whenever(postRestoreUseCase.postRestoreRequest(anyOrNull(), anyOrNull(), anyOrNull()))
+                .thenReturn(postRestoreNetworkError)
+
+        val uiStates = initObservers().uiStates
+        val errorEvents = initObservers().errorEvents
+
+        viewModel.start(site, restoreState, parentViewModel)
+
+        ((uiStates.last().items).first { it is ActionButtonState } as ActionButtonState).onClick.invoke()
```
ashiagr: 💡: How about we move these repeating lines in every test case to function like `triggerRestore`?

ashiagr (commented):
Thanks @zwarm! Since most of the classes were cloned and reviewed originally, I found everything fine overall. I just left one comment for `PostRestoreUseCase`.

Also few suggestions (not related to the PR):

1. How about we show a warning on the "Restore" screen that "This will override and remove all content created after this point" (similar to web version) before we try restore. I know you just referred the mockups so cc'ing @osullivanchris. I think it's an important message to be displayed to the user.

restore warning

2. On unchecking all restore items on the "Restore" screen, we might want to disable "Restore site" action button. I'm not sure what will restore if all items are unchecked. The button gets disabled on the web, just fyi.

osullivanchris:
> How about we show a warning on the "Restore" screen that "This will override and remove all content created after this point" (similar to web version) before we try restore. I know you just referred the mockups so cc'ing @osullivanchris. I think it's an important message to be displayed to the user.

Hey thanks for the ping @ashiagr!

I think we are having pretty much the same discussion [over here](https://github.com/wordpress-mobile/WordPress-iOS/pull/15640#issuecomment-760593364)? I hadn't seen this message on web. Seems like a good one to include. Similar to the iOS discussion did you mean to replace the warning message I had in there, or have this in addition?

ashiagr:
> I think we are having pretty much the same discussion over here? I hadn't seen this message on web. Seems like a good one to include. Similar to the iOS discussion did you mean to replace the warning message I had in there, or have this in addition?

Yeah I meant the same warning. We can just reuse the one shown in the web version.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/PostRestoreUseCase.kt
zwarm: @ashiagr - this all goes back to "should warning be a step" or really a state in details because we show network error as a snackbar back on details view in backup. I left this as an issue to address outside of this PR - handling "back" with error in the wizard.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/restore/RestoreWarningViewModelTest.kt
zwarm: Love this idea - in commit 57d0630

zwarm:
> 1. How about we show a warning on the "Restore" screen that "This will override and remove all content created after this point" (similar to web version) before we try restore. I know you just referred the mockups so cc'ing @osullivanchris. I think it's an important message to be displayed to the user.

I'm with you on this and I agree that we should closely match the language on the web. 
 
> 2. On unchecking all restore items on the "Restore" screen, we might want to disable "Restore site" action button. I'm not sure what will restore if all items are unchecked. The button gets disabled on the web, just fyi.

I can definitely implement this. @osullivanchris - just say the word

zwarm:
@ashiagr Thanks for the review, I pushed the minor comments. Leaving the UI message changes as a separate task. I'll merge this PR when CI completes

osullivanchris:
> I'm with you on this and I agree that we should closely match the language on the web.

Agree but still not 100% sure what we are proposing. Are we changing text, or adding an additional line of text? Can you show me what we're talking about, or need me to mock up? Just not exactly clear on this.

> I can definitely implement this. @osullivanchris - just say the word

Yep making the restore action disabled when nothing is selected makes sense to me. Go for it.

ashiagr:
> Agree but still not 100% sure what we are proposing. Are we changing text, or adding an additional line of text? Can you show me what we're talking about, or need me to mock up? Just not exactly clear on this.

I was looking for the message on the screen with the checkboxes 🤦‍♀️ since it appeared on clicking checkboxes on the web, I think it's fine to use the existing warning message unless iOS wants any changes.

osullivanchris:
Discussed with @ashiagr on slack and realised the word 'All' is the only thing causing conflict for me now. Its the same as web. But it seems wrong to tell people that we are doing "All" whether they select some, or all, boxes. Hope that makes sense :) Will revert with same on the iOS issue

ashiagr (approved):

PR #13778, opened 2021-01-14
Issue/13328 restore progress

zwarm:
Parent #13328

This PR implements the "progress" wizard step/view for the restore request and is mostly a clone of Backup Download progress with a few temporary navigations until the last step in the wizard has been implemented. 

**Merge Instructions**
- Make sure #13773 & #13775 have been merged (sorry, but they need to build upon each other)
- Remove Not Ready For Merge Label
- Merge as normal

**Cloned Classes**
RestoreProgressAdapter
RestoreProgressFragment
RestoreProgressStateListItemBuilder

**Notes**:
- Ignore styling - this is being handled separately 
- Ignore unit test placement in packages - this is being handled separately 
- RestoreActivity & RestoreViewModel (+unit tests) will continue to get updated as I move throughout the steps of the wizard

**To test:**
**Warning**: The Restore site button is hooked up, so if you press it your site will be backed up!
- Navigate to ActivityLog
- Tap the more menu
- Select the restore option
- Tap the Restore Site button on the Details View
- Tap the Confirm restore site on the Warning View
Verify
- (1) The progress view is shown
- (2) If there is an error, the wizard will close (eventually it will go to the complete view)
- (3) When the restore finishes, the wizard will close (eventually it will go to the complete view)
- (4) Tapping the OK, notify me button, Back Button and X will close the wizard

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
```diff
@@ -154,6 +154,10 @@ class BackupDownloadViewModel @Inject constructor(
         _wizardFinishedObservable.value = Event(BackupDownloadCanceled)
     }
 
+    fun onBackupDownloadProgressExit() {
```
ParaskP7, line 157: Minor (🔍): Consider adding test for this and any other suchlike public functions.

PS: The same apply for the `RestoreViewModel`, its `onProgressExit()` function and its other public functions.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.wordpress.android.ui.jetpack.restore.RestoreRequestState.Complete
+import org.wordpress.android.ui.jetpack.restore.RestoreRequestState.Failure.NetworkUnavailable
+import org.wordpress.android.ui.jetpack.restore.RestoreRequestState.Failure.RemoteRequestFailure
+import org.wordpress.android.ui.jetpack.restore.RestoreRequestState.Progress
+import org.wordpress.android.util.NetworkUtilsWrapper
+import javax.inject.Inject
+
+const val DELAY_MILLIS = 10000L
+
+class GetRestoreStatusUseCase @Inject constructor(
```
ParaskP7, line 20: Warning (⚠️): Consider adding tests for this use case.

PS: Also the same applies to the `GetBackupDownloadStatusUseCase` use case, which doesn't have any as well.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
```diff
@@ -0,0 +1,64 @@
...
+    private val networkUtilsWrapper: NetworkUtilsWrapper,
+    private val activityLogStore: ActivityLogStore
+) {
+    suspend fun getRestoreStatus(site: SiteModel, restoreId: Long) = flow {
+        // start off with a delay until "queued" status is implemented in RewindStatusModel
+        delay(DELAY_MILLIS)
```
ParaskP7: 1) Question (❓): Why 10'' seconds?
2) Question (❓): There is another delay at the end of this function, do we need both of them?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.wordpress.android.util.NetworkUtilsWrapper
+import javax.inject.Inject
+
+const val DELAY_MILLIS = 10000L
+
+class GetRestoreStatusUseCase @Inject constructor(
+    private val networkUtilsWrapper: NetworkUtilsWrapper,
+    private val activityLogStore: ActivityLogStore
+) {
+    suspend fun getRestoreStatus(site: SiteModel, restoreId: Long) = flow {
```
ParaskP7, line 24: Praise (❤️): I am loving this `Flow` solution, kudos!

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
```diff
@@ -0,0 +1,64 @@
...
+        delay(DELAY_MILLIS)
+
+        while (true) {
+            if (!networkUtilsWrapper.isNetworkAvailable()) {
+                emit(NetworkUnavailable)
+                return@flow
+            }
+
+            val statusForSite = activityLogStore.getRewindStatusForSite(site)
+            val rewind = statusForSite?.rewind
```
ParaskP7: Minor (🔍): Consider removing `statusForSite` parameter since it is not used anywhere. What I mean is that these two lines can be merge: `val rewind = activityLogStore.getRewindStatusForSite(site)?.rewind`

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/viewholders/RestoreAdditionalInformationViewHolder.kt
```diff
@@ -0,0 +1,19 @@
...
+import org.wordpress.android.R
+import org.wordpress.android.ui.jetpack.common.JetpackListItemState
+import org.wordpress.android.ui.jetpack.common.viewholders.JetpackViewHolder
+import org.wordpress.android.ui.jetpack.restore.RestoreListItemState.AdditionalInformationState
+import org.wordpress.android.ui.utils.UiHelpers
+
+class RestoreAdditionalInformationViewHolder(
+    private val uiHelpers: UiHelpers,
+    parent: ViewGroup
+) : JetpackViewHolder(R.layout.backup_list_additional_information_item, parent) {
```
ParaskP7, line 14: Warning (⚠️): This `Restore` related view holder is using a `Backup` related layout file. This is most probably an oversight, right?

PS: The same applies for the `RestoreProgressViewHolder` class.

WordPress/src/main/res/layout/restore_progress_fragment.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
```
ParaskP7, line 6: Minor (🔍): This comment does not relate to this layout file but its corresponding `Backup` layout file, where there is an unused `xmlns:tools` namespace. Consider cleaning this up.

ParaskP7 (commented):
👋 @zwarm  !

I have reviewed and tested this PR as per the instruction, great job! Also, kudos for the `Flow` usage on the use case! 🌟 ❤️  

I have left some comments (🔍 + ⚠️) and few questions (❓), which I would like to discuss before approving this PR.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
zwarm: @ParaskP7 - While testing I discovered there is indeed another state called "queued" which is not accounted for in the existing services. It is not in FluxC either. Once that state is added to fluxc then I can properly handle it in the use case. Looking back at the existing service, they use a "now" parameter to either kick off the status check immediately OR delay for 10 seconds. Since this use case is always run after submittal of the restore request (again mimicking the existing), I just use the delay. I plan on updating FluxC when I update getting progress file names. 

What I will do is move the delay to after the network check - it will have the same effect. in commit 4bd7d42

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
zwarm: Noted and added the task. 👍

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
zwarm: Noted and added the task. 👍

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/viewholders/RestoreAdditionalInformationViewHolder.kt
zwarm: That is correct - "backup" is generic - I will be renaming to backup_restore later today :)

WordPress/src/main/res/layout/restore_progress_fragment.xml
zwarm: Noted - there will only be 1 fragment xml in the future - they are all the same.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
zwarm: In commit c5b9f6d

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
ParaskP7: Thanks for the detailed explanation @zwarm ! 🌟 

I am guessing the above is a response to `Question.1`. How about `Question.2`?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
ParaskP7: Thanks @zwarm ! Just to make it clear, this will be done on another PR, right?

zwarm:
@ParaskP7 - Let me know what else needs to be done before you give you approval? Thanks.

ParaskP7:
> @ParaskP7 - Let me know what else needs to be done before you give you approval? Thanks.

👋 @zwarm !

I have two pending questions for you, then I'll approve this PR. Can you check the comments I haven't yet resolved?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
zwarm: I removed the second delay in that same commit - moved it so that it can be reused.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
zwarm: Yes, use case unit tests will be in a separate PR - not yet started

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
ParaskP7: Oh sorry, missed that, you are right! 🙏

ParaskP7 (approved):

zwarm:
Many thanks @ParaskP7

PR #13780, opened 2021-01-14
Add initial Quick Start Task card item implementation to the improved My Site fragment

### LINKED ISSUE #13584: Create new Task card item

renanferrari:
Fixes #13584

This PR builds upon #13724 and #13725 and it basically introduces the check icon for done tasks, click handling and overall styling for the card.

Under the hood, the dummy classes were extracted and renamed and the layout is now using a `MaterialCardView` so we can take advantage of its built-in checkable states. This was customized quite a bit, first by tinting the check icon so it matches the color used by the parent card progress bar and also by introducing custom foreground and ripple colors, so we can fine tune what the card looks like when it's clicked and done, both in light and dark modes. We're also using a `ShapeableImageView` to control the corner radius of the illustrations, which matches the one for the card.

https://user-images.githubusercontent.com/830056/104662064-ba98c280-56a8-11eb-91d6-92ff065ac13d.mp4

### To test

Make sure all unit tests in `MySiteViewModelTest` pass. 

Additionally:

1. Go to the My Site screen.
1. Notice the sample Quick Start Task cards.
1. Make sure nothing looks out of place.
1. Click on a task card that is not done.
1. Notice the ripple effect.
1. Notice a snackbar with the task id.
1. Click on a task card that is done.
1. Notice how nothing happens.

---

PR submission checklist:


renanferrari:
I'm still going to add some illustrations to this PR, so I'm converting it to draft in the meantime.

renanferrari:
I have included the new illustrations for the "Customize your site" category and this PR is now ready for review:

https://user-images.githubusercontent.com/830056/104784709-a23bad00-5767-11eb-80f7-8606532336b6.mp4

Note: Since the illustrations are vector drawable files, there's now a size warning for this PR, but it's fine to ignore it.

planarvoid (approved):
Thanks for the changes! It's looking great. I have one doubt. When I click on an item, the background ripple end exactly at the same edge as the text and I don't think it looks great: 
 
PXL_20210118_114011101

renanferrari:
@planarvoid Thanks for the review! 

I agree the ripple effect looks a bit weird, given the current padding on the card. The way I see it, we could either add some padding to the card's title and description, or remove the ripple effect altogether, which was something @mbshakti was already considering (for other reasons), so I'd like to hear her input into this.

mbshakti:
It's looking soo good @renanferrari -- The only comment I had was the same one as @planarvoid, the ripple effect having the same edge as the text doesn't look great. I think it would be nice if the ripple effect had some kind of padding so the text is not at the edge, something like this:
Screen Shot 2021-01-18 at 5 01 02 PM

But I'm not sure how much work that's going to be, if it's gonna take a bit of time, I think maybe we can just remove the effect altogether since we were on the fence about it anyway.

renanferrari:
@mbshakti Thanks for sharing that design! I've just pushed some changes to achieve the desired effect:

https://user-images.githubusercontent.com/830056/105079056-a4ef1880-5a6d-11eb-8351-79f34e745181.mp4

Let me know what you think 🙂

mbshakti (approved):
Nice job!! Ripple effect is looking good :)

renanferrari:
Thanks @mbshakti. I'm going to merge this since it was previously approved.

PR #13784, opened 2021-01-15
Populate quick start cards with tasks

### LINKED ISSUE #13782: Populate quick start cards with tasks
This task is to setup the structure that populates the quick start cards in the new my site with un/completed quick start tasks.

planarvoid:
Fixes #13782

This PR creates a repository which provides a live endpoint of the quick start items shown on the my site screen. There is also a new builder which takes these items and transforms them into the UI items. I've added the `start` method call from the main activity which is now passed to the improved my site fragment and renamed the Dummy classes into normal names. Everything should be tested. 

To test:
- Open the app with the my site flag turned on
- Notice the cards are now visible and populated (you might have to create a new site for this)
- Notice the "Create a site" item is marked as "done" and the progress has shifted

Screenshot_1610715111

PR submission checklist:


renanferrari:
This is not a review just yet, as I've realized there's a bit of overlap between some of the changes here and the ones introduced by #13780 and we should probably decide how we want to deal with that first. Would you be ok rebasing this onto that branch? Otherwise, feel free to share your preferred approach in this case.

planarvoid:
I haven't realised that @renanferrari ! I've merged your PR into this one and fixed the conflicts - in the end there wasn't that much to solve. The problem I see here (and one I won't fix in this PR) is that we'll have updated new quick start items. We need to think about the current enums and updating the current state as well. This means there might be some unused drawables in this solution but we'll update that in another PR.

WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteItem.kt
```diff
@@ -52,13 +53,11 @@ sealed class MySiteItem(val type: Type) {
...
         @ColorRes val accentColor: Int,
+        val progress: Int,
         val onMoreClick: ListItemInteraction? = null
     ) : MySiteItem(QUICK_START_CARD) {
-        val doneTaskCards = taskCards.filter { it.done }
-        val progress = if (taskCards.isNotEmpty()) (doneTaskCards.size * 100f / taskCards.size).roundToInt() else 0
-
         data class QuickStartTaskCard(
-            val id: String,
+            val id: QuickStartTask,
```
renanferrari: Maybe we should rename this attribute from `id` to `task`?

WordPress/src/main/java/org/wordpress/android/ui/mysite/QuickStartRepository.kt
```diff
@@ -0,0 +1,79 @@
...
+        listOfNotNull(customizeCategory, growCategory)
+    }
+
+    fun startQuickStart() {
+        selectedSiteRepository.getSelectedSite()?.let { site ->
+            quickStartStore.setDoneTask(site.id.toLong(), CREATE_SITE, true)
+            refreshCustomizeTasks()
+            refreshGrowTasks()
+        }
+    }
```
renanferrari: One question: how are we planning to handle site selection changes? I'm asking because I'm wondering if we should adopt a more reactive approach here later on (in another PR) by listening to `SelectedSiteRepository::selectedSiteChange` instead? Let me know your thoughts on this.

renanferrari (approved):
Thanks for the changes, they're looking pretty good. I've tested the app and it works as expected. I left a couple of questions below but the PR is approved, so feel free to merge it once the lint errors are resolved and #13780 is merged to `develop` 👍

WordPress/src/main/java/org/wordpress/android/ui/mysite/QuickStartRepository.kt
planarvoid: this is a good point 👍 , I'll change the code to be more reactive

planarvoid:
thanks for the review @renanferrari . I've made some changes to make the QuickStartRepository reactive and actually shifted the task loading to a background thread. It would be great if you could give this PR another check 👍

.idea/codeStyles/Project.xml
```diff
@@ -32,7 +32,16 @@
     <JetCodeStyleSettings>
       <option name="PACKAGES_TO_USE_STAR_IMPORTS">
         <value>
-          <package name="kotlinx.android.synthetic" withSubpackages="true" static="false" />
+          <package name="kotlinx.android.synthetic" alias="false" withSubpackages="true" />
```
renanferrari: Do we really want to include these `Project.xml` changes here or were they included by mistake?

renanferrari (approved):
Thanks for the changes, the repository is looking even better now! I left one question below.

Again, feel free to merge this once conflicts have been resolved.

.idea/codeStyles/Project.xml
planarvoid: ah no, they are here by mistake. They keep on appearing in my changes and I always have to manually remove them :/

PR #13794, opened 2021-01-16
Fix stories `alt` tags

mzorz:
Fixes accessibility problems by setting the `alt` attribute on Media files properly, with the added text on Story frames images:
- we're setting the `alt` attribute on the Story block itself, for each media item composing a Story frame
- we're also setting the  `alt` attribute on the composed image that was uploaded to the Media section of the site, so in case it's used elsewhere the `alt` attribute will be set.

Related PRs:
- Stories library https://github.com/Automattic/stories-android/pull/631
- WordPressUtils https://github.com/wordpress-mobile/WordPress-Utils-Android/pull/60

_**Note: once the WordPressUtils PR is merged, let's update the `build.gradle` reference in this PR with the latest  `versionName` given.**_

To test:
1.  create a story with one or more slides
2. add some text to each slide
3. verify the uploaded html contains the `alt` attribute populated with the text you entered in step 2.
4. verify the media item in the media section contains the Alternative text set properly.

Example
The following html corresponds to a Story post with one slide, which has the word "hello" as an added view:
```
<!-- wp:jetpack/story {"mediaFiles":[{"alt":"hello","caption":"","id":"1767","link":"https://testmzorz3.files.wordpress.com/2021/01/wp_story1610815157231_0.jpg","mime":"image/jpeg","type":"image","url":"https://testmzorz3.files.wordpress.com/2021/01/wp_story1610815157231_0.jpg"}]} -->
<div class="wp-block-jetpack-story wp-story"></div>
<!-- /wp:jetpack/story -->
```

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
@@ -499,6 +499,10 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
                 }
             }
 
+            viewModel.onStorySaved()
+            // TODO add tracks
+            processStorySaving()
+
             val savedContentIntent = Intent()
             val blockId = intent.extras?.getString(ARG_STORY_BLOCK_ID)
             savedContentIntent.putExtra(ARG_STORY_BLOCK_ID, blockId)
@@ -509,6 +513,7 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
                 // if not, let's use the current Story
                 storyIndex = storyRepositoryWrapper.getCurrentStoryIndex()
             }
+
             // if we are editing this Story Block, then the id is assured to be a remote media file id, but
             // the frame no longer points to such media Id on the site given we are just about to save a
             // new flattened media. Hence, we need to set a new temporary Id we can use to identify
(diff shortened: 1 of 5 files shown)


WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
```diff
@@ -499,6 +499,10 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
                 }
             }
 
+            viewModel.onStorySaved()
+            // TODO add tracks
+            processStorySaving()
```
mzorz, line 504: I moved this call up here because we need the AddedViews to be read from PhotoEditor (in case they have just finished editing stuff on the last Story frame that was being shown on the Story composer) and added to the StoryRepository properly.
(see inside `processStorySaving()` --> `addCurrentViewsToFrameAtIndex()`)

WordPress/src/main/java/org/wordpress/android/ui/stories/media/StoryMediaSaveUploadBridge.kt
```diff
@@ -135,44 +141,45 @@ class StoryMediaSaveUploadBridge @Inject constructor(
...
-                        // find the MediaModel for a given Uri from composedFrameFile
-                        for (frame in story.frames) {
-                            // if the old URI in frame.composedFrameFile exists as a key in the passed map, then update that
-                            // value with the new (probably optimized) URL and also keep track of the new id.
-                            val oldUri = Uri.fromFile(frame.composedFrameFile)
-                            val mediaModel = oldUriToMediaFiles.get(oldUri)
-                            mediaModel?.let {
-                                val oldTemporaryId = frame.id ?: ""
-                                frame.id = it.id.toString()
+                    // find the MediaModel for a given Uri from composedFrameFile
```
mzorz, line 144: Side effect of this refactor: we were able to ditch this nested for loop running through all the Stories the StoryRepository currently holds, given the frames collection for the specific Story being saved is now being passed by parameter.

aforcier (approved):
Confirmed the alt text is extracted and set correctly in both the media metadata and in the block, and that it's properly reflected in the image alt text during playback.

Also updated the utils and stories lib references to reflected the merged PRs there.

:shipit:

PR #13795, opened 2021-01-18
Issue/13326 trigger fix threats

ashiagr:
Parent #13326

This PR triggers fix threats action on scan state screen and handles result from `FixThreatsUseCase`.
Child PR for `FixThreatsUseCase`: #13796

As fix threats API is asynchronous, polling for fix status is done in a separate PR: #13798 
[PR: #13798 also includes threat fixing flow (web vs mobile) video] 

**To test**

Prerequisite:

1. Make sure both Scan feature flag and MySiteImprovements flag are set to on in the App settings.
2. You have access to WP.com account with a site having scan capability and threats (e.g. https://pressable-jetpack-daily-scan.mystagingwebsite.com). Server credentials should be set on the site.
3. I've added a threat to the site. In case you run out of threats, re-install bad calendar plugin from [here](https://jetpackp2.files.wordpress.com/2020/05/calendar-1.zip) and re-run scan.

- Login to the app using above WP.com account and select the site on the My Site tab
- Click scan menu item on My Site
- Click "Fix `x` threats" button on scan state screen with threats
- Notice that
   - A confirmation dialog is displayed
   - Action buttons are disabled when fix threats action is triggered (when ok clicked on confirmation dialog)
   - (Action) Started message is displayed when action succeeds
   - Error message is displayed when action fails
   - Action buttons are enabled when action fails

(Tests included in `ScanViewModelTest`)

Notes:

1. Ignore all styling
2. Server credentials check is not handled
3. Message texts are kept close to web version but might not match exactly. Ensure that they'er relevant and we can update them during copy review.

Merge Instructions

- ~~Make sure child PR #13796 is merged to develop~~ ✅ 
- ~~Make sure #13768 is merged to develop~~ ✅ 
- ~~Remove the "Not Ready for Merge" label~~ ✅ 
- Merge this PR

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
```diff
@@ -76,6 +90,31 @@ class ScanFragment : Fragment(R.layout.scan_fragment) {
         ((recycler_view.adapter) as ScanAdapter).update(content.items)
     }
 
+    private fun SnackbarMessageHolder.showSnackbar() {
+        val snackbar = WPSnackbar.make(
+            scan_state_layout,
+            uiHelpers.getTextOfUiString(requireContext(), this.message),
```
ParaskP7: Minor (🔍): The `this` keyword can be removed from the `message`.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/ScanViewModelTest.kt
```diff
@@ -116,8 +127,109 @@ class ScanViewModelTest : BaseUnitTest() {
...
+    @Test
+    fun `when fix threats confirmation dialog action is triggered, then fix threats confirmation dialog is shown`() =
+        test {
+            val expectedDialogTitle = UiStringRes(R.string.threat_fix_all_warning_title)
+            val expectedDialogMessage = UiStringResWithParams(
+                R.string.threat_fix_all_warning_message,
+                listOf(UiStringText("1"))
+            )
+            val expectedPositiveButtonLabel = R.string.dialog_button_ok
+            val expectedNegativeButtonLabel = R.string.dialog_button_cancel
```
ParaskP7: 1) Minor (🔍): Consider moving the `expected` related variables to the `Assert` part of the test. IMHO the `expected` related variables are not to be confused with the `Arrange` part of any test as that part is only for what happens before the `Act`. Wdyt?
2) Minor (🔍): I would even consider doing the below to avoid some boilerplate:
```
                assertThat(title).isEqualTo(UiStringRes(R.string.threat_fix_all_warning_title))
                assertThat(message).isEqualTo(UiStringResWithParams(
                        R.string.threat_fix_all_warning_message,
                        listOf(UiStringText("1"))
                ))
                assertThat(positiveButtonLabel).isEqualTo(R.string.dialog_button_ok)
                assertThat(negativeButtonLabel).isEqualTo(R.string.dialog_button_cancel)
```

PS: If you agree on point.1 please apply it to the rest of the tests.

ParaskP7 (approved):
👋 @ashiagr !

I have reviewed and tested this PR as per the instruction, great job! 🌟

I have left a couple of minor (🔍) comments. I am going to approve this PR anyway, since none is blocking. Also, I am not going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

PS: The one thingy that I noticed while testing this PR is that I had to trigger `Scan now` twice for it to work. It seemed that the first time it gets triggered it returns immediately as green and only a second trigger would actually triggered a scan and show threats. Not sure if this was a timing matter, I just thought I should notice that here in case you experienced the same at some point.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
ashiagr: In 1fa403e

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/ScanViewModelTest.kt
ashiagr: In 21fb41a

ashiagr:
Thanks for the review @ParaskP7! I've included your suggestions. 

Ready for another look 🙇‍♀️.

ParaskP7:
> Ready for another look 🙇‍♀️.

🟢 from me @ashiagr ! Thank you so much. 🙏 

Feel free to merge when you are ready.

PR #13796, opened 2021-01-18
Issue/13326 fix threats usecase

ashiagr:
Parent: #13795

This PR adds fix threats use case.

To test:
That `FixThreatsUseCaseTest` tests run fine and cover action results.

PR submission checklist:

### DIFF
--- a/WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FixThreatsUseCaseTest.kt
+++ b/WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FixThreatsUseCaseTest.kt
@@ -1,0 +1,65 @@
+package org.wordpress.android.ui.jetpack.scan.usecases
+
+import com.nhaarman.mockitokotlin2.any
+import com.nhaarman.mockitokotlin2.whenever
+import kotlinx.coroutines.InternalCoroutinesApi
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Before
+import org.junit.Test
+import org.mockito.Mock
+import org.wordpress.android.BaseUnitTest
+import org.wordpress.android.TEST_DISPATCHER
+import org.wordpress.android.fluxc.action.ScanAction.FIX_THREATS
+import org.wordpress.android.fluxc.store.ScanStore
+import org.wordpress.android.fluxc.store.ScanStore.FixThreatsError
+import org.wordpress.android.fluxc.store.ScanStore.FixThreatsErrorType.INVALID_RESPONSE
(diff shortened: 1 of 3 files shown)


WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FixThreatsUseCaseTest.kt
```diff
@@ -0,0 +1,65 @@
...
+            OnFixThreatsStarted(FixThreatsError(INVALID_RESPONSE), FIX_THREATS)
+        )
+
+        val result = useCase.fixThreats(fakeSiteId, fakeThreatIds)
+
+        assertThat(result).isEqualTo(Failure.RemoteRequestFailure)
+    }
+
+    @Test
+    fun `when fix threats is triggered successfully, then Success is returned`() = test {
```
ParaskP7, line 58: Minor (🔍): Instead of the `successfully` wording with the `when` part, I would have added a `given` part and preserve consistency with the above test, like: `given valid response, when fix threats is triggered, then Success is returned`

PS: When using `NetworkUnavailable`, `RemoteRequestFailure` and `Success` words on the test sentences, I always advocate in favor of normal works, so that everyone can read the test sentence without worrying about the implementation details. For example, I would have done something along these lines:
- `..., then the call fails due to network unavailability`
- `..., then the call fails`
- `..., then the call succeeds`

ParaskP7 (approved):
👋 @ashiagr !

I have reviewed and tested this PR, it all works as expected, kudos on the tests, they look so beautiful! ❤️ 

PS: I have left a comment just as an FYI, feel free to merge this PR.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FixThreatsUseCaseTest.kt
ashiagr: > Instead of the successfully wording with the when part, I would have added a given part 

In 6c752ce

> When using NetworkUnavailable, RemoteRequestFailure and Success words on the test sentences, I always advocate in favor of normal works

I like it 👍, I'll start following this practice.

ashiagr:
Thanks for the review @ParaskP7! I liked the FYI part and included it in this PR itself. Need approval before I merge it :).

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FixThreatsUseCaseTest.kt
ParaskP7: Looks great @ashiagr ! ❤️ 

I am now going to merge this PR.

PR #13797, opened 2021-01-18
Issue/13326 fetch fix threats status usecase

ashiagr:
Parent #13798
This PR adds use case to fetch fix threats status.

To test:
That `FetchFixThreatsStatusUseCaseTest` tests run fine and cover action results.

Merge Instructions

1. Make sure child PR #13795 is merged to develop
2. Remove the "Not Ready for Merge label"
3. Merge this PR

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchFixThreatsStatusUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchFixThreatsStatusUseCase.kt
@@ -1,0 +1,81 @@
+package org.wordpress.android.ui.jetpack.scan.usecases
+
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.flowOn
+import org.wordpress.android.fluxc.model.scan.threat.FixThreatStatusModel
+import org.wordpress.android.fluxc.model.scan.threat.FixThreatStatusModel.FixStatus
+import org.wordpress.android.fluxc.store.ScanStore
+import org.wordpress.android.fluxc.store.ScanStore.FetchFixThreatsStatusPayload
+import org.wordpress.android.modules.IO_THREAD
+import org.wordpress.android.ui.jetpack.scan.usecases.FetchFixThreatsStatusUseCase.FetchFixThreatsState.Complete
+import org.wordpress.android.ui.jetpack.scan.usecases.FetchFixThreatsStatusUseCase.FetchFixThreatsState.Failure
+import org.wordpress.android.ui.jetpack.scan.usecases.FetchFixThreatsStatusUseCase.FetchFixThreatsState.InProgress
+import org.wordpress.android.util.AppLog
+import org.wordpress.android.util.AppLog.T
+import org.wordpress.android.util.NetworkUtilsWrapper
+import javax.inject.Inject
+import javax.inject.Named
+
+private const val FETCH_FIX_THREATS_STATUS_DELAY_MILLIS = 5000L
+
+class FetchFixThreatsStatusUseCase @Inject constructor(
+    private val networkUtilsWrapper: NetworkUtilsWrapper,
+    private val scanStore: ScanStore,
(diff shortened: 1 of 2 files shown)


WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchFixThreatsStatusUseCase.kt
```diff
@@ -0,0 +1,81 @@
...
+    private val scanStore: ScanStore,
+    @Named(IO_THREAD) private val ioDispatcher: CoroutineDispatcher
+) {
+    suspend fun fetchFixThreatsStatus(
+        remoteSiteId: Long,
+        fixableThreatIds: List<Long>,
+        delayInMs: Long = FETCH_FIX_THREATS_STATUS_DELAY_MILLIS
+    ): Flow<FetchFixThreatsState> = flow {
+        while (true) {
+            if (!networkUtilsWrapper.isNetworkAvailable()) {
```
malinajirka, line 35: 💡 We might want to consider moving the "internet connection" check above the "while" cycle or adding a retry behavior. I'm a bit worried the current implementation might show the error more often than necessary. Wdyt?
(even if you decide to make some changes, we can implement them later - up to you).

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchFixThreatsStatusUseCaseTest.kt
```diff
@@ -0,0 +1,123 @@
...
+
+        val useCaseResult = useCase.fetchFixThreatsStatus(fakeSiteId, listOf(fakeThreatId))
+            .toList(mutableListOf())
+            .last()
+
+        assertThat(useCaseResult).isEqualTo(Failure.RemoteRequestFailure)
+    }
+
+    @Test
+    fun `when threats fix status is fetched, then FixFailure is returned if result model contains error`() = test {
```
malinajirka: 🔍 Nitpick: We might want to move `if result model contains error` into the "given" part -> `given result model contains error, when threats fix status is fetched, then FixFailure is returned`. Wdyt?

malinajirka (approved):
Thanks @ashiagr! I've left two comments, but it looks great overall.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchFixThreatsStatusUseCase.kt
ashiagr: >  I'm a bit worried the current implementation might show the error more often than necessary.

There's a `return@flow` after `NetworkUnavailable` is emitted. I think it will take care of it. 

```
if (!networkUtilsWrapper.isNetworkAvailable()) {
    emit(Failure.NetworkUnavailable)
    return@flow
}
```
I can move it outside the while loop, if network is gone while fix status is in progress, `RemoteRequestFailure` will be returned instead of `NetworkUnavailable`. Is that fine?


> adding a retry behavior

I'd really like to add a retry behavior, adding it to the backlog.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchFixThreatsStatusUseCaseTest.kt
ashiagr: In c6c955a

ashiagr:
Thanks for the review @malinajirka! I've addressed your comments or added as action items. 

Ready for another look 🙇‍♀️.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchFixThreatsStatusUseCase.kt
malinajirka: > I can move it outside the while loop, if network is gone while fix status is in progress, RemoteRequestFailure will be returned instead of NetworkUnavailable. Is that fine?

I think that's fine. However, if you plan to implement the retry behavior anyway, I guess we can just wait until it's implemented and do not make any changes now.

malinajirka (approved):
LGTM, thanks for the changes!

PR #13800, opened 2021-01-18
Add prompt to update site title

### LINKED ISSUE #13786: Add functionality to "Update site title" quick start item
The "Update site title" item needs to point to the correct place when clicked and it should disappear once the site has been created.

planarvoid:
Fixes #13786

This PR adds the `activeTask` to the quick start repository. This requires the repository to now expose a model containing both the categories and the active task. The active task is passed to the site info block builder which uses it to show the quick start focus point. I've reused the previous component but for the focus point but it's now added directly to the layout which feels much cleaner than adding it dynamically. I couldn't figure out how to start the focus point animation immediately after setting it's visibility/adding it to the layout so instead there is a method which sets visibility and starts the animation. 

I'll add the rest of the functionality like the snackbar and the functionality to hide the focus point in the following PR.

To test:
- Turn the my site flag on and restart the app
- Click on the "Update site title" task in the first quick start card.
- Notice the quick start focus point appears over the Site title

update_quick_start_title_2

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mysite/SiteInfoBlockBuilder.kt
```diff
@@ -17,7 +19,8 @@ class SiteInfoBlockBuilder
         titleClick: (SiteModel) -> Unit,
         iconClick: (SiteModel) -> Unit,
         urlClick: (SiteModel) -> Unit,
-        switchSiteClick: (SiteModel) -> Unit
+        switchSiteClick: (SiteModel) -> Unit,
+        activeTask: QuickStartTask?
```
renanferrari: I think passing a boolean here instead of the active task would make the signature more consistent with the other parameters. What do you think?

renanferrari (approved):
Thanks for the changes, they're looking pretty good and the app works as expected. I left just one suggestion below, but the PR is approved, so feel free to merge it once the errors are resolved and #13784 is merged to `develop`.

WordPress/src/main/java/org/wordpress/android/ui/mysite/SiteInfoBlockBuilder.kt
planarvoid: I was thinking about this and I can't decide which option is better :-) but I'm ok with changing it to Boolean

renanferrari (approved):
Thanks again for the changes, LGTM!

PR #13802, opened 2021-01-18
Populate Scan History screen with data from /scan/history endpoint

malinajirka:
Parent issue #13326

Related PR in FluxC - https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1841

Merge instructions
1. Use FluxC tag
3. Merge this PR

This PR uses data provided by the /scan/history endpoint to populate Scan History screen (note: filtering by tab is not yet supported).
We decided to fetch data in the ScanHistoryViewModel (parent) and provide them to ScanHistoryListViewModel (children) through LiveData. Filtering is not yet supported, but I plan to do the actually filtering in the child view models.

To test:
2. Open Scan screen
3. Click on "History" button in the Toolbar
4. Notice the history screen is shown and threats get displayed after the request finished (it sometime takes 20+ seconds)

Note: 
- loading/error states are not implemented
- tab filtering doesn't work yet
- on item click is not implemented

PR submission checklist:

- [] I have considered adding unit tests where possible.


WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryViewModel.kt
```diff
@@ -1,17 +1,44 @@
...
     fun start(site: SiteModel) {
         if (isStarted) return
         isStarted = true
         this.site = site
+        fetchScanHistory()
+    }
+
+    private fun fetchScanHistory() {
+        launch {
+            val result = scanStore.fetchScanHistory(FetchScanHistoryPayload(site))
```
ashiagr, line 36: ❓: Are you planning to extract this logic to a use case? I think we're not using `ScanStore` directly in the VM.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryViewModel.kt
```diff
@@ -1,17 +1,44 @@
...
+            if (result.isError) {
+                // TODO malinjir handle error
+            } else {
+                _threats.value = scanStore.getScanHistoryForSite(site)
```
ashiagr, line 40: Same as above

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryListFragment.kt
```diff
@@ -30,15 +39,30 @@ class ScanHistoryListFragment : ViewPagerFragment(R.layout.scan_history_list_fra
...
-        viewModel.start(site)
     }
 
     private fun setupObservers() {
+        viewModel.uiState.observe(
+                viewLifecycleOwner,
+                { listItems ->
+                    refreshContentScreen(listItems)
+                }
+        )
```
ashiagr: Minor (🔍): It can fit in one line.

```suggestion
        viewModel.uiState.observe(viewLifecycleOwner, { refreshContentScreen(it) })
```

ashiagr (approved):
Thanks @malinajirka! So nice to start seeing threats history 🌟. 
I just left minor comments. Feel free to address them in subsequent PR. 

Approving this PR.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryViewModel.kt
malinajirka: I wasn't planning doing that, since I felt like the use case wouldn't do anything else but invoke this single method. However, I can wrap it in a use case when I add a network connection check.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryViewModel.kt
malinajirka: It seems the usecase won't contain anything else than this single line of code - do you think it's worth wrapping into a use case?

malinajirka: I feel like usecase are good in 3 scenarios
1. Converting event bus to a coroutine
2. Wrapping a code so it can be easily reused somewhere else
3. Wrapping a code which is complex and would deserve standalone test suite

I think this case doesn't fit any of the scenarios - tbh even the `fetchScanHistory` doesn't fit, since I doubt it'll ever get reused. Wdyt?

ashiagr: We can keep this one here, I agree with your reasoning. But I still feel we should have a usecase for `fetchScanHistory` to be consistent with the rest of the code. Either case, use your best judgement. I have no strong opinion.

malinajirka: > But I still feel we should have a usecase for fetchScanHistory to be consistent with the rest of the code.

👍 I'll introduce it in the upcoming PRs.

PR #13805, opened 2021-01-18
Revert "[Gutenberg] Audio block -  Consolidated & Normal Media Picker Integration"

develric:
Reverts wordpress-mobile/WordPress-Android#13612

The above mentioned #13612 PR was merged by mistake since we need to wait for other GB PRs to be deployed first. 

### To test
Smoke test the new and old media picker in adding media (images/videos) from the editor and from the media library.


jd-alexander (approved):
Thanks @develric I did the smoke test with both pickers for the File, Image, Video block all works as expected. LGTM 🚢

PR #13811, opened 2021-01-19
Issue/13455 make post as seen

### LINKED ISSUE #13455: Mark posts as seen/unseen in Reader
We should add the ability to mark posts in Reader as seen/unseen.

On Calypso, this is currently implemented through the post's dropdown menu. Current proposal is to use the same approach on mobile.

khaykov:
Fixes #13455

This PR adds an ability to mark posts as seen or unseen manually, or marking posts as seen by opening them, similarly to Calypso.

This functionality is available to all users logged in with wpcom account for posts that are a part of a feed (search or tag (topic) posts are excluded) . Logged-out users will not be able to access the toggle button in post card menu, and opening a post will not mark as seen, since there is no access token.

@develric I used your implementation of Comment Following as a reference, so the code should hopefully be familiar :)

[Image from Gyazo](https://gyazo.com/6b1a395981a31404437467bf21695361)

To test:
- Enable `UNREAD_POSTS_COUNT` feature.
- Open Reader and navigate to any stream (for test purposes navigate to the stream with a Site Filter)
- Open site filter and take a note of the unread post count (will not be visible if all posts are red).
- Open postcard menu and mark post as seen or unseen. Notice the Snackbar confirming the success of operation. Open the Reader in Calypso and confirm that seen status of the post was changed.
- Open site filter and confirm that the unread post counter was incremented or decremented.
- Turn off internet connection and try toggling seen status of the post. Notice Snackbar with error message.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
```diff
@@ -152,6 +157,20 @@ class ReaderPostListViewModel @Inject constructor(
...
 
+    fun onToggleSeenStatusClicked(post: ReaderPost, bookmarksList: Boolean) {
+        launch(bgDispatcher) {
+            readerPostCardActionsHandler.onAction(post, TOGGLE_SEEN_STATUS, bookmarksList)
+        }
+    }
+
+    fun onExternalPostOpened(post: ReaderPost) {
+        launch(bgDispatcher) {
+            seenStatusToggleUseCase.markPostAsSeenIfNecessary(post).flowOn(bgDispatcher).collect {
```
develric: Not sure, but since we are not basically collecting on this flow, wonder if we should use just a suspend function for `markPostAsSeenIfNecessary`? wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderSeenStatusToggleUseCase.kt
```diff
@@ -0,0 +1,139 @@
...
+                        readerPostTableWrapper.setPostSeenStatusInDb(post, true)
+                        readerBlogTableWrapper.decrementUnseenCount(post.blogId)
+                        analyticsUtilsWrapper.trackWithReaderPostDetails(
+                                AnalyticsTracker.Stat.READER_POST_MARKED_AS_SEEN,
+                                post,
+                                mutableMapOf(ACTION_SOURCE_PARAM_NAME to actionSource.toString())
+                        )
+                        PostSeenStateChanged(true, UiStringRes(string.reader_marked_post_as_seen))
+                    }
+                    is SeenStatusToggleCallResult.Failure -> {
```
develric: np: I think we should use the `SeenStatusToggleCallResult` or not both in Failure and Success case, wdyt?

WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderSeenStatusToggleUseCase.kt
```diff
@@ -0,0 +1,139 @@
...
+                        readerPostTableWrapper.setPostSeenStatusInDb(post, false)
+                        readerBlogTableWrapper.incrementUnseenCount(post.blogId)
+                        analyticsUtilsWrapper.trackWithReaderPostDetails(
+                                AnalyticsTracker.Stat.READER_POST_MARKED_AS_UNSEEN,
+                                post,
+                                mutableMapOf(ACTION_SOURCE_PARAM_NAME to actionSource.toString())
+                        )
+                        PostSeenStateChanged(false, UiStringRes(string.reader_marked_post_as_unseen))
+                    }
+                    is SeenStatusToggleCallResult.Failure -> {
```
develric: same np as above :)

develric (commented):
Hey @khaykov , this is looking great 👍 ! Good job 😄 ! I left a couple of comments inline with code, none of them are fundamental, so let me know wdyt 🙇 !

One thing I was thinking is I'm not fully sure we are all set on the web to release this and unread counts feature. Maybe we should merge it behind a feature flag?

Sharing two additional thoughts I had, but I think in case they make sense we can maybe manage them in a separate PR, let me know:

- Maybe there is already a plan for it, but I feel like it's important to mark visually a post that is seen/unseen like I see a post is grayed/not grayed in Calypso. Not sure there is a design already for the apps on this, but if it's not too much work maybe good to have it before to release the feature? Just my 2 cents, let's discuss about it 🙇 
- I think that with releasing this feature we should probably address the issue you noticed while reviewing a previous PR (opened an issue for it here #13821) and related to missing overflow menu in post cards when a filter is active. I found myself looking for the menu in a filtered stream 😄

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostListViewModel.kt
khaykov: Good idea 👍

WordPress/src/main/java/org/wordpress/android/ui/reader/usecases/ReaderSeenStatusToggleUseCase.kt
khaykov: Agree! I also renamed `PostSeenState.Failure` to `PostSeenState.Error` to differentiate from the requests `Failure`.

khaykov:
Thanks for the review, @develric !
To answer some of your questions:

> One thing I was thinking is I'm not fully sure we are all set on the web to release this and unread counts feature. Maybe we should merge it behind a feature flag?

We are releasing this one on mobile first to do a load/performance testing of the feature and make sure everything works as expected before releasing it for everyone on the web.

> Maybe there is already a plan for it, but I feel like it's important to mark visually a post that is seen/unseen like I see a post is grayed/not grayed in Calypso. Not sure there is a design already for the apps on this, but if it's not too much work maybe good to have it before to release the feature? Just my 2 cents, let's discuss about it 🙇

I confirmed the state of the design for that, and we will have aim to release it in the next version 👍 

> I think that with releasing this feature we should probably address the issue you noticed while reviewing a previous PR (opened an issue for it here #13821) and related to missing overflow menu in post cards when a filter is active. I found myself looking for the menu in a filtered stream 😄

Yep, that will be a good thing to address. I can take a look at it, unless you were planing to work on it :)

develric (approved):
Thanks for the changes and following up with clarifications @khaykov 🙇 , sounds all good to me 👍 !

Re the #13821 I didn't look into it yet as well 😊. , let's align on the other running priorities for the sprint on slack and get it covered 👍 

This said, good job and :shipit:

PR #13824, opened 2021-01-20
Update subheader and icon for threats in fixed and ignored status

malinajirka:
Parent issue #13326

This PR updates icon and subheader of Threat list items in Fixed and Ignored state.

I made two other changes
1. Changed default ThreatStatus for dummy object used for testing to CURRENT
2. Removed colors from related vector drawable resources and use tint to set the color programatically

Note: The UI is not polished, I made just necessary changes so the UI doesn't look completely broken (eg. vector dimensions).
Note2: Review by commits might be easier to follow.

To Test
1. Open Scan
2. Notice current threat has red background and "i"(notice) icon
3. Open History
4. Notice fixed threat has red background and shield icon with a checkmark
5. Notice ignored threat has grey background and "i"(notice) icon

To test:

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13827, opened 2021-01-20
Issue/13326 update tabs

malinajirka:
Parent issue #13326

This PR updates labels of tabs on Scan History screen and implements filtering by ThreatStatus type.

Merge instructions:
1. Make sure https://github.com/wordpress-mobile/WordPress-Android/pull/13824 is merged
2. Remove Not Ready for Merge label
3. Merge this PR

To test:
2. Click on Scan menu item on MySite screen
3. Click on History action in the toolbar
5. Notice labels of the tabs are "ALL, FIXED, IGNORED"
4. Wait until a request finishes (there is no progress and it sometime takes 30s)
6. Verify the filtering works

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryListFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryListFragment.kt
@@ -28,7 +28,7 @@ class ScanHistoryListFragment : ViewPagerFragment(R.layout.scan_history_list_fra
         super.onViewCreated(view, savedInstanceState)
         initDagger()
         initRecyclerView()
-        initViewModel(getSite(savedInstanceState))
+        initViewModel(getSite(savedInstanceState), getTabType())
     }
 
     private fun initDagger() {
@@ -44,12 +44,12 @@ class ScanHistoryListFragment : ViewPagerFragment(R.layout.scan_history_list_fra
         recycler_view.itemAnimator = null
     }
 
-    private fun initViewModel(site: SiteModel) {
+    private fun initViewModel(site: SiteModel, tabType: ScanHistoryTabType) {
         viewModel = ViewModelProvider(this, viewModelFactory).get(ScanHistoryListViewModel::class.java)
         parentViewModel = ViewModelProvider(parentFragment as ViewModelStoreOwner, viewModelFactory).get(
                 ScanHistoryViewModel::class.java
         )
-        viewModel.start(site, parentViewModel)
+        viewModel.start(tabType, site, parentViewModel)
         setupObservers()
     }
 
@@ -68,6 +68,9 @@ class ScanHistoryListFragment : ViewPagerFragment(R.layout.scan_history_list_fra
             savedInstanceState.getSerializable(WordPress.SITE) as SiteModel
         }
     }
+
+    private fun getTabType(): ScanHistoryTabType =
+            requireNotNull(arguments?.getParcelable(ARG_TAB_TYPE))
 
     override fun getScrollableViewForUniqueIdProvision(): View? = recycler_view
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryListFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryListFragment.kt
@@ -11,6 +11,7 @@ import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.ui.ViewPagerFragment
 import org.wordpress.android.ui.jetpack.scan.ScanListItemState
 import org.wordpress.android.ui.jetpack.scan.adapters.ScanAdapter
+import org.wordpress.android.ui.jetpack.scan.history.ScanHistoryViewModel.ScanHistoryTabType
 import org.wordpress.android.ui.utils.UiHelpers
 import org.wordpress.android.util.image.ImageManager
 import javax.inject.Inject
@@ -73,5 +74,16 @@ class ScanHistoryListFragment : ViewPagerFragment(R.layout.scan_history_list_fra
     override fun onSaveInstanceState(outState: Bundle) {
         outState.putSerializable(WordPress.SITE, viewModel.site)
         super.onSaveInstanceState(outState)
+    }
+
+    companion object {
+        private const val ARG_TAB_TYPE = "arg_tab_type"
+
+        fun newInstance(tabType: ScanHistoryTabType): ScanHistoryListFragment {
+            val newBundle = Bundle().apply {
+                putParcelable(ARG_TAB_TYPE, tabType)
+            }
+            return ScanHistoryListFragment().apply { arguments = newBundle }
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryListFragment.kt
```diff
@@ -68,10 +69,24 @@ class ScanHistoryListFragment : ViewPagerFragment(R.layout.scan_history_list_fra
         }
     }
 
+    private fun getTabType(): ScanHistoryTabType =
+            requireNotNull(arguments?.getParcelable(ARG_TAB_TYPE))
```
ParaskP7: Minor (🔍): This can become a one-liner.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryListViewModel.kt
```diff
@@ -21,18 +26,28 @@ class ScanHistoryListViewModel @Inject constructor(
...
-            }
-        }
+        uiState = parentViewModel
+                .threats
+                .map { threats ->
+                    threats
+                            .filter { mapTabTypeToThreatStatuses(tabType).contains(it.baseThreatModel.status) }
+                            .map {
+                                scanThreatItemBuilder.buildThreatItem(it, this::onItemClicked)
+                            }
```
ParaskP7: Minor (🔍): Consider formatting this a bit in order to save a couple of lines, like:
```
                .map { threats ->
                    threats.filter { mapTabTypeToThreatStatuses(tabType).contains(it.baseThreatModel.status) }
                            .map { scanThreatItemBuilder.buildThreatItem(it, this::onItemClicked) }
                }
```

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryViewModel.kt
```diff
@@ -1,14 +1,22 @@
 package org.wordpress.android.ui.jetpack.scan.history
```
ParaskP7, line 1: Minor (🔍): Consider running reformat on this class.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryListViewModelTest.kt
```diff
@@ -34,16 +43,54 @@ class ScanHistoryListViewModelTest {
...
 
         assertThat(viewModel.uiState.value).isNotEmpty
     }
+
+    @Test
+    fun `Only fixed threats are shown, when fixed tab is selected`() {
+        viewModel.start(ScanHistoryTabType.FIXED, site, scanHistoryViewModel)
+        viewModel.uiState.observeForever(mock())
+
+        verify(scanThreatItemBuilder, atLeast(1)).buildThreatItem(captor.capture(), anyOrNull())
```
ParaskP7: Idea (💡): What if instead of `atLeast(1)` you used a more explicit verification, like `times(3)`? This way the test becomes more solid as it will fail if you add/remove `FIXED` threats from the list or change the filtering logic.

PS: The same applies for the other two similar test below.

ParaskP7 (approved):
👋 @malinajirka  !

I have reviewed and tested this PR as per the instruction, good job!

I have left only minor (🔍) comments and one idea (💡 ). I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

PS: It seems that there is a CI issue with `Lint`.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryListViewModelTest.kt
malinajirka: Actually I had `times(3)` there and changed it to `atLeast(1)` to make it more solid/robust:D. If you add/remove threats into the list, the test will still work without changes and it'll still verify that only "FIXED" threats are returned (unless you remove all FIXED threats in which case it'll fail, but that's good because it won't be able to verify the behavior in that case). However, we'll need to update the test every time we touch the threat list if we use "times(3)".

Can you please elaborate how it becomes more solid with `times(3)`?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryListFragment.kt
malinajirka: Fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/13827/commits/c203e55da6ebf1c4921c38bd862250836a134a9b

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryListViewModel.kt
malinajirka: Fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/13827/commits/c203e55da6ebf1c4921c38bd862250836a134a9b

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryViewModel.kt
malinajirka: Fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/13827/commits/9af69b0ded024e77e024bc2781ee8c6d58220006

malinajirka:
Thanks for the review @ParaskP7! I've implemented the suggested changes. Let me know what you think ;).

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryListViewModelTest.kt
malinajirka: Ohh I see, you want the test to verify that all FIXED threats passed the filtering and non of them was filtered out even when it shouldn't. I'll update it with "times(x)", I guess it makes sense.

malinajirka: Updated in https://github.com/wordpress-mobile/WordPress-Android/pull/13827/commits/06744f918991f3ee1058587ae0cd522316938eb0

ParaskP7: Yes exactly your last comment Jirka! 💪 

Also, with `times(x)` if the main code filtering logic changes, say filtering by `FIXED` and something else, with the test list being untouched, then the test might still pass and produce a false positive. We would not know to update our tests since the main code got updated as well.

The solid/robust approach you were thinking is correct, but in this case, because it is applied to having the test more robust because of the test being updated (aka. adding/removing threats on the test list), then this doesn't provide much value. If the test constant is changing (in this case the test list), it is expected for the tests to change with it as well. Making that more solid/robust might create a problem with the production code.

I know, it is a thin line and I keep fighting with that line every single time! 😄

malinajirka: I totally agree, I tried to decrease the maintenance cost of the unit tests but I didn't realize it decreases the value of the test. Definitely a good call from you, thanks for the suggestion and elaborate reply!

ParaskP7 (approved):
Thanks for applying my suggestions @malinajirka ! 🌟 

I have checked the updated and it all looks good to me. Feel free to merge this PR when CI completes.

PR #13828, opened 2021-01-20
Feature/13786 add functionality to update site title

### LINKED ISSUE #13786: Add functionality to "Update site title" quick start item

planarvoid:
Fixes #13786

This PR adds logic to show a snackbar when QS item is clicked and it closes the current active task. You can use the commented out code to retest the functionality by making a task not done.

To test:
- Make sure the "My site" flag is turned on and the dynamic cards are visible
- Click on "Update site title" task
- Notice the snackbar saying "Tap *site title* to change your site title"
- Notice the focus point on Site title
- Click on the "Site title"
- Update the title/dismiss dialog
- Notice the focus point on Site title is gone
- Notice the card with "Update site title" task is gone 

remove_done_task

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mysite/QuickStartRepository.kt
```diff
@@ -74,6 +95,45 @@ class QuickStartRepository
...
+//            if (quickStartStore.hasDoneTask(site.id.toLong(), task)) {
+//                quickStartStore.setDoneTask(site.id.toLong(), task, false)
+//                refresh.value = false
+//                return
+//            }
+            if (task != activeTask.value) return
+            activeTask.value = null
+            if (quickStartStore.hasDoneTask(site.id.toLong(), task)) return
+            // If we want notice and reminders, we should call QuickStartUtils.completeTaskAndRemindNextOne here
+            quickStartStore.setDoneTask(site.id.toLong(), task, true)
```
renanferrari, line 131: Judging by the code in `QuickStartUtils`, I'm guessing we're missing a call to `AnalyticsTracker` here in order to track the completed task, what do you think?

renanferrari (commented):
Thanks for the changes! I've tested the app and everything seems to be working as expected. The behavior when switching sites still has problems, but I'm assuming we'll tackle this later on, right?

Code is looking pretty good, I left just one question below.

WordPress/src/main/java/org/wordpress/android/ui/mysite/QuickStartRepository.kt
planarvoid: I've added tracking in https://github.com/wordpress-mobile/WordPress-Android/pull/13828/commits/3d5e3975028c63fc455aeafce0e32354a69b8a2f

planarvoid:
Thanks for the check @renanferrari , I've added the tracking and merged develop

> I've tested the app and everything seems to be working as expected. The behavior when switching sites still has problems, but I'm assuming we'll tackle this later on, right?

What problems does it have? I've tried switching between 2 sites, one with the completed "Update site title" task and one without and it seems to update the UI correctly. Is there something I'm missing?

renanferrari:
@planarvoid Thanks for adding the tracking!

As for the problems, I found three but they are probably all related to each other:

1. After creating a new site, the Quick Start section appears, but it will remain there for every other site, which is different from the old behavior, as you can see in the comparison below. Also, notice how the other sites have the "Create your site" task uncompleted.

**Old behavior: Quick Start only appears for the site in which it was started**

https://user-images.githubusercontent.com/830056/105399128-faf4c500-5c01-11eb-907d-7eb3298a48f1.mp4

**New behavior: Quick Start appears for every site**

https://user-images.githubusercontent.com/830056/105399162-0516c380-5c02-11eb-9824-f55b1ff164de.mp4

2. If I start the "Set your site title" task after switching sites, the focus point won't appear:

https://user-images.githubusercontent.com/830056/105399208-165fd000-5c02-11eb-9229-9bf4c991a3ff.mp4

3. If I complete the "Set your site title" task and then switch to another site and try to start that task there, the focus point won't appear, but if I go back to the previous site, it will, even though it wasn't supposed to.

https://user-images.githubusercontent.com/830056/105399294-2d9ebd80-5c02-11eb-98e7-e5ec6259503e.mp4

planarvoid:
thanks for the check @renanferrari . I've found a bug - we were only updating the LiveData instance if quick start is in progress. This means that when switching to a site without QS we'd keep the QS items for the previous site. I've changed the logic (9c7fe36) to always refresh the data and remove the QS items when QS is not in progress. As far as I could check, it seems to work.

renanferrari (approved):
@planarvoid It seems that most of the issues I mentioned before have been solved, with the exception of issue 2. Let's keep track of it in #13867 and merge this PR since the rest looks pretty good.

LGTM :shipit:

PR #13834, opened 2021-01-20
Issue/13328 backup download single

zwarm:
Parent #13329 

This PR consolidates the BackupDownload feature into a single Activity/Fragment/VM. The wizard is still used to manage steps and movement. Non ui events are no longer have been moved out of the fragment - this is a deviation from the POC. Future enhancements will include moving the non-navigation events onClick handlers into the item state builders, although it may fall outside of the MVP.

Based on PR comments, I will move forward with the same consolidation for the restore feature.

**New**
- backup_download_activity.xml
- BackupDownloadFragment
- BackupDownloadStates.kt (consolidates the UI & toolbar states into a single file)

**Updated Classes**
- BackupDownloadActivity - stripped down to essentials, fragment reference is in layout
- BackupDownloadViewModelTest 
- BackupDownloadErrorTypes - Added int constructor
- BackupDownloadStep - Add new step for Error
- Dagger components were removed for old fragments/view models and updated for new fragment.

- 11 deletions

Test
- Run the backup download process from start to end. Note that the process works identically to the way it did before.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadFragment.kt
```diff
@@ -0,0 +1,186 @@
...
+        initAdapter()
+        initViewModel(savedInstanceState)
+    }
+
+    private fun initDagger() {
+        (requireActivity().application as WordPress).component().inject(this)
+    }
+
+    private fun initBackPressHandler() {
+        requireActivity().onBackPressedDispatcher.addCallback(
```
malinajirka, line 55: ❓ What do we need this dispatcher for? Can't we override onBackPressed?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadFragment.kt
```diff
@@ -0,0 +1,186 @@
...
+            }
+        })
+    }
+
+    private fun showView(state: BackupDownloadUiState) {
+        ((recycler_view.adapter) as JetpackBackupRestoreAdapter).update(state.items)
+    }
+
+    private fun updateToolbar(toolbarState: ToolbarState) {
+        val activity = requireActivity() as? AppCompatActivity
```
malinajirka, line 147: 💡 It seems we touch the toolbar from both activity and fragment. I'm wondering if it'd be easier to follow if we moved 
```
supportActionBar?.setHomeButtonEnabled(true)
supportActionBar?.setDisplayHomeAsUpEnabled(true)
```
from the activity into this method. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
```diff
@@ -131,6 +161,180 @@ class BackupDownloadViewModel @Inject constructor(
...
+                transitionToError(BackupDownloadErrorTypes.GenericFailure)
+            }
+        }
+    }
+
+    private fun buildProgress() {
+        _uiState.value = ProgressState(
+                toolbarState = ProgressToolbarState(),
+                items = stateListItemBuilder.buildProgressListStateItems(
+                        progress = 0,
```
malinajirka, line 195: ❓ Do we intentionally always set 0 to the progress?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
```diff
@@ -131,6 +161,180 @@ class BackupDownloadViewModel @Inject constructor(
...
+                backupDownloadState = backupDownloadState.copy(
+                        rewindId = result.rewindId,
+                        downloadId = result.downloadId,
+                        published = extractPublishedDate())
+                wizardManager.showNextStep()
+            }
+            is OtherRequestRunning -> {
+                _snackbarEvents.postValue(Event(OtherRequestRunningMsg))
+            }
+            else -> {
```
malinajirka, line 284: Feel free to ignore, it's not related to this PR.

🔍 Nitpick: If this is not an expected state, I'd consider throwing an exception (fail fast principle). Wdyt?

malinajirka (approved):
Thanks @zwarm, really great job ⭐ !

I've tested the flow and it works as expected. I've left a couple of comments, but none of them is vital. I'm approving this PR, so feel free to merge it (if you decide to make some changes, feel free to do them in a separate PR after this one is merged).

Note: One thing I noticed is that we replace up arrow with "X" in the toolbar after the first step of the wizard. I believe this is not a common practice on Android as "X" indicates it's a dialog -> it is a dialog even on the first step of the flow, so I'd consider showing "X" there as well. Wdyt?

zwarm:
> Note: One thing I noticed is that we replace up arrow with "X" in the toolbar after the first step of the wizard. I believe this is not a common practice on Android as "X" indicates it's a dialog -> it is a dialog even on the first step of the flow, so I'd consider showing "X" there as well. Wdyt?

This is a great point and fits better with the Android ecosystem. Any objections @osullivanchris to changing the first <- to an "x" to align with Android?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadFragment.kt
zwarm: @malinajirka - I feel like a simpleton here, but Fragment doesn't expose a override for onBackPressed, so I am using the dispatcher to intercept it from the activity, since I want to handle it all in the viewModel based on what step I am in. I didn't want to use an interface handler.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadFragment.kt
zwarm: Sounds good, I'll add it to my punch list for a future PR. 👍

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
zwarm: > If this is not an expected state, I'd consider throwing an exception (fail fast principle). Wdyt?

This is an edge case handler, we shouldn't fail if the app is left open on this view and someone clicks a different backup on the web. We show snackbar that there is already a request in progress. This is part of the unhappy state.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
zwarm: > Do we intentionally always set 0 to the progress?

Yes, but I can make this better by having an initial constant. Thanks for pointing this out. I'll note it and review again.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadFragment.kt
malinajirka: :headbang: Sorry, my bad - I shouldn't do reviews in morning. I mixed it up with "onActivityResult" 😞 .

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
malinajirka: ohh, I see the `when` doesn't not include all branches and some of them are reachable in edge cases. Thanks for the clarification!

osullivanchris:
> This is a great point and fits better with the Android ecosystem. Any objections @osullivanchris to changing the first <- to an "x" to align with Android?

Oh yeah, it looks like in the final designs post I have 'x' on all those screens as well! Let's do it 😄  I probably hadn't specifically addressed it before as we moved from normal deeper view to using full screen modals for these flows half way through the project.

PR #13836, opened 2021-01-21
Show postcard menu when site filter is active.

### LINKED ISSUE #13821: Missing overflow menu in reader post cards when filter is active.
While reviewing a not related PR [here](https://github.com/wordpress-mobile/WordPress-Android/pull/13766#pullrequestreview-568758343) it was noted the following.

### Expected behavior
When in the reader in a filterable tab like Following, if I select a filter, I would expect the posts shown to have the overflow menu like the posts card when no filter is selected like in the image below

<p align=center><image src="https://user-images.githubusercontent.com/47797566/105112447-579d9600-5ac3-11eb-8478-75c0d495a9ad.png" width=300/></p>

### Actual behavior
When in the reader in a filterable tab like Following, if I select a filter, the posts shown do not have the overflow menu

<p align=center><image src="https://user-images.githubusercontent.com/47797566/105112429-494f7a00-5ac3-11eb-9a3a-b54cad68f718.png" width=300/></p>

In the reader:
- go in a filterable tab like Following
- select a filter 
- notice the post cards presented do not have the overflow menu

##### Tested on [Pixel 3 Emu], Android [API 29], WPAndroid [current develop alpha-267]

NOTE: iOS seems ok already.

khaykov:
Fixes #13821 

When selecting a followed site from site filter, were were showing it's post as posts from Blog Preview, which does not have a menu button. This PR changes this, so the posts from followed sites will be treated in the same way as the unfiltered feed.

[Image from Gyazo](https://gyazo.com/f029fc62eb0b2435d97b046079870a75)

To test:
- Using any Reader's tab that has a filter, open it and select an individual site.
- Confirm that the menu button is visible and works as expected.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubFilterViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/subfilter/SubFilterViewModel.kt
@@ -254,7 +254,7 @@ class SubFilterViewModel @Inject constructor(
 
                 _readerModeInfo.value = (ReaderModeInfo(
                         null,
-                        ReaderPostListType.BLOG_PREVIEW,
+                        ReaderPostListType.TAG_FOLLOWED,
                         currentBlogId,
                         currentFeedId,
                         requestNewerPosts,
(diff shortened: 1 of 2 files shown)


develric:
Hey @khaykov , thanks for looking into this one 🙇 ! Had a first pass here and it seems I'm getting that, if I have multiple sites followed and try to filter on one of them, the list is still presenting posts from all sources like below

| Not filtered stream | Filtered stream with same posts as before | 
|---|---|
| image | image |


I'm afraid there are a few places in the logic where it relies on the list type being BLOG_PREVIEW so I'm a bit afraid on touching that part. Didn't look into code yet, but wonder if we should keep the `ReaderPostListType.BLOG_PREVIEW` and change the behaviour of showing/not showing the menu when the post card presented belongs to a feed in this top level reader fragment. Let's speculate about it later 👍

khaykov:
Closing in favor of #13897

PR #13837, opened 2021-01-21
Remove symlinks to gutenberg-mobile submodule (#13747)

### LINKED ISSUE #13747: Remove gutenberg-mobile symlinks in favor of reading assets from the AAR
To eventually remove the `gutenberg-mobile` submodule, we need to first get rid of a couple of symlinks that WordPress Android has for assets living in it.

I've already done it in my dirty experimental branch `react-native-from-bintray-experiment`. This is just a matter of doing it in isolation.

See also [the `gutenberg-mobile` counterpart](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2994) to add the assets to the AAR.

mokagio:
Fixes #13747 by removing symlinks to files in the `gutenberg-mobile` submodule in favor of loading them from the `gutenberg-react-native-bridge` library.

### To test

For `supported-blocks.json`:

- Add a breakpoint to `SupportedBlocks.kt` line 12, where we load the file
- Debug the app
- Load the screen to create a new site page

You'll see the breakpoint being hit and you should be able to step over successfully.

For `unsupported-blocks/`:

- Add a breakpoint to `WPGutenbergWebViewActivity.kt` line 143 and/or 149, where files from that folder are loaded
- From WP.com, create a post (or a draft) with a block that is not supported yet. I used `jetpack/business-hours`
- Launch the app and go to Posts > Drafts
- Load the post with unsupported block(s)
- Select the unsupported block > "Edit using web editor"

You'll see the breakpoint(s) being hit and you should be able to step over successfully. 

### PR submission checklist:

### DIFF
--- a/WordPress/src/main/assets/supported-blocks.json
+++ b/WordPress/src/main/assets/supported-blocks.json
@@ -1,1 +1,0 @@
-../../../../libs/gutenberg-mobile/src/block-support/supported-blocks.json
--- a/libs/editor/WordPressEditor/src/main/assets/unsupported-block-editor
+++ b/libs/editor/WordPressEditor/src/main/assets/unsupported-block-editor
@@ -1,1 +1,0 @@
-../../../../../gutenberg-mobile/resources/unsupported-block-editor/
--- a/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/WPGutenbergWebViewActivity.java
+++ b/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/WPGutenbergWebViewActivity.java
@@ -140,13 +140,13 @@ public class WPGutenbergWebViewActivity extends GutenbergWebViewActivity {
 
     @Override
     protected List<String> getOnGutenbergReadyExternalSources() {
-        String file = getFileContentFromAssets("unsupported-block-editor/remove-nux.js");
+        String file = getFileContentFromAssets("remove-nux.js");
         return Arrays.asList(file);
     }
 
     @Override protected List<String> getOnPageLoadExternalSources() {
         long userId = getIntent().getExtras().getLong(ARG_USER_ID, 0);
-        String file = getFileContentFromAssets("unsupported-block-editor/extra-localstorage-entries.js")
+        String file = getFileContentFromAssets("extra-localstorage-entries.js")
                 .replace("%@", Long.toString(userId));
         return Arrays.asList(file);
     }


WordPress/src/main/assets/supported-blocks.json
```diff
@@ -1 +0,0 @@
-../../../../libs/gutenberg-mobile/src/block-support/supported-blocks.json
```
mokagio, line 1: This diff looks like this because it was a symlink.

libs/editor/WordPressEditor/src/main/assets/unsupported-block-editor
```diff
@@ -1 +0,0 @@
-../../../../../gutenberg-mobile/resources/unsupported-block-editor/
```
mokagio, line 1: This diff looks like this because it was a symlink.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/WPGutenbergWebViewActivity.java
```diff
@@ -140,13 +140,13 @@ private String getAuthenticationPostData(String authenticationUrl,
...
     protected List<String> getOnGutenbergReadyExternalSources() {
-        String file = getFileContentFromAssets("unsupported-block-editor/remove-nux.js");
+        String file = getFileContentFromAssets("remove-nux.js");
         return Arrays.asList(file);
     }
 
     @Override protected List<String> getOnPageLoadExternalSources() {
         long userId = getIntent().getExtras().getLong(ARG_USER_ID, 0);
-        String file = getFileContentFromAssets("unsupported-block-editor/extra-localstorage-entries.js")
+        String file = getFileContentFromAssets("extra-localstorage-entries.js")
```
mokagio, line 149: These files live in the `resources/unsupported-block-editor` folder in `gutenberg-mobile`, but the AAR packaging (at least in the way it's configured [here](https://github.com/WordPress/gutenberg/pull/28216/files#diff-320e7dacc9e9a83ac8128d299319f82cf6dfa2609e0ebbc043d13de33ea92773R127)) puts them all on the same `assets/` folder.

Technically, I don't think that's a big deal. However, I can see how we lost a bit of context on those files.

If you think that we could benefit from having them in a dedicated `assets/` subfolder (and maybe can point me to a resource on how to make it happen 😅) I'll open a PR for it in Gutenberg and update this one accordingly.

oguzkocer (commented):
I am going to leave this one to @mchowning and @cameronvoell as they would have a lot more context than I do. The changeset is super small and it looks fine to me, but I'll leave the approval to them.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/WPGutenbergWebViewActivity.java
mchowning: The context is nice, but I agree that it's not a big deal. 👍

mchowning (approved):
Looks good, and works well. One step closer to not needing the submodule! 😄

PR #13842, opened 2021-01-21
Scan history UI state

malinajirka:
Parent issue #13326

This PR adds Error and Loading states to ScanHistoryFragment (contains viewpager and fullscreen error view) and ScanHistoryListFragment (contains RecyclerView).

Overview of the Scan History screen structure
1. ScanHistory (parent) - takes care of fetching the data from the server and showing an error when the request fails. It provides LiveData with a list of all threats.
2. ScanHistoryList (child, one per tab) - observes threats provided by ScanHistoryViewModel, filters them by ThreatStatus depending on the corresponding tab, and maps them to UiState. Takes care of showing list of skeleton items (loading), list of threats or empty screen when the list of threats is empty.

To test:
2. Open Scan
3. Click on History action in the toolbar (notice loading skeletons)
4. Turn on airplane mode before the request finishes
5. Notice a RequestFailed error is shown
7. Click on Retry
8. Notice NoConnection error is shown
9. Turn the airplane mode off
10. Click on retry
11. Notice threat items are shown when the request finishes (it might take 30s)
-----------------
Switch to a site which doesn't have any threats in history
1. Open Scan
2. Click on History action in the toolbar
3. Notice empty screen is shown when the request finishes (it might take 30s)

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryFragment.kt
```diff
@@ -8,19 +8,24 @@ import androidx.fragment.app.Fragment
...
+import org.wordpress.android.ui.jetpack.scan.history.ScanHistoryViewModel.UiState.ErrorUiState
 import org.wordpress.android.ui.utils.UiHelpers
 import org.wordpress.android.util.LocaleManagerWrapper
+import org.wordpress.android.util.image.ImageManager
 import javax.inject.Inject
 
 class ScanHistoryFragment : Fragment(R.layout.scan_history_fragment), ScrollableViewInitializedListener {
     @Inject lateinit var viewModelFactory: ViewModelProvider.Factory
     @Inject lateinit var uiHelpers: UiHelpers
+    @Inject lateinit var imageManager: ImageManager
```
zwarm: ❓ Is `imageManager` addition for a future PR? I don't see it being used yet.

zwarm (approved):
@malinajirka - all looks good, nicely done. Shimmer effect is neat. I left one question, but it's not a blocker and can be ignored. 👍

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/history/ScanHistoryFragment.kt
malinajirka: Good catch, I'll remove it. 

I wanted to use it, but it seems it doesn't work with vector drawables anymore. I'll look into it during a hack week.

malinajirka: Removed in https://github.com/wordpress-mobile/WordPress-Android/pull/13842/commits/dc1de4bd4594adb5135cdf64afe42d94334dc0d1

PR #13849, opened 2021-01-22
Issue/13328 restore single

zwarm:
Parent #13328 

This PR consolidates the Restore feature into a single Activity/Fragment/VM. The wizard is still used to manage steps and movement. Future enhancements will include moving the non-navigation events onClick handlers into the item state builders, although it may fall outside of the MVP.

**New**
- restore_activity.xml (this was a rename from jetpack_backup_restore_activity.xml) - now is a FragmentContainer
- RestoreFragment 
- RestoreStates.kt (consolidates the UI & toolbar states into a single file)

**Updated Classes**
- RestoreActivity - stripped down to essentials, fragment reference is in layout
- RestoreViewModelTest 
- RestoreErrorTypes - Added int constructor
- RestoreStep - Add new step for Error
- Dagger components - removed old fragments/view models and updated for new fragment.

- 13 deletions

Test
- Run the restore process from start to end. Note that the process works identically to the way it did before.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreFragment.kt
```diff
@@ -0,0 +1,182 @@
...
+        recycler_view.layoutManager = LinearLayoutManager(context, RecyclerView.VERTICAL, false)
+        initAdapter()
+    }
+
+    private fun initAdapter() {
+        recycler_view.adapter = JetpackBackupRestoreAdapter(imageManager, uiHelpers)
+    }
+
+    private fun initViewModel(savedInstanceState: Bundle?) {
+        viewModel = ViewModelProvider(requireActivity(), viewModelFactory)
```
malinajirka, line 79: ❓ I'm wondering if we need to attach the VM to the activity? Could we perhaps attach it to the fragment?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreFragment.kt
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    override fun onSaveInstanceState(outState: Bundle) {
+        viewModel.writeToBundle(outState)
+        super.onSaveInstanceState(outState)
+    }
+
+    companion object {
+        const val TAG = "RESTORE_FRAGMENT"
+        fun newInstance(bundle: Bundle?): RestoreFragment {
```
malinajirka, line 178: 💡 Idea: If we want to increase chances that whoever creates this fragments sets "KEY_RESTORE_ACTIVITY_ID_KEY" and "SITE", we might want to add them as parameters here instead of accepting generic "Bundle". Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
```diff
@@ -0,0 +1,106 @@
...
+        override val items: List<JetpackListItemState>,
+        override val toolbarState: ToolbarState
+    ) : RestoreUiState(ERROR)
+
+    sealed class ContentState(override val type: StateType) : RestoreUiState(type) {
+        data class DetailsState(
+            val activityLogModel: ActivityLogModel,
+            override val items: List<JetpackListItemState>,
+            override val toolbarState: ToolbarState,
+            override val type: StateType
```
malinajirka, line 29: ⚠️ We might want to consider removing this parameter from the constructor and keep it just in ContentState(..) constructor. It seems that we set the value twice on each instance. Wdyt?

Same applies to other states

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
```diff
@@ -0,0 +1,106 @@
...
+        val errorType: RestoreErrorTypes,
+        override val items: List<JetpackListItemState>,
+        override val toolbarState: ToolbarState
+    ) : RestoreUiState(ERROR)
+
+    sealed class ContentState(override val type: StateType) : RestoreUiState(type) {
+        data class DetailsState(
+            val activityLogModel: ActivityLogModel,
+            override val items: List<JetpackListItemState>,
+            override val toolbarState: ToolbarState,
```
malinajirka, line 28: ⚠️ We might want to turn this into a regular field (move it out of the constructor), since we always set this value to `DetailsToolbarState`. Wdyt?

Same applies to other states


```
data class DetailsState(
            val activityLogModel: ActivityLogModel,
            override val items: List<JetpackListItemState>,
        ) : ContentState(DETAILS) {
            override val toolbarState: ToolbarState = DetailsToolbarState
        }
```

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
```diff
@@ -0,0 +1,106 @@
...
+    COMPLETE(3),
+    ERROR(4)
+}
+
+sealed class ToolbarState {
+    abstract val title: Int
+    @DrawableRes val icon: Int = R.drawable.ic_close_24px
+
+    data class DetailsToolbarState(
+        @StringRes override val title: Int = R.string.restore_details_page_title
```
malinajirka, line 65: ⚠️ We might want to turn this into a regular field (move it out of the constructor), since we always set this value to `R.string.restore_details_page_title`. Wdyt?

```
object DetailsToolbarState : ToolbarState() {
        @StringRes override val title: Int = R.string.restore_details_page_title
    }
```

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreViewModel.kt
```diff
@@ -93,40 +146,216 @@ class RestoreViewModel @Inject constructor(
...
+        _uiState.value = ErrorState(
+                toolbarState = ErrorToolbarState(),
+                items = stateListItemBuilder.buildCompleteListStateErrorItems(
+                        onDoneClick = this@RestoreViewModel::onDoneClick
+                ),
+                errorType = errorType
+        )
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
```
malinajirka, line 239: 🔍 Nitpick: Do we really need the VisibleForTesting annotation? I haven't tested it, but it seems we could mock `wizardManager.navigatorLiveData` in tests? Wdyt?

malinajirka (approved):
Thanks @zwarm! It looks great overall! I've tested the app and it works as expected. I have left a few potential improvements - I missed them in the Backup PR. 

I'm approving this PR, but I think comments marked with ⚠️, ❓, 💡 should be considered. Having said that, feel free to create a separate PR and merge this PR. I don't want to block it, since it's quite big and keeping it up to date might get out of hands.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreFragment.kt
zwarm: Yep, an oversight on my part (copy & pasted from the activity). Next PR I'll take care of both features.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreFragment.kt
zwarm: I'm not even sure I should expose the fragment right now, it was used by the activity, which I nixed. If I decide to keep it, then yes, I agree I'll add them as params instead. I'll do the same for backup download.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
zwarm: I was actually considering removing the "individual" states and just using type as the differentiator - they are all the same, but I do want to know that the step & state match.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
zwarm: Yes, I noted the same thing last night, but it was late, so I will be revisiting today.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
zwarm: Good idea

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreViewModel.kt
zwarm: I'll go back and review this, thanks for the direction.

PR #13838, opened 2021-01-21
Issue/13812 fix site notifications on following tab

ashiagr:
Fixes #13812

This PR checks if the blog is an external feed [here](https://github.com/wordpress-mobile/WordPress-Android/blob/b7a62482dd80bc50c227d52be059c65b552999b3/WordPress/src/main/java/org/wordpress/android/datasets/ReaderBlogTableWrapper.kt#L15-L29), when site follow status is checked in the db. This resolves an occasional follow blog issue occurring when a blog is followed by url and feed id gets generated for it. 

It also fixes `SQLiteConstraintException: UNIQUE constraint failed` exception seen in the logs when site notifications are subscribed.

Details below:

To test:

1. Follow a simple WP.com site using the site url on Reader -> gear icon -> Manage Topics & Sites

2. Notice that the user's followed sites response does not contain `feed_id`. Details are updated in `tbl_blog_info` with `feed_id` = 0.
    Eg: [read-following-mine_meta=site%2Cfeed.txt](https://github.com/wordpress-mobile/WordPress-Android/files/5854302/read-following-mine_meta.site.2Cfeed.txt)

3. Return from "Manage Topics & Sites" to the following tab. Notice that followed site posts contains feed id and are inserted into `tbl_posts` with `feed_id` set.
    Eg: [read-following_number=20&meta=site,likes.txt](https://github.com/wordpress-mobile/WordPress-Android/files/5854296/read-following_number.20.meta.site.likes.txt)

4. Toggle follow site menu few times. Also toggle follow button on the post details screen.

5. Notice that site is followed/ unfollowed properly.
    
    With the external feed check in the `ReaderBlogTableWrapper` class, site is now retrieved from the `tbl_blog_info` using the `blog_id` instead of the `feed_id`. Earlier it wasn't found in the `tbl_blog_info` since `feed_id` was not set at step 2 and returned wrong followed status, sending incorrect follow request. 

https://user-images.githubusercontent.com/1405144/105350203-cd634800-5c10-11eb-80b7-2f9bf546e89a.mp4 

https://user-images.githubusercontent.com/1405144/105350250-deac5480-5c10-11eb-984f-2dbf6ae226f3.mp4

[...]

Issue #13812, opened 2021-01-19
Reader: Enable/Disable site notifications action doesn't work


malinajirka:
### Expected behavior
When I click on "Turn on/off site notifications" I'd expect the site notifications get enabled/disabled.

### Actual behavior
When I click on "Turn on/off site notifications" nothing happens.

```2021-01-19 09:31:24.885 8257-8509/org.wordpress.android.beta E/SQLiteDatabase: Error inserting SHOULD_EMAIL_COMMENTS=false BLOG_NAME=Writing Rhymes SUBSCRIPTION_ID=617136108 FEED_ID=83396915 URL=http://writingrhymes.wordpress.com SHOULD_NOTIFY_POSTS=false BLOG_ID=107850799 EMAIL_POSTS_FREQUENCY=instantly SHOULD_EMAIL_POSTS=true
    android.database.sqlite.SQLiteConstraintException: UNIQUE constraint failed: SubscriptionModel.FEED_ID, SubscriptionModel.URL (code 2067 SQLITE_CONSTRAINT_UNIQUE)
        at android.database.sqlite.SQLiteConnection.nativeExecuteForLastInsertedRowId(Native Method)
        at android.database.sqlite.SQLiteConnection.executeForLastInsertedRowId(SQLiteConnection.java:938)
        at android.database.sqlite.SQLiteSession.executeForLastInsertedRowId(SQLiteSession.java:790)
        at android.database.sqlite.SQLiteStatement.executeInsert(SQLiteStatement.java:88)
        at android.database.sqlite.SQLiteDatabase.insertWithOnConflict(SQLiteDatabase.java:1701)
        at android.database.sqlite.SQLiteDatabase.insert(SQLiteDatabase.java:1570)
        at com.yarolegovich.wellsql.InsertQuery.execute(InsertQuery.java:61)
        at org.wordpress.android.fluxc.persistence.AccountSqlUtils.updateSubscriptions(AccountSqlUtils.java:147)
        at org.wordpress.android.fluxc.store.AccountStore.updateSubscriptions(AccountStore.java:1307)
        at org.wordpress.android.fluxc.store.AccountStore.onAccountAction(AccountStore.java:912)
        at org.wordpress.android.fluxc.store.AccountStore.onAction(AccountStore.java:818)
        at java.lang.reflect.Method.invoke(Native Method)
        at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:507)
[...]

PR #13854, opened 2021-01-22
Issue/13812 site notifications discover

ashiagr:
This PR fixes sites notifications issue on the discover tab as part of #13812.
It prefetches the site if not available in db before the toggle follow and toggle notifications actions are performed.

To test:
- Fresh install the app and perform toggle (site, notification) actions on a discover tab post few times
- Verify that the post is appropriately followed/ unfollowed,  notifications are subscribed/ unsubscribed

Merge Instructions:

- Review and merge #13838
- Remove 'Not Ready for Merge' label
- Merge this PR

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
@@ -55,6 +55,9 @@ import org.wordpress.android.ui.reader.repository.usecases.PostLikeUseCase.PostL
 import org.wordpress.android.ui.reader.repository.usecases.UndoBlockBlogUseCase
 import org.wordpress.android.ui.reader.usecases.BookmarkPostState.PreLoadPostContent
 import org.wordpress.android.ui.reader.usecases.BookmarkPostState.Success
+import org.wordpress.android.ui.reader.usecases.ReaderFetchSiteUseCase
+import org.wordpress.android.ui.reader.usecases.ReaderFetchSiteUseCase.FetchSiteState
+import org.wordpress.android.ui.reader.usecases.ReaderGetSiteUseCase
 import org.wordpress.android.ui.reader.usecases.ReaderPostBookmarkUseCase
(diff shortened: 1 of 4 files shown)


WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
```diff
@@ -126,22 +131,71 @@ class ReaderPostCardActionsHandler @Inject constructor(
...
+            }
+            if (shouldPrefetchSite) {
+                handleActionAfterSiteFetch(post, type, isBookmarkList, fromPostDetails)
+            } else {
+                handleAction(post, type, fromPostDetails, isBookmarkList)
+            }
+        }
+    }
+
+    private suspend fun handleActionAfterSiteFetch(
```
malinajirka: 💡

1. It seems the usecase uses Flow, but it always returns a single result. Is the Flow really necessary and does it bring any benefits?

2. "Collect" is a suspend function, do we need to recursively call "handleAction"? We might want to consider taking advantage of the suspend and turn onAction into something like
```
withContext(bgDispatcher) {
            if (type == FOLLOW || type == SITE_NOTIFICATIONS) {
                val readerBlog = readerBlogTableWrapper.getReaderBlog(post.blogId, post.feedId)
                if (readerBlog == null) {
                    val result = fetchSite(post, type, isBookmarkList, fromPostDetails)
                    if (result != FetchSiteState.Success) return@withContext
                }
            } 
            handleAction(post, type, fromPostDetails, isBookmarkList)
        }
```
Wdyt?

malinajirka (commented):
Thanks @ashiagr! I haven't tested the app yet, I'll test it after you consider the proposed changes.

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderPostCardActionsHandler.kt
ashiagr: Thanks for quickly jumping on a call 🙇‍♀️, addressed in 4e4fde2.

malinajirka (approved):
All works as expected, thanks for the changes!

PR #13857, opened 2021-01-22
Fixing P2 and A8C tab shown traking in reader.

develric:
This PR aim to fix an issue in tracking the `READER_A8C_SHOWN` and `READER_P2_SHOWN` events. The goal of these 2 events should be to track when user swipe to or select the Automattic or Followed P2s tabs, whereas the way we were tracking was such that we were tracking when the tags related to those tabs were loaded, so checking the tracks in the logcat you could notice those 2 events were not fired every time you go to the related tabs but basically only at first load of the tabs.

### To test
- In logcat use the `Tracked: ` string as filter
- Swipe (or select) between all the available tabs 
- Check you get the track events related to `reader_a8c_shown` and `reader_p2_shown` every time you come to the relative tab
- Check you get `reader_following_shown`, `reader_discover_shown`, `reader_liked_shown`, `reader_saved_list_shown` when coming to the related tabs
- If you have custom lists in the reader you should get a single `reader_custom_tab_shown` until you change to a not generic custom tab

PR submission checklist:


khaykov (approved):
Thanks for spotting the issue and for the fix, @develric ! Works well 👍

Issue #13853, opened 2021-01-22
Snackbar appears behind FAB on the improved My Site fragment


renanferrari:
### Expected behavior

The FAB should slide up when a snackbar appears.

### Actual behavior

The FAB partially covers the snackbar:

<img width=300 src="https://user-images.githubusercontent.com/830056/105429578-72d9e400-5c30-11eb-94bc-5c28dfa6b532.png"/>

Issue #13844, opened 2021-01-22
Add functionality to "Enable post sharing" QS task


renanferrari:
When clicked, the QS task should scroll to the "Sharing" item in the list menu and show a focus point on the item. After clicking the item, it should redirect to the Sharing screen and show a focus point on the "Connections" section of that screen. The task should be completed after clicking any item on that screen.

Step 1 | Step 2
------------ | -------------
![sharing1](https://user-images.githubusercontent.com/830056/105430393-21caef80-5c32-11eb-9c66-76c495a49754.png) | ![sharing2](https://user-images.githubusercontent.com/830056/105430396-22fc1c80-5c32-11eb-9428-10957e3e7692.png)

PR #13858, opened 2021-01-22
Implement View Site QS task and scroll logic

### LINKED ISSUE #13840: Add functionality to "View site" updated QS task
When clicked the QS task should scroll to the "View site" item in the list menu and show a focus point on the item.

planarvoid:
Fixes #13840 

This PR adds the focus point and snackbar to the "View site" task. The task gets completed if you click on the "View site" button. I've also added scroll logic to the currently active task. Let me know if my approach makes sense, I'm not 100% it's good :-)

To test:
- Make sure the My Site improvements flag is turned on and the app is restarted
- Open a site with QS started
- Click on "View site" 
- Notice the app scrolls to the "View site" item and it's showing a focus point
- Go back to the Quick start card but make sure the Site info block is not visible
- Click on "Upload site icon" or "Update site title" task
- Notice the app scrolls up to the Site info block

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
```diff
@@ -178,6 +178,11 @@ class ImprovedMySiteFragment : Fragment(),
                 }
             }
         })
+        viewModel.onScrollTo.observe(viewLifecycleOwner, {
+            it?.applyIfNotHandled {
+                recycler_view.smoothScrollToPosition(this.second)
```
renanferrari: Maybe we should try to set an offset here by using `LinearLayoutManager::scrollToPositionWithOffset`. Again, I haven't tested this myself, so let me know what you think.

Edit: You've probably already addressed this, but I got the chance to test this as part of #13844 and found that the following solution seems to work pretty well. Let me know what you think.

```suggestion
                (recycler_view.layoutManager as LinearLayoutManager).scrollToPositionWithOffset(this.second, 0)
```

WordPress/src/main/res/values/styles.xml
```diff
@@ -518,9 +518,6 @@
         <item name="android:layout_height">@dimen/my_site_list_row_icon_size</item>
         <item name="android:layout_alignParentStart">true</item>
         <item name="android:layout_centerVertical">true</item>
-        <item name="android:layout_marginTop">@dimen/margin_large</item>
-        <item name="android:layout_marginBottom">@dimen/margin_large</item>
-        <item name="android:layout_marginEnd">@dimen/my_site_list_row_icon_margin_right</item>
```
renanferrari: I believe some of these styling changes have caused the My Site item layout to break when using the app with the feature flag turned off:

renanferrari (commented):
Hey @planarvoid, thanks for the changes. I've tested the solution and I agree with the approach we're using to scroll, although I did find some issues with it.

The most apparent one is that when we scroll to the "View site" item, it gets covered by the Snackbar:

https://user-images.githubusercontent.com/830056/105773955-7a0c3500-5f43-11eb-9354-b79e90610407.mp4

I know we're still going to change how the Snackbar is being positioned on #13853, but I've left a comment below with a suggestion that I believe we could try here, even though I haven't tested it myself.

I also left a comment below with another issue that affects the app when the flag is turned off.

WordPress/src/main/res/values/styles.xml
renanferrari: I actually just realized this also affects the bottom sheet's items layout:

planarvoid: Good catch! I've reverted the style changes and created a copies called "Updated.."  in 
5f9c859

planarvoid:
Thanks for the review and the solution to the overlapping snackbar! It works well 👍 . I think in this case the Snackbar probably isn't the most amazing solution. Some kind of tooltip message above the button would probably worked better.

renanferrari (approved):
Thanks for the changes! Everything seems to be working as expected now. 

Feel free to merge this once the test errors are resolved.

PR #13864, opened 2021-01-22
Issue/13329 backup download error retry

zwarm:
Parent #13329 

This PR updates `GetBackupDownloadStatusUseCase` with the following:
(1) Reorders execution so that the Fetch from the server is first, then a get on the local db
(2) Enable 3 retries when error is received from server fetch
(3) Lowers the delay to 5 seconds (from 10)
(4) Emits an `Empty` state when an null status model is returned (Note: The FluxC part has not been implemented yet)

This PR also includes a new test class: `GetBackupDownloadStatusUseCaseTest`

**To test:**
-Run the app and make sure the backup/download still functions as expected

PR submission checklist:


WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCaseTest.kt
```diff
@@ -0,0 +1,128 @@
...
+import org.wordpress.android.ui.jetpack.backup.download.BackupDownloadRequestState.Progress
+import java.util.Date
+
+@InternalCoroutinesApi
+class GetBackupDownloadStatusUseCaseTest : BaseUnitTest() {
+    private lateinit var useCase: GetBackupDownloadStatusUseCase
+    @Mock lateinit var networkUtilsWrapper: NetworkUtilsWrapper
+    @Mock lateinit var activityLogStore: ActivityLogStore
+    @Mock private lateinit var site: SiteModel
+    @Mock lateinit var statusModel: BackupDownloadStatusModel
```
ParaskP7: Suggestion (💡 ): Consider stubbing the models instead of mocking them. Usually it is better to use fakes for data related classes and only use mocks for logic related classes.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCaseTest.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    @Before
+    fun setup() {
+        useCase = GetBackupDownloadStatusUseCase(networkUtilsWrapper, activityLogStore)
+        whenever(networkUtilsWrapper.isNetworkAvailable()).thenReturn(true)
+        whenever(activityLogStore.getBackupDownloadStatusForSite(site)).thenReturn(statusModel)
```
ParaskP7: Minor (🔍): This setup is only really required on the `failure` scenario test. I would consider moving that there instead of having it in setup for all other tests.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCaseTest.kt
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Test
+    fun `given site, when status is fetched, then NetworkUnavailable is returned on no network`() = test {
```
ParaskP7: Minor (🔍): Consider changing the `given` sentence to `given no network` instead of having that at the end of the `then` sentence like `... on no network`. Having the given as `given site` it gives no real value to it. Let me know and I can further elaborate on it.

PS: The same applies to all tests below:
- `given failure`
- `given success`
- `given empty`
- `given progress is not null`

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCaseTest.kt
```diff
@@ -0,0 +1,128 @@
...
+        whenever(networkUtilsWrapper.isNetworkAvailable()).thenReturn(false)
+
+        val result = useCase.getBackupDownloadStatus(site, downloadId).toList(mutableListOf())
+
+        Assertions.assertThat(result).contains(Failure.NetworkUnavailable)
+    }
+
+    @Test
+    fun `given site, when status is fetched, then RemoteRequestFailure is returned on failure`() = test {
```
ParaskP7: Warning (⚠️): This test will always run with a delay of 5 seconds, given the fact that the `DELAY_MILLIS` is set to this value. This means that this test suite will deteriorate how long the tests are run in general.

Consider using the `MainCoroutineScopeRule` approach to fake the delay, by replacing `test { ... }` with `coroutineScope.runBlockingTest { ... }` and then add the `advanceTimeBy(DELAY_MILLIS)` right after the test `act` line. Let me know if you need help with that and I can explain in more detail.

PS: The same applies for the `progress` related test at the end which also delays to completion.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCaseTest.kt
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Test
+    fun `given progress is not null, when backup is running, then Progress is returned`() = testWithSuccessResponse {
+        whenever(activityLogStore.getBackupDownloadStatusForSite(site))
+                .thenReturn(inProgressModel)
+                .thenReturn(completeStateModel)
+        val result = useCase.getBackupDownloadStatus(site, downloadId).toList(mutableListOf())
+
+        Assertions.assertThat(result).contains(progressStatus)
```
ParaskP7: Question (❓): I am guessing this test is kind of incomplete, right? Would the `completeStatus` need to be asserted as well, since it is used in the `thenReturn(...)`? Like this: `assertThat(result).contains(progressStatus, completeStatus)`

ParaskP7 (commented):
👋 @zwarm  !

Great work on the testing part. I have left some comments for you to check there. The one comment I would really like to address is the warning one (⚠️). Let me know if you need help with that.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCaseTest.kt
zwarm: That's there to get us out of the progress loop, but it won't hurt. 👍

ParaskP7: Ah, I see now, thanks for sharing! 🌟

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCaseTest.kt
zwarm: Nice catch. 🙇

zwarm:
@ParaskP7 - all set and ready for next review. Thanks!

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCaseTest.kt
ParaskP7: 👋 @zwarm !

Thanks for making the change. 🙏

May I suggest going with using the `coroutineScope.runBlockingTest { ... }` instead of using `runBlockingTest { ... }` directly, as per my comment above?

I only insist on that for two reasons:
1) Using the `coroutineScope` to trigger the `runBlockingTest { ... }` helps making sure that all `runBlockingTest { ... }` related tests are running as part of the test scope and so they don't compete with each other, which makes the test more coroutine robust.
2) Searching for `runBlockingTest { ... }` on the whole `WordPress-Android` repo, I saw that the team has defined this `MainCoroutineScopeRule` and I am guessing this is what we should be using in such case. You can search for it and see how they set up a test with that (there are a couple of test with that pattern and `runBlockingTest` in general so far). @malinajirka maybe you can help us with the history here?

Apologies if I wasn't too clear on my initial comment. Let me know when you come back online and we will deal with that together so that to merge this with develop asap.

ParaskP7:
> @ParaskP7 - all set and ready for next review. Thanks!

👋 Annmarie! Thank for applying my suggestions. I have only one comment left unresolved. Let's deal with that and merge this with `develop` afterwards.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCaseTest.kt
zwarm: @ParaskP7 I think I got it. See commit a707713. 😉

ParaskP7: Awesome, thanks @zwarm , it looks good! 🌟

ParaskP7 (approved):

PR #13866, opened 2021-01-22
Try to get post with feed item id in reader post details

khaykov:
Readers post can belong to multiple tags or streams, we sometimes have a multiple, almost identical versions of it at the same time - they only difference is additional data, like tag and feed item id. When opening a reader post details we use post and blog id to retrieve the post from DB, and because there are multiple posts that match give blog and post ID, we only take the first post from the set.

This PR changes the SQL query, so the posts with `feed_item_id` will appear at the top of the set and will get picked.
Reader logic is pretty complicated, so to avoid any potential complications I limited use of the the sorting to post details.

To test:
- Enable `UNREAD_POSTS_COUNT` feature.
- Pick a post in followed, a8c or or p2 streams. Confirm that is has a Mark as Seen/Unseen in postcard menu.
- Open post details and confirm that the Mark as Seen/Unseen is available in post menu (in top right corner).

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/datasets/ReaderPostTable.java
+++ b/WordPress/src/main/java/org/wordpress/android/datasets/ReaderPostTable.java
@@ -355,6 +355,15 @@ public class ReaderPostTable {
                        excludeTextColumn);
     }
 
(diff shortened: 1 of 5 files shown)


khaykov:
Going to close this for now, since we are trying a different approach.

PR #13869, opened 2021-01-23
Mark reference strings as non translatable.

khaykov:
@designsimply reported an issue where title of Contact Support screen is broken for some non-english locales.

[Image from Gyazo](https://gyazo.com/4cc67e6a5e58b5656455f18c59581882)

I turned out that `contact_fragment_title` string referenced `@string/contact_support` but was not marked as `translatable="false"`, and GlotPress escaped the `@` character as a result (like [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/res/values-es/strings.xml#L278))

@loremattei If I understand the issue correctly, this PR should hopefully fix this, and the offending string should be removed from the translated resources.

To test:
- There are no immediate changes.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/res/values/strings.xml
+++ b/WordPress/src/main/res/values/strings.xml
@@ -1115,7 +1115,7 @@
     <!-- threat fix -->
     <string name="threat_fix_default_header">How will we fix it?</string>
     <string name="threat_fix_fixed_header">How did Jetpack fix it?</string>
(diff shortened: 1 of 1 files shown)


loremattei (approved):
Yes, this looks good to me. 
With `translatable=false` GlotPress won't import the string at all.

PR #13870, opened 2021-01-24
Adjust states, context and others

zwarm:
Parent #13329 

This PR make minor enhancements to the Backup Download code base.  

Included in this PR are the following:
- Use fragment context instead of activity for viewmodel providers
- Remove newInstance - it’s not being used, entry is only through activity
- Move toolbar handling for home button and display home from activity to fragment
- Use val for starting progress value
- Tidy up toolbarState by making title a regular field and moving icon to parent
- Tidy up uiState by moving toolbar to a regular field
- Update onBackPressed when conditional to include “error” 

**To test:**
- Run the backup download process from start to end. 
- Note that the process works identically to the way it did before.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadStates.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadStates.kt
@@ -20,7 +20,7 @@ abstract class BackupDownloadUiState(open val type: StateType) {
 
     data class ErrorState(
         val errorType: BackupDownloadErrorTypes,
-        override val items: List<JetpackListItemState>,
+        override val items: List<JetpackListItemState>
     ) : BackupDownloadUiState(ERROR) {
         override val toolbarState: ToolbarState = ErrorToolbarState
     }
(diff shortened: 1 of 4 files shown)


WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadStates.kt
```diff
@@ -51,27 +59,23 @@ enum class StateType(val id: Int) {
 
 sealed class ToolbarState {
     abstract val title: Int
-    abstract val icon: Int
+    @DrawableRes val icon: Int = R.drawable.ic_arrow_back
```
malinajirka: ❓ If I recall correctly we wanted to use "close" everywhere, not "back". Or have we decided to switch to "back" icon?

malinajirka (approved):
Thanks @zwarm for making these improvements! ;) I've left one comment but it looks great overall. Feel free to merge it even if you update the icon.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadStates.kt
zwarm: 🤦 Yes, I left the wrong one in...fantastic catch.

PR #13871, opened 2021-01-24
Adjust states, context and other minor changes

zwarm:
Parent #13329 

This PR make minor enhancements to the Restore flow.  

Included in this PR are the following:
- Use fragment context instead of activity for viewmodel providers
- Remove newInstance - it’s not being used, entry is only through activity
- Tidy up toolbarState by making title a regular field and moving icon to parent
- Tidy up uiState by moving toolbar to a regular field

**To test:**
- Run the restore process from start to end. 
- Note that the process works identically to the way it did before.

PR submission checklist:


ParaskP7 (approved):
👋 @zwarm !

I have reviewed and tested this PR as per the instruction, everything works as expected. Good job on refactoring the toolbar state and usages! 🌟

PR #13872, opened 2021-01-25
Issue/13326 fix threats status progress labels

ashiagr:
Parent #13326

This PR 
- Adds different progress labels (for showing "Fixing threat" and threat info) to the `jetpack_list_progress_item` layout and updates corresponding `ProgressState`.
- Hides action buttons when threats are fixing.

Before | After
-------|------
before | after

To test:

Prerequisite:

- You have access to WP.com account with a site having scan capability and threats (e.g. https://pressable-jetpack-daily-scan.mystagingwebsite.com). Server credentials should be set on the site.
- A threat is added to the site. In case you run out of threats, re-install bad calendar plugin from here and re-run scan.

1. Login to the app using above WP.com account and select the site on the My Site tab
2. Click scan menu item on My Site
3. Click fix threats button on the scan screen + ok in the confirmation dialog
4. Notice that

   - Fix threats state and info labels are shown  in the progress layout
   - Action buttons are not shown

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
```diff
@@ -120,33 +122,33 @@ class ScanViewModel @Inject constructor(
...
         launch {
             @StringRes var messageRes: Int? = null
-            var isFixing: Boolean
+            var fixingThreatIds: List<Long>
+            val scanStateModel = requireNotNull(scanStore.getScanStateForSite(site))
             fetchFixThreatsStatusUseCase.fetchFixThreatsStatus(
                 remoteSiteId = site.siteId,
                 fixableThreatIds = fixableThreatIds
             ).collect { status ->
+                fixingThreatIds = emptyList()
```
ParaskP7: Minor (🔍): Consider changing that to `var fixingThreatIds = emptyList<Long>()` and thus avoid having it being declared further up, thus saving a line. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
```diff
@@ -195,11 +197,20 @@ class ScanViewModel @Inject constructor(
         }
     }
 
-    private fun updateFixThreatsStatusProgressBar(isVisible: Boolean) {
+    private fun updateFixThreatsStatusProgressBar(
+        scanStateModel: ScanStateModel,
+        fixingThreatIds: List<Long> = emptyList()
```
ParaskP7: Question (❓): The `emptyList()` default value is not necessary here, isn't that right? Do you want this function to work without having the caller pass a `fixingThreadIds` in the future? If not, I would consider removing that to avoid setting this expectation.

ParaskP7 (approved):
👋 @ashiagr  !

I have reviewed and tested this PR as per the instruction, great job! 🌟 

I have left one minor (🔍) comment and question (❓). I am going to approve this PR anyway, since none is blocking. I am NOT going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

PS: What is the history behind the `ScanStateListItemsBuilder` class and its test? I am seeing all of them being commented out. I really think that if test are written for this class it would help enormously, even with the code review process. For example, having this `buildFixThreatsProgressInfoLabel(...)` function being tested with all the various results it can produce it could have been a great help. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
ashiagr: Progress state can change from `InProgress` to `Complete` or `Failure` states, `fixingThreatIds` should be set to empty in all these cases. Instead of putting it in every when case, I added it to the top of the block that executes on flow result.
With this context, let me know if you have a better idea about its placement 😅.

EDIT: 
Sorry I misunderstood 🤦‍♀️, done in d3602da

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
ashiagr: Makes sense, removed in 8951875, b46df8f.

ashiagr:
Thanks for the review Petros, I've addressed your comment and answered the question.

> What is the history behind the ScanStateListItemsBuilder class and its test? 

These tests were introduced when screen items were not added as list items, I have to tweak them for the updated structure. Adding it to the backlog.

Ready for another look 🙇‍♀️.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
ParaskP7: Thank you Ashita, sorry that I wasn't clear enough! 🤗

ParaskP7:
👋 @ashiagr !

Thanks for applying my suggestions, it all looks good to me!

🙏

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
ashiagr: Nope, I misunderstood. Thank you for the review 🌟.

PR #13876, opened 2021-01-25
Update Zendesk to 5.0.2

### LINKED ISSUE #13861: ZenDesk Push Notifications don't work
- Submit a support ticket
- Respond to it from within ZenDesk
- You should receive a push notification from ZenDesk saying that your message was responded to, but that's not working

I suspect the issue is related to the fact that the Push Token changes over time, and we may not be updating the token with the ZenDesk SDK properly.  Viewing web hooks coming back to WP.com from ZenDesk, there are nearly none that contain an Android push token.

zwarm:
Fixes #13861 

This PR updates the Zendesk support library from 5.0.1 to 5.0.2.

It fixes an issue where the push notification is not received on the first message sent to a user. According to the Zendesk [release log](https://developer.zendesk.com/embeddables/docs/android-support-sdk/release_notes) this is the only change in the update.

### DIFF
--- a/WordPress/build.gradle
+++ b/WordPress/build.gradle
@@ -340,7 +340,7 @@ dependencies {
         exclude group: 'org.wordpress', module: 'utils' 
     }
 
-    implementation (group: 'com.zendesk', name: 'support', version: '5.0.1') {
+    implementation (group: 'com.zendesk', name: 'support', version: '5.0.2') {
         exclude group: 'com.google.dagger'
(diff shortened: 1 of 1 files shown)


jkmassel (approved):
I tested this on-device and can confirm that it works.

I'll merge it into `release/16.6` – this will result in it being shipped as part of that release (for delivery in two weeks, thus giving us ~2 weeks to test), rather than into `develop` which would put it out for release in 4 weeks.

Thanks for handling this @zwarm!!

PR #13883, opened 2021-01-26
Fix strings in release/16.6

loremattei:
Translators noticed that some strings which reference other strings resources have been sent to GlotPress. 
I think the only issue with those is that they miss the `translatable=false` attribute, so this PR adds it. 

Adding @zwarm and @malinajirka as reviewers to make sure this is the expected behaviour.

After this PR is merged, we'll need to merge the release branch into `develop` to update GlotPress.

cc @JavonDavis.

PR submission checklist:


zwarm (approved):
@loremattei - Works as expected. Thanks for catching this.

PR #13885, opened 2021-01-26
Extract site content fragment

### LINKED ISSUE #13884: Create child my site content fragment which always has selected site
The current implementation is used for both cases when we do and don't have a selected site. This causes some issues after selecting a site when the previous data is visible for a short time. The cleaner solution would be to extract the logic depending on the selected site into a child fragment and recreate this fragment each time a site is selected. This helps with the logic that shows the "Scan" item in the list because it gets loaded and updated only after a site was selected.

planarvoid:
Fixes #13884

We've been discussing with @malinajirka that the need to react to site changes in the my site view model makes the code unnecessarily complicated. There is also a chance of wrong behaviour after you change the selected site because we always use and update the previous view model. In the end we've reached a conclusion that the best possible solution is to extract the items related to the selected site into a separate fragment with its own view model. This component is only shown when there is a selected site and the fragment is recreated when the site is updated. Let me know what you think of my changes! There should be no functionality change to the My Site logic but the result looks cleaner 👍 

To test:
- Turn on the My Site improvements flag and restart the app
- Check the My Site works as expected (I've smoke tested the functionality and it seems OK, the tests pass with some minor changes).

PR submission checklist:


renanferrari (changes requested):
Hey @planarvoid, thanks for the changes. I've tested this solution and found that apparently it has some performance issues. The app got noticeably slow after switching sites a few times, which eventually caused an ANR:



I believe this may have something to do with the fact that we are recreating fragments with each site switch, although I might be wrong.

Other than that, while I do agree the resulting solution is cleaner than what we currently have, I'd argue that reacting to site changes inside the My Site view model still makes sense to me conceptually speaking and that maybe we could simplify some of that logic somewhere other than at the fragment level. 

That said, it's possible I'm just missing some context on what kind of problems we're trying to avoid and/or what other alternatives we have considered so far, so maybe you could expand on that a little bit?

WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
```diff
@@ -353,110 +128,33 @@ class ImprovedMySiteFragment : Fragment(),
...
-        recycler_view.setVisible(true)
+    private fun loadData() {
+        content_container.setVisible(true)
         actionable_empty_view.setVisible(false)
-        (recycler_view.adapter as? MySiteAdapter)?.loadData(items)
+    }
+
+    private fun showFragment(fragment: Fragment) {
+        val fragmentTransaction: FragmentTransaction = childFragmentManager.beginTransaction()
+        childFragmentManager.findFragmentByTag(CONTENT_FRAGMENT_TAG)?.let { previousFragment ->
```
malinajirka, line 141: ❓ I'm wondering if we need to remove the fragmenty manually, doesn't `replace` take care of it?

WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
```diff
@@ -153,190 +86,32 @@ class ImprovedMySiteFragment : Fragment(),
...
-                inputDialog.setTargetFragment(this, 0)
-                inputDialog.show(parentFragmentManager, TextInputDialogFragment.TAG)
-            }
-        })
-        viewModel.onQuickStartMenuShown.observe(viewLifecycleOwner, {
-            it?.getContentIfNotHandled()?.let { id ->
-                ((parentFragmentManager.findFragmentByTag(id) as? QuickStartMenuFragment)
-                        ?: QuickStartMenuFragment.newInstance(id))
-                        .show(parentFragmentManager, id)
+        viewModel.onSiteChange.observe(viewLifecycleOwner, {
```
malinajirka: ⚠️ Tbh I haven't noticed any slowness when I tested the changes on my device + emulator. However, I'm wondering if they might be caused by the way LiveData behave -> we start re-creating the content fragment after onStart/onResume, because the LiveData won't receive any updates until then -> perhaps using a different approach which would start the process sooner would speed things up. Wdyt?

malinajirka:
> I believe this may have something to do with the fact that we are recreating fragments with each site switch, although I might be wrong.

Tbh I wouldn't expect that to cause any issues. When you open eg. Stats, Activity Log, .. - you are creating a new activity a new fragment(s) and it's still really fast, not mentioning that Activity creation is more expensive operation. My guess is that the issue is hidden somewhere else, but I might be wrong of course.

planarvoid:
Thanks for the check @renanferrari and @malinajirka , while trying to improve the performance I've actually realized this solution doesn't work because we need to observe not only the actual site switching but also updates to the site model and that would destroy the purpose of change.

planarvoid:
Thanks for the reviews @malinajirka @renanferrari . 
1) Performance
I think I've fixed the performance issues. The problem was that the `distinct` on the `onSiteChanged` that was setting up the fragment didn't work so we were basically recreating the fragment 3-4 times on each site change. This should now be fixed. Could you let me know if it works better now? 
2) Is this change necessary?
This is an unusually complex and difficult problem to solve. The issue we want to avoid is to ever show data for a site that's not selected. Previously we had the `selectedSiteChange` event which would update the list but it wouldn't update all the sources at the same time. This could be solved with a complex structure either with flow or with hacky live data to something like this: 
`flow(siteChange) -> if site changed, drop current state -> flatMap all the sources based on the selected site`
This is doable but it's not simple and it's not really future proof. 
Another approach would be something like this:
`partial state source emits a map of site ID to value -> view model handles this state and only uses value for the selected site`
This approach is not great either. 

The change in this PR splits the problem into two parts. One is site switch - this is simplified a lot - we switch the fragment and recreate the view model when a different site is selected. This assures that the data in the fragment is always up to date. 
The other one is when the selected site data changes. This is handled within the content fragment and the content view model only if the updated site ID is the same as of the selected site (that's permanently set within the content fragment).

I'm open to any better solution to this problem but as I mentioned, it's very complex.

renanferrari:
Closed in favor of #13909.

PR #13893, opened 2021-01-27
Issue/13326 threat detail

malinajirka:
Parent issue #13326

This PR updates ThreatDetail screen
- Displays Fixed header next to the icon for Fixed threats
- Displays Found header below the icon for Fixed threats
- Displays Found header next to the icon for non-fixed threats (current, ignored)
- Update icons background based on threat status (red for current, grey for ignored, green for fixed)

To test:
2. Open threat detail for Current, Fixed and Ignored threats and make sure it matches the behavior described above

Notes:
1. `threat_details_list_header.xml ` is using styles from Jetpack styles (eg. Jetpack.Header). I'm not sure if that's fine, wdyt?
2. Updating the tests was a bit tricky - I think the existing tests are trying to test too much within one method. We might want to consider splitting them into multiple smaller tests where each test verifies only a "single" behavior.  
 
PR submission checklist:


WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/details/ThreatDetailsListItemsBuilderTest.kt
```diff
@@ -17,26 +18,37 @@ import org.wordpress.android.fluxc.model.scan.threat.ThreatModel.Fixable
...
 import org.wordpress.android.ui.jetpack.common.JetpackListItemState.ActionButtonState
 import org.wordpress.android.ui.jetpack.common.JetpackListItemState.DescriptionState
 import org.wordpress.android.ui.jetpack.common.JetpackListItemState.HeaderState
-import org.wordpress.android.ui.jetpack.common.JetpackListItemState.IconState
 import org.wordpress.android.ui.jetpack.scan.TEST_FILE_PATH
 import org.wordpress.android.ui.jetpack.scan.ThreatTestData
 import org.wordpress.android.ui.jetpack.scan.builders.ThreatItemBuilder
 import org.wordpress.android.ui.jetpack.scan.details.ThreatDetailsListItemState.ThreatContextLinesItemState
 import org.wordpress.android.ui.jetpack.scan.details.ThreatDetailsListItemState.ThreatFileNameState
+import org.wordpress.android.ui.jetpack.scan.details.ThreatDetailsListItemState.ThreatDetailHeaderState
```
ashiagr: Minor 🔍: Import-ordering needed here.

ashiagr (approved):
Thanks @malinajirka! All looks good to me 🌟. I just left a minor comment.

Regarding your notes

> threat_details_list_header.xml is using styles from Jetpack styles (eg. Jetpack.Header). I'm not sure if that's fine, wdyt?

How about we override `Jetpack.Header` in `scan_style`:
`<style name="Scan.ThreatDetails.Header" parent="Jetpack.Header">`
and use this style on the threat details header?

>  I think the existing tests are trying to test too much within one method. We might want to consider splitting them into multiple smaller tests where each test verifies only a "single" behavior.

We can surely do this, let's discuss more on how granular they should be in our Android sync call.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/details/ThreatDetailsListItemsBuilderTest.kt
malinajirka: Fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/13893/commits/978361bccc6342e577570a48c3f1aa9977198637

malinajirka:
Thanks for the review @ashiagr! 


> How about we override Jetpack.Header in scan_style: <style name="Scan.ThreatDetails.Header" parent="Jetpack.Header">
and use this style on the threat details header?

Fixed in 4f152ae

ashiagr (approved):
LGTM, thanks!

PR #13897, opened 2021-01-27
Feature/enable reader post menu for all posts

khaykov:
This PR makes the reader post menu button visible for all the reader posts. Previously it was not visible for posts in Blog Preview and Search. We already to a post-specific check to populate menu, so posts that previously had no menu button will have the following menu:

[Image from Gyazo](https://gyazo.com/8c8fe99201fe793b51123c9162f2870b)

In addition:
- We are showing report button for all the posts.
- We are not showing block button for followed posts.

To test:
- Maker sure the "Block this site" button is only visible for posts from sites that you are following. 
- From Reader post tap on a blog name, and navigate to to blogs post list. Notice that the postcard menu button is visible, and contains the "Report this post" (and "Block this site" button if you follow the site). Navigate to post details and confirm that the menu  button is visible and contains same options.
- Repeat the previous step for Search post results.

PR submission checklist:


develric (approved):
Thanks for this PR @khaykov 🙇 ! Code looks good and the app worked as described for me. I checked the actions from the menu (share, Block/Report a site or post etc) in filtered lists and in blog_preview streams and worked well. Also smoke test didn't show unexpected behaviours. Good job 👍 . :shipit:

Issue #13228, opened 2020-10-26
ANR sometimes when adding from free photo library to gallery block.


develric:
Noticed during internal call for testing (comment here pbMoDN-1oy-p2#comment-2763). 

### Expected behavior
When adding multiple images to a gallery block from Free Photo Library (could possibly happen also with other sources), I would expect those images to be uploaded into the gallery block

### Actual behavior
When adding multiple images to a gallery block from Free Photo Library (could possibly happen also with other sources), it has been reported sometimes the app goes ANR.

### Steps to reproduce the behavior
There is a suspect it can happen when the post becomes too big, but note I was not able currently to reproduce it. (cc @AmandaRiu )

- In GB add a gallery block
- Add images from Free Photo Library
- App could go in ANR

<p align=center>
<img src="https://user-images.githubusercontent.com/47797566/97166916-4a079700-1786-11eb-8d8b-be7e9092c7d0.png" width=300/>
</p>

PR #13900, opened 2021-01-28
Issue/13328 progress labels

zwarm:
Parent #13228

This PR enhances the Restore progress view by:
1. Updating the labels using the API message value and current entry.  (Note that current entry is not yet returning data, but when it does it will be shown under the progress bar)
2. Uses a indeterminate progress bar until the progress value is > 0
Also this PR adds back the "R.string" in `RestoreViewModel` I don't know why AS keeps dropping it. 

**Notes**
- View copy will be updated in the separate PR

|on submit|starting|in progress
|---|---|---|
|on submit|starting|progress

**To test:**
- Run the Restore Process from start to finish
- Take note that Progress view
-- Shows an indeterminate progress bar until progress > 0
-- Shows a message on the right hand side of the screen - this message is straight from the API

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreViewModel.kt
@@ -341,7 +341,11 @@ class RestoreViewModel @Inject constructor(
                                         } else {
                                             null
                                         },
-                                    progressStateLabel = UiStringText("${restoreStatus.message}"),
+                                    progressStateLabel = if (restoreStatus.message != null) {
+                                        UiStringText("${restoreStatus.message}")
+                                    } else {
+                                        null
+                                    },
                                     isIndeterminate = (restoreStatus.progress ?: 0) <= 0
                             )
                         } else {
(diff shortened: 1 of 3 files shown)


WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreViewModel.kt
```diff
@@ -330,11 +331,18 @@ class RestoreViewModel @Inject constructor(
...
-                                    )
+                                            R.string.restore_progress_label,
+                                            listOf(UiStringText(restoreStatus.progress?.toString() ?: "0"))
+                                    ),
+                                    progressInfoLabel = if (restoreStatus.currentEntry != null) {
+                                            UiStringText("${restoreStatus.currentEntry}")
+                                        } else {
+                                            null
+                                        },
+                                    progressStateLabel = UiStringText("${restoreStatus.message}"),
```
malinajirka: ⚠️ It seems that the "message" field is also nullable. I think this logic might display "null" to the user, wdyt?

malinajirka (commented):
Thanks @zwarm! The changes LGTM. I've left just one comment.

Issues not related to this PR which we might want to add to our todo list:
1. When I click on "Restore site" it navigates me to Warning screen -> there is a noticable animation glitch -> the cancel button appears after a delay and the confirmation button changes it's vertical position a bit (https://i.gyazo.com/ce0763521d7729b09b45ef048dd22110.gif)
2. There is a noticable delay after I click on "Confirm restore site" on the warning screen, it takes +- 1-2 seconds before the apps navigates me to the progress screen.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreViewModel.kt
zwarm: Downstream it's `UiHelpers.setTextOrHide` that handles setting the view itself and nulls are properly handled there, so we are good. 👍  I updated to handle null message in commit eb6088e

malinajirka (approved):

PR #13901, opened 2021-01-28
Add seen/unseen support to all posts.

khaykov:
Depends on #13836

This PR expands the seen/unseen toggle support to all the posts made before 2020-07-13

What changed:
- Updated the main endpoint for toggling seen/unseen status. Now we are using post and blog ID's to toggle the status.
- We use the previous endpoint to toggle seen/unseen status of non-wpcom posts.
- Instead of using `feed_item_id` to decide either to show the menu button or not we are using absence of `is_seen` flag to determine this. Note: this field is still missing from some endpoints (like tags and search) so until it's there I will be using a date comparison to determine if the feature if available for a specific post.

Note:
- Seen/Unseen button will not be visible right after DB upgrade, since we need to update posts and populate `is_seen` and `is_seen_supported` flag into DB.
- I found a couple of posts where seen/unseen toggle is not working. It also doesn't work on calypso for same posts, so there is not much we can do on client side until this is fixed. 

To test:
- Enable `SEEN_UNSEEN_WITH_COUNTER` feature and refresh posts.
- Confirm that the Seen/Unseen functionality works with all the posts in Reader - Followed, Discover, Blog Preview, P2, Likes, Tags.
- Confirm that the Seen/Unseen functionality works with non-wpcom posts (eg. from cnn.com).
- Confirm that the posts created before 2020-07-13 do not have a Seen/Unseen button in the post menu. It's easy to get to old posts in discover.wordpress.com or with some less popular tags/topics. 

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/utils/PostSeenStatusApiCallsProvider.kt
```diff
@@ -66,4 +61,39 @@ class PostSeenStatusApiCallsProvider @Inject constructor(
...
+                if (askingToMarkAsSeen) {
+                    endpoint = if (post.isExternal) {
+                        requestJson.put("feed_id", post.feedId.toString())
+                        requestJson.put("feed_item_ids", JSONArray(arrayListOf(post.feedItemId)))
+                        "/seen-posts/seen/new"
+                    } else {
+                        requestJson.put("blog_id", post.blogId.toString())
+                        requestJson.put("post_ids", JSONArray(arrayListOf(post.postId)))
+                        "/seen-posts/seen/blog/new"
+                    }
```
develric, line 80: Like this usage of if as an expression function ❤️

develric (approved):
Thanks for this one @khaykov , code LGTM; while testing I also found a couple of posts for which the seen/unseen behaviour and status seems not consistent, but I confirm also in my case it was the same happening (for those specific posts) operating on calypso, so seems something going on with those posts and maybe to report to backend folks. This said for the rest the behaviour was as described 👍 . :shipit:

PR #13902, opened 2021-01-28
Jetpack Section: Restore - Replace Rewind Status Service with Restore Use Cases

ParaskP7:
Parent #13328

This PR replaces the `RewindStatusService` with the `PostRestoreUseCase` and `GetRestoreStatusUseCase`. This step is necessary in order to reuse those use cases, remove the complicated status services and in general reduce the complexity within the `ActivityLogViewModel`. Also, this step is a prerequisite and as much necessary for another feature that is about to be added to this screen, which is showing a progress row when a `Backup Download` has been triggered.

⚠️ When this PR is merged into `develop` is will affect the next release as this refactor is not guarded on a feature flag. Make sure to do a thorough code review. @zwarm @malinajirka and @ashiagr I have added all of you as reviewers to this PR as this is a complicated one and I would love to get as much comments as possible. At least, testing the PR and finding out bugs would be a major help.

ℹ️ Don't be scared by the amount of commits (77), the file changes are not that many (9). The commits are that many because of the many refactors I had to do in order to improve the state of the `ActivityLogViewModel` and its corresponding tests to be able to remove the status service and replace that with the use cases. If you do this code review commit-by-commit it will help you enormously.

Future TODOs:
* Timestamp is not being shown on restore progress row and its corresponding snackbars (started and completed) when triggered from the web. An item for that will be added on the parent ticket.
* `RewindStatusService` has not been completely removed due to the fact that the `ActivityLogDetailViewModel` also uses it just to show/hide the `RESTORE` button. An item for that will be added on the parent ticket.

To test:

#### Scenario.1 (App flow)
* Find a rewindable item from the list and: 
  * Either, click on the `RESTORE` icon and then on the dialog that appears, click on the `RESTORE SITE` button,
  * Or, click on the item itself to launch the `Activity Log Details` screen, click the `RESTORE` button and then on the dialog that appears, click on the `RESTORE SITE` button,
* You should automatically be navigated back to the `Activity Log` screen, on top of the list, and a restore will be triggered,
* A snackbar should immediately appear informing that the restore just started. Make sure the restore timestamp is available on the snackbar.
* Make sure the `RESTORE` icon, for every rewindable item, is not shown while restore progress is ongoing.
* Wait for restore to finish while having the `Activity Log` screen open both, within the app and the web (so you can see the actual progress),
* When restore finishes (looking on the web), a the restore (push) notification should appear.
* Then, the restore progress item should disappear (after at least 5'' seconds, due to the while loop check status delay).
* A pull-to-refresh should be automatically triggered and a new restore item might (or might not) appear (depending on how fast the restore progress item disappears and how fast this new item becomes available on the backend, via the `activityLogStore.fetchActivities(...)` API call). In case the new restore item does not appear, wait a bit (maybe 10'' seconds) and trigger a manual pull-to-refresh. Verify that the new restore item appears after some time.
* A snackbar should immediately appear informing that the restore just completed.  Make sure the restore timestamp is available on the snackbar.
* Make sure the `RESTORE` icon, for every rewindable item, is now shown again.

#### Scenario.2 (Web flow)
* Trigger restore from web,
* Wait for it to start,
* Either, launch `Activity Log` screen, or trigger a pull-to-refresh if you are already within that screen.
* The restore progress item should immediately appears. **CHANGE** (⚠️): The timestamp will not be available when triggered from the web, or it might become available based on some constraints (to be improved).
* Make sure the `RESTORE` icon, for every rewindable item, is not shown while restore progress is ongoing.
* Wait for restore to finish while having the `Activity Log` screen open both, within the app and the web (so you can see the actual progress),
* When restore finished (looking on the web), a the restore (push) notification should appear.
* Then, the restore progress item should disappear (after at least 5'' seconds, due to the while loop check status delay).
* A pull-to-refresh should be automatically triggered and a new restore item might (or might not) appear (depending on how fast the restore progress item disappears and how fast this new item becomes available on the backend, via the `activityLogStore.fetchActivities(...)` API call). In case the new restore item does not appear, wait a bit (maybe 10'' seconds) and trigger a manual pull-to-refresh. Verify that the new restore item appears after some time.
* A snackbar should immediately appear informing that the restore just completed. **CHANGE** (⚠️): The timestamp will not be available when triggered from the web, or it might become available based on some constraints (to be improved).
* Make sure the `RESTORE` icon, for every rewindable item, is now shown again.

#### Scenario.3 (Go wild flow)
* Trigger any of the above 2 scenarios until you reach the point where the restore progress item is shown.
* Trigger a pull-to-refresh manually and make sure the progress item is still shown.
* Go to the bottom of the list and trigger another page load, go to the top of the list and make sure the progress item is still shown.
* Go to the bottom of the list and trigger another page load, click on any item to navigate to the `Activity Log Detail` page, and then go back. Go to the top of the list and make sure the progress item is still shown.
* Go to the bottom of the list and trigger another page load, wait there until the restore finishes.
* Then, make sure you are being navigated automatically to the top of the list and the restore progress item have disappeared.

#### Scenario.4 (Restore flow)
* Since this PR updates the `GetRestoreStatusUseCase` it is recommended that the `Restore Flow` is tested as well.
* To do so, open app and go to `My Site` tab,
* Enable `RestoreFeatureConfig`, scroll down and click the `RESTART THE APP` button. 
* Find a rewindable item from the list, click on the ellipsis menu item and select `Restore to this point` to launch the `Restore screen,
* Wait for the restore to finish, click `Done` and in general verify everything works as expected.

PS: I **REALLY** suggest doing a commit-by-commit review as this will make it much easier for you to review the overall solution.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
@@ -105,13 +105,13 @@ sealed class RestoreRequestState {
         val progress: Int?,
         val message: String? = null,
         val currentEntry: String? = null,
-        val date: Date? = null
+        val published: Date? = null
     ) : RestoreRequestState()
 
     data class Complete(
         val rewindId: String,
         val restoreId: Long,
-        val date: Date? = null
+        val published: Date? = null
     ) : RestoreRequestState()
 
     sealed class Failure : RestoreRequestState() {
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
@@ -5,16 +5,17 @@ import androidx.annotation.StringRes
 import org.wordpress.android.R
 import org.wordpress.android.fluxc.model.activity.ActivityLogModel
 import org.wordpress.android.ui.jetpack.common.JetpackListItemState
-import org.wordpress.android.ui.jetpack.restore.StateType.DETAILS
-import org.wordpress.android.ui.jetpack.restore.StateType.WARNING
-import org.wordpress.android.ui.jetpack.restore.StateType.PROGRESS
 import org.wordpress.android.ui.jetpack.restore.StateType.COMPLETE
+import org.wordpress.android.ui.jetpack.restore.StateType.DETAILS
 import org.wordpress.android.ui.jetpack.restore.StateType.ERROR
+import org.wordpress.android.ui.jetpack.restore.StateType.PROGRESS
+import org.wordpress.android.ui.jetpack.restore.StateType.WARNING
 import org.wordpress.android.ui.jetpack.restore.ToolbarState.CompleteToolbarState
 import org.wordpress.android.ui.jetpack.restore.ToolbarState.DetailsToolbarState
 import org.wordpress.android.ui.jetpack.restore.ToolbarState.ErrorToolbarState
 import org.wordpress.android.ui.jetpack.restore.ToolbarState.ProgressToolbarState
 import org.wordpress.android.ui.jetpack.restore.ToolbarState.WarningToolbarState
+import java.util.Date
 
 abstract class RestoreUiState(open val type: StateType) {
     abstract val items: List<JetpackListItemState>
@@ -103,10 +104,15 @@ sealed class RestoreRequestState {
         val rewindId: String,
         val progress: Int?,
         val message: String? = null,
-        val currentEntry: String? = null
+        val currentEntry: String? = null,
+        val date: Date? = null
     ) : RestoreRequestState()
 
-    data class Complete(val rewindId: String, val restoreId: Long) : RestoreRequestState()
+    data class Complete(
+        val rewindId: String,
+        val restoreId: Long,
+        val date: Date? = null
+    ) : RestoreRequestState()
 
     sealed class Failure : RestoreRequestState() {
         object NetworkUnavailable : Failure()
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
@@ -30,6 +30,7 @@ class GetRestoreStatusUseCase @Inject constructor(
         restoreId: Long? = null
     ) = flow {
         if (restoreId == null) {
+            if (isNetworkUnavailable()) return@flow
             if (fetchActivitiesRewind(site)) return@flow
         }
         while (true) {
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
@@ -22,7 +22,10 @@ class GetRestoreStatusUseCase @Inject constructor(
     private val networkUtilsWrapper: NetworkUtilsWrapper,
     private val activityLogStore: ActivityLogStore
 ) {
-    suspend fun getRestoreStatus(site: SiteModel, restoreId: Long) = flow {
+    suspend fun getRestoreStatus(
+        site: SiteModel,
+        restoreId: Long
+    ) = flow {
         while (true) {
             if (!networkUtilsWrapper.isNetworkAvailable()) {
                 emit(NetworkUnavailable)
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
@@ -36,7 +36,7 @@ class GetRestoreStatusUseCase @Inject constructor(
     ) = flow {
         if (restoreId == null) {
             if (!isNetworkAvailable()) return@flow
-            if (!isFetchActivitiesRewindSuccessful(site)) return@flow
+            if (!fetchActivitiesRewind(site)) return@flow
         }
         while (true) {
             if (!isNetworkAvailable()) return@flow
@@ -58,7 +58,7 @@ class GetRestoreStatusUseCase @Inject constructor(
                 }
             }
 
-            if (!isFetchActivitiesRewindSuccessful(site)) return@flow
+            if (!fetchActivitiesRewind(site)) return@flow
 
             delay(DELAY_MILLIS)
         }
@@ -71,7 +71,7 @@ class GetRestoreStatusUseCase @Inject constructor(
         } else true
     }
 
-    private suspend fun FlowCollector<RestoreRequestState>.isFetchActivitiesRewindSuccessful(site: SiteModel): Boolean {
+    private suspend fun FlowCollector<RestoreRequestState>.fetchActivitiesRewind(site: SiteModel): Boolean {
         val result = activityLogStore.fetchActivitiesRewind(FetchRewindStatePayload(site))
         return if (result.isError) {
             emit(RemoteRequestFailure)
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
@@ -50,12 +50,8 @@ class GetRestoreStatusUseCase @Inject constructor(
                         emitFailure()
                         return@flow
                     }
-                    RUNNING -> {
-                        emitProgress(rewind)
-                    }
-                    QUEUED -> {
-                        emitProgress(rewind)
-                    }
+                    RUNNING -> emitProgress(rewind)
+                    QUEUED -> emitProgress(rewind)
                 }
             }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
@@ -43,11 +43,7 @@ class GetRestoreStatusUseCase @Inject constructor(
                     (restoreId == null || rewind.restoreId == restoreId)) {
                 when (rewind.status) {
                     FINISHED -> {
-                        if (rewind.rewindId != null) {
-                            emitComplete(rewind)
-                        } else {
-                            emitFailure()
-                        }
+                        emitFinished(rewind)
                         return@flow
                     }
                     FAILED -> {
@@ -77,6 +73,9 @@ class GetRestoreStatusUseCase @Inject constructor(
         }
         return false
     }
+
+    private suspend fun FlowCollector<RestoreRequestState>.emitFinished(rewind: Rewind) =
+            if (rewind.rewindId != null) emitComplete(rewind) else emitFailure()
 
     private suspend fun FlowCollector<RestoreRequestState>.emitComplete(rewind: Rewind) {
         val rewindId = rewind.rewindId as String
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
@@ -33,10 +33,7 @@ class GetRestoreStatusUseCase @Inject constructor(
             if (fetchActivitiesRewind(site)) return@flow
         }
         while (true) {
-            if (!networkUtilsWrapper.isNetworkAvailable()) {
-                emit(NetworkUnavailable)
-                return@flow
-            }
+            if (isNetworkUnavailable()) return@flow
 
             val rewind = activityLogStore.getRewindStatusForSite(site)?.rewind
             if (rewind != null &&
@@ -59,6 +56,14 @@ class GetRestoreStatusUseCase @Inject constructor(
 
             delay(DELAY_MILLIS)
         }
+    }
+
+    private suspend fun FlowCollector<RestoreRequestState>.isNetworkUnavailable(): Boolean {
+        if (!networkUtilsWrapper.isNetworkAvailable()) {
+            emit(NetworkUnavailable)
+            return true
+        }
+        return false
     }
 
     private suspend fun FlowCollector<RestoreRequestState>.fetchActivitiesRewind(site: SiteModel): Boolean {
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
@@ -35,11 +35,11 @@ class GetRestoreStatusUseCase @Inject constructor(
         restoreId: Long? = null
     ) = flow {
         if (restoreId == null) {
-            if (isNetworkUnavailable()) return@flow
+            if (!isNetworkAvailable()) return@flow
             if (fetchActivitiesRewind(site)) return@flow
         }
         while (true) {
-            if (isNetworkUnavailable()) return@flow
+            if (!isNetworkAvailable()) return@flow
 
             val rewind = activityLogStore.getRewindStatusForSite(site)?.rewind
             if (rewind != null &&
@@ -64,12 +64,12 @@ class GetRestoreStatusUseCase @Inject constructor(
         }
     }.flowOn(bgDispatcher)
 
-    private suspend fun FlowCollector<RestoreRequestState>.isNetworkUnavailable(): Boolean {
+    private suspend fun FlowCollector<RestoreRequestState>.isNetworkAvailable(): Boolean {
         if (!networkUtilsWrapper.isNetworkAvailable()) {
             emit(NetworkUnavailable)
-            return true
+            return false
         }
-        return false
+        return true
     }
 
     private suspend fun FlowCollector<RestoreRequestState>.fetchActivitiesRewind(site: SiteModel): Boolean {
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
@@ -43,7 +43,9 @@ class GetRestoreStatusUseCase @Inject constructor(
                 when (rewind.status) {
                     FINISHED -> {
                         if (rewind.rewindId != null) {
-                            emit(Complete(rewind.rewindId as String, rewind.restoreId))
+                            val rewindId = rewind.rewindId as String
+                            val date = activityLogStore.getActivityLogItemByRewindId(rewindId)?.published
+                            emit(Complete(rewind.rewindId as String, rewind.restoreId, date))
                         } else {
                             emit(RemoteRequestFailure)
                         }
@@ -54,10 +56,14 @@ class GetRestoreStatusUseCase @Inject constructor(
                         return@flow
                     }
                     RUNNING -> {
-                        emit(Progress(rewind.rewindId as String, rewind.progress, rewind.message, rewind.currentEntry))
+                        val rewindId = rewind.rewindId as String
+                        val date = activityLogStore.getActivityLogItemByRewindId(rewindId)?.published
+                        emit(Progress(rewindId, rewind.progress, rewind.message, rewind.currentEntry, date))
                     }
                     QUEUED -> {
-                        emit(Progress(rewind.rewindId as String, rewind.progress, rewind.message, rewind.currentEntry))
+                        val rewindId = rewind.rewindId as String
+                        val date = activityLogStore.getActivityLogItemByRewindId(rewindId)?.published
+                        emit(Progress(rewindId, rewind.progress, rewind.message, rewind.currentEntry, date))
                     }
                 }
             }
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
@@ -4,6 +4,7 @@ import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.FlowCollector
 import kotlinx.coroutines.flow.flow
 import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.model.activity.RewindStatusModel.Rewind
 import org.wordpress.android.fluxc.model.activity.RewindStatusModel.Rewind.Status.FAILED
 import org.wordpress.android.fluxc.model.activity.RewindStatusModel.Rewind.Status.FINISHED
 import org.wordpress.android.fluxc.model.activity.RewindStatusModel.Rewind.Status.QUEUED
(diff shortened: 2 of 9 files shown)


WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
```diff
@@ -103,10 +104,15 @@ sealed class RestoreRequestState {
         val rewindId: String,
         val progress: Int?,
         val message: String? = null,
-        val currentEntry: String? = null
+        val currentEntry: String? = null,
+        val date: Date? = null
```
malinajirka: 🔍 Nitpick: We might want to consider renaming this value. The name doesn't tell us anything about the field except of its type. Is it current date or date of the last update of the progress or date when the action started?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
```diff
@@ -103,10 +104,15 @@ sealed class RestoreRequestState {
...
     ) : RestoreRequestState()
 
-    data class Complete(val rewindId: String, val restoreId: Long) : RestoreRequestState()
+    data class Complete(
+        val rewindId: String,
+        val restoreId: Long,
+        val date: Date? = null
```
malinajirka: Same as above

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -177,24 +153,157 @@ class ActivityLogViewModel @Inject constructor(
...
-        activityLogStore.getRewindStatusForSite(site)
-
-        reloadEvents(done = true)
+        reloadEvents(true, currentRestoreEvent)
         requestEventsUpdate(false)
 
         showFiltersIfSupported()
     }
 
+    @VisibleForTesting
```
malinajirka, line 157: 🔍 Nitpick: In general we try to avoid using @VisibleForTesting as it's not necessary in most cases. The test would ideally test the state the class is providing through public methods/fields. Usually the need for using this annotation indicates that the code could/should be extracted into a separate class. Having said all that, feel free to keep it.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
```diff
@@ -1,65 +1,100 @@
...
-                emit(RemoteRequestFailure)
-                return@flow
-            }
+            if (fetchActivitiesRewind(site)) return@flow
 
             delay(DELAY_MILLIS)
         }
+    }.flowOn(bgDispatcher)
+
+    private suspend fun FlowCollector<RestoreRequestState>.isNetworkUnavailable(): Boolean {
```
malinajirka: ⚠️ We might want to consider renaming this method to "isNetworkAvailable". 
1. We use isNetworkAvailable throughout the codebase so the current name might lead to mistakes.
2. Both names isNetworkAvailable/isNetworkUnavailable are extremely similar so it's easy to miss the difference when scanning the code.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
```diff
@@ -1,65 +1,100 @@
...
+        if (!networkUtilsWrapper.isNetworkAvailable()) {
```
malinajirka: 🔍 Nitpick: I like to have ideally a single return statement within a method when possible. We might want consider putting the return in front of the `if`. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
```diff
@@ -1,65 +1,100 @@
...
+            return true
+        }
+        return false
+    }
+
+    private suspend fun FlowCollector<RestoreRequestState>.fetchActivitiesRewind(site: SiteModel): Boolean {
+        val result = activityLogStore.fetchActivitiesRewind(FetchRewindStatePayload(site))
+        if (result.isError) {
+            emit(RemoteRequestFailure)
+            return true
```
malinajirka: ❓ It feels a bit unintuitive to me that the method returns "true" on failure and "false" on success. Could you please elaborate why you made this decision?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/PostRestoreUseCase.kt
```diff
@@ -28,7 +28,7 @@ class PostRestoreUseCase @Inject constructor(
     suspend fun postRestoreRequest(
         rewindId: String,
         site: SiteModel,
-        types: RewindRequestTypes
+        types: RewindRequestTypes? = null
```
malinajirka, line 31: ❓ Could we perhaps pass RewindRequestTypes(true,true,...) instead of introducing a nullable type here? I realize it'll slightly change the current behavior (the request will contain the types instead of an empty request) but I think it's worth it. Wdyt?

malinajirka (changes requested):
Really great job @ParaskP7 🥇 ! The logic is incomparably easier to follow now and the concerns are nicely separated ❤️ . I have left a few comments, but nothing vital. I tested the behavior with feature flags on/off and it worked pretty well - see possible issues below.

Note: This PR is quite huge and I found it difficult to review. You did a great job documenting your every step, but review requires focus and being focused while reviewing 70+ commits quite tricky (even if each commit would be extremely simply, which some of them aren't). Tbh I gave up the review-by-commit and I reviewed just the changes - which turned out to be ok, since most of the changes are in tests which don't require such a focused review.
Please always consider splitting PRs into smaller chunks - each of the PRs can be merged into a working branch and you can request that only the final PR needs to be tested. Thanks!


Possible issues
1. ⚠️ When I turn off all feature flags -> open Activity Log -> click on Restore -> it takes +- 5s before the item at the top with progress appears. Before this PR, the item would appear right away.
2. When I enable "restore" feature flag -> open Activity Log -> Click on "Restore" -> Click on "Restore site" -> Click on "Confirm restore site" -> click on "Ok notify me" -> the progress item doesn't appear, unless I pull down to refresh. (I think this behavior was the same even before this PR)

ParaskP7:
> Really great job @ParaskP7 🥇 ! The logic is incomparably easier to follow now and the concerns are nicely separated ❤️ . I have left a few comments, but nothing vital. I tested the behavior with feature flags on/off and it worked pretty well - see possible issues below.

Thanks for the review @malinajirka ! 🌟 

> Note: This PR is quite huge and I found it difficult to review. You did a great job documenting your every step, but review requires focus and being focused while reviewing 70+ commits quite tricky (even if each commit would be extremely simply, which some of them aren't). Tbh I gave up the review-by-commit and I reviewed just the changes - which turned out to be ok, since most of the changes are in tests which don't require such a focused review.

Yes, I could probably merge some of the commits together, like the import commits. Even with test refactoring, maybe there I could have been more relaxed as well, will consider it for next time. Many thanks! 🙏 

> Please always consider splitting PRs into smaller chunks - each of the PRs can be merged into a working branch and you can request that only the final PR needs to be tested. Thanks!

I really considered that, it is just that the overall solution was keep getting a bit more involved as I was progressing with it. Now that you know what was involved, how would you split the PR further, can you make a suggestion? 🙏 

> ⚠️ When I turn off all feature flags -> open Activity Log -> click on Restore -> it takes +- 5s before the item at the top with progress appears. Before this PR, the item would appear right away.

Yes, nice catch, I should have mentioned that in the description as well. This is expectable because of the way the use case works and the 5' seconds delay. I didn't want to complicate things further thus I decided to not deal with it now. Do you this it is fine if you launch with that as an MVP and improve later. I'll then add a task on the parent ticket. 🌟 

> When I enable "restore" feature flag -> open Activity Log -> Click on "Restore" -> Click on "Restore site" -> Click on "Confirm restore site" -> click on "Ok notify me" -> the progress item doesn't appear, unless I pull down to refresh. (I think this behavior was the same even before this PR)

Yes, this is going to be part of my next PR. The current `develop` version is also not working for the `Ok, notify me!` or when going back manually. See `Hook wizard back into Activity Log List` task on the parent ticket. 🌟

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
ParaskP7: Your are totally right @malinajirka ! 🌟 

I am going to rename that to `published` and match that with the `ActivityLogModel`, since this is where that date comes from in the first place. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
ParaskP7: Same as above. 😄 

PS: Usually I try to avoid duplicate comment by adding a `PS: Xyz` on my first comment letting the author know that this comment applies elsewhere.

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
ParaskP7: @malinajirka 🙏 

I am so glad that you mentioned that and actually for saying that this is an indication that the code should be extracted into a separate class. This is exactly why I did that!

But, actually in this case I kind of had little luck in avoiding exposing the `reloadEvents(...)` since unfortunately I couldn't easily create all the combinations otherwise, in order to do the necessary refactoring, especially before adding the new solution.

Do you think now that a more simpler solution is in place I could possible make that back into a `private` function and simulate the exact some setup through the `public` entry points? I could probably achieve that but I might also miss some of it cases, not sure.

PS: I would personally prefer to leave it as such, hoping that later on it will be extracted out to a separate class.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
ParaskP7: I agree.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
ParaskP7: I kind of agree @malinajirka , for sure I understand why!

Would you prefer this version:

```
        return if (!networkUtilsWrapper.isNetworkAvailable()) {
            emit(NetworkUnavailable)
            true
        } else {
            false
        }
```

Or, this more compact one:

```
        return if (!networkUtilsWrapper.isNetworkAvailable()) {
            emit(NetworkUnavailable)
            true
        } else false
```

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
ParaskP7: I agree @malinajirka , it felt unintuitive to me too.

I was just extracting this logic to remove duplication and this was the default structure the IDE came up automatically. I then was thinking a bit about it and didn't like it but couldn't come up with a better naming, then I moved forward, thanks for bringing that up again.

What if I rename that function to `fetchActivitiesRewindAndReturnOnError(...)` or revert the logic and name it as `isFetchActivitiesRewindSuccessful(...)` to match the `isNetworkAvailable()`?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/PostRestoreUseCase.kt
ParaskP7: Yeap, another good catch @malinajirka , you, you, you are good! 😄 

So, the reason why I decided to go with `null` instead of `RewindRequestType(true, true, ...)` is because the `RewindStatusService.rewind(...)` function is also doing the same, so I didn't want to change the behavior as I wasn't sure if that will end up being a functional change. See `val rewindResult = activityLogStore.rewind(RewindPayload(site, rewindId))`.

On the `FluxC` level passing `null` versus `types` is different, see `ActivityLogRestClient.rewind(...)` function and more specifically this block of code:
```
        val typesBody = if (types != null) {
            mapOf("types" to types)
        } else {
            mapOf()
        }
```

So on what case you would add an `empty map`, while on the other case you would add a `map of types`.

Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
ParaskP7: This is done here: 4b5a9fda72cf636abcd3864be970fb1f02de7624

PS: Let me know if you want it renamed to something else. 🙏

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
ParaskP7: This is done here: 4b5a9fd

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
ParaskP7: This is done here: 830cc94e407f6313edd1bd97ac86fff9034db6f2

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
ParaskP7: This is done here: 5cc156dd031b34c4f95c64055633cb04e672c36e

I went with the more compact approach, let me know if you have another idea. 🙏

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
ParaskP7: This is done here: 05e4b5bfcb78ead282be53a1dcf6f93d2abbfb3f

I went with the `isFetchActivitiesRewindSuccessful(...)` approach, let me know if you have another idea. 🙏

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/PostRestoreUseCase.kt
zwarm: iirc, the empty map is acceptable in the API and is the default. I'll have to go back and test to be positive, that was so long ago.

zwarm:
@ParaskP7 - So much easier to understand, very nicely done! 🎆 

I've have a few takeaways that I will be refactoring in GetBackupDownloadStatus based off of the comments here:
- Move the flowOn operator/dispatcher to use case
- Replace Date with published in the states
- IsNetworkUnavailable to NetworkAvailable
However, I don't know if you have the backup integration waiting in a separate PR. Let me know and I won't touch them.

One comment not related to the code in this PR, but related to the use case.
-I still have an item on my list to add 3X error retry on fetch error which may alter `if (!isFetchActivitiesRewindSuccessful(site)) return@flow` - not truly a success, but not a failure ... don't want to exit flow, but restart the loop and not go to the next step. May be nice to sync up on this one. (I have already added the retry to the backup use case)

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
malinajirka: 👍 Thanks for the clarification! I understand now, why/when it was added. I agree with you that keeping it as is, is good enough for now ;)!

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreStates.kt
malinajirka: Awesome, thanks!

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
malinajirka: Thanks for the changes! I'm still on the edge - when I see `isFetchActivitiesRewindSuccessful` I don't expect the method will actually hit the server. What if we just named it "fetchActivitiesRewind" and return "true" on success and "false" on failure? Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/PostRestoreUseCase.kt
malinajirka: Yeah, I realize it'd slightly change the behavior but from my understanding the result will be the same. Having said that, let's not worry about that now and lets just stick to the nullable type. Thanks!

malinajirka:
> I really considered that, it is just that the overall solution was keep getting a bit more involved as I was progressing with it. Now that you know what was involved, how would you split the PR further, can you make a suggestion? 🙏

I usually prefer to have PRs with a single focus and this PR was making a lot of improvements not directly related to the task. I really love these improvements. I was wondering if we could have for example created a separate PR for test refactoring + the new tests added before you actually started working on the changes. Perhaps it wasn't feasible, it's hard to tell as a reviewer. Thank you so much for considering it! 

> Yes, nice catch, I should have mentioned that in the description as well. This is expectable because of the way the use case works and the 5' seconds delay. I didn't want to complicate things further thus I decided to not deal with it now. Do you this it is fine if you launch with that as an MVP and improve later. I'll then add a task on the parent ticket. 🌟

👍 

> Yes, this is going to be part of my next PR. The current develop version is also not working for the Ok, notify me! or when going back manually. See Hook wizard back into Activity Log List task on the parent ticket. 🌟

❤️

malinajirka (approved):

WordPress/src/test/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModelTest.kt
```diff
@@ -662,30 +568,568 @@ class ActivityLogViewModelTest {
...
-        birthday.set(1985, 8, 27)
-
-        val list = mutableListOf<ActivityLogModel>()
-        val activity = ActivityLogModel(
-                "", "", null, "", "", "",
-                "", true, "", birthday.time
-        )
-        list.add(activity)
-        list.add(activity.copy(rewindable = false))
+    private fun event() = ActivityLogListItem.Event(
```
ashiagr, line 1086: ❓: Functions with no arguments might be interchangeable with read-only properties. Any reason why we used a function `event()` here? Might help me understand when to prefer one over another.

ashiagr (approved):
@ParaskP7 , I reviewed each commit and appreciate the level of effort put into building the solution. I really liked the refactoring effort like
- re-ordering functions close to the places they're used
- creating private group
- extracting smaller functions out of bigger ones, increasing readability

Great job!

I agree with @malinajirka that this PR could've been split into smaller PRs. But I also understand that sometimes it is difficult to do that while being so much involved in building a solution. It happens to everyone of us. Nice thing is commits are split into separate ones specific to their purpose. We might consider cherry picking some of the similar commits (like for tests refactoring) and creating a separate PR out of it. It might be too much of an effort, and as always we have to find a balance. 

I just left one question. Approving it from side.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
ParaskP7: I agree. 🙏

ParaskP7: This is done here: 4de839318642e1b3beda35bda2c0c45bf24f2072

ParaskP7:
👋 @zwarm !

> @ParaskP7 - So much easier to understand, very nicely done! 🎆

Thank YOU! ❤️ 

> However, I don't know if you have the backup integration waiting in a separate PR. Let me know and I won't touch them.

I am going to create 2 PRs today, one for the restore integration and one for the backup integration. It is probably better if I do the change on the `GetBackupDownloadStatusUseCase` myself so that you can I don't block you and you can focus on other things, wdyt? 🙏 

> I still have an item on my list to add 3X error retry on fetch error which may alter if (!isFetchActivitiesRewindSuccessful(site)) return@flow - not truly a success, but not a failure ... don't want to exit flow, but restart the loop and not go to the next step. May be nice to sync up on this one. (I have already added the retry to the backup use case)

I remember the 3x error retry login on the `GetBackupDownloadStatusUseCase`. Feel free to introduce a similar logic for the `GetRestoreStatusUseCase` when this PR is merged. And definitely, let's sync! 🌟

ParaskP7:
👋 @malinajirka !

> I usually prefer to have PRs with a single focus and this PR was making a lot of improvements not directly related to the task. I really love these improvements. I was wondering if we could have for example created a separate PR for test refactoring + the new tests added before you actually started working on the changes. Perhaps it wasn't feasible, it's hard to tell as a reviewer. Thank you so much for considering it!

I so much agree, and apologies for making the move and created one PR instead of two, I could definitely do what you suggested (refactor + tests) and then the actual solution. Please expect that next time. I probably got caught in the fever of delivering since I have been delaying this solution more than I expected (or wanted). 🙏

ParaskP7:
👋 @ashiagr !

> @ParaskP7 , I reviewed each commit and appreciate the level of effort put into building the solution. I really liked the refactoring effort like

❤️ 

> ? I agree with @malinajirka that this PR could've been split into smaller PRs. But I also understand that sometimes it is difficult to do that while being so much involved in building a solution. It happens to everyone of us. Nice thing is commits are split into separate ones specific to their purpose. We might consider cherry picking some of the similar commits (like for tests refactoring) and creating a separate PR out of it. It might be too much of an effort, and as always we have to find a balance.

Yes, I totally agree with you all, I should have split this PR at least into two (refactor + tests). Thanks so much such suggestions. Please expect me to do that next time. 🙏

WordPress/src/test/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModelTest.kt
ParaskP7: Thanks for the question @ashiagr ! And what a great one indeed. 🌟

You will understand the answer quickly if you check the other such function further above, the `activity()` one, which accept parameters, which all of them default to some value. So, in the end you can use this function as `activity()` and get the default behavior of the (data) class, or update say the `rewindable` to `false` and get an activity that is not rewindable. With the `event()` I went with the same approach, but didn't need to add any variables (yet).

To be honest, there is a longer story to that. This is actually a `stub/fake` pattern that can be applied to any codebase, in terms of testing, where ALL those suchlike function can be created in one place, with all its parameters being available from the function, with default, so that whenever a developer needs to create say an `activity` for their test, they would just import this helper function from that common place and right away use it as `activity()`, without the need to do all this boilerplate setup over and over again. This really removes the clutter in setting up tests by a lot. Also, this helps removing the need to use `@Mock` on a `Site` class, which is not recommended, and have a common `site()` test helper function to have a site created as a stub/fake and not as a mock.

Does the above make sense or am I mumbling again? 😅

ashiagr: Thank you for an elaborate answer 🙏, it does make sense!

malinajirka:
> I so much agree, and apologies for making the move and created one PR instead of two, I could definitely do what you suggested (refactor + tests) and then the actual solution. Please expect that next time. I probably got caught in the fever of delivering since I have been delaying this solution more than I expected (or wanted). 🙏

Please don't apologize :). It's very difficult to find the balance before you start working on a task. As you progress it becomes more and more difficult to split the work up. Thank you for all the effort you put into documenting your every step and for the great detailed PR description! It helped all of us ;)

ParaskP7:
👋 @malinajirka !

> Yes, this is going to be part of my next PR. The current develop version is also not working for the Ok, notify me! or when going back manually. See Hook wizard back into Activity Log List task on the parent ticket. 🌟

The PR I was referring to is now ready, you can check it really quick if you are curious: #13914

PR #13903, opened 2021-01-28
Issue/13326 start scan after delay

ashiagr:
This PR fetches scan state after a delay once scan is started. It fixes an issue when scan state did not update immediately after starting the scan.

To test:
Prerequisite:

- You have access to WP.com account with a site having scan capability (e.g. https://pressable-jetpack-daily-scan.mystagingwebsite.com).

1. Login to the app as well Calypso using above WP.com account and select the site on the My Site tab
2. Click scan menu item on My Site in the app
3. Add/ delete threat to/ from the site from Calypso
4. Start scan few times in the app
5. Notice that scan state is updated appropriately in the app

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanStateUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanStateUseCase.kt
@@ -3,7 +3,6 @@ package org.wordpress.android.ui.jetpack.scan.usecases
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.flowOn
 import org.wordpress.android.fluxc.model.SiteModel
@@ -29,30 +28,29 @@ class FetchScanStateUseCase @Inject constructor(
         polling: Boolean = true,
         delayInMs: Long = FETCH_SCAN_STATE_DELAY_MILLIS
     ): Flow<FetchScanState> = flow {
-        if (!networkUtilsWrapper.isNetworkAvailable()) {
-            emit(Failure.NetworkUnavailable)
-            return@flow
-        }
+        while (true) {
+            if (!networkUtilsWrapper.isNetworkAvailable()) {
+                emit(Failure.NetworkUnavailable)
+                return@flow
+            }
 
-        val result = scanStore.fetchScanState(FetchScanStatePayload(site))
-        if (result.isError) {
-            emit(Failure.RemoteRequestFailure)
-            return@flow
-        } else {
-            val scanStateModel = scanStore.getScanStateForSite(site)
-            if (scanStateModel != null) {
-                emit(Success(scanStateModel))
-                if (polling && scanStateModel.state == ScanStateModel.State.SCANNING) {
-                    fetchScanState(site = site, polling = true).collect {
-                        emit(it)
-                        delay(delayInMs)
-                    }
-                } else {
-                    return@flow
-                }
-            } else {
+            val result = scanStore.fetchScanState(FetchScanStatePayload(site))
+            if (result.isError) {
                 emit(Failure.RemoteRequestFailure)
                 return@flow
+            } else {
+                val scanStateModel = scanStore.getScanStateForSite(site)
+                if (scanStateModel != null) {
+                    emit(Success(scanStateModel))
+                    if (polling && scanStateModel.state == ScanStateModel.State.SCANNING) {
+                        delay(delayInMs)
+                    } else {
+                        return@flow
+                    }
+                } else {
+                    emit(Failure.RemoteRequestFailure)
+                    return@flow
+                }
             }
         }
     }.flowOn(bgDispatcher)
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanStateUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanStateUseCase.kt
@@ -25,7 +25,6 @@ class FetchScanStateUseCase @Inject constructor(
 ) {
     suspend fun fetchScanState(
         site: SiteModel,
-        delayInMs: Long = FETCH_SCAN_STATE_DELAY_MILLIS,
         startWithDelay: Boolean = false
     ): Flow<FetchScanState> = flow {
         while (true) {
@@ -47,7 +46,7 @@ class FetchScanStateUseCase @Inject constructor(
                 if (scanStateModel != null) {
                     emit(Success(scanStateModel))
                     if (scanStateModel.state == ScanStateModel.State.SCANNING) {
-                        delay(delayInMs)
+                        delay(FETCH_SCAN_STATE_DELAY_MILLIS)
                     } else {
                         return@flow
                     }
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanStateUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanStateUseCase.kt
@@ -27,7 +27,7 @@ class FetchScanStateUseCase @Inject constructor(
         site: SiteModel,
         polling: Boolean = true,
         delayInMs: Long = FETCH_SCAN_STATE_DELAY_MILLIS,
-        startAfterDelay: Boolean = false
+        startWithDelay: Boolean = false
     ): Flow<FetchScanState> = flow {
         while (true) {
             if (!networkUtilsWrapper.isNetworkAvailable()) {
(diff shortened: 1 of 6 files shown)


WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanStateUseCase.kt
```diff
@@ -27,32 +26,36 @@ class FetchScanStateUseCase @Inject constructor(
     suspend fun fetchScanState(
         site: SiteModel,
         polling: Boolean = true,
```
ParaskP7: Minor (🔍): Consider removing this `polling` parameters as it is no longer necessary?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanStateUseCase.kt
```diff
@@ -27,32 +26,36 @@ class FetchScanStateUseCase @Inject constructor(
...
-        delayInMs: Long = FETCH_SCAN_STATE_DELAY_MILLIS
+        delayInMs: Long = FETCH_SCAN_STATE_DELAY_MILLIS,
```
ParaskP7: Warning (⚠️): I understand that you added this `delayInMs` as a parameter in order to be able to override that within its test suite. Note that this is problematic as it is generally advised to not do that, meaning adding functionality to the production code for test to work. For example, in this case, this allows clients to trigger the `fetchScanState(...)` function with any delay they like, which might cause confusion for future developers, since we actually don't want to allow that.

There is a really nice test-only related solution to this coroutines related problem that I advised Annmarie to follow as well (see below). I actually did the same on my own PR for that matter.

Test-only solution: Consider using the `MainCoroutineScopeRule` approach to fake the delay, by replacing `test { ... }` with `coroutineScope.runBlockingTest { ... }` and then add the `advanceTimeBy(DELAY_MILLIS)` right after the test act line. There are a couple of tests already using this approach, please search for usages of `MainCoroutineScopeRule` to find out more about it.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanStateUseCase.kt
```diff
@@ -27,32 +26,36 @@ class FetchScanStateUseCase @Inject constructor(
...
+        startAfterDelay: Boolean = false
```
ParaskP7: Minor (🔍): This is totally opinionated so feel free to ignore. I would have named this like:
- `startWithDelay` or,
- `startDelay`, removing the `After` word altogether.

PS: I don't know why but `After` sounds a bit tricky, maybe I am used to see this kind of parameters with `With` instead. Again, please ignore if you feel `After` is much better.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanStateUseCase.kt
```diff
@@ -27,32 +26,36 @@ class FetchScanStateUseCase @Inject constructor(
...
                 } else {
+                    emit(Failure.RemoteRequestFailure)
                     return@flow
                 }
-            } else {
-                emit(Failure.RemoteRequestFailure)
-                return@flow
             }
         }
     }.flowOn(bgDispatcher)
```
ParaskP7, line 59: Praise (❤️): I really like that you put the `flowOn(...)` operator here. I know this is not part of this PR, but I wanted to comment that because I think all our use cases should switch to that approach. @zwarm , I am mentioning you here so that you are aware.

@malinajirka , wdyt?

PS: I believe this is the most robust way of using the use cases seems the client should not (or in those case, we don't want them) to manage the dispatcher themselves. The dispatcher needs to be a background dispatcher every time this use case is being used and as such it should be placed within the use case itself

ParaskP7 (commented):
👋 @ashiagr  !

I have left some comments for you to consider.

PS: While testing this solution I actually managed to trigger `Scan now` twice for it to work, which we don't want to, right? I managed to do that by applying the `calendar-1.zip` plugin (which is a threat) via the web interface, then right after I saw that being applied and being a warning (looking on the web) I immediately triggered scan on the app. 5'' seconds past (the start delay) and that wasn't enough to trigger the progress. However, the second time I triggered it, 10'' seconds past this time, the progress start showing. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanStateUseCase.kt
ashiagr: Passing on praise to @malinajirka, it was suggested by him in a PR review :).

ParaskP7: Praise to the lord, the almighty! 😄

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanStateUseCase.kt
ashiagr: In ecbd477

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanStateUseCase.kt
ashiagr: Nice suggestion 👍
In 1243a65, bd5d4a2

I've used `MainCoroutineScopeRule` earlier, this is indeed cleaner!

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanStateUseCase.kt
ashiagr: In 43afa89

ashiagr:
Thanks for the review @ParaskP7! I've addressed your comments.

Regarding this: 

> While testing this solution I actually managed to trigger Scan now twice for it to work, which we don't want to, right? I managed to do that by applying the calendar-1.zip plugin (which is a threat) via the web interface, then right after I saw that being applied and being a warning (looking on the web) I immediately triggered scan on the app. 5'' seconds past (the start delay) and that wasn't enough to trigger the progress. However, the second time I triggered it, 10'' seconds past this time, the progress start showing. Wdyt?

Can you confirm if it happens also on Calypso or just in the app?

ParaskP7:
> Thanks for the review @ParaskP7! I've addressed your comments.

Thank you @ashiagr ! 🌟 I checked the updates and it all looks good to me now. 🙏 

> Can you confirm if it happens also on Calypso or just in the app?

Working on it, will let you know asap. 🏃

ParaskP7:
> Working on it, will let you know asap. 🏃

👋 @ashiagr ! Unfortunately, scan is not available for my site `petrosatwork.blog` in Calypso, or I am not able to find it somehow. If you remember I told you that as some point in the past. Could you suggest me a site that I can test this with, one without any current threats? All I do is add the `calendar-1.zip` and then trigger the scan.

ParaskP7 (approved):

ParaskP7:
I approved this PR to unblock you anyway, since everything else is okay.

It is just that I still triggered `Scan` twice to get it launch and start progressing for my site from within the app. On the web I cannot trigger it because the scan feature is not available (somehow). If you want we can sit together tomorrow and figure that out.

ashiagr:
Thanks @ParaskP7! 

>  Could you suggest me a site that I can test this with, one without any current threats? All I do is add the calendar-1.zip and then trigger the scan.

We can uninstall calender plugin on https://pressable-jetpack-daily-scan.mystagingwebsite.com and re-install it.
 
>  then right after I saw that being applied and being a warning (looking on the web) I immediately triggered scan on the app. 

> It is just that I still triggered Scan twice to get it launch and start progressing for my site from within the app. On the web I cannot trigger it because the scan feature is not available (somehow). If you want we can sit together tomorrow and figure that out.

Sure, let's sync and figure it out. Thanks for offering your time  🙇‍♀️.

ParaskP7:
> We can uninstall calender plugin on https://pressable-jetpack-daily-scan.mystagingwebsite.com and re-install it.

Great, I am on it as we speak and testing this site.

> Sure, let's sync and figure it out. Thanks for offering your time 🙇‍♀️.

Thank YOU! 🙏

ParaskP7:
> Great, I am on it as we speak and testing this site.

👋 @ashiagr !

Ok, so I just tried that 3 times on this site via the Calypso interface and I got it to not work as well once, when I did it fast (less than 5'' seconds). When I tried to wait for 5'' seconds it worked as expected. Maybe this is because it takes some few seconds to get everything in place backend-wise (db, etc) before a scan can know that a new plugin is available or something has changed on the website. Taking into consideration that a user would never be as fast with applying a plugin and then scanning right away, probably this will not be an issue.

PS: I think this 5'' seconds delay helps anyway with all the other scenarios that we have seen where a scan is triggered at any point.

ashiagr:
> Taking into consideration that a user would never be as fast with applying a plugin and then scanning right away, probably this will not be an issue.

> I think this 5'' seconds delay helps anyway with all the other scenarios that we have seen where a scan is triggered at any point.

Thank you @ParaskP7 for testing it out 🙏. I'll merge this PR now.

PR #13909, opened 2021-01-28
Handle site dependent site state in the improved my site fragment

### LINKED ISSUE #13884: Create child my site content fragment which always has selected site

planarvoid:
Fixes #13884 

I'm extracting the My Site sources in this PR. The source reacts to the selected site ID change and reloads/refetches all the data in the My Site fragment based on this ID. The implementation is done with LiveData and it makes it easy to add new sources of data.

To test:
- Make sure the Improved my site flag is turned on and the app is restarted
- Smoke test My Site screen with and without a site

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteSource.kt
```diff
@@ -0,0 +1,8 @@
+package org.wordpress.android.ui.mysite
+
+import kotlinx.coroutines.flow.Flow
+import org.wordpress.android.ui.mysite.MySiteUiState.PartialState
+
+interface MySiteSource<T : PartialState> {
+    fun buildSource(siteId: Int): Flow<T?>
+}
```
renanferrari: I really like the concept of `MySiteSource`!

WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteStateProvider.kt
```diff
@@ -0,0 +1,72 @@
...
+    val state: Flow<MySiteUiState> = selectedSiteRepository.siteSelected.asFlow().flatMapMerge { siteId ->
+        if (siteId != null) {
+            mySiteSources.map { it.buildSource(siteId).distinctUntilChanged() }.asFlow().flattenMerge()
+        } else {
+            flowOf()
+        }
+    }.let { partialStates ->
+                flow {
+                    var accumulator = MySiteUiState()
+                    emit(accumulator)
```
renanferrari: Emitting this here causes the empty view to flash every time we open the app or switch sites. If I'm not missing anything, I'd probably remove it.

WordPress/src/main/java/org/wordpress/android/ui/mysite/SelectedSiteRepository.kt
```diff
@@ -17,13 +24,18 @@ class SelectedSiteRepository
 ) {
     private var siteSettings: SiteSettingsInterfaceWrapper? = null
     private val _selectedSiteChange = MutableLiveData<SiteModel>()
+    private val _siteSelected = MutableLiveData<Int>()
     private val _showSiteIconProgressBar = MutableLiveData<Boolean>()
     val selectedSiteChange = _selectedSiteChange as LiveData<SiteModel>
+    val siteSelected = _siteSelected as LiveData<Int?>
```
renanferrari, line 23: I'm not sure if I understand the purpose of introducing `siteSelected`. Wouldn't we be able to use `selectedSiteChange` instead?

WordPress/src/main/java/org/wordpress/android/ui/mysite/ScanAndBackupSource.kt
```diff
@@ -0,0 +1,34 @@
...
+            val itemsVisibility = jetpackCapabilitiesUseCase.getJetpackPurchasedProducts(site.siteId)
+            emit(
+                    JetpackCapabilities(
+                            scanAvailable = scanScreenFeatureConfig.isEnabled() && itemsVisibility.scan,
+                            backupAvailable = backupScreenFeatureConfig.isEnabled() && itemsVisibility.backup
+                    )
+            )
+        }
+    }
+}
```
renanferrari, line 31: This is really neat! The only thing I'm not sure if I understand is why we still need to compare the `siteId` argument with the id from `selectedSiteRepository.getSelectedSite()`. I'm not very familiar with the Flow API, but since we're calling `buildSource` and creating a new Flow every time the site is switched, I'd expect the previous Flow to be cancelled, since we're not "subscribed" to it anymore. Is that not the case?

WordPress/src/main/java/org/wordpress/android/ui/mysite/DomainRegistrationHandler.kt
```diff
@@ -1,43 +1,58 @@
...
+) : MySiteSource<DomainCreditAvailable> {
+    private var continuation: CancellableContinuation<OnPlansFetched>? = null
+
+    override fun buildSource(siteId: Int) = flow {
+        continuation?.cancel()
+        continuation = null
+        val site = selectedSiteRepository.getSelectedSite()
+        if (site == null || site.id != siteId) {
+            return@flow
+        }
```
renanferrari, line 36: Same question I had about `ScanAndBackupSource`: I'm not sure why we need to compare the site ids here.

renanferrari (commented):
Hey, I just finished taking a good look at this and I really like the direction this is going! I still want to try out some more things tomorrow but I'll leave some thoughts and questions below for now.

WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteStateProvider.kt
planarvoid: the idea here (and maybe it's unnecessary) is to clear the data when you switch site before anything else is loaded. When I'm thinking about it, it might not be necessary because the "SiteModel" is emitted right after so it shouldn't be necessary to emit the empty model.

WordPress/src/main/java/org/wordpress/android/ui/mysite/SelectedSiteRepository.kt
planarvoid: `SiteModel` doesn't have an `equals` method so it emits much more often than the site is actually changed and we can't get distinct items. This causes the model to be redrawn much more often and it causes the performance issues. What would probably work is some custom distinct where we check the emitted items by IDs. However, the approach with 2 live data seemed simpler. 

This whole issue is caused by us misusing one event - OnSiteChanged - for two purposes, site selection and site updates.

WordPress/src/main/java/org/wordpress/android/ui/mysite/DomainRegistrationHandler.kt
planarvoid: we probably don't need that, I just wanted to be super sure that we're loading data for the correct site.

planarvoid:
hey @renanferrari , I've replied to all of your comments and made some quite significant changes to the code (especially introducing a source not depending on a selected site for the current avatar). Let me know what you think :-). I think it works well now 👍

renanferrari (approved):
Thanks for the changes! The code is looking pretty good and I think we're ready to give this a try 👍

PR #13910, opened 2021-01-28
Log error when throwing exception

zwarm:
Parent #13328 & #13329

This PR logs an error when:
1. An exception is thrown while checking for required intent parameters
2. When GetBackupDownloadStatusUseCase exceeds the retry limit

Please let me know if you feel this code is overkill. 

**Notes**
I was too optimistic with this PR. The AppLog is in WordPress.Utils and I not realize that there is already a [PR](https://github.com/wordpress-mobile/WordPress-Utils-Android/pull/59) out there that updated values, but that PR hasn't been merged yet. I have a request out to figure out how to get that pushed, then I will come back and update this one with the tag names.

To test:
- I couldn't find a way to test this without adding code to force an error, so there is nothing to test in this PR

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCase.kt
```diff
@@ -31,6 +33,8 @@ class GetBackupDownloadStatusUseCase @Inject constructor(
             val result = activityLogStore.fetchBackupDownloadState(FetchBackupDownloadStatePayload(site))
             if (result.isError) {
                 if (retryAttempts++ >= MAX_RETRY) {
+                    AppLog.e(T.JETPACK_BACKUP_DOWNLOAD,
+                            "GetBackupDownloadStatusUseCase: Exceeded 3 retries while fetching status")
```
ParaskP7: Minor (🔍): Consider using `${this.javaClass.simpleName}` instead of hardcoded class name, just like you did with the other two app logs.

ParaskP7 (commented):
👋 @zwarm !

I am just leaving a `Comment` here and not approving yet since the CI is not building due to what you have noted in the PR description. When this is ready again, let me and I'll approve right afterwards.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCase.kt
zwarm: > Minor (🔍): Consider using `${this.javaClass.simpleName}` instead of hardcoded class name, just like you did with the other two app logs.

That was explicit because when in coroutine you don't get the name of the outer class .. you get an obscure reference to "coroutine" or in this case the message would have been "SafeCollector: Exceeded 3 retries while fetching status". I want the actual name of the class that is throwing the error.

Thanks for taking a quick look. Still waiting on the Utils piece to be sorted. :)

zwarm: Hi @ParaskP7 

The utils pieces have been merged in. Once CI finishes we should be good to go. Thanks again!

ParaskP7: 👋 @zwarm !

Sorry, I was too quick into the `{this.javaClass.simpleName}` and thanks for noting that mistake. What about this alternative: `this@GetBackupDownloadStatusUseCase.javaClass.simpleName`

ParaskP7: PS: I am seeing the CI complaining still, is that because we need to trigger another run?

zwarm: > PS: I am seeing the CI complaining still, is that because we need to trigger another run?

Me too ... 👀  as I'm not sure why the latest doesn't get pulled in. It's not complaining about the pom like it does for FluxC.

ParaskP7: Maybe @malinajirka can help? Shameless mention... 😄

zwarm: >  `this@GetBackupDownloadStatusUseCase.javaClass.simpleName`

I bet this _would_ work .. let me test. 👍

malinajirka: ~It seems `1.30.3-beta.1` doesn't exist.~

~I think the build is not failing on missing pom because the Utils library is referenced (transitive dependency) from other dependencies we rely on. So it fetches a different version of the utils library and that's enough for gradle - however, the other version is missing the newest changes, hence the build fails. Wdyt?~

malinajirka: Another comment - we also have a duplicate version of AppLog in `test/org.wordpress.android.util`. Perhaps we need to add `JETPACK_BACKUP` there as well. Wdyt?

zwarm: > Another comment - we also have a duplicate version of AppLog in test/org.wordpress.android.util. Perhaps we need to add JETPACK_BACKUP there as well. Wdyt?

Yes, that would explain the failing test with java.lang.NoSuchFieldError. 🙇

zwarm: @ParaskP7 - finally. Many thank to Jirka, that failing test was driving me nuts - that's a great little nugget of WP knowledge to stash away.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreFragment.kt
```diff
@@ -89,7 +89,7 @@ class RestoreFragment : Fragment(R.layout.jetpack_backup_restore_fragment) {
                 site to activityId
             }
             else -> {
-                AppLog.e(T.JETPACK_RESTORE, "Error initializing ${this.javaClass.simpleName}")
+                AppLog.e(T.JETPACK_REWIND, "Error initializing ${this.javaClass.simpleName}")
```
ParaskP7, line 92: Question (❓): Why did this change to `REWIND`? Since rewind is used in general with backup download as well, as a term that is (like `rewindId`), I think it is better if we use `RESTORE` as a naming convention on restore specific cases. I try to do that everywhere, when refactoring field, parameters, functions, etc. Thus, I asking this question to understand if I am missing any information in regards to that.

ParaskP7 (approved):
👋 @zwarm !

I am approving this PR, I bet it wasn't fun on your part to go through all these for such a simple change. 😅 

Since I don't have the full context here, I am going to let you merge that to `develop` when you know it is the right time (don't want to break anything myself).

WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/RestoreFragment.kt
zwarm: > Why did this change to REWIND?

Because I chose RESTORE before I realized that the value was already defined in Utils. I used the value that was preexisting.

PR #13914, opened 2021-01-29
Jetpack Section: Connect Restore with Activity Log/Backup Screen

ParaskP7:
Parent #13328

This PR connects the restore flow with the activity log/backup screens. This means that this PR makes sure that when a user presses the `Ok, notify me!` button or presses/swipes back (triggering the `onBackPressed()`), while restore is in progress from within the restore screen, then a restore progress row appears on the activity log screen immediately. Before, a user would had to do a manual re-launch or pull-to-refresh on the activity log/backup screen to get the progress row appear.

To test:
* Enable `RestoreFeatureConfig`, scroll down and click the `RESTART THE APP` button. 
* Find a rewindable item from the list, click on the ellipsis menu item and select `Restore to this point` to launch the `Restore` screen,
* You should automatically be navigated back to the `Activity Log` screen and a restore status will be triggered,
* A snackbar should immediately appear informing that the restore just started. Make sure the restore timestamp is available on the snackbar.
* Make sure the ellipsis menu item, for every rewindable item, is not shown while restore progress is ongoing.
* Wait for restore to finish while having the `Activity Log` screen open both, within the app and the web (so you can see the actual progress),
* When restore finishes (looking on the web), a restore (push) notification should appears.
* Then, the restore progress item should disappear (after at least 5'' seconds, due to the while loop check status delay).
* A pull-to-refresh should be automatically triggered and a new restore item might (or might not) appear (depending on how fast the restore progress item disappears and how fast this new item becomes available on the backend, via the `activityLogStore.fetchActivities(...)` API call). In case the new restore item does not appear, wait a bit (maybe 10'' seconds) and trigger a manual pull-to-refresh. Verify that the new restore item appears after some time.
* A snackbar should immediately appear informing that the restore just completed. Make sure the restore timestamp is available on the snackbar.
* Make sure the ellipsis menu items, for every rewindable item, is now shown again.

PS: As always, I suggest doing a commit-by-commit review, from within the IDE, as this will make it much easier for you to review the overall solution. If you need to focus on the feature itself and not the refactor or tests filter the commits by `Feature` to get a quick overall.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/RequestCodes.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/RequestCodes.java
@@ -28,6 +28,7 @@ public class RequestCodes {
     public static final int SMART_LOCK_READ = 1500;
     public static final int NOTIFICATION_SETTINGS = 1600;
     public static final int ACTIVITY_LOG_DETAIL = 1700;
+    public static final int RESTORE = 1720;
     public static final int PAGE_PARENT = 1800;
     public static final int HISTORY_DETAIL = 1900;
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/RequestCodes.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/RequestCodes.java
@@ -28,6 +28,7 @@ public class RequestCodes {
     public static final int SMART_LOCK_READ = 1500;
     public static final int NOTIFICATION_SETTINGS = 1600;
     public static final int ACTIVITY_LOG_DETAIL = 1700;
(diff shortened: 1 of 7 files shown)


zwarm (approved):
@ParaskP7 - Nice work. I was able to see the starting snackbar, the progress bar, and the final snackbar on complete. Dates were visible. I really like the improvements in the Viewmodels. 👍

ParaskP7:
> @ParaskP7 - Nice work. I was able to see the starting snackbar, the progress bar, and the final snackbar on complete. Dates were visible. I really like the improvements in the Viewmodels. 👍

So many thanks @zwarm ! 🌟

PR #13920, opened 2021-01-31
Issue/13326 scan full screen loading & error handling

ashiagr:
Parent #13326

This PR displays full screen loading and errors on fetch scan and start scan requests. 

Full screen loading/ errors
Loading (0) | No Network (1) | Generic Request Failed (3)  | Start Scan Request Failed (4) 
---------|-------------|------------------------|--------------------------
loading_scan| no-network-full screen |generic_request_failed|Something went wrong_start_scan

Snack bars
No Network (5)| Generic Request Failed (6)
-----------| -----------------------
no network snackbar|generic_request_failed_snackbar

#### Loading (0), No Network (1)

1. Fresh install the app and logon to wp.com account with site having scan capability
2. Switch to site with scan capability
3. Enable airplane mode
4. Click Scan menu
5. Notice that fullscreen "Loading scan" is shown, followed by full screen "No network available" message
6. Disable airplane mode
7. Click retry
8. Notice that scan content is shown
 
------------------------------------------------------------

#### No network snackbar (5)

1. Continuing from above, while being on the scan screen, enable airplane mode
2. Click scan button
3. Notice that a snackbar is shown with  no network available message

------------------------------------------------------------

#### Generic Request Failed (3), Start Scan Request Failed (4), Generic Request Failed snackbar (6)
These states can be tested by using
`ScanViewModel` tests:

Generic Request Failed (3)
`given fetch scan fails, when scan state fetched over empty scan state, then request failed ui state shown`
`given request failed error ui state, when contact support is clicked, then contact support screen is shown`

Start Scan Request Failed (4)
`given scan start request fails, when scan button is clicked, then request failed state is shown`
`given start scan request failed error state, when contact support is clicked, then contact support shown`

Generic Request Failed snackbar (6)
`given fetch scan state fails, when scan state fetched over last scan state, then request failed msg shown`

Notes: 

1. Start scan request fail displays a full screen error with contact support button, message string is copied from the iOS app. 
2. I plan to implement pull-to-refresh which will enable users to retry fetch scan state on full screen errors. 

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
@@ -49,6 +49,7 @@ class ScanFragment : Fragment(R.layout.scan_fragment) {
         recycler_view.addItemDecoration(
             HorizontalMarginItemDecoration(resources.getDimensionPixelSize(R.dimen.margin_extra_large))
         )
+        recycler_view.setEmptyView(actionable_empty_view)
         initAdapter()
     }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
@@ -45,15 +45,15 @@ class ScanFragment : Fragment(R.layout.scan_fragment) {
     }
 
     private fun initRecyclerView() {
+        recycler_view.itemAnimator = null
+        recycler_view.addItemDecoration(
+            HorizontalMarginItemDecoration(resources.getDimensionPixelSize(R.dimen.margin_extra_large))
+        )
         initAdapter()
     }
 
     private fun initAdapter() {
         recycler_view.adapter = ScanAdapter(imageManager, uiHelpers)
-        recycler_view.itemAnimator = null
-        recycler_view.addItemDecoration(
-            HorizontalMarginItemDecoration(resources.getDimensionPixelSize(R.dimen.margin_extra_large))
-        )
     }
 
     private fun initViewModel(site: SiteModel) {
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
@@ -77,7 +77,8 @@ class ScanFragment : Fragment(R.layout.scan_fragment) {
                     is FullScreenLoadingUiState -> updateFullScreenLoadingLayout(uiState)
 
                     is ErrorUiState.NoConnection,
-                    is ErrorUiState.GenericRequestFailed -> updateErrorLayout(uiState as ErrorUiState)
+                    is ErrorUiState.GenericRequestFailed,
+                    is ErrorUiState.StartScanRequestFailed -> updateErrorLayout(uiState as ErrorUiState)
                 }
             }
         )
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
@@ -17,6 +17,7 @@ import org.wordpress.android.ui.RequestCodes
 import org.wordpress.android.ui.jetpack.scan.ScanNavigationEvents.OpenFixThreatsConfirmationDialog
 import org.wordpress.android.ui.jetpack.scan.ScanNavigationEvents.ShowThreatDetails
 import org.wordpress.android.ui.jetpack.scan.ScanViewModel.UiState.ContentUiState
+import org.wordpress.android.ui.jetpack.scan.ScanViewModel.UiState.FullScreenLoadingUiState
 import org.wordpress.android.ui.jetpack.scan.adapters.HorizontalMarginItemDecoration
 import org.wordpress.android.ui.jetpack.scan.adapters.ScanAdapter
 import org.wordpress.android.ui.jetpack.scan.details.ThreatDetailsFragment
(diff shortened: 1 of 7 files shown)


WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
```diff
@@ -262,7 +291,41 @@ class ScanViewModel @Inject constructor(
         )
     )
 
-    sealed class UiState { // TODO: ashiagr add states for loading, error as needed
-        data class Content(val items: List<JetpackListItemState>) : UiState()
+    sealed class UiState {
+        object FullScreenLoadingUiState : UiState() {
+            @DrawableRes val image = R.drawable.img_illustration_empty_results_216dp
```
malinajirka: 🔍 Nitpick: The illustration looks a bit weird on "loading" screen. I'd consider showing a progress instead of the "result empty" illustration. Wdyt?


P.S. Or a skeleton with a Shimmer effect :).

WordPress/src/main/res/values/strings.xml
```diff
@@ -1061,6 +1061,14 @@
     <string name="scan_history">Scan History</string>
     <string name="scan_history_menu_title">History</string>
 
+    <string name="scan_loading_title">Loading scan</string>
```
malinajirka: 🔍 Nitpick: We might want to consider adding "..." at the end of this string resource. Wdyt?

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/ScanViewModelTest.kt
```diff
@@ -89,6 +93,18 @@ class ScanViewModelTest : BaseUnitTest() {
         )
     }
 
+    @Test
+    fun `given last scan state not present in db, when vm starts, then loading scan state is displayed`() = test {
+        whenever(scanStore.getScanStateForSite(site)).thenReturn(null)
+        val uiStates = init().uiStates
+
+        val state = uiStates.first() as FullScreenLoadingUiState
```
malinajirka: 🔍 Nitpick: It seems that this tests doesn't have just a single reason to fail. We might want to consider splitting it into two tests.
1. One that check that `uiStates.first()` is instance of `FullScreenLoadingUiState`
2. One that checks that content of `FullScreenLoadingUiState` is what we expect

Same applies to some of the other tests, wdyt?

WordPress/src/main/res/values/strings.xml
```diff
@@ -1061,6 +1061,14 @@
...
+    <string name="scan_no_network_title" translatable="false">@string/no_network_title</string>
+    <string name="scan_no_network_subtitle" translatable="false">@string/no_network_message</string>
+    <string name="scan_request_failed_title" translatable="false">Something went wrong</string>
```
malinajirka: ⚠️ It seems this string uses "translatable=false" attribute, even though it contains a translation and not just a reference. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
```diff
@@ -66,8 +71,14 @@ class ScanFragment : Fragment(R.layout.scan_fragment) {
         viewModel.uiState.observe(
             viewLifecycleOwner,
             { uiState ->
-                if (uiState is Content) {
-                    refreshContentScreen(uiState)
+                when (uiState) {
```
malinajirka, line 78: ❓ Do we need to update visibility of these views? Or we don't need to do that, because the RV is always empty when a loading/error fullscreen view is shown?

malinajirka (changes requested):
Thanks @ashiagr! Great job ;). It looks good overall, I've left one blocking comment ("translatable=false") and a few nitpicks.

🔍 Nitpick: When I click on "retry" when my internet connection is still disabled, there is no loading indicator  ->  it seems like the button doesn't work. We usually add "500ms" delay in such scenarios so the loading is displayed for a bit.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanFragment.kt
ashiagr: We don't need to do that, because the RV is always empty when a loading/error fullscreen view is shown.

WordPress/src/main/res/values/strings.xml
ashiagr: Good catch. After I added it, I searched for "Something went wrong" in the `strings.xml `as I remembered I saw it earlier. But it wasn't the same and had an extra string added to it, then I forgot to remove "translatable=false" 🤦‍♀️.

In cf7dfd6.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/ScanViewModelTest.kt
ashiagr: It confuses me always, split done in 273866b.

WordPress/src/main/res/values/strings.xml
ashiagr: Nitpick accepted, added '...' in cf7dfd6.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
ashiagr: ~~I'll add a skeleton with a Shimmer effect.~~

ashiagr: As discussed, ended up using progress bar in d9ea5d2

ashiagr:
Thanks for the review @malinajirka! I've addressed your comments and answered questions.

>  Nitpick: When I click on "retry" when my internet connection is still disabled, there is no loading indicator -> it seems like the button doesn't work. We usually add "500ms" delay in such scenarios so the loading is displayed for a bit.

Added retry delay in d9ea5d2.

Ready for another round.

malinajirka (approved):
LGTM, thanks! ;)

PR #13921, opened 2021-01-31
Remove setHasStableIds

zwarm:
parent #13328 

This PR addresses a [comment](https://github.com/wordpress-mobile/WordPress-Android/pull/13900#pullrequestreview-578171638) in PR #13900

"When I click on "Restore site" it navigates me to Warning screen -> there is a noticeable animation glitch -> the cancel button appears after a delay and the confirmation button changes it's vertical position a bit"

To fix the issue I removed `setHasStableIds(true)` because the adapter is being reused and the view holders are not the same. 

**To test:**
- Launch the app
- Navigate to Activity Log
- Tap on the more menu for an item
- Tap on Restore to this point
- Tap on Restore Site to get to the warning view
- Note that the Cancel button doesn't jump and confirmation button doesn't change position.

PR submission checklist:


ParaskP7 (approved):
👋 @zwarm  !

I have reviewed and tested this PR as per the instruction, good fix! 🌟

PR #13924, opened 2021-02-01
Issue/13326 navigate to scan screen from detail

malinajirka:
Parent issue #13326

This PR update the approach used to propagate data from ThreatDetail screen to Scan screen. 

The previous approach was using "startActivityForResult(ThreatDetail)/setResult". The ThreatDetail screen would set a flag in the result and the Scan screen would read this flag and force update the screen. This approach is great when the ThreatDetail screen is always started from Scan screen, but it's not the case anymore. We need to start ThreatDetail screen even from Scan History screen and perhaps even from notifications in the near future.

The new approach uses "startActivity(ThreatDetail)/startActivity(Scan)" . The ThreatDetail screen requests "startActivity(Scan screen)" with "clear top | single top" flags. If the Scan screen is already on the backstack, it's "onNewMethod" is invoked and the screen is force updated. If the Scan screen is not on the backstack, it's started from scratch in which case we don't need to force update it, because the screen always fetches new data on initialization.

Known issues:
1. When "fix threat" action is invoked from a detail of an ignored threat, the progress indicator on the Scan screen is not visible.

To test:
Prerequisites:
- A threat is added to the testing site. In case you run out of threats, re-install bad calendar plugin and re-run scan.
- Scan feature flag is enabled
1. Open Scan screen
2. Start scan
3. Open detail of the found threat
4. Click on "Fix threat" button
5. Notice the app redirects you back to the "Scan" screen and the correct progress state + snackbar is shown

----------------------------
Prerequisites:
- A threat is added to the testing site. In case you run out of threats, re-install bad calendar plugin and re-run scan.
- Scan feature flag is enabled
1. Open Scan screen
2. Start scan
3. Open detail of the found threat
4. Click on "Ignore threat" button
5. Notice you get redirected back to the "scan" screen and snackbar is shown
6. Open History
7. Open detail of the ignored threat
8. Click on "Fix threat" button
9. Notice the app redirects you back to the "Scan" screen, snackbar is shown (there is a known bug, the progress indicator is not visible)

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
@@ -10,6 +10,7 @@ import android.text.TextUtils;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.StringRes;
 import androidx.core.app.TaskStackBuilder;
 import androidx.fragment.app.Fragment;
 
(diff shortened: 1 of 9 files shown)


ashiagr (approved):
Works like a charm 🎉 except of the progress bar display issue which we discussed will be fixed in a separate PR.

Thanks for this refactoring to support fixing from scan history 👍!

PR #13926, opened 2021-02-01
Jetpack Section: Connect Backup Download with Activity Log/Backup Screen

ParaskP7:
Parent #13329

This PR connects the backup download flow with the activity log/backup screens. This means that this PR makes sure that when a user presses the `Ok, notify me!` button or presses/swipes back (triggering the `onBackPressed()`), while backup download is in progress from within the backup download screen, then a backup progress row appears on the activity log screen immediately.

⚠️ It seems that on some very rare occasions when the backup download `data` is empty and a `null` model is returned back to the `store` then this can end up not overriding the previous progressing state, leaving the progress item loading on screen indefinitely (see `ActivityLogStore.fetchBackupDownloadState(...)`). @zwarm we should discuss that together to figure out what to do with that since just returning a `null` model it seems to not cover all cases to a 100%.

To test:

#### Scenario.1 (Backup download flow)

* Enable `BackupDownloadFeatureConfig`, scroll down and click the `RESTART THE APP` button. 
* Find a rewindable item from the list, click on the ellipsis menu item and select `Download backup` to launch the `Backup Download` screen,
* Click the `Create downloadable file` button.
* You should automatically be navigated back to the `Activity Log` screen and a backup status will be triggered,
* The backup progress item should immediately appears. Make sure the backup timestamp is available on this progress item.
* A snackbar should immediately appear informing that the backup just started. Make sure the backup timestamp is available on the snackbar.
* Make sure the ellipsis menu item, for every rewindable item, is not shown while backup progress is ongoing.
* Wait for backup to finish while having the `Activity Log` screen open both, within the app and the web (so you can see the actual progress),
* When backup finishes (looking on the web), a backup (push) notification should appears.
* Then, the backup progress item should disappear (after at least 5'' seconds, due to the while loop check status delay).
* A pull-to-refresh should be automatically triggered.
* A snackbar should immediately appear informing that the backup just completed. Make sure the backup timestamp is available on the snackbar.
* Make sure the ellipsis menu items, for every rewindable item, is now shown again.

#### Scenario.2 (Restore and backup download flow)

* Do all the above, but combine that with triggering a `Restore` first.
* Make sure that both the restore and backup progress item appear (and disappear) on/from the screen.
* Make sure that both the restore and backup snackbars appear (and disappear) on/from the screen

PS: As always, I suggest doing a commit-by-commit review, from within the IDE, as this will make it much easier for you to review the overall solution. If you need to focus on the feature itself and not the refactor or tests filter the commits by `Feature` to get a quick overall.

**Merge Instructions**

1. ~~Make sure PR #13914 is merged to `develop`~~.
2. ~~Remove the `Not Ready for Merge` label~~.
3. Merge this PR.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -554,6 +618,56 @@ class ActivityLogViewModel @Inject constructor(
...
+                reloadEvents(
+                        backupDownloadEvent = BackupDownloadEvent(
+                                displayProgress = true,
+                                isCompleted = false,
+                                rewindId = state.rewindId,
+                                published = state.published
+                        )
+                )
+            }
+            is BackupDownloadRequestState.Complete -> if (isBackupDownloadProgressItemShown) {
```
ashiagr: Nitpick 🔍: In a when statement, if a branch is more than a single line, consider separating it from adjacent case blocks with a blank line.

Feel free to ignore it if you preferred consistency over this.

ashiagr (commented):
Great job @ParaskP7! I reviewed each commit and every commit message included the right information! If I noted any minor issue in one commit, I saw it fixed in the next few commits, so I don't really have any review comments :).

I just noticed that sometimes progress bar wasn't displayed even after snackbar to inform download has started was shown.

https://user-images.githubusercontent.com/1405144/106577154-0631e400-6564-11eb-81e4-6ea058d2f718.mov

Is it related to the issue you mentioned in the description?

Tested on: https://pressable-jetpack-complete.mystagingwebsite.com/

ParaskP7:
> Great job @ParaskP7! I reviewed each commit and every commit message include the right information! If I noted any minor issue in one commit, I saw it fixed in the next few commits, so I don't really have any review comments :).

🙏 

> I just noticed that sometimes progress bar wasn't displayed even after snackbar to inform download has started was shown.

Thanks so much for the thorough testing @ashiagr ! 🌟 

From the video you shared, I am seeing a couple of things happening:
1) In general, the `Backup Download` is very fast, it might even last few seconds, like less than `5 secs`. This is at least what I experienced when testing my site (https://petrosatwork.blog). This means that if you exclude items from the download, it would be even faster. Also, we have a delay of `5 secs` which might magnify this behavior. For example, if we had a delay of `1 secs` maybe that would have been better. What I tried to do is to launch a `Backup Download` from its dedicated screen (as you did) and immediately press the `Ok, notify me` or back button to check behavior. I am not sure If this can be further optimised.
2) I noticed that you triggered a pull-to-refresh right after triggering `Ok, notify me`. This again might caused an issue as such. A user would normally not need to do that and the `Backup` progress row should appear immediately (or after a quick second at least).

> Is it related to the issue you mentioned in the description?

Actually no, this is not related to the issue I mentioned. The issue I mentioned is causing a `Backup` progress row to keep being persisted in the UI even when a `Backup Download` is complete. This is due to the fact that we do not persist in the database when a empty response is returned from the API (or clear the state in the database for that matter). Thus, we might get a `Backup Download` persisted at ie. `88% progress`, then instead of getting a complete event, we might get an empty event, at which case the `88% progress` persisted state will not be reset and shown in the UI until another `Backup Download` get triggered. We already discuss about it with @zwarm yesterday and we said we will add a couple of action items on the parent tickets, both `Restore` *13328 and** `Backup Download` #13329, since it might even happen for `Restore`, and deal with them later on. @malinajirka I am shamelessly mentioning you here as well so you are aware of that.

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
ParaskP7: 👋 @ashiagr !

Thanks for the comment. 🙏 Can you please elaborate on that because I am not 100% sure I understand?

Also, what if I do the below and extract the logic in private fucntions:
```
    private fun handleBackupDownloadStatus(state: BackupDownloadRequestState) {
        when (state) {
            is BackupDownloadRequestState.Progress -> handleBackupDownloadStatusForProgress(state)
            is BackupDownloadRequestState.Complete -> handleBackupDownloadStatusForComplete(state)
            else -> Unit // Do nothing
        }
    }
```

ParaskP7: PS: Same apply for the `Restore` status:
```
    private fun handleRestoreStatus(state: RestoreRequestState) {
        when (state) {
            is RestoreRequestState.Progress -> handleRestoreStatusForProgress(state)
            is RestoreRequestState.Complete -> handleRestoreStatusForComplete(state)
            else -> Unit // Do nothing
        }
    }
```

ashiagr: It's such a minor thing, I just wanted to know what you think of this example from [Formatting control flow statements](https://kotlinlang.org/docs/reference/coding-conventions.html#formatting-control-flow-statements) 

```
private fun parsePropertyValue(propName: String, token: Token) {
    when (token) {
        is Token.ValueToken ->
            callback.visitValue(propName, token.value)
<blank line>
        Token.LBRACE -> { // ...
        }
    }
}
```

> what if I do the below and extract the logic in private fucntions:

Works for me 👍

ashiagr:
> In general, the Backup Download is very fast, it might even last few seconds, like less than 5 secs. This is at least what I experienced when testing my site (https://petrosatwork.blog). This means that if you exclude items from the download, it would be even faster. Also, we have a delay of 5 secs which might magnify this behavior. For example, if we had a delay of 1 secs maybe that would have been better. 

Yep I agree, progress bar might not have been shown as the download might have been completed very fast.
Shouldn't we still see a snackbar that download was complete? I think I didn't see it either.

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
ParaskP7: Hmm... Didn't know that this convention existed @ashiagr , thanks for letting me know! 🙏 

I think what that convention tries to say applies when you don't have braces, then it does make sense to have an empty line in between in order to avoid confusion. 👍 However, we you do have braces, maybe this does not apply since the end brace is already adding this effect, wdyt?

PS: I make the extract refactoring for now, but will definitely keep that in mind going forward, thanks for sharing! ❤️

ParaskP7: This is now done here: b0c8190125bd76c4d8c2d20d5bb382a06f4fd945

ashiagr: Thank you for the explanation, makes sense 🙇‍♀️.

ParaskP7:
> Yep I agree, progress bar might not have been shown as the download might have been completed very fast.
Shouldn't we still see a snackbar that download was complete? I think I didn't see it either.

Unfortunately, the way it is done atm, we should only see the snackbar if the code goes through the `Complete` status, which will then trigger the `reloadEvent(...)` with a `BackupDownloadEvent.isCompleted` field set to `true`. But, in that case we are not going through that flow since the `Activity Log` screen will not receive this event at all.

ashiagr:
> Unfortunately, the way it is done atm, we should only see the snackbar if the code goes through the Complete status, which will then trigger the reloadEvent(...) with a BackupDownloadEvent.isCompleted field set to true. But, in that case we are not going through that flow since the Activity Log screen will not receive this event at all.

Alright, I just wanted to discuss this scenario before approving the PR. I think it is fine to be with this behaviour for now. I'm approving this PR now. Feel free to merge it when CI is done. Issue you mentioned in the description can be fixed in another PR.

ashiagr (approved):

ParaskP7:
> Alright, I just wanted to discuss this scenario before approving the PR. I think it is fine to be with this behaviour for now. I'm approving this PR now. Feel free to merge it when CI is done. Issue you mentioned in the description can be fixed in another PR.

Once again, thank you so much for the thorough review! 🙏 

I believe you are correct, I have the same feeling about it. I think it is okay to have such a behavior for the MVP. Also, it might be okay to proceed with merging even with the warning (⚠️) issue I mentioned in my description. We can keep improving as we go.

Issue #13846, opened 2021-01-22
Add functionality to "Follow other sites" QS task


renanferrari:
When clicked, the QS task should show a focus point on the Reader tab.  The task should be completed after clicking that tab. On the Reader screen, a Snackbar should indicate that the user should click on the "Search" button in the toolbar.

Step 1 | Step 2
------------ | -------------
![follow1](https://user-images.githubusercontent.com/830056/105430211-c698fd00-5c31-11eb-8744-b486a1a379a5.png) | ![follow2](https://user-images.githubusercontent.com/830056/105430214-c862c080-5c31-11eb-9628-a1188e58c972.png)

PR #13928, opened 2021-02-01
Add post publish QS task

### LINKED ISSUE #13845: Add functionality to "Publish a post" QS task
When clicked, the QS task should show a focus point on the My Site FAB. After clicking the FAB, it should open the main bottom sheet and show a focus point on the "Blog post" item.  The task should be completed after clicking that item.

Step 1 | Step 2
------------ | -------------
publish1 | publish2

planarvoid:
Fixes #13845

This PR adds logic to show the quick start focus point on the FAB from the improved my site fragment. I was considering moving the FAB to the fragment (because I think it logically belongs there) but the changes I'd have to make were massive and the code with two access point gets really complicated. In the end the code shows the FAB from the fragment but handles the logic to hide as before - through the view model which now calls QuickStartRepository when the flag is enabled.

publish_post_video

To test:
- Turn on the MySiteImprovements flag and restart the app
- Make sure you are using a new site with Quick Start in progress
- Click on "Publish post" QS card
- Notice a Snackbar with a correct message
- Notice a focus point on the FAB
- Click on another QS card ("Update title")
- Notice the focus point is gone
- Click on the "Publish post" QS card again
- Click on the FAB
- Notice the focus point on the "Blog post" item
- Click on the "Blog post" item 
- Go back
- Notice the "Publish post" card is now gone

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
@@ -426,7 +426,7 @@ public class WPMainActivity extends LocaleAwareActivity implements
             switch (createAction) {
                 case CREATE_NEW_POST:
                     // complete quick start task outside of QS process
-                    if (getSelectedSite() != null) {
+                    if (getSelectedSite() != null && !mMySiteImprovementsFeatureConfig.isEnabled()) {
                         QuickStartUtils.completeTaskAndRemindNextOne(
                                 mQuickStartStore,
                                 QuickStartTask.PUBLISH_POST,
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
@@ -13,6 +13,7 @@ import android.text.TextUtils;
 import android.view.HapticFeedbackConstants;
 import android.view.View;
 import android.view.ViewGroup;
+import android.view.ViewParent;
 import android.widget.TextView;
 import android.widget.Toast;
 
@@ -228,6 +229,26 @@ public class WPMainActivity extends LocaleAwareActivity implements
         boolean onActivityBackPressed();
     }
 
+    private Runnable mShowFabFocusPoint = new Runnable() {
+        @Override public void run() {
+            if (isFinishing()) {
+                return;
+            }
+            boolean focusPointVisible =
+                    findViewById(R.id.fab_container).findViewById(R.id.quick_start_focus_point) != null;
+            if (!focusPointVisible) {
+                int horizontalOffset = getResources().getDimensionPixelOffset(
+                        R.dimen.quick_start_focus_point_my_site_right_offset
+                );
+                int focusPointSize = getResources().getDimensionPixelOffset(R.dimen.quick_start_focus_point_size);
+                QuickStartUtils.addQuickStartFocusPointAboveTheView(
+                        findViewById(R.id.fab_container), mFloatingActionButton, horizontalOffset,
+                        (mFloatingActionButton.getHeight() - focusPointSize) / 2
+                );
+            }
+        }
+    };
+
     @Override
     public void onCreate(Bundle savedInstanceState) {
         ProfilingUtils.split("WPMainActivity.onCreate");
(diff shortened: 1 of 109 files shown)


renanferrari (commented):
Thanks for the changes! Code looks pretty good and I've tested the app and everything seems to work fine, except for the scenario in the video below:

https://user-images.githubusercontent.com/830056/106543906-6897cf80-64e5-11eb-9720-9f19d46096ac.mp4

While working on #13846, I also came across similar issues myself as that task requires us to show a focus point on the navigation bar. I decided to try a different approach there, by treating that focus point as an event rather than a state and it seemed to work well. 

I've made [a draft of that implementation available here](https://github.com/wordpress-mobile/WordPress-Android/pull/13934), so you can take a look. If you think it's worth it, we can try following the same approach here. If not, we could try something else to achieve some consistency between the implementations. Either way, let me know what you think.

planarvoid:
Great catch @renanferrari . I've looked into it and I've actually implemented the logic differently so it makes more sense. I've added the `focusPoint` field to the FAB view model. Since it's already hidden/shown based on the live data, I've just added one additional source and that's the active task. The change of the active task now redraws the FAB and hides/shows it. I'm much happier with this solution :-). Her's how it works now:

publish_post_fixed

WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
```diff
@@ -420,13 +441,31 @@ private void initViewModel() {
...
+                    mHandler.removeCallbacks(mShowFabFocusPoint);
+                    mHandler.post(() -> {
+                        View focusPointView =
+                                findViewById(R.id.fab_container).findViewById(R.id.quick_start_focus_point);
+                        if (focusPointView != null) {
+                            ViewParent directParent = focusPointView.getParent();
+                            if (directParent instanceof ViewGroup) {
+                                ((ViewGroup) directParent).removeView(focusPointView);
+                            }
+                        }
```
renanferrari, line 458: It seems that we could replace this with `QuickStartUtils::removeQuickStartFocusPoint`?

WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
```diff
@@ -173,7 +173,9 @@ class ImprovedMySiteFragment : Fragment(),
             it?.let { uiModel ->
                 loadGravatar(uiModel.accountAvatarUrl)
                 when (val state = uiModel.state) {
-                    is State.SiteSelected -> loadData(state.items)
+                    is State.SiteSelected -> {
+                        loadData(state.items)
+                    }
```
renanferrari: Big nitpick here, just to maintain consistency.

```suggestion
                    is State.SiteSelected -> loadData(state.items)
```

WordPress/src/main/java/org/wordpress/android/ui/mysite/QuickStartRepository.kt
```diff
@@ -60,9 +63,10 @@ class QuickStartRepository
     private val detailsMap: Map<QuickStartTask, QuickStartTaskDetails> = QuickStartTaskDetails.values()
             .associateBy { it.task }
     private val refresh = MutableLiveData<Boolean?>()
-    private val activeTask = MutableLiveData<QuickStartTask?>()
+    private val _activeTask = MutableLiveData<QuickStartTask?>()
     private val _onSnackbar = MutableLiveData<Event<SnackbarMessageHolder>>()
     val onSnackbar = _onSnackbar as LiveData<Event<SnackbarMessageHolder>>
+    val activeTask = _activeTask as LiveData<QuickStartTask?>
```
renanferrari, line 68: Good call on making this public! 👍

WordPress/src/main/java/org/wordpress/android/ui/mysite/QuickStartRepository.kt
```diff
@@ -74,14 +78,14 @@ class QuickStartRepository
     override fun buildSource(siteId: Int) = flow {
         emit(QuickStartUpdate())
         refresh.asFlow().map {
-            if (quickStartUtils.isQuickStartInProgress(siteId)) {
+            if (quickStartUtils.isQuickStartInProgress(siteId) || true) {
```
renanferrari: This seem to have been left here by mistake?

```suggestion
            if (quickStartUtils.isQuickStartInProgress(siteId)) {
```

WordPress/src/main/java/org/wordpress/android/ui/mysite/QuickStartRepository.kt
```diff
@@ -143,14 +147,18 @@ class QuickStartRepository
...
         eventBus.postSticky(QuickStartEvent(task))
     }
 
     fun clear() {
         job.cancel()
     }
 
+    fun shouldShowPublishPostQuickStartTask(): Boolean {
+        return activeTask.value == PUBLISH_POST
+    }
```
renanferrari: Since `activeTask` is now public, I wonder if we could remove this?

renanferrari (approved):
Thanks for the changes! I can confirm the app now works as expected and I really like the solution you found as well :)

I left a few comments below, but they are mostly nitpicks. Feel free to merge this after addressing them and the errors in the tests.

WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
planarvoid: I've previously used that method but the problem I had there is that it wasn't removing the correct focus point. For example when you clicked on "Publish post" and then on "Update site title", the site title focus point was removed because that method just finds any view in the parent with that ID and it was basically a race condition.

WordPress/src/main/java/org/wordpress/android/ui/mysite/QuickStartRepository.kt
planarvoid: yeah, good catch! I use it for testing :-)

renanferrari (approved):
LGTM! Feel free to merge once the lint errors are fixed.

PR #13931, opened 2021-02-01
Issue/13326 missing server credentials

malinajirka:
Parent issue #13326

This PR fixes an issue where snackbar was not being shown when the server returned a response containing all threats in "NOT_FIXED" status. It also adds "invokedByUser" flag so the snackbar is shown only when the action is invoked by the user - it's not shown, when it's eg invoked by vm's "init()" method.

To test:
Prerequisites
- Make sure Scan feature flag is enabled
1. Select a site which doesn't have sever credentials and has some threats
2. Select Scan
3. Make sure snackbar is not shown when you open the screen
4. Click on "Fix threat" button
5. Verify a snackbar saying that you need to contact support is shown

PR submission checklist:


ashiagr (approved):
Works as expected, thank you for the changes!

PR #13934, opened 2021-02-02
Implement "Follow other sites" QS task

### LINKED ISSUE #13846: Add functionality to "Follow other sites" QS task

renanferrari:
Fixes #13846

This task implements the "Follow other sites" Quick Start task, which is part of the "Grow your audience" category.

This task has two steps: the first one shows a snackbar message and focus point on top of the "Reader" button on the bottom navigation bar and the second one shows a snackbar message on the Reader screen. The task is completed as soon as the Reader screen is shown.

https://user-images.githubusercontent.com/830056/106625548-84d35500-6555-11eb-980f-be8503056b10.mp4

## To test

1. Go to the My Site screen and select a site where Quick Start is active.
1. Under the "Grow your audience" category, tap the "Follow other sites" task.
1. Notice that:
   1. A focus point appears over the Reader button on the bottom navigation bar.
   1. The correct snackbar message appears.
1. Tap on the "Follow other sites" task a second time.
1. Notice that there's still just one focus point over the Reader button.
1. Tap another task, like "Set your site title".
1. Notice that the focus point on top of the Reader button disappears.
1. Tap on the "Follow other sites" task again.
1. Tap on the Reader button.
1. On the Reader screen, notice that the correct snackbar message appears.
1. Go back to the My Site screen and scroll to the "Grow your audience" category.
1. Notice that the "Follow other sites" task is now completed.

PR submission checklist:


renanferrari:
<s>@planarvoid I've changed the base of this PR from `develop` to `feature/13845-add-post-publish-qs-task`, but since `develop` was merged here and not there, the diff is kinda messy right now. This should be fixed once we merge `develop` to that branch or the other way around. I'm keeping this as a draft to avoid any confusion, but the PR should be ready for review.</s>

Edit: I've rebased this on top of `feature/13845-add-post-publish-qs-task` and that should make it easier to review.

WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteViewModel.kt
```diff
@@ -143,6 +147,10 @@ class MySiteViewModel
     val onNavigation = merge(_onNavigation, siteStoriesHandler.onNavigation)
     val onMediaUpload = _onMediaUpload as LiveData<Event<MediaModel>>
     val onUploadedItem = siteIconUploadHandler.onUploadedItem
+    val onExternalQuickStartFocusPointVisibilityChange = quickStartRepository.activeTask
```
planarvoid: What do you think about moving this to the Main Activity view model and observing the active task directly there? I think it makes sense to do it there since the focus point is shown on that level and not in the MySite fragment/viewmodel.

renanferrari: Thanks for the suggestion, it makes total sense. I have now moved the logic to the `MainActivity` level. 

Since I have also moved the logic to add/remove a focus point view from the `ImprovedMySiteFragment` to the `MainActivity`, it seemed to make sense to reuse that for the `PUBLISH_POST` task as well (the logic on the view model remains the same). I've tested this with the flag both on and off and it seems to be working well. Let me know what you think.

renanferrari:
@planarvoid This is ready for another round of review 👍

planarvoid (approved):
I've tested the "Follow other sites" task, switching tabs while the task is active. I've also retested the "Publish post" task to make sure it works as expected. Everything is looking great 👍  thanks!

PR #13938, opened 2021-02-02
Issue/13326 get free estimate

malinajirka:
Parent issue #13326

This PR implements "get free estimate" action which is shown on ThreatDetail screen for non-fixable threats.

To test:
Prerequisites:
- enable scan feature flag
- select a site which either has a current non-fixable threat or it has a non-fixable threat in Scan History (eg. pressable-jetpack-daily-scan)
1. Open scan
2. Optional: (Open Scan History)
3. Open detail of a non-fixable ignored threat
4. Click on the "Get free estimate" button
5. Notice the app redirects you to an external webview

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/details/ThreatDetailsNavigationEvents.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/details/ThreatDetailsNavigationEvents.kt
@@ -17,4 +17,8 @@ sealed class ThreatDetailsNavigationEvents {
     data class ShowUpdatedScanStateWithMessage(@StringRes val messageRes: Int) : ThreatDetailsNavigationEvents()
 
     data class ShowUpdatedFixState(val threatId: Long) : ThreatDetailsNavigationEvents()
+
+    object ShowGetFreeEstimate : ThreatDetailsNavigationEvents() {
+        const val url = "https://codeable.io/partners/jetpack-scan/"
+    }
 }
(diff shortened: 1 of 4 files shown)


WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/details/ThreatDetailsNavigationEvents.kt
```diff
@@ -17,4 +17,8 @@ sealed class ThreatDetailsNavigationEvents {
     data class ShowUpdatedScanStateWithMessage(@StringRes val messageRes: Int) : ThreatDetailsNavigationEvents()
 
     data class ShowUpdatedFixState(val threatId: Long) : ThreatDetailsNavigationEvents()
+
+    object ShowGetFreeEstimate : ThreatDetailsNavigationEvents() {
+        const val url = "https://codeable.io/partners/jetpack-scan/"
```
zwarm, line 22: ❓ - Do we have an alternative for hardcoding or is this sufficient for the MVP? Post MVP 🤔  perhaps an endpoint similar to jetpack capabilities, but for upsells?

zwarm (approved):
@malinajirka - Looks good 🎈  - all works as expected, external browser launched correctly. I left one question/comment on hardcoding, but it won't hold up the merge. Just a thought for the future.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/details/ThreatDetailsNavigationEvents.kt
malinajirka: I asked the same question yesterday on Slack - tbh I'm not sure. I don't like hardcoding urls - especially 3rd party. The 3rd party can change the url and we'll never be able to remove it from this version of the app. 
I was thinking if we could for example point to our url and redirect from there. Downsides of this approach is that some browsers might complain and suppress the redirect to a different domain.

malinajirka: I'll merge this PR and update the code if we make a different decision on Slack.

PR #13940, opened 2021-02-02
Issue/13326 scan fixing in builder

ashiagr:
Parent #13326

This PR moves threat fixing state item building logic to the builder class. It fixes the issue when progress bar was not displayed when an ignored threat was fixed.

It also 
- updates header, description
- removes labels from the progress bar 
- adds loading spinner on threat list items when threats are fixing
to have a bit of similarity with the iOS app.

https://user-images.githubusercontent.com/1405144/106699637-4303e680-6609-11eb-8baf-a32cc766b85f.mov

**To test**

Prerequisite:

- A threat is added to the testing site. In case you run out of threats, re-install bad calendar plugin and re-run scan.
- Scan feature flag is enabled

1. Open Scan screen
2. Start scan
3. Click on "Fix threats" button
4. Notice that 
    - app displays "Fixing Threats" state and progress indicator is displayed on the threat items
    - on pressing back and returning to the screen while threats are fixing, "Fixing Threats" state is re-shown
    - as threats get fixed, they are removed from the list
    - once fixing threats list is empty, scan state is re-fetched and displayed

** Also test the flow when an ignored threat is fixed with steps from PR #13924.

**Notes**
- Ignore styling

PR submission checklist:


WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/ScanViewModelTest.kt
```diff
@@ -517,22 +518,23 @@ class ScanViewModelTest : BaseUnitTest() {
...
-
-            fetchFixThreatsStatus(observers)
-
-            val indeterminateProgressBars = (observers.uiStates.last() as ContentUiState).items
-                .filterIsInstance<ProgressState>()
-                .filter { it.isIndeterminate && it.isVisible }
-
-            assertThat(indeterminateProgressBars.isNotEmpty()).isTrue
-        }
+    // TODO ashiagr fix test
```
malinajirka: Is this intentionally commented out or were you planning on getting back to it in this PR?

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/builders/ScanStateListItemsBuilderTest.kt
```diff
@@ -60,8 +68,8 @@ class ScanStateListItemsBuilderTest : BaseUnitTest() {
             htmlMessageUtils,
             resourceProvider,
             threatItemBuilder,
-            uiHelpers,
-            contextProvider
+            threatDetailsListItemsBuilder,
+            scanStore
         )
 //        whenever(htmlMessageUtils.getHtmlMessageFromStringFormatResId(anyInt(), any())).thenReturn(SpannedString(""))
```
malinajirka: ❓ Could we remove this commented out code as well? We can always find it in the git history if needed. Wdyt?

malinajirka (approved):
Looks great @ashiagr, thanks!

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/ScanViewModelTest.kt
ashiagr: Oh yes, I wanted to fix it here. Thanks for pointing it.

I've removed vm test cases related to updating button and progress bar on threats fixing in 7e1742e as threats fixing content building is now done in the builder class.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/builders/ScanStateListItemsBuilderTest.kt
ashiagr: Yes, removed in d61a3e8.

ashiagr:
Thanks for the review @malinajirka, ready for another look.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/builders/ScanStateListItemsBuilder.kt
```diff
@@ -242,16 +246,28 @@ class ScanStateListItemsBuilder @Inject constructor(
...
+    ): DescriptionState {
+        val descriptionText = htmlMessageUtils
+            .getHtmlMessageFromStringFormatResId(
+                R.string.scan_idle_threats_found_description,
+                "<b>$threatsCount</b>",
+                "<b>${site.name ?: resourceProvider.getString(R.string.scan_this_site)}</b>"
+            )
+
+        val clickableText = resourceProvider.getString(R.string.scan_here_to_help)
+        val startIndex = descriptionText.indexOf(clickableText)
```
malinajirka: ⚠️ I'm worried that this might not work as expected. No-one can guarantee that translations to other languages will have the exact same translation for `scan_here_to_help` as what is used in `scan_idle_threats_found_description` -> so I think this might return -1. Or am I missing something?


What if we used a string argument instead of relying on the translation. Wdyt?

```
<string name="scan_idle_threats_found_description">The scan found %1$s potential threats on %2$s. Please review them below and take action. We\'re %3$s if you need us.</string>
<string name="scan_here_to_help">here to help</string>
```

Wdyt?

malinajirka (changes requested):
Thanks for the improvements @ashiagr! I found one potential issue, let me know what you think ;).

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/builders/ScanStateListItemsBuilder.kt
ashiagr: Oops, yes you're right....will fix it shortly.

ashiagr: Fixed in 6dbdce9.

malinajirka (approved):
LGTM, thanks!

PR #13942, opened 2021-02-02
Show post/page updated notification if this is not a first time publish

### LINKED ISSUE #3018: Updating a Published Page/Post notifies "Page Published" instead of updated
**Describe the bug**
Updating a Published Page/Post notifies "Page Published" instead of "Page Updated."

**To Reproduce**
1. Go to a published Page / Post
2. Make a change
3. Save the change
4. The toast says, “Page Published”; however, “Page Updated” is more accurate to the action taken.

**Expected behavior**
Update the toast to say, “Page Updated.” 

**Screenshots**
<kbd><a href="https://user-images.githubusercontent.com/3384451/104965324-81b06480-59ac-11eb-92e2-aeafd4b11445.gif"></a></kbd>

**Additional context**
This was called out in one of our recent Empathy Challenge.

antonis:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/3018

## To test:

1. Go to a published Page / Post
2. Make a change
3. Save the change
4. Verify that the toast says “Page Updated”

Note: This implementation makes the assumption that the post/page is new if it was created in the last minute

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/posts/PostUtils.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/posts/PostUtils.java
@@ -344,14 +344,25 @@ public class PostUtils {
     }
 
     public static boolean isPublishDateInThePast(String dateCreated, Date currentDate) {
-        Date pubDate = DateTimeUtils.dateFromIso8601(dateCreated);
+        // just use half an hour before now as a threshold to make sure this is backdated, to avoid false positives
+        return isPublishDateInThePast(dateCreated, currentDate, 30);
+    }
 
+    /**
+     * Checks if the post publish date is in the past
+     *
+     * @param dateCreated date created
+     * @param currentDate current date
+     * @param threshold   threshold in minutes
+     * @return true if the post was published in the past
+     */
+    public static boolean isPublishDateInThePast(String dateCreated, Date currentDate, int threshold) {
+        Date pubDate = DateTimeUtils.dateFromIso8601(dateCreated);
         Calendar cal = Calendar.getInstance();
         cal.setTime(currentDate);
-        // just use half an hour before now as a threshold to make sure this is backdated, to avoid false positives
-        cal.add(Calendar.MINUTE, -30);
-        Date halfHourBack = cal.getTime();
-        return pubDate != null && pubDate.before(halfHourBack);
+        cal.add(Calendar.MINUTE, -threshold);
+        Date timeThreshold = cal.getTime();
+        return pubDate != null && pubDate.before(timeThreshold);
     }
 
     // Only drafts should have the option to publish immediately to avoid user confusion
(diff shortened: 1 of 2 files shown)


WordPress/src/main/java/org/wordpress/android/ui/posts/PostUtils.java
```diff
@@ -344,14 +344,25 @@ public static boolean isPublishDateInTheFuture(String dateCreated, Date currentD
     }
 
     public static boolean isPublishDateInThePast(String dateCreated, Date currentDate) {
-        Date pubDate = DateTimeUtils.dateFromIso8601(dateCreated);
+        // just use half an hour before now as a threshold to make sure this is backdated, to avoid false positives
```
enejb, line 347: Instead of using a threshold can we not used the previous post status. So when the post changes from "draft" => "published"?

antonis: Hey @enejb 👋 
Thank you for your feedback on this 🙇 

> Instead of using a threshold can we not used the previous post status. So when the post changes from "draft" => "published"?

The correct approach is the one you describe. My solution is a workaround based on the [`isPublishDateInThePast` utility method](https://github.com/wordpress-mobile/WordPress-Android/blob/47a600c17a4a00403c56da43061e91b84684505d/WordPress/src/main/java/org/wordpress/android/ui/posts/PostUtils.java#L346) that is [used in one more place](https://github.com/wordpress-mobile/WordPress-Android/blob/a9eeed79f6d85fd3a3a55157ff9b2932248c00cc/WordPress/src/main/java/org/wordpress/android/ui/posts/PostSettingsUtils.kt#L27).

At this point I've found two ways to get the previous post status and do the correct check to determine if this post is published or just updated:
* Fetch the post revisions. This would have the overhead of a network call just to change the wording from "published" to "updated".
* Track down all the cases where the post upload is triggered and pass the information of the previous status. This solution looks good to me but involves changes in a lot of places in two repositories:
  * The `onPostUploadedSnackbarHandler` method is called by the [equivalent wrapper](https://github.com/wordpress-mobile/WordPress-Android/blob/bcb2cb26824adc83f3eda82baf61110e4e83eb7e/WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadUtilsWrapper.kt#L50) method.
  * The later is [triggered in 9 occasions](https://github.com/wordpress-mobile/WordPress-Android/search?q=onPostUploadedSnackbarHandler&type=code) and all seem to be a result of a [`OnPostUploaded` event](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/e2b21ad10c70ad4f0147e48eb6704585ba8858a7/fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java#L310).
  * We can keep the store the old post or just it's status [when the post is updated](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/e2b21ad10c70ad4f0147e48eb6704585ba8858a7/fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java#L1015) and add it in the payload of the `OnPostUploaded` event.
  
I will investigate the second solution further and if I don't see any implications I will proceed with this.

antonis:
Hey @enejb, @mkevins 👋 
I'm closing this in favor of a more proper approach at https://github.com/wordpress-mobile/WordPress-Android/pull/13970

PR #13943, opened 2021-02-02
Implement EDIT_HOMEPAGE QS action and tests

### LINKED ISSUE #13877: Add functionality to "Edit your homepage" QS task
When clicked, the QS task should show a focus point on the Pages quick action button. On the Site Pages screen, a Snackbar should indicate that the user should click on the "Homepage" item. The task should be completed after clicking the item.

Step 1 | Step 2
------------ | -------------
homepage1 | homepage2

planarvoid:
Fixes #13877 

This PR adds functionality to the edit homepage dynamic card. It now scrolls the app to the Quick actions, shows a focus point and on click triggers the previously implemented logic (with the sticky event). The only change to the previous logic is that the `complete` action is now passed to the QuickStartRepository from the pages screen (if the flag is enabled). Don't mind the editor crash in the video. 

To test:
- Choose a site with a "Homepage" selected
- Click on the "Edit homepage" card
- Notice the app scrolls to the quick actions card
- Notice the focus point on the "Pages" action
- Click on the "Pages" action
- Notice the focus point on the Homepage
- Click on the Homepage
- Go back to My Site
- Notice the "Edit Homepage" action is now completed
- Switch to a site without a Homepage
- Notice the "Edit Homepage" task is autocompleted

edit_homepage_task

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteViewModel.kt
```diff
@@ -335,9 +340,12 @@ class MySiteViewModel
         _onNavigation.value = Event(getStatsNavigationActionForSite(site))
     }
 
-    private fun quickActionPagesClick() {
+    private fun quickActionPagesClick(activeTask: QuickStartTask?) {
         val site = requireNotNull(selectedSiteRepository.getSelectedSite())
         analyticsTrackerWrapper.track(QUICK_ACTION_PAGES_TAPPED)
+        if (activeTask == EDIT_HOMEPAGE) {
+            eventBusWrapper.postSticky(QuickStartEvent(activeTask))
+        }
```
renanferrari: Maybe we could use `QuickStartRepository::requestNextStepOfTask` here instead? Then we wouldn't even need to pass the current active task as a parameter to `ListItemInteraction::create`.

renanferrari (commented):
Thanks for the changes! I've tested the app and everything seems to be working as expected. Code looks good overall, I just left a suggestion below.

WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteViewModel.kt
planarvoid: This method calls `activeTask.value = null` and I'm wondering if that's what we expect in this case since we complete the task only if the user clicks on the Homepage, not on the "Pages" item. What do you think? The difference is that click on Pages and back would not remove the focus point. I guess removing the focus point is OK though since the point is to show the user the menu item and by clicking it we're sure they've seen it. And they can also restart the task by clicking it again

renanferrari (approved):

PR #13944, opened 2021-02-02
Issue/backup restore reduce fetch delay

zwarm:
Parent #13328 & #13329

Merge Instructions:
1. Make sure [FluxC PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1860) is merged first
2. Update FluxC tag in gradle file
3. Remove "Not ready for merge label"

To test:
1. Launch app
2. Turn on the Restore feature flag
3. Navigation to Activity Log
4. Tap on the more menu on a row 
5. Select the "Restore to this point' option
6. Follow the restore process to end
7. Note that the process runs as expected

------------
1. Launch app
2. Turn on BackupDownload feature flag
3. Navigation to Activity Log
4. Tap on the more menu on a row 
5. Select the "Download backup' option
6. Follow the download process to end
7. Note that the process runs as expected

PR submission checklist:


ParaskP7 (changes requested):
👋 @zwarm !

Warning (⚠️): What I just noticed while running through the app and flows is that the `GetRestoreStatusUseCase` is running indefinitely when there is no restore in progress and the activity log screen launches. This is due to the fact that the `activityLogStore.getRewindStatusForSite(site)?.rewind` returns as `null` within the `GetRestoreStatusUseCase` and since it is `null` it never enters the next `if` statement, then the loop starts over resulting in the same output.

I am not sure when this was introduced or whether we missed that from the very beginning, but I am confident that if we add the `emit(Empty)` logic on this use case, just like we did on the `getBackupDownloadStatus(...)`, everything will work fine. I actually tried it myself and it indeed works okay. Wdyt?

GetRestoreStatusUseCase
```
val rewind = activityLogStore.getRewindStatusForSite(site)?.rewind
if (rewind == null) {
    emit(Empty)
    return@flow
}
if (restoreId == null || rewind.restoreId == restoreId) { ... }
```

RestoreViewModel
```
private fun handleQueryStatus(restoreStatus: RestoreRequestState) {
    when (restoreStatus) {
        ...
        is Empty -> transitionToError(RestoreErrorTypes.RemoteRequestFailure)
        else -> ...
        }
    }
```

zwarm:
>  I am confident that if we add the emit(Empty) logic on this use case

@ParaskP7 - 🤦  Wow, this was a great catch. I addressed as you suggested and added a new test in the `GetRestoreUseCaseTest` to make sure it emit properly when a null status model is returned from the get call. 🙇

ParaskP7:
> @ParaskP7 - 🤦 Wow, this was a great catch. I addressed as you suggested and added a new test in the GetRestoreUseCaseTest to make sure it emit properly when a null status model is returned from the get call. 🙇

👋 @zwarm !

Many thanks 🙏 I am looking at the changes as we speak...

WordPress/src/test/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCaseTest.kt
```diff
@@ -240,6 +241,14 @@ class GetRestoreStatusUseCaseTest {
                 )
             }
 
+    @Test
+    fun `given get status model is null, when restore status triggers, then return empty`() = test {
```
ParaskP7: Minor (🔍): Consider adding a test for the `without restore id` case. All the tests above are duplicated to `with restore id` and `without restore id`. Maybe we should keep that pattern for this test as well.

PS: After the so many update we did on this use case maybe the `with` and `without` restore id is no longer necessary. But, before deleting and consolidating these and the `Backup Download` related test, I would keep consistency first.
PSS: Feel free to ignore this comment.

ParaskP7 (approved):
👋 @zwarm  !

Just reviewed the changed and they look good to me, many thanks! 🙏 

I have left a small test comment, feel free to ignore it and merge this to `develop`, it is now approved.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCaseTest.kt
zwarm: > PS: After the so many update we did on this use case maybe the `with` and `without` restore id is no longer necessary. 

I didn't think it was necessary, so I started on the road to eliminating it; however for this PR I will stay consistent. I added in the null restoreId test. :)

ParaskP7: Thank you @zwarm ! 🌟 

Yes indeed, I think it is no longer necessary and we can do an cleanup with our `Restore` and `Backup Download` use case tests next week when we will get the chance to do it.

PS: LGTM, feel free to merge this PR when it is ready.

PR #13955, opened 2021-02-03
Better handling of empty state in My Site fragment

planarvoid:
There are two issues that cause the My Site screen to sometimes show the "No Sites" state. This PR fixes both of them. 

The first thing is that we don't need to emit empty state when site changes. The other one is that when a Site ID is emitted but the SiteModel is not yet there, we don't want to emit an empty state (because it's logically incorrect - there is a site). In this PR this state is filtered out now. 

To test:
- Go to Plans
- Go back
- Notice you can't see the "No Site" state
- You might need to turn on the "Don't keep activites" developer options to make sure the main activity is destroyed when you leave it

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteStateProvider.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteStateProvider.kt
@@ -37,7 +37,6 @@ class MySiteStateProvider(
             mySiteSources.filterIsInstance(SiteIndependentSource::class.java)
                     .map { source -> source.buildSource().distinctUntilChanged().asLiveData(bgDispatcher) }
         }
-        result.value = MySiteUiState()
         for (newSource in currentSources) {
             result.addSource(newSource) { partialState ->
                 if (partialState != null) {
--- a/WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteStateProvider.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteStateProvider.kt
@@ -11,6 +11,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.distinctUntilChanged
 import org.wordpress.android.modules.BG_THREAD
 import org.wordpress.android.ui.mysite.MySiteSource.SiteIndependentSource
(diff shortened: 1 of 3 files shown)


WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteStateProvider.kt
```diff
@@ -30,22 +31,26 @@ class MySiteStateProvider(
     }
 
     val state: LiveData<MySiteUiState> = selectedSiteRepository.siteSelected.switchMap { siteId ->
-        val result = MediatorLiveData<MySiteUiState>()
+        data class SiteIdToState(val siteId: Int?, val state: MySiteUiState = MySiteUiState()) {
+            fun update(partialState: PartialState): SiteIdToState {
+                return this.copy(state = state.update(partialState))
+            }
+        }
```
renanferrari: This declaration's location seems a bit odd for me. Could we move it to the outside of the function?

WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteStateProvider.kt
```diff
@@ -30,22 +31,26 @@ class MySiteStateProvider(
...
         for (newSource in currentSources) {
             result.addSource(newSource) { partialState ->
                 if (partialState != null) {
-                    result.value = (result.value ?: MySiteUiState()).update(partialState)
+                    result.value = (result.value ?: SiteIdToState(siteId)).update(partialState)
                 }
             }
         }
-        result
+        result.filter { it.siteId == null || it.state.site != null }.map { it.state }
```
renanferrari: We might want to document this logic as it probably won't be obvious for everyone that reads it. What do you think?

renanferrari (commented):
Thanks for working on this, the app feels much better now and I really like the solution you found! That said I left two minor comments below.

WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteStateProvider.kt
planarvoid: good point 👍

planarvoid:
Thanks for the review @renanferrari , I've implemented your suggestions 👍

renanferrari (approved):

PR #13956, opened 2021-02-03
Jetpack Section: Backup Screen Tracking Events

ParaskP7:
Parent #13629

This PR adds five new track events specific to the new `Backup` screen and enhances an existing `Activity Log Detail` event with a `source` property so that it differentiates between the `activity_log` or `backup` entry points.

To test:
- Launch app and go to `My Site` tab.
- Got to: `My Site` -> `Toolbar Avatar` -> `App Settings` -> `Privacy settings`
- Switch on `Collect information` to enable tracking on your device.
- Go to `My Site` -> `Toolbar Avatar` -> `App Settings` -> `Test feature configuration`.
- Enable `BackupsFeatureConfig`, scroll down and click the `RESTART THE APP` button. **DO NOT** confuse this feature flag with the `BackupFeatureConfig` (singular). Make sure to enable the right feature flag.
- Open `Android Logcat`, filter the logcat by `tracked` and start noticing the `🔵 Tracked` events.
- Go back to `My Site` and click on the `Backup` menu item. Notice the `jetpack_backup_list_opened` track event is triggered.
- Find a rewindable item and click on it. Notice the `activity_log_detail_opened` track event is triggered with `backup` as its `source`.
- Click the `RESTORE` button and then the `RESTORE SITE` dialog button. Notice that you will be redirected back to the `Backup` screen and that the `jetpack_backup_rewind_started` track event is triggered.
- Click the `Date Range` chip. Notice the `jetpack_backup_filterbar_range_button_tapped` track event is triggered.
- Select a date range and click `SAVE`. Notice that you will be redirected back to the `Backup` screen and that the `jetpack_backup_filterbar_select_range` track event is triggered.
- Click the `x` inner button within the chip to clear the filter. Notice the `jetpack_backup_filterbar_reset_range` track event is triggered.

PR submission checklist:


ParaskP7:
👋 @leandroalonso !

I added you as a reviewer as well just as an FYI to you so that you are aware of this since you were working on the `Backup` screen on iOS.

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
```diff
@@ -614,16 +618,26 @@ public static void viewBackupList(Activity activity, SiteModel site) {
...
-        AnalyticsUtils.trackWithSiteDetails(AnalyticsTracker.Stat.ACTIVITY_LOG_LIST_OPENED, site);
+        AnalyticsUtils.trackWithSiteDetails(AnalyticsTracker.Stat.JETPACK_BACKUP_LIST_OPENED, site);
         Intent intent = new Intent(activity, ActivityLogListActivity.class);
         intent.putExtra(WordPress.SITE, site);
         intent.putExtra(ACTIVITY_LOG_REWINDABLE_ONLY_KEY, true);
         activity.startActivity(intent);
     }
 
-    public static void viewActivityLogDetailForResult(Activity activity, SiteModel site, String activityId) {
+    public static void viewActivityLogDetailForResult(
```
malinajirka, line 628: 🔍 Nitpick: We usually don't use "parameter per line" formatting in Java classes. We might want to update this to 
```
public static void viewActivityLogDetailForResult(Activity activity, SiteModel site, String activityId,
                                                      boolean rewindableOnly) {
```

Wdyt?

malinajirka (commented):
Thanks @ParaskP7! I'm having some doubts about some of the events we came up with. Let's discuss this during the Android Sync meeting today.

I'm not approving the PR yet, I'll approve it when we make a final decision on how to proceed.


> Click the RESTORE button and then the RESTORE SITE dialog button. Notice that you will be redirected back to the Backup screen and that the jetpack_backup_rewind_started track event is triggered.

Should this trigger also `jetpack_restore_opened` and when you confirm the dialog -> `jetpack_restore_confirmed`?

Related question: should `jetpack_restore_opened` event be fired when I click on "MySite -> Backup -> more button on one of the items -> "Restore to this point" ? 


> jetpack_backup_rewind_started

I'm not sure about this event. I tried to keep the separation between backup/activityLog only on the level of the entry points - -> so we don't need to pass backup/activityLog flag into all the flows and keep having `ifs` all over the place. I understand that `activity_log_rewind_started` is an existing event, but I'm wondering if we should probably make an exception here and stop recording it (or record the same event even for backup).

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
ParaskP7: Oh cool, thanks for letting me know. I saw that this was not being done on the other methods as well and I was wondering why. Will update to your suggestion. 🙏

ParaskP7:
Many thanks for the thorough review @malinajirka ! 🌟 

> Thanks @ParaskP7! I'm having some doubts about some of the events we came up with. Let's discuss this during the Android Sync meeting today.

Definitely! 👌 

> I'm not approving the PR yet, I'll approve it when we make a final decision on how to proceed.

Makes sense. 👍 

> Should this trigger also `jetpack_restore_opened` and when you confirm the dialog -> `jetpack_restore_confirmed`?

No, this actual event is a bit tricky and it is not a "new" one per se. This was located within the `RewindStatusService.rewind(...)` function and was being triggered then. If you traverse back and check the history on the `ActivityLogViewModel` class you will notice that this was triggered during `onRewindConfirmed(...)`, which is now renamed to `onRestoreConfirmed(...)`. So, the new `Backup` screen related event is actually replicating the existing event for the `Activity Log` screen, thus me keeping the `rewind` naming convention, like `jetpack_backup_rewind_started`, since the other one is named as `activity_log_rewind_started`. Does this make sense?

> Related question: should `jetpack_restore_opened` event be fired when I click on `MySite` -> `Backup` -> `more button` on one of the items -> `Restore to this point`?

This track events are related to the `Restore Flow Screen` and thus @zwarm will handle that (or I'll take this over from Annmarie, to be discussed). But, definitely I didn't want them as part of this PR, trying to keep them separate.

> I'm not sure about this event. I tried to keep the separation between backup/activityLog only on the level of the entry points - -> so we don't need to pass backup/activityLog flag into all the flows and keep having ifs all over the place. I understand that `activity_log_rewind_started` is an existing event, but I'm wondering if we should probably make an exception here and stop recording it (or record the same event even for backup).

I touched on that a bit above, now realising that you kind of have the picture. This is up to us to decide, I am comfortable in us stopping recording it, but at the same time why not just keep it for a while to get some results, figure out whether it is valuable and then drop it afterwards, especially since we already have data for the `activity_log_rewind_started` and can start comparing that to the new `jetpack_backup_rewind_started` track event. Wdyt?

PS: Currently, we cannot record the same event for backup since within the `ActivityLogDetail` screen we only have the `RESTORE` flow. This event is purely connecting the `ActivityLogDetail` with that restore trigger.

malinajirka (approved):
Thanks for the clarification @ParaskP7!

As discussed `activity_log_rewind_started` is currently tracked only in the legacy flows. We need to make decision if it's find to stop tracking this event in the new flows or if we need to retain it.

All looks good to me ;)!

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
malinajirka: Ohh, you are right - there is a lot of methods which use the formatting you used. In that case, don't worry about it ;).

ParaskP7:
> Thanks for the clarification @ParaskP7!

🙏 

> As discussed activity_log_rewind_started is currently tracked only in the legacy flows. We need to make decision if it's find to stop tracking this event in the new flows or if we need to retain it.

👌 

> All looks good to me ;)!

🙏

PR #13957, opened 2021-02-03
Enable/disable submit button on checkbox click

zwarm:
Parent #13328 & 13329

This PR addresses the following comment
"On unchecking all checkbox items on the "Details" screen, we might want to disable action button."
By ..
1. Disabling the details action button when no items are selected
2. Enabling the action button when at least 1 checkbox is selected

This behavior now mimics Calypso.

To Test
- Launch the app
- Ensure BackupDownload and Restore feature flags are enabled
------------------
- Navigate to ActivityLog
- Tap More menu on an item
- Select "Download backup" option
- On the details view, uncheck all the checkboxes
- The "Create downloadable file" button should toggle to disabled
- Check a checkbox
- The "Create downloadable file" button should toggle to enabled
------------------
- Navigate to ActivityLog
- Tap More menu on an item
- Select "restore to this point" option
- On the details view, uncheck all the checkboxes
- The "Restore Site" button should toggle to disabled
- Check a checkbox
- The "Restore Site" button should toggle to enabled
------------------

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
@@ -338,9 +338,9 @@ class BackupDownloadViewModel @Inject constructor(
     }
 
     private fun onCheckboxItemClicked(itemType: JetpackAvailableItemType) {
-        _uiState.value?.let {
+        (_uiState.value as? DetailsState)?.let {
             val updatedItems = stateListItemBuilder.updateCheckboxes(it, itemType)
-            _uiState.postValue((it as DetailsState).copy(items = updatedItems))
+            _uiState.postValue(it.copy(items = updatedItems))
         }
     }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
@@ -42,7 +42,6 @@ import org.wordpress.android.ui.jetpack.backup.download.usecases.PostBackupDownl
 import org.wordpress.android.ui.jetpack.common.JetpackListItemState
 import org.wordpress.android.ui.jetpack.common.JetpackListItemState.CheckboxState
 import org.wordpress.android.ui.jetpack.common.ViewType
-import org.wordpress.android.ui.jetpack.common.ViewType.CHECKBOX
 import org.wordpress.android.ui.jetpack.common.providers.JetpackAvailableItemsProvider
 import org.wordpress.android.ui.jetpack.common.providers.JetpackAvailableItemsProvider.JetpackAvailableItemType
 import org.wordpress.android.ui.jetpack.common.providers.JetpackAvailableItemsProvider.JetpackAvailableItemType.CONTENTS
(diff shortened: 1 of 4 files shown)


WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
```diff
@@ -339,20 +338,9 @@ class BackupDownloadViewModel @Inject constructor(
...
-                        contentState.copy(checked = !contentState.checked)
-                    } else {
-                        contentState
-                    }
-                } else {
-                    contentState
-                }
-            }
-            _uiState.postValue(details.copy(items = updatedList))
+        _uiState.value?.let {
```
malinajirka: 💡 Idea: We might want to consider using `(_uiState.value as? DetailsState)?.let {...}` instead of force casting the state in `_uiState.postValue((it as DetailsState).copy(items = updatedItems))`. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/builders/BackupDownloadStateListItemBuilder.kt
```diff
@@ -229,4 +232,44 @@ class BackupDownloadStateListItemBuilder @Inject constructor(
             label = UiStringRes(labelRes),
             itemBottomMarginResId = itemBottomMarginResId
     )
+
+    fun updateCheckboxes(
+        uiState: BackupDownloadUiState,
+        itemType: JetpackAvailableItemType
+    ): List<JetpackListItemState> {
+        var checkedCount = 0
```
malinajirka: 💡 This code took me a while to understand. We might want to consider separating the "change state" logic and "count" logic. Perhaps something like this might work:
```
        val updateListItems = uiState.items.map { state ->
                if (state is CheckboxState && state.availableItemType == itemType) {
                    state.copy(checked = !state.checked)
                } else {
                    state
                }
        }
        val atLeastOneChecked = updateListItems.filterIsInstance<CheckboxState>().find { it.checked } != null
        return updateDetailsActionButtonState(updateListItems, enableActionButton = atLeastOneChecked)
```

Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/builders/BackupDownloadStateListItemBuilder.kt
```diff
@@ -229,4 +232,44 @@ class BackupDownloadStateListItemBuilder @Inject constructor(
...
+        }
+        return updateDetailsActionButtonState(updateCheckboxes, checkedCount > 0)
+    }
+
+    private fun updateDetailsActionButtonState(
+        details: List<JetpackListItemState>,
+        enableActionButton: Boolean
+    ): List<JetpackListItemState> {
+        return details.map { state ->
+            if (state.type == PRIMARY_ACTION_BUTTON) {
```
malinajirka: ❓ Is this equivalent to `if (state is ActionButtonState && !state.isSecondary)` ? If it is, I think it might be safer to use this approach, wdyt? ( or could we perhaps drop the "isSecondary" check completely?)

malinajirka (approved):
Thanks @zwarm! I've tested the app and it works as expected. I've left some suggestions, but they are not blocking this PR.

P.S. The changes are not unit tested - I know that the whole class is not unit tested yet, but adding at least tests for the introduced logic might be worth it. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
zwarm: Changed in 14d3a34

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/builders/BackupDownloadStateListItemBuilder.kt
zwarm: Nice upgrade. In commit 14d3a34

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/builders/BackupDownloadStateListItemBuilder.kt
zwarm: `if (state is ActionButtonState) ` is sufficient ... in this case I can be opinionated because there is only 1 button on the details view. Updated in 14d3a34

zwarm (commented):
> P.S. The changes are not unit tested - I know that the whole class is not unit tested yet, but adding at least tests for the introduced logic might be worth it. Wdyt?

Yes, unit testing is going to be my focus today ... with the builders at the top of the list. 
@malinajirka Thanks for the review. :)

PR #13958, opened 2021-02-04
Force text color on reader post.

### LINKED ISSUE #13841: In Dark Mode Text Color Does Not Change Automatically (to a light one - presumably white)
### Expected behavior
In dark mode the "Reader" should change the dark text color to "white" or some light color (perhaps light grey etc.) automatically.

### Actual behavior
On Android in dark mode the text color of posts stays the same as they are/were in light mode. Yellow is yellow, green is green, black is black etc. 

1. Publish a post with different color text in it. 
2. Use Android phone in dark mode to access that post in "Reader"
3. Success!? ... Shows the same color text as in light mode. 

Phone: OnePlus5 
OS: Oxygen OS 10.0.1 (based on Android 10)
WP App version: 14.7

Images on Android: 

Screenshot_20210121-204005 Screenshot_20210121-204011

Same Images on iOS: 

IMG_1021 IMG_1020

khaykov:
Fixes #13841 

This PR forces text inside blocks (`<p>` tags) and elements with `.has-inline-color` class to use a preferred text color. 
This fixes issued of text color readability in light/dark modes and aligns us with iOS.

I'm not aware of other ways we change text color in our editors, so please let me know if you think I'm missing something. 

Before-After comparisons:

[Image from Gyazo](https://gyazo.com/d1935d01e5b960084c1ee67e675799b9)

[Image from Gyazo](https://gyazo.com/d69148ca76f6b0535d73c22e7614540c)

PS. Red is colored using GB block level setting, and it was not showing up in Reader anyway.

To test:
- Open post with colored text in Reader.
- Confirm that the colors are not shown.

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -5,6 +5,7 @@
 * [*] Adds tablet mode in site and post/page previews [https://github.com/wordpress-mobile/WordPress-Android/pull/13787]
 * [*] Reader: show post menu for posts in Blog Preview and Search results [https://github.com/wordpress-mobile/WordPress-Android/pull/13897]
 * [**] Site Creation: Enables dot blog subdomains for each site design. [https://github.com/wordpress-mobile/WordPress-Android/pull/13917]
+* [*] Reader: render all colored text using a single color. [https://github.com/wordpress-mobile/WordPress-Android/pull/13958]
 
 16.6
 -----
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostRenderer.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostRenderer.java
@@ -344,7 +344,8 @@ public class ReaderPostRenderer {
               // force font style
         sbHtml.append(" body.reader-full-post__story-content { font-family: 'Noto Serif', serif; font-weight: 400; ")
               .append("font-size: 16px; margin: 0px; padding: 0px; }")
-              .append(" p, div, li { line-height: 1.6em; font-size: 100%; }")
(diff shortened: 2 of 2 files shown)


malinajirka:
Thanks @khaykov! I'll look into this on Monday. We are wrapping up a project so I won't have time today.

malinajirka:
Thanks @khaykov! I've tested the app and it works as described. However, we might want to consider placing these styles into `https://wordpress.com/calypso/reader-mobile.css` instead of hardcoding them into the app. This approach would bring us two benefits. Firstly, we'd fix the issue even in the older versions of the app. Secondly, we could be sure that we are aligned with iOS since they are using the same stylesheet. Wdyt? (@leandroalonso your opinion on this issue would be very welcome)

khaykov:
Hey @malinajirka @leandroalonso I didn't know about `reader-mobile.css`! Sounds really handy, and I'm up for adding this fix there, but I'm not sure how this will affect iOS - this issues does not seems to be present there, and I'm not sure how this will influence post rendering (I guess this is a question to @leandroalonso ?).

If there are not issues with this fix, how do I go about adding it to `reader-mobile.css` ?

leandroalonso:
Hi @khaykov!

> I'm not sure how this will affect iOS - this issues does not seems to be present there, and I'm not sure how this will influence post rendering

The main difference why this doesn't happen in iOS is because we strip the `style` attributes, thus this color doesn't appear there.

Adding the text color CSS wouldn't be an issue if it's like this:

```
p, div, li, .has-inline-color { color: var(--color-text) !important; }
```

(the font-size and line-height are Android specific)

I can help with testing this change in iOS just to make sure!

khaykov:
I opened a [PR](https://github.com/Automattic/wp-calypso/pull/50291) in calypso to address this. Will close this one after/if fix on calypso side is merged.

PR #13962, opened 2021-02-04
Issue/13326 hide scan on wpcom sites

malinajirka:
Parent issue #13326

This PR hides Scan menu item on .com sites (even atomic). The scan feature is supported on sites with Business plan, however, it's auto-managed so the Scan feature shouldn't be visible to the users on such sites.

To test:
2. Select a site with Business plan
3. Make sure Scan item is not visible
-------------------
1. Make sure Scan feature flag and MySiteImprovements feature flag are enabled
2. Select a site with Business plan
3. Make sure Scan item is not visible

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
@@ -154,6 +154,7 @@ import org.wordpress.android.util.QuickStartUtils.Companion.isQuickStartInProgre
 import org.wordpress.android.util.QuickStartUtils.Companion.removeQuickStartFocusPoint
 import org.wordpress.android.util.QuickStartUtilsWrapper
 import org.wordpress.android.util.SiteUtils
+import org.wordpress.android.util.SiteUtilsWrapper
 import org.wordpress.android.util.ToastUtils
 import org.wordpress.android.util.ToastUtils.Duration.SHORT
 import org.wordpress.android.util.WPMediaUtils
@@ -211,6 +212,7 @@ class MySiteFragment : Fragment(),
     @Inject lateinit var uiHelpers: UiHelpers
     @Inject lateinit var themeBrowserUtils: ThemeBrowserUtils
     @Inject lateinit var jetpackCapabilitiesUseCase: JetpackCapabilitiesUseCase
+    @Inject lateinit var siteUtilsWrapper: SiteUtilsWrapper
     @Inject lateinit var quickStartUtilsWrapper: QuickStartUtilsWrapper
     @Inject @Named(UI_THREAD) lateinit var uiDispatcher: CoroutineDispatcher
     @Inject @Named(BG_THREAD) lateinit var bgDispatcher: CoroutineDispatcher
@@ -281,12 +283,11 @@ class MySiteFragment : Fragment(),
             uiScope.launch {
                 val itemsVisibility = jetpackCapabilitiesUseCase.getJetpackPurchasedProducts(site.siteId)
                 row_scan.setVisible(
-                        scanScreenFeatureConfig.isEnabled() &&
-                                itemsVisibility.scan &&
-                                !site.isWPCom &&
-                                !site.isWPComAtomic
+                        siteUtilsWrapper.isScanEnabled(scanScreenFeatureConfig.isEnabled(), itemsVisibility.scan, site)
                 )
-                row_backup.setVisible(backupScreenFeatureConfig.isEnabled() && itemsVisibility.backup)
+                row_backup.setVisible(
+                        siteUtilsWrapper.isBackupEnabled(backupScreenFeatureConfig.isEnabled(), itemsVisibility.backup)
+                )
             }
         }
     }
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
@@ -281,8 +281,10 @@ class MySiteFragment : Fragment(),
             uiScope.launch {
                 val itemsVisibility = jetpackCapabilitiesUseCase.getJetpackPurchasedProducts(site.siteId)
                 row_scan.setVisible(
-                        scanScreenFeatureConfig.isEnabled() && itemsVisibility.scan &&
-                                !site.isWPCom && !site.isWPComAtomic
+                        scanScreenFeatureConfig.isEnabled() &&
+                                itemsVisibility.scan &&
+                                !site.isWPCom &&
+                                !site.isWPComAtomic
                 )
                 row_backup.setVisible(backupScreenFeatureConfig.isEnabled() && itemsVisibility.backup)
             }
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
@@ -280,7 +280,10 @@ class MySiteFragment : Fragment(),
         if (scanScreenFeatureConfig.isEnabled() || backupScreenFeatureConfig.isEnabled()) {
             uiScope.launch {
                 val itemsVisibility = jetpackCapabilitiesUseCase.getJetpackPurchasedProducts(site.siteId)
-                row_scan.setVisible(scanScreenFeatureConfig.isEnabled() && !site.isWPCom && itemsVisibility.scan)
+                row_scan.setVisible(
+                        scanScreenFeatureConfig.isEnabled() && itemsVisibility.scan &&
+                                !site.isWPCom && !site.isWPComAtomic
+                )
                 row_backup.setVisible(backupScreenFeatureConfig.isEnabled() && itemsVisibility.backup)
             }
         }
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
@@ -280,7 +280,7 @@ class MySiteFragment : Fragment(),
         if (scanScreenFeatureConfig.isEnabled() || backupScreenFeatureConfig.isEnabled()) {
             uiScope.launch {
                 val itemsVisibility = jetpackCapabilitiesUseCase.getJetpackPurchasedProducts(site.siteId)
-                row_scan.setVisible(scanScreenFeatureConfig.isEnabled() && itemsVisibility.scan)
+                row_scan.setVisible(scanScreenFeatureConfig.isEnabled() && !site.isWPCom && itemsVisibility.scan)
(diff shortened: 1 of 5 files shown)


WordPress/src/main/java/org/wordpress/android/ui/mysite/ScanAndBackupSource.kt
```diff
@@ -22,7 +22,8 @@ class ScanAndBackupSource
             val itemsVisibility = jetpackCapabilitiesUseCase.getJetpackPurchasedProducts(site.siteId)
             emit(
                     JetpackCapabilities(
-                            scanAvailable = scanScreenFeatureConfig.isEnabled() && itemsVisibility.scan,
+                            scanAvailable = scanScreenFeatureConfig.isEnabled() && itemsVisibility.scan &&
```
ParaskP7: 1) Minor (🔍): Since the `&&` logic is getting more complicated, I would suggest splitting it in multiple line to increase readability. For example:
```
scanAvailable = scanScreenFeatureConfig.isEnabled() &&
        itemsVisibility.scan &&
        !site.isWPCom &&
        !site.isWPComAtomic,
backupAvailable = backupScreenFeatureConfig.isEnabled() &&
        itemsVisibility.backup
```
2) Minor (🔍) It seems that this class can be reformatted.
3) Minor (🔍) The `@Inject constructor` line can be merged with its above line. This is not your change so feel free to ignore.

PS: Readability is very subjective so please ignore this comment if you feel about this differently.

WordPress/src/test/java/org/wordpress/android/ui/mysite/ScanAndBackupSourceTest.kt
```diff
@@ -109,8 +111,51 @@ class ScanAndBackupSourceTest : BaseUnitTest() {
...
+
+    @Test
+    fun `Scan visible on non-wpcom sites when Scan product is available and feature flag enabled`() = test {
+        init(scanScreenFeatureEnabled = true)
+        whenever(site.siteId).thenReturn(siteRemoteId)
+        whenever(jetpackCapabilitiesUseCase.getJetpackPurchasedProducts(siteRemoteId)).thenReturn(
+                JetpackPurchasedProducts(scan = true, backup = false)
+        )
+        whenever(site.isWPCom).thenReturn(false)
+        whenever(site.isWPComAtomic).thenReturn(false)
```
ParaskP7, line 161: Suggestion (💡): Since you explicitly added the `.thenReturn(false)` here, for both cases (`isWPCom` and `isWPComAtomic`), despite the fact that this is being setup again during `setUp()`, which I like, I suggest adding the missing `.thenReturn(false)` case on the other two cases above, to make that explicit as well.

WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
```diff
@@ -280,7 +280,10 @@ class MySiteFragment : Fragment(),
         if (scanScreenFeatureConfig.isEnabled() || backupScreenFeatureConfig.isEnabled()) {
             uiScope.launch {
                 val itemsVisibility = jetpackCapabilitiesUseCase.getJetpackPurchasedProducts(site.siteId)
-                row_scan.setVisible(scanScreenFeatureConfig.isEnabled() && itemsVisibility.scan)
+                row_scan.setVisible(
```
ParaskP7, line 285: 1) Minor (🔍) It seems that this class can be reformatted. 
2) Suggestion (💡): This `&&` logic is duplicated here and within the `ScanAndBackupSource` class. I suggest extracting this duplicated logic as it might cause problems in the future where a developer will forget updating one or the other. Wdyt?

PS: I would have extracted the logic into utility functions within the `ScanAndBackupSource` class and reuse those functions here. But, I am sure there are better ways, this was just the first thing that came to my mind.

ParaskP7 (approved):
👋 @malinajirka  !

I have left only minor (🔍) comments and some suggestions (💡) for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

WordPress/src/test/java/org/wordpress/android/ui/mysite/ScanAndBackupSourceTest.kt
malinajirka: Hmm, I usually explicitly list only things which are relevant for test/what the tests is trying to verify.

"Scan not visible on atomic sites even when Scan product is available" is trying to verify Atomic sites, so only "whenever(site.isWPComAtomic).thenReturn(true)" is listed.

"Scan not visible on wpcom sites even when Scan product is available" is trying to verify wpcom sites, so only "whenever(site.isWPCom).thenReturn(true)" is listed.

"Scan visible on non-wpcom sites when Scan product is available and feature flag enabled`" is trying to verify non-wpcom sites, which are defined as !wpcom && !atomic -> therefore 
"whenever(site.isWPCom).thenReturn(false)
        whenever(site.isWPComAtomic).thenReturn(false)" is listed.
        
        
What do you think about this approach? Would you still prefer if it was explicitly listed everywhere?

WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
malinajirka: > It seems that this class can be reformatted.

This class is going to be removed soon - when the ImprovemedMySiteFragment is finished. 

> This && logic is duplicated here and within the ScanAndBackupSource class. I suggest extracting this duplicated logic as it might cause problems in the future where a developer will forget updating one or the other. Wdyt?

Definitely a good comment! I thought about it and I'd agree in general, but since this class is going to be removed in 2-4 weeks I decided to not worry about the copy paste. The only shared place is the JetpackCapabilitiesUseCase and this logic doesn't fit there. Introducing a new class which we could share felt like an overkill. Wdyt?

malinajirka:
Thanks for the review @ParaskP7! I've left some comments with my thoughts - I'd really like to hear your thoughts before we merge this PR. Thanks!

WordPress/src/main/java/org/wordpress/android/ui/mysite/ScanAndBackupSource.kt
malinajirka: Fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/13962/commits/fdc9032df95dba6292afcbf1b076450c0809e79c

WordPress/src/test/java/org/wordpress/android/ui/mysite/ScanAndBackupSourceTest.kt
ParaskP7: 👋 @malinajirka !

Yeap, yeap, I got that, thanks for sharing your process. I would just add them explicitly just to show that one of them is `false` since this bit of information matters in that test context and the last test contain both as well. But, again, this is very opinionated everyone would do it differently, I just wanted to mention what I would have done.

PS: This is very minor so please ignore as such.

WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
ParaskP7: Oh, you are right, maybe it is not worth it after all, thanks for considering! 🤗

malinajirka: Actually, thinking about it -> I could move it to SiteUtils. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/mysite/ScanAndBackupSource.kt
ParaskP7: 👋 @malinajirka !

Thanks for this. 🙏 Just a quick note that the `&&` must be at the end of the line. A new line must not begin with `&&`. At least this is how I know this convention to be. Is this not the case here?

WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
ParaskP7: I would personally do it, just to make sure that I don't get in trouble in between, even if that is for 2-4 weeks. But, it is up to you.😄

WordPress/src/main/java/org/wordpress/android/ui/mysite/ScanAndBackupSource.kt
malinajirka: I'm never sure which one is correct and I always pick the wrong one :D.

ParaskP7: Same, but I have `Detekt` for that! 😄 

PS: Soon you will too! 🎉

malinajirka: Fixed in 078f00e2583508fd0b93227b89e719c5863eedba

WordPress/src/test/java/org/wordpress/android/ui/mysite/ScanAndBackupSourceTest.kt
malinajirka: Updated in 078f00e2583508fd0b93227b89e719c5863eedba

malinajirka: So bad news - the tests were complaining about "unnecessary stubbing" so I had to revert this change 😞 .

malinajirka:
Thanks @ParaskP7!

I've extracted the shared logic into SiteUtils - I added TODOs there, so we inline the logic into the ScanAndBackupSource when the legacy MySiteFragment is removed. I also decided to use `thenCallRealMethod()` in tests in order to keep the tests in the `ScanAndBackupSourceTest`.

WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
malinajirka: Done in https://github.com/wordpress-mobile/WordPress-Android/pull/13962/commits/4376c3ed5c1e762cd92aa9cbe400509ba8446333

ParaskP7:
Thank YOU @malinajirka ! 🌟 

> I've extracted the shared logic into SiteUtils - I added TODOs there, so we inline the logic into the ScanAndBackupSource when the legacy MySiteFragment is removed. I also decided to use thenCallRealMethod() in tests in order to keep the tests in the ScanAndBackupSourceTest.

I have reviewed the changes and they LGTM! 🌟 

Please feel free to merge this PR when CI is complete. 🙏

ParaskP7 (approved):

Issue #13456, opened 2020-11-24
Add support for invite links to P2 sites


khaykov:
We should add support for invite links to P2 sites. This will include a feature for generating (and disabling) invite links with specific privileges.

I'm not yet sure if we should add the functionality to open P2 links in the app.

This might also include a refactor of the People Management screen, since it's pretty old.

designsimply:
I would **love** to be able to accept invitations through the app instead of by email—I'm not sure if this is on the table as an option or in scope for this issue and would like to add that this is a personal note from me as a user 🙂

develric:
Adding designs for Android.

![image](https://user-images.githubusercontent.com/47797566/106901453-1084e700-66f8-11eb-987e-4b7eee5e44fb.png)

develric:
>I would love to be able to accept invitations through the app instead of by email—I'm not sure if this is on the table as an option or in scope for this issue and would like to add that this is a personal note from me as a user 🙂

Hi @designsimply , in my understanding is not currently covered and not sure what it could require tbh 🤔 . Maybe you could place a comment about it on our team project backlog (pctCYC-3s-p2)? I think it sounds like an interesting scenario to consider 👍  (cc @aerych)

PR #13963, opened 2021-02-04
Moving invite people to material.

develric:
As part of the #13456 , and before adding invite links functionality, this PR aims to move the Invite People screen to material design components. 

| Before | After |
|---|---|
| image | image |

And here is the dark theme one:
<p align=center><image src="https://user-images.githubusercontent.com/47797566/106925367-50f05f00-6710-11eb-8b3d-80e2997ddf5c.png" width=400/></p>

For the `Usernames or emails` field, we wanted to host usernames on chips inside a TextInputLayout styled with OutlinedBox as the other components in the screen. For this we created a custom component (more later). 

@mattmiklic a part the already discussed usage of chips, a deviation from original design (there can be other I'm not aware of 😄  ) is the remaining characters counter. I used the behaviour/aspect from the standard material design component (here `0/500`). We can eventually try to hack it but to me it seems consistent and I would rather stay with it (let me know wdyt 🙇  )

This PR do not bring the people fragment to VM architecture (but it's something we should probably do also to remove some of the logic from inside the views that is still the case here). Note that the fragment is a retained fragment.

### Component notes

As of our understanding, currently the TextInputLayout (that is a LinearLayout) works in OutlinedBox mode having
a single TextInputEditText as a direct child, inflating needed views into an internal not accessible FrameLayout;
so not easily possible to add additional views over the main TextInputLayout keeping the usual
behaviours of the TextInputEditText.

The new component tries to cover the usage of chips in combination with an EditText while mimic an OutlinedBox with TextInputLayout material design behaviour.

The implementation is pretty basic and do not cover all possible use cases. We should leave this component usage
once we find a more out of the box solution with material design lib (keep an eye on future releases) that
covers our use cases.

Supports hint/label animation (with RTL support).

#### New component
Would be good to check with different API levels
- Check the behaviour of the label animation is consistent with the standard control
- Check the control keeps the state on configuration changes
- check rtl and dark mode
- Whatever you can think of 😄 

#### People screen functionality
Note that the chps (and buttons before) can be created by delimiters (space, comma and enter) or the `done` keyboard button. Also changing focus or tapping `SEND` creates a chip/button from the text if available

- Test the functionality and compare there is no deviation from original
- Check that the Role control is not active when having only a Role on the site (I checked forcing by debugger)

PR submission checklist:


mattmiklic (approved):
This looks great to me @develric. I think the chips work really well here and I totally agree about not hacking the character count, it seems just fine as you've done it.

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
```diff
@@ -0,0 +1,729 @@
...
+
+        @ColorInt fun colorFromState(context: Context): Int {
+            val color = context.getColorResIdFromAttribute(this.colorAttr)
+
+            return ContextCompat.getColor(context, color)
+        }
+
+        companion object {
+            fun stateFromColor(context: Context, color: Int): ItemValidationState {
+                return values().first { value ->
```
khaykov, line 121: Wanted to confirm that it's ok to crash here if the wrong color is passed.

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
```diff
@@ -0,0 +1,729 @@
...
+        loadOutlineDrawable()
+        loadDimensions()
+        loadColors()
+
+        styleView(isEditorFocused(), hasItemsOrText(), false)
+
+        setListeners()
+    }
+
+    fun isEditorEmpty(): Boolean {
```
khaykov: I noticed that this function is not used.

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
```diff
@@ -0,0 +1,729 @@
...
+        return !this.hasText()
+    }
+
+    fun getTextIfAvaliableOrNull(): String? {
```
khaykov: Typo in available :)

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
```diff
@@ -0,0 +1,729 @@
...
+    fun addOrUpdateChip(item: String, state: ItemValidationState) {
+        throwExceptionIfChipifyNotEnabled()
+        chipify(item, state)
+    }
+
+    fun removeChip(item: String) {
+        throwExceptionIfChipifyNotEnabled()
+        val chips = getChipsMap()
+
+        if (chips.containsKey(item)) {
```
khaykov: I'm curious about the checks here - we can only have unique chips in the map, so I wonder in what situation we can have multiple chips be filtered into matchedChips? This is certainly a future proof approach, but if we only have unique chips, maybe we can simplify the logic to something like this:

 ```
   chips[item]?.let {
            flexbox.removeView(it)
            if (flexbox.childCount == 1) {
                styleView(isEditorFocused(), hasItemsOrText(), true)
            }
        }
```
I don't have any context, so feel free to ignore this one :)

File: WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt [hunk 5]
```diff
@@ -0,0 +1,729 @@
...
+
+                override fun newArray(size: Int): Array<SavedState?> {
+                    return arrayOfNulls(size)
+                }
+            }
+        }
+    }
+
+    companion object {
+        private const val MAX_NUMBER_OF_CHIPS_DEFAULT = 10
```
khaykov: Should we use this constant to specify a default value for `maxChips` when attributes is missing?

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
```diff
@@ -0,0 +1,729 @@
...
+            editor.setOnKeyListener { _, _, keyEvent ->
+                keyEvent.keyCode == KeyEvent.KEYCODE_DEL &&
+                        keyEvent.action == KeyEvent.ACTION_DOWN && removeLastEnteredItem()
+            }
+        }
+    }
+
+    private fun canAddMoreChips() = flexbox.childCount < maxChips + 1 || maxChips == 0
+
+    private fun manageAddItemTrigger(item: String) {
```
khaykov: Looking at the name I'm not too sure what this method does :)

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
```diff
@@ -0,0 +1,729 @@
...
+            val itemName = chipToRemove.text.toString()
+            flexBox.removeViewAt(flexbox.childCount - 2)
+            itemsManager?.onRemoveItem(itemName)
+            return true
+        }
+        return false
+    }
+
+    private fun endsWithDelimiter(string: String): Boolean {
+        if (string.isNullOrBlank()) {
```
khaykov: Since sting is not nullable in the declaration, maybe we can use `isBlank` instead?

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
```diff
@@ -0,0 +1,729 @@
...
+                setLabelColor(label, colorSurface, outlineColorAlphaFocused)
+
+                hint.visibility = View.VISIBLE
+                label.visibility = View.INVISIBLE
+            }
+        }
+    }
+
+    private fun animateViewTo(targetState: HelperTextState) {
+        val duration = 167L // this should be the material design default for the animation
```
khaykov: What do you think about having this duration in a global variable of some sort?

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
```diff
@@ -0,0 +1,729 @@
...
+        textView.setTextColor(colorWithAlpha)
+    }
+
+    private fun hasItemsOrText() = hasItems() || hasText()
+    private fun hasItems() = flexbox.childCount > 1
+    private fun hasText() = (editor.text?.length ?: 0) > 0
+    private fun isEditorFocused() = editor.isFocused
+    private fun hasHint() = hintResourceId != 0
+
+    override fun onSaveInstanceState(): Parcelable? {
```
khaykov: Since we are creating a `SavedState`, I don't think it's necessary to mark return object as nullable.

khaykov (commented):
Wow, great work on such complicated custom view, @develric !

Code looks great, and I only have some minor nitpicks.

Custom view works very well, I only found one issue during the testing - removing the last chip also removes text that is not chipified yet.
[Image from Gyazo](https://gyazo.com/53daf6f04b5739ec1dd66469ba99fded)

Not sure if this PR covers it, but I found a very minor issues with general layout - when you select a role, the error messages repopulate, which causes layout to jump.
[Image from Gyazo](https://gyazo.com/2109b8de1056f3c1e9e995a3e0b0466b)

Also, I wonder if we can add a bit of padding to the bottom of text field hint - when you have too much usernames entered it can move to the bottom, like this:
[Image from Gyazo](https://gyazo.com/fcb376beb0aec21b41c15da204ac1490)

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
develric: I think in this case yes; all the logic involved is internal, so if the color is not found there is an issue a logic issue to find. Let me know if you agree 🙇

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
develric: You are right 👍 . Strangely AS didn't report it in my case 🤔 . Anyway, removed it here f4115f5480c49c478560dac85e7f345ddc182965

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
develric: 👍 changed here f4115f5480c49c478560dac85e7f345ddc182965

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
develric: NIce! yep the idea could be to make the component cover also the case with duplicated chip items, but agree to change since now it's not covered 👍 . Changed here 507aca5cda1dfba582a342be69d8b8a433cbbff6

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
develric: ahah, that was the original intention 😄 . I then went to something like: if you don't set the wpMaxNumberChips then no limits in number of chips (0 is used); use the set limit otherwise. I removed the unused constant if you agree 🙇

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
develric: 😄 the intention was to say like `manage a trigger to add item`, where trigger is any condition that want to add an item...maybe just `addItem`is clearer 😄

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
develric: Agree, changed 👍

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
develric: 👍 changed.

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
khaykov: Yep, that's totally fine with me :) Just wanted to make sure 👍

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
khaykov: Manage is a bit broad, so if the methods adds an item I'm ok with `addItem` :)

WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
develric: I used auto complete to create the function prototype and was going to say I would prefer to keep it, but checked that kt complains if you return a nullable type there, so changed it 👍

develric:
>Wow, great work on such complicated custom view, @develric !

🙇 

>removing the last chip also removes text that is not chipified yet

Great catch 😄 . Fixed in 65240d2f63a9ce0b57f3f27596d9688a651bce97

>Not sure if this PR covers it, but I found a very minor issues with general layout - when you select a role, the error messages repopulate, which causes layout to jump.

Noticed that one myself, but didn't change the behaviour for now. Just to share, the principle to recheck the errors listed is correct, since changing role is possible that for example users that already are associated to that role (so an error is reported) could be used for another role (so error should be removed). I can create an issue for this one to be covered later if you agree, given we could want to refactor toward VM 🙇 

>add a bit of padding to the bottom of text field hint

Makes sense, changed here 1aaf075ea6ab4d39ff22090568e5ec0da4cc6565

PR #13965, opened 2021-02-04
[Tooling/Fastlane] Delete unused helpers

AliSoftware:
Part of https://github.com/wordpress-mobile/release-toolkit/issues/204

This only removes the helpers that were either unused or are already present in release-toolkit.

Does not seem like much of a diff, but that change required me to compare each helper from this repo with the helpers from release-toolkit to ensure that even if they diverged over time and the diff wasn't empty, removing the local copies would not break anything.

For this first step I let the helpers and actions related to screenshots as they seem to only be present in WPAndroid. A separate step / PR will move those remaining ones into the release-toolkit and then remove them from the WPAndroid repo, but that's a larger piece of work (especially for screenshots we still need repo-specific files locally, like the `.ini` files, so the split won't be that trivial if we want it to be flexible and adapt to multiple repos in the future)

To test:

* Ensure the various actions still work (they should use the helpers from the release-toolkit now)


loremattei (commented):
Removing `android_git_helper` and `android_version_helper` look good to me. I briefly tested it on a modified version of `new_beta_release` and it worked as expected. 

I can't find `android_virtual_device_helper.rb` in `release-toolkit` tho 🤔 Am I blind or missing something?

AliSoftware:
> I can't find android_virtual_device_helper.rb in release-toolkit tho 🤔 Am I blind or missing something?

It's not, but iirc it's also not used anywhere (at least I couldn't find a reference to any of the methods in that helper anywhere in WPiOS, but would love a confidence check on that too)

jkmassel:
> iirc it's also not used anywhere

It looks like it's used in `TakeAndroidEmulatorScreenshotsAction` still, though it's quite possible that they could be replaced by things built into fastlane at some point?

AliSoftware:
> It looks like it's used in `TakeAndroidEmulatorScreenshotsAction` still, though it's quite possible that they could be replaced by things built into fastlane at some point?

Ah! You're right indeed! Not sure how I missed it (my guess is because there was multiple modules in that file, which is unusual and not idiomatic in Ruby, I might have only searched for references of the first one… 🤷 )… thank god for peer reviews! 😉 

Fixed, so we should be all good now!

AliSoftware:
Note: the Detekt linter failure is not coming from this PR, and in fact also fails in `develop`. [p1613163870090800/1613130377.085000-slack-C02QANACA]

jkmassel (approved):
This is ok in my view – I've verified both these helpers are now present over in the release toolkit (and freshly refactored!) 

IMHO we can leave it to @JavonDavis to press merge on this, since he'd be the one dealing with any breakage 😁

AliSoftware:
I've rebased the branch on top of `develop` to fix the CI issue during lint, so should soon be green again.

@JavonDavis please merge whenever you're ready (and CI is green) 🙂  – preferably before next release so you can test the changes on it

PR #13967, opened 2021-02-04
Add Remove functionality to the quick start dynamic cards

### LINKED ISSUE #13852: Add "Don't show again" functionality to the QS bottom sheet menu
When clicking on the "Don't show again" action on the Quick Start Card bottom sheet menu, the app should hide the card and never show it again.

https://user-images.githubusercontent.com/830056/106677014-b6204500-6596-11eb-8f53-ed6e243d0633.mov

Note: a prototype can be found in the design spec document linked on pbArwn-1fl-p2.

planarvoid:
Fixes #13852

This PR adds remove functionality to the QS dynamic cards. The removal is permanent and there is no going back except reinstalling the app (or clearing cache). The data is kept in the App preferences.

To test:
- Make sure your site has QuickStart started
- Click on card/hamburger menu/don't show again
- Notice the card disappears
- Restart the app
- Notice the card is still gone

https://user-images.githubusercontent.com/1079756/106914093-f69dd100-6704-11eb-9156-419ecf367a15.mp4

PR submission checklist:


renanferrari (commented):
Thanks for the changes! I can confirm the app is working as expected and that the code looks good.

I do have a question though as I've noticed that there's an [existing logic to remove the entire Quick Start section which also sets all the tasks as completed](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt#L1300-L1308). I don't think we need to do that here (setting tasks as completed), but I'm wondering if that won't have any implications or side effects on other behaviors. I wouldn't think so, but could you please double check?

planarvoid:
as far as I can see I don't think there could be any implications in either of the approaches. If we ever implement reminders, we need to be somehow finishing the quick start. Removing both cards from the my site screen doesn't seem to be the same action as "Finish QS" though so I'm not sure. I think we can keep this as it is for now @renanferrari

renanferrari (approved):
Thanks for looking into it, @planarvoid! 

LGTM :shipit:

PR #13968, opened 2021-02-04
Issue/backup restore design tweaks

zwarm:
Parent #13328 & #13329

This PR addresses a few design tweaks
1. Drops explicit color references and uses default. So instead of referencing `error_50` directly, it now uses `error`. Same for `warning`, `success`, and `default`.
2. Swaps subheader for description in the generic error view

To Test
- Run the restore & backup download process to the end and ensure all works correctly.

PR submission checklist:


malinajirka (approved):
LGTM, thanks!

PR #13970, opened 2021-02-04
Show post/page updated notification if this is not a first time publish

### LINKED ISSUE #3018: Updating a Published Page/Post notifies "Page Published" instead of updated

antonis:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/3018

Depends on: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1864

## Description

Shows post/page updated notification if this is not a first time publish

## To test:

**Note**: The tests should be executed **both for .com and self-hosted sites**

### Published Page / Post
1. Open a published Page / Post
2. Make a change
3. Save the change
4. Verify that the toast says “Page/Post Updated”

### Draft Page / Post
1. Open a draft Page / Post
2. Publish it
3. Verify that the toast says “Page/Post Published”

### Draft Page / Post List
1. Open the draft pages/posts list
2. Select Publish Now / Publish from the actions
3. Verify that the toast says “Page/Post Published”

### New Page / Post
1. Create a new Page / Post
2. Publish it
3. Verify that the toast says “Page/Post Published”

## Notes

Other approaches where examined to resolve this:
1. Fetch the post revisions. This would have the overhead of a network call just to change the wording from "published" to "updated".
2. [A workaround based on the isPublishDateInThePast utility method that is used in one more place](https://github.com/wordpress-mobile/WordPress-Android/pull/13942).
3. Fetching the status of the post before updating it in the `PostStore`. This though seemed promising did not work in all cases since the post status was set before uploading.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadUtils.java
```diff
@@ -511,9 +512,9 @@ public void onClick(View v) {
                         snackbarButtonRes = R.string.button_view;
 
                         if (post.isPage()) {
-                            snackbarMessageRes = R.string.page_published;
+                            snackbarMessageRes = isFirstTimePublish ? R.string.page_published : R.string.page_updated;
                         } else if (userCanPublish) {
-                            snackbarMessageRes = R.string.post_published;
+                            snackbarMessageRes = isFirstTimePublish ? R.string.post_published : R.string.post_updated;
```
mkevins, line 517: I wonder if it makes sense to introduce new strings here. :thinking: Perhaps, to distinguish between updating a draft post / page with a published one. Maybe something like `"Published (post|page) updated"`, or `"Updated published (post|page)"`. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
```diff
@@ -296,7 +296,7 @@ class ImprovedMySiteFragment : Fragment(),
                     is ItemUploadedModel.PostUploaded -> {
                         uploadUtilsWrapper.onPostUploadedSnackbarHandler(
                                 activity,
-                                requireActivity().findViewById(R.id.coordinator), true,
+                                requireActivity().findViewById(R.id.coordinator), true, false,
```
mkevins, line 299: I was having trouble tracking down why this should be `false` here, but I was also seeing local errors in AS for this file, even though it built, installed, and ran properly :man_shrugging: `Type mismatch: inferred type is () -> [ERROR : <ERROR FUNCTION RETURN TYPE>] but Observer<in Event<SnackbarMessageHolder>!> was expected`.

Not suggesting a change, but curious why this one is `false` (since my code navigation was failing here while reviewing, due to the weird errors I was seeing).

mkevins (approved):
The code looks correct. It's a lot of changes for a change in a toast :smile: , but it is technically more correct than the time-based approach.

I tested all flows listed on a Pixel 3a (physical device), and everything is working as described.

I had one comment about considering new strings for this case, but not a blocker.

WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
antonis: > Not suggesting a change, but curious why this one is `false` (since my code navigation was failing here while reviewing, due to the weird errors I was seeing).

Good question @mkevins 👍 
The `ItemUploadedModel.PostUploaded` has only one (that is not test code) [instance creation](https://github.com/wordpress-mobile/WordPress-Android/blob/9a94470259e16d228d73685d0b21e612ec1de665/WordPress/src/main/java/org/wordpress/android/ui/mysite/SiteIconUploadHandler.kt#L61). This is triggered by an `UploadErrorEvent` and always leads to the `isError` if branch of the `UploadUtils.onPostUploadedSnackbarHandler` method.

WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadUtils.java
antonis: That is a good idea @mkevins 👍 
It makes perfect sense to inform the user on this. At this point we have the following messages per case:
1. New post/page published -> `post_published` / `page_published`
2. Published post/page edited -> `post_updated` / `page_updated`
3. Draft post/page published -> `post_published` / `page_published`
4. Draft post/page updated -> [`editor_draft_saved_online`](https://github.com/wordpress-mobile/WordPress-Android/blob/c8f3f5dc35e39e23bd58cac39b39fa29bb68b5b7/WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadUtils.java#L500) 

I agree that the `"Published (post|page) updated"`/`"Updated published (post|page)"` would be more clear for case 2 and is an easy change (if I'm not missing some other corner case). I suggest to leave it as is for now thought since the same string is used in the [equivalent notification](https://github.com/wordpress-mobile/WordPress-Android/blob/02d13098ed55216215da2265ef391f6fbb128a98/WordPress/src/main/java/org/wordpress/android/ui/uploads/PostUploadNotifier.java#L325) and the original reported issue (internal ref `pb3aDo-Eu#comment-1181`) suggests using the `"Page Updates"` string.

It would be also nice to be able to distinguish between cases 1 and 3 but since `onPostUploadedSnackbarHandler` method does not have all the information needed it would probably make the implementation even more complex and escapes the scope of the original [issue](https://github.com/wordpress-mobile/gutenberg-mobile/issues/3018).

mkevins: Thanks for listing the other cases / scenarios as well. I agree, cases 1 and 3 would be harder to tackle (I'm not sure it's worth the compexity). Also, you raise a good point on the consistency with the notifications for case 2.

WordPress/src/main/java/org/wordpress/android/ui/mysite/ImprovedMySiteFragment.kt
mkevins: Thanks for the explanation!

PR #13976, opened 2021-02-05
Issue/13326 fix scan backup menu animation

malinajirka:
Parent issue #13326

This PR fixes two issues
1. List item added animation is played whenever you open MySite screen. This is solved by synchronously loading cached value (even when it's expired) of JetpackCapabilities. So unless the permissions changes or the user switches between sites the animation should not be played.

2. The app would crash if fetchJetpackCapabilities was requested before the previous request finished. This is fixed in https://github.com/wordpress-mobile/WordPress-Android/commit/f74b76ecbf773d9b291270fd97c6310d8d22497a. New approach adds support for parallel request for different sites. If a duplicate request is initiated for the same site, the previous request is force finished with an error.

To test:
- clear app data
1. Log in with an account with multiple jetpack sites (eg. the one from the doc)
2. Select "pressable-jetpack-complete" site and notice "Scan" and "Backup" items are shown (with animation)
3. Open Reader and go back to MySite
4. Notice "Scan" and "Backup" items are shown (withOUT animation)
5. Select pressable-jetpack-daily-scan site and notice "Scan" is shown (backup disappears with an animation)
6. Kill the app
7. Re-open the app and notice "Scan" is shown (no animation is played)
----------------------------------
- clear app data
1. Log in with an account with multiple jetpack sites (eg. the one from the doc)
2. Enable MySiteImprovements feature flag and repeat steps 2-7 from the above section
-------------------------
1. Select pressable-jetpack-daily-backup
2. Open Activity Log and notice filters are shown
3. Select pressable-jetpack-daily-scan
4. Open Activity Log and notice filters are NOT shown

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/JetpackCapabilitiesUseCase.kt
```diff
@@ -31,58 +30,62 @@ class JetpackCapabilitiesUseCase @Inject constructor(
...
-    @param:Named(BG_THREAD) private val bgDispatcher: CoroutineDispatcher
+    private val currentDateProvider: CurrentTimeProvider
 ) {
-    private var continuation: Continuation<OnJetpackCapabilitiesFetched>? = null
+    private var continuation: HashMap<Long, Continuation<OnJetpackCapabilitiesFetched>?> = hashMapOf()
 
-    suspend fun getJetpackPurchasedProducts(remoteSiteId: Long): JetpackPurchasedProducts {
-        val capabilities = withContext(bgDispatcher) {
-            getOrFetchJetpackCapabilities(remoteSiteId)
+    suspend fun getJetpackPurchasedProducts(remoteSiteId: Long) = flow {
```
malinajirka, line 37: This method now uses `flow` (it's used only for the new ImprovedMySiteFragment). It always returns cached capabilities and than initiates fetch of the up-to-date capabilities if needed.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/JetpackCapabilitiesUseCase.kt
```diff
@@ -31,58 +30,62 @@ class JetpackCapabilitiesUseCase @Inject constructor(
...
     }
 
-    suspend fun getOrFetchJetpackCapabilities(remoteSiteId: Long): List<JetpackCapability> {
-        return if (hasValidCache(remoteSiteId)) {
-            getCachedJetpackCapabilities(remoteSiteId)
-        } else {
-            fetchJetpackCapabilities(remoteSiteId)
-        }
-    }
+    fun getCachedJetpackPurchasedProducts(remoteSiteId: Long): JetpackPurchasedProducts =
```
malinajirka, line 44: getCachedJetpackPurchasedProducts and fetchJetpackPurchasedProducts are public since they are used directly from the legacy (currently used) My Site fragment. The legacy fragment doesn't use getJetpackPurchasedProducts (flow) -> it needs to synchronously load cached value since the state is refreshed from onResume().

WordPress/src/main/java/org/wordpress/android/ui/jetpack/JetpackCapabilitiesUseCase.kt
```diff
@@ -31,58 +30,62 @@ class JetpackCapabilitiesUseCase @Inject constructor(
...
+
+        val capabilities: List<JetpackCapability> = response.capabilities ?: listOf()
+        if (!response.isError) {
+            updateCache(remoteSiteId, capabilities)
+        }
+        return capabilities
+    }
+
+    private fun forceResumeDuplicateRequests(remoteSiteId: Long) {
+        continuation[remoteSiteId]?.let {
```
malinajirka, line 83: Instead of throwing an exception, we simply resume any duplicate requests with an exception.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/JetpackCapabilitiesUseCase.kt
```diff
@@ -96,12 +99,16 @@ class JetpackCapabilitiesUseCase @Inject constructor(
...
     fun onJetpackCapabilitiesFetched(event: OnJetpackCapabilitiesFetched) {
-        dispatcher.unregister(this@JetpackCapabilitiesUseCase)
-        continuation?.resume(event)
-        continuation = null
+        continuation[event.remoteSiteId]?.let {
+            continuation.remove(event.remoteSiteId)
+            it.resume(event)
+        }
+        if (continuation.isEmpty()) {
+            dispatcher.unregister(this@JetpackCapabilitiesUseCase)
```
malinajirka, line 110: We need to make sure to unregister the instance only when no requests are running (previously, multiple parallel requests for different sites were not supported at all)

WordPress/src/main/java/org/wordpress/android/ui/mysite/ScanAndBackupSource.kt
```diff
@@ -15,17 +16,19 @@ class ScanAndBackupSource
     private val jetpackCapabilitiesUseCase: JetpackCapabilitiesUseCase
 ) : MySiteSource<JetpackCapabilities> {
     override fun buildSource(siteId: Int) = flow {
-        emit(JetpackCapabilities(scanAvailable = false, backupAvailable = false))
```
malinajirka, line 18: We don't need to always set visibility to false before the request to fetch capabilities finished, since the use case now always returns the cached value first.

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -363,7 +363,7 @@ class ActivityLogViewModel @Inject constructor(
             !site.hasFreePlan -> refreshFiltersUiState()
             else -> {
                 viewModelScope.launch {
-                    val purchasedProducts = jetpackCapabilitiesUseCase.getJetpackPurchasedProducts(site.siteId)
+                    val purchasedProducts = jetpackCapabilitiesUseCase.getCachedJetpackPurchasedProducts(site.siteId)
```
malinajirka, line 366: ActivityLogViewModel will always use the cached value now.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/JetpackCapabilitiesUseCase.kt
```diff
@@ -31,58 +30,62 @@ class JetpackCapabilitiesUseCase @Inject constructor(
...
-            getCachedJetpackCapabilities(remoteSiteId)
-        } else {
-            fetchJetpackCapabilities(remoteSiteId)
-        }
-    }
+    fun getCachedJetpackPurchasedProducts(remoteSiteId: Long): JetpackPurchasedProducts =
+            mapToJetpackPurchasedProducts(getCachedJetpackCapabilities(remoteSiteId))
 
-    private suspend fun hasValidCache(remoteSiteId: Long): Boolean {
-        return withContext(bgDispatcher) {
```
malinajirka, line 58: I've removed all context switches from this class, since it doesn't perform any long-running operation so it can run even on the Ui thread. I previously considered loading data from sharedPreferences for a long-running operation - but it's causing more troubles than it's worth it.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/JetpackCapabilitiesUseCase.kt
```diff
@@ -31,58 +30,62 @@ class JetpackCapabilitiesUseCase @Inject constructor(
...
-            val lastUpdated = appPrefsWrapper.getSiteJetpackCapabilitiesLastUpdated(remoteSiteId)
-            lastUpdated > currentDateProvider.currentDate().time - MAX_CACHE_VALIDITY
-        }
+    suspend fun fetchJetpackPurchasedProducts(remoteSiteId: Long): JetpackPurchasedProducts =
+            mapToJetpackPurchasedProducts(fetchJetpackCapabilities(remoteSiteId))
+
+    private fun mapToJetpackPurchasedProducts(capabilities: List<JetpackCapability>) =
```
malinajirka, line 50: This has not been changed - just extracted into a standalone method

WordPress/src/main/java/org/wordpress/android/ui/jetpack/JetpackCapabilitiesUseCase.kt
```diff
@@ -31,58 +30,62 @@ class JetpackCapabilitiesUseCase @Inject constructor(
...
+            JetpackPurchasedProducts(
+                    scan = capabilities.find { SCAN_CAPABILITIES.contains(it) } != null,
+                    backup = capabilities.find { BACKUP_CAPABILITIES.contains(it) } != null
+            )
+
+    fun hasValidCache(remoteSiteId: Long): Boolean {
```
malinajirka, line 56: This method was not changed at all - I'm not sure why github marked it as changed.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/JetpackCapabilitiesUseCase.kt
```diff
@@ -31,58 +30,62 @@ class JetpackCapabilitiesUseCase @Inject constructor(
...
-                continuation = cont
-                dispatcher.dispatch(SiteActionBuilder.newFetchJetpackCapabilitiesAction(payload))
-            }
-
-            val capabilities: List<JetpackCapability> = response.capabilities ?: listOf()
-            if (!response.isError) {
-                updateCache(remoteSiteId, capabilities)
-            }
-            return@withContext capabilities
+        forceResumeDuplicateRequests(remoteSiteId)
```
malinajirka, line 66: I've removed the "withContext" and replace throw exception with `forceResumeDuplicateRequests(remoteSiteId)`, the rest is without changes.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/JetpackCapabilitiesUseCase.kt
zwarm: Happens to me sometimes too .. I think it's AS .

WordPress/src/main/java/org/wordpress/android/ui/jetpack/JetpackCapabilitiesUseCase.kt
zwarm: I agree with the decision made here.

zwarm (approved):
@malinajirka - Everything is working as expected. 👏  In addition to the tests you specified, I also attempted to recreate the crash and was unable to (switching site quickly - over & over again, back & forth through the tabs with site switching thrown in). 

Thanks for the walkthrough, it was very helpful.

PR #13979, opened 2021-02-05
Jetpack Section: Connect Activity Log Detail with Restore

ParaskP7:
Parent #13328

This PR connects the restore flow the activity log detail screen. This means that this PR makes sure that when a user presses the `RESTORE` button within the `Activity Log Detail` screen, then instead of getting a `Rewind Dialog` they get redirected to the `Restore` screen and the new restore flow. Also, when a restore gets triggered from within the new restore flow, this PR makes sure that the user get redirected back to the `Activity Log` screen, skipping the `Activity Log Detail` screen and starts seeing the restore progress row within that screen.

To test:
* Open app and go to `My Site` tab.
* Launch `Activity Log` or `Backup` screen,
* Find a rewindable item from the list and click on it. You will now enter the `Activity Log Detail` screen.
* On that screen, click the `RESTORE` button to launch the `Restore` screen,
* You should automatically be navigated back to the `Activity Log` screen, skipping the `Activity Log Detail` screen and a restore status will be triggered,
* Try doing the same steps again, but this time when within the `Restore` screen and flows, go back.
* Make sure that going back will navigate to to the `Activity Log Detail` screen.
-----
* Since the restore feature flag now defaults to `true`, in order to test the old flow you will need to manually override the default to `false`.
* Do that and make sure that the old flow, through the `Rewind Dialog` works as well.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/activitylog/detail/ActivityLogDetailActivity.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/activitylog/detail/ActivityLogDetailActivity.kt
@@ -40,6 +40,7 @@ class ActivityLogDetailActivity : LocaleAwareActivity(), BasicDialogPositiveClic
     }
 
     override fun onNegativeClicked(instanceTag: String) {
+        // Do nothing.
     }
 
     override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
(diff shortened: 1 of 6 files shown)


zwarm (approved):
@ParaskP7 - Everything works as expected. I am returned to activity log when explicitly exiting the flow and remain on detail when if I wait to end. LGTM 🎉

PR #13983, opened 2021-02-05
Issue/13328 restore tracks details

zwarm:
Parent #13328 

This PR
(1) updates the Tracks source when transitioning from Activity Log Detail view to the Restore flow as follows:
- backup/detail
- activity_log/detail

(2) Refactor to use `SOURCE_TRACK_EVENT_PROPERTY_KEY` in `showBackupDownloadForResult()` and `showRestoreForResult()` instead of hardcoding the word "source"

**To test:**
-Launch the app
- Navigate to Activity Log
- Select an item that is available for restore
- Tap on the item row
- Tap on the Restore icon
- Note the log shows: activity_log/detail

---------------
-Launch the app
- Navigate to Activity Log
- Select an item that is available for restore
- Tap on the item row
- Tap on the Restore icon
- Note the log shows: backup/detail

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
@@ -633,14 +633,13 @@ public class ActivityLauncher {
     ) {
         Map<String, Object> properties = new HashMap<>();
         properties.put(ACTIVITY_LOG_ACTIVITY_ID_KEY, activityId);
-        String source = null;
+        String source;
         if (rewindableOnly) {
             source = BACKUP_TRACK_EVENT_PROPERTY_VALUE;
-            properties.put(SOURCE_TRACK_EVENT_PROPERTY_KEY, BACKUP_TRACK_EVENT_PROPERTY_VALUE);
         } else {
             source = ACTIVITY_LOG_TRACK_EVENT_PROPERTY_VALUE;
-            properties.put(SOURCE_TRACK_EVENT_PROPERTY_KEY, ACTIVITY_LOG_TRACK_EVENT_PROPERTY_VALUE);
         }
+        properties.put(SOURCE_TRACK_EVENT_PROPERTY_KEY, source);
         AnalyticsUtils.trackWithSiteDetails(AnalyticsTracker.Stat.ACTIVITY_LOG_DETAIL_OPENED, site, properties);
 
         Intent intent = new Intent(activity, ActivityLogDetailActivity.class);
--- a/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
@@ -1445,7 +1445,6 @@ public class ActivityLauncher {
 
     public static void showRestoreForResult(Activity activity, @NonNull SiteModel site, String activityId,
                                                    int resultCode, String source) {
-
         Map<String, String> properties = new HashMap<>();
         properties.put(SOURCE_TRACK_EVENT_PROPERTY_KEY, source);
         AnalyticsTracker.track(Stat.JETPACK_RESTORE_OPENED, properties);
--- a/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
@@ -633,9 +633,12 @@ public class ActivityLauncher {
     ) {
         Map<String, Object> properties = new HashMap<>();
         properties.put(ACTIVITY_LOG_ACTIVITY_ID_KEY, activityId);
+        String source = null;
         if (rewindableOnly) {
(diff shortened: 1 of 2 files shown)


WordPress/src/main/java/org/wordpress/android/ui/activitylog/detail/ActivityLogDetailFragment.kt
```diff
@@ -74,12 +79,21 @@ class ActivityLogDetailFragment : Fragment() {
...
                 val spannable = activityLogModel?.content?.let {
-                    notificationsUtilsWrapper.getSpannableContentForRanges(it, activityMessage, { range ->
-                        viewModel.onRangeClicked(range)
-                    }, false)
+                    notificationsUtilsWrapper.getSpannableContentForRanges(
+                            it,
+                            activityMessage,
+                            { range ->
+                                viewModel.onRangeClicked(range)
+                            },
```
ParaskP7, line 88: Minor (🔍): Consider making this lambda a one-liner: `{ range -> viewModel.onRangeClicked(range) },`

WordPress/src/main/java/org/wordpress/android/ui/activitylog/detail/ActivityLogDetailFragment.kt
```diff
@@ -168,11 +188,27 @@ class ActivityLogDetailFragment : Fragment() {
...
+                            item.createdTime
+                    ),
                     getString(R.string.activity_log_rewind_site),
                     getString(R.string.cancel)
             )
             dialog.show(requireFragmentManager(), it)
         }
     }
+
+    private fun buildTrackingSource() =
```
ParaskP7: 1) Minor (🔍): I really liked the `forwardSlash` variable that you initially had, since this will make the whole string concatenation more explicit and readable. However, I would suggest having that as a constant instead, like `FORWARD_SLASH` (see full example below).
2) Minor (🔍): When a function is converted into an expression body, and there are braces then the equal can be on the same line if it fits. This will make the function look as a body function (see full example below).
3) Minor (🔍): Consider getting the string and checking if that is nullable instead of using `containKey(...)`. This will make it a bit more readable since the `SOURCE_TRACK_EVENT_PROPERTY_KEY` will only be utilised once (see full example below).

```
    private fun buildTrackingSource() = requireActivity().intent?.extras?.let {
        val source = it.getString(SOURCE_TRACK_EVENT_PROPERTY_KEY)
        when {
            source != null -> source + FORWARD_SLASH + DETAIL_TRACKING_SOURCE
            else -> DETAIL_TRACKING_SOURCE
        }
    }
```

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
```diff
@@ -633,16 +633,20 @@ public static void viewActivityLogDetailForResult(
     ) {
         Map<String, Object> properties = new HashMap<>();
         properties.put(ACTIVITY_LOG_ACTIVITY_ID_KEY, activityId);
+        String source = null;
```
ParaskP7: 1) Minor (🔍): Consider removing the optional ` = null`.
2) Minor (🔍): Consider reusing the `TRACK_EVENT_PROPERTY_VALUE` instead of duplicating them. This will make the code more robust and less error-prone in the future. The below example removes an extra duplication as well:

```
String source;
if (rewindableOnly) {
    source = BACKUP_TRACK_EVENT_PROPERTY_VALUE;
} else {
    source = ACTIVITY_LOG_TRACK_EVENT_PROPERTY_VALUE;
}
properties.put(SOURCE_TRACK_EVENT_PROPERTY_KEY, source);
```

ParaskP7 (approved):
👋 @zwarm  !

I have left only minor (🔍) comments for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

PS: I have noted that on the first commit the whole class was reformatted. I am not sure why that happened because in my case it wouldn't reformat it. What is your line length configuration (should be `120`)? Btw, I also suspect that this was done automatically due to the fact that some commit options was checked while on the commit dialog. Usually, I have all options checked, like `Reformat code`, `Rearrange code`, etc. But, after committing, what I do is then check whether additional changes where included in my commit, ones that I didn't wanted. If that is the case, I undo my commit, run reformat on another commit and commit my initially intended code as a second commit. This helps me keep my commits clear from unwanted automatic IDE changes. Let me know if you are interested in this process of mine and we can do a pair programming at some point to exchange ideas since I would also love to hear how you are doing it on your side!

WordPress/src/main/java/org/wordpress/android/ui/activitylog/detail/ActivityLogDetailFragment.kt
zwarm: @ParaskP7 - this was an auto AS thing. We should review settings as a team during our next sync

PR #13994, opened 2021-02-08
Issue/13326 extract to use case

malinajirka:
Parent issue #13326

This PR wraps "fetchScanHistory" call into a usecase as per [this comment](https://github.com/wordpress-mobile/WordPress-Android/pull/13802#discussion_r559942517).

To be honest, I'm still not sure if we should be wrapping these types of code blocks into use cases. If you look at the changes in `ScanHistoryViewModel.kt`, the before/after code is almost identical, but we have an extra class which doesn't do almost anything. I'm not even sure this make the code more consistent, because AFAIK we mostly don't use usecases for basic calls to FluxC or do we? It feels to me that we are just adding one extra layer which has basically no purpose and doesn't bring any benefits right now. I'd personally introduce use cases only in [the following scenarios](https://github.com/wordpress-mobile/WordPress-Android/pull/13802#discussion_r560046539).

Wdyt @ParaskP7 @zwarm @ashiagr?

To test:
tbd

PR submission checklist:

### DIFF
--- a/WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanHistoryUseCaseTest.kt
+++ b/WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanHistoryUseCaseTest.kt
@@ -1,0 +1,65 @@
+package org.wordpress.android.ui.jetpack.scan.usecases
+
+import com.nhaarman.mockitokotlin2.anyOrNull
+import com.nhaarman.mockitokotlin2.whenever
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Before
+import org.junit.Test
+import org.mockito.Mock
+import org.wordpress.android.BaseUnitTest
+import org.wordpress.android.fluxc.action.ScanAction.FETCH_SCAN_HISTORY
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.store.ScanStore
+import org.wordpress.android.fluxc.store.ScanStore.FetchScanHistoryError
+import org.wordpress.android.fluxc.store.ScanStore.FetchScanHistoryErrorType.GENERIC_ERROR
+import org.wordpress.android.fluxc.store.ScanStore.OnScanHistoryFetched
+import org.wordpress.android.test
+import org.wordpress.android.ui.jetpack.scan.usecases.FetchScanHistoryUseCase.FetchScanHistoryState.Failure
+import org.wordpress.android.ui.jetpack.scan.usecases.FetchScanHistoryUseCase.FetchScanHistoryState.Success
+import org.wordpress.android.util.NetworkUtilsWrapper
+
+class FetchScanHistoryUseCaseTest : BaseUnitTest() {
+    private lateinit var fetchScanHistoryUseCase: FetchScanHistoryUseCase
+
+    @Mock private lateinit var networkUtilsWrapper: NetworkUtilsWrapper
+    @Mock private lateinit var scanStore: ScanStore
+    private val site: SiteModel = SiteModel()
+
+    @Before
+    fun setUp() {
+        fetchScanHistoryUseCase = FetchScanHistoryUseCase(networkUtilsWrapper, scanStore)
+
+        whenever(networkUtilsWrapper.isNetworkAvailable()).thenReturn(true)
+    }
+
+    @Test
+    fun `Network failure returned, when the device is offline`() = test {
+        whenever(networkUtilsWrapper.isNetworkAvailable()).thenReturn(false)
+
+        val result = fetchScanHistoryUseCase.fetch(site)
+
+        assertThat(result).isInstanceOf(Failure.NetworkUnavailable::class.java)
+    }
+
+    @Test
+    fun `Request failure returned, when the request fails`() = test {
+        whenever(scanStore.fetchScanHistory(anyOrNull())).thenReturn(
+                OnScanHistoryFetched(site.siteId, FetchScanHistoryError(GENERIC_ERROR), FETCH_SCAN_HISTORY)
+        )
+
+        val result = fetchScanHistoryUseCase.fetch(site)
+
+        assertThat(result).isInstanceOf(Failure.RemoteRequestFailure::class.java)
+    }
+
+    @Test
+    fun `Success returned, when the request succeeds`() = test {
+        whenever(scanStore.fetchScanHistory(anyOrNull())).thenReturn(
+                OnScanHistoryFetched(site.siteId, null, FETCH_SCAN_HISTORY)
+        )
+
+        val result = fetchScanHistoryUseCase.fetch(site)
+
+        assertThat(result).isInstanceOf(Success::class.java)
+    }
+}
--- a/WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanHistoryUseCaseTest.kt
+++ b/WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanHistoryUseCaseTest.kt
@@ -6,7 +6,6 @@ import com.nhaarman.mockitokotlin2.verify
 import com.nhaarman.mockitokotlin2.whenever
 import kotlinx.coroutines.InternalCoroutinesApi
 import org.assertj.core.api.Assertions.assertThat
-import org.greenrobot.eventbus.ThreadMode
 import org.junit.Before
 import org.junit.Test
 import org.mockito.Mock
@@ -104,5 +103,4 @@ class FetchScanHistoryUseCaseTest : BaseUnitTest() {
 
         assertThat((result as Success).threatModels).isEqualTo(threats)
     }
-
 }
--- a/WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanHistoryUseCaseTest.kt
+++ b/WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanHistoryUseCaseTest.kt
@@ -81,18 +81,7 @@ class FetchScanHistoryUseCaseTest : BaseUnitTest() {
     }
 
     @Test
-    fun `Success returned, when the request succeeds`() = test {
-        whenever(scanStore.fetchScanHistory(anyOrNull())).thenReturn(
-                OnScanHistoryFetched(site.siteId, null, FETCH_SCAN_HISTORY)
-        )
-
-        val result = fetchScanHistoryUseCase.fetch(site)
-
-        assertThat(result).isInstanceOf(Success::class.java)
-    }
-
-    @Test
-    fun `Data from db returned, when the request suceeds`() = test {
+    fun `Data from db returned, when the request succeeds`() = test {
         val threats = listOf<ThreatModel>(mock(), mock())
         whenever(scanStore.fetchScanHistory(anyOrNull())).thenReturn(
                 OnScanHistoryFetched(site.siteId, null, FETCH_SCAN_HISTORY)
--- a/WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanHistoryUseCaseTest.kt
+++ b/WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanHistoryUseCaseTest.kt
@@ -1,14 +1,20 @@
 package org.wordpress.android.ui.jetpack.scan.usecases
 
 import com.nhaarman.mockitokotlin2.anyOrNull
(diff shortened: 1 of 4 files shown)


ParaskP7:
👋 @malinajirka !

> To be honest, I'm still not sure if we should be wrapping these types of code blocks into use cases. If you look at the changes in ScanHistoryViewModel.kt, the before/after code is almost identical, but we have an extra class which doesn't do almost anything. I'm not even sure this make the code more consistent, because AFAIK we mostly don't use usecases for basic calls to FluxC or do we? It feels to me that we are just adding one extra layer which has basically no purpose and doesn't bring any benefits right now. I'd personally introduce use cases only in the following scenarios.

I am also very opinionated on the `use case` argument, so please take this with a grain of salt.

I have been using the `clean architecture` pattern for a while now and I do see the benefit of always going through use cases, even if that is the simplest thing. It indeed creates more boilerplate code, but this is done for a good reason, separation of concerns and consistency (even ease of modularization).

In our codebase, we have both approached, with and without use cases. Usually we do utilise use cases when the logic gets too complicated, but even then sometimes we don't. In my opinion, but having a rule to always go through a use case, will ease this. Every developer will not need to worry about what to do, where, since they will always follow that path. Also, doing that will help with testing, since any view model will usually only mock the use case, making testing and setting it up for view models a bit more straightforward.

When a use case is very simple, I think about the use case as a stepping stone for future complexity, much like when you create the first test on any class. It is a stepping stone for making it easy for future enhancements on that class.

-----

Suggestion (💡): I rewrote your solution a bit to help with the `use case` case. Consider the following alternative, it will probably make the sense why utilising a `use case` at least in this case is a good thing:

`ScanHistoryViewModel.kt`
```
    private fun fetchScanHistory(isRetry: Boolean = false) {
        launch {
            _uiState.value = ContentUiState
            when (val result = fetchScanHistoryUseCase.fetch(site, isRetry)) {
                is Success -> _threats.postValue(result.threadModels)
                is NetworkUnavailable -> _uiState.value = NoConnection(this@ScanHistoryViewModel::onRetryClicked)
                is RemoteRequestFailure -> _uiState.value = RequestFailed(this@ScanHistoryViewModel::onRetryClicked)
            }
        }
    }
```

`FetchScanHistoryUseCase.kt`
```
    suspend fun fetch(
        site: SiteModel,
        isRetry: Boolean
    ): FetchScanHistoryState {
        if (isRetry) {
            delay(RETRY_DELAY)
        }
        return if (!networkUtilsWrapper.isNetworkAvailable()) {
            scanTracker.trackOnError(ScanTracker.ErrorAction.FETCH_SCAN_HISTORY, ScanTracker.ErrorCause.OFFLINE)
            FetchScanHistoryState.Failure.NetworkUnavailable
        } else {
            val result = scanStore.fetchScanHistory(FetchScanHistoryPayload(site))
            if (result.isError) {
                scanTracker.trackOnError(ScanTracker.ErrorAction.FETCH_SCAN_HISTORY, ScanTracker.ErrorCause.REMOTE)
                FetchScanHistoryState.Failure.RemoteRequestFailure
            } else {
                withContext(bgDispatcher) {
                    FetchScanHistoryState.Success(scanStore.getScanHistoryForSite(site))
                }
            }
        }
```

PS: By pushing as many moving part as makes sense to the `use case` will even make testing easier and more complete.

ParaskP7:
👋 again @malinajirka !

I also noticed an inconsistency while playing a bit with the UI. I know this is not part of this PR, but I though to mention it here for you to maybe check it really quick or create a todo task for us to fix it later on.



PS: This happens when you have the screen opened, then you lock your phone, wait for 5'' secs, and unlock it.

malinajirka:
> I rewrote your solution a bit to help with the use case case. Consider the following alternative, it will probably make the sense why utilising a use case at least in this case is a good thing:

Thanks for the suggestions;)! I always try to keep use cases "screen independent" so I try to avoid putting any ui and tracking related logic there. Having said that, since we are technically tracking request errors I think it's a good call to move the tracking into the usecase. I'm not so sure about the "isRetry" delay though - it's purely a UI related logic -> we want have the delay there, so a progress is shown on the UI at least for a bit after the user hits the retry button.

> I have been using the clean architecture pattern for a while now and I do see the benefit of always going through use cases, even if that is the simplest thing. It indeed creates more boilerplate code, but this is done for a good reason, separation of concerns and consistency (even ease of modularization).

Whether a vm invokes "store.fetch()" or "usecase.fetch()" doesn't make much difference regarding separation of concerns imo. In both cases, the VM is responsible for handling the result so the concerns are identical, aren't they? 
I agree that if we decide that all communication with FluxC should go through use-cases than we should use it no matter the complexity of the call for consistency. However, we don't have such rule now. I personally think it's an overkill/over-engineering. There is tons of scenarios when a code is used just once in a codebase and an extra layer just makes the code harder to follow.

Having said all that, I agree that moving the tracking code into the use case make sense here (since it's not related to UI but to the request's result). With that, the use-case can be easily re-used and whoever re-uses it can't forget to track the result.
Wdyt?

> When a use case is very simple, I think about the use case as a stepping stone for future complexity, much like when you create the first test on any class. It is a stepping stone for making it easy for future enhancements on that class.

Preparing code for something which might never happen is always a bit tricky. First test has a purpose as soon as you introduce it, however, use-case with a single line of code (or single line + mapping) imo doesn't have any purpose at the time of creation. I agree that it'll facilitate work to future devs if they need to re-use it or add some logic there. But if it's not in a use case yet, they can create it and spent the same time as you would have. If no-one ever needs to create the usecase, your time working on the usecase would been just wasted. 

Anyway, this all depends a lot on personal preferences. I'll make some of the changes you proposed and turn this into a non-draft PR. 

> I also noticed an inconsistency while playing a bit with the UI. I know this is not part of this PR, but I though to mention it here for you to maybe check it really quick or create a todo task for us to fix it later on.

Nice catch! I've added it to the list of tasks.

ParaskP7:
> I always try to keep use cases "screen independent" so I try to avoid putting any ui and tracking related logic there. Having said that, since we are technically tracking request errors I think it's a good call to move the tracking into the usecase. I'm not so sure about the "isRetry" delay though - it's purely a UI related logic -> we want have the delay there, so a progress is shown on the UI at least for a bit after the user hits the retry button.

1) Track Events: Yes, I was thinking the same and ended up putting the tracking related logic within the use-case because of the nature of tracking, which is `request errors`. I think they owned to be there so that others that might re use this use-case will benefit as well.
1) Retry Logic: Yeap I wasn't sure about that and because we where putting this logic within the use-case with all the recent use-cases that we created, I though it might fit there. Most probably this is not the case here, will have to think on that a bit.
1) Store Get: What I like with the use-case approach I suggested is that I added the background dispatcher and the logic to get the list of models within the use case itself so that to make it easier for the client.
1) With Context: Since we are always using the store to fetch (from network) and get (from database), switching context to a background or io dispatcher should always be the default. Maybe having this logic within the use-case can be another reason to utilise use-cases instead of having this on the view model layer. Wdyt? Obviously, this need to be part of the store itself and pushed to the lowest level, but it is a good start I believe.

> Whether a vm invokes "store.fetch()" or "usecase.fetch()" doesn't make much difference regarding separation of concerns imo. In both cases, the VM is responsible for handling the result so the concerns are identical, aren't they?

Actually, since we kind of always utilise the `fetch` and `get` from store approach, maybe having the use-case to combine those make even more sense than having the view model deal with that explicitly.

> I agree that if we decide that all communication with FluxC should go through use-cases than we should use it no matter the complexity of the call for consistency. However, we don't have such rule now. I personally think it's an overkill/over-engineering. There is tons of scenarios when a code is used just once in a codebase and an extra layer just makes the code harder to follow.

I agree with the `over-engineering` comment, but this is what you get with clean architecture, more boilerplate which aims at sacrificing simplicity for the 90% of the cases over separation of concerts and better testability. I know this is not as easy decision to be made and thus teams agree to make a decision in favor or against that and then move forward knowing the limitation of each approach.

> Having said all that, I agree that moving the tracking code into the use case make sense here (since it's not related to UI but to the request's result). With that, the use-case can be easily re-used and whoever re-uses it can't forget to track the result.
Wdyt?

👍 

> Preparing code for something which might never happen is always a bit tricky. First test has a purpose as soon as you introduce it, however, use-case with a single line of code (or single line + mapping) imo doesn't have any purpose at the time of creation. I agree that it'll facilitate work to future devs if they need to re-use it or add some logic there. But if it's not in a use case yet, they can create it and spent the same time as you would have. If no-one ever needs to create the usecase, your time working on the usecase would been just wasted.

Yes, this is tricky indeed, I think I covered that above. When it comes to this it is an architectural decision with pros and cons that the team has to agree and feel good about. I personally I am in favor of it, even with this cons that you described. But, this is my experience and everyone has their own.

> Anyway, this all depends a lot on personal preferences. I'll make some of the changes you proposed and turn this into a non-draft PR.

Yes indeed, this is purely theoretical discussion and I am fine with anything the team decides to do going forward, thus I started this thread with the `grain of salt` sentence. 😅

malinajirka:
> With Context: Since we are always using the store to fetch (from network) and get (from database), switching context to a background or io dispatcher should always be the default. Maybe having this logic within the use-case can be another reason to utilise use-cases instead of having this on the view model layer. Wdyt? Obviously, this need to be part of the store itself and pushed to the lowest level, but it is a good start I believe.

I'd personally prefer to have the "withContext" in FluxC and provide a suspendable function from there.

> Actually, since we kind of always utilise the fetch and get from store approach, maybe having the use-case to combine those make even more sense than having the view model deal with that explicitly.

I agree that for the "get-fetch" cases using the usecase is a good approach. Tbh I missed that you moved the "get" into the usecase.



👍 With the improvements you did, this codeblock is definitely a candidate for a use-case. For other cases, I think you made some great points and in most scenarios using a use-case should be preferred (still just my opinion not division). Thank you so much for all your input!

ParaskP7:
> I'd personally prefer to have the "withContext" in FluxC and provide a suspendable function from there.

💯 

> I agree that for the "get-fetch" cases using the usecase is a good approach.

Yeap, it might be something to consider going forward now that we are getting more and more comfortable with Coroutines.

> Tbh I missed that you moved the "get" into the usecase.

😅 I kind of figured out... 😝 

> 👍 With the improvements you did, this codeblock is definitely a candidate for a use-case. For other cases, I think you made some great points and in most scenarios using a use-case should be preferred (still just my opinion not division). Thank you so much for all your input!

You made some excellent points as well, and I am definitely comfortable with any approach we take having discussed as many arguments and pros/cons as possible based on the needs of our codebase.

Thank YOU for starting this discussion, always a pleasure to exchange opinion with you Mr. Jirka! 🙏

malinajirka:
Thanks @ParaskP7! I've updated the code as per your suggestions with one exception - I kept the isRetry delay in the VM.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanHistoryUseCaseTest.kt
```diff
@@ -0,0 +1,106 @@
...
+                OnScanHistoryFetched(site.siteId, null, FETCH_SCAN_HISTORY)
+        )
+
+        val result = fetchScanHistoryUseCase.fetch(site)
+
+        assertThat(result).isInstanceOf(Success::class.java)
+    }
+
+    @Test
+    fun `Data from db returned, when the request suceeds`() = test {
```
ParaskP7: 1) Minor (🔍): There is a typo in `suceeds`.
2) Suggestion (💡): Consider removing the above `isInstanceOf` test since this assert logic in included within this test as well (`result as Success`). I understand that you wanted to split the test and have them do exactly one thing, but in that case a `Success` has no real value without the list of threat models and as such I would consider them the same (as you did with this test). If the `Success` returned value and the `scanStore` logic was doing two separate things I would indeed have them as separate test, but in this instance they both work for the same single result. Wdyt?

ParaskP7 (approved):
👋 @malinajirka  !

I have reviewed and tested this PR as per the instruction, everything works as expected, great job! 🌟 🌟 

I have left one minor (🔍) comment and a suggestion (💡) for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/usecases/FetchScanHistoryUseCaseTest.kt
malinajirka: Fixed in https://github.com/wordpress-mobile/WordPress-Android/pull/13994/commits/aa14dcdffc630331fef6d69b83b4f71ec6600bc3

PR #13290, opened 2020-11-04
Detekt Integration - Showcase

ParaskP7:
🚫  This draft PR is just for showcasing purposes and as such it should not be merged to `develop`.

PR #13996, opened 2021-02-08
Detekt Integration - Phase.1

ParaskP7:
External Links: [Detekt](https://github.com/detekt/detekt) - [Detekt Documentation](https://detekt.github.io/detekt/)
Internal Link: `paqN3M-eg-p2` (Detekt Integration)
Related PR: [Showcase](https://github.com/wordpress-mobile/WordPress-Android/pull/13290)
Next Phases: [Phase.2](https://github.com/wordpress-mobile/WordPress-Android/issues/14011) - [Phase.3](https://github.com/wordpress-mobile/WordPress-Android/issues/14012)

This PR adds `Detekt` to the `WordPress` repo. Detekt is a static code analysis tool for the Kotlin programming language. Detekt does for Kotlin what CheckStyle does for Java, only so much more.

To test (🧪 ):
- Make sure to read the external and internal links mention above to get an overall idea on what `Detekt` has to offer, how it can be configured and utilised.
- Pull `detekt-integration` on a local branch.
- Run `./gradlew WordPress:detekt` and verify that the build is successful.
- Delete the `config/detekt/baseline.xml` file, comment out the `Detekt's` baseline configuration within the `WordPress/build.gradle` file and run `./gradlew WordPress:detekt` again. Verify that the build fails, reporting `880` issues.
- Install the `detekt` IDE plugin and configure it following the instructions within the `coding-style.md` readme file. Verify that everything works and that the IDE reports `Detekt` violations as well.
- Locate the `Detekt` related `Circle CI` job and verify correctness.

Questions (❓):
- For @loremattei: `Detekt` was configured on the root `build.gradle` file, within the `allprojects { ... }` scope. However, I have noticed that the `subprojects { ... }` is utilised as well for static analysis tools like `KtLint`. Do I need to add an additional `task` for `Detekt` within the `subprojects { ... }`?
- For @loremattei: The exist a task within the WordPress `build.gradle` file, called `violationCommentsToGitHub`. This task has a `violations` configuration for `CheckStyle`. Do I need to add something similar for `Detekt`?

**Merge Instructions** (⚠️):
1. Make sure 2 reviewers have approved the PR (@loremattei and @malinajirka).
1. Remove the `Not Ready for Merge` label.
1. Merge this PR.

PR submission checklist:


build.gradle
```diff
@@ -79,6 +82,21 @@ allprojects {
         configFile file("${project.rootDir}/config/checkstyle.xml")
     }
 
+    detekt {
+        config = files("${project.rootDir}/config/detekt/detekt.yml")
+        autoCorrect = false
+        buildUponDefaultConfig = false
+        disableDefaultRuleSets = false
+        ignoreFailures = false
+        parallel = false
```
malinajirka, line 91: ❓ Is there any reason why we don't want to run the build in parallel? Or are we keeping it to false since it's the default configuration?

build.gradle
```diff
@@ -79,6 +82,21 @@ allprojects {
...
+        debug = false
+        reports {
+            html.enabled = true
```
malinajirka, line 94: 💡 Idea: Unless we are planning to use all three formats, I'd personally disabled those we don't need.

config/detekt/detekt.yml
```diff
@@ -0,0 +1,689 @@
...
+    thresholdInObjects: 11
+    thresholdInEnums: 11
+    ignoreDeprecated: false
+    ignorePrivate: false
+    ignoreOverridden: false
+
+coroutines:
+  active: true
+  GlobalCoroutineUsage:
+    active: false
```
malinajirka, line 132: ❓ We might want to consider enabling this, wdyt?

config/detekt/detekt.yml
```diff
@@ -0,0 +1,689 @@
...
+    exceptionNames:
+      - Error
+      - Exception
+      - Throwable
+      - RuntimeException
+
+formatting:
+  active: true
+  android: false
+  autoCorrect: true
```
malinajirka: ⚠️  Is it ok to have "autocorrect=true" on CI? Won't this introduce local changes?

config/detekt/detekt.yml
```diff
@@ -0,0 +1,689 @@
...
+    active: true
+    exceptionNames:
+      - Error
+      - Exception
+      - Throwable
+      - RuntimeException
+
+formatting:
+  active: true
+  android: false
```
malinajirka: ❓ What is this flag for? I haven't found it in the docs.

config/detekt/detekt.yml
```diff
@@ -0,0 +1,689 @@
...
+  LateinitUsage:
+    active: false
+    excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']
+    excludeAnnotatedProperties: []
+    ignoreOnClassesPattern: ''
+  MapGetWithNotNullAssertionOperator:
+    active: false
+  MissingWhenCase:
+    active: true
+    allowElseExpression: true
```
malinajirka, line 494: I personally think that `else/default` branches should be avoided, so I'd turn this flag on. But I guess we can start with this setup and turn it on later when people get use to Detekt.

config/detekt/detekt.yml
```diff
@@ -0,0 +1,689 @@
...
+  active: true
+  ClassOrdering:
+    active: false
+  CollapsibleIfStatements:
+    active: false
+  DataClassContainsFunctions:
+    active: false
+    conversionFunctionPrefix: 'to'
+  DataClassShouldBeImmutable:
+    active: false
```
malinajirka, line 526: I'd love to turn this on one day :).

config/detekt/detekt.yml
```diff
@@ -0,0 +1,689 @@
...
+  ExplicitCollectionElementAccessMethod:
+    active: false
+  ExplicitItLambdaParameter:
+    active: false
+  ExpressionBodySyntax:
+    active: false
+    includeLineWrapping: false
+  ForbiddenComment:
+    active: true
```
malinajirka, line 539: ⚠️ Since we are merging code behind feature flags into develop I don't think we can avoid using "TODO" comments. Wdyt?

config/detekt/detekt.yml
```diff
@@ -0,0 +1,689 @@
...
+    values: ['TODO:', 'FIXME:', 'STOPSHIP:']
+    allowedPatterns: ''
+  ForbiddenImport:
+    active: false
+    imports: []
+    forbiddenPatterns: ''
```
malinajirka, line 545: ❓ Could we forbid imports ending with `.*`? I think ktlint/lint doesn't allow them so it might make sense to have the same rule here. Wdyt?

config/detekt/detekt.yml
```diff
@@ -0,0 +1,689 @@
...
+    excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']
+    ignoreNumbers: ['-1', '0', '1', '2']
+    ignoreHashCodeFunction: true
+    ignorePropertyDeclaration: false
+    ignoreLocalVariableDeclaration: false
+    ignoreConstantDeclaration: true
+    ignoreCompanionObjectPropertyDeclaration: true
+    ignoreAnnotation: false
+    ignoreNamedArgument: true
+    ignoreEnums: false
```
malinajirka: 💡 We might want to consider ignoring magic numbers in enums. I think introducing an enum like this "VARIANT_A(1), VARIANT_B(2), VARIANT_C(3), ..." is fine and used in our codebase. Wdyt?

config/detekt/detekt.yml
```diff
@@ -0,0 +1,689 @@
...
+    excludedFunctions: 'equals'
+    excludeLabeled: false
+    excludeReturnFromLambda: true
+    excludeGuardClauses: false
+  SafeCast:
+    active: true
+  SerialVersionUIDInSerializableClass:
+    active: false
+  SpacingBetweenPackageAndImports:
+    active: false
```
malinajirka, line 629: 💡  We might want to consider enabling this rule.

config/detekt/detekt.yml
```diff
@@ -0,0 +1,689 @@
...
+  UnnecessaryInheritance:
+    active: true
+  UnnecessaryLet:
+    active: false
+  UnnecessaryParentheses:
+    active: false
+  UntilInsteadOfRangeTo:
+    active: false
+  UnusedImports:
+    active: false
```
malinajirka, line 654: 💡 We might want to enable all the UnsedXyz rules, wdyt?

malinajirka (commented):
Thank you for working on this @ParaskP7! 

I've tested the configuration and it seems to work as expected. I've left a couple of comments, let me know what you think.

P.S. I installed the IDE plugin and I created a method with 5 returns statements. The cmd task found the error, should the IDE plugin underline the method or somehow indicate an error?

build.gradle
ParaskP7: 👋 @malinajirka !

From the documentation what this does is: `Builds the AST in parallel. Rules are always executed in parallel. Can lead to speedups in larger projects. `false` by default.`

AST, to the best of my knowledge, means `Abstract Syntax Tree`, which seems to speed up Detekt. I am not sure if that is the reason why this is not the default in general. I choose to have this `false` for two reasons:
1) It is the default, just to be on the safe side for now,
1) Running `Detekt` is already very fast, I don't think we need to worry in making it faster for our project. On the other hand, I worry that by executing rules in parallel can end up buggy and not report or report more violations than normal. Thus, I would start with `false`.

PS: We can definitely turn that on to `true`, I don't think it will hurt. But, the default `false` seems safer for me as a starting point.

build.gradle
ParaskP7: 👋 @malinajirka !

Yes, I was thinking about that too. I choose to not disable any of the formats for two reasons:
1) The output files are very small anyway, thus there will not be a build overhead,
1) I don't know how and which format our CI will going to be using in the future, for reporting purposes, so I enabled all possible formats to make it easier for future integration.

PS: We can definitely have only one of those enabled if you like. Let me know which one you prefer. @loremattei

config/detekt/detekt.yml
ParaskP7: 👋 @malinajirka !

Yes, I totally agree, but I would still stick to the default atm, before enabling any additional rule. What we can do is create an issue with all the rules that we want to enable on the 2nd phase of Detekt and keep those suggestions there, alongside any explanation why. Wdyt?

config/detekt/detekt.yml
ParaskP7: Good catch @malinajirka !

Actually, this will not be doing anything in terms with `autoCorrect` since the global configuration within the root `build.gradle` is set to `false`:
```
detekt {
    ...
    autoCorrect = false
    ...
}
```

PS: I am not even sure what this `autoCorrect` is doing. Maybe this is used only when the global configuration is set to `true` to fine grain to child configuration in turning them to `false` to avoid formatting them. Think about it the other way around.

Question (❓): Should I turn this off just to be on the save side. Wdyt?

config/detekt/detekt.yml
ParaskP7: Another good catch @malinajirka !

Maybe we should indeed consider changing this to `true`. To my understanding this will apply additional formatting rule if the module is Android specific. Since `Detekt` is Kotlin specific, not only Android, they probably decided to have them separated like that.

Question (❓): Should I turn this on. Wdyt?

config/detekt/detekt.yml
ParaskP7: 👋 @malinajirka !

Yeap, I agree with both your statement, to avoid `else/default` branches, but also to start with the default for now. Let's document that in this issue that I suggested we create for Detekt's 2nd phase?

config/detekt/detekt.yml
ParaskP7: Oh yes, me too! 😄

config/detekt/detekt.yml
ParaskP7: Good catch @malinajirka !

We probably need to disable that for now. But, I would strongly suggest to utilise the `allowedPatterns` and enabled that in the near future. We should document that on the issue for Detekt's 2nd phase as well.

config/detekt/detekt.yml
ParaskP7: 👋 @malinajirka !

True, we can do that. Should we do it now or keep that for Detekt's 2nd phase?

config/detekt/detekt.yml
ParaskP7: Good catch @malinajirka !

You are probably right, we might want to ignore magic number in enums.

config/detekt/detekt.yml
ParaskP7: 👋 @malinajirka !

True, we can do that. Should we do it now or keep that for Detekt's 2nd phase?

config/detekt/detekt.yml
ParaskP7: 👋 @malinajirka !

True, we can do that. Should we do it now or keep that for Detekt's 2nd phase?

ParaskP7:
👋 @malinajirka !

> I've tested the configuration and it seems to work as expected. I've left a couple of comments, let me know what you think.

Thank you so much for this thorough review and all the comments you left Jirka, much much appreciated! 💪 

> P.S. I installed the IDE plugin and I created a method with 5 returns statements. The cmd task found the error, should the IDE plugin underline the method or somehow indicate an error?

The IDE plugin should highlight the whole line with yellow color and indicate that this is a warning. You want want to restart your IDE or sync it or the plugin might be misbehaving on its current version or due to the new Detekt version, not sure. However, I also noticed that my IDE wasn't highlighting those warnings for me too yesterday, but skipped the debugging for later.

config/detekt/detekt.yml
malinajirka: 👍 That's sounds like a great approach!

config/detekt/detekt.yml
malinajirka: Ahh, I see. I didn't realize there is also a global flag. I think it's fine to keep the current setup. Thanks for the clarification.

config/detekt/detekt.yml
malinajirka: I tried to turn it on and it didn't find any issues. Since I haven't found it in the docs, I'm wondering if it requires some extra "experimental" flag in the gradle. Wdyt?

config/detekt/detekt.yml
malinajirka: 👍 sounds good

config/detekt/detekt.yml
malinajirka: I'd just remove "TODO" from there, the others seems fine to me. Wdyt?

config/detekt/detekt.yml
malinajirka: Both is fine with me - this is technically not a new rule, since ktlint/lint is failing the build.

config/detekt/detekt.yml
malinajirka: 2nd phase SGTM

config/detekt/detekt.yml
malinajirka: 2nd phase SGTM

config/detekt/detekt.yml
ParaskP7: Thinking about it I believe it is better to have this as `false` as well and avoid any future confusion. 😅 I convinced you to convince me to convince you, and me back...

config/detekt/detekt.yml
ParaskP7: There is no extra `experimental` flag in gradle for that matter, I am pretty sure of it. 😄 

But, I do believe it is better if we switch it on, as per your suggestion, just to avoid any future confusion.

config/detekt/detekt.yml
ParaskP7: Yeap, I agree! 🙏

config/detekt/detekt.yml
ParaskP7: Based on the fact that we will someday decommission `KtLint` I think we do need to take this into consideration. But, on the other hand, I am sure that when that decommissioning `KtLint` day comes, we will do a thorough 1-to-1 rule checking to make sure we are not removing any rules that are already working fine with that tool.

Thanks for noticing! 🙏

config/detekt/detekt.yml
ParaskP7: This has been now documented within this [issue's comment](https://github.com/wordpress-mobile/WordPress-Android/issues/14011#issuecomment-775885924).

config/detekt/detekt.yml
ParaskP7: This has been now documented within this [issue's comment](https://github.com/wordpress-mobile/WordPress-Android/issues/14011#issuecomment-775885924).

config/detekt/detekt.yml
ParaskP7: This has been now documented within this [issue's comment](https://github.com/wordpress-mobile/WordPress-Android/issues/14011#issuecomment-775885924).

config/detekt/detekt.yml
ParaskP7: This has been now documented within this [issue's comment](https://github.com/wordpress-mobile/WordPress-Android/issues/14011#issuecomment-775885924).

config/detekt/detekt.yml
ParaskP7: This has been now documented within this [issue's comment](https://github.com/wordpress-mobile/WordPress-Android/issues/14011#issuecomment-775885924).

config/detekt/detekt.yml
ParaskP7: This has been now documented within this [issue's comment](https://github.com/wordpress-mobile/WordPress-Android/issues/14011#issuecomment-775885924).

config/detekt/detekt.yml
ParaskP7: This has been now documented within this [issue's comment](https://github.com/wordpress-mobile/WordPress-Android/issues/14011#issuecomment-775885924).

config/detekt/detekt.yml
ParaskP7: Also, this has been now documented within this [issue](https://github.com/wordpress-mobile/WordPress-Android/issues/14012).

config/detekt/detekt.yml
ParaskP7: This is now done here: 68e24e9e45c180b7dcaf1018b875a01af0ef013b

config/detekt/detekt.yml
ParaskP7: This is now done here: 9b1025f19e4afd325316665a0c19b7c5bb1fe9fb

config/detekt/detekt.yml
ParaskP7: Also, this is now done here: c77bf70f0b87af6faef1397fc616db96722c4a16

config/detekt/detekt.yml
ParaskP7: This is now done here: 56fdf264f151938cf5680e60304e36f765a8b779

ParaskP7:
👋 @malinajirka !

I just re-requested your review on this PR, could you please check out my responses per comment and any new commit that fixes what was described.

Many thanks in advance! 🙏

malinajirka (approved):
LGTM, thanks for the changes and for creating the issue for 2nd phase ;).

loremattei (approved):
Thank you for this @ParaskP7!

I tested following the steps you provided and it worked as expected, other than what @malinajirka already noted. 

> Detekt was configured on the root build.gradle file, within the allprojects { ... } scope. However, I have noticed that the subprojects { ... } is utilised as well for static analysis tools like KtLint. Do I need to add an additional task for Detekt within the subprojects { ... }?

I've never dug into how KtLint was configured, but looking at the Detekt documentation it seems to me that what you did is enough. 

> The exist a task within the WordPress build.gradle file, called violationCommentsToGitHub. This task has a violations configuration for CheckStyle. Do I need to add something similar for Detekt?

Yes, that task is called on CI and it adds a comment with the violations to the GH PR. It [looks like Detekt is supported as well](https://github.com/tomasbjerre/violation-comments-to-github-gradle-plugin). I _think_ that if we use the default output configuration for the Detekt report, it should pick it up automatically.

build.gradle
loremattei: I think `xml` is enough for CI as it's the one used by the [plugin we use to post comments on PRs](https://github.com/tomasbjerre/violation-comments-to-github-gradle-plugin). 👍

ParaskP7:
👋 @loremattei !

> I tested following the steps you provided and it worked as expected, other than what @malinajirka already noted.

Thank you so much for the review! 🙏 Yeap, @malinajirka was really thorough! ❤️ 

> I've never dug into how KtLint was configured, but looking at the Detekt documentation it seems to me that what you did is enough.

Thanks for double checking on that, I agree! 🙏 

> Yes, that task is called on CI and it adds a comment with the violations to the GH PR. It looks like Detekt is supported as well. I think that if we use the default output configuration for the Detekt report, it should pick it up automatically.

Coolio, let me check on that as well before merging that to `develop`. Or, question (❓), do you think we can merge it even without this `violations` configuration and add that one as a separate task afterwards?

build.gradle
ParaskP7: Many thanks for adding to that @loremattei !

So, since `xml` is required by CI, and IMHO `html` is a real nice to have locally, to get a nice representation of the issue at hand and how to fix it, I would have those two enabled. This might leave us with only `txt` to decide to disable, but even that I suggest to keep it enabled since I find myself sometimes looking at this find to figure what the problem more rapidly. I don't think those files are too much overhead and as such I would have the enabled (at least for now).

@loremattei @malinajirka your thoughts?

PS: I will be fine even if we decide to disable all of them and ending up having `xml` the only one being enabled.

ParaskP7:
👋 @loremattei !

> Coolio, let me check on that as well before merging that to develop.

I checked the tool ([violation-comments-to-github-gradle-plugin](https://github.com/tomasbjerre/violation-comments-to-github-gradle-plugin)) and maybe the below configuration is what we need to have `violation comments` work for both `Checkstyle` and `Detekt` now. Wdyt?

image

I'll try to push a commit with that configuration with both, some `CheckStyle` and/or `Detekt` issues, to see this `violation comments` setup work as expected.

build.gradle
malinajirka: If it doesn't take any significant amount of time, I agree that keeping all of them is fine.

ParaskP7:
👋 @loremattei !

I have been trying to update the `violations` configuration while having some issues with `Detekt` (7 of them), but I am not seeing any output. I am not familiar with this tool and how it outputs the violations. Can you maybe help with that?

PS: I am now going to create a separate `test` branch with a `CheckStyle` violation, then open a `test` PR and observe the output (if any). Will keep you posted. #14033
PSS: I also checked `Circle CI` and the `Violations` job and saw this below, maybe something is wrong, I am not sure:
```
#!/bin/bash -eo pipefail
if [ -n "$GITHUB_API_TOKEN" ]; then
  ./gradlew --stacktrace violationCommentsToGitHub -DGITHUB_PULLREQUESTID=${CIRCLE_PULL_REQUEST##*/} -DGITHUB_OAUTH2TOKEN=$GITHUB_API_TOKEN
else
  echo "Not posting lint errors to Github because \$GITHUB_API_TOKEN is not found"
fi
```

PR #13998, opened 2021-02-08
Lower min value of bar chart in stats to 4

### LINKED ISSUE #13671: Stat Graph Anomalies with low stats
### Expected behavior

The user expected to see more accurate representation of the numbers in the values shown in the graph. 

### Actual behavior

The difference between the value shown and the numbers received doesn't always equal the difference between the numbers trying to be reflected. 

They mention the following and share two screenshots:

> note two things:
> - the vertical scale is labelled 0-1-2-4-5. i don't know the reason
for this but it's extremely odd. i have never seen a graph or chart
where the scale changes then changes _back_.
> - the graph purports to show views. note that _every_ day with views
is a non-integer value -- it looks to me as if the ones that 'should'
be 1 hit are a little less -- to my eye around .8. what is .8 of a
view? or 1.6 (three of them look like that)? the last one i can't
guess because i don't want to guess what's going on in the 2-4 part of
the graph.

Image 2020-12-23 at 3 30 07 pm

Image 2020-12-23 at 3 32 42 pm

To reproduce this, view stats for a site that doesn't have a lot of stats and check where the graph end bar lands compared to the actual 1 on the graph. Note the scale. 

##### Tested on Samsung S9+, Android 10, WPAndroid 16.3-rc-2

User reported on Pixel 4a in  3594277-zen

planarvoid:
Fixes #13671 

As discussed in the issue, this PR solves the anomaly in the stats Day/Week/Month/Year graph with very low data by lowering the minimal top value of the graph to 4. This way each row in the graph represents one whole item.

To test:
- Check stats for a site with very low views
- Notice the Left axis now says 0, 1, 2, 3, 4 where 4 is the top value and no number there is missing
- Sanity check the other graphs with more data

Screenshot_1612798529

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/BarChartViewHolder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/BarChartViewHolder.kt
@@ -27,7 +27,7 @@ import org.wordpress.android.ui.stats.refresh.utils.LargeValueFormatter
 import org.wordpress.android.util.DisplayUtils
 
 private const val MIN_COLUMN_COUNT = 5
-private const val MIN_VALUE = 5f
+private const val MIN_VALUE = 4f
 
 private typealias BarCount = Int
 


ParaskP7 (approved):
👋 @planarvoid !

Actually my side had two views when looking at the `INSIGHTS` tab and that wasn't showing correctly with `MIN_VALUE = 5f`. When that was update to `MIN_VALUE = 4f` then everything works as expected, the UI was showing correctly.

PS: Also other sites with more stats where showing as expected as well.

PR #14000, opened 2021-02-08
Show snackbar in activity context

### LINKED ISSUE #13853: Snackbar appears behind FAB on the improved My Site fragment

planarvoid:
Fixes #13853 

This PR changes the target view of the snackbar from the fragment to the main activity coordinator in order to show the snackbar under the FAB.

To test:
- Make sure the MySiteImprovements flag is turned on and the app is restarted
- Make sure you're on a site with QS active
- Click on a task
- Notice the snackbar is shown and the FAB is moved up

Screenshot_1612803991

PR submission checklist:


renanferrari (approved):
Thank you! This seems to be working fine and the code changes are looking good.

The only thing that bothers me is that there seems to be an unnecessary background behind the Snackbar. I knew this was not an issue with this PR specifically, but I did some digging and it seems to me that once we remove the My Site feature flag, we might be able to make some changes to the `main_activity.xml` layout so it can properly use a `CoordinatorLayout`, which would probably allow us to fix that. I've opened an issue for it here: #14003

LGTM :shipit:

PR #14002, opened 2021-02-08
Add backoff delay

zwarm:
Parent #13328 & #13329

This PR adds a backoff strategy to `GetBackupDownloadStatusUseCase` and `GetRestoreStatusUseCase` when retrying because of a failed fetch status. The strategy implemented starts with the normal delay and ups it by a factor of 2 for each subsequent retry.

I also did a small POC on using flowRetry, but I couldn't figure out how to reset the retryAttempts on a successful fetch in that scenario. I chose to submit this solution and keep plugging away at flow/retry separately.

To test:
- Run the backup download/restore process to end and ensure it still works as expected.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCase.kt
```diff
@@ -24,29 +24,34 @@ import javax.inject.Named
...
         while (true) {
             if (!isNetworkAvailable()) return@flow
 
             val result = activityLogStore.fetchBackupDownloadState(FetchBackupDownloadStatePayload(site))
             if (result.isError) {
-                if (retryAttempts++ >= MAX_RETRY) {
-                    AppLog.d(T.JETPACK_BACKUP, "$tag Exceeded 3 retries while fetching status")
-                    emit(RemoteRequestFailure)
-                    return@flow
+                when (exceedsRetryAttempts(retryAttempts++)) {
```
malinajirka: 🔍 Nitpick: We might want to use `if` instead of `when` here. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCase.kt
```diff
@@ -24,29 +24,34 @@ import javax.inject.Named
...
+                    true -> return@flow
+                    false -> {
+                        delay(backoffDelay)
+                        backoffDelay = (backoffDelay * DELAY_FACTOR)
```
malinajirka: We never clear the backoffDelay, so it might grow indefinitely. Eg. first three attempts fail -> delay is 4s -> one succeeds -> 3 fail and delay is 16s -> etc.

We might want to consider removing "backoffDelay" variable altogether and use something like `delay(DELAY_MILLIS * (max(1, DELAY_FACTOR * retryAttempts)))`. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCase.kt
```diff
@@ -79,4 +84,12 @@ class GetBackupDownloadStatusUseCase @Inject constructor(
             false
         }
     }
+
+    private suspend fun FlowCollector<BackupDownloadRequestState>.exceedsRetryAttempts(retryAttempts: Int): Boolean {
+        return if (retryAttempts >= MAX_RETRY) {
+            AppLog.d(T.JETPACK_BACKUP, "$tag: Exceeded $MAX_RETRY retries while fetching status")
+            emit(RemoteRequestFailure)
```
malinajirka, line 85: 💡 It feels a bit unexpected that a method called `exceedsRetryAttempts` emits a value. We might want to consider renaming it and moving even the delay here. Something like this:

```
private suspend fun FlowCollector<BackupDownloadRequestState>.handleError(retryAttempts: Int): Boolean {
        return if (retryAttempts >= MAX_RETRY) {
            AppLog.d(T.JETPACK_BACKUP, "$tag: Exceeded $MAX_RETRY retries while fetching status")
            emit(RemoteRequestFailure)
            true
        } else {
            delay(DELAY_MILLIS * (max(1, DELAY_FACTOR * retryAttempts)))
            false
        }
    }
```

Wdyt?

malinajirka (changes requested):
Thanks @zwarm! I've left a couple of in-code comments. It'd be also great if we introduced unit tests for this new behavior. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCase.kt
zwarm: Makes perfect sense. 👍

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCase.kt
zwarm: I started with if and moved to when 🤦 , but will changed it back now that the delay is handled in the fun.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCase.kt
zwarm: I originally named the fun `handleError`, but felt the name didn't invoke a question that matched with a boolean return value. `exceedsRetryAttempts` feels like a question that would return a true/false value. The reverse can be said with emitting a value with the name `handleError` vs. `exceedsRetryAttempts`. So I followed the `isNetworkAvailable` pattern here .. ask question, emit if needed, return boolean.

I agree with moving the backoff logic here.

zwarm: Renamed in 83916f7

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCase.kt
zwarm: Updated in 83916f7

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCase.kt
zwarm: Updated in 83916f7

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCase.kt
malinajirka: > I originally named the fun handleError, but felt the name didn't invoke a question that matched with a boolean return value

I see, but I think it's a common pattern on Android to return "true" when an event was handled and "false" when it wasn't handled. If we wanted to follow the Android pattern even more we could call it "OnErrorReceived" but I don't think it's necessary, since this is not a system/user event.

zwarm: @malinajirka - That's fine, I had already renamed it, just wanted you to know why I had the original name.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCaseTest.kt
```diff
@@ -169,6 +169,57 @@ class GetBackupDownloadStatusUseCaseTest : BaseUnitTest() {
...
+                                            BackupDownloadStatusError(GENERIC_ERROR),
+                                            FETCH_BACKUP_DOWNLOAD_STATE
+                            ))
+                            .thenReturn(
+                                    OnBackupDownloadStatusFetched(
+                                            BackupDownloadStatusError(GENERIC_ERROR),
+                                            FETCH_BACKUP_DOWNLOAD_STATE
+                            ))
+
+                val result = useCase.getBackupDownloadStatus(site, downloadId).toList()
```
malinajirka, line 181: ⚠️ 
1. `runBlockingTests` invokes `dispatcher.advanceUntilIdle()` internally, so it skips all delays.
2. `toList()` method on Flow suspends the current thread until the Flow completes -> the `advanceTimeBy` doesn't have any effect here.

We can discuss this during the call. It's quite tricky.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCaseTest.kt
```diff
@@ -169,6 +169,57 @@ class GetBackupDownloadStatusUseCaseTest : BaseUnitTest() {
...
+                advanceTimeBy(DELAY_MILLIS)
+
+                assertThat(result).contains(RemoteRequestFailure)
```
malinajirka: 💡 Idea: To make this test even more robust, we might want to consider using 
`assertThat(result).isEqualTo(listOf(RemoteRequestFailure))` to make sure the usecase doesn't emit anything else.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCaseTest.kt
```diff
@@ -169,6 +169,57 @@ class GetBackupDownloadStatusUseCaseTest : BaseUnitTest() {
...
+                                OnBackupDownloadStatusFetched(
+                                        BackupDownloadStatusError(GENERIC_ERROR),
+                                        FETCH_BACKUP_DOWNLOAD_STATE
+                                ))
+                        .thenReturn(OnBackupDownloadStatusFetched(FETCH_BACKUP_DOWNLOAD_STATE))
+
+                val result = useCase.getBackupDownloadStatus(site, downloadId).toList()
+                advanceTimeBy(DELAY_MILLIS)
+
+                assertThat(result).contains(progressStatus, completeStatus)
```
malinajirka: 💡 Idea: To make this test even more robust, we might want to consider using
assertThat(result).isEqualTo(listOf(progressStatus, completeStatus)) to make sure the usecase doesn't emit anything else. Wdyt?

malinajirka (approved):
Thanks @zwarm! I've tested the flows and it works as expected. I've left some comments, but feel free to merge this PR.

I plan to discuss the comment marked with ⚠️ during our call today - this wasn't introduced in this PR so I don't think we need to block it.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCaseTest.kt
zwarm: Made the change in 1983a4a. Better than using contains. Thanks

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCaseTest.kt
zwarm: Made the change in 1983a4a. Better than using contains. Thanks

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/GetBackupDownloadStatusUseCaseTest.kt
zwarm: Based on our discussions, commit 1983a4a
(1) Replaced `MainCoroutineScopeRemove` with `test`
(2) Removed `advanceTimeBy`, as it wasn't needed

malinajirka: Nice, thanks!

malinajirka (approved):

PR #14009, opened 2021-02-09
Replace flow with live data in sources

planarvoid:
I've been looking a bit into the flows/livedata and I think it doesn't make sense to build flows and call asLiveData. I think it also quite lowers the performance and doesn't improve readibility. I've changed the interface in this PR so that the `buildSource` method now returns live data. The only problem with this approach is that it stops us from running slower operations (flow runs as a suspend function by default). I've fixed this by adding coroutine scope as the parameter to the build source. I think the result is quite neat and more readable tbh. Let me know what you think! 

To test:
- Make sure you have the my site improvements flag turned on and the app is restarted
- Click around the my site to check if everything looks ok
- Switch between a few sites to see if the screen gets updated.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/mysite/CurrentAvatarSource.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/mysite/CurrentAvatarSource.kt
@@ -1,10 +1,9 @@
 package org.wordpress.android.ui.mysite
 
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MediatorLiveData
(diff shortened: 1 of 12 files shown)


renanferrari (approved):
Thank you for the changes! I've tested the app and everything seems to be working as expected. 

I also agree that the solution is pretty good and that the code is more readable now and although I feel like we can still simplify some of these sources a bit further, I'm gonna go ahead and merge this PR since there's nothing wrong with it and those optimizations can be done separately.

LGTM :shipit:

Issue #4131, opened 2016-05-30
Gradient is black on the Country view in Stats


maxme:
![4372b28a2c5069cb69395d9f39f076591080a971](https://cloud.githubusercontent.com/assets/40213/15642556/6386a358-2648-11e6-8824-5edf4231eeab.png)

maxme:
addressed by #4172

PR #14013, opened 2021-02-09
Update geochart version to the latest stable

### LINKED ISSUE #13637: Stats: update GeoChart to resolve map border issues
A user reported that the borders for Morrocco in the stats map were not up to date and were missing content of the Sahara. They mentioned that Google had updated their maps to reflect this which seems to be the case on Google Maps.
On looking at the [GeoChart map](https://developers.google.com/chart/interactive/docs/gallery/geochart#Example) which the app uses, it doesn't look like the difference is showing there. 

On checking, @frosty noticed that the app doesn't use the most recent version of GeoChart, so it may be worth looking if it can be updated or not for cases where Google made some updates which the app isn't reflecting.

Internal discussion: p1608298268293900-slack-C011BKNU1

planarvoid:
Fixes #13637 

We were using an old version of geochart on purpose ([original issue from 2016](https://github.com/wordpress-mobile/WordPress-Android/issues/4131), fix in this [PR](https://github.com/wordpress-mobile/WordPress-Android/issues/4131)). However, I'm not seeing the missing label issue so it might have been fixed already by google (the testing steps in the original issue just say to open stats/period and check the countries block).  

To test:
- Open Stats on a site with some views
- Check one of Day/Week/Month/Year tab that has countries block
- Scroll to the countries block
- Notice the legend is present
- Notice the map looks OK

Screenshot 2021-02-09 at 13 37 09
Screenshot 2021-02-09 at 13 36 04

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/MapViewHolder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/MapViewHolder.kt
@@ -31,10 +31,6 @@ class MapViewHolder(parent: ViewGroup) : BlockListItemViewHolder(
     fun bind(item: MapItem) {
         coroutineScope.launch {
             delay(100)
-            // See: https://developers.google.com/chart/interactive/docs/gallery/geochart
-            // Loading the v42 of the Google Charts API, since the latest stable version has a problem with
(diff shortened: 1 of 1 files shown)


renanferrari (approved):
Thanks for the changes. The code looks good and the app is working as expected.

LGTM :shipit:

PR #14017, opened 2021-02-09
Remove APKs from GH releases

loremattei:
Currently, the universal APKs we upload to GH releases are unsigned, so they are not easily installable. 
We already upload the app bundle from which a valid APK can be generated with the proper key, so there's no advantage in having the unsigned APK as well. 
The standard expectation is for people to be able to install the APK directly, so the unsigned APK generates confusion. 

This PR disables the upload of the APK from CI. 
We may evaluate sharing a replaceable key on CI for signing the APKs in the future if there will be a need for it.

PR submission checklist:

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -714,11 +714,13 @@ ENV["validate_translations"]="lintVanillaRelease"
     version = options[:version]
     apk_file_path = universal_apk_file_path(version)
     aab_file_path = bundle_file_path(version)
+    # APKs built on CI are unsigned, so don't upload them until we have a fix
+    release_assets = is_ci ? aab_file_path : "#{apk_file_path},#{aab_file_path}" 
     create_release(repository:GHHELPER_REPO, 
       version: version["name"], 
       release_notes_file_path:release_notes_path,
       prerelease: set_prerelease_flag,
-      release_assets:"#{apk_file_path},#{aab_file_path}"
+      release_assets: release_assets
     )
   end
 
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -715,7 +715,8 @@ ENV["validate_translations"]="lintVanillaRelease"
     apk_file_path = universal_apk_file_path(version)
     aab_file_path = bundle_file_path(version)
     # APKs built on CI are unsigned, so don't upload them until we have a fix
-    release_assets = is_ci ? aab_file_path : "#{apk_file_path},#{aab_file_path}" 
+    release_assets = [aab_file_path]
+    release_assets.append(apk_file_path) unless is_ci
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -714,11 +714,13 @@ ENV["validate_translations"]="lintVanillaRelease"
     version = options[:version]
     apk_file_path = universal_apk_file_path(version)
     aab_file_path = bundle_file_path(version)
+    # APKs built on CI are unsigned, so don't upload them until we have a fix
+    release_assets = is_ci ? aab_file_path : "#{apk_file_path},#{aab_file_path}"
```
AliSoftware: I'm unsure if it makes sense to have a different behavior depending on if we create the release from CI vs a local machine tho, as that might mean that some releases (created locally) would have the (signed) apk but some (created from CI) would not, which… might create as much confusion as having the unsigned apk.

What do you think about always just uploading the aab for consistency?

loremattei: If we create the release on a local machine, the APK is signed, so it can be downloaded and installed without any additional step. I think that if we have a valid APK, it's worth it to upload it. 
I don't think it would add much confusion: we may still continue to manually upload some APKs for some testing purposes anyway, so we'll likely have some releases with the APK, and others without. 
While this is not perfect, I think that as long as the APK you download is installable, there shouldn't be big issues.

AliSoftware: nit (part 1):

```suggestion
    release_assets = [aab_file_path]
    release_assets.append(apk_file_path) unless is_ci
```

fastlane/Fastfile
```diff
@@ -714,11 +714,13 @@ ENV["validate_translations"]="lintVanillaRelease"
...
     apk_file_path = universal_apk_file_path(version)
     aab_file_path = bundle_file_path(version)
+    # APKs built on CI are unsigned, so don't upload them until we have a fix
+    release_assets = is_ci ? aab_file_path : "#{apk_file_path},#{aab_file_path}" 
     create_release(repository:GHHELPER_REPO, 
       version: version["name"], 
       release_notes_file_path:release_notes_path,
       prerelease: set_prerelease_flag,
-      release_assets:"#{apk_file_path},#{aab_file_path}"
+      release_assets: release_assets
```
AliSoftware: nit (part 2)

```suggestion
      release_assets: release_assets.join(',')
```

AliSoftware (approved):
Right. SGTM then!

Added a nit in the way to write the logic in a more idiomatic ruby, but feel free to ignore and merge as is, that works too!

loremattei:
Thanks @AliSoftware! I committed your suggestions! 🙇

PR #14020, opened 2021-02-09
Fix "Edit your homepage" Quick Start task completion

### LINKED ISSUE #14016: Make sure "Edit homepage" task completes
We're now calling `quickStartRepository.requestNextStepOfTask(EDIT_HOMEPAGE)` when leaving the my site to the pages list. This call sets the activeTask to null. In the pages list we try to call `completeTask` which requires the active task to be the same as the task we're trying to complete. This causes the QS task to never be completed.

renanferrari:
Fixes #14016

This PR basically updates the `QuickStartRepository::requestNextStepOfTask` method so that it doesn't clear the currently active task when called. The reason it used to do that, was because it was originally implemented to be used in conjunction with the `QuickStartUtils::completeTaskAndRemindNextOne` method. However, if we tried to use it in conjunction with `QuickStartRepository::completeTask` it would not work, as that method needs the active task to not be `null`, which was not very intuitive.

One problem with not clearing the currently active task is that if the user doesn't complete the next step of the task and comes back to the My Site fragment, they would still see the current active task focus point, which is not desirable. To fix that, this PR also makes sure to clear the active task inside `QuickStartRepository::refresh` which is always called when the My Site fragment is resumed. This also prevents a focus point from incorrectly appearing after switching sites. Let me know what you think! 

### To test

1. Go to the My Site screen and select a site with a "Homepage" where Quick Start is active.
1. Under the "Customize your site" category, tap the "Edit your homepage" task.
1. Notice that a focus point appears over the Pages button.
1. Tap on the Pages button.
1. On the Pages screen, notice the focus point over the "Homepage" item.
1. Without clicking anything, go back to the previous screen.
1. On the My Site screen, notice that the focus point has disappeared.
1. Scroll to the "Customize your site" category.
1. Notice that the "Edit your homepage" task is still uncompleted. Tap it again.
1. Repeat steps 4, 5 and 6.
1. On the Pages screen, tap the "Homepage" item.
1. Go back to the My Site screen and scroll to the "Customize your site" category.
1. Notice that the "Edit your homepage" task is now completed.

Also make sure all other tasks are still working as expected.

---

PR submission checklist:


planarvoid:
thanks for the PR @renanferrari . There is one side effect of this PR. The refresh in onResume clears the active task for example on orientation change. I'm wondering if that's acceptable. I've checked the previous version and the focus point doesn't disappear there. I think the desired behaviour would be to keep the focus point when you're on the my site screen (even through rotation/pause/resume) and hide it when you leave the screen. One solution I could think of is introducing another field - "task in progress". Leave the active task as it was but save the "task in progress" when you call `requestNextStepOfTask`. In `completeTask` we'd check if the task is active (== the focus point is visible) or if the task is in progress to complete the task. The task in progress would be cleared active task is set.

renanferrari:
@planarvoid Thanks for the review! I've implemented your suggestion and added a `pendingTask` variable to `QuickStartRepository` to control which task we expect to complete as a next step. 

One thing missing from this solution is that it would only hide the focus point when leaving the screen to complete the next step of a task. If we tapped the Notifications tab and then come back to My Site for example, I'd expect the focus point to be gone (this is the previous behavior we had).

To fix this, I've also added a `QuickStartRepository::clearActiveTask` that is only called from the fragment's `onPause`, but only if there's no configuration changing. This is more or less what [we were doing previously](https://github.com/wordpress-mobile/WordPress-Android/blob/93ebe4fe6dce242db01811fc81e9198da48bd43b/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt#L1525). Let me know what you think.

planarvoid (approved):
Thanks for the changes, it's looking good and working well 👍 Feel free to merge it once the conflicts are resolved

PR #14029, opened 2021-02-10
Use Activity context to style QS snackbar text color with an attr color

### LINKED ISSUE #14015: The bold text in a QS snackbar is white on white background
When in the dark mode the bold text shown in the snackbar is white on light grey background and that is not very readable. We need to change the color to a darker shade.

Screenshot 2021-02-09 at 14 54 50

planarvoid:
Fixes #14015 

We were previously using fixed white color to stylize the QS snackbars. This PR replaces the color with attr colorSurface which works well both in light and dark modes. In order to obtain it we need to pass the activity context into the the utils so I've moved the logic around a bit. 

To test:
- Make sure you're on a site with QuickStart in progress
- Turn on the light mode
- Click on a few of the QS tasks
- Notice the snackbar message is consistently in white (light gray) color and the background is dark
- Turn on the dark mode
- Click on a few of the QS tasks
- Notice the snackbar message is consistently in black (dark gray) color and the background is light

Screenshot 2021-02-10 at 13 24 18
Screenshot 2021-02-10 at 13 21 42

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/util/QuickStartUtils.kt
+++ b/WordPress/src/main/java/org/wordpress/android/util/QuickStartUtils.kt
@@ -97,7 +97,7 @@ class QuickStartUtils {
             )
             // nothing to highlight
             if (startOfHighlight != -1 && endOfHighlight != -1) {
-                val highlightColor = resourceProvider.getAttr(activityContext, R.attr.colorSurface)
+                val highlightColor = activityContext.getColorFromAttribute(R.attr.colorSurface)
                 mutableSpannedMessage.setSpan(
                         ForegroundColorSpan(highlightColor),
                         startOfHighlight, endOfHighlight, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE
--- a/WordPress/src/main/java/org/wordpress/android/util/QuickStartUtils.kt
+++ b/WordPress/src/main/java/org/wordpress/android/util/QuickStartUtils.kt
@@ -15,7 +15,6 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.view.ViewGroup.MarginLayoutParams
-import androidx.core.content.ContextCompat
 import androidx.core.graphics.drawable.DrawableCompat
 import androidx.core.text.HtmlCompat
 import org.wordpress.android.R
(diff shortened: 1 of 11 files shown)


WordPress/src/main/java/org/wordpress/android/util/QuickStartUtils.kt
```diff
@@ -67,6 +67,7 @@ class QuickStartUtils {
         @JvmOverloads
         fun stylizeQuickStartPrompt(
             resourceProvider: ResourceProvider,
+            activityContext: Context,
```
renanferrari: Since we're passing a Context here, I think we could probably replace the `ResourceProvider` usages with it?

WordPress/src/main/java/org/wordpress/android/util/QuickStartUtils.kt
```diff
@@ -96,7 +97,7 @@ class QuickStartUtils {
             )
             // nothing to highlight
             if (startOfHighlight != -1 && endOfHighlight != -1) {
-                val highlightColor = resourceProvider.getColor(android.R.color.white)
+                val highlightColor = resourceProvider.getAttr(activityContext, R.attr.colorSurface)
```
renanferrari: And then here we could probably use [this](https://github.com/wordpress-mobile/WordPress-Android/blob/635536d1190c49f40d9a00995440af056e391b13/WordPress/src/main/java/org/wordpress/android/util/ContextExtensions.kt#L22-L27) instead:

```suggestion
                val highlightColor = activityContext.getColorFromAttribute(R.attr.colorSurface)
```

Edit: btw, it seems that `highlightColor` is supposed to be a color int, but `getAttr` returns a color id. If you decide to keep the `getAttr` method, you should probably change it and annotate it with `@ColorInt`.

renanferrari (commented):
Thanks for the changes! Everything seems to be working as expected. Code looks good, but I left a couple of comments below.

Edit: I just realized that the Snackbars don't look right on some other screens:

| Pages Screen | Sharing Screen |
|---|---|
|||

WordPress/src/main/java/org/wordpress/android/util/QuickStartUtils.kt
planarvoid: Fixed in 64ab82e

planarvoid:
Good catch @renanferrari. I didn't expect the other dialogs to use WPDialogSnackbar and not WPSnackbar. I didn't know WPDialogSnackbar even exists. It should be refactored since it doesn't use correct styles. However, in this PR I've removed the changed the behaviour so that we have 2 methods, one is themed and called from the Improved my site and the other one behaves as it did before so the behaviour stays as it was.

renanferrari (approved):
LGTM! :shipit:

PR #14044, opened 2021-02-12
P2 people invite links

develric:
Part of #13456

This PR introduce possibility for P2 sites to create/share/disable invite links for the roles on their site.

<p align=center><image src="https://user-images.githubusercontent.com/47797566/107771650-59155380-6d3b-11eb-9950-855ec831f29a.gif" width=300/></p>

Hey @mattmiklic wanted to check with you couple of deviations from original, let me know if it's ok (and as always any other feedback welcome 🙇 ):

- As shown in the gif above, introduced loading views while transitioning between states
- As shown below, I used for the Expire the date format we have already in the People screen

| image | image |
|---|---|

Note: a dedicated PR is on the way for analytics and unit testing.

## To test
First of all check the Invite Link section is only available for P2 sites, then use a P2 site for below tests

### Generating links
- On a P2 site with no invitation links already generated tap on Generate new link
- Check the links are created and the roles are all the roles available on the web for that site
- Check the web reports those links now
- Using the app share link button copy the links and check they are consistent with those available on the web for every role

### Sharing an invite link
- Check you can share every role link and copy it or open it in one of the apps available in the share bottom sheet
- Check the shared link is consistent
- Try to use it with another user and see the link allows you to join the site in the expected role (check you are also able to join as administrator)

### Disabling links
- Keep a copy of an invite link
- Disable the links and check you are landed again in the screen with Generate new link button
- Try to use the previously copied link and check it's no more valid

### Snackbars
- Try to make an action for invite links with flight mode on (for example try to generate or delete links)
- Check you get a snackbar to notify the user about the issue

### Smoke test
Smoke test the Invite People screen overall, including the already present people invitation mechanism

PR submission checklist:


mattmiklic (approved):
Everything is looking good here! I tried creating a link, sharing it, and disabling it and didn't find any visual issues. Both light and dark mode look good. 🚀

WordPress/src/main/java/org/wordpress/android/ui/people/InviteLinksUseCase.kt
```diff
@@ -0,0 +1,142 @@
...
+
+        data class InviteLinksError(
+            override val scenarioContext: UseCaseScenarioContext,
+            val error: UiString
+        ) : InviteLinksState()
+    }
+
+    companion object {
+        // Pretty arbitrary amount to allow the progress bar to appear to the user
+        private const val PROGRESS_DELAY_MS = 600L
```
khaykov, line 140: Nice touch!

WordPress/src/main/java/org/wordpress/android/ui/people/InviteLinksUseCase.kt
```diff
@@ -0,0 +1,142 @@
...
+    ) {
+        if (!accountStore.hasAccessToken()) {
+            flow.emit(UserNotAuthenticated)
+        } else {
+            flow.emit(InviteLinksLoading(scenarioContext))
+            delay(PROGRESS_DELAY_MS)
+            if (!networkUtilsWrapper.isNetworkAvailable()) {
+                flow.emit(InviteLinksError(scenarioContext, UiStringRes(string.error_network_connection)))
+            } else {
+                strategy.invoke(flow, blogId, scenarioContext)
```
khaykov, line 63: Love the strategy + context approach!

WordPress/src/main/java/org/wordpress/android/ui/people/PeopleInviteFragment.java
```diff
@@ -158,16 +191,132 @@ public View onCreateView(final LayoutInflater inflater, ViewGroup container, Bun
...
+        mShareLinksButton = rootView.findViewById(R.id.share_links);
+        mLinksRoleTextView = rootView.findViewById(R.id.links_role);
+        mLinksRoleContainer = rootView.findViewById(R.id.links_role_container);
+        mDisableLinksButton = rootView.findViewById(R.id.disable_button);
+        mExpireDateTextView = rootView.findViewById(R.id.expire_date);
+
         return rootView;
     }
 
+    @Override public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceState) {
```
khaykov: Tiny nitpick - can we put `@Override` at the new line above the method?

WordPress/src/main/res/values/strings.xml
```diff
@@ -2271,6 +2272,21 @@
     <string name="role_follower">Follower</string>
     <string name="role_viewer">Viewer</string>
 
+    <string name="invite_links_section_title">Invite Link</string>
+    <string name="invite_links_get_status_retry">Get links retry</string>
```
khaykov: Maybe we could change this one to something a bit more user-friendly, eg. "Refresh link status"?

WordPress/src/main/java/org/wordpress/android/ui/people/PeopleInviteFragment.java
```diff
@@ -158,16 +191,132 @@ public View onCreateView(final LayoutInflater inflater, ViewGroup container, Bun
...
+        mShareLinksButton.setOnClickListener(v -> {
+            mViewModel.onShareButtonClicked(
+                    mLinksRoleTextView.getText() != null
+                            ? mLinksRoleTextView.getText().toString()
+                            : ""
+            );
+        });
+
+        mDisableLinksButton.setOnClickListener(v -> {
+            if (!isAdded()) return;
```
khaykov, line 231: Curious if there is a particular reason we have `isAdded()` check for this button, but not for other ones?

WordPress/src/main/java/org/wordpress/android/ui/people/PeopleInviteFragment.java
```diff
@@ -158,16 +191,132 @@ public View onCreateView(final LayoutInflater inflater, ViewGroup container, Bun
...
+                    return Unit.INSTANCE;
+                })
+        );
+
+        mViewModel.getInviteLinksUiState().observe(getViewLifecycleOwner(), uiState -> {
+            manageLinksControlsVisibility(uiState);
+
+            manageShimmerSection(uiState.isShimmerSectionVisible(), uiState.getStartShimmer());
+
+            manageActionButtonsEnable(uiState.getEnableActionButtons());
```
khaykov: Just to be on a safe side of clarity, what do you think about modifying the name of `manageActionButtonsEnable` to `manageActionButtonsEnabledState` and `getEnableActionButtons` to `isEveryActionButtonEnabled`?

WordPress/src/main/java/org/wordpress/android/models/InvitePeopleUtils.kt
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+
+    fun getDisplayNameForRole(
+        siteModel: SiteModel,
+        roleName: String
+    ): String {
+        val roles = getInviteRoles(siteStore, siteModel, contextProvider.getContext())
+
+        return roles?.firstOrNull { roleModel ->
```
khaykov: Looks like we don't need `?.` here since `getInviteRoles` is non-nullable.

WordPress/src/main/java/org/wordpress/android/ui/people/InviteLinksApiCallsProvider.kt
```diff
@@ -0,0 +1,142 @@
...
+            }
+        } ?: Failure(contextProvider.getContext().getString(string.invite_links_null_response))
+    }
+
+    private fun getErrorStringAndLog(
+        functionName: String,
+        blogId: Long,
+        volleyError: VolleyError?
+    ): String {
+        var error = VolleyUtils.errStringFromVolleyError(volleyError)
```
khaykov: I think we can change the `var` to `val` here.

WordPress/src/main/java/org/wordpress/android/ui/people/PeopleInviteDialogFragment.kt
```diff
@@ -0,0 +1,101 @@
...
+import androidx.lifecycle.ViewModelStoreOwner
+import com.google.android.material.dialog.MaterialAlertDialogBuilder
+import org.wordpress.android.R
+import org.wordpress.android.WordPress
+import org.wordpress.android.ui.people.PeopleInviteDialogFragment.DialogMode.DISABLE_INVITE_LINKS_CONFIRMATION
+import org.wordpress.android.ui.people.PeopleInviteDialogFragment.DialogMode.INVITE_LINKS_ROLE_SELECTION
+import org.wordpress.android.viewmodel.ContextProvider
+import javax.inject.Inject
+
+class PeopleInviteDialogFragment : DialogFragment() {
```
khaykov, line 28: Since this is a multipurpose dialog, maybe we should add a comment explaining what it does.

khaykov (approved):
This looks really good, @develric ! And works really smooth too! Apart from the issue with the JSON returned from endpoint, I didn't encounter anything critical.

I left a couple of comments related to naming and other minor things.

WordPress/src/main/java/org/wordpress/android/ui/people/PeopleInviteFragment.java
develric: No good reasons 😄 , changed.

develric:
Hi @khaykov, thanks for the review 🙇 ! I made changes based on your feedbacks. Ready for a last check 😄 

PS: Eric has deployed a fix for the JSON issue into the endpoint so hopefully we should be safe there now 👍

khaykov (approved):
Thanks for the changes! Looks great 👍

PR #14049, opened 2021-02-12
Issue/13326 threat details builder tests

ashiagr:
Parent #13326

This PR refactors threat details builder tests: 

- Makes them more granular by splitting into multiple tests.
- Reformats test statements into `given/when/then `format.

PR submission checklist:


WordPress/src/test/java/org/wordpress/android/ui/jetpack/scan/details/ThreatDetailsListItemsBuilderTest.kt
```diff
@@ -93,182 +92,319 @@ class ThreatDetailsListItemsBuilderTest : BaseUnitTest() {
...
+
     @Test
-    fun `builds fix details items correctly for a fixable threat`() {
+    fun `given fixable threat not in fixed status, when items are built, then fix header exists`() {
         // Arrange
         val fixableThreat = GenericThreatModel(
             ThreatTestData.baseThreatModel.copy(
-                fixable = Fixable(file = null, fixer = Fixable.FixType.EDIT, target = null),
-                status = ThreatStatus.CURRENT
+                fixable = Fixable(file = null, fixer = Fixable.FixType.EDIT, target = null)
```
ashiagr, line 292: Let me know if we should add separate tests for each `FixType` for this and the following test for fix header and description items.

malinajirka: It's difficult for me to evaluate that. It really depends on how much value do you think it'd add and if it takes 5 minutes or 5 hours. :) Use your best judgment.

malinajirka (approved):
Thanks @ashiagr! Really great improvements, it's much easier to read the tests now ❤️ !

I tried to remove a line from the class and a test failed. It was very easy to know what is wrong from the name of the test! I think this is already really good. 

One thing which feels a bit unusual is that all the tests are testing if something is added (positive) but none of them is testing if something is not added (negative). I think it's a good practice to try to have a negative test for each positive test.
For example:
- We have `given fixable threat in current status, when items are built, then ignore threat button exists`
- I'd consider adding `given fixable threat not in current status, when items are built, then ignore threat button doesn't exists`.

I think having tests in pairs positive-negative can bring a good value for a very little effort - it's usually just a copy paste and update of a single field + name of the test. (the effort is little if it's added when you write the tests, it might not be little if you want to add such tests retrospectively)

Without negative tests, I can change the logic of the class quite a bit and all the tests still pass:
Screenshot 2021-02-12 at 11 56 20

I realize you spent quite some time on these tests already and it might not be worth improving them even more. I just wanted to share this in case you'd find it useful for future. Feel free to merge this PR and thank you so much for all these improvements!

ashiagr:
Thanks for the review @malinajirka! Appreciate your time for reviewing all tests 🙏. 

> that all the tests are testing if something is added (positive) but none of them is testing if something is not added (negative). I think it's a good practice to try to have a negative test for each positive test.

I'll add at least for some of the cases I think might be good to have. I'll keep this PR open for few more mins.

malinajirka (approved):
Awesome, thank you, it looks perfect now 👏

Issue #13267, opened 2020-10-30
Jetpack Section: Project Issue


malinajirka:
This is the main project issue for the Jetpack Section project

### Tasks
- [x] Create feature flags for Activity Log, Restore, Backup and Scan
- [x] Activity Log Filters: https://github.com/wordpress-mobile/WordPress-Android/issues/13268
- [x] Restore: #13328 
- [x] Backup Download: #13329 
- [x] Scan: https://github.com/wordpress-mobile/WordPress-Android/issues/13326
- [x] Backup Screen: #13629

 #### Activity Log
- [x] Final Design Review 
- [x] Release
- [x] Remove feature flags two weeks after the release

#### Restore
- [x] Final Design Review
- [x] Release
- [x] Remove feature flags two weeks after the release

 #### Backup
- [x] Final Design Review 
- [x] Release
- [x] Remove feature flags two weeks after the release

 #### Scan
- [x] Final Design Review 
- [x] Release
- [x] Remove feature flags two weeks after the release

PR #14050, opened 2021-02-12
Update retry logic for Scan/Restore/DownloadBackup use cases

malinajirka:
Parent issue #13267 

I recommend reviewing this PR by commits.

- Moves "startWithDelay" outside of `while` loop in FetchScanStatus use case (it was a bug)
- Adds retry logic to FetchScanStateUseCase
- Updates retry logic in GetBackupDownloadStatus use case so it retries even on NetworkNotAvailable state
- Updates retry logic in GetReestoreStatus use case so it retries even on NetworkNotAvailable state
- Removes unnecessary nesting in all of the above usecases

To test:
1. Select a site with Backup plan
2. Open Activity Log
3. Click on "more" button on one of the rewindable items
4. Click on "Download backup"
5. Make sure the download backup flow works as expected - you can try to turn off internet connection for 2 seconds during the process and notice that the flow recovers without showing any errors
-------------------------
1. Select a site with Backup plan
2. Open Activity Log
3. Click on "more" button on one of the rewindable items
4. Click on "Restore to this point"
5. Make sure the restore flow works as expected - you can try to turn off internet connection for 2 seconds during the process and notice that the flow recovers without showing any errors
----------------------------------
1. Select a site with Scan plan
2. Introduce an active threat on the site (eg. install the calendar plugin)
2. Open Scan
3. Click on ScanNow button
4. Make sure the Scan works as expected - you can try to turn off internet connection for 2 seconds during the process and notice that the flow recovers without showing any errors

PR submission checklist:


zwarm (approved):
@malinajirka - nice improvements :) All flows work as expected. 👍

PR #14052, opened 2021-02-12
Add a "Set as Featured Image" Button to Image Block

### LINKED ISSUE #1011: [Proposal] Allow easier setting of Featured Image
## Problem

IMG_0450

The current Featured Image flow has a few issues:

* Featured Image is only available in Post Settings screens, which means it can get buried and not easy to locate

* If you've set an image from the post body as a Featured Image, it's not obvious which image is selected because there is no visual cue

## Solution

To solve the above issues, we should we should supply an intuitive "shortcut" of sorts to achieve this – in context. This could be on the first image of the post, which is typically the image users expect to use.

* Add a quick action button to the Image block settings sheet

* Add an indicator over the image that is being used as the Featured Image – similar indication that we have on GIFs in Aztec

* Optionally, if the Featured Image is already set, we could display a dialog that confirms whether the user wants to replace the current image.

It might look something like this:

IMG_0451

## Aztec

If we do any work on the Aztec end, I'd recommend we simply add the `Featured` badge to the image that's being used as the Featured Image.

SiobhyB:
With this PR, we'll be implementing a "set as featured image" button with the purpose of making it easier for users to assign a new featured image directly from the post area.

Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1011

`gutenberg-mobile`: wordpress-mobile/gutenberg-mobile#3116
`gutenberg`: https://github.com/WordPress/gutenberg/pull/28854

Please refer to the Gutenberg PR as the central place for steps to tests and screenshots.

PR submission checklist:


SiobhyB:
The PR is still in draft mode because, though the code technically works to set a featured image, the approach I’ve taken feels fragile/error-prone. 

I’ve added a new `updateFeaturedImage` method under [the _EditorFragmentListener methods_ section](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L2915) in **EditPostsActivity**. That method utilizes the existing `updateFeaturedImage` function [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostSettingsFragment.java#L1001) in **EditPostSettingsFragment**. I then call this function in **GutenbergEditorFragment**.

By calling the function in this way, our data tools won't distinguish between any clicks to the image block's new featured image button and the button within the Post Settings screen. It also feels fragile, like any issues with **EditPostSettingsFragment** may then lead to issues with the image block's button.   

I'm looking to explore other patterns/ways to set the featured image from **GutenbergEditorFragment**. 

cc-ing @hypest (who noted he'd be able to take a look over this later in the week) and @jd-alexander (who has familiarity with the code from buddying me with it these past weeks). Thanks in advance for your thoughts and any pointers!

hypest:
Looks like the bundling job fails with `error index.js: Cannot find module 'babel-plugin-transform-remove-console'` and that's the error I also see locally while trying to build WPAndroid on this PR. Will try to understand what's going on there.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
```diff
@@ -795,6 +808,39 @@ public void onClick(DialogInterface dialog, int id) {
...
+
+            AlertDialog dialog = builder.create();
+            dialog.show();
+        }
+    }
+
+    private void featuredImageSet(int mediaId) {
+        mEditorFragmentListener.updateFeaturedImage(mediaId, true);
+        setFeaturedImageId(mediaId);
+        ToastUtils.showToast(getActivity(), R.string.featured_image_confirmation).show();
```
hypest: 👋 Siobhan, I'm assuming that the Toast here is mostly for debugging purposes and not part of the final design, right? If so, what do you think about adding a "TODO" comment here to help us remember to remove it before we finalize the feature?

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -2974,6 +2974,11 @@ public void onEditPostPublishedSettingsClick() {
      * EditorFragmentListener methods
      */
 
+    @Override
+    public void updateFeaturedImage(final long mediaId, final boolean imagePicked) {
+        mEditPostSettingsFragment.updateFeaturedImage(mediaId, imagePicked);
```
hypest: Any particular reason to not reuse the [already defined `setFeaturedImageId`](https://github.com/wordpress-mobile/WordPress-Android/blob/05ddeced286ff7ab0ee67cfe856b178f0af6f0d7/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L2550)? That one calls the same function in `mEditPostSettingsFragment` and also checks if the reference is null, making it slightly more safe.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
```diff
@@ -795,6 +808,39 @@ public void onClick(DialogInterface dialog, int id) {
...
+                }
+            });
+
+            AlertDialog dialog = builder.create();
+            dialog.show();
+        }
+    }
+
+    private void featuredImageSet(int mediaId) {
+        mEditorFragmentListener.updateFeaturedImage(mediaId, true);
```
hypest: Looking at [the code that uses the `imagePicked` parameter](https://github.com/wordpress-mobile/WordPress-Android/blob/9f9a9c9d3b024debc62bf5b3211691d6390b1059/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostSettingsFragment.java#L1002-L1008) (passed as `true` here), my impression is that it uses that param just to trigger a Tracks event during the media picking flow.

During the set-feature-image flow from inside an image block, I'd expect to pass `false` so to avoid triggering that event so, I'm not sure why the `true` here @SiobhyB , can you elaborate?

This also seems related to the concern you expressed in [the comment](https://github.com/wordpress-mobile/WordPress-Android/pull/14052#issuecomment-801029547).

hypest:
> It also feels fragile, like any issues with EditPostSettingsFragment may then lead to issues with the image block's button.

Can you elaborate on that @SiobhyB ? Like, what could be an example of an issue with EditPostSettingsFragment that'd break the image block button?

In the meantime, in case that's what you have in mind too, it _is_ a bit awkward that the editor fragment needs to call a function in the Post Settings fragment just to update the featured image id. Feels like such a function could live in a headless "service" fragment inside the EditPostActivity instead of lying inside a UI fragment.

I still think that the current PR is a good first take on implementing the set-featured-image functionality, and a cleaner approach (perhaps by extracting the shared code that sets the featured image id in the local DB) could be worked on in a follow up PR.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
SiobhyB: You're right, I'll add a "TODO" and this is a great lesson for me to be more diligent in doing similar in the future.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
SiobhyB: Ah, only my lack of knowledge. :D I hadn't paid much attention to that function within EditPostActivity, I'd been too focused on inspecting the functions under EditPostSettings. I've updated this now and it works great.

SiobhyB:
Thank you for looking into this @hypest! 🙇‍♀️ 

> Can you elaborate on that @SiobhyB ? Like, what could be an example of an issue with EditPostSettingsFragment that'd break the image block button?

This was more of a _feeling_ than anything concrete I felt certain about. Your note about it being a bit awkward to call the function in the Post Settings fragment matches up to the hunch I had and I wasn't sure if I was missing wider problems this could bring due to my lack of experience.

> In the meantime, in case that's what you have in mind too, it is a bit awkward that the editor fragment needs to call a function in the Post Settings fragment just to update the featured image id. Feels like such a function could live in a headless "service" fragment inside the EditPostActivity instead of lying inside a UI fragment.

Great, thank you! If there aren't any major issues then I'll work with this approach for now. :)

hypest:
👋 @SiobhyB , looks like this PR is not up-to-date with the latest from the gutenberg-mobile/Gutenebrg side PRs and won't compile correctly. Can you update it? While at it, I wonder if it might be a better strategy to test/develop by using the WPAndroid app anyway (instead of the gb-mobile demo apps), since the feature includes many changes in the bridge anyway.

SiobhyB:
I've updated this PR now, thanks again for looking!

> While at it, I wonder if it might be a better strategy to test/develop by using the WPAndroid app anyway (instead of the gb-mobile demo apps), since the feature includes many changes in the bridge anyway.

I've been developing/testing using the WPAndroid app in order to allow me to test the changes I'm making. I'm still actively working on and committing to this issue while it's in draft mode, which I think is why you caught the PR while it's out-of-date. Let me know if I'm missing something else here.

_Edited to add: I'm aware that there are merge conflicts here, it's on my list to look into!_

hypest:
Just a quick note: we weren't exactly trying to merge this PR but got merged automatically because we merged [a PR that wholly included](https://github.com/wordpress-mobile/WordPress-Android/pull/14451) the commits already.

PR #14058, opened 2021-02-12
Merge Release/16.7 into develop

JavonDavis:
Merge Release/16.7 into develop


loremattei (approved):
I assume this is to update the release notes on GlotPress, even though it looks like there's also an unreleased change that gets merged (which is fine with me).

JavonDavis:
> I assume this is to update the release notes on GlotPress, even though it looks like there's also an unreleased change that gets merged (which is fine with me). 

@loremattei yes it is and I also checked with @antonis that those changes don't require a new beta build

PR #14062, opened 2021-02-12
WebViewActivity : Stop audio playback after exiting post preview

### LINKED ISSUE #14046: Audio Block doesn't stop playback after exiting post preview
### Expected behavior

Audio Block should stop audio playback after exiting post preview

### Actual behavior

After previewing a post with an audio block and playing the audio, exiting the preview and continuing use of the app does not stop audio playback.

1. Create a post with an audio block 
2. Add an audio file to the block with a decent amount of playback time
3. Preview the post and play the audio file 
4. Exit the preview and observe the playback doesn't stop

Google Pixel 4a, Android version 11, WPAndroid 16.7-rc-4

jd-alexander:
Fixes #14046

## Findings

This issue was occurring because the basic activity for displaying a WebView was not able to pause the audio playback when the user leaves the activity and `onPause` was run.   Currently, the `onPause` of the `WebViewActivity` contains logic for stopping flash video via `mWebView.onPause();` Since this solution does not extend to rendered `<audio controls> </audio>` an alternative solution was found.  

### Option 1  

The first idea that came to mind was to do some form of Javascript injection into the preview page, that would traverse through the elements in the HTML document, find the active <audio> component, and stop it.   

### Option 2

  Another idea I thought of was to simply clear the view in some way or form. After verifying with the Android documentation to validate that this is an accepted practice, I implemented the behavior they showcased as the optimal approach. This simply involves clearing the `WebView` with `WebView.load(“about:blank”)`. Doing this reliably resets the `WebView` state and releases all the page resources. In our case, the target page resources is that of the post/page preview that contains the playing content. Once the post/page preview’s resources are cleared up any running JavaScript will be released.  

###  Choice : Option 2 

Based on the simplicity, readability, and maintainability I decided to go with Option 2. I like the approach because it doesn’t just load any random URL, It loads the “about:blank” page that is built in the web browser. Other engineers who see this will be able to associate it with seeing this address in their browser with an empty page.   I considered putting the contents of the “load” method in a constant but decided against this, as when I checked in other places where a similar solution was utilized, they kept the “about:blank” inline to communicate intent. 

## Testing

1. Create a post. 
2. Add an Audio block.
3. Add a playable file with good length; over 20 seconds.
4. Ensure you have a moderate volume setting so the sound file is audible.
3. Start the playback. 
4. Exit the preview by pressing the back button to go back to the post content.
5. Notice that the audio has stopped.

## Regression testing

1. Ensure that you can open the link externally and navigate back with you no issues. 
2. Ensure that you can switch between mobile, desktop, and tablet, and no content or resources are lost.

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/WebViewActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/WebViewActivity.java
@@ -120,7 +120,9 @@ public abstract class WebViewActivity extends LocaleAwareActivity {
     private void pauseWebView() {
         if (mWebView != null) {
             mWebView.onPause();
-            loadUrl("about:blank");
+            if (isFinishing()) {
+                loadUrl("about:blank");
+            }
         }
     }
 
--- a/WordPress/src/main/java/org/wordpress/android/ui/WebViewActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/WebViewActivity.java
@@ -120,6 +120,7 @@ public abstract class WebViewActivity extends LocaleAwareActivity {
     private void pauseWebView() {
         if (mWebView != null) {
             mWebView.onPause();
+            loadUrl("about:blank");
         }
     }
 


jd-alexander:
The detekt plugin was introduced to the codebase recently and the lint failure is due to this. On Monday, I will check out what's causing this.

jd-alexander:
> The detekt plugin was introduced to the codebase recently and the lint failure is due to this. On Monday, I will check out what's causing this.

This issue should now be resolved with the `develop` merge as `develop` now contains a fix.

mzorz:
Thanks a lot for this change @jd-alexander! Going to try it in a few. I have some questions / notes in the meantime:

> they kept the “about:blank” inline to communicate intent. 

I like this, I also prefer this given it has meaning and gives context right away 👍 


> After verifying with the Android documentation to validate that this is an accepted practice, I implemented the behavior they showcased as the optimal approach.

Could you provide a link to the documentation? 🙏

> 5. Start the playback.
> 6. Exit the preview by pressing the back button to go back to the post content.


The steps here got me wondering, should the audio block also stop playing when sending the app to the background / locking the phone? I understand this PR as is does indeed fix the issue https://github.com/wordpress-mobile/WordPress-Android/issues/14046, but just a thought.

jd-alexander:
> Thanks a lot for this change @jd-alexander! Going to try it in a few. I have some questions / notes in the meantime:

🙇 

> I like this, I also prefer this given it has meaning and gives context right away 👍

Indeed!

> Could you provide a link to the documentation? 🙏

Whoops! Of course! Here you go https://developer.android.com/reference/android/webkit/WebView#clearView()

> The steps here got me wondering, should the audio block also stop playing when sending the app to the background / locking the phone? I understand this PR as is does indeed fix the issue #14046, but just a thought.

This is a good point. I used this approach because I assumed that backgrounding the app or locking the phone wouldn't be supported if the activity is finishing. So if they are just pausing for some other reason that doesn't cause the activity to be finished then the preview should continue. One eg, of this is navigating from the app to the external browser. Let me know if you have any thoughts on my ideas, as lifecycle behavior gets tricky sometimes, so I am open to other interpretations or perspectives.

mzorz (approved):
Thanks for the explanation @jd-alexander !  I think this is right and fits the purpose 👍 

LGTM :shipit:

PR #14063, opened 2021-02-13
With some keyboard creates chips on dot

### LINKED ISSUE #14066: With some keyboard creates chips on dot
### Expected behavior
In the Invite People screen, while adding `Usernames or emails`I would expect to be able to add emails with dot sign without the string being transformed in a chip.

### Actual behavior
In the Invite People screen, while adding `Usernames or emails`I can have the string validated and being transformed in a chip when using a keyboard different from the Google Gboard standard one. Not sure on other keyboards but I noticed this for example with the MS SwiftKey keyboard.

- Install and enable the MS SwiftKey keyboard
- Try to add a user email with a dot (like name.surname@domain.com)
- Check the string is validated and transformed into a chip when pressing dot (so getting a chip with `name.` string)

develric:
Fixes #14066

In the `Invite People` screen, currently the validation and transformation into a chip of the `Usernames or emails`field is triggered by `space`, `comma`, `return` keys. 

Using some keyboards (like for example the MS SwiftKey) the insertion of a `space` is done when inserting a `dot`; this is unwanted since it triggers the validation and transformation into a chip of the text inserted until then and it could prevent the insertion of emails (containing a dot).

To avoid this we have added the `textEmailAddress` qualifier to the TextInputEditText `android:inputType` field 

| Before with MS SwiftKey | After with MS SwiftKey |
|---|---|
| image | image |

While looking into this I also noticed some warnings when creating the chip. The warnings are like below

```
W/IInputConnectionWrapper: endBatchEdit on inactive InputConnection
W/IInputConnectionWrapper: clearMetaKeyStates on inactive InputConnection
W/IInputConnectionWrapper: beginBatchEdit on inactive InputConnection
```
This seems to be a [known issue ](https://stackoverflow.com/a/30454234) due to setting the text to empty string with setText(""). AFAIU this is not causing issues but I confirmed that using `text?.clear()` removes the warnings. Since the verification of this seems pretty confined I thought it made sense to fix this in the beta update.

### Warnings are not present any more
- Using develop (or current release 16.7) confirm that you can see warnings similar to the above mentioned when creating a chip
- Now use the apk from this PR and check the warning are no more present 

### Chips are not created with dot
- Install and enable the MS SwiftKey keyboard
- Using develop (or current release 16.7) confirm that adding a user email with a dot (like name.surname@domain.com), a chip is created when pressing the dot (so getting a chip like `name.`)
- Now use the apk from this PR and check the issue is not present and you can fully insert an email like `name.surname@domain.com`; only after tapping `space`, `comma`, `return` keys (or leaving the control focus) the chip is created as expected
- Try to send invitation and smoke test

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/people/WPEditTextWithChipsOutlined.kt
@@ -181,7 +181,7 @@ class WPEditTextWithChipsOutlined @JvmOverloads constructor(
         throwExceptionIfChipifyNotEnabled()
 
         if (!hasText() || !canAddMoreChips()) {
-            editor.setText("")
+            resetText()
             return null
         }
 
@@ -333,13 +333,13 @@ class WPEditTextWithChipsOutlined @JvmOverloads constructor(
 
     private fun addItem(item: String) {
         if (!canAddMoreChips() && item.isNotBlank()) {
-            editor.setText("")
+            resetText()
         } else {
             val cleanedItem = removeDelimiterFromItemIfPresent(item)
 
             if (cleanedItem.isNullOrBlank()) return
 
-            editor.setText("")
+            resetText()
 
             itemsManager?.onAddItem(cleanedItem) ?: chipify(cleanedItem, ItemValidationState.NEUTRAL)
         }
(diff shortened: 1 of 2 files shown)


planarvoid (approved):
I've tested the people screen with the MS Swift key before and after the changes and the issue is fixed 👍  good job!

PR #14064, opened 2021-02-13
Show message when all QS tasks from the same category are completed

### LINKED ISSUE #14042: Show success message when all tasks from a Quick Start category are completed
We should display a toast message when all tasks from a Quick Start category are completed:

https://user-images.githubusercontent.com/830056/107710350-87485400-6ca5-11eb-899a-c4c7dd2a0287.mov

Here's the copy for each category:

| | Customize your site | Grow your audience |
|---|---|---|
| **Title** | All tasks completed | All tasks completed |
| **Subtitle** | You’ve customized your site. Well done! | You’ll reach more people. Nice job! | 
| **Image** |  |  |

renanferrari:
Fixes #14042

This PR introduces a snackbar message to the improved My Site screen that is shown when all Quick Start tasks from a single category are completed. If the last task from that category is completed on the My Site screen, the snackbar appears immediately. If the task is completed on another screen, the snackbar appears when the user comes back to the My Site screen.

https://user-images.githubusercontent.com/830056/107836013-538b2e00-6d7a-11eb-9f14-9fedc78014a2.mp4

### To test

1. Go to the My Site screen and select a site where Quick Start is active.
1. Complete all tasks under the "Customize your site screen" category, but leave for last a task that is completed on the My Site screen (`UPDATE_SITE_TITLE` or `UPLOAD_SITE_ICON`).
1. After completing the last task, notice the corresponding snackbar message for that category is shown immediately.
1. Complete all tasks under the "Grow your audience" category and make sure the corresponding snackbar message is shown after you go back to the My Site screen.

---

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/mysite/QuickStartRepository.kt
```diff
@@ -72,6 +75,9 @@ class QuickStartRepository
     val activeTask = _activeTask as LiveData<QuickStartTask?>
 
     private var pendingTask: QuickStartTask? = null
+    private var pendingCategoryCompletion: QuickStartTaskType? = null
```
planarvoid, line 78: The solution with the pending category seems a little bit messy :-). I guess the reason why we can't show the snackbar right when the task is completed is that we can't differentiate what screen we're on, right? Normally I'd expect the livedata to be only consumed when the fragment is visible but I guess that doesn't work in our case.

planarvoid (commented):
Thanks for the PR! It's looking great. I have one thing that is a little weird. 
- Finish all the tasks in the "Customize" category except "Update title
- Click on "Update title"
- The "Update title" dialog gets shown
- The snackbar with the success message gets shown under it

I think we need to move the completion of the update icon/title tasks to when the dialogs are dismissed.

WordPress/src/main/java/org/wordpress/android/ui/mysite/QuickStartRepository.kt
renanferrari: >I guess the reason why we can't show the snackbar right when the task is completed is that we can't differentiate what screen we're on, right?

Yes. But even if we're actually on the My Site screen at the time of completion, that would still be an issue considering some tasks are completed _right before we leave_ the screen. The [`EXPLORE_PLANS` task](https://github.com/wordpress-mobile/WordPress-Android/blob/bde6762d259976921c7dbae6e48d88fe8e1ed8f3/WordPress/src/main/java/org/wordpress/android/ui/mysite/MySiteViewModel.kt#L269-L272) is one example.

>Normally I'd expect the livedata to be only consumed when the fragment is visible but I guess that doesn't work in our case.

That's correct. The snackbar livedata can be consumed even if the fragment is not visible. It only stops being observed once the fragment hits `onStop`.

I guess we could try to find a more elegant solution after we deal with #14073 as well. Feel free to share any suggestions you may have.

renanferrari:
Thanks for the review @planarvoid!

>I think we need to move the completion of the update icon/title tasks to when the dialogs are dismissed.

Good idea! I've implemented this in 256f9f3 and 715be2a. 

The site title dialog calls the dismissal callback every time the dialog is closed, no matter which interaction actually happened, so we just needed to call `completeTask` from a single place. 

The site icon dialog is a bit different, it only calls the dismissal callback when the dialog is actually dismissed, so we need to call `completeTask` from a few additional places. I also realized that if the user actually goes to the media picker screen, we shouldn't refresh immediately, so I've also added a change in 61c6809 that should give us more flexibility when calling `completeTask`.

I also noticed a bug with the `ENABLE_POST_SHARING` and `FOLLOW_SITE` tasks, where the snackbar would not show up if we left one of those tasks for last, since they were not being completed via `QuickStartRepository`. It was fixed in 
4f27488.

planarvoid (approved):
Thanks for the changes! I've tested the flow with updating site title/icon (and you can see right away that the tasks are finished after the dialog is gone). I've also tested the flow with the enable sharing/follow sites and it works well, good job 👍

Issue #14074, opened 2021-02-15
Remove workaround for rewind_download_ready


zwarm:
Remove the workaround for intercepting **type=stat**  swapping it for **type=stat**  in NotificationsTable.prepareNote

See this PR for details

thehenrybyrd:
@zwarm I think you're referring to https://github.com/wordpress-mobile/WordPress-Android/pull/14075?
I see in that PR you state that:
> This code MUST be replaced once the API properly formats the Note

Do you know if the required API changes have been made at this point?

PR #14075, opened 2021-02-15
Issue/13329 backup download notification

zwarm:
Parent #13329 

This PR adds temporary support for Backup Download ready notifications.

Currently the notifications endpoint returns **type=stat** which is turned into a Stats deep link. This is not the correct behavior for the download notification. In this PR we intercept the note when it is returned from the API and swap **type=stat** with **type=rewind-download-ready** before it gets loaded into the database.  

This code MUST be replaced once the API properly formats the Note (see issue #14074)

**Notes**
When viewing the notification detail view, the User name  appears to be a link; however tapping on it will not take you anywhere. The type for this clickable range is marked as *User*; however *User* links are treated as if they are reader comment links and rely on siteId to be present in the node. In the case of backup download (and restore), the user node does not include siteId, so we can't forward the user anywhere (And I don't believe we should forward to the reader). Either way, it feels like the entire body response needs to be customized at the API level for both of these features. And we should identify where each link should forward too. Adding hacky workarounds is a poor solution.

**Merge Instructions:**
- Review [FluxC PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1876)  merge and create a tag in FluxC develop
- Replace FluxC hash with tag from develop in build.gradle
- Remove the "Not Ready for Merge label"
- Merge this PR

**To test:**
- Launch the app
- Navigate to the activity log
- Tap on a item that supports backup and choose 'Download backup" from the menu
- Kick off a backup & wait until in ends
- Navigate back to the home screen
- Tap the Notifications tab
- Look for the "Your backup is ready for download" item
- Tap on the item to launch the notification detail view
- Tap on the *site's Backups* text (should be blue)
- Note that the Backup view is launched

PR submission checklist:


malinajirka (approved):
Thanks @zwarm! It's the best solution we can implement considering the current constraints. Good job.

PR #14076, opened 2021-02-15
Remove fetch delay for normal processing

zwarm:
Parent #13328 

Addresses the following by removing the delay for normal processing:
The email informing of a successful restore arrives while the app is still “Importing SQL dump” or “renaming tables”, maybe it’s a latency issue but the delay was quite noticeable.

See p5T066-p2

NOTE: This change DOES NOT guarantee that the restore process notification comes before the view is finished, but it _should_ help with the noticeable delay.

**Merge Instructions**
I'm not sure if this should be pushed to the beta or go into the develop branch. 

- If it needs to go to beta, then please change the base branch. If it can wait for next release, then leave develop as the base.
- Remove the Not ready for merge label
- Merge the PR

cc @malinajirka @ashiagr @ParaskP7  - thoughts on when this should get pushed? Thanks.

To test:
- Launch the app
- Navigate to the Activity Log
- Chose an item that is restorable
- Tap on the menu for the item and select "Restore to this point"
- Kick off the restore process and wait
- Note that the progress view finishes before the email notification arrives (or that the timing is less noticeable)

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/restore/usecases/GetRestoreStatusUseCase.kt
@@ -74,7 +74,6 @@ class GetRestoreStatusUseCase @Inject constructor(
                     QUEUED -> emitProgress(rewind)
                 }
             }
-            delay(DELAY_MILLIS)
         }
     }.flowOn(bgDispatcher)
 


ParaskP7:
👋 @zwarm !

> Addresses the following by removing the delay for normal processing:
> The email informing of a successful restore arrives while the app is still “Importing SQL dump” or “renaming tables”, maybe it’s a latency issue but the delay was quite noticeable.
> NOTE: This change DOES NOT guarantee that the restore process notification comes before the view is finished, but it should help with the noticeable delay.

I saw the comment on the P2 post about it and having the email arriving faster than the UI completes, but I am still very skeptical about this change. Wouldn't that mean that the while loop will run as fast as the `activityLogStore.fetchActivitiesRewind(...)` returns, meaning that it might every run 10 times a second, am I missing something here? To imagine that this was 10'' seconds a loop before `Jeptack Section`, I think we are okay if we just keep it as it is (1'' sec per loop).

Maybe all we can do is wait for users to (potentially) complain and revisit that, maybe even talking to backend and ask them if they can introduce a small delay between the restore being complete and the email being sent.

@ashiagr @malinajirka your thoughts?

PS: @momo-ozawa @emilylagunacan @leandroalonso can you also chip-in here and let us know how you are handling this on your end? The question is: Do you have any delay in between while asking for the restore progress?

> See p5T066-p2

I am not sure about this link, can you help me figure out where it points. PS: I managed to successfully open it but I got confused as to its result.

> cc @malinajirka @ashiagr @ParaskP7 - thoughts on when this should get pushed? Thanks.

I would otherwise suggest it goes to `develop`, I don't think this is a blocker for this release.

ParaskP7 (commented):
👋 @zwarm !

> Addresses the following by removing the delay for normal processing:
> The email informing of a successful restore arrives while the app is still “Importing SQL dump” or “renaming tables”, maybe it’s a latency issue but the delay was quite noticeable.
> NOTE: This change DOES NOT guarantee that the restore process notification comes before the view is finished, but it should help with the noticeable delay.

I saw the comment on the P2 post about it and having the email arriving faster than the UI completes, but I am still very skeptical about this change. Wouldn't that mean that the while loop will run as fast as the `activityLogStore.fetchActivitiesRewind(...)` returns, meaning that it might every run 10 times a second, am I missing something here? To imagine that this was 10'' seconds a loop before `Jeptack Section`, I think we are okay if we just keep it as it is (1'' sec per loop).

Maybe all we can do is wait for users to (potentially) complain and revisit that, maybe even talking to backend and ask them if they can introduce a small delay between the restore being complete and the email being sent.

@ashiagr @malinajirka your thoughts?

PS: @momo-ozawa @emilylaguna @leandroalonso can you also chip-in here and let us know how you are handling this on your end? The question is: Do you have any delay in between while asking for the restore progress?

> See p5T066-p2

I am not sure about this link, can you help me figure out where it points. PS: I managed to successfully open it but I got confused as to its result.

> cc @malinajirka @ashiagr @ParaskP7 - thoughts on when this should get pushed? Thanks.

I would otherwise suggest it goes to `develop`, I don't think this is a blocker for this release.

malinajirka:
I completely agree with everything Petros said. Removing the delay completely is imho too risky and not worth this (minor) UX issue.

zwarm:
@ashiagr @ParaskP7 @malinajirka - Thanks for the input, I was on the fence about this to begin with. Without a real push model, we are stuck with an imperfect solution.  I'll close this PR. Thanks.

emilylaguna:
> PS: @momo-ozawa @emilylaguna @leandroalonso can you also chip-in here and let us know how you are handling this on your end? The question is: Do you have any delay in between while asking for the restore progress?

We update the status once every 5 seconds. There could be a delay but I haven't noticed one.

ParaskP7:
Thanks for sharing @emilylaguna ! 🙏

PR #14080, opened 2021-02-16
Invalidate bundler cache to fix CI failure

AliSoftware:
Fixes the CI failures encountered in #14077, #14078 and #14079 on the "Installable Build" job, due to a bundler cache apparently being out of sync with the rake dependency.

_(Similar fix to what I had to do on WCAndroid in https://github.com/woocommerce/woocommerce-android/pull/3583/commits/d46e8bb0212612160ef6af6ef26f8178ee1b5796)_

#### To test

 - Ensure CI passes in this PR
 - Rebase the aforementioned PRs on top of this one (or cherry-pick this PR's commit into them) and check it fixes the CI failure

---

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #14078, opened 2021-02-16
Jetpack Section: Scan - fix crash on snackbar display

ashiagr:
### Description

This PR fixes below crash when snackbar is displayed on Scan and Threat Details screen.

```
java.lang.NullPointerException: scan_state_layout must not be null
        at org.wordpress.android.ui.jetpack.scan.ScanFragment.showSnackbar(ScanFragment.kt:126)
        at org.wordpress.android.ui.jetpack.scan.ScanFragment.access$showSnackbar(ScanFragment.kt:31)
        at org.wordpress.android.ui.jetpack.scan.ScanFragment$setupObservers$2$1.invoke(ScanFragment.kt:89)
```

### Findings

We passed fragment's root constraint layout (using Kotlin synthetics) in the `showSnackbar` method. We started experiencing this crash after replacing `FragmentContainerLayout` with `Fragment` in the corresponding activity layout xml. 

As per this [SO answer](https://stackoverflow.com/a/34543172/193545), we instantiated `viewModels` for these screen inside `onViewCreated` so the layout referenced in the `showSnackbar` method should've been present at the time we had to display the snackbar.

I think it is the way Kotlin synthetic properties are cached and referenced in the different ways we add a fragment (directly in the xml or using `FragmentContainerView`). Or crash might be occurring on random devices as per [this comment](https://stackoverflow.com/questions/34541650/nullpointerexception-when-trying-to-access-views-in-a-kotlin-fragment/34543172#comment98324304_34543172), however I'm not 100% sure.

### Solution

It is sufficient to just pass fragment's view in the `showSnackbar` method, so I've removed reference to the fragment's constraint layout in the method. This fixes the crash and snackbar is displayed properly.

### To Test

Fix or ignore a threat from Threat Details screen (on a Jurassic Ninja site with scan capability and bad calendar plugin installed), see that app does not crash and snackbar is displayed properly.

(Or you can return `Failure.RemoteRequestFailure ` from `FixThreatsUseCase` so that snackbar is shown for the failed response.)

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #14077, opened 2021-02-16
P2 people invite links unit test

develric:
Fixes #13456

This PR adds unit testing for 13456 issue introduced components.

## To test
- Check the unit tests pass
- I needed to add RoleUtilsWrapper.kt and SimpleDateFormatWrapper.kt to make it testable. Would be good to make some smoke testing just in case 😄 

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #14082, opened 2021-02-16
[Tooling] Fix the bundler/rake failure on CI

AliSoftware:
This is the second attempt at https://github.com/wordpress-mobile/WordPress-Android/pull/14080 as it seems the fix wasn't sufficient and was merged too soon (before we finished testing the fix on the other PRs)
 - The first build with the fix, on the PR fixing it, passed – and PR #14080 went green.
 - But then the build failed on the other PRs I cherry-picked it on (which were triggered in parallel).
 - ~Hopefully that failure was only because the builds were racing each other as it seems like cache got picked by 2nd PR before 1st PR finished building it. So hopefully waiting until the job is done this time will help. I also used different cache keys for the different jobs just to be on the safe side.~ In fact the issue was deeper than this, see comments below and also complementary details in [the WCAndroid PR doing the same on Woo](https://github.com/woocommerce/woocommerce-android/pull/3585)

To test:
 - Ensure CI passes on this PR 
 - Then **once the CI on this PR has finished and passed**, then **cherry-pick the changes on one of the other failing PR** (#14077, #14078 and #14079)

_PR created as Draft to be sure to not merge it too fast this time_


AliSoftware:
After multiple attempts and debugging, it happens that the solution is to update `bundler` to `2.2.10`.

This is because CircleCI has updated their image to use `bundler 2.2.10` apparently recently, very likely because this version contains an [important security fix](https://github.com/rubygems/rubygems/blob/master/bundler/CHANGELOG.md#security-fixes) about an issue that made some noise on internet recently.

Bundle bundler 2.2.10 also generates a different `Gemfile.lock` (to separate the gem in the lockfile by source as part of solving that security issue) and I'm guessing likely a slightly different resolution algorithm taking that into account, which is why the `Gemfile.lock` which was resolved by bundler `2.1.4` (that we were using up until now) was causing trouble to bundler `2.2.10`.

AliSoftware:
After reading in more details [the whole PR on bundler](https://github.com/rubygems/rubygems/pull/3655/files) about this security issue, I realized that with the setup we had made bundler still use the multi-source mode, because the plugin we include in `Pluginfile` was not inside the `source` block so resolved its dependencies from a different set of sources.

<details><summary>Explanations of that side-effect issue</summary>

`bundler`'s PR explains that the whole point of fixing the security issue is to also recommend to avoid using multiple sources in Gemfiles, and using `source 'xxx' do … end` blocks to group some gems in a given source and some others (aka the ones outside the block) not having the same source will also cause issues.

In fact, as discussed in [this comment on the PR](https://github.com/rubygems/rubygems/pull/3655#issuecomment-779318620), when resolving transitive dependencies it will use the source of the parent gem, but in the case of our `Pluginfile` which doesn't specify a source, that only resolves to local gems, so wasn't able to find `rake` and `octokit` and others locally (when cache was empty at least), instead of sourcing them from rubygems.orgs like all other gems, and leading to the `Gemfile.lock` having multiple sources (hence the initially bigger diff on the `Gemfile.lock` after this change before the last commit)…
</details>

Only figured that out while trying to replicate the fix on WCAndroid in https://github.com/woocommerce/woocommerce-android/pull/3585 – so now we use a single global `source` at the top, applied to all gems including the plugin.

Pfew 😓 I hope that's it this time

AliSoftware:
* Fix cherry-picked and being tested on Gutenberg PR too here: p1613490458125400/1613486607.116800-slack-C0126N6JH7Z
* Fix also tested on #14077, #14078

.circleci/config.yml
```diff
@@ -249,7 +249,7 @@ jobs:
           init-submodules: true
       - checkout-submodules
       - bundle-install/bundle-install:
-          cache_key_prefix: installable-build-v2
+          cache_key_prefix: release-build-v2
```
jkmassel, line 252: Nice

.circleci/config.yml
```diff
@@ -548,7 +548,7 @@ jobs:
           init-submodules: true
       - checkout-submodules
       - bundle-install/bundle-install:
-          cache_key_prefix: installable-build-v2
+          cache_key_prefix: translation-review-build-v2
```
jkmassel, line 551: Very nice

jkmassel (approved):

PR #14084, opened 2021-02-16
Restoring the usage of commitAllowingStateLoss instead of commit in Bottom Nav

### LINKED ISSUE #13987: Reader: tapping the login link in the filters bottom sheet in logged-out Reader causes the app to freeze and close
To prepare:

* Add a self-hosted site.
* Log out of the app.

To test: 

1. Go to Reader > Following > Filter.
1. Observe there are zero sites and topics followed in the bottom sheet that appears.
1. Tap the "Log in to WordPress.com" link inside the bottom sheet.

**Result:** the app stops responding if you tap the login link in the filters bottom sheet in logged-out Reader. ([41s](https://cloudup.com/cd2nn3UnG16))

image
<sup>Tested with WPAndroid 16.6-rc-2 on Pixel 3 Android 11.</sup>

develric:
Fixes #13987 

<details>
  <summary>History left here for context; please look the update section below for final decisions</summary>
  
Was looking into the above mentioned issue and drafting this PR to further discuss and double check the approach we can use to fix (pinging you since you worked on the code involved and may have some context 🙇 )

AFAIU we were using `commitAllowingStateLoss` when setting the fragment page on the bottom navigation until some point like here below:

```
if (fragment != null) {
            fragmentManager
                    .beginTransaction()
                    .replace(R.id.fragment_container, fragment, getTagForPosition(position))
                    // This is used because the main activity sometimes crashes because it's trying to switch fragments
                    // after `onSaveInstanceState` was already called. This is the related issue
                    // https://github.com/wordpress-mobile/WordPress-Android/pull/10853
                    .commitAllowingStateLoss()
        }
```

in the specific case of this issue (attempt to login from reader bottom sheet when being in a self-hosted account) that was preventing the crash that can happen considering the timing of [this related part of code ](https://github.com/wordpress-mobile/WordPress-Android/blob/ef56b66e13cb50f2edc5217536f7755002179257/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java#L583).

You can forcely reproduce the issue setting to 1500 ms the current 500 ms delay there.

With this increased delay and using the `commitAllowingStateLoss´I was not able to spot odd conditions around that specific scenario or smoke testing; wondering if we could restore that behavior. wdyt? any insight I'm missing?

Another possibility could be to remove the `mBottomNav.postDelayed` and setting the bottom navigation to My Site when coming back from Me section. This could be maybe cleaner but tbh I'm not sure it's trivial to collect/discriminate all possible cases when you should set the My Site page or not 🤔.

Thanks for your thoughts/insights 🙇‍♂️
</details>

## UPDATE

We are keeping the commit and relying on View Model live data lifecycle awareness, separating the start of the login flow from the setting of main tab as main page. Re this last point, we found (at least) one flow that do not set the main page to my site. Talking about when the user actually decide to login. In this case the VM is restarted so we need to preventively set the main page to be My Site in the shared prefs.

### To test
- Starting from a self-hosted user, go to the reader subfilter
- Tap the "Log in to WordPress.com" 
- Check you are landed into Me section and no crash
- Come back and check you are landed into the My Site page
- Do the same steps as above but now do a login in the Me Site
- After the login check you are landed in the My Site page and not into the reader page that was the last selected page by the user

PR submission checklist:


malinajirka:
Hi @develric, 
it seems that `getStartLoginFlow().observe` is causing a side-effect (postDelayed + switch tabs) in the view layer which is the root cause of this issue. I'd consider handling this in the kotlin code(viewmodel) using `delay(500)`. 

Something like
```
fun onOpenLoginPage() {
        _startLoginFlow.value = Event(true)
       delay(500)
       _selectTab.value = MY_SITE (or just _switchToMySite.value = Event(Unit))
    }
```

Since LiveData are lifecycle aware, I believe the tab selection will never happens after `onSaveInstanceState`. Therefore, we'll still be able to use `commit` and not `commitAllowingStateLoss`. 
Wdyt?

develric:
Hi @malinajirka 👋 

>Since LiveData are lifecycle aware, I believe the tab selection will never happens after onSaveInstanceState.

Love this ❤️ ! As simple as effective 👍 !

As discussed on slack I agree this will prevent the crash, only additional scenario we need to cover is when the user actually decides to login after tapping on the "Log in to WordPress.com" link from the reader bottom sheet. In this case the VM is recreated and we risk to land the user to the reader that was the last page that was selected. Adding saving of My Site page in the shared prefs to cover this also. 

Let me know wdyt 🙇

WordPress/src/test/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModelTest.kt
```diff
@@ -341,10 +342,28 @@ class WPMainActivityViewModelTest : BaseUnitTest() {
...
+        }
+
         startViewModelWithDefaultParameters()
-        viewModel.onOpenLoginPage()
 
-        assertThat(viewModel.startLoginFlow.value!!.peekContent()).isTrue()
+        viewModel.onOpenLoginPage(0)
+
+        assertThat(loginFlowTriggered).isTrue
+        verify(appPrefsWrapper, times(1)).setMainPageIndex(eq(0))
```
malinajirka, line 368: 🔍 Nitpick: I realize this test was not introduced in this PR, but since we are adding more assertions I think the comment is still relevant. Having said that feel free to ignore it.

Usually it's better to have a single focused tests with a clear name explaining what they are testing. There are two good reasons why this practice can be useful.
1. Unit tests serve as a documentation for other devs (future-self) -> "the correct action is triggered" doesn't say much -> the documentation should say what the "correct actions" are.
2. Unit tests would ideally have a single reason to fail -> when it fails anyone should be able to quickly find the issue and fix it without (almost any) domain knowledge.
3. 
We might want to consider splitting this into 3 test cases
1. "...., the login page flow is started"
2. "..., the default bottom sheet page is set to MySite"
3. "..., the bottom sheet tab is switched to MySite"

Let me know what you think ;)

malinajirka (approved):
Thanks @develric, I've reviewed the code and tested the app and it works great! I've left one minor comment, but feel free to ignore it and merge this PR without any changes.

WordPress/src/test/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModelTest.kt
develric: makes sense 👍 ! Changed to follow your advice. Thanks for all the feedbacks @malinajirka 🙇 !

PR #14085, opened 2021-02-16
Add 10mb limit to loading thumbnails from videos

planarvoid:
Fixes one possible reason of #9997

I've found out one of the reasons for the OOMs. The reason is that in order to show the video thumbnails in all the media screens we use glide. Glide starts loading the whole video and allocates memory in order to keep it. If the video is bigger than available memory, we get OOM. This crash happens right after you enter the screen (before the whole video is loaded). In order to fix this I've added a check that pings the URL to find out the size of the video before it's checked. We don't show thumbnails for videos bigger than 10 megabytes. It's still possible to click on the video to see the thumbnail. I think this approach is a good compromise. I don't think users would expect all the videos on the screen to be loaded.

To test (media screen):
- Go to a site with some big videos uploaded
- Go to My Site/Media
- Notice the app hasn't crashed
- Notice the big videos don't show a thumbnail, instead you can see a grey background (in the light mode)

To test (GB post):
- Go to a site with some big videos uploaded
- Click on "Add post"
- Add a "Video" block
- Click on "Add video"
- Choose from Wordpress library
- Notice no crash
- Notice thumbnails get loaded only for videos < 10 mbs

To test (Story):
- Go to a site with some big videos uploaded
- Click on "Add a story"
- Click on "Wordpress" icon
- Notice no crash when showing the list of items
- Notice thumbnails get loaded only for videos < 10 mbs

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/media/VideoLoader.kt
```diff
@@ -0,0 +1,54 @@
...
+) {
+    fun runIfMediaNotTooBig(scope: CoroutineScope, filePath: String, onSuccess: () -> Unit, onFailure: () -> Unit) {
+        scope.launch {
+            var length = 0
+            withContext(bgDispatcher) {
+                try {
+                    val url = URL(filePath)
+
+                    val urlConnection = url.openConnection()
+                    for ((key, value) in authenticationUtils.getAuthHeaders(filePath).entries) {
```
malinajirka, line 37: ⚠️ ❓ Do we use this method when loading video's from 3rd party sites? It seems that we add the AuthHeaders no matter the destination and I'm wondering if it might cause any security issues. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/media/VideoLoader.kt
```diff
@@ -0,0 +1,54 @@
...
+                        urlConnection.addRequestProperty(key, value)
+                    }
+
+                    length = urlConnection.contentLength
+                    if (length <= 0) {
+                        length = urlConnection.getHeaderFieldInt("Content-Length", 0)
+                    }
+                } catch (ioe: IOException) {
+                    ioe.printStackTrace()
```
malinajirka: ⚠️ We might want to consider using AppLog instead of sout, wdyt?

WordPress/src/main/java/org/wordpress/android/ui/media/VideoLoader.kt
```diff
@@ -0,0 +1,54 @@
...
+                }
+            }
+            withContext(mainDispatcher) {
+                if (length in 1 until SIZE_LIMIT_10_MB) {
```
malinajirka: ❓ Is there any difference between `if (length in 1 until SIZE_LIMIT_10_MB)` and `if (length < SIZE_LIMIT_10_MB)`? IMO `if (length < SIZE_LIMIT_10_MB)` is more readable. Wdyt?

WordPress/src/main/java/org/wordpress/android/util/image/ImageManager.kt
```diff
@@ -111,23 +118,34 @@ class ImageManager @Inject constructor(private val placeholderManager: ImagePlac
...
-                .addFallback(imageType)
-                .addPlaceholder(imageType)
-                .applyScaleType(scaleType)
-                .attachRequestListener(requestListener)
-                .apply(RequestOptions().frame(0))
-                .into(imageView)
-                .clearOnDetach()
+        videoLoader?.runIfMediaNotTooBig(scope,
+                videoUrl, {
+            GlideApp.with(context)
```
malinajirka: 💡 We might want to consider explicitly specifying the parameters names `onSuccess = ..., onFailure = ...`. I find the current syntax a bit difficult to follow. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerAdapter.kt
```diff
@@ -12,7 +13,11 @@ import org.wordpress.android.ui.mediapicker.MediaPickerUiItem.Type
 import org.wordpress.android.ui.mediapicker.MediaPickerUiItem.VideoItem
 import org.wordpress.android.util.image.ImageManager
 
-class MediaPickerAdapter internal constructor(imageManager: ImageManager) : Adapter<ThumbnailViewHolder>() {
+class MediaPickerAdapter internal constructor(
+    private val imageManager: ImageManager,
+    private val coroutineScope: CoroutineScope
+) :
+        Adapter<ThumbnailViewHolder>() {
```
malinajirka: 🔍 Nitpick: It seems this extra line break is not necessary.

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerFragment.kt
```diff
@@ -499,7 +499,8 @@ class MediaPickerFragment : Fragment() {
     private fun setupAdapter(items: List<MediaPickerUiItem>) {
         if (recycler.adapter == null) {
             recycler.adapter = MediaPickerAdapter(
-                    imageManager
+                    imageManager,
+                    viewModel
```
malinajirka: 🔍 Nitpick: We might want to consider casting the VM so the code is more readable ` viewModel as CoroutineScope`. It scared me at first that we are passing the VM into the adapter, before I realized we are passing just the scope.

WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoThumbnailViewHolder.kt
```diff
@@ -2,16 +2,25 @@ package org.wordpress.android.ui.photopicker
...
-class PhotoThumbnailViewHolder(parent: ViewGroup, private val thumbnailViewUtils: ThumbnailViewUtils) :
+@Deprecated(
+        "This class is being refactored, if you implement any change, please also update " +
+                "{@link org.wordpress.android.ui.mediapicker.PhotoThumbnailViewHolder}"
+)
+class PhotoThumbnailViewHolder(
+    parent: ViewGroup,
+    private val thumbnailViewUtils: ThumbnailViewUtils,
+    private val imageManager: ImageManager
+) :
```
malinajirka: 🔍 Nitpick: The linebreak seems unnecessary.

malinajirka (commented):
Thanks @planarvoid! It looks good overall. I've left a couple of questions/suggestions. Let me know what you think.

WordPress/src/main/java/org/wordpress/android/ui/media/VideoLoader.kt
planarvoid: as discussed in DM, this call should be safe because it's checking if the URL can accept the auth headers and doesn't add them when it can't

WordPress/src/main/java/org/wordpress/android/ui/media/VideoLoader.kt
planarvoid: Good point 👍 . I was following AS advice but it doesn't lead to great readability. I've introduced a new constant for the min size and changed the logic in 99c9982eb0

WordPress/src/main/java/org/wordpress/android/util/image/ImageManager.kt
planarvoid: Sound good 👍 , I've renamed the params to `loadAction` and `fallbackAction` and added named calls in 99c9982eb0

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerAdapter.kt
planarvoid: Fixed in 99c9982eb0

WordPress/src/main/java/org/wordpress/android/ui/photopicker/PhotoThumbnailViewHolder.kt
planarvoid: Good catch, I've fixed the other linebreaks as well in e30225ad41

WordPress/src/main/java/org/wordpress/android/ui/mediapicker/MediaPickerFragment.kt
planarvoid: Good idea 👍 , casted in e30225ad41

planarvoid:
thanks for the review @malinajirka , I've implemented all your suggestions!

malinajirka (approved):
Thanks @planarvoid! It all LGTM. Nice investigation and fix!

ParaskP7:
👋 @planarvoid !

> I've found out one of the reasons for the OOMs. The reason is that in order to show the video thumbnails in all the media screens we use glide. Glide starts loading the whole video and allocates memory in order to keep it. If the video is bigger than available memory, we get OOM. This crash happens right after you enter the screen (before the whole video is loaded). 

Really nice investigation, thanks so much for figuring this out! 🙏 

> In order to fix this I've added a check that pings the URL to find out the size of the video before it's checked. We don't show thumbnails for videos bigger than 10 megabytes.

Really nice solution, given the circumstances! 🥇 

> It's still possible to click on the video to see the thumbnail. I think this approach is a good compromise. I don't think users would expect all the videos on the screen to be loaded.

Indeed, this is a good compromise, but I am a bit skeptical about it.

Today, I have been playing with this as well, and I must admit that I find the 10MB restriction a bit too restrictive for videos. Almost all the videos I have on my device are more than 10MB. It seems to me that with 10MB you get just a few seconds on a video, obviously always depending on the quality. As such I am curious, given the restriction, if our video thumbnails feature effectively becomes unusable as most of the video thumbnails will stop previewing. In might even be the case that this will make it hard for most of our users to quickly select their desired video from the list and thus might ending up frustrating them. I am just speculating here, with no data to support that, just mere intuition, so apologise in advance.

Having said that, and after confirming that assumption of mine with data, somehow, maybe the next step would be to create a task to figure out a way to load video thumbnails another way, a way which will lift this 10MB restriction. Wdyt?

malinajirka:
Good point @ParaskP7! I've also realized that perhaps we are using a wrong url for some of the items. For example MediaModel in FluxC has `mThumbnailUrl` field, but from my very quick investigation it seems that we are using `mUrl` field instead. There is also `ImageUtils.getVideoFrameFromVideo(...)` and `ReaderVideoUtils.retrieveVideoThumbnailUrl(..)` methods. Could you or @planarvoid please double check that there isn't a space for improvement there?

ParaskP7:
👋 @malinajirka !

> I've also realized that perhaps we are using a wrong url for some of the items. For example MediaModel in FluxC has mThumbnailUrl field, but from my very quick investigation it seems that we are using mUrl field instead. There is also ImageUtils.getVideoFrameFromVideo(...) and ReaderVideoUtils.retrieveVideoThumbnailUrl(..) methods. Could you or @planarvoid please double check that there isn't a space for improvement there?

Yea, I was thinking of that, and wow, thanks so much for doing an initial digging. I believe it is worth it if we dig a bit more and figure out whether there is indeed a space for improvement there, maybe even a quick win.

planarvoid:
hey @ParaskP7  and @malinajirka , if I remember correctly we can't use `mThumbnailUrl` (I think it's either empty for videos or the URLs cannot be open in the app). The other methods are promising. Feel free to investigate either of the options 👍

PR #14097, opened 2021-02-17
Use StringBuilder instead of += when building Site Settings blacklist

### LINKED ISSUE #14070: Improve performance of site settings sync
We're getting ANRs when the app syncs the site settings. The code does quite a lot on the main thread so I think a reasonable approach would be to look at the part of code and move some parts of it to a background thread.

Sentry: https://sentry.io/organizations/a8c/discover/wordpress-android:19181e61c67b44c59728544fab233c0f/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&query=user.email%3Atangoooo480%40gmail.com&sort=-timestamp&statsPeriod=24h&widths=-1&widths=-1&widths=-1&widths=-1&widths=-1

planarvoid:
Fixes #14070 

We've confirmed that this ANR happens to a site with a lot of commenters spammed out (42k). The building of the `blacklist` field is quite inefficient. I was able to reproduce this issue and it's gone when the blacklist handling is changed to StringBuilder.

To test:
- Put lot of items in the blacklist (ping me if you want the list)
- Switch to the site with the blacklist
- The app is normally responding

PR submission checklist:


malinajirka (approved):
LGTM thanks!

PR #14101, opened 2021-02-18
Fix EventBusException(already registered) in JetpackCapabilitiesUseCase

malinajirka:
Fixes https://sentry.io/share/issue/f425f33b2f5c4245835c7e0fa9e3a3fc/
```
Subscriber class org.wordpress.android.ui.jetpack.JetpackCapabilitiesUseCase already registered to event class org.wordpress.android.fluxc.store.SiteStore$OnJetpackCapabilitiesFetched
```

The crash happens when jetpack capabilities for multiple site are fetched at the same time. When I wrote the use case I didn't realize that multiple registration to the event bus result in an exception. This PR fixes the issue by moving the registration to "init" and leaving the unregistration to the parent of the use case.

To test:
1. Clear app data
2. Login to an account with at least 3 sites
3. Set Network type: GSM and Signal Strength: Poor in the emulator settings
4. Switch to SiteB
5. Quickly switch to SiteC and notice the app doesn't crash

PR submission checklist:

@JavonDavis I'm sorry for not noticing this issue sooner. Unfortunately I think this bug needs to be fixed before the release as it might crash to a lot of users. Trends in Sentry seems to be confirming this. Let me know if I can help with something.

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/JetpackCapabilitiesUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/JetpackCapabilitiesUseCase.kt
@@ -34,6 +34,14 @@ class JetpackCapabilitiesUseCase @Inject constructor(
 ) {
     private var continuation: HashMap<Long, Continuation<OnJetpackCapabilitiesFetched>?> = hashMapOf()
 
+    init {
+        dispatcher.register(this@JetpackCapabilitiesUseCase)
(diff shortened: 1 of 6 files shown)


ParaskP7 (approved):
👋 @malinajirka  !

I used an emulator with GSM network type and poor signal strength. I first manage to reproduce the exception and then verified that this PR fixes it.

PS: Feel free to merge this PR when CI completes.

PR #14105, opened 2021-02-18
Make sure site model is selected before the ID of the site is emitted

### LINKED ISSUE #14104: Jetpack Scan/Backup menu items always hidden after restart on ImprovedMySiteFragment
**This issue happens only when ImprovedMySite feature flag is enabled.** 

### Expected behavior
When I restart the app, "Scan" and "Backup" menu items on MySite screen should be shown if the user has access to those products.

### Actual behavior
When I restart the app, "Scan" and "Backup" menu items on MySite screen are always hidden at first even though the user has access to the products. The items appear for example after the user selects a different site.

1. Enable ImprovedMySite feature flag
2. Select a site which has Jetpack "Backup" or "Scan" features enabled
3. Select MySite screen
4. Force close the app
5. Start the app and notice "Backup"/"Scan" items are hidden
6. Switch to a different site and back to the current site
7. Notice "Backup"/"Scan" menu items are visible (if the current site has access to the products)

develop with ImprovedMySite feature flag enabled

Thank you @ParaskP7 for reporting this!

planarvoid:
Fixes #14104 

The original issue is caused by the selected site missing the first time the app is started. The fix is quite simple. When a site change happens, we first update the site object and as a second step emit the "SiteSelectedId" event.

To test:
- Enable ImprovedMySite feature flag
- Select a site which has Jetpack "Backup" or "Scan" features enabled
- Select MySite screen
- Force close the app
- Start the app and notice "Backup"/"Scan" items are visible

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/mysite/SelectedSiteRepository.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/mysite/SelectedSiteRepository.kt
@@ -2,6 +2,8 @@ package org.wordpress.android.ui.mysite
 
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.distinctUntilChanged
+import androidx.lifecycle.map
 import org.wordpress.android.fluxc.Dispatcher
 import org.wordpress.android.fluxc.generated.SiteActionBuilder
 import org.wordpress.android.fluxc.model.SiteModel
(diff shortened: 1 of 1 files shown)


renanferrari:
Couldn't `siteSelected` be simplified to something like: 

```kotlin
val siteSelected = _selectedSiteChange.map { it?.id }.distinctUntilChanged()
```

malinajirka (approved):
LGTM, thanks!

Issue #14107, opened 2021-02-18
NullPointerException: row_scan must not be null


loremattei:
**Sentry Url**: https://sentry.io/share/issue/d4d49740d05249ffaf6922b53df0d3e6/
**User Count**: 97
**Count**: 243
**First Release**: org.wordpress.android@16.7-rc-1+988
**First Seen**: 2021-02-10T14:31:13.613000Z
**Last Seen**: 2021-02-18T14:22:07Z
**24 Hours**: 31
**30 Days**: 241

```
NullPointerException: row_scan must not be null
     at org.wordpress.android.ui.main.MySiteFragment.updateScanAndBackupVisibility(MySiteFragment.kt:299)
     at org.wordpress.android.ui.main.MySiteFragment.access$updateScanAndBackupVisibility(MySiteFragment.kt:183)
     at org.wordpress.android.ui.main.MySiteFragment$updateScanAndBackup$1.invokeSuspend(MySiteFragment.kt:289)
...
(2 additional frame(s) were not displayed)
```

renanferrari:
I'm closing this as it has already been addressed by #14108.

PR #14108, opened 2021-02-18
Fix NPE when fetchJetpackPurchasedProduct finishes after onViewDestroyed

malinajirka:
Fixes https://github.com/wordpress-mobile/WordPress-Android/issues/14107

Hopefully last fix for the beta build. This PR fixes NPE happening when fetchJetpackPurchasedProduct finishes after the fragment's view gets already destroyed - eg. when user switches to a different tab or puts the app into the background.

To test:
1. Slow down internet connection on an emulator
2. Switch site and quickly switch to "Notifications" tab in the bottom navigation
3. Notice the app doesn't crash

PR submission checklist:

cc @JavonDavis


planarvoid (approved):
Looks good 👍 . I've tested the previously crashing flow and it works with this fix

PR #14115, opened 2021-02-19
Only show Snackbar in stories when the target view is present

### LINKED ISSUE #13679: NullPointerException: Attempt to invoke virtual method 'android.content.Context android.view.View.getContext()' on a nu...
Sentry Issue: [WORDPRESS-ANDROID-11TA](https://sentry.io/share/issue/df7ab6403ec9400e87c4a317bf3572ec/)

```
NullPointerException: Attempt to invoke virtual method 'android.content.Context android.view.View.getContext()' on a null object reference
    at org.wordpress.android.widgets.WPSnackbar.<init>(WPSnackbar.java:31)
    at org.wordpress.android.widgets.WPSnackbar.make(WPSnackbar.java:85)
    at org.wordpress.android.ui.stories.StoryComposerActivity$setupStoryEditorMediaObserver$2.onChanged(StoryComposerActivity.kt:374)
    at org.wordpress.android.ui.stories.StoryComposerActivity$setupStoryEditorMediaObserver$2.onChanged(StoryComposerActivity.kt:88)
    at org.wordpress.android.ui.stories.media.StoryEditorMedia$addNewMediaItemsToEditorAsync$1.invokeSuspend(StoryEditorMedia.kt:88)
...
(14 additional frame(s) were not displayed)
```

planarvoid:
Fixes #13679 

This fix is fairly straightforward problem. We're trying to show a snackbar on a view that's not there. The reason is that the target view is in the EditPostActivity and the crash happens in StoryComposerActivity. What is the desired behaviour? Do we want to show the snackbar in the EPA or is it an oversight? We might want to consider switching views to `compose_loop_frame_layout`. WDYT @mzorz ? 

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
@@ -6,6 +6,7 @@ import android.app.ProgressDialog
 import android.content.Intent
 import android.net.Uri
 import android.os.Bundle
+import android.view.View
 import androidx.fragment.app.DialogFragment
 import androidx.lifecycle.Observer
 import androidx.lifecycle.ViewModelProvider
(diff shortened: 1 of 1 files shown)


mzorz:
Nice catch @planarvoid ! Yes this snackbar is only used in case of error when adding media to the Story so it should only show in the Story composer - we can switch to `compose_loop_frame_layout` as suggested (also the extra checks in this PR look good to me) 👍

planarvoid:
thanks for the answer @mzorz , I've updated the code to point to the correct layout 👍

mzorz (approved):
Thank you for your changes @planarvoid ! 
LGTM :shipit:

PR #14123, opened 2021-02-19
Merge Release/16.7 into develop

JavonDavis:
Merges 16.7 into develop


Gemfile
```diff
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
-source "https://rubygems.org" do
-  gem 'fastlane', "~> 2"
-  gem 'nokogiri'
-end
+source "https://rubygems.org"
+
+gem 'fastlane', "~> 2"
+gem 'nokogiri'
```
AliSoftware, line 6: I'm a bit surprised to see this diff as we're supposed to have fixed that **both** in the `release/16.7` branch and in `develop` in parallel 🤔  (didn't we, @jkmassel?)

I mean, in practice the new content is what we do want, so it's looking good, just wondering if seeing that diff doesn't hide something else we'd have missed or something?

JavonDavis:
Resolving conflicts over in https://github.com/wordpress-mobile/WordPress-Android/pull/14124

Gemfile
AliSoftware: Ok I'm confused now because now that I've seen #14124 exists, I'd expect this PR to in fact be closed in favor of this new #14124? (So I should not review this?)

Still surprising that there would be a diff between `develop` and `release/*` here anyway, regardless of conflict resolution required for other stuff. I mean, both branches are supposed to have the same fix for this in both places…

JavonDavis: Hey @AliSoftware yep that's right, there's no real need to review this one, but we still needed to branch off to resolve the conflicts

JavonDavis:
@AliSoftware @jkmassel actually I'll still need a 👍  on this one to merge in the changes from finalising the release

jkmassel (approved):

JavonDavis:
Actually after discussing with @jkmassel I think it might be best to go with a new branch to get these changes in so closing this PR.

PR #14127, opened 2021-02-22
Issue/13326 display scan finished message

ashiagr:
This PR displays scan finished message to address p5T066-1OP#comment-7218.
(Strings matched with [iOS app](https://github.com/wordpress-mobile/WordPress-iOS/pull/15921)).

No threats | One threat | Multiple threats
--|--|--
 device-2021-02-22-114234 | device-2021-02-22-113934 | device-2021-02-22-114447

To test:

1. Launch the app
2. Open a site with Jetpack Scan enabled
3. Perform a scan where the result is no threats
4. Verify the notice appears with "no threats found"
5. Perform a scan where the result is 1 threat
6. Verify the notice appears with "1 potential threat found"
7. Perform a scan where the result is more than 1 threat
8. Verify the notice appears with "X potential threats found" where X is the number of threats

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
```diff
@@ -97,12 +99,12 @@ class ScanViewModel @Inject constructor(
         }
     }
 
-    private fun fetchScanState(startWithDelay: Boolean = false, isRetry: Boolean = false) {
+    private fun fetchScanState(invokedByUser: Boolean = false, isRetry: Boolean = false) {
```
ParaskP7, line 102: Suggestion (💡): This is a suggestion for the future, but obviously just something you already doing well. I would have committed this `invokedByUser` change separately, as a `Refactor` commit. I would do that just to separate this commit from the `Feature` related commit just to make it easier for the reviewer to understand immediately that this has nothing to do with the actual feature itself.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
```diff
@@ -132,14 +136,36 @@ class ScanViewModel @Inject constructor(
...
+    private fun showScanFinishedMessage(scanStateModel: ScanStateModel) {
+        val threatsCount = scanStateModel.threats?.size ?: 0
+        val message = UiStringText(
+            when (threatsCount) {
+                0 -> htmlMessageUtils.getHtmlMessageFromStringFormatResId(
+                    R.string.scan_finished_no_threats_found_message
+                )
+
+                1 -> htmlMessageUtils.getHtmlMessageFromStringFormatResId(
+                    R.string.scan_finished_threats_found_message_singular,
```
ParaskP7: Suggestion (💡): Since this string will always be used with a `1` argument, being a `singular`. What if we remove the `%s` from within the string and replace that with a hardcoded `1` instead? I consider this way a bit more safer due to the fact that it will not be possible to mix this string afterwards with a value other that `1`, which I am guessing it is our intention.

ParaskP7 (approved):
👋 @ashiagr  !

I have left only a couple of suggestions (💡 ) for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
ashiagr: I agree, a separate commit must've been better here. I'll keep this in mind for the future. Thank you for this suggestion 🙏.

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/ScanViewModel.kt
ashiagr: Great suggestion, this will indeed be safer. In 3feccfe.

ashiagr:
Thanks for the review @ParaskP7! I've addressed the comments.
Ready for another look.

ParaskP7 (approved):
Thanks for the update @ashiagr , LGTM! 🚀

PR #14131, opened 2021-02-22
Detekt - Update 'LongParameterList' Configuration

ParaskP7:
This PR further configures [Detekt](https://github.com/detekt/detekt) to loosen the `LongParameterList` so that it accounts for dependency injection and ignores this rule when a constructor have parameters that are being injected, like on `ViewModel` or `UseCase`. As such, it makes is easier to work with the `LongParameterList` while not completely ignoring the rule.

PS: For more information read internal discussion here: `paqN3M-hh#comment-481`

To test:
- Pull the branch locally.
- Run `./gradlew WordPress:detekt` and make sure the task succeeds.

PR submission checklist:


ashiagr (approved):
Thanks @ParaskP7! It works as described 👍. 

Just a heads up, seems like detekt will allow us to [configure severity level per rule](https://git.io/Jt7Tk) [`error`, `warning`, `info`, `ignore`] in a future version. But it won't allow different severity levels within the same rule (e.g. constructor vs function params list within `LongParameterList` rule).

As discussed in the p2 post, we might have preferred showing a warning rather than ignoring long param list for constructors. Until `detekt` allows us to do this, I think we're good with these changes.

ParaskP7:
👋 @ashiagr !

> Just a heads up, seems like detekt will allow us to configure severity level per rule [error, warning, info, ignore] in a future version. 

Yeap, this will make configuring `Detekt` much suited to our needs, very nice finding, thanks for sharing! ❤️ 

> But it won't allow different severity levels within the same rule (e.g. constructor vs function params list within LongParameterList rule).

True, we can anyway try it out, see how we can work with it, whether we completely ignore those future `LongParameterList` warnings or not, and if so we can fall back to an `error` severity just to make it more strict. 💪 

> As discussed in the p2 post, we might have preferred showing a warning rather than ignoring long param list for constructors. Until detekt allows us to do this, I think we're good with these changes.

Again, thanks for taking the time to read the P2 post and our comments and thanks for finding this `Detekt` issue that they discuss about configuring the severity level. I agree, until `Detekt` allows us to do this, we can process with these changes and see how that works for us, then adjust accordingly. 🙏

planarvoid (approved):
👍

PR #14139, opened 2021-02-23
Add max width constraint to the My Site fragment layout

renanferrari:
This PR fixes the improved My Site fragment layout on tablets, by adding a max width constraint to it:

|Normal|Empty|
|---|---|
|normal|empty|

To test:

1. Login with an account that has no sites.
1. Go to the My Site screen.
1. Notice the empty view.
1. Notice the margins are correct.
1. Add a new site.
1. Notice the newly added site appears in the My Site screen.
1. Notice the margins are correct.

---

PR submission checklist:


osullivanchris:
Looks good @renanferrari . The only thing is, its a pity that the cards you made are limited to the fixed width. It feels like they should be able to flow across the whole screen. We obviously didn't account for something like this when we did it before. I did try to see if I could make the layout work, but neither of these are perfect.

- First one has the fixed width layout to the left rather than centre of the screen, and the carousel spills over
- The second one keeps it centred. Carousel spills over. Would start aligned left with the row, spilling over to the right, then I've showed it as the user scrolls - with content spilling over to the left too.

As I say neither are perfect. And probably not worth your while investing too much on tablet here. But just sharing the ideas.

tablet wide quick start

planarvoid:
I couldn't review this because I only saw an empty screen when logging in without a site, here's the fix - https://github.com/wordpress-mobile/WordPress-Android/pull/14149

I'll get back to this PR once it's merged @renanferrari

renanferrari:
@osullivanchris Thank you for testing this out and sharing your ideas, really appreciate it! Also sorry for the delay on this, I'm on my Groundskeeping rotation and was AFK last Thursday.

>The only thing is, its a pity that the cards you made are limited to the fixed width. It feels like they should be able to flow across the whole screen.

I totally agree with this.

>The second one keeps it centred. Carousel spills over. Would start aligned left with the row, spilling over to the right, then I've showed it as the user scrolls - with content spilling over to the left too.

I think this option feels a bit more natural for me, so I'd probably pick that over the first one. That said, I can see it being a bit involved to implement, so you're right that it's probably not worth tackling this right away. I've created an issue about this, so we don't lose track of it: #14172

renanferrari:
@planarvoid I've merged the fix and this is now ready for another round.

planarvoid (approved):
LGTM 👍  both with and without a site

PR #14140, opened 2021-02-24
Use tabs for comment filters

khaykov:
[This](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1888) Flux-C PR is a perquisite for this one.

This PR changes the filter of the Comment List from dropdown to Tabs.

While I modified and moved things around quite a lot, I have not changed the basic architecture of comment list.
I will leave this until we start comment unification, so we can decided on how to reuse the code and nicely wrap it all into use cases.

Things that I have not covered in this PR and that are present in existing implementation:
- Preserving action mode selection on configuration change.
- Preserving scroll position on configuration change.

While on surface not much things changed, I had to reorganize a lot of business logic to transition from single fragment pattern to ViewPager + Tabs.

In addition, I fixed an issue with missing empty views and status messages.

Originally not expecting that this issue will take so much work, I converted CommentsActivity to kotlin 😓 

[Image from Gyazo](https://gyazo.com/c0ed98d3ddd90d180da287e31bb746aa)

To test:
- Moderation from Comment Details.
- Batch moderation (long press on comments)
- Empty states.
- Configuration change.
- Load More.
- Smoke test, preferably comparing functionality to old version side by side.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/comments/CommentsListFragment.java
```diff
@@ -69,7 +76,7 @@
         SPAM(R.string.comment_status_spam),
         DELETE(R.string.comment_status_trash);
 
-        private final int mLabelResId;
+        public final int mLabelResId;
```
develric: Maybe we can consider keeping it private and adding a method like `getLabelResId` or even using the `getLabel` with the setText fun taking the string as input in `tab.setText(commentListFilters[position].mLabelResId)`, wdyt?

develric (commented):
Hi @khaykov 👋 ! Thanks for this PR 🙇 ! It looks good overall 👍 ; I would like to have another round of check/testing on it also given the refactor to Kotlin (btw thanks for tackling it 👍 ) and the changes in the CommentsListFragment but wanted to anticipate some thoughts I had in this first round. Let me know wdyt 🙇 .

Used OnePlus 7T Android 10

- When trashing multiple comments in batch, the UNDO action on the snackbar is not restoring any of the selected comments for me (it's working well selecting only one single comment). Not sure if we can easily batch undo from the snackbar tbh; let's discuss about it.
- If I trash a Pending comment, it seems we loose the pending state and restoring it actually approves it also
- When scrolling a long list of comments, I get the PTR indicator to appear together with the progress loading bar; maybe we should not show the PTR indicator in that case, wdyt?
- Not sure it's worth given we are now using tabs, but to check with you, we are not restoring the last selected comments tab when coming to the comments screen

khaykov:
Thanks for the review @develric ! Ready for another review 🙇  Also, for additional test steps, you can try to moderate comments on Calypso and confirm that the status reflects correctly in the app.

> When trashing multiple comments in batch, the UNDO action on the snackbar is not restoring any of the selected comments for me (it's working well selecting only one single comment). Not sure if we can easily batch undo from the snackbar tbh; let's discuss about it.

Yeah, snackbar should not be visible for batch trash, since we have a confirmation dialog :) Fixed it!

> If I trash a Pending comment, it seems we loose the pending state and restoring it actually approves it also

Yep, that's the limitations of endpoint, and also how it works on Calypso too :)

> Not sure it's worth given we are now using tabs, but to check with you, we are not restoring the last selected comments tab when coming to the comments screen

This is intentional 👍

develric (commented):
Hi @khaykov, tested this on a OnePlus 7T Android 10 phone. Checked the following:

✅  - Fluxc db fresh install and db script update
✅  - View Pager implementation and swipe between pages + pagination
✅  - Batch actions on comments; they were correctly moved to the relevant list and refreshed as expected
✅  - Same check as before but for single comment in comment details
✅  - Played with moderating comments in Calypso or the app and checking the status was kept consistent (and it was)

Also code LGTM 👍 

As agreed I'll wait the update of the FluxC tag and the release note (there seems to be a conflict right now) then we can merge 🎉

develric (approved):
LGTM :shipit:

PR #14147, opened 2021-02-24
jetpack: backup download notice banner

### LINKED ISSUE #14030: Jetpack: show a notice to download the backup
#### Downloading the file

1. When a Backup is available (created either from the web or the app)
2. The user can see a notice to download the file
3. When the user taps download the file is downloaded

#### Dismissing the notice

1. When a Backup is available (created either from the web or the app)
2. The user can see a notice to download the file
3. When the user taps dismiss the notice is dismissed and it doesn't appear again

### Design

In iOS was used an already existent component to show this notice:

I would recommend the same in Android (if such component exists), but that's up to @osullivanchris.

### Technical information

The list of files available to download can be obtained from `GET: /wpcom/v2/sites/<siteID>/rewind/downloads?_envelope=1`. Remember to check `validUntil` before displaying the notice.

If the user dismisses the notice, a `POST` request should be made to `/v2/sites/<siteID>/rewind/downloads/<downloadID>?_envelope=1`. This dismisses the notice in the app and in the web (so no need to persist in the client that information, next time you request the list of downloads it won't be there).

zwarm:
Fixes #14030 

This PR adds the Backup Download Banner to the activity log list and backup list.

Notable changes ( It may be easiest to review this PR by commits and I'm available if reviewer wants to pair review)
- `activity_log_list_notice_item.xml` - New layout file for the notice banner
- `styles.xml` & `strings.xml` - Added new styles for the banner and a few strings
- `ActivityLogAdapter`, `ActivitiyLogDiffCallback` - both updated to support the `Notice` list item type and the new `NoticeItemViewHolder`, so that the view will actually show in the list
- Updated `Complete` to include *url* and *validUntil* and `GetBackupDownloadStatusUseCase` to send those values downstream
- Created `PostDismissBackupDownloadUseCase` which is post the dismiss request to the server. If there is an error, we log it and move on.
- Four new tracking collections points were added to track button taps from the notice banner. Separate tracks for activity log list notice vs backup list notice. 
      JETPACK_BACKUP_DOWNLOAD_FILE_NOTICE_DOWNLOAD_TAPPED - 
      JETPACK_BACKUP_DOWNLOAD_FILE_NOTICE_DISMISSED_TAPPED,
      ACTIVITY_LOG_DOWNLOAD_FILE_NOTICE_DOWNLOAD_TAPPED,
      ACTIVITY_LOG_DOWNLOAD_FILE_NOTICE_DISMISSED_TAPPED
 - Added a couple of new function to `ActivityLogTracker` to make the tracks call reusable
 - Added a new navigation event for `DownloadBackupFile`, which will forward the user to the web browser for file download. Updated the observer in `ActivityLogListFragment` to handle the browser launch
 - Added a `JetpackBackupDownloadActionState` which will let the calling activity know which state the download is in on return. This was a precursor to showing the correct messaging when returning to the activity log list or backup list. Changes were made to `BackupDownloadFragment` and `BackupDownloadViewModel` to support this new intent extra.
 - The majority of the changes occur in `ActivityLogViewModel`, which include:
 - - Updating `backupDownloadEvent` with a *displayNotice* field, so when `reloadEvents` executes it uses this property to determine if the notice banner should be shown. Also included downloadId, url, and validUntil.
 - - Removed the showBackupDownloadFinishedMessage from within reloadEvents, the snackbar is replaced with the notice banner
 - - Added a new function to build the notice list item
 - - Added two new functions to handle the *download* and *dismiss* button clicks from within the notice banner 
 - - Updated `onQueryBackupDownloadStatus` to use the JetpackBackupDownloadActionState to determine which snackbar message to be shown. Note that in the case of returning to the lists, we do want to show the snackbar in addition to the notice banner.
 - - Updated `handleBackupDownloadStatusForProgress` to no longer check if the progress item is shown, we always kick off a reloadEvents
 - - Updated `handleBackupDownloadStatusForComplete` to reloadEvents of the progressItem is not shown
 - Updated and/or added tests

**Merge Instructions:**
- Still need to decide which milestone to base against, so please do not merge yet
- Review [FluxC PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1891)  merge and create a tag in FluxC develop
- Replace FluxC hash with tag from develop in build.gradle
- Remove the "Not Ready for Merge label"
- Merge this PR

|light |dark|notification
|---|---|--|
|light|dark|backup-download

To test: 
#### Scenario.1 : BackupDownload Flow - Manual Exit
* Open app
* Find a rewindable item from the list, click on the ellipsis menu item and select `Download backup` to launch the `Backup Download` screen,
* Click the `Create downloadable file` button.
* Navigate back manually (by clicking the `X` toolbar button or pressing/swiping back).
* You should automatically be navigated back to the `Activity Log` screen and a backup status will be triggered,
* The backup progress item should immediately appears. 
* A snackbar should immediately appear informing that the backup just started. 
* Wait for backup to finish while having the `Activity Log` screen 
* When backup finishes
 - The backup progress item should disappear
 - The backup download notice should appear

#### Scenario.2 : BackupDownload Flow - Wait for flow to end and then exit
* Open app
* Find a rewindable item from the list, click on the ellipsis menu item and select `Download backup` to launch the `Backup Download` screen,
* Click the `Create downloadable file` button.
* Wait for the backup download flow to complete keeping the backup flow process open
* After the backup completes, navigate back manually (by clicking the `X` toolbar button or pressing/swiping back)
 - A snackbar should immediately appear informing that the backup is complete
 - The backup download notice should appear

#### Scenario.3 (Notice banner download)
- Run a backup using scenario 1 or 2 above
- Click on the "Download" button within the notice banner
- Note that a browser is launched and the download starts. In some cases, you may be prompted to verify the download.

#### Scenario.4 (Notice banner dismiss)
- Run a backup using scenario 1 or 2 above
- Click on the "Dismiss" button within the notice banner
- Note the the Notice banner disappears

PR submission checklist:


ParaskP7:
👋 @zwarm !

I have some questions on the second scenario: `Scenario.2 : BackupDownload Flow - Wait for flow to end and then exit`
1) `When backup finishes`: This is just a typo, correct?
1) `A snackbar should immediately appear informing that the backup is complete`: Since in the first scenario, this snackbar is not displayed, we might probably want to do the same here, meaning never showing the finished message?

ParaskP7:
👋 @zwarm !

I also tested a fifth scenario, given that was on the video you shared, opening the `Download Backup` notification and clicking the `site's Backups` link for both:
1) When the `Backup Download` notice is showing (not previously dismissed), and
1) When the `Backup Download` notice was dismissed.

Everything works as expected, kudos! 🌟

osullivanchris (approved):
I went through the 4 flows and tested on tablet and dark mode. All looks exactly as expected, nice work 👌 😃

zwarm:
> 👋 @zwarm !
> 
> I have some questions on the second scenario: `Scenario.2 : BackupDownload Flow - Wait for flow to end and then exit`
> 
> 1. `When backup finishes`: This is just a typo, correct?

Yes 🤦 

> 2. `A snackbar should immediately appear informing that the backup is complete`: Since in the first scenario, this snackbar is not displayed, we might probably want to do the same here, meaning never showing the finished message?

I decided to leave it because we are returning from another screen, where as in scenario 1 we are already in the context of the list. I could just as easily remove it here too. We need a tie breaker.

ParaskP7:
👋 @zwarm !

> I decided to leave it because we are returning from another screen, where as in scenario 1 we are already in the context of the list. I could just as easily remove it here too. We need a tie breaker.

Actually, thinking about it more and having you explained it it makes sense, let's leave it as is for now! 🙏

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/PostDismissBackupDownloadUseCase.kt
```diff
@@ -0,0 +1,54 @@
...
+            }
+            val result = activityLogStore.dismissBackupDownload(DismissBackupDownloadPayload(site, downloadId))
+            if (result.isError) break
+            dismissed = true
+            break
+        }
+        dismissed
+    }
+
+    private suspend fun handleError(
```
ParaskP7, line 43: Suggestion (💡): I have managed to make this a bit more compact. Feel free to update it to this new version or ignore:
```
    private suspend fun handleError(retryAttempts: Int) = if (retryAttempts >= MAX_RETRY) {
        AppLog.d(T.JETPACK_BACKUP, "$tag: Exceeded $MAX_RETRY retries while dismiss download backup file")
        true
    } else {
        delay(DELAY_MILLIS * max(1, DELAY_FACTOR * retryAttempts))
        false
    }
```

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadFragment.kt
```diff
@@ -126,13 +128,22 @@ class BackupDownloadFragment : Fragment(R.layout.jetpack_backup_restore_fragment
         viewModel.wizardFinishedObservable.observe(viewLifecycleOwner, {
             it.applyIfNotHandled {
                 val intent = Intent()
-                val (backupDownloadCreated, ids) = when (this) {
-                    is BackupDownloadCanceled -> Pair(false, null)
-                    is BackupDownloadInProgress -> Pair(true, Pair(rewindId, downloadId))
-                    is BackupDownloadCompleted -> Pair(true, null)
+                val (backupDownloadCreated, ids, actionType) = when (this) {
+                    is BackupDownloadCanceled -> Triple(false, null, JetpackBackupDownloadActionState.CANCEL)
```
ParaskP7, line 132: Suggestion (💡): Since all of those when entries are multiline, I suggest that become a multiline as well, just for someone to be able to see the diffs straight away. Wdyt? Something like:
```
is BackupDownloadCanceled -> Triple(
        false,
        null,
        JetpackBackupDownloadActionState.CANCEL
)
```

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
```diff
@@ -441,10 +454,15 @@ class BackupDownloadViewModel @Inject constructor(
         @Parcelize
         data class BackupDownloadInProgress(
             val rewindId: String,
-            val downloadId: Long
+            val downloadId: Long,
+            val inProgress: Boolean = true
```
ParaskP7, line 458: Question (❓): As far as I understand this `inProgress` and the other one further below are unused, isn't it?

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -73,7 +77,8 @@ class ActivityLogViewModel @Inject constructor(
     private val dateUtils: DateUtils,
     private val activityLogTracker: ActivityLogTracker,
     private val jetpackCapabilitiesUseCase: JetpackCapabilitiesUseCase,
-    private val restoreFeatureConfig: RestoreFeatureConfig
+    private val restoreFeatureConfig: RestoreFeatureConfig,
+    private val postDismissBackupDownloadUseCase: PostDismissBackupDownloadUseCase
```
ParaskP7, line 81: Minor (🔍): Since this is a backup related use case, consider grouping that below the other backup related use case further above. Wdyt?

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -150,7 +157,7 @@ class ActivityLogViewModel @Inject constructor(
     var rewindableOnly: Boolean = false
 
     private var currentRestoreEvent = RestoreEvent(false)
-    private var currentBackupDownloadEvent = BackupDownloadEvent(false)
+    private var currentBackupDownloadEvent = BackupDownloadEvent(false, displayNotice = false)
```
ParaskP7, line 160: Suggestion (💡): Since a named parameter is now used, I suggest updating all of them to match that pattern, like:
```
    private var currentRestoreEvent = RestoreEvent(displayProgress = false)
    private var currentBackupDownloadEvent = BackupDownloadEvent(displayProgress = false, displayNotice = false)
```

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -194,6 +203,12 @@ class ActivityLogViewModel @Inject constructor(
         if (withBackupDownloadProgressItem) {
             items.add(getBackupDownloadProgressItem(backupDownloadEvent.rewindId, backupDownloadEvent.published))
         }
+        if (withBackupDownloadNoticeItem) {
```
ParaskP7, line 206: Warning (⚠️): Consider adding a test for this case.

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -223,8 +238,7 @@ class ActivityLogViewModel @Inject constructor(
             currentRestoreEvent = RestoreEvent(false)
         }
         if (backupDownloadEvent.isCompleted) {
```
ParaskP7, line 240: 1) Question (❓): Why keeping the state is needed here?
2) Question (❓): If keeping the state is necessary, will copying the current state to `displayProgress = false` provide any value? This is due to the fact that the above `currentBackupDownloadEvent = backupDownloadEvent` will already be doing that for us, will it not?
3) Question (❓): If all my above assumptions are correct, probably this whole if logic can be removed altogether.
4) Warning (⚠️): Consider updating the corresponding tests no matter which approach is to be taken.

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -265,6 +279,23 @@ class ActivityLogViewModel @Inject constructor(
...
+    private fun getBackupDownloadNoticeItem(backupDownloadEvent: BackupDownloadEvent): ActivityLogListItem.Notice? {
+        val rewindDate = backupDownloadEvent.published
+                ?: backupDownloadEvent.rewindId?.let { activityLogStore.getActivityLogItemByRewindId(it)?.published }
+        return rewindDate?.let {
+            ActivityLogListItem.Notice(
+                    label = resourceProvider.getString(
+                            R.string.activity_log_backup_download_notice_description_with_two_params,
+                            rewindDate.toFormattedDateString(), rewindDate.toFormattedTimeString()
+                    ),
+                    primaryActionButtonClickListener = this@ActivityLogViewModel::onDownloadBackupFileClicked,
```
ParaskP7, line 291: Suggestion (💡): What if those two actions where to be written like the below:
```
primaryAction = { onDownloadBackupFileClicked(backupDownloadEvent.url as String) },
secondaryAction = {  dismissNoticeClicked(backupDownloadEvent.downloadId as Long) }
```

Wouldn't that remove some of the boilerplate? No need for explicitly passing the `url` and `downloadId` to the item itself.

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -511,6 +542,21 @@ class ActivityLogViewModel @Inject constructor(
         return true
     }
 
+    private fun onDownloadBackupFileClicked(url: String) {
```
ParaskP7, line 545: Warning (⚠️): Consider adding a test for the tracking trigger.

PS: The same apply for the `dismissNoticeClicked(...)` function below.

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -511,6 +542,21 @@ class ActivityLogViewModel @Inject constructor(
...
+        activityLogTracker.trackDownloadBackupDownloadButtonClicked(rewindableOnly)
+        _navigationEvents.value = Event(ActivityLogNavigationEvents.DownloadBackupFile(url))
+    }
+
+    private fun dismissNoticeClicked(downloadId: Long) {
+        activityLogTracker.trackDownloadBackupDismissButtonClicked(rewindableOnly)
+        dismissBackupDownloadJob?.cancel()
+        dismissBackupDownloadJob = viewModelScope.launch {
```
ParaskP7, line 553: Question (❓): Is canceling the job needed here? Probably not, right? If that is indeed the case I believe this can be updated to the below:
```
viewModelScope.launch { postDismissBackupDownloadUseCase.dismissBackupDownload(downloadId, site) }
reloadEvents(backupDownloadEvent = BackupDownloadEvent(displayNotice = false, displayProgress = false))
```

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -628,9 +674,14 @@ class ActivityLogViewModel @Inject constructor(
         }
     }
 
-    fun onQueryBackupDownloadStatus(rewindId: String, downloadId: Long) {
+    fun onQueryBackupDownloadStatus(rewindId: String, downloadId: Long, actionState: Int) {
```
ParaskP7, line 677: Warning (⚠️): Consider adding and updating any tests related to this change.

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -650,29 +701,32 @@ class ActivityLogViewModel @Inject constructor(
...
-            reloadEvents(
-                    backupDownloadEvent = BackupDownloadEvent(
-                            displayProgress = true,
-                            isCompleted = false,
-                            rewindId = state.rewindId,
-                            published = state.published
-                    )
-            )
-        }
+        reloadEvents(
```
ParaskP7, line 704: Blocker (🚫): By removing this if logic here, the UI will end up scrolling on top every 1 second, while the backup download is in progress and up until it completes.

Why was this removed in the first place?  I understand why the if logic was updated on the `Complete` status scenario, but I believe this is not necessary here, isn't it?

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -650,29 +701,32 @@ class ActivityLogViewModel @Inject constructor(
...
+                        displayProgress = true,
+                        displayNotice = false,
+                        isCompleted = false,
+                        rewindId = state.rewindId,
+                        published = state.published
+                )
+        )
     }
 
     private fun handleBackupDownloadStatusForComplete(state: BackupDownloadRequestState.Complete) {
```
ParaskP7, line 715: Warning (⚠️): Consider adding a test for the new logic added here.

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
```diff
@@ -650,29 +701,32 @@ class ActivityLogViewModel @Inject constructor(
...
+        val backupDownloadEvent = BackupDownloadEvent(
+                displayProgress = false,
+                displayNotice = isBackupDownloadFileValid(state.url, state.validUntil, state.downloadId),
```
ParaskP7, line 718: Suggestion (💡): Maybe this can be moved to the `GetBackupDownloadStatusUseCase` and be tested there as well? Then, this will be passed to the `BackupDownloadRequestState.Complete` state as `isValid`, above the current `validUntil` field?
```
displayNotice = state.isValid,
```

WordPress/src/main/res/layout/activity_log_list_notice_item.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="utf-8"?>
```
ParaskP7, line 1: Minor (🔍): Consider reformatting this file.

WordPress/src/main/res/layout/activity_log_list_notice_item.xml
```diff
@@ -0,0 +1,54 @@
...
+    android:layout_height="wrap_content">
+
+    <com.google.android.material.textview.MaterialTextView
+        android:id="@+id/label"
+        style="@style/ActivityLogList.Notice.Item.Label"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintBottom_toTopOf="@+id/button_wrapper"
+        tools:text="We successfully created a backup download file from October 21, 2020 at 3:40 AM."
```
ParaskP7, line 15: Suggestion (💡): Since the `android:text` is already set here, I suggest removing the `tools:text` and keep the text from the `android:text` itself. I understand that this text has a couple of params, which can seems weird in the preview, but I still believe that this is better since the `@string` can change in due time and we can have an outdated preview. Actually it is already kind of out of sync.

PS: Same applies for all other places where `tools:text` is used in this file (`Download` and `Dismiss` buttons).

WordPress/src/main/res/layout/activity_log_list_notice_item.xml
```diff
@@ -0,0 +1,54 @@
...
+        android:text="@string/activity_log_backup_download_notice_description_with_two_params"/>
+
+    <androidx.constraintlayout.widget.ConstraintLayout
```
ParaskP7, line 18: Question (❓): Couldn't the same result be achieved with only one layer of `ConstraintLayout`?

WordPress/src/main/res/values/styles.xml
```diff
@@ -1540,4 +1540,45 @@
     <style name="WordPress.Button.Invite.DisruptiveAction" parent="WordPress.Invite.Button.Secondary">
         <item name="android:textColor">?attr/wpColorError</item>
     </style>
+
+    <!-- ActivityLog List Backup Download Notice Item -->
+    <style name="ActivityLogList.Notice.Item.Label" parent="android:Widget.TextView">
```
ParaskP7, line 1545: Suggestion (💡): I usually suggest that the `layout_width` and `layout_height` to not be part of a style since they are totally dependent on the individual screen and how a specific element is positions in it. Also, I consider the `layout_margin` to be included in this category. However, the `padding` can be indeed considered part of the style.

PS: I see both approached being used on our styles, as such I am okay with anyway you decide to proceed with that. This was just me sharing my thoughts since I would have done it the way I described it.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/PostDismissBackupDownloadUseCaseTest.kt
```diff
@@ -0,0 +1,91 @@
...
+                        causeOfChange = DISMISS_BACKUP_DOWNLOAD
+                )
+        )
+
+        val result = useCase.dismissBackupDownload(downloadId, siteModel)
+
+        assertThat(result).isEqualTo(true)
+    }
+
+    private val downloadId = 100L
```
ParaskP7, line 90: Minor (🔍): Consider moving that on top, closer to all the other fields.

WordPress/src/test/java/org/wordpress/android/ui/jetpack/backup/download/usecases/PostDismissBackupDownloadUseCaseTest.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    @Test
+    fun `given generic error response, when dismiss download is triggered, then false is returned`() = test {
+        whenever(activityLogStore.dismissBackupDownload(any()))
+                .thenReturn(
+                        OnDismissBackupDownload(
+                                downloadId, DismissBackupDownloadError(
+                                GENERIC_ERROR
+                        ), DISMISS_BACKUP_DOWNLOAD
+                        )
```
ParaskP7, line 67: Minor (🔍): Consider updating the format of this `whenever/thenReturn` to the blelow:
```
whenever(activityLogStore.dismissBackupDownload(any())).thenReturn(
        OnDismissBackupDownload(
                downloadId, DismissBackupDownloadError(GENERIC_ERROR),
                DISMISS_BACKUP_DOWNLOAD
        )
)
```

PS: The same applies for the test further below.

WordPress/src/test/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModelTest.kt
```diff
@@ -1606,23 +1659,14 @@ class ActivityLogViewModelTest {
...
-                            eq(R.string.activity_log_backup_finished_snackbar_message),
-                            any(),
-                            any()
-                    )
-            ).thenReturn(BACKED_UP_DATE_TIME)
-        } else {
-            whenever(resourceProvider.getString(R.string.activity_log_backup_finished_snackbar_message_no_dates))
-                    .thenReturn(BACKED_UP_NO_DATE)
-        }
+                .thenReturn(BACKUP_NOTICE)
```
ParaskP7, line 1670: Minor (🔍): Consider moving that on top, next to the other line, like:
```
).thenReturn(BACKUP_NOTICE)
```

ParaskP7 (changes requested):
👋 @zwarm !

I have reviewed and tested this PR as per the instructions, everything almost works as expected, amazing job dealing with this huge fat task! 🌟 🌟 🌟

I have left some minor (🔍) comments and suggestions (💡) for you to consider. Also, I have left a handful of questions (❓) I would like to discuss with you. Most importantly, I have left some warnings (⚠️), but mainly around missing tests. Finally, there is one blocker (🚫) that I would really like us to fix before merging this to `develop`.

🙏

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/BackupDownloadViewModel.kt
zwarm: @ParaskP7 I noticed this when cherry picking and removed it already. 👍

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
zwarm: I caught this one too in the cherry pick

WordPress/src/main/java/org/wordpress/android/ui/jetpack/backup/download/usecases/PostDismissBackupDownloadUseCase.kt
zwarm: Good suggestion, incorporated into cherry picked branch. Thanks

WordPress/src/main/res/values/styles.xml
zwarm: Going to leave this as is for this PR, as there are multiple approaches to using styles. We should have a follow up on future direction overall. Thanks

WordPress/src/main/res/layout/activity_log_list_notice_item.xml
zwarm: Going to leave this as is for now, if I have time I will revisit

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
zwarm: Removed in the cherry pick PR

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
zwarm: Removed as discussed. 🙇

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
zwarm: Updated as discussed

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
zwarm: Added back as discussed. Thanks

WordPress/src/main/java/org/wordpress/android/viewmodel/activitylog/ActivityLogViewModel.kt
zwarm: Moved as discussed

zwarm:
@ParaskP7  - This PR was closed and replaced with #14155

WordPress/src/main/res/values/styles.xml
ParaskP7: Thanks for considering it @zwarm ! 🌟 

Cc @ashiagr here so that we can setup a quick call on that to first discuss it amongst us. 🙏

PR #14149, opened 2021-02-25
Fix empty screen when logging in without a site

planarvoid:
When logging in with an account without a site the app didn't show the empty view. The reason for this bug was that the `siteSelected` didn't emit the null after the change we did when it was set up in `map`. I've changed the initialization so it's now correctly set up with `null` value and added a unit test for it.

Before: 

Screenshot_1614178005

After: 

Screenshot_1614243465

To test:
- Log in with an account without a site
- Turn on the My Site Improvements flag and restart the app
- Go to "My site"
- The empty view is visible and the "Add new site" button is there

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/mysite/SelectedSiteRepository.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/mysite/SelectedSiteRepository.kt
@@ -21,7 +21,7 @@ class SelectedSiteRepository
     private val _selectedSiteChange = MutableLiveData<SiteModel>()
     private val _showSiteIconProgressBar = MutableLiveData<Boolean>()
     val selectedSiteChange = _selectedSiteChange as LiveData<SiteModel>
-    val siteSelected: LiveData<Int> by lazy {
+    val siteSelected: LiveData<Int?> by lazy {
         val result = MediatorLiveData<Int>()
         result.addSource(_selectedSiteChange) { site -> result.value = site?.id }
         result.value = null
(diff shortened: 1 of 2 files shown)


WordPress/src/main/java/org/wordpress/android/ui/mysite/SelectedSiteRepository.kt
```diff
@@ -21,7 +21,12 @@ class SelectedSiteRepository
     private val _selectedSiteChange = MutableLiveData<SiteModel>()
     private val _showSiteIconProgressBar = MutableLiveData<Boolean>()
     val selectedSiteChange = _selectedSiteChange as LiveData<SiteModel>
-    val siteSelected = _selectedSiteChange.map { it?.id }.distinctUntilChanged()
+    val siteSelected: LiveData<Int?> by lazy {
+        val result = MediatorLiveData<Int>()
+        result.addSource(_selectedSiteChange) { site -> result.value = site?.id }
+        result.value = null
+        result.distinctUntilChanged()
+    }
```
renanferrari: If emitting `null` as the very first value solves this issue, maybe we could just keep the `siteSelected` logic as it was and have the default value of `_selectedSiteChange` be `null`, like this:

```kotlin
  private val _selectedSiteChange = MutableLiveData<SiteModel?>(null)
  private val _showSiteIconProgressBar = MutableLiveData<Boolean>()
  val selectedSiteChange = _selectedSiteChange as LiveData<SiteModel?>
  val siteSelected = _selectedSiteChange.map { it?.id }.distinctUntilChanged()
```

I've tested the code above just to make sure and, as far as I can see, everything seems to work as expected (including the new test you added). Let me know what you think!

renanferrari (commented):
Thanks a lot for looking into this issue and for figuring it out! I've tested the app and can confirm it now works as expected. Code looks good, but I made one suggestion below that I think could make the solution a bit cleaner.

planarvoid:
Good suggestion @renanferrari , I've used your code 👍

renanferrari (approved):
LGTM :shipit:

PR #14151, opened 2021-02-25
View milestone notifications

planarvoid:
Related FluxC PR - https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1893

This PR adds changes to show the new view milestone notification view differently in the notification detail. The changes are quite small. I've added a new material button (which we can later use to replace the button in the layout). The current code is quite fragile so I tried to put everything behind the `isViewMilestone` check. 

I think the icon is not visible yet in the list because the icon is photonized and this makes it not load from the sandbox. I think this will work normally when the notification is released. We can double check that. All the icons are photonized and work correctly.

To test:
- Setup sandbox and create a view milestone notification
- Go to notifications
- Notice the notification in the list
- Click on the notification
- Notice the number is animated
- Notice the confetti
- Notice the bigger text
- Click on the blog name
- Notice the blog is opened
- Go back
- Click on the "Take a deep dive"
- Notice you're redirected to your blog stats 

https://user-images.githubusercontent.com/1079756/109144871-0a67b080-7762-11eb-8ad0-f66fb6e955da.mp4

https://user-images.githubusercontent.com/1079756/109144879-0d62a100-7762-11eb-8787-d8c721f84330.mp4

PR submission checklist:


renanferrari (commented):
Thanks for the changes, this is looking pretty good! 🎊 Using Lottie really was a great choice 👍

I did find two issues while testing the app, although I'm not sure if they are caused by these changes of if it's due my own sandbox setup:

1. The button style doesn't seem to be correctly applied:



2. The confetti animation is happening for other notification types:

https://user-images.githubusercontent.com/830056/111392188-d0b90400-8694-11eb-9544-c2cee8964b09.mp4

planarvoid:
Thanks for the check @renanferrari . The different button is shown because there was a change in the API response. That's fixed now with ff39f57. The confetti was shown in wrong places because we were loading more fragments in the view pager and the confetti was shown in `onActivityCreated`. I've moved it to `onResume` in 9e34bdf. Both issues should work ok now 👍

renanferrari (approved):
Thanks for the changes! Everything seems to be working fine now 🙂

LGTM 👍

Issue #14137, opened 2021-02-23
Likes Enhancements in notifications


develric:
We would like to enhance the current visualization of like-ers in notifications as per below flow image.

![image](https://user-images.githubusercontent.com/47797566/108850080-39a3e380-75e3-11eb-806b-30e91c4b2a96.png)

A few tasks as per current understanding of the ticket:

- [x] Create feature flag [#14361]
- [x] Bring list of like-ers to use two line list with avatar material guideline [#14361]
- [x] Current implementation limit number of items to 10 client side and to 90 server side (at least for posts). Eliminate the 10 limit items client side (server side will be managed later) [#14361, #14477, #14555]
- [x] Create the user profile bottom sheet to be visualized when user taps on a like-ers item [#14437, #14456]
- [x] Show tapped site feed like in reader site feeds when user tap on the like-er site when available [#14361]
- [x] Add analytics [#14525]
- [x] Add Unit testing [#14532, #14541]
- [x] Remove feature flag and release [#14563]

abdi5:
kedirabdi585@gmail.com

designsimply:
Gah! I love this so much!

PR #14161, opened 2021-03-01
Refactor NotificationsDetailListFragment to kotlin

develric:
As part of the #14137 this PR moves the NotificationsDetailListFragment to kotlin.

The intention was only to move the fragment to kotlin, without affecting the feature, so the behaviour should have remained the same as before.

A few notes:
- Started the translation from AS auto conversion tool
- AsyncTask was kept for now and the update value (that seems to be not used) was moved from Boolean to Void
- again inside the AsyncTask, AS auto translated into a while loop the `for (int i = 0; i < bodyArray.length(); i++) {` and it seems consistent to me
- One thing I would like to get attention is the variable nullability moving to kotlin, sometimes could be tricky to determine if a variable should be managed as nullable or not, given that kotlin would crash the app if a null value ends up to be passed to a not nullable variable/parameter.

## To test
Having 2 Emulators with the develop branch and this one branch, compare the behaviour of the NotificationsDetailListFragment and check it behaves the same. Some scenarios I could think of:

- From Notifications select each of the filtered list type and open notification details 
- Open notification from automatcher
- Trigger a notification (like by commenting on a post) and open the notification details by tapping on the Android system notification
- Smoke test notifications approving/spamming/liking etc...better if having 2 emulators/devices to compare behaviours side by side

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/notifications/NotificationsDetailListFragment.kt
```diff
@@ -0,0 +1,629 @@
...
+            handleNoteBlockSpanClick(activity as NotificationsDetailActivity, clickedSpan)
+        }
+
+        override fun showDetailForNoteIds() {
+            if (!isAdded || notification == null || activity !is NotificationsDetailActivity) {
+                return
+            }
+            val detailActivity = activity as NotificationsDetailActivity
+
+            requireNotNull(notification).let { note ->
```
khaykov, line 201: `requireNotNull` check is not present in Java file, and I think we might not need it here, since we already check if `notification == null`.

I see this pattern couple of times in code.

khaykov (approved):
Conversion looks pretty good to me, @develric ! I compared some methods side by side with Java and they look good. Thanks for extracting overwhelming logic from `doInBackground`, btw!

I did the general smoke testing (everything was good), but was not able to test push notifications - can't get them to work on emulators for some reason (any ideas?), and don't have a real device at hand at the moment.

I also tested a pingbacks, and from my understanding we always should have a pingback url if the notification is Pingback.

One last thing - I noticed is that we still have couple of variable names starting with `m`.

WordPress/src/main/java/org/wordpress/android/ui/notifications/NotificationsDetailListFragment.kt
develric: AFAIU being those mutable member variables, we need at least to assert them as not null and for example assign them to a local variable even after the check previously done; something like:

```
val localNote : Note = notification!!
```

The usage of the requireNotNull seemed to me a bit more synthetic maybe 🤷 ; not feeling strong about, or if I'm missing anything so let me know wdyt 😄

develric:
Hi @khaykov really thanks for taking the time to look into this one 🙇 ! 

>One last thing - I noticed is that we still have couple of variable names starting with m.

🤦‍♂️ ! Totally missed those, changed in 3c7e98e , thanks!

>I did the general smoke testing (everything was good), but was not able to test push notifications - can't get them to work on emulators for some reason (any ideas?), and don't have a real device at hand at the moment.

If you are building from sources, sometimes I had every sort of issues often related to the local configuration files used for the builds (updating them to last versions helped if that's the case); sometimes I was lukier using Wasabi and other times all seemed to being fixed removing all app versions from the emulator (and cold restarting it), installing one apk from the [releases](https://github.com/wordpress-mobile/WordPress-Android/releases) page and then installing side by side my build flavor like wasabi it started to work (a bit empiric I know 😄 ). Have you tried using the CI built version apk as well? Let me know 🙇 

PS: replied inline for the requireNotNull thing 🙇

WordPress/src/main/java/org/wordpress/android/ui/notifications/NotificationsDetailListFragment.kt
khaykov: Ah, right, it's not only about null 👍 `requireNotNull` is fine with me :)

khaykov (approved):
Thanks @develric ! Notifications worked with CI build 👍 LGTM 🚢

PR #14179, opened 2021-03-03
Start QS only for the newly created site

### LINKED ISSUE #14178: Do not start quick start for the actually selected site
If you have site that's elligible for quick start after a fresh install, if you create a new site, the QS gets started also for the previously selected site. This is because we're starting the QS too early before the newly created site is selected.

I can share an account on which this happens persistently.

Steps to repro:
- Make sure you're on a fresh install with at least one site 
- Turn on the My Site Improvements flag and restart the app 
- Notice the site does not have QS started
- Create a new site
- Notice the new site does have QS started
- Switch site to the previous site
- Notice the previous site has QS started as well

planarvoid:
Fixes #14178 

This PR removes the double call to the `passActivityResultToMySiteFragment`. In addition instead of starting the QS for the selected site, I'm passing the created site ID to the function. I think this approach should be much less error prone. Let me know what you think! 

To test:
- Make sure you're on a freshly installed app with a site
- Turn on the My Site Improvements flag and restart the app
- Create a site 
- Notice the new site has QS started
- Switch site to the previous site
- Notice the previous site doesn't have QS started

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/WPMainActivity.java
@@ -1060,7 +1060,6 @@ public class WPMainActivity extends LocaleAwareActivity implements
                 }
                 break;
             case RequestCodes.CREATE_SITE:
-                passOnActivityResultToMySiteFragment(requestCode, resultCode, data);
                 QuickStartUtils.cancelQuickStartReminder(this);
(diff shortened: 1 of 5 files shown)


renanferrari (approved):
Thanks for looking into this and for coming up with a solution! The code is looking good and I've tested the app with the feature flag both enabled and disabled and everything seems to be working as expected.

LGTM :shipit:

PR #14021, opened 2021-02-10
Issue/13326 pluralisation of strings based on threats count

ashiagr:
Parent #13326

This PR 

1. Handles pluralisation of strings based on threats count:

Key | Message | Commit
-- | -- | --
threat_fix_all_warning_message | Please confirm you want to fix %s active threats. | aa0e8dba
threat_fix_all_status_success_message | Threats were successfully fixed. | 4c12e1d2
scan_fixing_threats_description | We\'re hard at work in the background fixing these threats. In the meantime feel free to continue to use your site as normal, you can check the progress at anytime. | f21b4ef8
scan_fixing_threats_title | Fixing Threats | 34058f18
scan_idle_with_threats_description | The scan found %1s potential threats with %2s. Please review them below and take action or tap the fix all button. We are %3$s if you need us. | a3500cfb
threats_found | Threats found | cb1a5239

Does not make any changes to below messages:

Key | Message | Reason
-- | -- | --
threat_fix_all_warning_title | Fix all threats | Corresponds to Fix All button, no singular string found in [Calypso](https://opengrok.a8c.com/source/xref/calypso/client/components/jetpack/fix-all-threats-dialog/index.tsx?r=43d3beb3#36)
threat_fix_all_error_message | Error fixing threats. Please contact our support. | No singular string found in [Calypso](https://opengrok.a8c.com/source/xref/calypso/client/state/data-layer/wpcom/sites/scan/threats/fix-all-threats.js?r=8bda004c#57) 

2. Updates messages based on editorial review:

Key | Message | Commit
-- | -- | --
scan_fixing_threats_description (Issue #14008) | We\'re hard at work fixing these threats in the background. In the meantime feel free to continue to use your site as normal, you can check the progress at anytime. | a52d9aad
threat_ignore_warning (`p1612896373162300-slack-C0180B5PRJ4`)  | You shouldn\’t ignore a security unless you are absolute sure it\’s harmless. If you choose to ignore this threat, it will remain on your site &lt;b&gt;%s&lt;/b&gt;. | cab95b22

**Notes**

1. Due to translation issues with using [quantity plurals](https://developer.android.com/guide/topics/resources/string-resource#Plurals), I just duplicated messages in the `strings.xml`. Let me know if you have a better suggestion.

[...]

PR #14181, opened 2021-03-03
Update singular strings for threats to use 1

JavonDavis:
The Romanian translators are having trouble with the string formatting before the singular noun, This PR changes that to explicitly use 1. 

To test: Test that with a single threat it behaves as expected in https://github.com/wordpress-mobile/WordPress-Android/pull/14021

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/builders/ScanStateListItemsBuilder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/builders/ScanStateListItemsBuilder.kt
@@ -256,13 +256,10 @@ class ScanStateListItemsBuilder @Inject constructor(
     ): DescriptionState {
         val clickableText = resourceProvider.getString(R.string.scan_here_to_help)
 
-        val descriptionTextResId = if (threatsCount > 1) {
-            R.string.scan_idle_threats_description_plural
-        } else R.string.scan_idle_threats_description_singular
         val descriptionText = if (threatsCount > 1) {
             htmlMessageUtils
                     .getHtmlMessageFromStringFormatResId(
-                            descriptionTextResId,
+                            R.string.scan_idle_threats_description_plural,
                             "<b>$threatsCount</b>",
                             "<b>${site.name ?: resourceProvider.getString(R.string.scan_this_site)}</b>",
                             clickableText
@@ -270,7 +267,7 @@ class ScanStateListItemsBuilder @Inject constructor(
         } else {
             htmlMessageUtils
                     .getHtmlMessageFromStringFormatResId(
-                            descriptionTextResId,
+                            R.string.scan_idle_threats_description_singular,
                             "<b>${site.name ?: resourceProvider.getString(R.string.scan_this_site)}</b>",
                             clickableText
                     )
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/builders/ScanStateListItemsBuilder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/builders/ScanStateListItemsBuilder.kt
@@ -258,8 +258,8 @@ class ScanStateListItemsBuilder @Inject constructor(
 
         val descriptionTextResId = if (threatsCount > 1) {
             R.string.scan_idle_with_threats_description_plural
-        } else R.string.scan_idle_with_threats_description_singular
-        val descriptionText = if(threatsCount > 1) {
+        } else R.string.scan_idle_threats_description_singular
+        val descriptionText = if (threatsCount > 1) {
             htmlMessageUtils
                     .getHtmlMessageFromStringFormatResId(
                             descriptionTextResId,
--- a/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/builders/ScanStateListItemsBuilder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/builders/ScanStateListItemsBuilder.kt
@@ -257,7 +257,7 @@ class ScanStateListItemsBuilder @Inject constructor(
         val clickableText = resourceProvider.getString(R.string.scan_here_to_help)
 
         val descriptionTextResId = if (threatsCount > 1) {
-            R.string.scan_idle_with_threats_description_plural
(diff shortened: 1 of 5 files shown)


WordPress/src/main/res/values/strings.xml
```diff
@@ -1095,7 +1095,7 @@
     <string name="scan_idle_manual_scan_description">To review your site again run a manual scan, or wait for Jetpack to scan your site later today.</string>
     <string name="scan_idle_last_scan_description">The last Jetpack scan ran %s and did not find any risks. %s</string>
     <string name="scan_idle_with_threats_description_plural">The scan found %1s potential threats with %2s. Please review them below and take action or tap the fix all button. We are %3$s if you need us.</string>
-    <string name="scan_idle_with_threats_description_singular">The scan found %1s potential threat with %2s. Please review the threat below and take action or tap the fix all button. We are %3$s if you need us.</string>
+    <string name="scan_idle_with_threats_description_singular">The scan found 1 potential threat with %1s. Please review the threat below and take action or tap the fix all button. We are %2$s if you need us.</string>
```
AliSoftware: Wording nitpick: I know that the previous code was written in a way that the output was already using `1` (and not `one`) in singular form (since it used via `UiStringText("${fixableThreatIds.size}")` in both code paths), but if I'm not mistaken, at least in English it's better to spell-out numbers when using in that kind of context ([source](https://www.grammarly.com/blog/when-to-spell-out-numbers/)) 🙂 

```suggestion
    <string name="scan_idle_with_threats_description_singular">The scan found one potential threat with %1s. Please review the threat below and take action or tap the fix all button. We are %2$s if you need us.</string>
```
(ditto for the other string below of course)

JavonDavis: I thought about it but saw that using "1" was a common pattern, I'll leave the decision on this up to @ashiagr though 🙂

JavonDavis:
Also flagging that changing this to remove the positional argument, will affect all the other translated languages(hence the current lint failure), excluding Romanian... @ashiagr if we're going to do this it'd be probably be best we rename this key to and then merge the release branch into develop to get new a new set of translations and avoid any accidental breakages... wdyt?

WordPress/src/main/res/values/strings.xml
ashiagr: Actually yes, 1 is a common pattern in the code but I think it is fine to use "one" as well if it eases our translators job. I'm fine with replacing "1" with "one".

ashiagr:
Thanks for this PR @JavonDavis  🙏.

> Also flagging that changing this to remove the positional argument, will affect all the other translated languages(hence the current lint failure), excluding Romanian... @ashiagr if we're going to do this it'd be probably be best we rename this key to and then merge the release branch into develop to get new a new set of translations and avoid any accidental breakages... wdyt?

Yes, I agree it is better to rename the key and merge the release branch to develop. Should I go ahead and make this change in this PR?

JavonDavis:
> Yes, I agree it is better to rename the key and merge the release branch to develop. Should I go ahead and make this change in this PR?

Yep that'd be great! 🙂 Thanks Ashita! I'll change this PR from a draft since we're going ahead with the change

ashiagr:
This is done @JavonDavis , included screenshots below. Can I merge it to `release/16.8`?

|scan_idle_threats_description_singular | threat_fix_all_confirmation_message_singular | scan_finished_potential_threats_found_message_singular |
|---|---|---|
|device-2021-03-04-102637 | device-2021-03-04-102702 | device-2021-03-04-103437|

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/builders/ScanStateListItemsBuilder.kt
```diff
@@ -257,15 +257,24 @@ class ScanStateListItemsBuilder @Inject constructor(
...
+                            clickableText
+                    )
+        } else {
+            htmlMessageUtils
+                    .getHtmlMessageFromStringFormatResId(
+                            descriptionTextResId,
+                            "<b>${site.name ?: resourceProvider.getString(R.string.scan_this_site)}</b>",
+                            clickableText
+                    )
+        }
```
AliSoftware: nit: now that we have different code paths for calling `getHtmlMesageFromStringFormatResId` anyway, probably not worth having an `if` branch just for the `descriptionTestResId` separately.

i.e. remove the `val descriptionTextResId = …` lines and directly use:
```kt
        val descriptionText = if (threatsCount > 1) {
            htmlMessageUtils
                    .getHtmlMessageFromStringFormatResId(
                            R.string.scan_idle_threats_description_plural,
                            "<b>$threatsCount</b>",
                            "<b>${site.name ?: resourceProvider.getString(R.string.scan_this_site)}</b>",
                            clickableText
                    )
        } else {
            htmlMessageUtils
                    .getHtmlMessageFromStringFormatResId(
                            R.string.scan_idle_threats_description_singular,
                            "<b>${site.name ?: resourceProvider.getString(R.string.scan_this_site)}</b>",
                            clickableText
                    )
        }
```

AliSoftware (commented):
Code is looking good; only left a suggestion for code readability.

Not marking the PR as ready to merge yet though because @JavonDavis will need to re-run the scripts which updates the `./fastlane/resources/values/strings.xml` file (the one imported by GlotPress) before merging, in order for the new strings to be picked up and to have a chance to have then translated in time before release day (otherwise, those new strings won't be uploaded to GP… until we merge this PR to develop and then start the next code-freeze from develop for the next version next Monday)

@JavonDavis feel free to decide to merge this PR as is anyway… as is as long as you ensure that don't forget to update the file separately during the wrangling of the beta

WordPress/src/main/java/org/wordpress/android/ui/jetpack/scan/builders/ScanStateListItemsBuilder.kt
ashiagr: Nice suggestion @AliSoftware 🌟, applied in 23da7f2.

JavonDavis:
Thanks @AliSoftware  @ashiagr ! Can I get a 👍  on this? I'll go ahead and update strings.xml when wrangling the new beta and merging into the develop

AliSoftware (approved):

PR #14186, opened 2021-03-04
Use viewModelScope in the ScopedViewModel.kt class

planarvoid:
In order to start using viewModelScope instead of the ScopedViewModel I've implemented a simple solution where I've removed the `CoroutineScope` interface from the ScopedViewModel and I've implemented the `launch` with the `viewModelScope` instead. This should help to correctly handle termination of the sibling jobs while at the same time it requires almost no changes to the tests which are still passing. It also provides us with the handy `launch` method in the view model without the need to call `viewModelScope.launch`. Let me know what you think @renanferrari 

To test:
- I'd trust the unit tests with this change

PR submission checklist:


renanferrari (approved):
LGTM 👍

PR #14191, opened 2021-03-04
Add content logging to help debug a Story post crash

renanferrari:
This PR is a follow up to #14185 and helps to investigate #13697.

It basically adds a `SiteModel` parameter to the `findAllStoryBlocksInPostAndPerformOnEachMediaFilesJson` method which we then use to check if we should also log the Post's content in addition to reporting any caught `StringIndexOutOfBoundsException`. The logic used to determine that is based [on this one](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L1868-L1877), which makes sure we only log posts that are not private or password-protected, on WPCom sites that are also not private.

To test:

Again, there's not much to test here since we can't reproduce the issue, but you can smoke test the Story feature and make sure nothing looks off.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/media/services/MediaUploadReadyListener.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/media/services/MediaUploadReadyListener.java
@@ -3,6 +3,7 @@ package org.wordpress.android.ui.media.services;
 import androidx.annotation.Nullable;
 
 import org.wordpress.android.fluxc.model.PostModel;
+import org.wordpress.android.fluxc.model.SiteModel;
 import org.wordpress.android.util.helpers.MediaFile;
 
 /**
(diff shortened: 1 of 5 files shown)


mzorz (approved):
Thank you for these changes @renanferrari 💯 

I was able to test the try/catch path by adding an artificial `throw StringIndexOutOfBoundsException("TEST")`  line within the `try{}`  block, and confirmed the logcat showed what I expected:

```
2021-03-04 18:17:21.062 15273-15273/org.wordpress.android E/WordPress-EDITOR: Error while parsing Story blocks: TEST
2021-03-04 18:17:24.027 15273-15333/org.wordpress.android D/Volley: [534] BasicNetwork.logSlowRequests: HTTP response for request=<[ ] https://public-api.wordpress.com/wpcom/v2/sites/176139122/rewind/capabilities/?_locale=en_US 0x7b2dc254 NORMAL 39> [lifetime=20476], [size=19], [rc=200], [retryCount=0]
2021-03-04 18:17:29.517 15273-15273/org.wordpress.android E/WordPress-EDITOR: HTML content of the post before the crash: <!-- wp:jetpack/story {"mediaFiles":[{"alt":"asdfafdafa","caption":"","id":"1","link":"","mime":"image/jpeg","type":"image","url":""}]} -->
    <div class="wp-story wp-block-jetpack-story"></div>
    <!-- /wp:jetpack/story -->
```

Also confirmed the test case runs correctly
Finally, I gave a pass on the code which looks right and to the point 👍 

LGTM :shipit:

PR #14197, opened 2021-03-05
Dynamic cards tracking

planarvoid:
The original purpose of this PR was to add the feature config property to all the necessary quick start events. I've added it to the "complete task" event, "all tasks completed" event and I've created a "quick start started" event so we can easily build the funnels. 

While testing the flows I've discovered 2 bugs:
- Go through the original QS flow (without the MySiteImprovements feature flag)
- Finish all the steps
- The QS block never disappears
I'm pretty sure the reason for this is that we never call the `setQuickStartCompleted`. It's only called when we skip all the tasks when disabling the dialog. However, we use the `getQuickStartCompleted` when we're checking whether the QS is in progress. Please tell me if I missed something here! I've added the `setQuickStartCompleted` to both places where we complete the quick start (both old way and the new way) and now it works.

Second bug:
- Finish the flow (both with and without the MySiteImprovements flag)
- Notice the `QUICK_START_ALL_TASKS_COMPLETED` event is never sent
The reason for this was that the method `isEveryQuickStartTaskDone` used to see whether there is anything else the user can do was using `getDoneCount` and comparing it to `QuickStartTask.values().size`. The problem here is that we have the `Unknown` task which could never be finished unless the user has skipped the quick start. I think the event was never sent. 

To test 1:
- Make sure the MySiteImprovements flag is turned ON and the app is restarted
- Create a site
- In Logcat: Notice the `quick_start_started` event with the `my_site_improvements` property == true
- Complete a task
- In Logcat: Notice the `quick_start_task_completed` event with the `my_site_improvements` property == true
- Finish all the tasks
- In Logcat: Notice the `quick_start_all_tasks_completed` event with the `my_site_improvements` property == true

To test 2:
- Make sure the MySiteImprovements flag is turned OFF and the app is restarted
- Create a site
- In Logcat: Notice the `quick_start_started` event with the `my_site_improvements` property == false
- Complete a task
- In Logcat: Notice the `quick_start_task_completed` event with the `my_site_improvements` property == false
- Finish all the tasks
- In Logcat: Notice the `quick_start_all_tasks_completed` event with the `my_site_improvements` property == false
- Notice the QuickStart block is gone

PR submission checklist:


renanferrari (approved):
Hey, thanks for the changes and for taking care of those bugs. The code is looking pretty good and the app is working as expected both with and without the feature flag.

Feel free to merge this once those lint errors are fixed.

LGTM :shipit:

PR #14198, opened 2021-03-05
Isse/7900 timezone selection - Rewrite Timezone selection using Kotlin with BottomSheetDialogFragment

### LINKED ISSUE #7900: Timezone selection can be confusing if some major cities are not included in the timezone dropdown list
- If a customer lives in a place that's not listed here, it's hard to skim through a long list of cities to find something that could be near to them. Even if they find something, how would they know if it's on the same timezone as they are? To make it easier, there should be continents listed with city names (i.e. "Europe/Berlin", "Asia/Kuala Lumpur" etc)

- If a customer does not want to use timezone strings but UTC/GMT offsets instead, they are not listed. Timezone strings are special in that they follow summertime changes. Sometimes these work wonky or customers might have some other reasons to stick with "fixed" offsets. Adding a list of "UTC+/-" values would be helpful. 

- Not sure what happens if a customer has already chosen UTC-offset in WordPress wp-admin and then looks at this list?

Screenshots from Calypso:

image

image

ravishanker:
Timezone selection from site settings in Android app is outdated in comparison with WordPress Core and iOS.  This PR brings android on par with them by adding missing features like grouping by continents and manual offsets.  Please see #7900 for more details on this issue.

It also improves the user experience by introducing bottom sheet,  better search and timezone information similar to android system timezone selection feature.  As below.

It is also a rewrite using modern android practices like Kotlin and MVVM pattern from current everything in one Java file implementation.

Fixes #7900 

To do:
- <strike>Refine timezone offset information</strike>
- <strike>Refine search feature</strike>
- <strike>Suggested timezone</strike>

To test:
Navigate to My Site -> Settings -> Timezone

<table><tr><td></td>
<td></td>
<td></td></tr></table>

PR submission checklist:


khaykov:
Hey @ravishanker 👋 Thanks for the PR! I checked the functionality of timezone picker looks like it works pretty well. 

Pulling timezones from the `TimeZone` class is pretty great, since we don't have to deal with any networking, but I have one concern about this approach - `TimeZone` class has a lot of timezones, and a there is a good number of zones that are not relevant (if you scroll through the list there is a lot of weird stuff). What do you think about pulling the timezones from our endpoint and using `TimeZone` to give them more information (what you are currently using in the list)?

ravishanker:
> Hey @ravishanker 👋 Thanks for the PR! I checked the functionality of timezone picker looks like it works pretty well.
> 
> Pulling timezones from the `TimeZone` class is pretty great, since we don't have to deal with any networking, but I have one concern about this approach - `TimeZone` class has a lot of timezones, and a there is a good number of zones that are not relevant (if you scroll through the list there is a lot of weird stuff). What do you think about pulling the timezones from our endpoint and using `TimeZone` to give them more information (what you are currently using in the list)?

Hi @khaykov Thanks for the feedback 🙏.  I'm looking at filtering entries to remove all those extras  For e.g. limiting to Los Angeles -> Pacific Time (GMT-8:00) or Sydney -> Australian Eastern Time (GMT+11).  Perhaps, making an API call could be an option if we can't figure a way to filter these entries.

ravishanker:
@khakov 🙋‍♂️ Finally, managed to filter the data, update layout without the need to fallback on API.  Please take a look again. Thanks

khaykov:
@ravishanker Thanks for the changes, weird timezones are gone 👍 Thinking about functionality and use cases, what do you think about including timezone name (the second row in the list elements) into the search? I see cases where folks would know their timezone name, but not necessarily what major city it might be associated with. Eg. living in the middle of US, you might search for "Central Time" instead of a "Chicago" (or some other major city). In addition, we still probably would want to include manual (and searchable) UTC offsets for convenience.

ravishanker:
@khaykov 👋 Thanks for the review 🙏.  That's a good idea to include second row into search field.  I'll add back manual UTC offsets.  
- Should continents be included as well into searchable?
- We're showing GMT prefix (GMT+6:00) instead of UTC, should that be GMT or UTC?

ravishanker:
@khaykov I've added the following now

- timezone name in search results
- suggested time zone under the search box (see updated screenshot above)
However, the manual offsets are not working.  Looks like the api used in android is not handling manual offsets.  It is not accepting and updating the manual offsets.  I've even tried adding the list from api, but to no avail unless I'm missing something here!

ravishanker:
@khaykov  :wave: What I found on investigating the issue is
when you select a manual offset for e.g UTC+7, the back-end is updating it is as "gmt_offset: 7" and
when site settings is fetched it is actually returning empty ""  

see WpComSiteSettings.java, pushWpSettings method line 382 and fetchWpSettings method on line 168)
It is working alright when selecting anything other than manual offsets though.
I've also now reverted from use of TimeZone class to calling AP and augmenting it with timezone offset and time.

WordPress/src/main/java/org/wordpress/android/ui/prefs/timezone/SiteSettingsTimezoneBottomSheet.kt
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    companion object {
+        const val KEY_TIMEZONE = "timezone"
+
+        @JvmStatic
+        fun newInstance(timezone: String): SiteSettingsTimezoneBottomSheet =
+                SiteSettingsTimezoneBottomSheet().apply {
+                    arguments = Bundle().apply {
+                        putString(KEY_TIMEZONE, timezone)
```
khaykov: Looks like `KEY_TIMEZONE` migrated from previous implementation, but I don't think we are using it anywhere.

WordPress/src/main/java/org/wordpress/android/ui/prefs/timezone/SiteSettingsTimezoneBottomSheet.kt
```diff
@@ -0,0 +1,206 @@
...
+
+class SiteSettingsTimezoneBottomSheet : BottomSheetDialogFragment() {
+    @Inject lateinit var viewModelFactory: ViewModelProvider.Factory
+    private lateinit var timezoneViewModel: SiteSettingsTimezoneViewModel
+
+    private val timezoneAdapter = TimezoneAdapter { timezone ->
+        setSelectedTimezone(timezone)
+    }
+
+    private var searchJob: Job? = null
```
khaykov: This is a pretty minor thing, but what do you think about moving this `Job` into ViewModel? So instead of calling `search()` or `updateTimezonesFromInput()` when text changes, we would just let the VM know that the text in search filed have changed, and it can do its magic.

WordPress/src/main/java/org/wordpress/android/ui/prefs/timezone/SiteSettingsTimezoneBottomSheet.kt
```diff
@@ -0,0 +1,206 @@
...
+
+        timezoneViewModel.timezones.observe(viewLifecycleOwner, {
+            timezoneAdapter.submitList(it)
+        })
+
+        timezoneViewModel.suggestedTimezone.observe(viewLifecycleOwner, {
+            binding?.btnTimezoneSuggestion?.text = it
+        })
+
+        timezoneViewModel.timezoneSearch.observe(viewLifecycleOwner, {
```
khaykov: From the perspective of the list, `timezones` and `timezoneSearch` are the same, do you think we can only observe `timezones` and update VM logic to modify `timezones` on search?

WordPress/src/main/java/org/wordpress/android/ui/prefs/timezone/SiteSettingsTimezoneBottomSheet.kt
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    private fun setSelectedTimezone(timezone: TimezoneItem?) {
+        timezone?.let {
+            callback?.onSelectTimezone(it.value)
+            dismiss()
+        }
+    }
+
+    private fun showEmptyView(show: Boolean) {
```
khaykov: Pretty minor comment, but to make things a bit more clear, what do think about renaming `showEmptyView` to something like `toggleEmptyViewVisibility`. To me personally, `showEmptyView` implies exactly that - to show empty view, so having `show` attribute is a bit weird. Same goes for `showProgressView` method.

khaykov: Pretty minor comment, but to make things a bit more clear, what do think about renaming `showEmptyView` to something like `toggleEmptyViewVisibility`. To me personally, `showEmptyView` implies exactly that - to show empty view, so having `show` attribute is a bit weird. Same goes for `showProgressView` method.

WordPress/src/main/java/org/wordpress/android/ui/prefs/timezone/SiteSettingsTimezoneBottomSheet.kt
```diff
@@ -0,0 +1,206 @@
...
+
+            searchInputLayout.setEndIconOnClickListener {
+                searchInputLayout.editText?.text?.clear()
+                searchInputLayout.editText?.clearFocus()
+                timezoneViewModel.onSearchCancelled()
+                hideSearchKeyboard()
+            }
+
+            btnTimezoneSuggestion.setOnClickListener { it as MaterialButton
+                callback?.onSelectTimezone(it.text.toString())
```
khaykov: I see we have a similar code in `setSelectedTimezone`, I wonder if we can refactor it a bit by wiring it through ViewModel. Maybe observe something like `selectedTimezone` from VM, and notify it that timezone was selected (either from suggestion or adapter).

We can go even further, and separate call to callback and `dismiss()` (by observing something like `dismissBottomSheet`) and reacting to it. 

This might be going a bit too granular, but we can benefit from separating this logic when testing VM. We would be able to confirm that Timezone WAS selected, and dialog WAS dismissed. With current implementation we can't test what would happen after you click on the timezone.

WordPress/src/main/java/org/wordpress/android/ui/prefs/timezone/SiteSettingsTimezoneViewModel.kt
```diff
@@ -0,0 +1,211 @@
...
+import java.time.format.DateTimeFormatter
+import java.time.format.TextStyle.FULL
+import java.time.zone.ZoneRulesException
+import java.util.Locale
+import javax.inject.Inject
+
+class SiteSettingsTimezoneViewModel @Inject constructor() : ViewModel() {
+    private val timezonesList = mutableListOf<TimezonesList>()
+
+    private val _showEmpty = MutableLiveData<Boolean>()
```
khaykov: Very minor thing - historically, we call mutable and regular live data the same way + underscore (eg. `_showEmpty` will be `_ showEmptyView` :)

WordPress/src/main/java/org/wordpress/android/ui/prefs/timezone/SiteSettingsTimezoneViewModel.kt
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        return filteredTimezones
+    }
+
+    fun onSearchCancelled() {
+        _timezones.postValue(timezonesList)
+    }
+
+    fun getTimezones(context: Activity) {
```
khaykov: Dow we actually need an `Activity` here? Or the `context` will be enough?

WordPress/src/main/java/org/wordpress/android/ui/prefs/timezone/SiteSettingsTimezoneViewModel.kt
```diff
@@ -0,0 +1,211 @@
...
+
+    private val _suggestedTimezone = MutableLiveData<String>()
+    val suggestedTimezone = _suggestedTimezone
+
+    private val _timezones = MutableLiveData<List<TimezonesList>>()
+    val timezones = _timezones
+
+
+    fun searchTimezones(query: CharSequence?) {
+        searchInput.value = query.toString()
```
khaykov, line 74: `query` is marked as nullable, so if at any point it will actually be a null, we will filter with a "null" string, which might not be what want :)

WordPress/src/main/java/org/wordpress/android/ui/prefs/timezone/SiteSettingsTimezoneViewModel.kt
```diff
@@ -0,0 +1,211 @@
...
+    private fun getZoneId(zone: String): ZoneId? {
+        return try {
+            ZoneId.of(zone)
+        } catch (e: ZoneRulesException) {
+            AppLog.e(SETTINGS, "Error parsing zoneId", e)
+            null
+        }
+    }
+
+    enum class Continents(val s: String) {
```
khaykov: For readability, we try to avoid ambitious attributes names like `s`. Also, it looks like the continent names from here will be visible in UI, so they need to be in strings resources.

WordPress/src/main/java/org/wordpress/android/ui/prefs/timezone/SiteSettingsTimezoneViewModel.kt
```diff
@@ -0,0 +1,211 @@
...
+            val jsonResponse = JSONObject(responseJson.orEmpty())
+            val jsonTimezonesByContinent = jsonResponse.getJSONObject("timezones_by_continent")
+            val jsonTimezonesManualOffsets = jsonResponse.getJSONArray("manual_utc_offsets")
+
+            Continents.values().map {
+                timezonesList.add(TimezoneHeader(it.s))
+                addTimezoneItems(jsonTimezonesByContinent.getJSONArray(it.s))
+            }
+
+            timezonesList.add(TimezoneHeader("Manual Offsets"))
```
khaykov: Will this string be visible? If so, we would probably like to add it to `strings.xml`

WordPress/src/main/res/drawable/bg_rectangle_grey_stroke_radius_4dp.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
```
khaykov: Not really sure where this file is used?

WordPress/src/main/res/layout/site_settings_timezone_bottom_sheet_list_header.xml
```diff
@@ -0,0 +1,13 @@
...
+    android:id="@+id/header_time_zone"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:background="?attr/selectableItemBackground"
+    android:paddingBottom="@dimen/margin_medium"
+    android:paddingTop="@dimen/margin_extra_medium_large"
+    android:textAppearance="?attr/textAppearanceHeadline6"
+    tools:text="Africa" />
+
+
```
khaykov: A bit of extra lines here :)

WordPress/src/main/res/layout/site_settings_timezone_bottom_sheet_list_item.xml
```diff
@@ -0,0 +1,49 @@
...
+        app:layout_constraintTop_toTopOf="parent"
+        tools:text="Abidjan" />
+
+    <TextView
+        android:id="@+id/zone_offset"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:textAppearance="?attr/textAppearanceBody2"
+        android:maxLines="1"
+        android:textColor="@color/zs_request_list_light_text_color"
```
khaykov: We probably would want to use something like `?attr/wpColorOnSurfaceMedium` here, so it will be readable in dark mode. I think `zs_request_list_light_text_color` comes from one of the 3rd party libraries.

khaykov (commented):
Thank you for improving the TimeZone picker, @ravishanker ! I finished the first pass, checked the code and functionality, and I didn't found any major issues. I found couple of minor issues and have some suggestions. Let me know what you think!

Visual issues:

Bottom of the list is not accesible when keyboard is visible:
[Image from Gyazo](https://gyazo.com/9f0257aa268a276d54d02bfee8639d67)

The ripple animation is affected by side padding, and does not extend to full row:
[Image from Gyazo](https://gyazo.com/0e80d30357fd05b7f44137c5b1615c04)

When the list expanded there is a weird gap under it:
[Image from Gyazo](https://gyazo.com/12996b664417b29284ac52c5cab23b70)

In landscape mode the bottom sheep is collapsed. This might be a default behavior, but we probably want it to be expanded.
[Image from Gyazo](https://gyazo.com/31910dd82285aac029d1d500572b753c)

Suggested timezone looks a bit weird in the center, what do you think about moving it next to the label or sticking it to the end?
[Image from Gyazo](https://gyazo.com/f2f73367c77eecd790472fbd7a0f7384)

The scrollbar is not visible.

JavonDavis:
Hey @ravishanker! Bumping this to `17.1`, if you'd like to get it in for this release please merge this into release17.0 and then DM me to release a new Beta version.

ravishanker:
@khaykov  I've addressed all the PR review comments from above.  Please take a look again.  Thanks.

WordPress/src/main/java/org/wordpress/android/ui/prefs/timezone/SiteSettingsTimezoneViewModel.kt
```diff
@@ -28,34 +33,55 @@ import java.time.zone.ZoneRulesException
...
 
-    fun searchTimezones(query: CharSequence?) {
+
+    fun searchTimezones(query: CharSequence) {
         searchInput.value = query.toString()
     }
 
+    fun onTimezoneSelected(timezone: String) {
+        _selectedTimezone.value = timezone
+        _dismissBottomSheet.value = Unit
```
khaykov: TIP: There is a convenient `call()` or `asyncCall()` methods for situations like this :)

khaykov:
Thank you for the changes, @ravishanker! Looks better now 👍 
I still see couple of visual issues:
- Keyboard blocks the list (scroll down to the bottom of the list and open the keyboard).
- Dialog is opened in collapsed state in landscape orientation.

Also there is a certain issue with the way you set up a `TimezoneSelectionCallback` in both current and previous implementation. Could you please explain the problem with the current approach and address it if you can?

WordPress/src/main/java/org/wordpress/android/ui/prefs/timezone/SiteSettingsTimezoneViewModel.kt
ravishanker: Thanks for the tip, will use it

ravishanker:
> Thank you for the changes, @ravishanker! Looks better now 👍
> I still see couple of visual issues:
> 
>     * Keyboard blocks the list (scroll down to the bottom of the list and open the keyboard).
> 
>     * Dialog is opened in collapsed state in landscape orientation.
> 
> 
> Also there is a certain issue with the way you set up a `TimezoneSelectionCallback` in both current and previous implementation. Could you please explain the problem with the current approach and address it if you can?

- I've implemented keyboard to dismiss when scroll starts.  Do I need to implement softInputMode adjustResize or adjustPan
- I put in a fix for opening bottomsheet expanded in landscape mode.  Seems to work on my device, will check again.
- I'm not sure about the certain issue.  Are you referring to old implementation in Java using targetFragment, onActivityResult and this implementation or the previous version and current version of the present implementation?

WordPress/src/test/java/org/wordpress/android/ui/prefs/timezone/SiteSettingsTimezoneViewModelTest.kt
```diff
@@ -0,0 +1,56 @@
...
+    private lateinit var viewModel: SiteSettingsTimezoneViewModel
+
+    @Before
+    fun setUp() {
+        viewModel = SiteSettingsTimezoneViewModel(resourceProvider)
+    }
+
+    @Test
+    fun testSearchTimezones() {
+        viewModel.timezones.observeForever {}
```
khaykov: Just for the sanity check and to test the behavior of observer exactly, would you mind assigning result to an object, and asserting it, instead of checking `.value` ?

khaykov (commented):
Thanks for all great work, Ravi! I checked what's new in the dependency versions you bumped, and looks like there is nothing breaking, and just general improvements, so we are good to go 🚀 

I left one minor comment about the test, and as a last step would you mind pulling the latest changes from `develop` and add this feature to `RELEASE_NOTES.txt`?

WordPress/src/test/java/org/wordpress/android/ui/prefs/timezone/SiteSettingsTimezoneViewModelTest.kt
ravishanker: Updated to assign result to an object and assert it

ravishanker:
@khaykov 👋  Updated the RELEASE_NOTES.txt with 17.1 entry

khaykov:
@ravishanker Looks like the stories submodule changes snuck in with your last commit. Would you mind reverting it? Thank you!

khaykov:
@ravishanker I ran a CI checks, and looks like there are some lint errors. You can check what is wrong by running `./gradlew ktlint`

khaykov (approved):
Thanks for all the changes, Ravi! Looks good to me 🚀

PR #14200, opened 2021-03-05
Feature/manage translations

loremattei:
This PR adds support for translating `strings.xml` files which belong to internal libraries which are imported as binary dependencies. 
This PR only covers `mobile-gutenberg`, but the flow can be easily used for other libraries by adding them to the `binary_imported_libraries` array. 

`localize_binary_deps` downloads the `strings.xml` file from the GitHub release which is referred in `build.gradle` as the imported one. 

**To test:**
1. Checkout a new branch from this one. 
2. Update `ext.gutenbergMobileVersion` in `build.gradle` to a released version.
3. `bundle install`
4. Run `bundle exec fastlane localize_binary_deps` and verify that the main `strings.xml` gets updated. 

Related `release-toolkit` PR: https://github.com/wordpress-mobile/release-toolkit/pull/222

**Note:** Before merging this PR, I'll need to release a new version of `release-toolkit` and update the reference in this PR.

PR submission checklist:

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -507,7 +507,10 @@ ENV["validate_translations"]="lintVanillaRelease"
         github_release_prefix: lib[:github_release_prefix])
 
       if download_path.nil?
-        UI.error "Abort." unless UI.confirm("Can't download strings file for #{lib[:name]}. Do you want to continue?")
+        error_message = "Can't download strings file for #{lib[:name]}.\r\n"
+        error_message += "Strings for this library won't get translated.\r\n"
+        error_message += "Do you want to continue anyway?"
+        UI.user_error! "Abort." unless UI.confirm(error_message)
       else
         UI.message("Strings.xml file downloaded to #{download_path}.")
         sh(merge_strings_tool_file_path, lib[:name], download_path)
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -512,7 +512,7 @@ ENV["validate_translations"]="lintVanillaRelease"
         error_message += "Do you want to continue anyway?"
         UI.user_error! "Abort." unless UI.confirm(error_message)
       else
-        UI.message("Strings.xml file downloaded to #{download_path}.")
+        UI.message("Strings.xml file for #{lib[:name]} downloaded to #{download_path}.")
         sh(merge_strings_tool_file_path, lib[:name], download_path)
         File.delete(download_path) if File.exist?(download_path)
       end
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -42,7 +42,7 @@ import "./ScreenshotFastfile"
 ########################################################################
 Dotenv.load('~/.wpandroid-env.default')
 ENV[GHHELPER_REPO="wordpress-mobile/WordPress-Android"]
-ENV["PROJECT_ROOT_FOLDER"]="./"
+ENV["PROJECT_ROOT_FOLDER"]=File.dirname(File.expand_path(__dir__)) + "/"
 ENV["PROJECT_NAME"]="WordPress"
 ENV["HAS_ALPHA_VERSION"]="true"
 ENV["validate_translations"]="lintVanillaRelease"
@@ -70,15 +70,15 @@ ENV["validate_translations"]="lintVanillaRelease"
 
     android_bump_version_release()
     new_version = android_get_app_version()
-    extract_release_notes_for_version(version: new_version, 
-      release_notes_file_path:"#{ENV["PROJECT_ROOT_FOLDER"]}RELEASE-NOTES.txt", 
+    extract_release_notes_for_version(version: new_version,
+      release_notes_file_path:"#{ENV["PROJECT_ROOT_FOLDER"]}RELEASE-NOTES.txt",
       extracted_notes_file_path:release_notes_path)
     cleanup_release_files(files: [release_note_short_path])
     android_update_release_notes(new_version: new_version)
     setbranchprotection(repository:GHHELPER_REPO, branch: "release/#{new_version}")
     setfrozentag(repository:GHHELPER_REPO, milestone: new_version)
 
-    localize_gutenberg(skip_module_update: true)
+    localize_binary_deps()
     localize_libs()
     send_strings_for_translation()
     ensure_git_status_clean()
@@ -193,7 +193,7 @@ ENV["validate_translations"]="lintVanillaRelease"
   #####################################################################################
   # finalize_hotfix_release
   # -----------------------------------------------------------------------------------
-  # This lane finalizes the hotfix branch. 
+  # This lane finalizes the hotfix branch.
   # -----------------------------------------------------------------------------------
   # Usage:
   # bundle exec fastlane finalize_hotfix_release
@@ -247,7 +247,7 @@ ENV["validate_translations"]="lintVanillaRelease"
   # bundle exec fastlane build_and_upload_release
   # bundle exec fastlane build_and_upload_release skip_confirm:true
   # bundle exec fastlane build_and_upload_release skip_prechecks:true
-  # bundle exec fastlane build_and_upload_release create_release:true 
+  # bundle exec fastlane build_and_upload_release create_release:true
   #####################################################################################
   desc "Builds and updates for distribution"
   lane :build_and_upload_release do | options |
@@ -279,7 +279,7 @@ ENV["validate_translations"]="lintVanillaRelease"
   # Example:
   # bundle exec fastlane build_and_upload_pre_releases
   # bundle exec fastlane build_and_upload_pre_releases skip_confirm:true
-  # bundle exec fastlane build_and_upload_beta create_release:true 
+  # bundle exec fastlane build_and_upload_beta create_release:true
   #####################################################################################
   desc "Builds and updates for distribution"
   lane :build_and_upload_pre_releases do | options |
@@ -460,6 +460,10 @@ ENV["validate_translations"]="lintVanillaRelease"
     {library: "Stories Creator", strings_path: "./libs/stories-android/stories/src/main/res/values/strings.xml", exclusions: []}
   ]
 
+  binary_imported_libraries = [
+    { name: "Gutenberg Native",   import_key: "gutenbergMobileVersion",   repository: "wordpress-mobile/gutenberg-mobile",   strings_file_path: "bundle/android/strings.xml" },
+  ]
+
   private_lane :send_strings_for_translation do | options |
     sh("cd .. && mkdir -p #{update_strings_path} && cp #{main_strings_path} #{update_strings_path} && git add #{update_strings_path}strings.xml")
     sh("git diff-index --quiet HEAD || git commit -m \"Send strings to translation.\"")
@@ -485,14 +489,24 @@ ENV["validate_translations"]="lintVanillaRelease"
     end
   end
 
-  desc "Import Gutenberg strings"
-  lane :localize_gutenberg do | options |
-    unless options[:skip_module_update] then
-      sh("git submodule init")
-      sh("git submodule update")
-    end
+  desc "Import strings from binary dependencies"
+  lane :localize_binary_deps do | options |
+    merge_strings_tool_file_path = File.join(ENV["PROJECT_ROOT_FOLDER"], 'tools/merge_strings_xml.py')
+    binary_imported_libraries.each do  | lib |
+      download_path = android_download_file_by_version(
+        library_name: lib[:name],
+        import_key: lib[:import_key],
+        repository: lib[:repository],
+        file_path: lib[:strings_file_path])
 
-    sh("cd .. && python ./tools/merge_strings_xml.py")
+      if download_path.nil?
+        UI.error "Abort." unless UI.confirm("Can't download strings file for #{lib[:name]}. Do you want to continue?")
+      else
+        UI.message("Strings.xml file downloaded to #{download_path}.")
+        sh(merge_strings_tool_file_path, lib[:name], download_path)
+        File.delete(download_path) if File.exist?(download_path)
+      end
+    end
 
     is_repo_clean = ("git status --porcelain").empty?
     unless is_repo_clean then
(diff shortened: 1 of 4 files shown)


loremattei:
The failure doesn't seem to be related to these changes. It looks like the HEAD commit in the target branch has also failed, so I guess it may be related to it.

oguzkocer:
> The failure doesn't seem to be related to these changes. It looks like the HEAD commit in the target branch has also failed, so I guess it may be related to it.

Yeap, I believe we'll just need to remove that CI task. I'll fix it in the original PR and we can ignore that task for the merge of this one.

fastlane/Fastfile
```diff
@@ -42,7 +42,7 @@ import "./ScreenshotFastfile"
 ########################################################################
 Dotenv.load('~/.wpandroid-env.default')
 ENV[GHHELPER_REPO="wordpress-mobile/WordPress-Android"]
-ENV["PROJECT_ROOT_FOLDER"]="./"
+ENV["PROJECT_ROOT_FOLDER"]=File.dirname(File.expand_path(__dir__)) + "/"
```
mokagio, line 45: I was surprised by `__dir__`, not being uppercased like `__FILE__`.

I looked it up and it turns out here's [an interesting reason for it in the Ruby semantics](https://stackoverflow.com/questions/15190700/why-is-file-uppercase-and-dir-lowercase).

fastlane/Fastfile
```diff
@@ -460,6 +460,10 @@ ENV["validate_translations"]="lintVanillaRelease"
     {library: "Stories Creator", strings_path: "./libs/stories-android/stories/src/main/res/values/strings.xml", exclusions: []}
   ]
 
+  binary_imported_libraries = [
+    { name: "Gutenberg Native",   import_key: "gutenbergMobileVersion",   repository: "wordpress-mobile/gutenberg-mobile",   strings_file_path: "bundle/android/strings.xml", github_release_prefix: "v" },
```
mokagio: Nitpick. How come three spaces between pairs in the hash?

```suggestion
    { name: "Gutenberg Native", import_key: "gutenbergMobileVersion", repository: "wordpress-mobile/gutenberg-mobile", strings_file_path: "bundle/android/strings.xml", github_release_prefix: "v" },
```

Or, to avoid having to scroll horizontally:

```suggestion
    {
      name: "Gutenberg Native",
      import_key: "gutenbergMobileVersion",
      repository: "wordpress-mobile/gutenberg-mobile",
      strings_file_path: "bundle/android/strings.xml",
      github_release_prefix: "v"
    },
```

fastlane/Fastfile
```diff
@@ -485,15 +489,26 @@ ENV["validate_translations"]="lintVanillaRelease"
...
   end
 
-  desc "Import Gutenberg strings"
-  lane :localize_gutenberg do | options |
-    unless options[:skip_module_update] then
-      sh("git submodule init")
-      sh("git submodule update")
+  desc "Import strings from binary dependencies"
+  lane :localize_binary_deps do | options |
+    merge_strings_tool_file_path = File.join(ENV["PROJECT_ROOT_FOLDER"], 'tools/merge_strings_xml.py')
```
mokagio: Nitpick. For consistency:

```suggestion
    merge_strings_tool_file_path = File.join(ENV["PROJECT_ROOT_FOLDER"], 'tools', 'merge_strings_xml.py')
```

fastlane/Fastfile
```diff
@@ -460,6 +460,10 @@ ENV["validate_translations"]="lintVanillaRelease"
     {library: "Stories Creator", strings_path: "./libs/stories-android/stories/src/main/res/values/strings.xml", exclusions: []}
   ]
 
+  binary_imported_libraries = [
```
mokagio: Nitpick. I recently [learned](https://github.com/Automattic/simplenote-macos/pull/870#discussion_r578371300) that Ruby constants should be uppercased.

```suggestion
  BINARY_IMPORTED_LIBRARIES = [
```

fastlane/Fastfile
```diff
@@ -485,15 +489,26 @@ ENV["validate_translations"]="lintVanillaRelease"
...
+        library_name: lib[:name],
+        import_key: lib[:import_key],
+        repository: lib[:repository],
+        file_path: lib[:strings_file_path],
+        github_release_prefix: lib[:github_release_prefix])
+
+      if download_path.nil?
+        UI.error "Abort." unless UI.confirm("Can't download strings file for #{lib[:name]}. Do you want to continue?")
+      else
+        UI.message("Strings.xml file downloaded to #{download_path}.")
```
mokagio: Nitpick. 
```suggestion
        UI.message("Strings.xml file for #{lib[:name]} downloaded to #{download_path}.")
```

tools/merge_strings_xml.py
```diff
@@ -98,16 +98,23 @@ def merge_strings(main_xml, extra_sections):
...
+    xml_string = ('<?xml version="1.0" encoding="UTF-8"?>' + '\n' +
         ET.tostring(main_root, encoding='utf8').split('\n',1)[1])
     with open(main_xml, 'wb') as xml_file:
         xml_file.write(xml_string)
 
 def main():
+    if len(sys.argv) != 3:
+        print 'Usage: merge_strings_xml.py <section name> <strings.xml file path>'
+        sys.exit(1)
+
```
mokagio, line 110: Nitpick. What do you think of assigning the argv elements to variables after this?

```python
section_name = sys.argv[1]
input_file = sys.argv[2]
```

This way we don't need to repeat argv accesses and readers won't need to scroll back to the usage text to know what `argv[x]` is.

fastlane/Fastfile
```diff
@@ -70,15 +70,15 @@ ENV["validate_translations"]="lintVanillaRelease"
 
     android_bump_version_release()
     new_version = android_get_app_version()
-    extract_release_notes_for_version(version: new_version, 
-      release_notes_file_path:"#{ENV["PROJECT_ROOT_FOLDER"]}RELEASE-NOTES.txt", 
+    extract_release_notes_for_version(version: new_version,
+      release_notes_file_path:"#{ENV["PROJECT_ROOT_FOLDER"]}RELEASE-NOTES.txt",
```
mokagio, line 74: Nitpick. I'm pretty sure this was a change from your editor, but since it happens to touch `PROJECT_ROOT_FOLDER`, maybe we can adopt Ruby `File.join` syntax here, too, like later on in the file?

```suggestion
      release_notes_file_path: File.join(ENV["PROJECT_ROOT_FOLDER"], 'RELEASE-NOTES.txt'),
```

fastlane/Fastfile
mokagio: I also wonder if we need to keep the trailing `/`.

If we were 100% sure all of the path management is done with Ruby APIs instead of string interpolation we wouldn't, but I'm not sure of that, so maybe it's best to err on the side of caution?

fastlane/Fastfile
```diff
@@ -485,15 +489,26 @@ ENV["validate_translations"]="lintVanillaRelease"
...
+    binary_imported_libraries.each do  | lib |
+      download_path = android_download_file_by_version(
+        library_name: lib[:name],
+        import_key: lib[:import_key],
+        repository: lib[:repository],
+        file_path: lib[:strings_file_path],
+        github_release_prefix: lib[:github_release_prefix])
+
+      if download_path.nil?
+        UI.error "Abort." unless UI.confirm("Can't download strings file for #{lib[:name]}. Do you want to continue?")
```
mokagio: Just using `error`, which doesn't exit the lane, is intentional?

I ended up in this path and the lane kept running even though it looked like it should stop. _There's another comment below about the request to commit the strings file event though it didn't change_.

image

If that's not the intended behavior, maybe [`user_error!` or `shell_error!`](https://docs.fastlane.tools/advanced/actions/#user-input-and-output) could do the job.

fastlane/Fastfile
```diff
@@ -485,15 +489,26 @@ ENV["validate_translations"]="lintVanillaRelease"
...
+        sh(merge_strings_tool_file_path, lib[:name], download_path)
+        File.delete(download_path) if File.exist?(download_path)
+      end
     end
 
-    sh("cd .. && python ./tools/merge_strings_xml.py")
-
     is_repo_clean = ("git status --porcelain").empty?
     unless is_repo_clean then
       commit_strings(options)
```
mokagio, line 523: This logic got me in "trouble".

I run it with an untracked file and, even though I had no string changes, I still got the message to check and commit the strings file.

That might happen in a real execution of the release process lanes, because many have a check to ensure the repo is clean, but I wanted to point it out.

mokagio (commented):
The changes here look good @loremattei 🙌 

I tested these changes using the `1.46.0` release and got this diff.

```diff
diff --git a/WordPress/src/main/res/values/strings.xml b/WordPress/src/main/res/values/strings.xml
index 13d75d319c..ecfeaf450c 100644
--- a/WordPress/src/main/res/values/strings.xml
+++ b/WordPress/src/main/res/values/strings.xml
@@ -3275,9 +3275,7 @@
     <string name="gutenberg_native_double_tap_to_move_the_block_to_the_left" tools:ignore="UnusedResources">Double tap to move the block to the left</string>
     <string name="gutenberg_native_double_tap_to_move_the_block_to_the_right" tools:ignore="UnusedResources">Double tap to move the block to the right</string>
     <string name="gutenberg_native_double_tap_to_move_the_block_up" tools:ignore="UnusedResources">Double tap to move the block up</string>
-    <string name="gutenberg_native_double_tap_to_open_action_sheet_to_edit_replace_or_clear_the_imag" tools:ignore="UnusedResources">Double tap to open Action Sheet to edit, replace, or clear the image</string>
     <string name="gutenberg_native_double_tap_to_open_action_sheet_with_available_options" tools:ignore="UnusedResources">Double tap to open Action Sheet with available options</string>
-    <string name="gutenberg_native_double_tap_to_open_bottom_sheet_to_edit_replace_or_clear_the_imag" tools:ignore="UnusedResources">Double tap to open Bottom Sheet to edit, replace, or clear the image</string>
     <string name="gutenberg_native_double_tap_to_open_bottom_sheet_with_available_options" tools:ignore="UnusedResources">Double tap to open Bottom Sheet with available options</string>
     <string name="gutenberg_native_double_tap_to_redo_last_change" tools:ignore="UnusedResources">Double tap to redo last change</string>
     <string name="gutenberg_native_double_tap_to_select" tools:ignore="UnusedResources">Double tap to select</string>
@@ -3291,7 +3289,6 @@
     <string name="gutenberg_native_double_tap_to_undo_last_change" tools:ignore="UnusedResources">Double tap to undo last change</string>
     <string name="gutenberg_native_duplicate_block" tools:ignore="UnusedResources">Duplicate block</string>
     <string name="gutenberg_native_edit_file" tools:ignore="UnusedResources">Edit file</string>
-    <string name="gutenberg_native_edit_focal_point" tools:ignore="UnusedResources">Edit focal point</string>
     <string name="gutenberg_native_edit_using_web_editor" tools:ignore="UnusedResources">Edit using web editor</string>
     <string name="gutenberg_native_edit_video" tools:ignore="UnusedResources">Edit video</string>
     <string name="gutenberg_native_error" tools:ignore="UnusedResources">Error</string>
```

I left a minor question regarding the exit path. Once answered / addressed this is good to go for me.

fastlane/Fastfile
loremattei: Yeah, I think so. It's used in many different places and I'm not confident about changing it right now. It would be great if @AliSoftware manages to tackle this while working on the release-toolkit and Fastfile refactoring/alignment project.

fastlane/Fastfile
loremattei: Right! I remember choosing it for a reason, but now I don't remember what I wanted to do 🤨 I've switched to `user_error` now

fastlane/Fastfile
AliSoftware: Yeah there's an issue logged about this already – https://github.com/wordpress-mobile/release-toolkit/issues/205 – and it would be nice to tackle it at some point, but currently felt low priority. So I agree we should keep the trailing slash (and the env var) so far until we get time to tackle this.

fastlane/Fastfile
loremattei: Yeah! This is something which may use some improvements. 
Here, I've just kept the same logic we have in the rest of the lanes that work on strings, but we surely want to make it more robust in the future.

fastlane/Fastfile
mokagio: 👍

fastlane/Fastfile
mokagio: Sounds good to me. One isolated change at a time 👍

mokagio: I re-read my original comment and noticed I missed a "not"

> That might **not** happen in a real execution of the release process lanes, because many have a check to ensure the repo is clean, but I wanted to point it out.

😅

mokagio (approved):
Thank you for considering my suggestions. For me, this is good to go.

fastlane/Fastfile
```diff
@@ -546,12 +570,12 @@ ENV["validate_translations"]="lintVanillaRelease"
...
       sh("mkdir -p #{build_dir}")
       unless is_ci
-        # We pre-build the Gutenberg RN bundle on a different job on CI,  
-        # so cleaning here breaks the build 
+        # We pre-build the Gutenberg RN bundle on a different job on CI,
+        # so cleaning here breaks the build
         UI.message("Cleaning branch...")
         sh("echo \"Cleaning branch\" >> #{logfile_path}")
         sh("./gradlew clean >> #{logfile_path} 2>&1")
-      end
```
oguzkocer, line 554: I was just skimming through this PR again and noticed this. I don't think we need this `unless is_ci` logic anymore since there is no CI job to bundle the JS files.

loremattei: Right, but also, we don't need it on CI anyway because we always start from a clean machine 🤷‍♂️. It just saves up a bit of time. But if you want to remove it, I'm fine with this.

oguzkocer: I don't mind keeping the code, but I think we should definitely change/remove the comment because it's referring to an action that no longer exists. Looking at the `fastlane/Fastfile`, there are other things that needs to be updated. For example, we can probably remove the `gutenberg_dep_check` lane. We could handle these in a separate PR.

loremattei: Yeah! Thinking more about it, it occurs to me that enabling clean-up on CI is not that bad. If we build more than one scheme in the same machine, it assures that we always start from a clean state. 
I'll open a new PR to address this.

loremattei: Done here: https://github.com/wordpress-mobile/WordPress-Android/pull/14337

hypest:
👋 all, I'm not familiar with Fastlane so, would appreciate helping me with this question: how's `ext.gutenbergMobileVersion` picked up by our Fastlane scripts? Is it doing some form of grepping of the build.gradle file directly, or does it use gradle to evaluate and return the value? Thanks!

loremattei:
>  Is it doing some form of grepping of the build.gradle file directly

This :-)

PR #14203, opened 2021-03-05
Update terms in code to use more inclusive language

### LINKED ISSUE #13619: Update terms in code to use more inclusive language
A quick `ack` of the repo shows `whitelist` and `blacklist` referenced in several places.  Let's update these references in our code and repo documentation (e.g. README) to use more inclusive terms.  Suggestions: `allowlist` or `blockedlist`. For references in third-party libs, we can see if there is an updated version of that lib and whether there is any blockers to switching to that version. 

See wp.me/paaHJt-15h-p2 for more.

renanferrari:
Fixes #13619

This PR replaces all references to "master" with "main" and all references to "blacklist" and "whitelist" with "denylist" and "allowlist" respectively. This includes class, variable and parameter names, comments, layout ids, string keys and more. Please, check this internal ref for more context on this: paaHJt-15h-p2

Note: The search for these terms was mostly limited to the `WordPress/src` directory, meaning it doesn't libraries and many configuration files. 

There were a few special cases that were handled a bit differently than the rest:

1. String keys on translated `strings.xml` files, which should be picked up by Glotpress automatically.
2. String constants, such as JSON fields, database fields, preference keys and event names, which were left as is. Examples:

https://github.com/wordpress-mobile/WordPress-Android/blob/e796f20f39146eb92e75c8f274246f614b940ff2/WordPress/src/main/java/org/wordpress/android/models/SiteSettingsModel.java#L54

https://github.com/wordpress-mobile/WordPress-Android/blob/e796f20f39146eb92e75c8f274246f614b940ff2/WordPress/src/main/java/org/wordpress/android/ui/prefs/WPComSiteSettings.java#L260

https://github.com/wordpress-mobile/WordPress-Android/blob/e796f20f39146eb92e75c8f274246f614b940ff2/WordPress/src/main/res/values/key_strings.xml#L139

3. User-facing strings, which were replaced with their updated counterparts from Calypso. Example:

https://github.com/wordpress-mobile/WordPress-Android/blob/e796f20f39146eb92e75c8f274246f614b940ff2/WordPress/src/main/res/values/strings.xml#L623

4. References to repository paths, which were replaced with an updated permalink shortened with `git.io`. Example:

https://github.com/wordpress-mobile/WordPress-Android/blob/e796f20f39146eb92e75c8f274246f614b940ff2/WordPress/src/main/java/org/wordpress/android/util/WPMeShortlinks.java#L9

5. Anything else that would lose its meaning if changed. Example:

https://github.com/wordpress-mobile/WordPress-Android/blob/e796f20f39146eb92e75c8f274246f614b940ff2/WordPress/src/main/java/org/wordpress/android/WordPress.java#L690

### To test

Note: This PR looks pretty big, but most of the code changes are quite simple, so this should not be a problem.

There's not much to test, but you can smoke test the most affected areas (like Site Preferences) and make sure everything still works like before.

You can also use the "Find in Path" feature on Android Studio to search for the terms and make sure nothing was missing.

---

PR submission checklist:


planarvoid (approved):
Looks good :+1:

PR #14204, opened 2021-03-05
Use binary configure

jkmassel:
This PR uses the gradle `configure` plugin instead of bundler, which should remove the need for a ruby toolchain for day-to-day development of WPAndroid.

**To Test:**

- Ensure all CI things pass
- Check out the repo from scratch, then switch to this branch. Run ./gradlew applyConfiguration. Note that it fails with the error "Build configuration file gradle.properties doesn't exist, follow README instructions". Copy gradle.properties-example to gradle.properties, then run ./gradlew applyConfiguration. This should replace the default values with the proper ones.

PR submission checklist:


.configure
```diff
@@ -1,45 +1,31 @@
...
+    },
+    {
+      "file": "android/WPAndroid/firebase.secrets.json",
+      "destination": ".configure-files/firebase-secrets.json"
     }
-  ],
-  "file_dependencies": [
-
   ]
 }
```
loremattei, line 31: Nitpick: very minor, but since we've tried to keep this consistent, we should add a new line here.

.gitignore
```diff
@@ -84,10 +84,17 @@ google-upload-credentials.json
...
-WordPress/*.jks
+# Configuration Secrets
+.configure-files/*
+!.configure-files/*.enc
+
+# Secrets that live inside the project
+*.properties.bak
+debug_a8c.keystore
+sentry.properties
+release.jks
```
loremattei, line 96: This is already covered by the `*.jks` rule below, but given how critical this file is, I'm really happy to see an extra layer of protection here :-)

loremattei:
I tested the tool in a few different situations and it worked as expected 👍 👌. I left a couple of minor comments about the diffs. 
I see that we're adding `upload.jks.enc`. Do we still need `automattic_upload.jks.enc`? 

A question to understand the workflow: how can users update or add new secrets with the new tool? Does the plugin support it or should people download the configure binary?

loremattei:
Also, it's worth noting that since this PR changes the name of `firebase-secrets.json`, we're triggering again the issue where switching branches leaves the unencrypted file ready to be committed 😅

loremattei:
Tested it a bit more and it looks like the new tool has some issues with checking out the right commit of the secrets repository. 
It seems that it can checkout a commit if it's already pulled, but it doesn't run `git pull` before trying to checkout the referenced commit.

jkmassel:
> It seems that it can checkout a commit if it's already pulled, but it doesn't run git pull before trying to checkout the referenced commit.

@loremattei – do you recall if this was the case when running `configure_update`? [The first step of the process is to fetch the latest secrets](https://github.com/Automattic/configure/blob/fe160acd77ab28f36fdc97f4066cc0fb9e627a72/src/configure.rs#L217-L229), so it's super odd to me that this would happen – just wondering if it's possible you can repro it? 😅

PR #14211, opened 2021-03-08
Use View Binding in Jetpack feature ViewHolders

planarvoid:
This PR introduces view bindings in the ViewHolders in the Jetpack features - Scan, Threats and Backup. I've tried several approaches and in the end I think this one is the best. It still encapsulates the layout inflation to the parent and makes the binding accessible from all the children. It also feels quite safe. Other things I tried:
- make the abstract `onBind` method like this `binding.onBind` - the advantage is that the children doesn't need to ever call the `binding` so it could be private. However, I don't think it makes the code much better or more readable to have the overriden calls look like `JetpackHeaderItemBinding.onBind`
- build the bindings directly in the ViewHolder children - I think passing the inflate function to the parent is safer cause it never happens that the binding is initialized incorrectly (without a parent)

To test:
- Go to a premium Jetpack site (I can give you account details)
- Test the Backup, Threat and Scan features
- Check they look normal

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/jetpack/common/viewholders/JetpackBackupRestoreFootnoteViewHolder.kt
```diff
@@ -14,29 +12,34 @@ class JetpackBackupRestoreFootnoteViewHolder(
...
     parent: ViewGroup
-) : JetpackViewHolder(R.layout.jetpack_backup_restore_list_footnote_item, parent) {
+) : JetpackViewHolder<JetpackBackupRestoreListFootnoteItemBinding>(
+        parent,
+        JetpackBackupRestoreListFootnoteItemBinding::inflate
+) {
     override fun onBind(itemUiState: JetpackListItemState) {
-        val state = itemUiState as FootnoteState
-        val resources = itemView.context.resources
+        with(binding) {
```
renanferrari, line 20: Really like how clean this gets when using `with()` 👍

WordPress/src/main/java/org/wordpress/android/ui/jetpack/common/viewholders/JetpackViewHolder.kt
```diff
@@ -1,17 +1,15 @@
...
 
-abstract class JetpackViewHolder(
-    @LayoutRes layout: Int,
-    parent: ViewGroup,
-    override val containerView: View = LayoutInflater.from(parent.context).inflate(layout, parent, false)
-) : RecyclerView.ViewHolder(containerView), LayoutContainer {
+abstract class JetpackViewHolder<T : ViewBinding>(
+    val parent: ViewGroup,
+    val inflateBinding: (LayoutInflater, ViewGroup, Boolean) -> T,
+    val binding: T = inflateBinding(LayoutInflater.from(parent.context), parent, false)
```
renanferrari: Do you think it makes sense to adjust the visibility of these parameters? 

```suggestion
    parent: ViewGroup,
    inflateBinding: (LayoutInflater, ViewGroup, Boolean) -> T,
    protected val binding: T = inflateBinding(LayoutInflater.from(parent.context), parent, false)
```

renanferrari (commented):
Thanks for the changes! I've tested the app and everything seems to be working as expected. Code is looking good and I really like the approach we're using to delegate the layout inflation to `JetpackViewHolder` 👍 I initially wondered if the `inflateBinding` function was really necessary, but seeing how it makes it easier to write new view holders, I think it pays off. (edit: I've simplified this a bit further on #14225 by using an extension function on the parent view. Could you please check it out and let me know what you think?)

I left just one small suggestion below.

planarvoid:
thanks for the review! The `viewBinding` extension function looks good. I think passing the parent view and the inflate function to the super constructor works very similarly though and doesn't introduce any additional complexity. I'm fine with using any of the approaches (or both).

renanferrari (approved):
LGTM 👍

PR #14215, opened 2021-03-09
Test - Use RN Gutenberg bridge built from the gutenberg-mobile CI

mokagio:
This is a test PR to verify that the binary that the `gutenberg-mobile` CI uploads to Bintray works. See https://github.com/wordpress-mobile/gutenberg-mobile/pull/3230.


mokagio:
This no longer needs to be an open PR.

PR #14217, opened 2021-03-09
Use view binding in stats

planarvoid:
This PR introduces ViewBindings in the stats, stats detail and stats view all screens. There were several tricky things to figure out. 
- When using the FragmentContainerView we now can't initialize the activity action bar in activity `onCreate` so I've moved it to fragment `onViewCreated`. 
- The `@SuppressLint("ClickableViewAccessibility")` is there because the way we catch the touches on the view pager makes the AS think that it affects accessibility. As far as I could see this is not the case so I've added the suppression. Let me know if you think otherwise. 
- I'm not sure my approach to `binding` is the best. I'm trying to pass it around as a parameter as much as possible and only use the field when the binding needs to be accessible from other methods than `onViewCreated`. When using `views` this is the safest approach because you can be sure the view is there. Here I'm not 100% sure.

To test:
- Go to Stats and notice it's loaded correctly
- Notice the action bar has the correct title and the back button is active
- Click on `View more` on a block on a site with a lot of data (you might need to go to the `Year` view)
- Notice the "View all" screen is loaded correctly
- Notice the action bar has the correct title and the back button is active
- Go to My Site/Blog posts
- Click on the overflow menu on one of the older blog posts (not an X-post) and go to "Post Stats"
- Notice the screen is loaded correctly
- Notice the action bar has the correct title and the back button is active

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/StatsFragment.kt
```diff
@@ -38,28 +39,37 @@ import javax.inject.Inject
...
+        super.onCreate(savedInstanceState)
         setHasOptionsMenu(true)
-        return inflater.inflate(R.layout.stats_fragment, container, false)
     }
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
         val nonNullActivity = requireActivity()
+        binding = StatsFragmentBinding.bind(requireView())
```
renanferrari: I believe we can use the parameter `view` here.
 
```suggestion
        binding = StatsFragmentBinding.bind(view)
```

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/StatsFragment.kt
```diff
@@ -38,28 +39,37 @@ import javax.inject.Inject
 
 private val statsSections = listOf(INSIGHTS, DAYS, WEEKS, MONTHS, YEARS)
 
-class StatsFragment : DaggerFragment(), ScrollableViewInitializedListener {
+class StatsFragment : DaggerFragment(R.layout.stats_fragment), ScrollableViewInitializedListener {
     @Inject lateinit var viewModelFactory: ViewModelProvider.Factory
     @Inject lateinit var uiHelpers: UiHelpers
     private lateinit var viewModel: StatsViewModel
     private lateinit var swipeToRefreshHelper: SwipeToRefreshHelper
+    private lateinit var binding: StatsFragmentBinding
```
renanferrari: I'm not sure if we should make the `binding` a `lateinit var` on fragments. The best practice seems to be to initialize [like this](https://developer.android.com/topic/libraries/view-binding#fragments) and then clear it on `onDestroyView()` to avoid memory leaks. That said, I've used this approach on #14224 and I'm not so sure I like it. I'm 100% open to other suggestions.

renanferrari (commented):
Thanks for the changes! I've tested the app and everything seems to be working as expected. Code is looking good overall but I'm also not sure about the approach to `binding`. I agree that passing it around as a parameter is the best option whenever possible, but as for when it needs to be accessed from other methods I'm still not sure what's the best approach as well.

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/StatsFragment.kt
planarvoid: you're right, I've missed this part 👍 . I'll change the variables

planarvoid:
good catch @renanferrari . I've changed the logic around binding a bit. What do you think about my changes? Instead of passing the binding as a parameter the methods are now in binding scope. I've also added clearing for the binding field.

WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/StatsViewAllFragment.kt
```diff
@@ -89,35 +84,52 @@ class StatsViewAllFragment : DaggerFragment() {
...
+                viewModel.onPreviousDateSelected()
+            }
         }
     }
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
         val nonNullActivity = requireActivity()
+        with(StatsViewAllFragmentBinding.bind(requireView())) {
```
renanferrari: Sorry, I missed this one the first time!

```suggestion
        with(StatsViewAllFragmentBinding.bind(view)) {
```

renanferrari (approved):
>Instead of passing the binding as a parameter the methods are now in binding scope.

I really liked that! Looks really clean 👌

Feel free to merge this once you addressed the suggestion below.

LGTM 👍

PR #14219, opened 2021-03-09
Add observe event method

planarvoid:
I've been thinking about this change for a long time and I think it's a good one in sense of removing a lot of boiler plate. It introduces a `observeEvent` extension method which only propagates the object encapsulated in the `Event` when it hasn't been consumed yet. I think the change is fairly straightforward. LMK what you think 👍 

To test:
- Smoke test the app, especially some of the parts using the events like Reader Discover screen
- Carefully check the code for typos

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/activitylog/list/filter/ActivityLogTypeFilterFragment.kt
```diff
@@ -89,7 +89,7 @@ class ActivityLogTypeFilterFragment : DialogFragment() {
         val parentViewModel = ViewModelProvider(requireParentFragment(), viewModelFactory)
                 .get(ActivityLogViewModel::class.java)
 
-        viewModel.uiState.observe(viewLifecycleOwner, Observer { uiState ->
```
renanferrari, line 92: Thanks for taking care of these redundant `Observer`s as well! 🙏

renanferrari (approved):
I really like this change, thanks for taking care of it!

LGTM! 👍

PR #14224, opened 2021-03-10
Use View Binding in Reader Activities and Fragments

renanferrari:
This PR introduces View Bindings for the Reader Activities and Fragments. It initializes and uses the `binding` object in a pretty standard way, following the best practices described [here](https://developer.android.com/topic/libraries/view-binding), but I'm not sure I like how it looks like for Fragments and I'm open to other suggestions.

Note: I've added a `@Suppress("ComplexMethod")` to the method that handles navigation events, as I can't think of a way to simplify it any further. Let me know if you any suggestions.

To test:

Smoke test the Reader, with special attention for the Discover section, the Interests screen and the Post Detail screen. Make sure everything is working as expected and nothing looks out of place. Also try to reblog a post with an account with no sites and make sure everything looks fine on the No Site To Reblog screen.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderFragment.kt
```diff
@@ -107,66 +112,66 @@ class ReaderFragment : Fragment(R.layout.reader_fragment_layout), ScrollableView
         }
     }
 
-    private fun initToolbar() {
+    private fun initToolbar(binding: ReaderFragmentLayoutBinding) = with(binding) {
```
planarvoid: I think I like a bit more the approach with `ReaderFragmentLayoutBinding.initToolbar()`. It allows you to call the `with` just once in the `onViewCreated`. Let me know what you think 👍

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderFragment.kt
```diff
@@ -39,6 +39,9 @@ class ReaderFragment : Fragment(R.layout.reader_fragment_layout), ScrollableView
     private var searchMenuItem: MenuItem? = null
     private var settingsMenuItem: MenuItem? = null
 
+    private var _binding: ReaderFragmentLayoutBinding? = null
+    private val binding get() = _binding!!
```
planarvoid: I know this is recommended but tbh I don't think it's necessary. I think it's ok to use directly the nullable field. At least the person who uses it will understand the field is alive only when the view is alive (and they can call `!!` consciously)

WordPress/src/main/java/org/wordpress/android/ui/reader/discover/ReaderDiscoverFragment.kt
```diff
@@ -61,107 +60,95 @@ class ReaderDiscoverFragment : ViewPagerFragment(R.layout.reader_discover_fragme
...
+        }
+        viewModel.navigationEvents.observeEvent(viewLifecycleOwner) { handleNavigation(it) }
+        viewModel.snackbarEvents.observeEvent(viewLifecycleOwner) { it.showSnackbar() }
+        viewModel.preloadPostEvents.observeEvent(viewLifecycleOwner) { it.addWebViewCachingFragment() }
+
         viewModel.start(parentViewModel)
     }
 
+    @Suppress("ComplexMethod")
+    private fun handleNavigation(event: ReaderNavigationEvents) = when (event) {
```
planarvoid, line 125: the only thing I can think of is pulling this out into a separate class (`ReaderNavigator`). I think it's ok to suppress the warning though.

planarvoid (commented):
thanks for the PR, It's looking good 👍 , consider my comments

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderFragment.kt
renanferrari: I totally agree. This has been addressed in b188e15.

WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderFragment.kt
renanferrari: Addressed in d9fb317.

planarvoid (approved):

PR #14226, opened 2021-03-10
Use view binding in accounts

planarvoid:
This PR introduces ViewBinding in all the `accounts` packages. Mostly the changes are related to login. I've cleaned up the layouts because I think they were redundant/overly complicated. As far as I can see the `WordPress/src/main/res/layout-sw600dp/post_signup_interstitial_activity.xml` was totally unnecessary + the `post_signup_interstitial_landscape` could be renamed and moved to the `landscape` package. Let me know if I'm missing something.

To test:
- Test the login flow works
- Test the Post signup interstitial on a phone in landscape

PR submission checklist:


renanferrari (commented):
Thanks for the changes! Code looks good and everything seems to be working fine on the app, but I do have one observation to make regarding the layouts cleanup:

>As far as I can see the `WordPress/src/main/res/layout-sw600dp/post_signup_interstitial_activity.xml` was totally unnecessary + the `post_signup_interstitial_landscape` could be renamed and moved to the `landscape` package.

At the time those screens were implemented, there was a requirement for the tablet's layouts to be the same in both portrait and landscape mode while keeping separate portrait and landscape layouts for smaller screens, which is why these layouts were arranged in this way as this achieves both things without introducing any extra code.

I can see a few ways in which we could improve this, but I'd rather wait for us to update the [`ConstraintLayout`](https://developer.android.com/jetpack/androidx/releases/constraintlayout) lib and use the new [Flow](https://developer.android.com/reference/androidx/constraintlayout/helper/widget/Flow) helper to achieve that. For now, I think I'd just revert the changes and maybe add some documentation to it. What do you think?

planarvoid:
that makes sense! I've changed the code. The downside (and the original reason why I made this change) is that when we have 2 included layouts with a different name, we have to handle 2 different bindings in the code (check e612b5d). I guess that's OK for now. This should be ready for another check @renanferrari

planarvoid:
I've actually removed the view bindings in the post_signup_interstitial_activity because having 2 different IDs was causing lint to fail and the code wasn't that good either. Now it uses a simple findViewById @renanferrari

renanferrari (approved):
Thanks for the changes!

LGTM 👍

PR #14227, opened 2021-03-10
Use view binding in activity log

planarvoid:
This PR replaces synth accessors with view bindings in the activity log classes (list, detail and the filters). In several places I've used `findViewById` to reach a view in the parent activity. I think it's the cleanest way forward at the moment but it's not a great approach. Changing it would require the activity to use the view model instead of the fragment. 

This line `requireActivity().findViewById<View>(R.id.progress)` is IMO incorrect. We're trying to show/hide an item in a viewholder. However, changing that is out of scope of this PR.

To test:
- Go to activity log
- Check it looks OK
- Go to an item detail
- Try several filters
- Check everything works as expected

PR submission checklist:


renanferrari (approved):
LGTM! 👍

PR #14229, opened 2021-03-10
Merge Release/16.8.1 into develop

JavonDavis:


JavonDavis:
@fluiddot what hash should the libs/gutenberg-mobile be resolved to?

fluiddot:
> @fluiddot what hash should the libs/gutenberg-mobile be resolved to?

I checked that that current reference in `develop` points to the tag `v1.48.0` so we should use that one because the hotfix points to a previous one (`v1.47.1`).

The `libs/gutenberg-mobile` hash should be: `12d7b2c24b2d0c7c836725a626c18377809bbb94`.

jkmassel (approved):

PR #14230, opened 2021-03-10
Add My Site feature flag handling to connected tests

renanferrari:
When the My Site Improvements feature flag is enabled, the layout of the My Site screen uses a `RecyclerView` instead of a single static structure to display its elements, which was causing some connected tests to fail locally.

This PR updates the `MySitePage` class so it can check if there's a `RecyclerView` being displayed and then handle actions accordingly. Other than that, this also removes the `MainNavBarComponent` since it could be replaced by `MySitePage` and performs some other general code cleanups.

To test:

- Run the optional connected tests on CI and make sure they pass.
- Run the connected tests locally, with the `my_site_improvements_enabled` flag both enabled and disabled, and make sure they pass.

PR submission checklist:


planarvoid (approved):
Changes are looking good 👍

PR #14232, opened 2021-03-10
Use viewModelScope instead of casting ViewModel to CoroutineScope

renanferrari:
The changes introduced by #14186 meant that a `ScopedViewModel` no longer implemented the `CoroutineScope` interface, but we missed a couple of cases where a `ViewModel` was being casted to a `CoroutineScope`, which in turn was causing a `ClassCastException` when trying to opening the photo/media picker. This PR uses the `viewModelScope` to replace those castings.

To test:

1. Open the app.
1. On the My Site screen, tap the Media button.
1. On the Media screen, tap the + button and select "Choose from the device".
1. Make sure the app doesn't crash.

PR submission checklist:


cameronvoell (approved):
Changes look good. I verified that we're no longer crashing when uploading a picture from device. 👍

PR #14235, opened 2021-03-11
Use view binding in domain registration

planarvoid:
This is a fairly straightforward change. The only tricky thing was that for some reason when the toolbar has ID `toolbar_main`, the binding crashes in runtime because it's missing `toolbar_text` view. This view is not in the layout so I guess there must be a bug in the view binding generation. Changing the ID of the toolbar fixed the issue.

To test:
- Go to a site with no domain and free credits
- Go through the domain registration and check nothing is out of place

PR submission checklist:


renanferrari (approved):
LGTM 👍

PR #14241, opened 2021-03-12
Use View Binding in Me screen

renanferrari:
This PR introduces View Bindings for the Me screen. Most of the changes introduced here are pretty similar to what we have already done in other classes. Like in #14217, a toolbar setup that was in an Activity had to be moved to a Fragment. 

Other than the Me activity and fragment, I'm also including changes to the `MainBottomSheetFragment` here, just because it was in the same package and it was pretty small.

To test:

Smoke test the Me screen and make sure everything is working as expected and nothing looks out of place. Main points of attention:

1. Try to upload a new avatar and make sure the progress bar shows up correctly and the process completes successfully.
1. Try to logout and make sure it works as expected.
1. Login with a self-hosted site and make sure the screen looks as expected.
1. Change the device language via system settings while the Me screen is open and make sure everything is updated, including the toolbar title. Repeat this but changing the language via app settings.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/MainBottomSheetFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/MainBottomSheetFragment.kt
@@ -6,16 +6,14 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.FrameLayout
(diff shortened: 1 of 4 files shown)


planarvoid (approved):
LGTM 👍

PR #14250, opened 2021-03-12
Use view binding in settings and stories

planarvoid:
This PR introduces view binding in two minor places in the app - the manual feature configuration and the stories prologue. The changes are straightforward. 

To test:
- Go to "Test feature configuration" in app settings
- Smoke test it
- Create a story
- Check the "Prologue" works as expected

PR submission checklist:


renanferrari (approved):
LGTM 👍

WordPress/src/main/java/org/wordpress/android/util/config/manual/ManualFeatureConfigActivity.kt
```diff
@@ -3,18 +3,20 @@ package org.wordpress.android.util.config.manual
 import android.os.Bundle
 import android.view.MenuItem
 import kotlinx.android.synthetic.main.manual_feature_config_fragment.*
```
renanferrari, line 5: Sorry, I just realized we missed this one!

PR #14251, opened 2021-03-12
Use View Binding in Reader views

renanferrari:
This PR introduces View Bindings for some Reader custom views. 

Bindings are initialized as soon as the views are created, by calling the `inflate()` method like this:

```kotlin
private val binding = ReaderPostDetailHeaderViewBinding.inflate(LayoutInflater.from(context), this, true)
```

One thing to note is that we have to pass the view itself as the `parent` parameter and set the `attachToParent` parameter to `true`.

Other than that, the changes are mostly similar to what we already had done and there was also an unused custom view that was removed along with its resources.

To test:

Smoke test the Reader and make sure everything is working as expected and nothing looks out of place. Main points of attention:

1. Open the Reader and tap on a post.
1. On the Post Detail screen, make sure that the section above the post's body is loaded correctly and displays some info about the post like tags, title, the blog's and author's names, etc.
1. Scroll to the end of the Post Detail screen and make sure that the Related Posts section was loaded correctly.
1. Tap on a tag and make sure that the Topic screen header is loaded correctly and displays the topic's name and a follow button.

PR submission checklist:


planarvoid (approved):
LGTM 👍

PR #14268, opened 2021-03-16
Add missing ID to the progress layout

planarvoid:
Fixes a crash that happens because of the removed top layout element ID in the progress_layout.

To test:
- Go to reader
- Click on an item
- Click on the overflow menu/Visit site
- The site is opened in the WebView

PR submission checklist:

### DIFF
--- a/WordPress/src/main/res/layout/wpwebview_activity.xml
+++ b/WordPress/src/main/res/layout/wpwebview_activity.xml
@@ -45,10 +45,10 @@
                 android:layout_height="match_parent" />
 
             <include
+                android:id="@+id/progress_layout"
                 layout="@layout/progress_layout"
                 android:layout_width="match_parent"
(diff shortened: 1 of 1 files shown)


WordPress/src/main/res/layout/wpwebview_activity.xml
```diff
@@ -45,10 +45,10 @@
                 android:layout_height="match_parent" />
 
             <include
+                android:id="@+id/progress_layout"
                 layout="@layout/progress_layout"
                 android:layout_width="match_parent"
                 android:layout_height="match_parent"
-                android:layout_above="@+id/navbar_container"
```
planarvoid, line 51: AS marked this as wrong because the referenced view is in another Relative layout

zwarm (approved):
LGTM. Thanks for taking care of this so quickly. :)

PR #14269, opened 2021-03-16
Remove homepage picker feature config

### LINKED ISSUE #14261: Remove homepage picker flag
The `HOME_PAGE_PICKER` is always true in `build.gradle`. Let's remove it and cleanup the unused code.

planarvoid:
Fixes #14261 

This PR removes the homepage picker flag and all the code that's unused now.

To test:
- Go through the site creating flow
- Make sure the Home page picker is shown correctly

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/sitecreation/SiteCreationActivity.kt
```diff
@@ -30,25 +30,21 @@ import org.wordpress.android.ui.sitecreation.previews.SitePreviewViewModel.Creat
...
-import org.wordpress.android.ui.sitecreation.segments.SegmentsScreenListener
-import org.wordpress.android.ui.sitecreation.segments.SiteCreationSegmentsFragment
 import org.wordpress.android.ui.sitecreation.theme.HomePagePickerFragment
 import org.wordpress.android.ui.sitecreation.theme.HomePagePickerViewModel
 import org.wordpress.android.ui.utils.UiHelpers
-import org.wordpress.android.util.config.HomePagePickerFeatureConfig
 import org.wordpress.android.util.wizard.WizardNavigationTarget
 import javax.inject.Inject
 
+@Suppress("TooManyFunctions")
```
antonis, line 39: This is not a request for change. Just a question to make sure I'm following the correct pattern in other cases.
Should we add `@Suppress` annotations in the code is such cases or add the exception in the`baseline.xml`? cc @ParaskP7

antonis (approved):
Great job @planarvoid 👍 
I tested the app and the cleanup did not affect the site creation workflow. The code changes are consistent and LGTM 🎉 
Thank you for taking care of this 🙇

WordPress/src/main/java/org/wordpress/android/ui/sitecreation/SiteCreationActivity.kt
ParaskP7: 👋 @antonis !

Thanks for pinging me on this. 🙏 

The way I think about it is that we should be adding `@Suppress` annotations, that is if we really need to and can't avoid dealing with the warning.

The `baseline.xml` should be considered as temporary and was created to have us start from point zero, with some initial suppressed tech debt. As such, I personally suggest that we only update or remove any lines in this baseline file, but never add new ones. The goal would be to progressively clean this baseline file with the ultimate goal of getting rid of that file and configuration altogether.

If that helps, another way to think about it, is imagining what you would do if there was no baseline file to start with. In that case, one would only have one option, to suppress it with the annotation. The same should apply here.

Let me know if you have another opinion on that. 🙏

antonis: Thank you for the detailed response @ParaskP7 🙇 

> Let me know if you have another opinion on that.

No. I agree with that. Just wanted to make sure that I'll follow the correct approach in my code 🙏

PR #14275, opened 2021-03-18
Remove UL&S flag

renanferrari:
Fixes #14262

This PR simply removes the usage of the `UNIFIED_LOGIN_AVAILABLE` flag from the app and remove its from the `build.gradle` file. It doesn't remove most of the unused code and doesn't change anything on the login library. These were handled in #14276.

The biggest changes here are related to the Prologue layouts which was simplified a lot due to the removal of additional included layout variations. Some layout ids were also renamed, so some UI tests also had to be updated.

### To test

**WordPress.com**

- Log out from the app.
- On the Prologue screen, make sure everything looks as expected.
- Try rotating the device and make sure the landscape layout is loaded (also try this with a tablet, and make sure the layout is the same on both landscape and portrait).
- Go through the login flow and make sure everything works as expected.

**Self-Hosted**

- Login with a self-hosted site.
- From the Me screen, try to login with a WordPress.com account.
- Go through the login flow and make sure everything works as expected.

**Other**

Run UI tests and make sure they all pass.

---

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #14276, opened 2021-03-18
Remove unused UL&S code

### LINKED ISSUE #14274: Remove unused code from the login library
As a follow up to #14262, after the `UNIFIED_LOGIN_AVAILABLE` flag is removed, we should also remove all the unnecessary code and resources from the login library.

renanferrari:
Fixes #14274

This PR is a follow up to #14275 – which removed the `UNIFIED_LOGIN_AVAILABLE` flag from the app – and removes all the code that was not going to be used by the login library anymore. 

Note: I'm aware the changelog seems huge, but most of it consists of completely removed files and resources, so it _shouldn't_ be as hard to review as it looks.

The biggest change in terms of layout is that the base form layout now only has one button at the bottom (now called `bottom_button`) instead of two (previously called `primary_button` and `secondary_button`), as this change required successive changes to other layouts. 

Another significant change is that a few listener callbacks on `LoginActivity` are being completely removed (or just renamed, in case of `onTermsOfServiceClicked`). 

Other than that, it's also worth noting that all unused layouts, drawables and resources such as strings, dimensions and colors were either completely removed or moved out from the login library in cases where they were still being referenced outside the library. These changes were made in separate commits, so it's a bit easier to track them down.

Note: I'll prepare a new PR to push these changes back to the library once this is merged.

### To test

Run UI tests and make sure they all pass.

Smoke test the login flow and try out as many different flows as you feel is necessary and make sure they all are working as expected.

---


libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginEmailFragment.java
```diff
@@ -580,10 +440,10 @@ protected void buildToolbar(Toolbar toolbar, ActionBar actionBar) {
             return;
         }
 
-        if (mShouldUseNewLayout) {
-            actionBar.setTitle(R.string.get_started);
-        } else {
+        if (mOptionalSiteCredsLayout) {
```
planarvoid, line 443: Doesn't this change the previous behaviour a bit? If I understand correctly, previously it was just `actionBar.setTitle(R.string.get_started);` because the flag `mShouldUseNewLayout` was always true, right?

planarvoid (commented):
I've tested the following:
- self-hosed login
- google login
- smart lock login
- wpcom login
and all work well 👍 . I have one question about the code.

libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginEmailFragment.java
renanferrari: `mShouldUseNewLayout` wasn't always `true` though. I mean, it was for WordPress, but not for Woo. When `mOptionalSiteCredsLayout` was `true`, then `mShouldUseNewLayout` would necessarily be `false`. 

That said, I believe most Woo cases would fall under the check above this one, where we check if the current login mode is `WOO_LOGIN_MODE`. I decided to leave this the way it is just in case I was missing something. Does that make sense?

planarvoid: got it, thanks!

planarvoid (approved):
:shipit:

PR #14286, opened 2021-03-19
ExPlat: Add A/B experiment

renanferrari:
This PR is a follow up to #13960 and introduces a A/B experiment using our ExPlat client.

The experiment is pretty simple and consists of changing the position of the "Story post" action inside the menu that appears when the main FAB is tapped. 

When the variation is `Control`, the action is shown at its default position, at the bottom of the list. When the variation is `Treatment`, the action is shown at the top of the list:

| Control | Treatment |
| ------- | ----------- |
|story-last|story-first|

### To test

Notes: 
- You'll need a tool like Charles to intercept the ExPlat endpoint response.
- The ExPlat endpoint looks like this: `/wpcom/v2/experiments/0.1.0/assignments/wpandroid/`.

1. Open the app.
1. On Charles, make sure you see the normal response from the ExPlat endpoint.
1. On the My Site screen, tap the FAB.
1. Notice that the menu actions have the following order: "Blog post", "Site page" and "Story post".
1. Close the app and make sure to stop it.
1. Open the app again.
1. On Charles, intercept and edit the response so it looks like this: 
```json
{
	"variations": {
		"wpandroid_create_menu_story_first": "treatment"
	},
	"ttl": 60
}
```
8. On the My Site screen, tap the FAB.
9. Notice that the menu actions have the following order: "Story post", "Blog post", "Site page".

Here's a video demonstrating this process:

https://user-images.githubusercontent.com/830056/111847100-f2a9c500-88e6-11eb-8387-b71f40309c7f.mov

Here are my breakpoint settings on Charles:

<p float="left">
breakpoint-settings-1
breakpoint-settings-2
</p>

Let me know if you need additional assistance with any of the steps above:

---

PR submission checklist:


develric (approved):
Hi @renanferrari 👋 ! I tested this as per test steps and all works as described 👍 ! I'm not familiar with the project or this specific experiment but AFAIU also code looks good. Checked the analytics tracking as well and it is consistent!

**Side notes:** placing a couple of notes I happened to think about while reviewing; probably not relevant to this specific PR so in case feel free to skip them 🙇 

1. I played a bit with the `ttl` response parameter setting it to like 10 seconds, changing the endpoint response from having the variant or not having it and forcing a refresh putting the app in background and back again in foreground. The endpoint was correctly fetched with the new experiment configuration, but the behaviour in the menus list order was kept as the original one (only closing/opening back the app triggers a change in the list order as per your test steps). AFAIU I think this is a wanted behaviour given we are keeping the cached value in `ExPlat > activeVariations`, but thought to better mention in case I'm missing anything.
2. I see we are not changing the order of the actions in the FAB inside the Posts section. I guess this may be also wanted to keep the experiment scope more `focused`, but again I thought I better mention than not 😄 
 
Not sure if you need other people in the list to look into it (especially on the data/experiment side) so I'm approving without merging right now; if this is not the case please just let me know or given you are targeting upcoming milestone feel free to merge if the side notes above are not relevant to this PR as I would expect 🙇

mzorz:
> Not sure if you need other people in the list to look into it (especially on the data/experiment side) so I'm approving without merging right now; if this is not the case please just let me know or given you are targeting upcoming milestone feel free to merge if the side notes above are not relevant to this PR as I would expect

👋 @develric @renanferrari I'm planning to review this one tomorrow morning but as @develric said, given the plans to merge to this milestone please feel free to merge given @develric already gave it a pass. I went through the code briefly and it looks good to me 👍 .

renanferrari:
@develric Thanks a lot for the review! Both of your assumptions are correct: if the variation is active, the behavior shouldn't change and this experiment only affects the main FAB menu, not the one in the Posts section 👍

@mzorz Thank you for looking into this as well 🙂

I'm going to go ahead and merge this, since @aaronyan has also looked into it already.

PR #14289, opened 2021-03-22
Use HTTP API calls to trigger release jobs

loremattei:
This PR updates our release tools to trigger any CI release job via an API call instead of relying on the automatic trigger on tags.
This way, it will be easier and cleaner to handle any failure during the release build and, moreover, it'll be possible to move more steps of the releases to CI (for example, validation and lint steps that can fail by design).

Note that in order to use the release tools, users should configure a new token for CircleCI on their local environment, following these steps:

- Go to `https://app.circleci.com/settings/user/tokens`.
- Create a Personal API Token.
- Add `CIRCLE_CI_AUTH_TOKEN=<Your Token>` to your `~/.wpandroid-env.default` (you can check out the updated `fastlane/env.example` for a template).

**To test:**
Full testing is tricky because it's tight to the release cycle. Anyway, it's possible to test the critical part of it this way:

- checkout a new `release/test` branch from this one and push it.
- add a dummy commit (the build will fail, so adding a dummy commit will avoid the failure to be reported to this PR).
- `bundle install`
- Run `bundle exec fastlane trigger_release_build branch_to_build:"release/test"`
- Verify that the release build job has been started on CircleCI. The job will fail during the upload to PlayStore because the current version is already present. This is fine.
- Delete the test branch.

`Review`
I've asked the review to @JavonDavis because he is currently the release manager of WordPress Android. All the code is already in production in WPiOS, so it should be fine, but there may always be some nuances with porting that I missed, so I think that understanding the changes will help if something happens during the next release.

PR submission checklist:


loremattei:
cc @AliSoftware as this may be relevant to the lane refactoring  you're working on.

JavonDavis (approved):
Changes look good @loremattei 👏 and job triggered successfully on CI from the test branch I created https://app.circleci.com/pipelines/github/wordpress-mobile/WordPress-Android

JavonDavis:
@loremattei  I pulled in develop to resolve the conflict with the fastlane/env.example. I'll also leave a little more time in case @AliSoftware had any comments

.circleci/config.yml
```diff
@@ -15,13 +15,19 @@ parameters:
   generate_screenshots:
     type: boolean
     default: false
+  beta_build:
+    type: boolean
+    default: false
+  release_build:
+    type: boolean
+    default: false
```
AliSoftware: Maybe I missed something, but it seems to me that those are _always_ OR'd at call site, so wondering if we need 2 separate parameters insted of a single one?

 - Every workflow that runs when non is set is using `unless: { or: … }` containing both `beta_build` and `release_build` in the `or`
 - The "Release Build" workflow runs if either of those is set (`when: { or: … }` with both those params)
 - I haven't seen anywhere else in the config (scripts param substitutions, etc)

So maybe unifying them under a single one could simplify the config?

.circleci/config.yml
```diff
@@ -481,21 +491,16 @@ workflows:
     when: << pipeline.parameters.generate_screenshots >>
     jobs:
       - raw-screenshots
-      - promo-screenshots
+      - promo-screenshots:
+          requires: [raw-screenshots]
```
AliSoftware, line 490: 😱 Surprised we never had any issue before with this missing dependency… Nice catch!

AliSoftware (commented):
Overall code LGTM 👍  Pending the question where we could simplify the CI config by merging the params.

_Note: Haven't run the test steps yet (only reviewed code in GH)_

.circleci/config.yml
loremattei: Yeah, right! I initially wanted to keep it to align the infrastructure with the iOS one, but the rest of the flow is different enough that this actually doesn't make sense as we can discriminate beta builds in a safer way on Android.
Updated in 40c94d0.

loremattei:
> I pulled in develop to resolve the conflict with the fastlane/env.example.

Thank you @JavonDavis!

loremattei:
Hey @AliSoftware do the latest changes address your comment?

fastlane/Fastfile
```diff
@@ -109,7 +110,8 @@ ENV["validate_translations"]="lintVanillaRelease"
   desc "Creates a new release branch from the current develop"
```
AliSoftware, line 110: _Not to self (for my lanes alignment task): update this desc_

fastlane/Fastfile
```diff
@@ -397,6 +403,24 @@ ENV["validate_translations"]="lintVanillaRelease"
     end
   end
 
+  #####################################################################################
+  # trigger_release_build
+  # -----------------------------------------------------------------------------------
+  # This lane triggers a stable release build on CI
+  # -----------------------------------------------------------------------------------
+  # Usage:
+  # bundle exec fastlane trigger_release_build [branch_to_build:<branch_name>]
```
AliSoftware: small nit: param is not really optional so remove the `[]`

```suggestion
  # bundle exec fastlane trigger_release_build branch_to_build:<branch_name>
```

AliSoftware (approved):
Hey @loremattei sorry for the delay, wanted to test this yesterday but got caught up by other pings.
Thanks for addressing the feedback, LGTM now 👌  :shipit:

fastlane/Fastfile
loremattei: Right! Good catch!

PR #14325, opened 2021-03-23
Remove WP Stories feature config and the unused code

### LINKED ISSUE #14318: Remove WP Stories feature flag
The task is to remove `WPStoriesFeatureConfig` and the now unused code.

planarvoid:
Fixes #14318 

This PR removes the WPStoriesEnabled flag because it was already released to all the users and it will not be reverted. The code should be quite straightforward.

To test:
- Check you can correctly add stories on a WPCom site
- Check the stories are disabled on a non-Jetpack self-hosted site

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModel.kt
```diff
@@ -321,7 +319,7 @@ class WPMainActivityViewModel @Inject constructor(
     }
 
     private fun shouldShowStories(site: SiteModel?): Boolean {
-        return wpStoriesFeatureConfig.isEnabled() && SiteUtils.supportsStoriesFeature(site)
+        return SiteUtils.supportsStoriesFeature(site)
```
mzorz: What do you think about removing this simple method and replacing wherever it's called with `SiteUtils.supportsStoriesFeature(site)` directly?

mzorz: We can leave it if you think it adds clarity - it's just that the occurrences where this is used are already giving reading clarity by assigning the return value to a descriptive named val i.e. `val shouldShowStories = shouldShowStories(site)`

planarvoid: I've implemented your suggestion in a9b2936421, I think it makes sense 👍

planarvoid:
I've implemented your suggestion so this should be ready for another check @mzorz 👍

mzorz (approved):
Tested as per the instructions and worked as expected 👍 
LGTM :shipit:

PR #14337, opened 2021-03-24
Update lanes taking into account the new build process

loremattei:
This PR cleans up the Fastfile removing a couple of checks which aren't required anymore: 
1. removes the `gutenberg_dep_check` lane as it's based on having the submodule available. 
2. removes a check which prevented cleaning up the local build environment before building on CI. 

_About 1_:
I think we should implement an equivalent check based on the version referenced in `build.gradle`. We could mimic what we do on iOS where, during the code freeze, we verify that we don't import libraries referencing random commit hashes or beta versions. At least, we should add a manual check step to the release scenario before merging this feature branch into `develop`.

_About 2_: 
Cleaning up on CI is probably not required most of the times because we start from a clean image. But it shouldn't take a lot of time anyway and it may still help when we build multiple flavours in the same machine. 

PR submission checklist:

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -65,7 +65,6 @@ ENV["validate_translations"]="lintVanillaRelease"
   #####################################################################################
   desc "Creates a new release branch from the current develop"
   lane :code_freeze do | options |
-    gutenberg_dep_check()
     old_version = android_codefreeze_prechecks(options)
 
     android_bump_version_release()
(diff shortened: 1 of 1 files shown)


mchowning:
This looks good to me, but I'll let Oguz or Gio give the official 👍

oguzkocer (approved):
> I think we should implement an equivalent check based on the version referenced in build.gradle. We could mimic what we do on iOS where, during the code freeze, we verify that we don't import libraries referencing random commit hashes or beta versions. At least, we should add a manual check step to the release scenario before merging this feature branch into develop.

I agree. I think for the moment we just want to verify that the version we are using has the `develop-{commit_hash}` format. I'll open a PR to manually check this during release process. 

>  Cleaning up on CI is probably not required most of the times because we start from a clean image. But it shouldn't take a lot of time anyway and it may still help when we build multiple flavours in the same machine.

This should only add 10 seconds or so to the build, so I think it's fine to run it in CI as well.

PR #14340, opened 2021-03-24
Use view binding in stats detail, view all and list fragments

planarvoid:
This PR contains view binding changes and one thing that I've noticed while trying to access the nested stats list fragment in the stats detail fragment. 

I'm pretty sure that the date selector handling in the Stats detail fragment was redundant. The detail fragment contains the list fragment (initialized with the Detail type). This list fragment is already handling the date selection and updates the data based on the selected date. The only thing that the detail fragment was doing in addition was the dateSelector.clear method which I've moved to the `DetailListViewModel`. The `DetailListViewModel` and the detail use case are used only in the detail activity/fragment. I've tested the stats detail and it seems to work as expected. I think that this change removes duplicate data loading.  

To test:
- Go to a site with a lot of data
- Go to blog posts
- Click on an overflow menu on a new post
- Click on "Stats"
- Notice the Stats detail screen
- Try going back and forward to switch dates with "<" and ">"
- Notice the date is changing the bar chart correctly
- Try selecting the date in the bar chart
- Notice it updates the date selector
- Select an older date
- Leave the screen
- Go to a different post stats
- Notice the previously selected older date is NOT selected
- Notice Today's date is selected

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/StatsViewAllFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/StatsViewAllFragment.kt
@@ -215,7 +215,7 @@ class StatsViewAllFragment : DaggerFragment(R.layout.stats_view_all_fragment) {
         })
 
         viewModel.dateSelectorData.observe(viewLifecycleOwner, { dateSelectorUiModel ->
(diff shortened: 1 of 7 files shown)


renanferrari (approved):
Thanks for the changes! The app is working well and the code looks good. 

LGTM 👍

PR #14343, opened 2021-03-24
Do not track marking post as seen when opening it.

khaykov:
This PR remove tracking of marking post as seen when unseen post is opened. This is done to align stats with iOS and Calypso.

To test:
Put breakpoint here, and perform following actions:
- Open unseen post in reader. Confirm that the breakpoint is not being hit.
- Using post menu in Post Details, mark post as Seen, and then mark it as Unseen. Confirm that the breakpoint is being hit.
- From the reader post list, use post menu to mark post as Seen. Confirm that the breakpoint is being hit.

PR submission checklist:


develric (approved):
Hey @khaykov 👋 , code looks good and the tested behaviour is as expected 👍 . Sharing a generic thought I happen to think of but I just think we are fine to go here 🙇 . In general I prefer to have more data than less, so I would have considered to keep the track by adding another source (like `auto-bump-by-opening` or maybe better than this poor name I could think of 😄 ). But this said I think in this case we are fine since we are aligning to calypso/ios as you reported 👍  :shipit:

Issue #14299, opened 2021-03-22
Jetpack App (Basics): Handle Sites


ParaskP7:
Parent #14293

- Research requirements.
- Remove `Connect another site` from the epilogue.
- Restrict app to only show Jetpack Sites.
- Disable ability to create a new site.
- Update API's to filter for Jetpack blogs only. We can use same logic cloud.jetpack.com uses (`me/sites?filter=jetpack`).

PR #14356, opened 2021-03-26
Jetpack App Basics: Handle sites

ashiagr:
Issue #14299

This is a draft PR for Jetpack app sites handling. Based on the Jetpack or WordPress app build variant, it controls:
1. "Connect another site" visibility on  the login epilogue
2. Jetpack sites filtering on the My Sites tab -> Site Picker
3. "Add new site" menu visibility on the My Sites tab -> Site Picker

`SiteStore.fetchSites` is accessed from different classes mentioned below: 
- **WordPress class**

    - `WordPress.java`
    - `JetpackConnectionResultActivity.java`
    - `SitePickerActivity.java`

- **Login Lib classes**

    - `LoginBaseFormFragment`
    - `LoginWpcomService`

- **FluxC Example Test classes**

    - `MockedStack_SiteTest`
    - `ReleaseStack_ActivityLogTestJetpack`
    - `ReleaseStack_InsightsTestJetpack`
    - `ReleaseStack_ManageInsightsTestJetpack`
    - `ReleaseStack_MediaTestJetpack`
    - `ReleaseStack_PluginTestJetpack`
    - `ReleaseStack_PostSchedulingTestJetpack`
    - `ReleaseStack_ScanTestJetpack`
    - `ReleaseStack_SiteTestJetpack`
    - `ReleaseStack_SiteTestWPCom`
    - `ReleaseStack_ThemeTestJetpack`
    - `ReleaseStack_TimeStatsTestJetpack`
    - `ReleaseStack_WPComBase`

- **FluxC Example app**
    - `SiteSelectorDialog`
    - `SiteActionBuilder`
    - `MainInstafluxActivity`

Instead of introducing a new boolean param in the `SiteStore.fetchSites` for jetpack filtering and setting it from the WordPress or  Login modules based on the build variant, FluxC PR: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1936 adds the `filters=jetpack` [request param](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/fac74b5d124b0c8670706b28e77339ba0d7ceab6/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java#L158) in the `SiteRestClient` if it has a Jetpack app context. Note that [app context already exists](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java#L145-L147) in the class constructor.

Going this route, we don't need to make any changes to the login lib for this task and also lesser number of classes will be affected since we're not adding new param to the `fetchSites` method.

@zwarm, I discussed with you already, here's how it will look. 
@zwarm, @ParaskP7 let me know what you think of it. I'll proceed accordingly.  

**Pre test**

You must update your local copy of `google-services.json` before running the app.
-- Copy existing `org.wordress.android`, `org.wordress.android.prealpha`, `org.wordress.android.beta` sections and replace with `com.jetpack.android`,` com.jetpack.android.prealpha`, `com.jetpack.android.beta`.

**To Test**

Build and launch below product flavors:
`wordpressJalapenoDebug`
`jetpackJalapenoDebug`

Verify that for Jetpack build variant:
- "Connect another site" is removed from the login epilogue
- Only "jetpack" sites for the user  are shown on the My Sites tab -> Site Picker
- Create a new site ability is disabled on the My Sites tab -> Site Picker

Verify that for WordPress build variant:
- "Connect another site" is present on the login epilogue
- All sites for the user are shown on the My Sites tab -> Site Picker
- Create a new site ability is present on the My Sites tab -> Site Picker

**Merge Instructions**

This PR targets `feature/jetpack-app-basics`.
We'll decide next week if we want to split these changes for WordPress and Jetpack app based on how we'd like to merge corresponding FluxC PR. 
Till that time, do not merge this PR.

PR submission checklist:


ParaskP7:
> Instead of introducing a new boolean param in the SiteStore.fetchSites for jetpack filtering and setting it from the WordPress or Login modules based on the build variant, FluxC PR: wordpress-mobile/WordPress-FluxC-Android#1936 adds the filters=jetpack request param in the SiteRestClient if it has a Jetpack app context. Note that app context already exists in the class constructor.
> 
> Going this route, we don't need to make any changes to the login lib for this task and also lesser number of classes will be affected since we're not adding new param to the fetchSites method.
> 
> @zwarm, I discussed with you already, here's how it will look.
> @zwarm, @ParaskP7 let me know what you think of it. I'll proceed accordingly.

👋 @ashiagr !

I just saw this change and I left a single comment on the `FluxC` PR. Let me know what you think about that. I actually feel that it is better to have the `isJetpack` value flow from the client and into the `FluxC` library and suggested to utilise a `BuildConfig` boolean to derive whether this is a `Jetpack` app or not (instead of depending on the package name, which is a string and potentially more dynamic, susceptible to change).

ashiagr:
@ParaskP7, @zwarm 

I updated this PR as per our `isJetpackApp` handing discussion and also updated corresponding FluxC PR. 
Can you give it another look?

🙇‍♀️

libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginBaseFormFragment.java
```diff
@@ -332,11 +336,20 @@ public void onAccountChanged(OnAccountChanged event) {
             mDispatcher.dispatch(AccountActionBuilder.newFetchSettingsAction());
         } else if (event.causeOfChange == AccountAction.FETCH_SETTINGS) {
             // The user's account settings have also been fetched and stored - now we can fetch the user's sites
-            mDispatcher.dispatch(SiteActionBuilder.newFetchSitesAction());
+            fetchUserSites();
```
ParaskP7: 1) Warning (⚠️): Consider using `ArrayList<>` instead of a raw parameterized class `ArrayList`. PS: Same apply everywhere.
2) Suggestion (💡): Instead of extracting the `fetchUserSites()` I would have done it a bit differently. I would have extracted the `fetchSitesPayload()` instead and keep the same line as before, like `mDispatcher.dispatch(SiteActionBuilder.newFetchSitesAction(fetchSitesPayload()));`.

PS: Having that specific `fetchSitesPayload()` function extracted, I would further extract the `isJetpackAppLogin()` within this one, so in the end it would look like:

```
private FetchSitesPayload fetchSitesPayload() {
    ArrayList<SiteFilter> siteFilters = new ArrayList<>();
    if (isJetpackAppLogin()) siteFilters.add(SiteFilter.JETPACK);
    return new FetchSitesPayload(siteFilters);
}

private boolean isJetpackAppLogin() {
    return (mLoginListener instanceof LoginListener)
                    && ((LoginListener) mLoginListener).getLoginMode() == LoginMode.JETPACK_APP_LOGIN_MODE;
}
```

Then, since the same will apply on the `LoginWpcomService.java` class, I would do it there similarly:
```
private FetchSitesPayload fetchSitesPayload() {
    ArrayList<SiteFilter> siteFilters = new ArrayList<>();
    if (mIsJetpackAppLogin) siteFilters.add(SiteFilter.JETPACK);
    return new FetchSitesPayload(siteFilters);
}
```

This way, it become obvious that this duplicate code can be unified and reused on both placed through a utility static method of some kind, which will accept the `isJetpackAppLogin` as a method parameter, because that parameter is the differentiator here.

PSS: Similarly, and on the `WordPress` module now, I would have a similar method be extracted, put in a utils class and reuse it. This time, this method would not need any method parameter as it will always be reused with the `BuildConfig.IS_JETPACK` logic. This way a lot of the duplication will get removed and the code will become more readable.

```
private FetchSitesPayload fetchSitesPayload() {
    ArrayList<SiteFilter> siteFilters = new ArrayList<>();
    if (BuildConfig.IS_JETPACK) siteFilters.add(SiteFilter.JETPACK);
    return new FetchSitesPayload(siteFilters);
}
```

libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginMode.java
```diff
@@ -10,7 +10,8 @@
     WPCOM_LOGIN_DEEPLINK,
     WPCOM_REAUTHENTICATE,
     SHARE_INTENT,
-    WOO_LOGIN_MODE;
+    WOO_LOGIN_MODE,
+    JETPACK_APP_LOGIN_MODE;
```
ParaskP7: Question (❓): What if this enum was called similar to the `WPCOM_LOGIN_ONLY` one, like `JETPACK_LOGIN_ONLY`. Since its starting point comes from the `ActivityLauncher.showSignInForResult(...)` method, I find it to be a more consistent naming. Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
```diff
@@ -274,6 +275,7 @@ public LoginMode getLoginMode() {
     private void loggedInAndFinish(ArrayList<Integer> oldSitesIds, boolean doLoginUpdate) {
         switch (getLoginMode()) {
```
ParaskP7, line 278: Suggestion (💡): Consider adding the default `WOO_LOGIN_MODE` case as well to get rid of the switch warning.

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
```diff
@@ -1172,13 +1173,15 @@ public static void showMainActivityAndSiteCreationActivity(Activity activity) {
...
+        showSignInForResult(activity, false, false);
     }
 
-    public static void showSignInForResult(Activity activity, boolean isWpComOnly) {
+    public static void showSignInForResult(Activity activity, boolean isWpComOnly, boolean isJetpackOnly) {
         Intent intent = new Intent(activity, LoginActivity.class);
         if (isWpComOnly) {
             WPCOM_LOGIN_ONLY.putInto(intent);
+        } else if (isJetpackOnly) {
+            JETPACK_APP_LOGIN_MODE.putInto(intent);
```
ParaskP7: 1) Question (❓): Is the `if/else if` what we really want here? Doesn't that mean that only one of the intents will get set no matter the parameter input. What I mean is that if someone tries to call this method with both the `isWpComOnly` and `isJetpackOnly` set as `true`, only the first value will matter, isn't it? This might become a bit misleading from the api point of view. Wdyt?
2) Suggestion (💡): In addition to the above, I was having a bit of trouble going through this logic and more specifically on the usage and api side of things. What if we completely rewrite that to the below:
```
public static void showSignInForResult(Activity activity) {
    Intent intent = new Intent(activity, LoginActivity.class);
    activity.startActivityForResult(intent, RequestCodes.ADD_ACCOUNT);
}

public static void showSignInForResultWpComOnly(Activity activity) {
    Intent intent = new Intent(activity, LoginActivity.class);
    WPCOM_LOGIN_ONLY.putInto(intent);
    activity.startActivityForResult(intent, RequestCodes.ADD_ACCOUNT);
}

public static void showSignInForResultJetpackOnly(Activity activity) {
    Intent intent = new Intent(activity, LoginActivity.class);
    JETPACK_APP_LOGIN_MODE.putInto(intent);
    activity.startActivityForResult(intent, RequestCodes.ADD_ACCOUNT);
}
```

PS: To me the above suggestion make is much more straightforward what is what and make the intention more clear. Wdyt? For example, with the above suggestion in place this:

```
if (BuildConfig.IS_JETPACK) {
    ActivityLauncher.showSignInForResult(this, false, true);
} else {
    ActivityLauncher.showSignInForResult(this);
}
```

Will become:

```
if (BuildConfig.IS_JETPACK) {
    ActivityLauncher.showSignInForResultJetpackOnly(this);
} else {
    ActivityLauncher.showSignInForResult(this);
}
```

And, `ActivityLauncher.showSignInForResult(activity, true, false)` will become as straightforward as `ActivityLauncher.showSignInForResultWpComOnly(activity)`.

In the end it will become apparent that there is only those 2 exceptions and everything else flows through the main `ActivityLauncher.showSignInForResult(...)` method.

WordPress/build.gradle
```diff
@@ -81,6 +81,7 @@ android {
         buildConfigField "boolean", "BACKUP_DOWNLOAD_AVAILABLE", "false"
         buildConfigField "boolean", "SEEN_UNSEEN_WITH_COUNTER", "false"
         buildConfigField "boolean", "RESTORE_AVAILABLE", "false"
+        buildConfigField "boolean", "IS_JETPACK", "false"
```
ParaskP7: Suggestion (💡): Maybe adding the `APP` to make it `IS_JETPACK_APP` will make it more clear when using it later on, like `BuildConfig.IS_JETPACK` vs. `BuildConfig.IS_JETPACK_APP`. It seems to me that the first one can be interpreted in many ways, while the one with an explicit `APP` prefix will be much harder to misinterpret. Wdyt?

ParaskP7 (changes requested):
👋 @ashiagr !

I have left a warning (⚠️) comment, some questions (❓) comments and a few suggestions (💡) for you to consider.

WordPress/build.gradle
ParaskP7: Updated in deffc9ca6461c6fc39a5382c1d975fd38f1c0991 🙏

libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginBaseFormFragment.java
ParaskP7: Update in 8bcb107aae62a2143b2e9404dba240bcdc718f03 and a5947d1089513e4af3a4c29032594bf714ad5bb7 🙏

WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
ParaskP7: Updated in 1b8f8dd76a673739a5a17b175b1060b523f2ce2d 🙏

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
ParaskP7: Update in d0f597dc78b493c9ee128ceab32c622c1fbaa92f 🙏

libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginMode.java
ParaskP7: Updated in 42534581ad8f4a13ccfe804bf957b6df3753f7bd 🙏

ParaskP7 (approved):
👋 @ashiagr !

PS: Thanks for the updates! ❤️

ashiagr:
Thank you for all the inputs on this PR @ParaskP7, and also for reviewing linked PRs 🙏.
I'm merging this PR with target branch now.

PR #14361, opened 2021-03-27
List all likes in notifications

develric:
Part of the #14137 (companion Fluxc PR [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1935))

This PR shows all who liked a post or comment in a notification.

Hi @mattmiklic 👋 . The user profile bottom sheet is being added in another PR but happy if you could have a quick sanity check on the images below, thanks 🙇 ! 
A few thins to highlight:

1. I kept the header element directly in the list (so it moves up and down with the elements and does not stay sticky on top). I did this to keep the list consistent with what we have right now in the other lists of notifications details (like followers). 
2. There is an extra margin in the first liker element that is introduced by the header element. I kept it to be consistent with the other kind of lists for now.
3. About the followers list; currently the two lists (likers and followers) are rendered in a different way, probably we will need to align the followers list to this likers list, but I guess it's ok to release the likers feature first and align the followers later? wdyt?

| Description | Image |
|---|---|
| Dedicated fragment for Engaged People (to be reused for Followers as well in future) | <p align=center><image src="https://user-images.githubusercontent.com/47797566/112737547-a0f2e180-8f5b-11eb-8f6d-30b323420431.png" width=300><p> |
| A loading state has been introduced | <p align=center><image src="https://user-images.githubusercontent.com/47797566/112737664-6178c500-8f5c-11eb-8807-33b0b85e000e.png" width=300><p> |
| Empty state for no network | <p align=center><image src="https://user-images.githubusercontent.com/47797566/112737614-12cb2b00-8f5c-11eb-8fd3-6b37bd27aa36.png" width=300><p> |
| Empty state when errors in getting likes | <p align=center><image src="https://user-images.githubusercontent.com/47797566/112810621-7f891700-907b-11eb-814b-09f108bdeb79.png" width=300><p> |

**REVIEW NOTES**
1. User profile bottom sheet, unit tests and analytics will be added in dedicated PR(s)
2. Pagination will be aligned with the server side pagination of the endpoint when it is available; currently likes are taken from the endpoint up to the 90 limit that is coded in the endpoint sources
3. The order of likers in the list is currently different than what we get from notifications itself. An endpoint modification is ongoing to align this (cc @aerych).
4. About the confetti effect that is present in the `NotificationsDetailsListFragment`: in my understanding confetti celebration effect does not apply to `NOTE_LIKE_TYPE` and `NOTE_COMMENT_LIKE_TYPE` but only to `NOTE_VIEW_MILESTONE` so we should be fine to not include logic for that in these likers list (cc @planarvoid for a sanity check 🙇 )

## To test
The feature is behind a Feature Flag to which we are adding in next PRs. Good to compare the test below side by side with a develop installation.

- Open a notification from the ALL notification filter and check the notifications is loaded correctly
- Try to swipe in between the opened notification details and check the UI is rendered correctly in both expected notifications order and content.
- When you are swiping to a `LIKE` notification (post or comment) check you get the loading state and you get the likers list. Possibly check also list with lot of post or comment likers
- Do same checks above opening from the `LIKES` notification filter
- Smoke test trying to open notifications details from different type of notifications
- Check the empty states illustrated in the images in the beginning of the PR description
- From a post like details (and later also from a comment like details) check side by side that you get the same opened screens tapping on the 
    - header author avatar
    - header itself
    - an liker item in the list (for this point and tapping on the header elements would be good to force (by debugger or finding a suitable notification) to get an webview renderer the site or an app reader screen
- With feature flag off check that we are having same behaviour as before. Some things to check: notifications list 

PR submission checklist:


mattmiklic (changes requested):
Hey @develric I had some trouble testing the APK linked here so I pinged you about that. In the meantime, two changes to request on the empty states: 

- Let's remove the illustrations from the no network and error screens -- we are moving away from this illustration style, so I think it'd be best not to introduce it to more new places.
- The button style is still using the old android button style (all caps button text) instead of the new-style buttons we introduced in Login.
- Can we add punctuation to the error messages in the snackbar messages? Just an ending period on each error would make them feel more finished.

Speaking of the snackbar messages, do we need them at all here? It seems like they just duplicate the error message displayed in the middle of the screen.

develric:
Hey @mattmiklic 👋 

>had some trouble testing the APK linked here so I pinged you about that

My bad! I missed to share the steps to enable the feature flag if using the CI APK 🤦‍♂️! Shared some more detailed instructions on slack, but here the TL;DR: in `Me > App Settings > Test feature configuration` enable the `LikesEnhancementsFeatureConfig` flag. Sorry again 🙇 

Good points on the empty views 👍 . I should have covered them overall now and have a thought about snackbars; here below what we get:

| image | image | image |
|---|---|---|

As for snackbars, I was keeping it because of 2 reasons:
1. When we have an issue but we have likes data in cache we show the likes we have without an empty view, in that case the snackbar can maybe be good to have (see last right image above)
2. In cases different from No Network, there can be meaningful errors reported from the API, and I was using a generic empty state text like `An error occurred getting likes data` and the extra message error received from the server placed in the snackbar

Maybe we could:
1. No Network case: remove the snackbar when we are showing the empty state. I would keep the snackbar if we are not showing the empty state but listing the available cached data
2. Other errors: show the snackbar always (when we show empty state it will possibly give some more error details, when we show the cached data at least can trigger the user to know why not all the likes could be there eventually)

The current APK in this PR has the logic as described above, but of course let me know if you think it can work or have different advices and I will do my best to follow up 🙇

mattmiklic:
> No Network case: remove the snackbar when we are showing the empty state. I would keep the snackbar if we are not showing the empty state but listing the available cached data
> Other errors: show the snackbar always (when we show empty state it will possibly give some more error details, when we show the cached data at least can trigger the user to know why not all the likes could be there eventually)

Yeah, both of these sound good to me! 

The empty states without illustrations look good to me now. I will test the rest later this afternoon!

mattmiklic (commented):
Thanks @develric! Had a chance to test and things mostly look good. Can you confirm two things for me -- 

- the user's display name on each like is `body` (16dp), High Emphasis color
- the @username is `body 2` (14dp), Medium Emphasis color

It all looks pretty good just eyeballing it, but the username looks like it might be the high emphasis color.

planarvoid:
hey @develric 

> About the confetti effect that is present in the NotificationsDetailsListFragment: in my understanding confetti celebration effect does not apply to NOTE_LIKE_TYPE and NOTE_COMMENT_LIKE_TYPE but only to NOTE_VIEW_MILESTONE so we should be fine to not include logic for that in these likers list (cc @planarvoid for a sanity check 🙇 )

Yep, the confetti is here just for the view milestone notifications

develric:
>Yep, the confetti is here just for the view milestone notifications
 
Thanks for confirming @planarvoid 🙇 

>Thanks @develric! Had a chance to test and things mostly look good. Can you confirm two things for me --

Designer eyes are always designer eyes @mattmiklic sir 😄! Yep there was a visual glitch. I was using the `textAppearanceBody1` and `textAppearanceBody2` respectively thinking they were already containing the desired definitions color wise but it seems it's not the case. In more details the @username was not using the medium emphasis. I changed this as I think it should be changed style wise (eventually I'm sure @khaykov could suggest me if a better way exists during the code review) but the final result seems better at least to my eyes; looking forward to your comment/advice 🙇 

| Before | After |
|---|---|
| image | image |

mattmiklic (approved):
Yeah, that looks just right now. Thanks for humoring me. :) This is good to go!

WordPress/src/main/java/org/wordpress/android/ui/engagement/EngagedPeopleAdapter.kt
```diff
@@ -0,0 +1,49 @@
...
+            is Liker -> (holder as LikerViewHolder).bind(item)
+        }
+    }
+
+    override fun getItemCount(): Int {
+        return itemsList.size
+    }
+
+    override fun getItemViewType(position: Int): Int {
+        return itemsList[position].type.ordinal
```
khaykov, line 47: Pretty clever to use an `ordinal` for view type!

WordPress/src/main/java/org/wordpress/android/ui/engagement/LikedItemViewHolder.kt
```diff
@@ -0,0 +1,75 @@
...
+import org.wordpress.android.ui.engagement.AuthorName.AuthorNameString
+import org.wordpress.android.util.GravatarUtils
+import org.wordpress.android.util.getDrawableResIdFromAttribute
+import org.wordpress.android.util.image.ImageManager
+import org.wordpress.android.util.image.ImageType
+
+class LikedItemViewHolder(
+    parent: ViewGroup,
+    private val imageManager: ImageManager,
+    private val context: Context
```
khaykov: We can get a context from the view, so no need to pass one to view holder :)

WordPress/src/main/java/org/wordpress/android/ui/engagement/EngagedPeopleListFragment.kt
```diff
@@ -0,0 +1,213 @@
...
+        initObservers()
+
+        viewModel.start(listScenario!!)
+    }
+
+    private fun initObservers() {
+        viewModel.uiState.observe(viewLifecycleOwner, { state ->
+            if (!isAdded) return@observe
+
+            loadingView.visibility = if (state.showLoading) View.VISIBLE else View.GONE
```
khaykov: We also have a `updateVisibility`, alongside `setTextOrHide` in `uiHelpers` :) But setting visibility in either way is fine 👍

WordPress/src/main/java/org/wordpress/android/ui/engagement/ListScenarioUtils.kt
```diff
@@ -0,0 +1,93 @@
...
+        )
+        headerNoteBlock.setIsComment(note.isCommentType)
+
+        val spannable: Spannable = notificationsUtilsWrapper.getSpannableContentForRanges(headerNoteBlock.getHeader(0))
+        val spans = spannable.getSpans(
+                0, spannable.length,
+                NoteBlockClickableSpan::class.java
+        )
+        for (span in spans) {
+            span.enableColors(activity)
```
khaykov: I think we can just use a Context for this one.

WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderUtilsWrapper.kt
```diff
@@ -41,4 +41,8 @@ class ReaderUtilsWrapper @Inject constructor(
     fun isExternalFeed(blogId: Long, feedId: Long): Boolean = ReaderUtils.isExternalFeed(blogId, feedId)
 
     fun getReportPostUrl(blogUrl: String): String = ReaderUtils.getReportPostUrl(blogUrl)
+
+    fun postAndCommentExists(blogId: Long, postId: Long, commentId: Long): Boolean {
```
khaykov, line 45: Wonder if it makes sense to add `is` to the beginning  of the name here, since we return a Boolean?

WordPress/src/main/res/values/strings.xml
```diff
@@ -1465,6 +1465,11 @@
         <item>@string/notif_events</item>
     </string-array>
 
+    <!-- likes -->
+    <string name="get_likes_no_network_error">No network available.</string>
+    <string name="get_likes_unknown_error">An unknown error occurred getting likes.</string>
```
khaykov: Maybe we can add `while` (or `when`?) to the message, like this `An unknown error occurred while getting likes.`, to match the tone of similar error messages.

khaykov (commented):
Done with the first round! Great work on this one, @develric ! I just have a couple of small suggestions :)

WordPress/src/main/java/org/wordpress/android/ui/engagement/ListScenarioUtils.kt
develric: 👍 ! Just noting I initially tried to use the injected ContextProvider (that is an Application Context) but AFAIU I think we need a full activity/view context since we use `getColorFromAttribute` calls inside the `span.enableColors`

WordPress/src/main/java/org/wordpress/android/ui/reader/utils/ReaderUtilsWrapper.kt
develric: That is a good point 😄  ! I thought to keep the `postAndCommentExists` name since we are wrapping an already existing function. I would have no doubt in changing this if it was a new piece of code, but maybe I would stay with the original wrapped name here? Wdyt?

WordPress/src/main/res/values/strings.xml
develric: 👍 ! Definitely better 😄

develric:
hi @khaykov 👋 ! Thanks for the review and advices 🙇 ! I should have followed up to all (hopefully 😄 ). Let me know wdyt 🙇

khaykov:
Sorry, pushed the Flux-C tag to wrong branch. It is supposed to be here, so I guess it's not a problem :)

khaykov (approved):
Looks good, thanks for the changes! I confirmed that the new likes are secure behind the FF 🚀

PR #14369, opened 2021-03-30
Deprecate wordpress blue and replace it with blue

### LINKED ISSUE #14026: [Colors] Remove usages of WordPress Blue
In the [2.4.0 release of the Color Studio palette](https://github.com/Automattic/color-studio/blob/master/CHANGELOG.md), WordPress Blue was deprecated in favor of Blue. We should update any remaining items in the app that use WordPress Blue to now use Blue instead.

We should also take this opportunity to update our color palette resources to use the latest colors. The Color Studio repo contains a builder script that should automatically generate xml.

planarvoid:
Fixes #14026 

This PR replaces the `wordpress_blue` with the equivalent `blue` color based on the updated guidelines. This changes the `primary` color all over the app. 

To test:
- I can't see how this change could cause any problems or how exactly to test it. 

## Regression Notes
- none

- there are no areas of impact

- none (no reason to add automated tests)

PR submission checklist:


renanferrari (commented):
These changes look good to me but unless I'm missing something, it looks like the `wordpress_blue` colors are not being used anywhere now, so I wonder if it wouldn't be better to just remove them, to prevent it from being used by mistake?

planarvoid:
Thanks for the check @renanferrari, since we're copying the colors from the color-studio project, we're importing colors also not used in our app. I think I'd rather copy all the colors than manually remove parts of them when updating the palette. WDYT? I know @ParaskP7  is implementing some more complex changes to the colors so maybe this could be part of that project.

ParaskP7:
👋 @planarvoid @renanferrari !

> Thanks for the check @renanferrari, since we're copying the colors from the color-studio project, we're importing colors also not used in our app. I think I'd rather copy all the colors than manually remove parts of them when updating the paletter. WDYT? I know @ParaskP7 is implementing some more complex changes to the colors so maybe this could be part of that project.

Yeap, it seems indeed to be wrong having those unused colors hanging out in our repo, but that's mainly because those are not being perceived as a palette and are included within this one big `colors.xml` file. However, as @planarvoid mentioned, it would be wise for us to have all colors copied, no matter whether they are used or not, actually copying the whole palette file, getting it from the source `color-studio` repo, and keep that in sync with every studio update. This way, the manual updates of palette colors will be kept to a minimum, which will end up being to just copy-paste the whole palette file as is from the source `color-studio` repo to our repo. Later on, we might even consider automating this task with gradle, and even further on pushing that automation to CI so that it can run daily/weekly or something.

renanferrari (approved):
@planarvoid @ParaskP7 Thanks for the clarification! That makes sense.

LGTM 👍

PR #14373, opened 2021-03-30
Feature/unreplied comments filter

khaykov:
This PR adds an Unreplied comments filter:
[Image from Gyazo](https://gyazo.com/c12548882921f88bee35c16afcb73104)

[Here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1937) is the companion Flux-C PR.

Unreplied filter shows recent comments that are:
- Top level (not a reply to other comment)
- Not made by a logged in user.
- Do not have a reply from logged in user.
- Is Approved or Pending.

Unreplied filter has no paging, and only shows unreplied comments from 100 most recent comments.

To test:

Comment structure you are looking to test looks like this:
[Image from Gyazo](https://gyazo.com/0653937d52dd0c4b5d19f1afbc91164a)

- Navigate to Comments List and select an Unreplied filter.
- Make sure that only comments that meet following conditions are visible
  - Top Level (direct reply to post, and not to another comment).
  - Not made by logged in user.
  - Do not have a reply from logged in user.
  - Is Approved or Pending.
- Open an Unreplied comment and reply to it.
- Navigate back to Comments List, and confirm that the comment you replied too is no longer in Unreplied comments list.

PR submission checklist:


develric (commented):
hi @khaykov , did a first pass here and code looks good overall 👍 , also made some basic testing with .COM and self.hosted and didn't find issues; I'm going to make some more test tomorrow and setting up the scenarios you were describing in PR description, in case you want to consider them meanwhile, leaving couple of np.

1. The empty state for the UNREPLIED tab is the generic one, maybe we could use a specific one like in the other tabs 

<p align=center><image src="https://user-images.githubusercontent.com/47797566/113222050-8172f600-9286-11eb-8955-4231b743b258.png" width=250/></p>

2. Not sure if you planned to do it separately, maybe worth to add unit testing to CommentLeveler. wdyt?

khaykov:
Thanks, @develric ! Implemented your suggestions 👍

develric (commented):
Hey @khaykov 👋 ! Thanks for the changes. From the tests I've done it seems to work well for me 👍 ! Just wanted to check with you couple of things to be sure I'm not missing anything 🙇 

1. Tested with pure self hosted sites. Not sure if it's my setup but I was not able to get the account associated with an email so the comment leveler logic was not working as expected for me. What I tried was to set the email in Help and Support > Contact email but no luck. Note I was using a pure self hosted and not logged in in the app with any other WP.com account. Not sure if we can set the account email somewhere else in the app. Again, sorry in case it's obvious 🙃
2. This one is not relevant to the PR but more as a topic to face in one of our upcoming comments enhancements projects I think. In my scenario I noticed that whereas when commenting from the web page (also from mobile browser) a submitted comment was immediately visible on the threaded comments, when commenting from the app it wasn't (one time only I actually got it but after lot of time). Wondering if we should trigger some dispatcher from the API endpoint (cc @aerych )

I think we are actually fine to go but let me know wdyt (mostly about point 1) 🙇

khaykov:
Good catch! I added a separate logic for self-hosted sites - we will try to get an email from Site object (we add it there by requesting a user profile).

develric (approved):
Yep 👍! Was looking into the last pushes and tested it already. I think it works really well, thanks @khaykov 🙇 ! :shipit: 
Feel free to merge it after fluxc is tagged 👍

Issue #14138, opened 2021-02-23
Likes Enhancements in Reader


develric:
We would like to enhance the current visualization of like-ers in reader as per below flow image.

![image](https://user-images.githubusercontent.com/47797566/108851328-9784fb00-75e4-11eb-894c-20b788b86ae0.png)

Currently this issue is intended to be tackled after #14137 so we could reuse the list of like-ers improvements (since basically it's the same screen)

A few tasks as per current understanding of the ticket:

- [x] Create feature flag (may make sense to use same as #14137) [#14361]
- [x] Add the like-ers faces row in reader detail view [#14375]
- [x] Reuse the list of like-ers, User Profile bottom sheet and site feed from #14137 [#14375]
- [x] Add analytics [#14525]
- [x] Add Unit testing [#14532, #14541]
- [x] Remove feature flag and release [#14563]

PR #14375, opened 2021-03-31
Adding liker faces train in reader and allow to access the list of likers

develric:
Part of the #14138. This PR shows liker faces (up to 5) in the Reader Post Details and allow to open the list of likers already used in the Notifications section (see #14361).

| image | image |
|---|---|

Hey @mattmiklic 👋. I'm still looking a couple of improvements code-wise but this should be ready for a design review (as for the previous one please note the `LikesEnhancementsFeatureConfig`  feature flag must be activated using the CI built apk 🙇 ). Want to highlight a deviation from original design spec. Unless I'm doing something wrong, it seems to me I'm obtaining a result more similar to spec using a spacing of -4dp (the apk in here is using this spacing) instead of -2dp (as per spec). Below the result with -2dp (top image) and -4dp (bottom image)

<p align=center><image src="https://user-images.githubusercontent.com/47797566/113186945-8d47c380-9258-11eb-97ca-34669ebc7203.png" width=300/></p>

## To test

The feature is behind a Feature Flag to which we are adding in next PRs. Good to compare the test below side by side with a develop installation.

- Try to open a post in the reader and check the like faces are loaded (notice the initial loading state)
- Check number of likers are consistent with the number next to the like star
- Check you can tap on the faces row to get the list of likers (this should be consistent with the list in the previous notifications PR)
- Try to like/unlike and check you get the number updated
- Try to make the above checks both with posts with lot of likes and under 5 likes or no likes
- Use a super short post to be able to check the loading phases of like  section without scrolling the post webview
- Check screen rotation states
- When having issues loading likers data (for example setting air mode) check that you have:   
    - if no likers data were loaded before -> you get an empty state error string
    - if some likers data were loaded before -> you get the rendered cached data and a snackbar with the issue message
- Check dark mode is consistent
- With feature flag off, smoke test to check that we are having same behaviour as before.

## Regression Notes
This feature is still behind a feature flag; regression notes will be highlighted as the feature parts are shipped to user. Also unit testing is being addressed in a dedicated PR.

PR submission checklist:


mattmiklic (changes requested):
@develric This is looking really good. One change to request; let's go from 1dp to 2dp for the white border on each avatar. That might or might not affect the offset you want to use between them, but what you have here (-4dp) looks good right now.

develric:
Hi @mattmiklic , thanks for the follow up. As we discussed a bit on slack here below the versions with 2dp white border and -4dp spacing (on the left) and -6dp spacing (on the right). If I got it right we both seems to prefer the -6dp, isn't it? If that's correct, I will push the code changes accordingly 🙇 

| -4dp spacing | -6dp spacing |
|---|---|
| image | image |

mattmiklic (approved):
Everything looks perfect now; thanks!

tw875768 (approved):

khaykov:
Sorry, @develric , I didn't get a chance to look at this one today. Will prioritize it for tomorrow!

WordPress/src/main/java/org/wordpress/android/ui/engagement/EngagedPeopleListActivity.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.wordpress.android.R
+import org.wordpress.android.ui.LocaleAwareActivity
+
+class EngagedPeopleListActivity : LocaleAwareActivity() {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        setContentView(R.layout.engaged_people_list_activity)
+
+        val listScenario = intent.getParcelableExtra<ListScenario>(KEY_LIST_SCENARIO)
```
khaykov, line 15: Do we want to check if `listScenario` is null here, and do something if it's not there (eg. throw an exception)?

WordPress/src/main/java/org/wordpress/android/ui/reader/viewholders/PostLikerItemDecorator.kt
```diff
@@ -0,0 +1,22 @@
...
+
+class PostLikerItemDecorator(context: Context, @DimenRes leftOffsetResId: Int) : ItemDecoration() {
+    private val leftOffset: Int = context.resources.getDimensionPixelSize(leftOffsetResId)
+
+    override fun getItemOffsets(outRect: Rect, view: View, parent: RecyclerView, state: State) {
+        super.getItemOffsets(outRect, view, parent, state)
+        val position = parent.getChildAdapterPosition(view)
+
+        if (position > 0) {
+            outRect.set((outRect.left - leftOffset), outRect.top, outRect.right, outRect.bottom)
```
khaykov, line 19: I think we need to add a case for for RTL, or it will look like this:
[Image from Gyazo](https://gyazo.com/1adf2b79a9e380e0d15197072db87560)

WordPress/src/main/res/values/strings.xml
```diff
@@ -1473,6 +1473,11 @@
     <string name="get_likes_no_network_error">No network available.</string>
     <string name="get_likes_unknown_error">An unknown error occurred while getting likes.</string>
     <string name="get_likes_empty_state_title">An error occurred while getting likes data</string>
+    <string name="like_title_single_text">1 Like</string>
```
khaykov, line 1476: While the meaning is pretty clear, I think we could change the string name to `like_title_singular`, so it will be easier for translators to get the idea :)

WordPress/src/main/res/values/strings.xml
```diff
@@ -1473,6 +1473,11 @@
...
+    <string name="like_title_multiple_text">%d Likes</string>
```
khaykov, line 1477: Similarly, here we could change it to `like_title_plural` (and same for other strings). 

Also, It might be an overkill just for 2 strings, but Android supports plurals natively, so you can have something like this in the strings.xml:

```
<plurals name="like_title">
        <item quantity="one">%d Like</item>
        <item quantity="other">%d Likes</item>
</plurals>
```

And retrieve them with:
`getResources().getQuantityString(R.plurals.like_title, numLikes, numLikes);`

WordPress/src/main/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostDetailViewModel.kt
```diff
@@ -145,13 +180,40 @@ class ReaderPostDetailViewModel @Inject constructor(
...
+        if (likesEnhancementsFeatureConfig.isEnabled()) {
+            _updateLikesState.addSource(getLikesHandler.likesStatusUpdate) { state ->
+                _updateLikesState.value = state
+            }
+        }
+    }
+
+    fun onRefreshLikersData(post: ReaderPost) {
+        if (!likesEnhancementsFeatureConfig.isEnabled()) return
+        val isLikeDataChanged = lastRenderedLikesData?.let {
```
khaykov, line 206: What do you think about adding a method to `RenderedLikesData` that will do this comparison? Something like `isSameAsInPost()` or `isMatchingThePost()`, or something like this. This way we can test it too :)

khaykov (approved):
Looks and works great, @develric ! I left couple of comments in code.

In addition, I noticed that the Likers are not aligned with post content in landscape mode:
[Image from Gyazo](https://gyazo.com/190c3eee68ffb5359065c34b0a9a3d13)

Also, when opening likes activity from the post, you can tap on the first row to open the same post again (and again, etc.). What do you think about closing the likes activity if the post in the first row is the same as the one we navigated from originally? This adds some extra logic, so I'm not sure it's worth it for this pretty edge case.

develric:
Hey @khaykov 👋 ! Thanks for the review and advices 🙇 ! I think you already checked that with the feature flag off, the behaviour in the reader details is correct (that is not train of faces and every element is in it's location with the expected content). If it is the case, I would merge this one (since we are not removing the feature flag on this PR) and address your suggestions in a dedicated PR, so that it is more manageable hopefully. Let me know wdyt, thanks sir 🙇 !

khaykov (approved):
Sure, sounds like a good plan. I made sure the feature flag is working as intended and going to merge this one 👍

PR #14381, opened 2021-03-31
Limit Classic editor access on Simple WP.com sites to existing Classic posts

cameronvoell:
The next step after [notifying users that the classic editor is going away](https://github.com/wordpress-mobile/gutenberg-mobile/issues/3209) is to actually remove it, starting with Simple sites.

I think this change doesn't warrant inclusion in RELEASE-NOTES.txt [since we already notified users about this change in 16.9](https://github.com/wordpress-mobile/WordPress-iOS/pull/16008), but if anyone thinks differently I'd love to hear. Perhaps we could include it in release notes but exclude it from the Play Store version notes.

Addresses https://github.com/wordpress-mobile/gutenberg-mobile/issues/3092 (for WPAndroid)

## To test

_Note: The toggle switch referred to below is the "Use block editor" toggle switch found in My Site → Settings. This is only available to site admins._

1. With APK from this PR, log into a WP.com account
2. Choose a Simple site where you are an admin
3. Verify that there **is no** toggle switch
4. Using the current Play Store version of the app – not the app built from this branch – log into the same WP.com account
5. Choose the same Simple site where you are an admin, locate the toggle switch and turn it off (meaning use the classic editor) 
6. Moving back over to the app built from this branch – locate the same Simple site
7. Create a new post and ensure that the block editor is shown (not the classic editor)

## Regression Notes

This PR removes the classic editor on Simple sites. However, we're leaving it in place for scenarios where the block editor cannot be used instead. These scenarios include editing a classic post (i.e. a post that contains only classic content). We have to check to make sure these scenarios still work on Simple sites. We should also check to make sure all other site types (Atomic/Jetpack/self-hosted) are unaffected by this change.

- [x] Test Simple WP.com sites do not show a toggle switch (used camiams.home.blog)
- [x] Test Simple WP.com sites defaulted on mobile to Classic use Gutenberg for new posts (used camiams.home.blog)
- [x] Regression test Atomic sites show a toggle switch (https://americanswimming9.wpcomstaging.com/)
- [x] Regression test Jetpack-connected self-hosted sites show a toggle switch (used coderovers.io)
- [x] Regression test self-hosted sites show a toggle switch (used codrovers.io)
- [x] Regression test Simple WP.com site users open existing Classic posts in the Classic editor (camiams.home.blog)

We are disabling the Classic Editor tests until they can be migrated to executing on existing classic posts, instead of creating new ones. That work is tracked with this ticket here: https://github.com/wordpress-mobile/WordPress-Android/issues/14389

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
+++ b/WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
@@ -210,7 +210,8 @@ public class SiteUtils {
             // Default to block editor when mobile editor setting is empty
             return true;
         } else {
-            return site.getMobileEditor().equals(SiteUtils.GB_EDITOR_NAME);
+            boolean isSimpleWPCom = site.isWPCom() && !site.isWPComAtomic();
+            return isSimpleWPCom || site.getMobileEditor().equals(SiteUtils.GB_EDITOR_NAME);
         }
     }
 
--- a/WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
+++ b/WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
@@ -210,9 +210,12 @@ public class SiteUtils {
             // Default to block editor when mobile editor setting is empty
             return true;
         } else {
-            boolean isSimpleWPCom = site.isWPCom() && !site.isWPComAtomic();
-            return isSimpleWPCom || site.getMobileEditor().equals(SiteUtils.GB_EDITOR_NAME);
+            return alwaysDefaultToGutenberg(site) || site.getMobileEditor().equals(SiteUtils.GB_EDITOR_NAME);
         }
+    }
+
+    public static boolean alwaysDefaultToGutenberg(SiteModel site) {
+        return site.isWPCom() && !site.isWPComAtomic();
     }
 
     public static String getSiteNameOrHomeURL(SiteModel site) {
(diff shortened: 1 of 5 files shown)


guarani (approved):
Tested on Simple sites using the test steps above and this looks good! 
The code looks good from my point of view, but I'm not an Android dev 😄 

nit pick: there's a mention of App Store above that probably refers to the Play Store

cameronvoell:
@mchowning An additional review on this would be welcome since after Paul approved, I added the plan we discussed for disabling classic tests until they are migrated to executing on an existing classic post. 🙏

WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
```diff
@@ -210,7 +210,8 @@ public static boolean isBlockEditorDefaultForNewPost(SiteModel site) {
             // Default to block editor when mobile editor setting is empty
             return true;
         } else {
-            return site.getMobileEditor().equals(SiteUtils.GB_EDITOR_NAME);
+            boolean isSimpleWPCom = site.isWPCom() && !site.isWPComAtomic();
```
mchowning: Not a blocking issue, but what would you think about instead extracting out something along these lines:

```
public static boolean alwaysDefaultToGutenberg(SiteModel site) {
  return site.isWPCom() && !site.isWPComAtomic();
}
```

and then using that method both here and at `SiteSettingsFragment::1005`? 

My thinking is that we always want those two checks to be the same (anytime the editor toggle is unavailable we want to also want to force the default editor to be Gutenberg), and having the logic in a shared method makes it more likely they'll stay in sync.

mchowning:
This all looks good to me. 👍  I had one comment, but it's not a blocker.

WordPress/src/main/java/org/wordpress/android/util/SiteUtils.java
cameronvoell: Nice suggestion, added here: 8d7b3dd28a78751f7698ca2dbb031df31ae2c295

mchowning (approved):

PR #9957, opened 2019-05-29
Changes related to deep linking email marketing

maxme:
Fixes #9864

To test:
- Should not open the app: `https://wordpress.com/post/`
- Should not open the app: `https://public-api.wordpress.com/nope/`
- Should not open the app: `https://public-api.wordpress.com/bar/?redirect_to=https%3A%2F%2FCHANGEME.wordpress.com%2Fpost`
- Should open the app's editor on previously selected site: `https://wordpress.com/post/wrong-hostname`
- Should open the app's editor with selected site CHANGEME: `https://wordpress.com/post/CHANGEME.wordpress.com`
- Should open the app's editor with selected site CHANGEME: `https://public-api.wordpress.com/mbar/?redirect_to=https%3A%2F%2Fwordpress.com%2Fpost%2FCHANGEME.wordpress.com`
- Should "open" the app, but then redirect to the browser: `https://public-api.wordpress.com/mbar/?redirect_to=https%3A%2F%2Fwordpress.com%2Fnope`

<s>Note: this is a draft PR because backend isn't updated yet and the deep linking urls might changes (`mobile-api`)</s>

This can be tested with ADB:

```shell
$ adb shell am start -a android.intent.action.VIEW \
 -c android.intent.category.BROWSABLE \
 -d "https://public-api.wordpress.com/mbar/?redirect_to=https%3A%2F%2Fwordpress.com%2Fpost%2FCHANGEME.wordpress.com"
```

Real test:
- Create a new site.
- Publish a post.
- Switch to another site (to make sure the editor is opened for the right site later).
- Wait for the "Congrats for your first post" marketing email.
- Tap the "Create new post" button, you should be able to choose between your browser or the WordPress app, select the WordPress app, the editor should open on this site.

[Demo video](https://bia.is/s/lhOY/mbar-email-demo.webm) of this ^

cc @diegoreymendez

PR #14387, opened 2021-04-01
Add "create a site" deeplink handling

### LINKED ISSUE #14371: Handle "Create a site" deeplink
When a user has an account without a site, they're sent an email that links to the "Create site" flow. These links are now handled in the browser. We want to add a deeplink handling to the app so that these links now link to the in-app "Create site" flow.

planarvoid:
Fixes #14371 

This solution introduces handling for the `public-api.wordpress.com/mbar/...redirect_to=wordpress.com/wp-login.php?...redirect_to="wordpress.com/start"` links from emails and opens a site creation flow when these links are clicked.
There is now a view model which handles all the `mbar` links and all the links that open the editor (just to have them in one place). 

To test - Site creation:
- Log in in the app with a WP.com account
- Click on the "Create your site" in an email (let me know if you need the link from the email)
- Notice this line of code in the LogCat - "DeepLink tracking URI successfully requested"
- Notice the "Site creation" flow is started
- Finish the flow and create a site
- Notice you end up on the My Site screen with the site is selected 

To test 2 - Site creation not logged in
- Do not log in to the app with a WP.com account
- Click on the "Create your site" in an email (let me know if you need the link from the email)
- Notice this line of code in the LogCat - "DeepLink tracking URI successfully requested"
- Notice the "WP.com signup" flow is started
- Log in/signup
- You end up logged in on the My Site screen

To test 3 - edit an existing post
- Log in to the app
- Click on a link like `https://wordpress.com/post/site_url/post_id` that exists for a site on an account you're logged in with 
- Notice this line of code in the LogCat - "DeepLink tracking URI successfully requested"
- Notice the post is opened in the editor

To test 4 - edit a non-existing post
- Log in to the app
- Click on a link like `https://wordpress.com/post/site_url/post_id` that doesn't exists for a site on an account you're logged in with 
- Notice this line of code in the LogCat - "DeepLink tracking URI successfully requested"
- Notice the editor is opened for site with site_url

To test 5 - edit a non-existing post on a non-logged in site
- Log in to the app
- Click on a link like `https://wordpress.com/post/site_url/post_id` that doesn't exists for a site that you're not logged in
- Notice this line of code in the LogCat - "DeepLink tracking URI successfully requested"
- Notice the editor is opened for your current site

## Regression Notes
- All the `mbar` links are impacted but they should all be handled as before

- Unit tested the classes that could be tested

- unit tests in this PR

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/DeeplinkWebViewActivity.kt
```diff
@@ -0,0 +1,52 @@
...
+import android.app.Activity
+import android.content.Intent
+import android.os.Bundle
+import android.webkit.WebResourceRequest
+import android.webkit.WebView
+import android.webkit.WebViewClient
+import org.wordpress.android.R
+import org.wordpress.android.ui.accounts.LoginMagicLinkInterceptActivity
+
+class DeeplinkWebViewActivity : WebViewActivity() {
```
renanferrari: If we decide to handle `mbar` links instead of `bar` ones, as we have been discussing recently, then I think we should be able to get rid of this class, right?

WordPress/src/main/java/org/wordpress/android/ui/DeepLinkingIntentReceiverActivity.java
```diff
@@ -91,14 +96,29 @@ protected void onCreate(Bundle savedInstanceState) {
...
-                // not handled
-                finish();
+                UriWrapper uriWrapper = new UriWrapper(uri);
+                // https://public-api.wordpress.com/bar/
+                if (mEmailUriHandler.shouldHandleUri(uriWrapper)) {
+                    mEmailUriHandler.handleUri(uriWrapper);
+                } else {
+                    // not handled
+                    finish();
+                }
```
renanferrari: Again, if we decide to handle `mbar` links, I believe we should be able to merge this logic with the existing logic in `handleOpenEditorFromTrackingUrl` as they seem to do the similar things, right? Let me know what you think.

WordPress/src/main/java/org/wordpress/android/ui/DeeplinkNavigator.kt
```diff
@@ -0,0 +1,32 @@
...
+        when (navigateAction) {
+            is OpenInWebView -> {
+                DeeplinkWebViewActivity.openUrl(activity, navigateAction.uri.toString())
+            }
+            is StartCreateSiteFlow -> {
+                if (navigateAction.isSignedToWpCom) {
+                    ActivityLauncher.newBlogForResult(activity)
+                } else {
+                    ActivityLauncher.addSelfHostedSiteForResult(activity)
+                }
```
renanferrari: Two thoughts here:

- If the user is not logged in to WP.com, I think it would probably be better for them to be redirected to the WP.com login flow instead of the add site flow.
- In addition to that, we should also check if the user is not logged in at all, in which case I think they should be redirected to the full login flow.

Let me know what you think!

renanferrari (commented):
Thanks for submitting these changes!

I tested both the create site URL and the magic link URL and everything seems to be working as expected.

Code is also looking pretty good and it seems like a solid starting point for us! Left a few thoughts and questions below.

WordPress/src/main/java/org/wordpress/android/ui/DeeplinkNavigator.kt
planarvoid: hmm, good point but I think we should only redirect to the WP.com flow because the link in the email only comes to WP.com users and I don't see a usecase where we'd like to redirect a user who got the "Create a site" email to the self-hosted flow. WDYT?

planarvoid:
thanks for the review @renanferrari , I've completely rewritten the whole solution to handle the `mbar` links, I've removed the WebView, tested the code, moved handling of all the `mbar` links and the editor links to the ViewModel, and overall improved the solution. I think this is ready now for another round.

WordPress/src/main/java/org/wordpress/android/ui/DeepLinkNavigator.kt
```diff
@@ -0,0 +1,47 @@
...
+                    navigateAction.site,
+                    navigateAction.postId
+            )
+        }
+        activity.finish()
+    }
+
+    sealed class NavigateAction {
+        data class OpenInBrowser(val uri: UriWrapper) : NavigateAction()
+        data class OpenInEditor(val site: SiteModel, val postId: Int) : NavigateAction()
```
renanferrari: It's a nitpick, but maybe we could rename `OpenInEditor` to `OpenEditorForPost` for consistency with `ActivityLauncher::openEditorForPostInNewStack` and `OpenEditorForSite`?

WordPress/src/main/java/org/wordpress/android/ui/DeepLinkingIntentReceiverViewModel.kt
```diff
@@ -0,0 +1,124 @@
...
+    @Named(UI_THREAD) private val uiDispatcher: CoroutineDispatcher
+) : ScopedViewModel(uiDispatcher) {
+    private val _navigateAction = MutableLiveData<Event<NavigateAction>>()
+    val navigateAction = _navigateAction as LiveData<Event<NavigateAction>>
+    val toast = editorLinkHandler.toast
+
+    /**
+     * URIs like `public-api.wordpress.com/mbar/...` come from emails and should be handled here
+     */
+    fun shouldHandleEmailUrl(uri: UriWrapper): Boolean {
```
renanferrari: These links currently come from emails, but in the future they might come from somewhere else. Wouldn't be better to keep the `shouldHandleTrackingUrl` name (as it was before)? Then we could also update the comment to indicate that this _usually_ comes from emails. Same thing for `handleEmailUrl` → `handleTrackingUrl`.

WordPress/src/main/java/org/wordpress/android/util/UriWrapper.kt
```diff
@@ -1,15 +1,27 @@
...
 data class UriWrapper(val uri: Uri) {
     constructor(uriString: String) : this(Uri.parse(uriString))
+
     val pathSegments: List<String> = uri.pathSegments
     val host: String? = uri.host
 
     override fun toString() = uri.toString()
     fun getQueryParameter(key: String): String? = uri.getQueryParameter(key)
+    fun copy(path: String): UriWrapper {
+        val newUri = Builder()
```
renanferrari: I was almost sure there was a method to easily copy `Uri`s, so I just looked it up and this is the one: `Uri::buildUpon`. Have you considered using that? It seems like it would be safer.

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
```diff
@@ -1171,6 +1171,14 @@ public static void showMainActivityAndSiteCreationActivity(Activity activity) {
         activity.startActivity(intent);
     }
 
+    public static void showMainActivityAndSignUp(Activity activity) {
+        final Intent intent = new Intent(activity, WPMainActivity.class);
+        intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
+        intent.addFlags(Intent.FLAG_ACTIVITY_NO_ANIMATION);
+        intent.putExtra(WPMainActivity.ARG_WP_COM_SIGN_UP, true);
+        activity.startActivity(intent);
+    }
```
renanferrari: I'm not sure if we need to add this. Calling `ActivityLauncher::showSignInForResult` with `isWpComOnly = true` should be enough in this case. Let me know what you think.

renanferrari (commented):
Again, thank you for the changes!

I've tested the app with the scenarios you provided and also the ones that were listed in #9957 and everything worked as expected.

Code is also looking pretty good. I just left a few comments below.

WordPress/src/main/java/org/wordpress/android/ui/DeepLinkNavigator.kt
planarvoid: makes sense, changed in f32d99fcb4

WordPress/src/main/java/org/wordpress/android/ui/DeepLinkingIntentReceiverViewModel.kt
planarvoid: tracking URL is a better name than email URL, changed in 0730a8218b

WordPress/src/main/java/org/wordpress/android/util/UriWrapper.kt
planarvoid: good catch! The previous approach was something used before, I've just copy&pasted it but I didn't know a method `buildUpon` exists. Changing it in in 2a83158eed

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
planarvoid: I think you're right. I see that the signup flow triggers the WPMain activity itself, it doesn't just rely on the result. Why I haven't used it before is that the login activity is triggered by `startActivityForResult` and I expected that this is started from the Main activity and that the result is necessary (as it is for all the other deeplink flows). I've changed it in a1d4d99b4b

renanferrari (approved):
Thanks for the changes again!

LGTM 👍

PR #14401, opened 2021-04-05
Fixes small errors in release toolkit

JavonDavis:
This PR does two things
1. Fixes an incorrectly named variable in the FastFile 
2. Updates the release-toolkit tag to the version with the regex change needed to find the gutenberg-mobile versin

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -469,7 +469,7 @@ ENV["validate_translations"]="lintVanillaRelease"
     {library: "Stories Creator", strings_path: "./libs/stories-android/stories/src/main/res/values/strings.xml", exclusions: []}
   ]
 
-  BINARY_IMPORTED_LIBRARIES = [
+  binary_imported_libraries = [
     {
       name: "Gutenberg Native",
       import_key: "gutenbergMobileVersion",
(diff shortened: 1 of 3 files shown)


oguzkocer (approved):

fastlane/Fastfile
```diff
@@ -469,7 +469,7 @@ ENV["validate_translations"]="lintVanillaRelease"
     {library: "Stories Creator", strings_path: "./libs/stories-android/stories/src/main/res/values/strings.xml", exclusions: []}
   ]
 
-  BINARY_IMPORTED_LIBRARIES = [
+  binary_imported_libraries = [
```
mokagio, line 472: Nice catch!

No big deal, but I would suggest keeping this as all uppercase, because that's how [constants are defined in Ruby](https://www.rubyguides.com/2017/07/ruby-constants/).

JavonDavis: Ahhh you're right 🤦‍♂️ overlooked that detail! I'll get this change in with another PR

mokagio: Yeah no worries. I kept confusing them till it was pointed out to me. It's a curious design choice, imho. Also, it's a minor optimization in the context of our `Fastfile`. No stress 😄

PR #14417, opened 2021-04-08
Feature/update looks of comment rows

khaykov:
This PR updates the look of Comment List items:
- Removes status label.
- Removes date from individual comments
- Adds a Date sub-header.
- Adds a yellow dot to indicate pending comments.
- Added a divider with offset between comments.

[Image from Gyazo](https://gyazo.com/077f57e1b04a336468e4c6592fd09dfc)

[Image from Gyazo](https://gyazo.com/0edff239c348f0bb47cfb2c9a2c371a4)

[Image from Gyazo](https://gyazo.com/200f19e6fb47618ef029fc3c9f5bf25b)

Behind the scenes I expanded view holders and moved a lot of logic into them, added some abstraction, implemented `DiffUtils` when loading comments.

Original plan was to move all the comment loading and management logic from adapter, but the amount of changes would be very large, so I decided to leave that for another time.

To test:
- Confirm that pending dot appears for pending comments.
- Confirm that moderation of individual comments works (and it's correctly
- Confirm that batch moderation works.

## Regression Notes
Comment moderation.

Manual testing.

With current implementation automated tests are pretty difficult.

PR submission checklist:


mattmiklic (changes requested):
This is looking really good, @khaykov. Just two requests, if possible:

It looks like previously the avatars were 18dp from the left edge of the screen, now they're 16dp. While this is technically more correct (and what I'd spec'd in the design), I think if we moved them back over to 18dp from the edge, it would give us more space around the pending indicator dot. And the avatars would optically align better with the back arrow and "All" tab in the navigation bar, since they don't render exactly on the 16dp keyline, but slightly to the right of it. This is what it would look like (a little less cramped than currently):

adjust-spacing

Second, could we move up the text in each row by 2dp (but not move up the avatar)? That would make the author/title better vertically aligned with the avatar. It would also reduce the amount of space between the date overlines and the text below them, which feels a little too open currently.

khaykov:
@mattmiklic Thanks for the pointers! I checked the original layout, and avatar was `20dp` from the edge of the screen. With this number it looks pretty close to your mock :) I also had to reduce the margin between text and avatar so the main content will remain `72dp` from the edge.

I removed the extra font padding from the title, so now it's perfectly aligned with the avatar. There is still a bit of a space to accommodate tall letters, like Ö.

I also removed an extra margin between sub-header and comment - divider was not drawn there, but it still took space.

[Image from Gyazo](https://gyazo.com/06939f5c2c22b405a76b12d216f57245)

[Image from Gyazo](https://gyazo.com/02ebfc2f3dec99258d4e42a142996c23)

mattmiklic (approved):
Everything looks really good here now. Thanks for these adjustments!

WordPress/src/main/java/org/wordpress/android/ui/comments/CommentAdapter.java
```diff
@@ -296,14 +153,31 @@ public CommentModel getItem(int position) {
...
     public long getItemId(int position) {
-        return mComments.get(position).getRemoteCommentId();
+        CommentListItem listItem = getItem(position);
+
+        if (listItem instanceof SubHeader) {
+            return listItem.longId();
+        } else if (listItem instanceof Comment) {
+            return ((Comment) listItem).getComment().getRemoteCommentId();
+        }
+        return -1;
```
develric, line 163: Not fully sure about the following, but since I happened to have the doubt, sharing below 😄 . If it's a dummy point please just ignore it and do not judge me too hard 😜.

I see SubHeader and Comment are different domains so not immediate to create a unique Id (as I think it's required since we are declaring has stable ids as true). I also see that the position of an element can change in the dataset so we should not use the default implementation that returns position here. This said I'm not super sure that the longId for a SubHeader could not collide with a RemoteCommentId 🤔 ; wonder if we could separate domains like (just an example 🙃 ):

- RemoteCommentId -> long >= 0 (and maybe this is a counter on a site and not site/post(?); so on high traffic sites could be pretty high number)
- NO_ID -> -1
- SubHeader -> long < 10 (like longId * -10 or given a hashcode could be negative in principle, using the timestamp * -10 🤔 )

As said not sure how much the above is near to a real case consideration, so let me know your thoughts 🙇

develric: Not fully sure about the following, but since I happened to have the doubt, sharing below 😄 . If it's a dummy point please just ignore it and do not judge me too hard 😜.

I see SubHeader and Comment are different domains so not immediate to create a unique Id (as I think it's required since we are declaring has stable ids as true). I also see that the position of an element can change in the dataset so we should not use the default implementation that returns position here. This said I'm not super sure that the longId for a SubHeader could not collide with a RemoteCommentId 🤔 ; wonder if we could separate domains like (just an example 🙃 ):

RemoteCommentId -> long >= 0 (and maybe this is a counter on a site and not site/post(?); so on high traffic sites could be pretty high number)
NO_ID -> -1
SubHeader -> long < 10 (like longId * -10 or given a hashcode could be negative in principle, using the timestamp * -10 🤔 )
As said not sure how much the above is near to a real case consideration, so let me know your thoughts 🙇

develric (commented):
Hi Klym 👋 ! Did a first pass here and it looks pretty good 👍 ! Code LGTM and the functionality also is working well (checked pending dot, single and batch moderation, also changing state from inside the comment detail and coming back to lists. RTL support works as expected as well). Tested on a Pixel 3 Emu API 29 comparing to an equivalent emulator side by side running the previous version of the app. I happened to think to something related to the adapter getItemId that I'm sharing for your thoughts. Not fully sure it is relevant so let's check with you 🙇 .

WordPress/src/main/java/org/wordpress/android/ui/comments/CommentAdapter.java
khaykov: Thanks, @develric ! This is a good point, and can be a concern for sites with a lot of comments. I think making the SubHeader ID negative should solve this :)

khaykov: Update: As we discussed, I went with `timestamp * -10` for subheader ID :)

develric (approved):
Hey @khaykov ! Everything looks good to me, thanks for the changes 🙇  ! :shipit:

PR #14424, opened 2021-04-08
Fix initial login prologue screen

ashiagr:
Fixes initial login prologue screen issue introduced in [this PR](https://github.com/wordpress-mobile/WordPress-Android/pull/14356). 

To test:

Test 1:
1. Fresh install the app/ Clear storage for any WordPress app flavor.
2. Launch the app.
3. Notice that the login initial prologue screen is displayed.

## Regression Notes
N/A

N/A

N/A
PR submission checklist:


ParaskP7 (approved):
👋 @ashiagr !

Wow, I really missed that while reviewing #14356, apologies! 🙏

PS: That's why having tests save the day, and vice-versa, not having tests, we solely depend on our error prone code review... 🤕 And in this case, even if we had this login on the `ViewModel` level we would still trigger the `ActivityLauncher` on the `Activity` level, which wouldn't be tested as well. That's why having tests for navigation and having a `Navigator` pattern has its benefits. We should start thinking about that since our activity launchers are very complicated and untested.

PR #14431, opened 2021-04-08
Reverted materialVersion back to 1.2.1

khaykov:
While working in TimeZone selector we bumped material library version from 1.2.1 to 1.3.0, and looks like this might have caused some issues (potentially related to new version of `ConstraintLayout` that 1.3.0 is using).

I only spotted the issue in reader, but it might appear somewhere else, so until we have more info on how to migrate to 1.3.0 I feel it's better to revert to 1.2.1

To test:
- Open Reader.
- Confirm that posts cards without Feature Images looks ok.

Post Cards should look like this:
[Image from Gyazo](https://gyazo.com/f39651beb9d1885a1dd2df5e51e1b7d0)

And not like this:
[Image from Gyazo](https://gyazo.com/ebbbd4085598dfae2a6e8a3f77260ca6)

PR submission checklist:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -140,7 +140,7 @@ ext {
     coreVersion = '1.3.2'
     lifecycleVersion = '2.2.0'
     constraintLayoutVersion = '1.1.3'
-    materialVersion = '1.3.0'
+    materialVersion = '1.2.1'
     preferenceVersion = '1.1.0'
     swipeToRefresh = '1.1.0'
     uCropVersion = '2.2.4'


ParaskP7 (approved):
👋 @khaykov !

> While working in TimeZone selector we bumped material library version from 1.2.1 to 1.3.0.

Was bumping the material library a requirement for this work or was that an extra nice-to-have improvement? I am asking that because if that was a requirement I am not sure if I should be looking at a specific screen/feature to verify correctness with this revert. I did check the commit (2b085a61e9aba504eb3dcc769be25cbbf95b7fe8) that pumped the material library to `1.3.0`, but I am still not 100% sure. Also, I am not sure whether bumping the core ktx is somehow related to that.

PS: Let me know your thoughts on the above. In the meanwhile, I will not merge this PR just yet.

khaykov:
Hey @ParaskP7 ! Thank you for the review! Regarding your question, it was nice to have improvement (We confirmed that there is no adverse effect on the original feature). I checked the change-log for material library, and it looked pretty innocent, so we went ahead with it.

ParaskP7:
👋 @khaykov !

Perfect, I was thinking that, but played it safe and asked! 😅 Feel free to merge it with `develop` 🟢 light from my side.

PR #14437, opened 2021-04-09
Adding user profile bottom sheet

develric:
Part of the #14137 and #14138 

This PR adds a user profile bottom sheet to the list of likers for both notifications and reader. Companion Fluxc PR [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1947)

NOTE: it requires D59220 endpoint changes to be shipped before it can be merged (now deployed). Note also that to check the CI built APK you need to deploy the above mentioned differential to your sandbox until those endpoint changes are not shipped.

<p align=center><image src="https://user-images.githubusercontent.com/47797566/114324279-8444d500-9b29-11eb-96fb-71cc2d29e577.png" width=300/></p>

### To test

Design has already been verified and approved. Should be enough to go to list of likers (both from reader or notifications) and open the user profile bottom sheet from there for users listed. 

Check that:
- the information presented are consistent with those of the user tapped on.
- when bio is missing you the relevant section is not shown
- when no preferred blog is present, all the Site section is not shown

During testing, the site name was not decoded correctly so that for example things like "It's great" should have resulted in something like "it&#039;s great". In our understanding now the text fields on the bottom sheet should be encoded/decoded correctly but good to have some attention for that, thanks 🙇 .

## Regression Notes
Feature still behind feature flag.

Feature still behind feature flag.

Feature still behind feature flag.

PR submission checklist:


develric:
Hi @mattmiklic , reporting here some advices/enhancements we discussed on slack, and an image that shows how I tried to implement them (I basically copied from the list of specs request you shared so it should be all, but feel free to correct in case something is wrong/missing of course 🙇 ).

<p align=center><image src="https://user-images.githubusercontent.com/47797566/114753561-dff4a580-9d57-11eb-8dc4-cceb7fa1385b.png" width=300/></p>

- Avatar size moved to 56dp
- User name moved to Headline6 with 20sp
- User login moved to textAppearanceBody with 16sp
- User bio moved to textAppearanceBody2 with 14sp;
- Site section label moved to Subtitle 14sp medium weight high emphasis
- Blavatar size moved to 40dp
- Site title and url: I removed the font padding and placed a 2dp in the middle to make the vertical alignment respect the avatar/blavatar more natural (hopefully 😬 )

In addition to the above I increased the max lines for bio to 5 before to ellipsize. 

PS: I like your idea to eventually expand the bio on tap and will create an GH issue for it to probably tackle in a future iteration I guess 👍 !

mattmiklic:
Thanks @develric, things are looking really good here. I just have one other question; could you check the line height of the profile text? It's 14sp, so the line height should be 20sp. It looks like it's currently a bit less than that.

develric:
Hi sir 👋 , I confirm the bio text has a 14sp text size. I was not adding any `lineSpacingExtra`explicitly; using 6sp I get the result below. Not fully sure this maps exactly to your intention, let me know 🙇 

| Before | After |
|---|---|
| image | image |

mattmiklic:
That's a bit closer; in the "after" screenshot it looks like the line height is 22 now, so maybe by using 4sp of lineSpacingExtra that would do the trick?

develric:
>it looks like the line height is 22 now

You are right @mattmiklic 😄! I think we need to take into account the font padding at least. I ended up using a `lineSpacingMultiplier` of 1.2 that seems to behave nicely at least to me (in the following images the grid is 4dp).

| Before | After |
|---|---|
| image | image |

mattmiklic (approved):
Looks good! Thanks for making those edits. Design is approved. 🥇

WordPress/src/main/java/org/wordpress/android/ui/engagement/BottomSheetUiState.kt
```diff
@@ -0,0 +1,17 @@
...
+    data class UserProfileUiState(
+        val userAvatarUrl: String,
+        val blavatarUrl: String,
+        val userName: String,
+        val userLogin: String,
+        val userBio: String,
+        val siteTitle: String,
+        val siteUrl: String,
+        val siteId: Long,
+        val onClick: ((siteId: Long, siteUrl: String) -> Unit)? = null
```
khaykov: Since this `onClick` is triggered when you click on the Site inside user profile, and not on user profile, I think it makes sense to rename it to something like `onSiteClickListener` or something like this, to void confusion :)

khaykov (approved):
Works really well! I left one small comment.

Not related to this PR, but some sites (google, yahoo) have a blog ID, and we try to open them in reader.

PR #14441, opened 2021-04-12
Jetpack App (Basics): Update Tab Bar Colors (Top Bar)

### LINKED ISSUE #14435: Jetpack App (Basics): Update Tab Bar Colors (Top Bar)
Parent: #14293

Jetpack App:
- In light mode, a selected top tab has `black` text and `green` underline ().
- In dark mode, both the text and line are `green`. Should be `white` text with `green` underline.
For example, see `Reader` and `Stats` screens.

PS: As discussed this exists on `WordPress` app too. So we should change on both.

ParaskP7:
Fixes #14435

This PR updates the `Jetpack`, but also `WordPress`, related tab bar colors based on the latest design feedback.

App | Light Theme | Dark Theme (Before) | Dark Theme (After)
-----|-----|-----|-----
WordPress | wordpress_light_theme | wordpress_dark_theme_before | wordpress_dark_theme_after
Jetpack | jetpack_light_theme | jetpack_light_theme_before | jetpack_light_theme_after

**To Test**
- Build, launch and smoke test the `white/blue` tab bar colors (text and line) for the `WordPress` app (e.g. `wordpressWasabiDebug`).
- Build, launch and smoke test the `white/green` tab bar colors (text and line) for the `Jetpack` app (e.g. `jetpackWasabiDebug`).

**Merge Instructions**
- ~~A quick design review by @osullivanchris on the tab color changes for the Jetpack app.~~ 🟢 
- ~~Remove `Needs Design Review` label.~~ 🟢 
- A quick review by @khaykov on why the `android:layout_marginBottom` is needed on the light theme. And, why the difference with the dark theme. #12675
- Remove `Not Ready for Merge` label.
- Merge as normal.

## Regression Notes
2. What I did to test those areas of impact (or what existing automated tests I relied on) 🟢
3. What automated tests I added (or what prevented me from doing so) 🟢

PR submission checklist:

### DIFF
--- a/WordPress/src/main/res/values-night/styles.xml
+++ b/WordPress/src/main/res/values-night/styles.xml
@@ -99,12 +99,6 @@
         <item name="android:navigationBarColor">?attr/colorSurface</item>
     </style>
 
-    <style name="WordPress.TabLayout" parent="Widget.MaterialComponents.TabLayout">
-        <item name="android:elevation">0dp</item>
-        <item name="android:background">@drawable/tab_layout_background</item>
-        <item name="android:duplicateParentState">true</item>
-    </style>
-
     <style name="WordPress.FloatingActionButton" parent="Widget.MaterialComponents.FloatingActionButton">
         <item name="android:backgroundTint">?attr/colorSecondaryVariant</item>
         <item name="tint">?attr/colorOnSurface</item>
--- a/WordPress/src/main/res/values/styles.xml
+++ b/WordPress/src/main/res/values/styles.xml
@@ -383,7 +383,7 @@
         <item name="actionTextColorAlpha">1.0</item>
     </style>
 
-    <style name="WordPress.TabLayout" parent="Widget.MaterialComponents.TabLayout.Colored">
+    <style name="WordPress.TabLayout" parent="Widget.MaterialComponents.TabLayout">
         <item name="android:background">@drawable/tab_layout_background</item>
         <item name="android:elevation">0dp</item>
         <item name="android:duplicateParentState">true</item>


ParaskP7:
👋 @khaykov !

Apologies for dragging you in this PR, but I wasn't sure why this `android:layout_marginBottom` is needed on the light theme for the `WordPress.TabLayout` style. And, why the difference with the dark theme. Maybe it is not needed after all and can be deleted that as well within this PR. Wdyt? 🙂 

PS: This PR (#12675) actually introduced this but I am still not sure about it.

osullivanchris (approved):
Looks good, thanks @ParaskP7 . This is exactly what I wanted. And glad you are addressing across both apps.

Good to ping @khaykov too. As maybe there is a reason we missed why it was that way on dark mode before.

ParaskP7:
👋 @osullivanchris !

> Looks good, thanks @ParaskP7 . This is exactly what I wanted. And glad you are addressing across both apps.

Thanks for the review and for the 👍 ! 🙏 

> Good to ping @khaykov too. As maybe there is a reason we missed why it was that way on dark mode before.

True, I agree.

ashiagr (approved):
Reviewed code changes and tested on both apps, 🟢 from my side ✨.

khaykov:
Hey @ParaskP7 ! I honestly don't remember the reason behind bottom margin for the light theme, but I believe it might have been related to the shadow that appears during lift on scroll (it only appears in light mode). I tested the screens in the PR you referenced with and without bottom margin, and nothing looks out of place for me :)

One thing I wanted to note - removing `WordPress.TabLayout` from dark mode style switches base style of tabs from `Widget.MaterialComponents.TabLayout` to `Widget.MaterialComponents.TabLayout.Colored`, and this affects the ripple color - it because almost invisible in dark mode.

ParaskP7:
👋 @khaykov !

Thank you so much for your input! 🙏 

> I honestly don't remember the reason behind bottom margin for the light theme, but I believe it might have been related to the shadow that appears during lift on scroll (it only appears in light mode). I tested the screens in the PR you referenced with and without bottom margin, and nothing looks out of place for me :)

Yeap, I tested it as well and couldn't see any difference, I am indeed not a designer, but for the life of me I couldn't... Also, what I did is to increase the `android:layout_marginBottom` value from `0.3dp` to `30dp`, then `300dp` and couldn't see anything changing as well. 🤷 

> One thing I wanted to note - removing WordPress.TabLayout from dark mode style switches base style of tabs from Widget.MaterialComponents.TabLayout to Widget.MaterialComponents.TabLayout.Colored, and this affects the ripple color - it because almost invisible in dark mode.

Oh you are right, I actually missed the fact that the dark and light theme doesn't share the exact same `TabLayout` parent. Since it is almost invisible in dark mode, and once again, in my eyes it is pretty invisible, do you think it is worth it to have the styles duplicated for that reason only? I actually I am not afraid of the duplication that much but rather the fact that someone would miss adding/removing/updating an attribute on light/dark theme and not the other, just like it happened already and we just noticed that now. Wdyt?

👋 @osullivanchris , what are your thoughts on it? Do you have any concerns about the ripple color on our tab bars being affected negatively somehow?

osullivanchris:
> 👋 @osullivanchris , what are your thoughts on it? Do you have any concerns about the ripple color on our tab bars being affected negatively somehow?

In the app right now, it looks like we have already lost the ripple effect on the tabs in light mode to achieve this styling. So if we were ok with it on light mode, it probably gives us precedent to do it on dark mode?

It is a pity to lose that little native touch by going with this colour styling. But if we can't have the ripple in this one place, I think its ok.

ParaskP7:
👋 @osullivanchris !

Thanks for your input! 🙏 

> In the app right now, it looks like we have already lost the ripple effect on the tabs in light mode to achieve this styling. So if we were ok with it on light mode, it probably gives us precedent to do it on dark mode?

This makes sense to me. 👍 

> It is a pity to lose that little native touch by going with this colour styling. But if we can't have the ripple in this one place, I think its ok.

True. 👍

khaykov:
@ParaskP7 @osullivanchris I deleted my original rambling comment, and here is the summary:

We should maintain consistent behavior between components and fix it when we find out inconsistencies. We should fix tab ripple in light mode, and make sure it works in similar way in the dark mode. 

We can change base tab style in light mode to `Widget.MaterialComponents.TabLayout` and specify tab colors that we want manually. This will restore the ripple on the light mode, keep it in the dark mode and will give us desired colors.

ParaskP7:
👋 @khaykov !

Thanks again for your input! 🙏 

> We should maintain consistent behavior between components and fix it when we find out inconsistencies. We should fix tab ripple in light mode, and make sure it works in similar way in the dark mode.

💯 

> We can change base tab style in light mode to `Widget.MaterialComponents.TabLayout` and specify tab colors that we want manually. This will restore the ripple on the light mode, keep it in the dark mode and will give us desired colors.

True, I can do that, will do that indeed!

Btw, on the `...specify tab colors that we want manually`, what would you suggest?

khaykov:
@ParaskP7 
> Btw, on the ...specify tab colors that we want manually, what would you suggest?

I just mean that if `Widget.MaterialComponents.TabLayout` is referencing colors you do not want, you can manually point them to correct ones with `tabIndicatorColor`.

ParaskP7:
👋 @khaykov !

> I just mean that if Widget.MaterialComponents.TabLayout is referencing colors you do not want, you can manually point them to correct ones with tabIndicatorColor.

Ah got you now! Yeap, will remove the `.Colored` from `Widget.MaterialComponents.TabLayout` then and double check all that again. If I find anything difference or weird, color-wise, I'll let you know. 🙏

ParaskP7:
👋 @khaykov and @osullivanchris !

I tried the change that Klymentiy suggested, which adds back the ripple effect for the dark theme, but also enables in on the light theme. I actually like that and I am seeing that apart from it adding this ripple effect, the rest of the colors remain as we want them to be. See `WordPress` videos below to get a feeling for it on both the light and dark modes:

https://user-images.githubusercontent.com/9729923/114852919-ecc2d900-9deb-11eb-85d5-5891737b9a59.mp4

https://user-images.githubusercontent.com/9729923/114853036-0ebc5b80-9dec-11eb-921a-68767c77bdfe.mp4

You will notice that, when you tab again on a tab that is already selected, the ripple effect color ends up being this soft `blue` color for the `WordPress` app (soft `green` for the `Jetpack` app). Instead, when tab on an unselected tab, the ripple effect color ends up being a soft `grey` instead.

Let me know what you think, we can either go with this one, or I can revert to the previous one, the one without ripple effects at all (on both themes). However, I feel that we either do one or the other and not mix and match, the way it was done previously, for consistency purposes (both UI and code-wise). But, then again, it is up to the design to decide, so I'll be perfectly okay with any decision you make! 🎨

osullivanchris:
Now we have the desired colours, and the system rippled effect. Sounds to me like we got everything we wanted without compromise - how could I object?! Good job! 😍

ParaskP7:
👋 @osullivanchris !

> Now we have the desired colours, and the system rippled effect. Sounds to me like we got everything we wanted without compromise - how could I object?! Good job! 😍

Great! 😍 Thanks for the 👍 Chris!

Now, I'll just wait for @khaykov to give us another 👍 and I'll be merging this!

🚀

khaykov (approved):
Looks good to me, @ParaskP7 ! Thanks for fixing the ripple 👍

ParaskP7:
👋 @khaykov !

> Looks good to me, @ParaskP7 ! Thanks for fixing the ripple 👍

Thank YOU for driving that! 🙏

PR #14444, opened 2021-04-12
Improve firebase syncing

planarvoid:
There were a few issues I'm trying to fix with this PR.
1. We retrieved the `MySiteImprovements` flag value before we initialized firebase with the default values. This caused the the first `isEnabled` call end up with `false` and we were caching this value until the app was restarted. I've added an `init` call to the app `onCreate` method which should fix this issue.
2. I've removed caching and replaced with calling `fetchAndActivate` method in `onCreate`, simply `fetch` on return to foreground and `activate in `onCreate`. The idea here is that the first start of the app will trigger `fetchAndActivate`, this value will not change until the app calls `onCreate` again (so it doesn't change under the user's hands` and we call `fetch` multiple times and `activate` the next time the app is restarted. Let me know if it makes sense.
3. I've added multiple feature states for tracking purposes. This will let us know where the flag value came from if we need to further analyze the data. To make this work I'm checking whether the flag for the given remote field has been initialized before  to see whether we're in the default state.

Before you test:
I recommend overriding the `isEnabled` field in the `MySiteImprovementsFeatureConfig` like this: 
```
        return super.isEnabled().also { Log.d("my_site_flag", "isEnabled:  ${isEnabled()}") }
    }
```
to see the actual value of the flag.

To test:
- Start with clean installation 
- Notice in Logcat the `my_site_flag` field starts as enabled
- Login/create an account
- The field is assigned and changed based on the AB test
- The My Site screen is correctly loaded based on the updated flag
- You might need to try this several times to get both the AB values

## Regression Notes
- This might cause the My Site screen to show the old or the new value for more users than we want

- The test steps

- Parametrized unit tests for AppConfig

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/WordPress.java
+++ b/WordPress/src/main/java/org/wordpress/android/WordPress.java
@@ -260,6 +260,7 @@ public class WordPress extends MultiDexApplication implements HasAndroidInjector
         initDaggerComponent();
         component().inject(this);
         mDispatcher.register(this);
+        mAppConfig.init();
 
         // Start crash logging and upload any encrypted logs that were queued but not yet uploaded
         mCrashLogging.start(getContext());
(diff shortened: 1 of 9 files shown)


planarvoid:
Closing this PR in favour of https://github.com/wordpress-mobile/WordPress-Android/pull/14445 which targets release branch.

PR #14445, opened 2021-04-12
Improve firebase syncing

planarvoid:
There were a few issues I'm trying to fix with this PR.
1. We retrieved the `MySiteImprovements` flag value before we initialized firebase with the default values. This caused the the first `isEnabled` call end up with `false` and we were caching this value until the app was restarted. I've added an `init` call to the app `onCreate` method which should fix this issue.
2. I've removed caching and replaced with calling `fetchAndActivate` method in `onCreate`, simply `fetch` on return to foreground and `activate in `onCreate`. The idea here is that the first start of the app will trigger `fetchAndActivate`, this value will not change until the app calls `onCreate` again (so it doesn't change under the user's hands` and we call `fetch` multiple times and `activate` the next time the app is restarted. Let me know if it makes sense.
3. I've added multiple feature states for tracking purposes. This will let us know where the flag value came from if we need to further analyze the data. To make this work I'm checking whether the flag for the given remote field has been initialized before  to see whether we're in the default state.

Before you test:
I recommend overriding the `isEnabled` field in the `MySiteImprovementsFeatureConfig` like this: 
```
        return super.isEnabled().also { Log.d("my_site_flag", "isEnabled:  ${isEnabled()}") }
    }
```
to see the actual value of the flag.

To test:
- Start with clean installation 
- Notice in Logcat the `my_site_flag` field starts as enabled
- Login/create an account
- The field is assigned and changed based on the AB test
- The My Site screen is correctly loaded based on the updated flag
- You might need to try this several times to get both the AB values

## Regression Notes
- This might cause the My Site screen to show the old or the new value for more users than we want

- The test steps

- Parametrized unit tests for AppConfig

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/util/config/RemoteConfig.kt
```diff
@@ -12,20 +12,38 @@ import javax.inject.Inject
...
+    fun init() {
         val firebaseRemoteConfig = FirebaseRemoteConfig.getInstance()
         val configSettings = Builder()
                 .setMinimumFetchIntervalInSeconds(BuildConfig.REMOTE_CONFIG_FETCH_INTERVAL)
                 .build()
         firebaseRemoteConfig.setConfigSettingsAsync(configSettings)
         firebaseRemoteConfig.setDefaultsAsync(RemoteConfigDefaults.remoteConfigDefaults)
-        firebaseRemoteConfig.fetchAndActivate()
-                .addOnCompleteListener { task: Task<Boolean?> ->
+        firebaseRemoteConfig.activate()
```
renanferrari: Maybe we could attach an `addOnCompleteListener` to this as well? 

Besides that I think it would also be interesting to track something similar to the `feature_flag_set` event, but only for activated Firebase values. It seems that the Task result [returns `true` if the current call activated the fetched configs](https://firebase.google.com/docs/reference/android/com/google/firebase/remoteconfig/FirebaseRemoteConfig?authuser=1#public-taskboolean-activate) so we could check if that's `true` and then track something like a `firebase_configs_activated` event with all the current activated configs, which we could retrieve from [`FirebaseRemoteConfig::getAll`](https://firebase.google.com/docs/reference/android/com/google/firebase/remoteconfig/FirebaseRemoteConfig?authuser=1#public-mapstring,-firebaseremoteconfigvalue-getall). We could then do the same for the listener attached to `fetchAndActivate`. 

I'm thinking that tracking something like this would allow us to accurately check what is coming from Firebase (before we introduce our own logic) and, unless I'm missing something, this event would necessarily have to show a 50/50 split on Tracks, so we could check that as well. Let me know what you think!

WordPress/src/main/java/org/wordpress/android/util/config/AppConfig.kt
```diff
@@ -37,17 +48,42 @@ class AppConfig
...
+            }
+            feature.buildConfigValue -> {
+                BuildConfigValue(feature.buildConfigValue)
+            }
+            !remoteConfig.isInitialized(feature.remoteField) -> {
+                DefaultValue(remoteConfig.isEnabled(feature.remoteField))
+            }
+            else -> {
+                RemoteValue(remoteConfig.isEnabled(feature.remoteField))
+            }
```
renanferrari: Maybe instead of relying on `RemoteConfig::isInitialized` we could take advantage of [`FirebaseRemoteConfigValue::getSource`](https://firebase.google.com/docs/reference/android/com/google/firebase/remoteconfig/FirebaseRemoteConfigValue?authuser=1#public-abstract-int-getsource)? It looks like it would work really well for what we want to achieve here.

WordPress/src/main/java/org/wordpress/android/util/analytics/AnalyticsTrackerWrapper.kt
```diff
@@ -16,7 +21,32 @@ class AnalyticsTrackerWrapper
...
+
+    fun track(stat: Stat, remoteField: String, featureState: FeatureState) {
+        AnalyticsTracker.track(
+                stat,
+                mapOf(
+                        remoteField to featureState.isEnabled,
+                        "${remoteField}_state" to featureState.toName()
+                )
+        )
+    }
```
renanferrari: I think we can probably extract a common logic between this two methods?

renanferrari (commented):
Thanks a lot for taking care of this!

I've tested a bunch of scenarios with this today and it seems like the biggest issue we had was indeed that we retrieved the `MySiteImprovements` flag value before we initializing Firebase with the default values and splitting the initialization and refresh logic seems to work pretty well!

>I've removed caching and replaced with calling `fetchAndActivate` method in `onCreate`, simply `fetch` on return to foreground and `activate` in `onCreate`. The idea here is that the first start of the app will trigger `fetchAndActivate`, this value will not change until the app calls `onCreate` again (so it doesn't change under the user's hands) and we call `fetch` multiple times and `activate` the next time the app is restarted. Let me know if it makes sense.

I think this makes sense. I was initially worried that we would call both `fetchAndActivate` and `fetch` during the first app start and considered using [`FirebaseRemoteConfigInfo::getLastFetchStatus`](https://firebase.google.com/docs/reference/android/com/google/firebase/remoteconfig/FirebaseRemoteConfigInfo?authuser=1#public-abstract-int-getlastfetchstatus) to check if we had not fetched anything else before making the call to `fetchAndActivate`, but I don't think that's really necessary.

I've left a few comments below with a few suggestions, let me know if they make sense. Other than that, I'm also thinking we should add the feature flag value to the `my_site_accessed` and `site_created` events and that would probably allow us to create more comprehensive funnels as well. What do you think?

WordPress/src/main/java/org/wordpress/android/util/config/AppConfig.kt
planarvoid: this is a good idea 👍 , I haven't notice that's available (but I was wondering why it isn't). I've implemented it (and removed the appPrefs usage) in 22659e4

WordPress/src/main/java/org/wordpress/android/util/analytics/AnalyticsTrackerWrapper.kt
planarvoid: done in a752e10

WordPress/src/main/java/org/wordpress/android/util/config/RemoteConfig.kt
planarvoid: I've added the `addOnCompleteListener` in a752e10. I've added tracking of the current sync state in 85fb47a
I think it makes sense 👍. I'm wondering if `feature_flag_set` still makes sense. Since we no longer cache the values, I've changed it to `feature_flag_value` but it's now tracked every time we read the value. I can keep it for now but maybe we should reconsider if it's not unnecessary noise.

WordPress/src/main/java/org/wordpress/android/util/analytics/AnalyticsTrackerWrapper.kt
planarvoid: Fixed in a752e10

planarvoid:
Thanks for the review @renanferrari , I've implemented all your suggestions and added feature config tracking for both the events you mentioned. I think this is ready for another round but I'll still spend some more time testing it as well

renanferrari (approved):
Changes are looking really good, thank you! 

LGTM 👍 

Not merging it since we are now going to target `develop`.

PR #14448, opened 2021-04-13
Crash Fix: Replace Instance with Static Reader Tracker

ParaskP7:
This crash has been reported today by @yoavf in this https://wp.me/p4a5px-2Jq-p2 post. Also, there are already a few Sentry issues being reported as well (see [Sentry issues here](https://href.li/?https://sentry.io/organizations/a8c/issues/?query=is%3Aunresolved+ReaderPostListFragment&statsPeriod=14d)).

Since the `trackTagLoaded(...)` method is also being triggered through the static `newInstanceForTag(...)` method, this means that the `mReaderTracker` field instance will not be yet available, and as thus cannot be used, or else it will result in a null pointer exception.

This PR fixes this crash by replacing the instance with the static reader tracker equivalent function.

**To Test**
- Launch app and navigate to `Reader` tab.
- Change tabs, from `FOLLOWING`, to `DISCOVER`, to `LIKES`, `SAVED`, `AUTOMATTIC`, etc and verify that app doesn't crash.

**To Replicate The Crash**
- Find the `ReaderPostListFragment.trackTagLoad(...)` method and replace the second `return` with `stat = AnalyticsTracker.Stat.READER_LIST_LOADED;`
- Replace the static `ReaderTracker` with the `mReaderTracker` field instance.
- Rebuild, launch app and navigate to `Reader` tab.
- Change tabs, from `FOLLOWING`, to `DISCOVER`, to `LIKES`, `SAVED`, `AUTOMATTIC`, etc and notice that the app crashes.

## Regression Notes
2. What I did to test those areas of impact (or what existing automated tests I relied on) 🟢 
3. What automated tests I added (or what prevented me from doing so) 🟢 

PR submission checklist:

- [ ] I have completed the Regression Notes.

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderPostListFragment.java
@@ -2520,6 +2520,15 @@ public class ReaderPostListFragment extends ViewPagerFragment
         setCurrentTag(tag);
     }
 
+    /**
+     * WARNING: Do not replace the static reader tracker with the corresponding instance reader tracker
+     * as this will result into a {@link NullPointerException} crash on specific scenarios.
+     * <p>
+     * This is because this method is also being triggered through the static
+     * {@link ReaderPostListFragment#newInstanceForTag} method, which means that the
+     * {@link ReaderPostListFragment#mReaderTracker} field instance will not be yet available, and
+     * as thus cannot be used, or else it will result in a {@link NullPointerException}.
+     */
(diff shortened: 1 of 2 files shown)


yoavf:
I can confirm no more crashes for me with this version! Thanks for the quick fix @ParaskP7

ashiagr (approved):
Thanks for quickly fixing it @ParaskP7!  I've tested and it works correctly 👍 and the reasoning is perfect!

ParaskP7:
👋 @yoavf and @ashiagr !

Thanks both for taking the time to test and review this fix on such short notice, much appreciated! ❤️ 

Merging then... 🚀

PR #14452, opened 2021-04-13
Add random value to the MySiteImprovementsFeatureConfig.kt based on user ID

planarvoid:
This is a simple solution that fixes the unreliability of the dynamic cards experiment we're experiencing now. It uses simple function to extract the value from the user ID.

To test:
- sign up with a user
- Notice the my site flag is either On or Off (based on User ID) and the dynamic cards or quick start is shown
- try with a different user until the value is different
- another option is to try creating sites with your current user

## Regression Notes
🤞  none

PR submission checklist:


renanferrari (commented):
Thanks for the changes! Code is looking and I can confirm the solution works as expected.

The only thing I'm wondering is if we should introduce another event similar to `feature_flag_set`? I think it would be good to have an event that is always fired when the flag is first set/read. I initially thought we could use `my_site_accessed` for that but then I realized it isn't consistently called at the same moment for every scenario (a new user can create a new site before seeing the My Site screen, for example). This probably won't be necessary for any of the funnels that we need to build to measure the feature's performance, but I think it should help us verify if our assignment distribution is working as expected. Let me know what you think!

planarvoid:
Good idea about the additional tracking. I've added it in 2a6b299. I've also added a small piece of logic where I cache the last value of `hasUserId` and use that in the `name()` method to make sure that we track the last actually used value. Some more tests were added in d3317ca @renanferrari

renanferrari (commented):
Thanks again for all the changes!

> I've also added a small piece of logic where I cache the last value of `hasUserId` and use that in the `name()` method to make sure that we track the last actually used value.

This is excellent, thank you!

The new code is looking good and I can confirm it works well, however I did find a bug while re-testing the solution. It's not related to the latest changes, but it is related to this particular "userId-based" approach we're going for. 

We currently initialize each main Fragment (My Site, Reader and Notifications) on `WPMainActivity::onCreate`, more specifically in this line:

```java
mBottomNav.init(getSupportFragmentManager(), this, mMySiteImprovementsFeatureConfig.isEnabled());
```

The problem is that we won't always have a `userId` at that point since the `WPMainActivity` can be created before we finish the login process in some cases. I think this only happens when using a magic link, but I could be wrong. This means that if a user with an odd `userId` uses a magic link to login, they'll see the new My Site fragment instead of the old one, but we will still track events with `my_site_improvements_random_enabled=false`. 

I'm now trying to figure out a solution that wouldn't require us to change a lot of things (specially on the login side). Let me know what you think.

planarvoid:
Thanks for the review @renanferrari ! I've looked into a few solutions and in the end I've decided to try to use AUTH_TOKEN hashcode instead of user ID. The reason is that we receive the AUTH token from the magic link so we can decide the value of the flag before we start main activity with the smallest possible changes to login flow. Let me know what you think!

renanferrari (approved):
LGTM 👍

Issue #14302, opened 2021-03-22
Jetpack App (Basics): Update Login UI


ParaskP7:
Parent: #14293

- [x] - Update prologue messages (for more info refer to `pcdRpT-ld-p2`).@ashiagr 
- [x] - Update prologue Images (for more info refer to `pcdRpT-ld-p2`). @ashiagr 
- [x] - Update no jetpack sites view @zwarm 
- [x] - Update missing jetpack view @zwarm 
~~Review existing verbiage for entering email address and update verbiage to reflect no signup.~~ Confirmed with iOS that no changes are needed.
- [x] - Merge layout files into single resource @zwarm 
**Note**: Changes to `strings.xml` in the Login Lib _may_ need corresponding changes in Wordpress and Woo.

PR #14459, opened 2021-04-14
Jetpack App (Basics): Handle login

### LINKED ISSUE #14303: Jetpack App (Basics): Handle Login
Parent: #14293

Option 1 (not pursuing)
- Restrict to WP.com login only
- Disable sign up during login, if possible
- Display an error if the user logs in without any Jetpack sites available
- Disable social logins

**Option 2 - To be implemented for Basics** [See for discussion](https://wp.me/pcdRpT-kx) 
The first 3 tasks are in PR #14459
* Add an epilogue screen that is displayed when the user doesn't have any Jetpack site ("empty screen"). UI TBD
* When a user enters a self-host URL, check if it's Jetpack and show .com login view if it is
* When a user enters a self-host URL,  if it's not a Jetpack site or not connected, show error view. UI TBD
* Disable sign up PR #14453

zwarm:
Fixes #14303 

This PR does the following:

- If the user enters a Jetpack site address and it is linked to a WordPress.com account, we redirect them to login to their WordPress.com account
- If the user enters a Jetpack site that is not linked, we display an error
- If the user logs into a WordPress.com account that does not contain any Jetpack sites we display an error

**Details**
- This PR contains changes to the following Login Lib Classes
- - LoginEmailFragment
- - LoginSiteAddressFragment
- There will be a separate PR to merge these changes back into the subtree

- To start us down the road of unit testing `LoginActivity`, `LoginViewModel` was introduced. This class extracts  `handleSiteAddressError` and `handleNoJetpackSites` from LoginActivity and posts back LoginNavigationEvents. This allowed for the creation of two basic unit tests in `LoginViewModelTest`. 

- `LoginNoSitesErrorFragment` was created to handle the case when No Jetpack Sites are associated with the user. 
- `LoginSiteCheckErrorFragment` was created to handle the case when the site address entered is not a Jetpack site.
-  Placeholder error views were added and the UI will be addressed separately.

**Notes**
- Connected tests are failing because CI is having trouble with locating flavor builds. These tests will need to be run after the needed CI changes have been applied by P9.
- The PR for the Login lib will need to be created after we merge the `feature/jetpack-app-basics` because the LoginMode is the lib has not been updated yet. FYI @ashiagr - let's sync about getting that in there. Thanks,

Pretest
- You must update your local copy of google-services.json before running the app.
- Download this branch

**To test:**
#### WordPress.com account with Jetpack sites
1. Tap the 'Login with WordPress' button
2. Enter WordPress.com credentials for an account that has at least 1 Jetpack site linked
3. After logging in you should be brought to the login epilogue (the view that displays your sites)

#### WordPress.com account with NO Jetpack sites
1. Tap the 'Login with WordPress' button
2. Enter WordPress.com credentials for an account that is not linked to any Jetpack sites
3. After logging in you should be brought to the error view with the message 'No Jetpack Sites'

#### WordPress.com account with NO Jetpack sites Part 2
1. After seeing the error from the step above, force quit the application, and restart it
2. You should be brought to the login prologue view again

#### Enter Site Address: Jetpack site that's connected to WordPress.com
1. Tap the 'Enter Site Address' button
2. Enter Jetpack site URL that is linked to a WordPress.com account, look here for examples: http://do.wpmt.co/
3. You should be brought to the WordPress.com email login view

#### Enter Site Address: Jetpack site that's **NOT** connected to WordPress.com
1. Tap the 'Enter Site Address' button
2. Enter `http://do.wpmt.co/jp-disconnected/` or another site that has Jetpack installed but not connected
3. You should be brought to an error view with the message "The website {site} is not a Jetpack site"

#### Enter Site Address: Site that has Jetpack installed but not activated
1. Tap the 'Enter Site Address' button
2. Enter `https://emily.tools/jetpack-installed-but-not-active`
3. You should be brought to an error view with the message "The website {site} is not a Jetpack site""

## Regression Notes
The WordPress login flow.

I tested using the same accounts on the WordPress scheme to verify the login flow still works correctly.

I created two new unit tests in `LoginViewModelTest`

PR submission checklist:


libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginSiteAddressFragment.java
```diff
@@ -432,6 +431,16 @@ private void handleConnectSiteInfoForWordPress(ConnectSiteInfoPayload siteInfo)
...
     }
 
+    private void handleConnectSiteInfoForJetpack(ConnectSiteInfoPayload siteInfo) {
+        endProgressIfNeeded();
+
+        if (siteInfo.hasJetpack && siteInfo.isJetpackConnected && siteInfo.isJetpackActive) {
+            mLoginListener.gotWpcomSiteInfo(UrlUtils.removeScheme(siteInfo.url));
+        } else {
+            mLoginListener.handleSiteAddressError(siteInfo);
+        }
```
ashiagr, line 441: I tried Jetpack app login via site address putting an atomic site address. Login was successful but since we filter out atomic sites in the Jetpack app, the site was not shown in the site picker.

This is the atomic site I used, notice it has `jetpack` and `jetpack_connection` set to `true`
atomic site

I think we should restrict this login and display an error otherwise it will be confusing why the atomic site was not shown.

WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
```diff
@@ -913,6 +938,12 @@ public void loginViaSiteCredentials(String inputSiteAddress) {
 
     @Override
     public void handleSiteAddressError(ConnectSiteInfoPayload siteInfo) {
-        // Not used in WordPress app
+        mViewModel.handleSiteAddressError(siteInfo);
+    }
+
+    public void handleNoJetpackSites() {
+        // hide keyboard if it you can
+        org.wordpress.android.util.ActivityUtils.hideKeyboard(this);
```
ashiagr: Minor 🔍: Consider 
1. importing `org.wordpress.android.util.ActivityUtils`
2. removing `it` from the `hide keyboard if it you can` comment

WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginViewModel.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val resourceProvider: ResourceProvider
+) : ViewModel() {
+    private val _navigationEvents = MediatorLiveData<Event<LoginNavigationEvents>>()
+    val navigationEvents: LiveData<Event<LoginNavigationEvents>> = _navigationEvents
+
+    fun handleSiteAddressError(siteInfo: ConnectSiteInfoPayload) {
+        val protocolRegex = Regex("^(http[s]?://)", IGNORE_CASE)
+        val siteAddressClean = siteInfo.url.replaceFirst(protocolRegex.toString().toRegex(), "")
+        val errorMessage: String = resourceProvider.getString(R.string.login_not_a_jetpack_site, siteAddressClean)
+        val frag = newInstance(siteAddressClean, errorMessage)
```
ashiagr: Warning ⚠️:  `ViewModel` should not reference a `fragment`. Consider replacing `SlideInFragment` navigation event with an event like `ShowSiteAddressError` (similarly  `ShowNoJetpackSitesError`  in the `handleNoJetpackSites` method) and creating an appropriate fragment instance in the observer inside the `Activity`. We did something similar [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/reader/ReaderFragment.kt#L158-L160) when opening Reader interest fragment for a view model event.

Suggestion 💡:  How about we rename `handleSiteAddressError` called from the Activity to `onHandleSiteAddressError` to follow the convention in the rest of the VMs?

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginNoSitesErrorFragment.kt
```diff
@@ -0,0 +1,91 @@
...
+        (requireActivity().application as WordPress).component().inject(this)
+
+        val binding = FragmentLoginNoSitesErrorBinding.bind(view)
+
+        binding.loginErrorMsg.text = errorMsg
+
+        viewModel = ViewModelProvider(this@LoginNoSitesErrorFragment, viewModelFactory)
+                .get(LoginNoSitesErrorViewModel::class.java)
+        viewModel.signOutWordPress(requireActivity().application as WordPress)
+
```
ashiagr: Suggestion💡: We might want to include any startup action like signout inside `viewModel.start` or `viewModel.onViewCreated` instead of directly calling from the `fragment`.

Suggestion💡:  Method can be split into `initDagger`, `initErrorMessageView`,  `initViewModel`. 

PS: Same can be applied to `LoginSiteCheckErrorFragment`.

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginNoSitesErrorFragment.kt
```diff
@@ -0,0 +1,91 @@
...
+    override fun onDetach() {
+        loginListener = null
+        super.onDetach()
+    }
+
+    override fun onResume() {
+        super.onResume()
+
+        unifiedLoginTracker.track(step = Step.NO_JETPACK_SITES)
```
ashiagr: Question❓: Since we have a dedicated VM for this fragment, how about we move this tracking to `vm.onFragmentResume`?

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginNoSitesErrorViewModel.kt
```diff
@@ -0,0 +1,23 @@
...
+import javax.inject.Named
+
+class LoginNoSitesErrorViewModel @Inject constructor(
+    @Named(UI_THREAD) mainDispatcher: CoroutineDispatcher,
+    @Named(BG_THREAD) val bgDispatcher: CoroutineDispatcher
+) : ScopedViewModel(mainDispatcher) {
+    fun signOutWordPress(application: WordPress) {
+        launch {
+            withContext(bgDispatcher) {
+                application.wordPressComSignOut()
```
ashiagr, line 38: Suggestion 💡: We might want to add a test for signout action. It can be similar to [this test case](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/test/java/org/wordpress/android/ui/main/MeViewModelTest.kt#L29-L31) in the `MeViewModelTest`.

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginNoSitesErrorFragment.kt
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    private fun initBackPressHandler() {
+        requireActivity().onBackPressedDispatcher.addCallback(
+                viewLifecycleOwner,
+                object : OnBackPressedCallback(
+                        true
+                ) {
+                    override fun handleOnBackPressed() {
+                        ActivityLauncher.showSignInForResultJetpackOnly(requireActivity())
```
ashiagr: Suggestion 💡: How about we get this done using `vm.onBackPressed` that passes a navigation event `ShowSignInForResultJetpackOnly`? We can then add a test that appropriate action is triggered on back press.

WordPress/src/test/java/org/wordpress/android/ui/accounts/LoginViewModelTest.kt
```diff
@@ -0,0 +1,60 @@
...
+        whenever(resourceProvider.getString(R.string.login_no_jetpack_sites)).thenReturn("No Jetpack sites.")
+    }
+
+    @Test
+    fun `given no jetpack sites, then SlideInFragment navigation event is posted`() {
+        val navigationEvents = initObservers().navigationEvents
+
+        viewModel.handleNoJetpackSites()
+
+        Assertions.assertThat(navigationEvents.last()).isInstanceOf(SlideInFragment::class.java)
```
ashiagr: Suggestion 💡:  This test case and the next one will pass even if the error message is the same for both scenarios. We might want to refactor the navigation event slightly: first by not passing the fragment (as suggested in another comment), second by including the exact error message that we can equate in our tests here.

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginSiteCheckErrorFragment.kt
```diff
@@ -0,0 +1,104 @@
...
+        (activity as AppCompatActivity).setSupportActionBar(toolbar)
+
+        (activity as AppCompatActivity).supportActionBar?.let {
+            it.setDisplayHomeAsUpEnabled(true)
+            it.setDisplayShowTitleEnabled(true)
+        }
+
+        binding.loginErrorMsg.text = errorMsg
+    }
+    override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
```
ashiagr: Minor 🔍:  Seems we need to format the file.

ashiagr (commented):
Thanks for including detailed instructions @zwarm! I tested both WordPress and Jetpack app and all the described scenarios passed! I really appreciate that you added tests for the new login conditions and their navigation 👌. It must have been very tricky. 

I've left few suggestions and asked about one more scenario that we might want to handle (Jetpack app: login via site address -> enter atomic site address that has jetpack by default -> login successful without atomic site in the site picker).  

Hope it helps.

libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginSiteAddressFragment.java
zwarm: @ashiagr Good point, the payload doesn't include information to determine if the site is Atomic - and it's treated as kind of a fall through situation. Let me check with iOS, as we are using the same jetpack checks. Thanks.

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginNoSitesErrorViewModel.kt
zwarm: @ashiagr _ lol, I didn't because that functionality is already tested in the MeViewModelTest ... our test would test the same exact functionality, but I'll add it.

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginNoSitesErrorFragment.kt
zwarm: Sure

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginNoSitesErrorFragment.kt
zwarm: `LoginActivity` doesn't listen for events from `LoginNoSitesErrorFragment`, only from `LoginViewModel`, so yes we can set up livedata/observers for this fragment/vm pair (a nice improvement), but ultimately we still need to call ActivityLauncher in the `LoginNoSitesErrorFragment`.

ashiagr: > but ultimately we still need to call ActivityLauncher in the LoginNoSitesErrorFragment.

Correct, just that our fragment will now get direction from the VM about the navigation to be done on the back press.

libs/login/WordPressLoginFlow/src/main/java/org/wordpress/android/login/LoginSiteAddressFragment.java
zwarm: @ashiagr - I confirmed with iOS and this is the expected behavior (Woo follows the same pattern). We redirect from Site Address to WP.comLogin, then we handle No Jetpack Sites. There is not enough info on the endpoint to make that call before login

WordPress/src/test/java/org/wordpress/android/ui/accounts/LoginViewModelTest.kt
zwarm: That is correct, these were tests for NavigationEvents only, not messaging.  I'll add new tests for the error messages

zwarm:
@ashiagr - I have addressed all your comments except for the one about Atomic site, which is the expected behavior. Thanks. :)

zwarm:
@ashiagr FYI: I created a new task in #14302 for reviewing all verbiage for existing flows, as making changes in LoginFlow strings _may_ require changes to WordPress and Woo. I'd rather we do this once, instead of twice (once with placeholder and once with actual content

ashiagr (approved):
Thank you @zwarm, everything looks good! We can update the related UI and verbiage as part of #14302 when the design is ready, thank you for adding the sub-tasks ❤️.

PR #14466, opened 2021-04-15
Add stats email deeplink handling

### LINKED ISSUE #14449: Add email deeplink handling for stats
Add handling of `public-api.wordpress.com/mbar?redirect_to=wordpress.com/stats` links to the Stats screen

planarvoid:
Fixes #14449 

This PR adds handling of email links like `public-api.wordpress.com/mbar/...redirect_to=wordpress.com/stats/day/example.com`. As addition I've pulled the logic for simple stats links like `wordpress.com/stats/week/example.com` to be handled by the same `StatsLinkHandler`. One more improvement is that the deeplinks now handle the stats timeframe as well as the site so if there is a link to `day`, it opens the day tab. Let me know if you need links to test this feature out. 

To test:
- Make sure you're logged in with an account that has the `your_site.com` site 
- Open a deeplink like `public-api.wordpress.com/mbar/...redirect_to=wordpress.com/stats/day/your_site.com`
- Notice it's opened in the app
- Notice it opens the Stats screen with the "Day" tab selected for `your_site.com`
- Test this with replacing `day` in the link with `week`, `year`, `insights` for all the screens

To test:
- Make sure you're not logged in with an account that has the `another_site.com` site 
- Open a deeplink like `public-api.wordpress.com/mbar/...redirect_to=wordpress.com/stats/day/another_site.com`
- Notice it's opened in the app
- Notice it opens the Stats screen for the currently selected site

To test:
- Make sure you're logged in with an account that has the `your_site.com` site 
- Open a deeplink like `wordpress.com/stats/day/your_site.com`
- Notice it's opened in the app
- Notice it opens the Stats screen with the "Day" tab selected for `your_site.com`

To test:
- Open a deeplink like `public-api.wordpress.com/bar/...redirect_to=wordpress.com/stats/day/your_site.com`
- Notice it's opened in the browser

## Regression Notes
- none 🤞 

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
@@ -388,7 +388,7 @@ public class ActivityLauncher {
         viewStatsInNewStack(context, site, null);
     }
 
-    public static void viewStatsInNewStack(Context context, SiteModel site,  @Nullable StatsTimeframe statsTimeframe) {
+    public static void viewStatsInNewStack(Context context, SiteModel site, @Nullable StatsTimeframe statsTimeframe) {
         if (site == null) {
             handleMissingSite(context);
             return;
--- a/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
@@ -85,6 +85,7 @@ import org.wordpress.android.ui.publicize.PublicizeListActivity;
 import org.wordpress.android.ui.sitecreation.SiteCreationActivity;
 import org.wordpress.android.ui.stats.StatsConnectJetpackActivity;
 import org.wordpress.android.ui.stats.StatsConstants;
+import org.wordpress.android.ui.stats.StatsTimeframe;
 import org.wordpress.android.ui.stats.StatsViewType;
 import org.wordpress.android.ui.stats.refresh.StatsActivity;
 import org.wordpress.android.ui.stats.refresh.StatsViewAllActivity;
@@ -400,6 +401,29 @@ public class ActivityLauncher {
         Intent mainActivityIntent = getMainActivityInNewStack(context);
 
         Intent statsIntent = StatsActivity.buildIntent(context, site);
+
+        taskStackBuilder.addNextIntent(mainActivityIntent);
+        taskStackBuilder.addNextIntent(statsIntent);
+        taskStackBuilder.startActivities();
+    }
+
+    public static void viewStatsInNewStack(Context context, SiteModel site, StatsTimeframe statsTimeframe) {
+        if (site == null) {
+            AppLog.e(T.STATS, "SiteModel is null when opening the stats from the deeplink.");
+            AnalyticsTracker.track(
+                    STATS_ACCESS_ERROR,
+                    ActivityLauncher.class.getName(),
+                    "NullPointerException",
+                    "Failed to open Stats from the deeplink because of the null SiteModel"
+                                  );
+            ToastUtils.showToast(context, R.string.stats_cannot_be_started, ToastUtils.Duration.SHORT);
+            return;
+        }
+        TaskStackBuilder taskStackBuilder = TaskStackBuilder.create(context);
+
+        Intent mainActivityIntent = getMainActivityInNewStack(context);
+
+        Intent statsIntent = StatsActivity.buildIntent(context, site, statsTimeframe);
 
         taskStackBuilder.addNextIntent(mainActivityIntent);
         taskStackBuilder.addNextIntent(statsIntent);
--- a/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
@@ -385,54 +385,55 @@ public class ActivityLauncher {
     }
 
     public static void viewStatsInNewStack(Context context, SiteModel site) {
-        if (site == null) {
-            AppLog.e(T.STATS, "SiteModel is null when opening the stats from the deeplink.");
-            AnalyticsTracker.track(
-                    STATS_ACCESS_ERROR,
-                    ActivityLauncher.class.getName(),
-                    "NullPointerException",
-                    "Failed to open Stats from the deeplink because of the null SiteModel"
-                                  );
-            ToastUtils.showToast(context, R.string.stats_cannot_be_started, ToastUtils.Duration.SHORT);
-            return;
-        }
-        TaskStackBuilder taskStackBuilder = TaskStackBuilder.create(context);
-
-        Intent mainActivityIntent = getMainActivityInNewStack(context);
-
-        Intent statsIntent = StatsActivity.buildIntent(context, site);
-
-        taskStackBuilder.addNextIntent(mainActivityIntent);
-        taskStackBuilder.addNextIntent(statsIntent);
-        taskStackBuilder.startActivities();
+        viewStatsInNewStack(context, site, null);
     }
 
-    public static void viewStatsInNewStack(Context context, SiteModel site, StatsTimeframe statsTimeframe) {
+    public static void viewStatsInNewStack(Context context, SiteModel site,  @Nullable StatsTimeframe statsTimeframe) {
         if (site == null) {
-            AppLog.e(T.STATS, "SiteModel is null when opening the stats from the deeplink.");
-            AnalyticsTracker.track(
-                    STATS_ACCESS_ERROR,
-                    ActivityLauncher.class.getName(),
-                    "NullPointerException",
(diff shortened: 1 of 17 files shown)


renanferrari (commented):
Thanks for the changes! 

I've tested the app and can confirm everything works as expected! The only thing it's not clear for me is if a link like `public-api.wordpress.com/mbar/...redirect_to=wordpress.com/stats/week/another_site.com` is supposed to open the Stats screen with the "Week" tab selected for the current selected site or if it's just supposed to open the Stats screen with whatever tab was selected before (this was the case for me). Both should be fine, I just wanna make sure we have the intended behavior.

Code-wise everything is looking pretty good. I just left one minor comment below, but it's barely related to your changes, so don't feel compelled to address it here.

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
```diff
@@ -406,6 +407,29 @@ public static void viewStatsInNewStack(Context context, SiteModel site) {
...
+        TaskStackBuilder taskStackBuilder = TaskStackBuilder.create(context);
+
+        Intent mainActivityIntent = getMainActivityInNewStack(context);
+
+        Intent statsIntent = StatsActivity.buildIntent(context, site, statsTimeframe);
+
+        taskStackBuilder.addNextIntent(mainActivityIntent);
+        taskStackBuilder.addNextIntent(statsIntent);
+        taskStackBuilder.startActivities();
+    }
```
renanferrari: This is not critical at all, but I wish there was a way for us to skip the transition animation when using `TaskStackBuilder`, as it looks kinda weird right now:

https://user-images.githubusercontent.com/830056/114952494-c2454000-9e2c-11eb-9be7-81ee9f826089.mp4

I tried adding a `Intent.FLAG_ACTIVITY_NO_ANIMATION` to the `mainActivityIntent` but it didn't work. Just wondering if you know a way to achieve this?

planarvoid:
thanks for the review!

>  Both should be fine, I just wanna make sure we have the intended behavior.

I was thinking about this. It is intentional but it might make more sense to open the `week` tab for the selected site. I think this would be a very rare usecase though. 

> This is not critical at all, but I wish there was a way for us to skip the transition animation when using TaskStackBuilder, as it looks kinda weird right now:

It does, right? I've used the previous solution and it doesn't look great tbh. I'll look into some possible improvements.

planarvoid:
I've tried to get rid of the animation and I couldn't find a way to do this well. If we remove the clear top flag, the animation is gone. However, the previous stack is still there so if you opened the app before, you're gonna have 2 instances of main activity. Doing single top introduces the animation again.

I've added a possibility of opening stats for the current site with a given timeframe (when the site doesn't exist for the current account) and I've refactored and simplified the view model code in fd50d2c. There is also a lot of new tests.

This should be ready for another round @renanferrari , thanks!

WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
```diff
@@ -384,25 +385,42 @@ public static void openEditorForReblog(
     }
 
     public static void viewStatsInNewStack(Context context, SiteModel site) {
+        viewStatsInNewStack(context, site, null);
+    }
+
+    public static void viewStatsInNewStack(Context context, SiteModel site,  @Nullable StatsTimeframe statsTimeframe) {
```
renanferrari: This should fix the checkstyle error.

```suggestion
    public static void viewStatsInNewStack(Context context, SiteModel site, @Nullable StatsTimeframe statsTimeframe) {
```

WordPress/src/main/java/org/wordpress/android/ui/DeepLinkNavigator.kt
```diff
@@ -37,6 +38,10 @@ class DeepLinkNavigator
                     navigateAction.postId
             )
             OpenStats -> ActivityLauncher.viewStatsInNewStack(activity)
+            is OpenStatsForTimeframe -> ActivityLauncher.viewStatsForTimeframeInNewStack(
+                    activity,
+                    navigateAction.statsTimeframe
+            )
```
renanferrari, line 44: Good call adding this, it makes total sense 👍

WordPress/src/main/java/org/wordpress/android/ui/DeepLinkingIntentReceiverViewModel.kt
```diff
@@ -29,117 +26,83 @@ class DeepLinkingIntentReceiverViewModel
...
+     * In these cases this function returns true
      */
-    fun shouldHandleTrackingUrl(uri: UriWrapper): Boolean {
-        // https://public-api.wordpress.com/mbar/
-        return uri.host == HOST_API_WORDPRESS_COM &&
-                uri.pathSegments.firstOrNull() == MOBILE_TRACKING_PATH
+    fun shouldHandleUrl(uri: UriWrapper): Boolean {
+        return isTrackingUrl(uri) ||
+                editorLinkHandler.isEditorUrl(uri) ||
+                statsLinkHandler.isStatsUrl(uri)
```
renanferrari: I really like the refactor you did here, I'm just not sure if we need to make the same checks here and on `buildNavigateAction`. Maybe we should just call `handleUrl` instead and make it return `true` if it was able to handle the URL? What do you think?

WordPress/src/main/java/org/wordpress/android/ui/DeepLinkingIntentReceiverViewModel.kt
```diff
@@ -29,117 +26,83 @@ class DeepLinkingIntentReceiverViewModel
...
 
     companion object {
-        private const val HOST_WORDPRESS_COM = "wordpress.com"
+        const val HOST_WORDPRESS_COM = "wordpress.com"
         private const val HOST_API_WORDPRESS_COM = "public-api.wordpress.com"
         private const val MOBILE_TRACKING_PATH = "mbar"
         private const val REGULAR_TRACKING_PATH = "bar"
         private const val REDIRECT_TO_PARAM = "redirect_to"
-        private const val POST_PATH = "post"
-        private const val STATS_PATH = "stats"
```
renanferrari: I think we can also remove the `STATS_PATH` constant from `DeepLinkingIntentReceiverActivity`.

renanferrari (commented):
>I've tried to get rid of the animation and I couldn't find a way to do this well.

Thanks for looking into it anyway!

>I've added a possibility of opening stats for the current site with a given timeframe (when the site doesn't exist for the current account) and I've refactored and simplified the view model code in fd50d2c. There is also a lot of new tests.

Thanks for all those changes, they are looking good! I left a few comments below, but nothing major. As a side note, we'll probably want to merge this before we merge #14471, so I can take care of the refactor changes on my side.

WordPress/src/main/java/org/wordpress/android/ui/DeepLinkingIntentReceiverViewModel.kt
planarvoid: I was thinking about this approach, I think you're right. It would be better. I haven't used it so far basically just because it's different than what we use in the deeplinking activity. I'll look into it.

WordPress/src/main/java/org/wordpress/android/ui/DeepLinkingIntentReceiverViewModel.kt
planarvoid: fixed in 
8990aa1

WordPress/src/main/java/org/wordpress/android/ui/DeepLinkingIntentReceiverViewModel.kt
planarvoid: Implemented in d572755

planarvoid:
thanks for the review 👍  I've addressed all of your comments and now this should be ready for another round

WordPress/src/main/java/org/wordpress/android/ui/DeepLinkingIntentReceiverViewModel.kt
```diff
@@ -25,30 +25,35 @@ class DeepLinkingIntentReceiverViewModel
...
      * `wordpress.com/stats/...`
      * `public-api.wordpress.com/mbar/`
      * In these cases this function returns true
      */
-    fun shouldHandleUrl(uri: UriWrapper): Boolean {
+    private fun shouldHandleUrl(uri: UriWrapper): Boolean {
         return isTrackingUrl(uri) ||
                 editorLinkHandler.isEditorUrl(uri) ||
                 statsLinkHandler.isStatsUrl(uri)
     }
```
renanferrari: This is definitely an improvement but I was wondering if it wouldn't be possible to completely remove this method now. I tinkered a bit with this idea and I believe it's possible, but it could potentially change some implicit behaviors, so I'll leave it for another PR.

renanferrari (approved):
Thanks again for the changes!

LGTM 👍

Issue #14301, opened 2021-03-22
Jetpack App (Basics): Update MySite Tab Icon


ParaskP7:
Parent: #14293

Change the MySite icon for Jetpack to the Home icon. Text stays as "My Site"

PR #14470, opened 2021-04-15
Jetpack App (Basics) - Add About App icon

### LINKED ISSUE #14464: Jetpack App (Basics): Update app icon in the About screen
Parent #14293

Add app icon in the About screen - Jetpack app.

- drawable: ic_my_sties_white_24dp.xml

zwarm:
Fixes #14464 

Adds the Jetpack icon to the About me view.

Note: The WordPress icon has not changed.

Theme | WordPress | Jetpack
-----|-----|-----
Light | wordpress_light | jetpack_light
Dark | wordpress_dark | jetpack_dark

**To test:**
- Build, launch and smoke test a build flavor for the WordPress app (e.g. wordpressWasabiDebug).
- Navigate to Me -> App Settings -> About WordPress for Android
- Note the Wordpress icon is shown
- Build, launch and smoke test a build flavor for the Jetpack app (e.g. jetpackWasabiDebug).
- Navigate to Me -> App Settings -> About Jetpack for Android
- Note the Jetpack icon is shown

## Regression Notes
None

Manual tested for presence of Jetpack and WordPress icon

N/A

PR submission checklist:

### DIFF
--- a/WordPress/src/jetpack/res/drawable/ic_about_app_white_24dp.xml
+++ b/WordPress/src/jetpack/res/drawable/ic_about_app_white_24dp.xml
@@ -1,0 +1,12 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <path
+        android:pathData="M12.0001,2C17.523,2 22,6.4771 22,12C22,17.5229 17.523,22 12.0001,22C6.4772,22 2,17.5229 2,12C2,6.4771 6.4772,2 12.0001,2ZM17.495,10.327L12.5071,10.327L12.5071,20.0193L17.495,10.327ZM11.4925,3.9817L6.5046,13.6741L11.4925,13.6741L11.4925,3.9817Z"
+        android:strokeWidth="1"
+        android:fillColor="@color/white"
+        android:fillType="evenOdd"
+        android:strokeColor="@color/transparent"/>
+</vector>
--- a/WordPress/src/main/res/drawable/ic_about_app_white_24dp.xml
+++ b/WordPress/src/main/res/drawable/ic_about_app_white_24dp.xml
@@ -1,0 +1,3 @@
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:drawable="@drawable/ic_my_sites_white_24dp"/>
+</layer-list>
--- a/WordPress/src/main/res/layout/about_activity.xml
+++ b/WordPress/src/main/res/layout/about_activity.xml
@@ -49,7 +49,7 @@
                     android:layout_height="50dp"
                     android:layout_marginTop="@dimen/margin_medium"
                     android:importantForAccessibility="no"
-                    android:src="@drawable/ic_my_sites_white_24dp" />
+                    android:src="@drawable/ic_about_app_white_24dp" />
 
                 <org.wordpress.android.widgets.WPTextView
                     android:id="@+id/about_first_line"


ParaskP7 (approved):
👋 @zwarm !

This is a nice tricky you did there with the `layer-list` attribute that reused the `ic_my_sites_white_24dp` drawable (❤️). However, I am a bit skeptical whether this is what we need to be doing. 🤔 

Doing a quick `find usages` on the `ic_my_sites_white_24dp` I realised that this is used not only with the `My Site` screen or the `Nav Tab` but if various other places, which has nothing to do with `My Site`. This brings me to believe that what I was thinking to suggest holds even more. So, long story short, I suggest instead that:
1) We rename the `ic_my_sites_white_24dp` to `ic_app_icon_white_24dp`,
2) Reused this `ic_app_icon_white_24dp` everywhere, including the `About Activity` screen, basically how it used to be, and
3) Add a new `ic_app_icon_white_24dp` for the `Jetpack` dimension, basically just renaming the `ic_about_app_white_24dp.xml` in there to this naming convension.
4) Then, the `WordPress` dimension related `ic_about_app_white_24dp.xml` with the `layer-list` trick with end up becoming obsolete and thus can be deleted.

I find that this way will make the diff between `WordPress` and `Jetpack` app much more straightforward and help with the reusability of this icon going forward, just like it is done on the `WordPress` dimension side.

PS: Having said all the above I understood that we might want to keep the `WordPress` such icons, even on the `Jetpack` app, as we where discussing yesterday, like the `View Admin` menu item icon on the `My Site` screen. Maybe those instances should now become the exception to the above rule. As should, we might want to duplicate the `ic_app_icon_white_24dp` to `ic_wordpress_icon_white_24dp` and use them as such to make it explicit that this is what we want.
PSS: Then, while working on #14301, we would add another exception, duplicating the `ic_app_icon_white_24dp` to `ic_my_sites_white_24dp` for the `WordPress` dimension and adding a new `ic_my_sites_white_24dp` with the designed `home` icon for the `Jetpack` dimension.

As always, I have written lots! 😄 But, it is totally up to you on how you want to proceed with that at this point! I am going to approve it for now, but not merge it, giving you some space to think about the above.

👋 @ashiagr ! What are your thoughts on all that?

osullivanchris (approved):
Visually looks fine to me, thanks for the ping 👍  

I can't speak to the file naming stuff @ParaskP7  raised, its above my level of understanding. At a high level I would think we could either:

- change it so every time we show WP icon in the app, we show Jetpack icon instead
- manually change each instance to point to a different icon in the Jetpack app

I can see pros and cons to each. There might be places we need the WP icon in future. But as a default, automatically changing WP->Jetpack. And then having places where we want WP as the exception, could be an approach.

ParaskP7:
Thanks for your input @osullivanchris , much appreciated! 🙏

ashiagr:
Thanks for the ping @ParaskP7!  I like your suggestions. 

Adding another angle to it 😅: I see that the WordPress app icon that is referenced here maps to [gridicons-my-sites](https://github.com/Automattic/gridicons/blob/master/sources/svg/gridicons-my-sites.svg) and seems like it derives its name from it. Keeping the name of the icon `ic_my_sites_xxx` might help us map it with the correct grid icon. 

If we were to use the same icon everywhere, I would've voted to keep the existing name. But with the exception that we have to deal with in the two apps, we now have two options for places of exception:

1. duplicate the entire vector icon 
2. use a separate layer-list icon that references this common vector icon

In both options, we're creating a new asset. If we need to change this common icon in the future, it will be easier with option 2. Option 1 is fine as well if everyone agrees to it and it simplifies things.

ParaskP7:
👋 @ashiagr !

Thanks for much for your input! 🙏 

Nice, I actually wasn't aware of that, `ic_my_sites_xxx` mapping to [gridicons-my-sites](https://github.com/Automattic/gridicons/blob/master/sources/svg/gridicons-my-sites.svg). You suggestions make sense. I just wonder then, maybe this `ic_my_sites_xxx` should only be used on `My Sites` for `WordPress` only and nowhere else, and we should duplicate our own `ic_app_icon_xxx` from the `ic_my_sites_xxx` to use it everywhere else.

I think what makes it difficult for us is that our `App` icon, or more specifically `WordPress` icon,  match, design-wise, the `My Site` icon and as such we ended up reusing it everywhere. But it might not have been this case, and the `My Site` icon might have been just a `Home` icon, like what we suppose to be doing for the `Jetpack` app. It seems pure coincidence that `My Site` and `WordPress` icon are the same, right?

Having said the above, I am a bit closer to your `Option.1` from above and have them duplicated and explicitly separate even thought the raw UI path and everything is exactly the same.

zwarm:
@ashiagr @ParaskP7 

I was keeping things a bit more simple for the MVP because this icon is _used everywhere_ and we have not yet discussed where else it needs to be changed. My idea was to KISS for this task and create a new task to review all the other places this icon is used, take screenshots and then get design involved as to what needs to be a jetpack icon and what needs to be WordPress.

I don't disagree with the above comments, but I prefer to leave this PR as written and move forward with the evaluation phase. We can then revisit naming as it relates to the full scope of modifications needed.

ParaskP7:
👋 @zwarm !

> I was keeping things a bit more simple for the MVP because this icon is used everywhere and we have not yet discussed where else it needs to be changed. My idea was to KISS for this task and create a new task to review all the other places this icon is used, take screenshots and then get design involved as to what needs to be a jetpack icon and what needs to be WordPress.

👌 Definitely, feel free to merge it when you are ready!

> I don't disagree with the above comments, but I prefer to leave this PR as written and move forward with the evaluation phase. We can then revisit naming as it relates to the full scope of modifications needed.

👍 

PS: I totally agree with the above Annmarie and thanks for keeping it simple, as I said, it is totally up to you. My only fear is that keeping it simple in some cases might create a mental and structural overhead going further forward into the future, that is adding complexity to an already complicated and involved structure. The way I personally go about this is to take this opportunity and add functionality while at the same time trying to unbundle the complexity. Even if `Jetpack` as MVP ends up not being used, we sure will be in a better position at the end, that is to the one we found ourselves in when we fist started this project.

zwarm:
> My only fear is that keeping it simple in some cases might create a mental and structural overhead going further forward into the future, that is adding complexity to an already complicated and involved structure. The way I personally go about this is to take this opportunity and add functionality while at the same time trying to unbundle the complexity. Even if `Jetpack` as MVP ends up not being used, we sure will be in a better position at the end, that is to the one we found ourselves in when we fist started this project.

@ParaskP7 I don't disagree and I am not saying don't do it. I'm saying we should minimize scope creep for this task. We have identified that this task uncovered a potential problem that needs to be addressed. We should create a new task for that, scope it properly, and give it the attention it deserves.

ParaskP7:
👋 @zwarm !

> I don't disagree and I am not saying don't do it. I'm saying we should minimize scope creep for this task. We have identified that this task uncovered a potential problem that needs to be addressed. We should create a new task for that, scope it properly, and give it the attention it deserves.

Definitely, I think we are on the same page, especially since a new task is going to be created on top of it, I am perfectly fine!

🙏

osullivanchris:
I hope this doesn't throw a spanner in the works. But just to point out that I think we do have a GridIcon for the Jetpack logo as well as a GridIcon for the WordPress logo. You can see the icon I'm talking about:

- on a non Jetpack site
- My Site menu 
- on the item **Plan** 

Just thinking maybe its an asset that's already implemented that we can utilise.

zwarm:
> I hope this doesn't throw a spanner in the works. But just to point out that I think we do have a GridIcon for the Jetpack logo as well as a GridIcon for the WordPress logo. You can see the icon I'm talking about:
> 
> * on a non Jetpack site
> * My Site menu
> * on the item **Plan**
> 
> Just thinking maybe its an asset that's already implemented that we can utilise.

Oops, @osullivanchris , I didn't see your comment before merging. I'm just getting back to this now and I am digging into all the places in which we show this icon. I'll cc you on the task after my research is complete. Thanks

PR #14471, opened 2021-04-15
Add deeplink support for posts sent via email

### LINKED ISSUE #14461: Add deeplink support to posts sent by email
When users are first starting to use WordPress, they receive multiple emails with links to posts that help them create and setup their site. We want to add deeplink support to most of these posts, so that they can be opened in the app's Reader. This includes links wrapped with `public-api.wordpress.com/mbar?redirect_to=*`.

renanferrari:
Fixes #14461

Our app already handles some post links and successfully redirect them to the Reader. This is done by the `ReaderPostPagerActivity`, which supports the following URL patterns:

- `http[s]://wordpress.com/read/blogs/{blogId}/posts/{postId}`
- `http[s]://wordpress.com/read/feeds/{feedId}/posts/{feedItemId}`
- `http[s]://{username}.wordpress.com/{year}/{month}/{day}/{postSlug}`

However, when posts are sent via email they are usually wrapped into a tracking URL like `public-api.wordpress.com/mbar/?redirect_to={url}` which was not being handled by the app. This PR adds support for this.

To achieve that, I had to take into account that supported URLs were already defined in the Manifest, as intent filters of the `ReaderPostPagerActivity` class, which means it wouldn't make sense to have a `shouldOpenReader` method that would just check if we can handle the same URL patterns already defined there, as we would then have to maintain the same logic in two different places. To address this, I initially considered moving all deeplink handling from `ReaderPostPagerActivity` to `DeepLinkingIntentReceiverActivity`, but since there's quite a lot of code there I looked for a simpler solution. 

In the end, I went with an approach that introduces very little changes to both `DeepLinkingIntentReceiverActivity` and `ReaderPostPagerActivity`:
1. A custom action (`org.wordpress.android.action.VIEW_POST`) was added as an intent filter of `ReaderPostPagerActivity`.
2. `ReaderPostPagerActivity` was updated to handle this custom action the exact same way it would handle a normal deeplink action.
3. `DeepLinkingIntentReceiverActivity` was updated to check if a redirect URL could be handled by `ReaderPostPagerActivity`. To do that, we build a new `Intent` with the custom action + the redirect URL and check if it can be resolved by any Activity. If so, since the custom action is only used by `ReaderPostPagerActivity`, we can safely assume the URL can be opened in the Reader.

Let me know if think this approach is sound.

### To test

1. Get any supported post URL and wrap into the tracking URL. Here are a couple of examples:
    - https://public-api.wordpress.com/mbar/?stat=groovemails-events&bin=wpcom_email_click&redirect_to=https://discover.wordpress.com/2018/03/22/how-to-create-website-logo/
    - https://public-api.wordpress.com/mbar/?stat=groovemails-events&bin=wpcom_email_click&redirect_to=https://wordpress.com/read/blogs/53424024/posts/32898
1. Click the link.
1. Notice the corresponding post is opened in the app's Reader.
1. Notice logcat shows the following message: "DeepLink tracking URI successfully requested".

## Regression Notes
Probably none, but maybe Reader.

Extensive manual testing and existing unit tests.

Unit tests.

PR submission checklist:


planarvoid (commented):
The solution looks good and works well. I think for now it's the best approach.
In the future I'd love for us to consolidate all the deeplink handling into one activity.

I have one concern about the functionality though. The editor links are opened in a new stack and not over the main activity. This leads to the back button going back to the email. I think it would be better to build the reader over the Main activity with the Reader selected. This way we can keep the user in the app. I know this is the previous behaviour but I don't think it's the best one possible since our goal is to bring the users to the app. WDYT?

renanferrari:
Hey @planarvoid, thanks for the review! 

I've addressed your suggestion and now the post links are opened with the main activity in the backstack. This only applies for links wrapped with the `mbar` URL though. To change the existing behavior for plain post links, I believe we would need to move the intent filters from `ReaderPostPagerActivity` to another class. This is in line with the goal of consolidating all the deeplink handling into one activity, which I very much agree, but I think this should be done in a different PR, so I've opened an issue for it here: #14496. Let me know if you agree.

I have also merged the changes from `develop`, resolved all conflicts and added a `ReaderLinkHandler` for consistency with the other changes you've made. 

However, after merging those changes I found that one of my tests was failing. It was testing a scenario where we intercept a `mbar` URL but can't handle the redirect URL ([example](https://public-api.wordpress.com/mbar/?stat=groovemails-events&bin=wpcom_email_click&redirect_to=https://wordpress.com/nope/)). In this case, I'd expect for the browser to be opened and for the server tracking to not be triggered, which was not happening (the browser was being opened, but the server tracking was still being triggered).

Good news is that I've tested this with the improvements I mentioned I tinkered with in https://github.com/wordpress-mobile/WordPress-Android/pull/14466#discussion_r616954366 and they seem to solve the issue, so I've just disabled this test for now and submitted those improvements via #14497, so you can take a look at it.

Let me know what you think!

planarvoid (approved):
> I think this should be done in a different PR, so I've opened an issue for it here: #14496. Let me know if you agree.

this makes sense 👍 , thanks for creating the issue!

> Good news is that I've tested this with the improvements I mentioned I tinkered with in #14466 (comment) and they seem to solve the issue, so I've just disabled this test for now and submitted those improvements via #14497, so you can take a look at it.

thanks for handling this!

I've tested the changes and it works well 👍

PR #14477, opened 2021-04-16
Adding pagination to list of post and comment likers

develric:
Part of the #14137 . This PR adds pagination support for both post and comment likers list. Pagination is using a similar pattern that @planarvoid used for Media that I think should work well in this context as well.

Companion FluxC PR [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1965)
Depends on the D60325 endpoint differential (now merged).

<p align=center><image src="https://user-images.githubusercontent.com/47797566/115088327-373d7600-9f10-11eb-9036-7d54f7459199.gif" width=300/></p>

Note that pages are reverse ordered by date_added; also pagination works with 2 parameters sent to API:

- `before`: this contains the oldest date already parsed from the endpoint so that the endpoint knows likers before that date should be sent back. Note that the precision of this date is down to seconds.
- `exclude`: this is an array of user ids (at least one should be present for pages after the first one). Since precision of the `before` date is down to seconds, in the cases where more than liker liked at the same exact second, this reports to the endpoint all the already client known user ids belonging to that second time, so that they are not sent back in the response from the endpoint.

### To test
Note that currently page length is 20.

- Test list of likers with less than 20 items is consistent and rendered as expected
- Test list of likers with more than 20 items is consistent and rendered as expected. Note that in between the pages of length 20 there is a Loader progress element while the next page is recovered
- Test list of likers with more than 1 page (the more likers, the best). Check that no holes are in the list of likers.
- Test the `exclude` pagination pattern. To test this you can use the debugger to force date_added overlaps in the list; also Charles Proxy could in theory be setup to this goal but in case all seems too tricky, ping me and we can setup a test scenario against my sandbox 🙇 
- Smoke test user profile bottom sheet on multiple pages list 
- Retry case: on errors (for example when no network) a Retry button element should be presented instead of the loader progress element. Set airplane mode and swipe to the end of the list of likers. Check you see a retry button, tap on it and confirm nothing changes. Remove the airplane mode and tap again on the Retry button. This time the next page should load (please note that sometimes after removing the airplane some taps are needed until the network connection is actually restored)

## Regression Notes
Feature behind feature flag.

Feature behind feature flag.

Feature behind feature flag.

PR submission checklist:


khaykov (approved):
Works well, @develric ! I found no issues :) As we discussed, I skipped the `exclude` test for now, since the setup is pretty complicated, and we are going to cover it with tests.

PR #14481, opened 2021-04-19
Jetpack App (Basics): Update Toolbar Title Typography

### LINKED ISSUE #14443: Jetpack App (Basics): Update Typography
Parent: #14293

This task is based on `pcdRpT-gx-p2` where updating the large titles to `Sans Serif` is described as a bullet point under the `Diving into The Basics` section.

ashiagr:
Fixes #14443

This PR updates the toolbar title font family to `san-serif` in the Jetpack app. There are different toolbar styles for 
- collapsing toolbar (with expanded and collapsed toolbar title)
- default (non-collapsing) toolbar
- toolbar for Login and 
- toolbar for Zendesk

These toolbar styles are extracted to a separate `styles_toolbar.xml` and the toolbar's `titleTextAppearance` is overridden for the Jetpack app. 

Collapsing Toolbar Expanded
WordPress | Jetpack
------------|----------
wp_toolbar_expanded| jp_toolbar_expanded

Collapsing Toolbar Collapsed
WordPress | Jetpack
------------|----------
wp_toolbar_collapsed| jp_toolbar_collapsed

Default (non-collapsing)
WordPress | Jetpack
------------|----------
wp_toolbar_title| jp_toolbar_title

Login
WordPress | Jetpack
------------|----------
wp_login_toolbar_title | jp_login_toolbar_title

Zendesk
WordPress | Jetpack
------------|----------
wp_zendesk_toolbar_title | jp_zendesk_toolbar_title

Merge Instructions

- Design review by @osullivanchris on the typography of the title.
- Remove Needs Design Review label.
- Remove Not Ready for Merge label.
- Merge the PR.

## Regression Notes
N/A - impacts only toolbar styles.

Tested toolbars on different screens for both WordPress and Jetpack app. 

N/A

PR submission checklist:

### DIFF
--- a/WordPress/src/jetpack/res/values/styles_toolbar.xml
+++ b/WordPress/src/jetpack/res/values/styles_toolbar.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <style name="TextAppearance.App.Toolbar.Title" parent="TextAppearance.MaterialComponents.Headline6">
+        <item name="fontFamily">san-serif</item>
+        <item name="android:textStyle">bold</item>
+    </style>
+
+    <style name="TextAppearance.App.CollapsingToolbar.Expanded" parent="TextAppearance.MaterialComponents.Headline4">
+        <item name="android:textColor">?attr/colorOnSurface</item>
+        <item name="fontFamily">san-serif</item>
+        <item name="android:textStyle">bold</item>
+    </style>
+
+    <style name="TextAppearance.App.CollapsingToolbar.Collapsed" parent="TextAppearance.MaterialComponents.Headline6">
+        <item name="android:textColor">?attr/colorOnSurface</item>
+        <item name="fontFamily">san-serif</item>
+        <item name="android:textStyle">bold</item>
+    </style>
+
+</resources>
+
--- a/WordPress/src/jetpack/res/values/styles_toolbar.xml
+++ b/WordPress/src/jetpack/res/values/styles_toolbar.xml
@@ -2,21 +2,15 @@
 <resources>
 
     <!-- TextAppearance -->
-    <style name="TextAppearance.App.Toolbar.Title" parent="TextAppearance.MaterialComponents.Headline6">
-        <item name="fontFamily">san-serif</item>
-        <item name="android:textStyle">bold</item>
-    </style>
+    <style name="TextAppearance.App.Toolbar.Title" parent="TextAppearance.MaterialComponents.Headline6"/>
 
     <style name="TextAppearance.App.CollapsingToolbar.Expanded" parent="TextAppearance.MaterialComponents.Headline4">
         <item name="android:textColor">?attr/colorOnSurface</item>
-        <item name="fontFamily">san-serif</item>
         <item name="android:textStyle">bold</item>
     </style>
 
     <style name="TextAppearance.App.CollapsingToolbar.Collapsed" parent="TextAppearance.MaterialComponents.Headline6">
         <item name="android:textColor">?attr/colorOnSurface</item>
-        <item name="fontFamily">san-serif</item>
-        <item name="android:textStyle">bold</item>
     </style>
 
 </resources>
--- a/WordPress/src/jetpack/res/values/styles_toolbar.xml
+++ b/WordPress/src/jetpack/res/values/styles_toolbar.xml
@@ -1,12 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
 
-    <!-- Login -->
-    <style name="Widget.LoginFlow.Toolbar" parent="Widget.MaterialComponents.Toolbar.Surface">
-        <item name="titleTextAppearance">@style/TextAppearance.App.Toolbar.Title</item>
-        <item name="android:theme">@style/ThemeOverlay.LoginFlow.Toolbar</item>
-    </style>
-
     <!-- TextAppearance -->
     <style name="TextAppearance.App.Toolbar.Title" parent="TextAppearance.MaterialComponents.Headline6">
         <item name="fontFamily">san-serif</item>
--- a/WordPress/src/jetpack/res/values/styles_toolbar.xml
+++ b/WordPress/src/jetpack/res/values/styles_toolbar.xml
@@ -1,11 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
 
+    <!-- Login -->
     <style name="Widget.LoginFlow.Toolbar" parent="Widget.MaterialComponents.Toolbar.Surface">
         <item name="titleTextAppearance">@style/TextAppearance.App.Toolbar.Title</item>
         <item name="android:theme">@style/ThemeOverlay.LoginFlow.Toolbar</item>
     </style>
 
+    <!-- TextAppearance -->
     <style name="TextAppearance.App.Toolbar.Title" parent="TextAppearance.MaterialComponents.Headline6">
         <item name="fontFamily">san-serif</item>
         <item name="android:textStyle">bold</item>
--- a/WordPress/src/jetpack/res/values/styles_toolbar.xml
+++ b/WordPress/src/jetpack/res/values/styles_toolbar.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<resources xmlns:android="http://schemas.android.com/apk/res/android">
+<resources>
 
     <style name="Widget.LoginFlow.Toolbar" parent="Widget.MaterialComponents.Toolbar.Surface">
         <item name="titleTextAppearance">@style/TextAppearance.App.Toolbar.Title</item>
@@ -24,4 +24,3 @@
     </style>
 
 </resources>
-
--- a/WordPress/src/jetpack/res/values/styles_toolbar.xml
+++ b/WordPress/src/jetpack/res/values/styles_toolbar.xml
@@ -1,6 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:android="http://schemas.android.com/apk/res/android">
 
+    <style name="Widget.LoginFlow.Toolbar" parent="Widget.MaterialComponents.Toolbar.Surface">
+        <item name="titleTextAppearance">@style/TextAppearance.App.Toolbar.Title</item>
+        <item name="android:theme">@style/ThemeOverlay.LoginFlow.Toolbar</item>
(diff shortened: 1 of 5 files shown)


WordPress/src/jetpack/res/values/styles_toolbar.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <style name="Widget.LoginFlow.Toolbar" parent="Widget.MaterialComponents.Toolbar.Surface">
+        <item name="titleTextAppearance">@style/TextAppearance.App.Toolbar.Title</item>
```
ashiagr: I'm not sure if I should override the login toolbar style here or add a `titleTextAppearance` (which doesn't exist right now) for the style in the login module. I'd appreciate any inputs on this 🙏.

WordPress/src/jetpack/res/values/styles_toolbar.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:android="http://schemas.android.com/apk/res/android">
```
ParaskP7: Minor (🔍): This `xmlns:android` namespace is not necessary for now. Consider removing it.

PS: Same apply for the `main` such file.

WordPress/src/jetpack/res/values/styles_toolbar.xml
ParaskP7: I think that for now what you did is okay and I would have done the same. Having said that I have a additional suggestion.

Suggestion (💡): Consider moving this style away from the `jetpack` to into the `main` `styles_toolbar.xml`. This way it will become more explicit that this `TextAppearance.App.Toolbar.Title` will be used for both the `WordPress` and `Jetpack` apps, just like it currently happens on the `WordPress.ToolBa` style. Wdyt?

WordPress/src/main/res/values-night/styles.xml
```diff
@@ -138,11 +138,6 @@
         <item name="android:navigationBarColor">@android:color/black</item>
     </style>
 
-    <style name="WordPress.ZenDesk.Toolbar" parent="Widget.MaterialComponents.Toolbar.Surface">
```
ParaskP7, line 141: Praise (❤️): I really love what you did there, extracting the all the `styles_toolbar.xml` files. Doing that makes things so much more structured. Kudos!

WordPress/src/main/res/values/styles_toolbar.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <style name="WordPress.ToolBar" parent="Widget.MaterialComponents.Toolbar.Surface">
```
ParaskP7, line 11: Suggestion (💡): What about further enhancing the structure of these file with extract block comments? For example, if you apply my above suggestions, within this file we can end up with a structure like the below:

```
<?xml version="1.0" encoding="utf-8"?>
<resources>

    <!-- Login -->

    ...

    <!-- App -->

    ...

    <!-- ZenDesk -->

    ...

    <!-- TextAppearance -->

    ...

</resources>
```

PS: Same apply to all such files.

WordPress/src/jetpack/res/values/styles_toolbar.xml
```diff
@@ -0,0 +1,27 @@
...
+    </style>
+
+    <style name="TextAppearance.App.CollapsingToolbar.Collapsed" parent="TextAppearance.MaterialComponents.Headline6">
+        <item name="android:textColor">?attr/colorOnSurface</item>
+        <item name="fontFamily">san-serif</item>
+        <item name="android:textStyle">bold</item>
+    </style>
+
+</resources>
+
```
ParaskP7: Minor (🔍): There exist an extra empty line here that can be potentially removed.

ParaskP7 (approved):
👋 @ashiagr !

I have reviewed and tested this PR as per the instructions, everything works as expected, great job! 🌟 🌟 

I have left some minor (🔍) comments and few suggestions (💡) for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself after you get approval from @osullivanchris as well.

WordPress/src/jetpack/res/values/styles_toolbar.xml
ashiagr: In 0cc8acd

WordPress/src/main/res/values/styles_toolbar.xml
ashiagr: In 6ad5f16

WordPress/src/jetpack/res/values/styles_toolbar.xml
ashiagr: In 0cc8acd, 3cbf8c4

ashiagr:
Thank you for the suggestions @ParaskP7! Included all of them 🙂.
Ready for another look.

ParaskP7:
👋 @ashiagr !

> Thank you for the suggestions @ParaskP7! Included all of them 🙂.
> Ready for another look.

Everything LGTM, thank you so much for taking care of all that! 🚀 

PS: Feel free to merge when CI completes and you get the 🟢 light from @osullivanchris !

osullivanchris (changes requested):
@ashiagr  thanks for the ping. Looks good so far. I have a couple of related thoughts:

- In my mind, there should be a 'default' app bar header provided by the system. I would have then assumed we customised this header in WordPress to achieve the serif font. So I'm wondering whether in Jetpack its a case of removing that customisation, returning to the default sans serif? Or is that the wrong mental model? Just wondering if we can steer close to the system rather than adding 2 steps of customisation. Maybe you have already done it this way 😃 
- From what I understand, referring to [the design guides](https://material.io/design/typography/the-type-system.html#type-scale), the default app bar header is 20dp medium. I used 20dp medium as well. In the screenshots you provided, it looks like 20dp bold. 
- However, you have used 20dp bold in the large titles, and that looks correct. So my question is if its possible to shift from bold to medium when the title shrinks? That's why I had in the mocks. But If that is crazy, we can stick to **bold** for both like you currently have.
- Noting that I also don't know how much comes 'out of the box' with the large titles, or if they are fully custom. Just trying to learn how it works while we discuss!

ashiagr:
Thanks for the feedback and questions @osullivanchris 🙇‍♀️! I've tried to answer them below: 

> In my mind, there should be a 'default' app bar header provided by the system. I would have then assumed we customised this header in WordPress to achieve the serif font. 

Correct, the default toolbar title uses [TextAppearance.MaterialComponents.Headline6](https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/typography/res/values-v21/styles.xml#L19) style which is applied to the toolbar's  `titleTextAppearance` property. In the WordPress app, we [override this style](https://github.com/wordpress-mobile/WordPress-Android/blob/9356d67b59d385911e50a87b200b6716fbff69e2/WordPress/src/main/res/values/styles_toolbar.xml#L41-L44) to achieve the `serif` font and `bold` text style. 

> So I'm wondering whether in Jetpack its a case of removing that customisation, returning to the default sans serif? Or is that the wrong mental model? Just wondering if we can steer close to the system rather than adding 2 steps of customisation. Maybe you have already done it this way 😃

> From what I understand, referring to the design guides, the default app bar header is 20dp medium. I used 20dp medium as well. In the screenshots you provided, it looks like 20dp bold.

For the Jetpack app,  in order to remove the customization coming from the WordPress app, we need to override that style again. In 9356d67 I've updated it to just use the default `Headline6` style (that internally uses the `san-serif-medium` `normal` font) for the [default](https://github.com/wordpress-mobile/WordPress-Android/blob/9356d67b59d385911e50a87b200b6716fbff69e2/WordPress/src/jetpack/res/values/styles_toolbar.xml#L5) and [shrunk](https://github.com/wordpress-mobile/WordPress-Android/blob/9356d67b59d385911e50a87b200b6716fbff69e2/WordPress/src/jetpack/res/values/styles_toolbar.xml#L12) toolbar titles. 
Before (san-serif, bold) | After (san-serif-medium, normal)
-------|-----
bold | normal

This should now remove the `bold` effect that we were seeing earlier.

> However, you have used 20dp bold in the large titles, and that looks correct. So my question is if its possible to shift from bold to medium when the title shrinks? That's why I had in the mocks. But If that is crazy, we can stick to bold for both like you currently have.

We can have different styles for the expanded (large) and the collapsed (shrunk) toolbar titles. I have [retained the bold style for the large titles](https://github.com/wordpress-mobile/WordPress-Android/blob/9356d67b59d385911e50a87b200b6716fbff69e2/WordPress/src/jetpack/res/values/styles_toolbar.xml#L9).

> Noting that I also don't know how much comes 'out of the box' with the large titles, or if they are fully custom. Just trying to learn how it works while we discuss!

Large titles use the [TextAppearance.MaterialComponents.Headline4](https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/typography/res/values/styles.xml#L46) style. We currently override it to add a `bold` text style. We can customize other properties as well and the customization can be done separately for the large titles and the shrunk ones. Hope this clarifies 🙂.  Please feel free to post any more questions you might have regarding this 🙏.

osullivanchris:
> Correct, the default toolbar title uses TextAppearance.MaterialComponents.Headline6 style which is applied to the toolbar's titleTextAppearance property. In the WordPress app, we override this style to achieve the serif font and bold text style.

🙌 thanks for explaining. That's exactly how I hoped it would work!

> For the Jetpack app, in order to remove the customization coming from the WordPress app, we need to override that style again. In 9356d67 I've updated it to just use the default Headline6 style (that internally uses the san-serif-medium normal font) for the default and shrunk toolbar titles.

Perfect.

> We can have different styles for the expanded (large) and the collapsed (shrunk) toolbar titles. I have retained the bold style for the large titles.

Yes! that's great. So glad we can have both. 🔨  Nailed it!

> Large titles use the TextAppearance.MaterialComponents.Headline4 style. We currently override it to add a bold text style. We can customize other properties as well and the customization can be done separately for the large titles and the shrunk ones. Hope this clarifies 🙂. Please feel free to post any more questions you might have regarding this 🙏.

Totally clarifies. Glad most of what we are doing here steers so close to the system. I think the bold on the large titles is worth having, as it looks better. But otherwise its great that we are doing so much out of the box. And glad to understand a bit better too!

This has my 👍  now. Thanks for all that!

PR #14491, opened 2021-04-20
Jetpack App (Basics): Update Zendesk ticket subject

### LINKED ISSUE #14314: Jetpack App (Basics): Configure Zendesk
Parent: #14293

Change the subject from "WordPress for Android Support" to "Jetpack for Android Support" (the latter without internationalization).

### To test

1. Contact the support from the Jetpack app and enter your message
2. Check that the ticket was sent with the "Jetpack for Android Support" subject'
3. Contact the support from WP app
4. Check that the ticket was sent with "WordPress for Android Support" subject

ashiagr:
Fixes #14314

To test:

WordPress app 

1. Put a breakpoint inside [ZendeskHelper. buildZendeskConfig](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/support/ZendeskHelper.kt#L390) where the Zendesk ticket subject is set.
2. Install and launch the WordPress app
3. Go to My Site -> Me -> Help & Support -> Contact Support
4. Notice that the `ticketSubject` is set to "WordPress for Android Support".

Jetpack app 

1. Repeat above steps 1- 3 for the Jetpack app
2. Notice that the `ticketSubject` is set to "Jetpack for Android Support".

Note: Since the subject is not visible to the app user, the text is not localized (similar to the behaviour earlier), and `translatable` is set to `false` in the strings.xml with a note. Then the string is overridden in the Jetpack app. We can also keep it as a constant in the `BuildConfig`, I can't think of any reason for one approach being better than the other and open to suggestions. 

## Regression Notes
N/A

N/A

N/A

PR submission checklist:

### DIFF
--- a/WordPress/src/jetpack/res/values/strings.xml
+++ b/WordPress/src/jetpack/res/values/strings.xml
@@ -5,4 +5,8 @@
 
     <!-- About View -->
     <string name="app_title">Jetpack for Android</string>
+
+    <!-- Contact us -->
+    <!--Note: Support ticket subject, only visible to the end users of the Zendesk support portal, doesn't need a translation -->
+    <string name="support_ticket_subject" translatable="false">Jetpack for Android Support</string>
 </resources>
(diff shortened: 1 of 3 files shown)


zwarm (approved):
@ashiagr - All works as expected. I also tested the Help & Support feature on the login email address view. 👍 .

As for `strings.xml` vs `BuildConfig` - I like the `strings` solution. It is a string and we're not making logic decision based on the value. Keeping it out of BuildConfig feels right.

PR #14494, opened 2021-04-20
Remove steps to install buldletool and install new jdk

JavonDavis:
Per discussion in Slack(p1618410985357000), we shouldn't need to install bundletool and the openjdk anymore and we can save ourselves some time in CI by not doing this. This PR removes that step in CI for the release build job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -172,16 +172,16 @@ jobs:
           name: Build Zalpha
           command: |
             if [[ ${APP_VERSION} == *"-rc-"* ]]; then
-              bundle exec fastlane build_alpha skip_confirm:true skip_prechecks:true create_release:true upload_to_play_store:true
+              bundle exec fastlane build_alpha skip_confirm:true skip_prechecks:true create_release:false upload_to_play_store:false
             fi
           no_output_timeout: 15m
       - run:
           name: Build Vanilla
           command: |
             if [[ ${APP_VERSION} == *"-rc-"* ]]; then
-              bundle exec fastlane build_beta skip_confirm:true skip_prechecks:true create_release:true upload_to_play_store:true
+              bundle exec fastlane build_beta skip_confirm:true skip_prechecks:true create_release:false upload_to_play_store:false
             else
-              bundle exec fastlane build_and_upload_release skip_confirm:true skip_prechecks:true create_release:true upload_to_play_store:true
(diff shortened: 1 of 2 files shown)


AliSoftware (changes requested):
Tested by:

 - Adding a temporary commit 3bb2b8c to temporarily disable the creation of GH release and upload to PlayStore on Release Builds, which I planned to revert after the tests
 - Then triggered a release build using `bundle exec fastlane trigger_release_build branch_to_build:update/ci_install_other_tools`

The CI [failed (see here)](https://app.circleci.com/pipelines/github/wordpress-mobile/WordPress-Android/23209/workflows/9778e0f4-dde6-4c01-acb9-ca93072cf13c/jobs/103897) while trying to run `bundletool` – so it seems we still have places where it's invoked.

I still think we don't need those invocations to `bundletool`, so the solution is to just remove them.

Note: I kept the 3bb2b8c temporary commit in for now so that you can test as you fix the PR. We should just be sure not to forget to revert that temporary commit before merging this PR.

.circleci/config.yml
```diff
@@ -180,16 +172,16 @@ jobs:
           name: Build Zalpha
           command: |
             if [[ ${APP_VERSION} == *"-rc-"* ]]; then
-              bundle exec fastlane build_alpha skip_confirm:true skip_prechecks:true create_release:true upload_to_play_store:true
+              bundle exec fastlane build_alpha skip_confirm:true skip_prechecks:true create_release:false upload_to_play_store:false
```
AliSoftware: Introduced via temporary commit 3bb2b8c4a797d6cdc1a01863f6b4dc9e236fa5e2, which I made in order to be able to test the PR by trigger a Release Build… without creating a GitHub Release nor a PlayStore entry.

🚨 So just a reminder that this commit needs to be reverted once we've tested everything (and ensured `fastlane trigger_release_build` succeeds) before we finally merge this PR.

JavonDavis:
Actually thought we had removed the invocations prior to this and just needed to remove the installation! Thanks for testing this out @AliSoftware  🙇

AliSoftware (approved):
Latest changes looks good so approving. 👍 

But let's wait for the application of that same change on the release branch then testing this with the creation of a real beta (since we need one anyway) – to ensure it works in the real case before merging this one, just in case.

JavonDavis:
Merging and closing since this worked well in the real case in the 17.2 Beta 3

PR #6196, opened 2017-06-30
Login Rework: prologue screen visual touches

hypest:
Finishes up the visual and textual touches for the login prologue screen.

* 2 loops per animation then auto-advance
* Using Material Design button style

device

To test:
* With the app data cleaned up, run the app (make sure it's a `wasabi` build variant)
* Observe the prologue screen. Notice the auto-advancing of the view-pager as well. All 5 animation screens should perform OK.

PR #14503, opened 2021-04-21
Add a "Set as Featured Image" Button to Image Block

SiobhyB:
Partial fix for: https://github.com/wordpress-mobile/gutenberg-mobile/issues/1011

`gutenberg-mobile`: wordpress-mobile/gutenberg-mobile#3116
`gutenberg`: https://github.com/WordPress/gutenberg/pull/31705

## Description

This PR adds a **Set as Featured** button to the image block's settings, with the purpose being to make it simpler for users to set a featured image within the post's editor. Users will also be able to **Remove as Featured** directly from the block's settings. 

A couple of notable changes to the native code have been highlighted below, with further details around the JS side highlighted in [the Gutenberg PR](https://github.com/WordPress/gutenberg/pull/31705). That PR can also be referred to as the central place for up-to-date screenshots and testing flows.

### OnSetFeaturedImageListener

[OnSetFeaturedImageListener](https://github.com/wordpress-mobile/WordPress-Android/pull/14503/files#diff-12a344ad8965875cb092fcc0c32a77e4909d9e5cb7092fbaf0d9ddc7d6c71d56R344
) listens for requests to set a featured image via an image block that are passed over the bridge. Please refer to [the Gutenberg PR ](https://github.com/WordPress/gutenberg/pull/31705)for more details of how this implemented from the JS side.

When a request is received from the native side, `onSetFeaturedImageButtonClicked()` [checks for whether a featured image has already been set](https://github.com/wordpress-mobile/WordPress-Android/pull/14503/files#diff-12a344ad8965875cb092fcc0c32a77e4909d9e5cb7092fbaf0d9ddc7d6c71d56R346), in which case `showFeaturedImageConfirmationDialog()` is called and users are able to confirm their wish to replace the featured image or not. If there isn't already an existing featured image or the user is removing an image as featured, then [the `setFeaturedImage()` function](https://github.com/wordpress-mobile/WordPress-Android/pull/14503/files#diff-12a344ad8965875cb092fcc0c32a77e4909d9e5cb7092fbaf0d9ddc7d6c71d56R809) is called directly.

_**Note: The code for setting a featured image has been iterated on as part of https://github.com/wordpress-mobile/WordPress-Android/pull/14521, with tracks events being added as part of that PR also. It'll be merged separately as an enhancement to this one.**_

### GutenbergDialogFragment

A [new `GutenbergDialogFragment` class](https://github.com/wordpress-mobile/WordPress-Android/pull/14503/files#diff-8889a43bfa80f8522889ae2f69695ac5eda5b29189bb29c9915a9c4825dae6e7) has been created as part of this PR. This class is based on [the `BasicFragmentDialog` class](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/posts/BasicFragmentDialog.kt), with some key differences being the removal of the third `setNeutralButton`, [an addition of a mediaId](https://github.com/wordpress-mobile/WordPress-Android/pull/14503/files#diff-8889a43bfa80f8522889ae2f69695ac5eda5b29189bb29c9915a9c4825dae6e7R39) to the constructor, and the ability for [the parent fragment to implement the positive/negative button interfaces](https://github.com/wordpress-mobile/WordPress-Android/pull/14503/files#diff-8889a43bfa80f8522889ae2f69695ac5eda5b29189bb29c9915a9c4825dae6e7R86). 

`GutenbergDialogFragment` [is called from the `showFeaturedImageConfirmationDialog()` function](https://github.com/wordpress-mobile/WordPress-Android/pull/14503/files#diff-12a344ad8965875cb092fcc0c32a77e4909d9e5cb7092fbaf0d9ddc7d6c71d56R795), to ensure that the dialog survives screen rotation. 

The **mediaId** associated with the image is [passed down from the `onSetFeaturedImageButtonClicked()` function](https://github.com/wordpress-mobile/WordPress-Android/pull/14503/files#diff-12a344ad8965875cb092fcc0c32a77e4909d9e5cb7092fbaf0d9ddc7d6c71d56R346) (which in turn retrieves it from the JS side).  

The functionality for when a user chooses to **replace** a featured image is defined in [the `onGutenbergDialogPositiveClicked()` function](https://github.com/wordpress-mobile/WordPress-Android/pull/14503/files#diff-12a344ad8965875cb092fcc0c32a77e4909d9e5cb7092fbaf0d9ddc7d6c71d56R1384) and the dialog is dismissed when the **cancel** button is clicked, as per [the `onGutenbergDialogNegativeClicked` function](https://github.com/wordpress-mobile/WordPress-Android/pull/14503/files#diff-12a344ad8965875cb092fcc0c32a77e4909d9e5cb7092fbaf0d9ddc7d6c71d56R1393).

As outlined in https://github.com/wordpress-mobile/WordPress-Android/issues/14676, there are other dialogs in the GutenbergEditorFragment file that are lost/dismissed with screen rotation. Part of the goal behind the new `GutenbergDialogFragment` class is to allow for easier refactoring of those other Dialogs, beginning with `showCancelMediaUploadDialog` and `showRetryMediaUploadDialog`, as those functions also accept a medaId.

## Testing

Please refer [to the Gutenberg PR](https://github.com/WordPress/gutenberg/pull/31705) as the central place for steps to tests and screenshots. Of note for the Android side, testing whether the **replace** dialog works as expected when the screen is rotated will be helpful.

## Regression Notes

The main unintended issues would be around the incorrect featured image being set/removed or other unintended/buggy issues with featured images.

The flows outlined in [the Gutenberg PR](https://github.com/WordPress/gutenberg/pull/31705) were tested.

N/A

<hr />

PR submission checklist:


SiobhyB:
Status update: I'm currently investigating an issue where `mFeaturedImageId` is reset to zero when following these steps:

1. Open a new post that already has a featured image, or set a featured image on the post if it doesn't already have one.
2. Select a different image block, tap the cog/gear icon, and then **Set as Featured** image.
3. When the **replace** dialog pops up, switch the screen's orientation before tapping the **replace** option.
4. Confirm the new image has been set as featured by navigating to _Post Settings_ and observing the **Featured** banner that overlays it within the editor.
5. Switch the orientation of the screen again. Select a different image block, tap the cog/gear icon, and then **Set as Featured** image.
6. Notice how the new image is automatically set as featured, with no _replace_ dialog displaying.

Even though the correct images are being set as featured in the above flow, the replace dialog should still display at step number 6. _Log.d_ shows that `mFeaturedImageId` is being reset to zero in this scenario, which explains why the dialog isn't popping up. The images themselves are being set as featured, so this doesn't seem like a major bug and more of an edge case, but would still be nice for the behaviour to be consistent. I'll keep investigating. 🕵️‍♀️

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergDialogFragment.kt
```diff
@@ -0,0 +1,129 @@
...
+import android.app.Dialog
+import android.content.Context
+import android.content.DialogInterface
+import android.os.Bundle
+import androidx.appcompat.app.AppCompatDialogFragment
+import androidx.fragment.app.Fragment
+import com.google.android.material.dialog.MaterialAlertDialogBuilder
+
+class GutenbergDialogFragment() : AppCompatDialogFragment() {
+    private lateinit var mTag: String
```
hypest, line 12: 👋 @SiobhyB , only took a quick glance here, and since I know you've taken inspiration from the BasicDialog implementation I wonder if you can elaborate on the choice of having these dialog "properties" (Tag, message, etc) be members of the Fragment while [they are part of a model class over at the BasicDialog implementation](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/posts/BasicDialogViewModel.kt#L38-L45). Not saying there's an issue with this, haven't really checked deep enough yet, just wondering what's the context/insight on this. Thanks!

SiobhyB: Ah, I used [BasicFragmentDialog.kt](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/posts/BasicFragmentDialog.kt) as the template, which implements the "properties" in this way. I've gone ahead to edit the commit message that incorrectly referenced BasicDialog.kt rather than BasicFragmentDialog.kt, sorry for the confusion!

hypest: Ahh, I see, thanks for the update!

SiobhyB:
[This bug](https://github.com/wordpress-mobile/WordPress-Android/pull/14503#issuecomment-844950869) has been fixed with the following commit: https://github.com/wordpress-mobile/WordPress-Android/pull/14503/commits/f57eeaf8703a2ef816cc98949a54b99ab6432584

I've asked for some help on Slack regarding the failing test: p1621524064033200-slack-C02QANACA

Edited to add: The issue causing the failing test has been resolved now. 🎉

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergDialogFragment.kt
```diff
@@ -0,0 +1,139 @@
...
+        builder.setMessage(mMessage)
+
+        mTitle?.let {
+            builder.setTitle(mTitle)
+        }
+
+        mPositiveButtonLabel?.let {
+            builder.setPositiveButton(mPositiveButtonLabel) { _, _ ->
+                dismissedByPositiveButton = true
+                val parentFragment: Fragment? = parentFragment
```
hypest: Hey Siobhan, I know this pattern of having a same-name local const picking up the value from a getter (`getParentFragment()` in this case) [is used in similar places](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/posts/BasicFragmentDialog.kt#L83-L86), but makes me a bit uncomfortable when reading the code. Trusted, in AndroidStudio with its syntax highlighting it's kinda easy to see the difference between the two, but it's not super intuitive otherwise.

I was thinking, would you be interested in using the following to do the same null-check and casting directly, without the temporary val?

```kotlin
               (parentFragment as? GutenbergDialogPositiveClickInterface)?.let {
                    it.onGutenbergDialogPositiveClicked(mTag, mMediaId);
                }
```

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
```diff
@@ -333,6 +343,16 @@ public void onQueryCurrentProgressForUploadingMedia() {
...
                 },
+                new OnSetFeaturedImageListener() {
+                    @Override
+                    public void onSetFeaturedImageButtonClicked(int mediaId) {
+                        if (mediaId == 0 || mFeaturedImageId == 0) {
+                            setFeaturedImage(mediaId);
+                        } else if (mFeaturedImageId != mediaId) {
+                            showFeaturedImageConfirmationDialog(mediaId);
+                        }
+                    }
```
hypest, line 370: Looks like the function only implicitly covers the case where the IDs are non-zero but equal. I know that that case is practically a no-op case (we don't want anything to happen if we're trying to set the same image as the current featured one) but it's best if we leave a comment for that to make it clear. Something like `// don't do anything if mediaId is the same as the current featuredImageId`.

Alternatively, and a bit of a nit, what do you think about restructuring the code in such a way that it's more clear what's going on with the special values (`0`) and the various combinations, like this:

```kotlin
public void onSetFeaturedImageButtonClicked(int mediaId) {
    if (mediaId == mFeaturedImageId) {
        // nothing special to do, trying to set the image that's already set as featured
        return;
    }

    if (mediaId == 0) {
        // user tries to clear the featured image setting
        setFeaturedImage(mediaId);
        return;
    }

    if (mFeaturedImageId == 0) {
        // current featured image is not set so, go ahead and set it to the provided one
        setFeaturedImage(mediaId);
        return;
    }

    // ask the user to confirm changing the featured image since there's already one set
    showFeaturedImageConfirmationDialog(mediaId);
}
```

It's of course more verbose, but that's actually part of the goal, to use verbosity for the sake of readability and simplifying understanding the combinations.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
```diff
@@ -774,6 +794,33 @@ public void onClick(DialogInterface dialog, int id) {
...
+        );
+
+        dialog.show(getChildFragmentManager(), TAG_REPLACE_FEATURED_DIALOG);
+    }
+
+    private void setFeaturedImage(int mediaId) {
+        mEditorFragmentListener.updateFeaturedImage(mediaId, false);
+        // Necessary to update image in Post Settings
+        setFeaturedImageId(mediaId);
+        sendToJSFeaturedImageId(mediaId);
```
hypest: Hmm, do we need the explicit `sendToJSFeaturedImageId()` call here? Not sure if I'm missing some codepath but looks like the `updateFeaturedImage()` two lines up is including a call to send the data to the JS side? 🤔

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
```diff
@@ -774,6 +794,33 @@ public void onClick(DialogInterface dialog, int id) {
...
+                mediaId
+        );
+
+        dialog.show(getChildFragmentManager(), TAG_REPLACE_FEATURED_DIALOG);
+    }
+
+    private void setFeaturedImage(int mediaId) {
+        mEditorFragmentListener.updateFeaturedImage(mediaId, false);
+        // Necessary to update image in Post Settings
+        setFeaturedImageId(mediaId);
```
hypest, line 829: 👋 Siobhan! Assuming the comment (`// Necessary to update image in the Post Settings`) is about the `setFeaturedImageId()` call, can you elaborate on the necessasity? I'm trying to follow the code and looks to me that [the member variable](https://github.com/wordpress-mobile/WordPress-Android/blob/74323e97c25ac24447dce5030f01f84f900eaf6b/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragmentAbstract.java#L143-L145) doesn't play a huge role and doesn't seem to be involved [when saving the data](https://github.com/wordpress-mobile/WordPress-Android/blob/332ca7f48d3142fae0267beecf0212166adc8fc6/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostSettingsFragment.java#L999-L1020) 🤔. Maybe I'm missing a use?

libs/editor/WordPressEditor/src/main/res/values/strings.xml
```diff
@@ -154,6 +154,13 @@
     <string name="dialog_body_try_block_editor" translatable="true">We’ll be removing the classic editor for new posts soon, but this won’t affect editing any of your existing posts or pages. Get a head start by enabling the Block Editor now in site settings.</string>
     <string name="dialog_dismiss_try_block_editor" translatable="true">Dismiss</string>
 
-
+    <!-- Featured image replacement -->
+    <string name="replace_current_title">Replace current?</string>
```
hypest: Wdyt about tweaking the title to not include an implied reference to the body (`current`)? I feel it might be tougher to translate. Some ideas: "Confirm replace?" or just "Replace" even?

While at it, maybe all the string names could be prefixed with `featured_image_` to provide context?

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergDialogFragment.kt
SiobhyB: Thank you for this, it led me down a road of learning [a bit more about Kotlin](https://discuss.kotlinlang.org/t/let-vs-if-not-null/3542/11) and I've gone ahead to implement your suggestion. :)

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
SiobhyB: Implemented!

> It's of course more verbose, but that's actually part of the goal, to use verbosity for the sake of readability and simplifying understanding the combinations.

This is useful insight, thank you! I think knowing when to favour verbosity is something I'm going to need to work at.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
SiobhyB: Aha, yes, I missed that. 🤦‍♀️ Thank you for catching! I've removed it in the latest commit.

libs/editor/WordPressEditor/src/main/res/values/strings.xml
SiobhyB: I agree that, upon reflection, the names here a bit confusing! I've made some changes so that they're now prefixed with `featured_image_replace_dialog_`, following the lead of some of the other strings for dialogs in the `strings.xml` file. Do you think these new names seem okay?

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergDialogFragment.kt
jd-alexander: @SiobhyB I have taken a look at the `BasicFragmentDialog.kt` you mentioned as your point of reference and compared it to the `GutenbergDialogFragment` and the main difference I saw was the inclusion of the `mediaId`. I am thinking we could reuse `BasicFragmentDialog.kt` and track the `mediaId` within the `GutenbergDialogFragment` itself. Wdyt?

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
```diff
@@ -333,6 +343,30 @@ public void onQueryCurrentProgressForUploadingMedia() {
...
                 },
+                new OnSetFeaturedImageListener() {
+                    @Override
+                    public void onSetFeaturedImageButtonClicked(int mediaId) {
+                        if (mediaId == mFeaturedImageId) {
+                            // nothing special to do, trying to set the image that's already set as featured
+                            return;
+                        }
+
+                        if (mediaId == 0) {
```
jd-alexander: I am thinking we could hardcode this `0` which serves as the `DEFAULT_MEDIA_ID` so it's easier to identify its purpose and if we need to change its value it is in a reusable state.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
```diff
@@ -333,6 +343,30 @@ public void onQueryCurrentProgressForUploadingMedia() {
...
+                            // user tries to clear the featured image setting
+                            setFeaturedImage(mediaId);
+                            return;
+                        }
+
+                        if (mFeaturedImageId == 0) {
+                            // current featured image is not set so, go ahead and set it to the provided one
+                            setFeaturedImage(mediaId);
```
jd-alexander, line 358: Within `EditorFragmentAbstract` `mFeaturedImageSupported` is being persisted via the `savedInstanceState` bundle. We may need to persist `mFeaturedImageId`. Have you tried rotating the device while a `mFeaturedImageId` is set, attempt to set the same media id, and see if the function returns? Let me know what that is like. I will do another test of this PR tomorrow.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
jd-alexander: Hi @hypest to add to the conversation here, I can see where setting member variable would be useful if it's being persisted across configuration changes. 

For the "Necessary to update image in the Post Settings" comment, this will be removed once https://github.com/wordpress-mobile/WordPress-Android/pull/14521 is merged into this one. It is now under review.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
hypest: > I think knowing when to favour verbosity is something I'm going to need to work at.

Thanks for looking at this that way! And you are 100% correct with "when to favour" as it's always a balance or compromise between too verbose and too compacted without it being clear usually which side one should lean to 😬

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
hypest: > I've removed it in the latest commit

May I offer a suggestion (preference?) for how to link and close "the loop"? I personally add messages like "Fixed with <this commit hash link>" ([example here](https://github.com/wordpress-mobile/WordPress-Android/pull/6196#discussion_r127898836)) to save everyone a couple of seconds from scrolling to find the commit in question. I actually do that in all threads that end up in code changes to help focus the PR review. Of course, the reviewer will probably have a look at the overall diff as well but often times it helps to quickly check the commit that addresses a particular feedback and move on, until another overall pass.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergDialogFragment.kt
hypest: Ah, nice! While at it, wdyt about doing the same treatment to [this null-check and cast case in `onDismiss()`](https://github.com/wordpress-mobile/WordPress-Android/pull/14503/files#diff-8889a43bfa80f8522889ae2f69695ac5eda5b29189bb29c9915a9c4825dae6e7R116-R122) too?

libs/editor/WordPressEditor/src/main/res/values/strings.xml
hypest: > Do you think these new names seem okay?

Yeap, names look good to me!

Just making sure, do you have thoughts about rephrasing the string as well?

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
hypest: Good point about naming the special value. In this case, I guess it's probably closer to a `MEDIA_ID_NO_FEATURED_IMAGE_SET` that denotes the special interpretation of the value, right?

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
hypest: > would be useful if it's being persisted across configuration changes.

Hmm, the comment would still need some rephrasing then since it currently mentions "Post Settings" which makes me think of persisting in the local device prefs or the DB, which is different than preserving it on configuration changes (rotations for example). In any case, I feel that either the code doesn't correspond to the comment anymore, or vice versa 🙂

SiobhyB: It looks like the original bug referenced in the comment is no longer reproducible, but the functionality's still breaking if I comment out the call to `setFeaturedImageId`. I was able to reproduce the following with that line commented out:

* Toggle between **Set as Featured** and **Remove as Featured** on any image block, or between more than one image block.
* Notice how sometimes the button is unresponsive or the **replace** dialog pops up when there isn't actually a current featured image.

From logging, I see that `mFeaturedImageId` isn't always automatically updated when going through that flow. I'm not exactly sure if that's okay or expected. Is there another approach for updating that variable that should/could be taken here?

For now, I've updated the `Necessary to update image in Post Settings` to `Ensure "mFeaturedImageId" member variable is updated to specified "mediaId"`, in order to try to be clearer. Looking forward to hearing any extra thoughts!

SiobhyB: @hypest, we must have been drafting our comments at the same time and I missed yours until after I clicked published. :) You're right that the comment is now outdated, interested to hear what you think of the updated phrasing.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergDialogFragment.kt
SiobhyB: @jd-alexander, the new class also removes the third `setNeutralButton` and adds [the ability for the parent fragment to implement the positive/negative button interfaces](https://github.com/wordpress-mobile/WordPress-Android/pull/14503/files#diff-8889a43bfa80f8522889ae2f69695ac5eda5b29189bb29c9915a9c4825dae6e7R86) (rather than [the host activity](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/ui/posts/BasicFragmentDialog.kt#L83), which is the default with `BasicFragmentDialog.kt` and would require changes to `EditPostActivity` in our case). Would you still lean towards re-using `BasicFragmentDialog.kt` over a new class with those additional changes considered?

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
SiobhyB: Noted for the future! Thank you so much, I appreciate tips around etiquette like that and see how that small change would make life a lot easier for the review. 🙇‍♀️

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
SiobhyB: @jd-alexander, `mFeaturedImageId` was added to `savedInstanceState` within `GutenbergEditorFragment` in https://github.com/wordpress-mobile/WordPress-Android/pull/14503/commits/f57eeaf8703a2ef816cc98949a54b99ab6432584, which I believe fixes any issues with it being reset with screen rotation. Would it be better practice to do this in `EditorFragmentAbstract` rather than `GutenbergEditorFragment`?

libs/editor/WordPressEditor/src/main/res/values/strings.xml
SiobhyB: Ah, yes, I agree with the reasoning for rephrasing. I went with `Replace?` in https://github.com/wordpress-mobile/WordPress-Android/pull/14503/commits/b4dc8252d0706ec223a38614546cffbb94d5c292, as that seemed to flow best in my mind, let me know if you think anything else is needed here.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
hypest: > `Ensure "mFeaturedImageId" member variable is updated to specified "mediaId"`

Hmm, I guess the comment now is more accurate, but unfortunately ended up not adding any context. I mean, the `setFeaturedImageId(mediaId)` itself is nothing more than a classic (Java) setter so it is in fact already conveying the message "let's set the `FeaturedImageId` to `mediaId`. With a line comment, we are seeking to add more info that is not obvious just by looking at the code. So in this case, I'd expect a comment that would answer the question: "**Why** are we setting `mFeaturedImageId` to `mediaId`?"

I understand that perhaps we are not quite sure yet, and I guess that's what we need to do, to make sure we understand the code and the need. Let's continue reading and researching the code to see what's going and. That should also help answer your other question:
> "From logging, I see that mFeaturedImageId isn't always automatically updated when going through that flow. I'm not exactly sure if that's okay or expected. Is there another approach for updating that variable that should/could be taken here?".

libs/editor/WordPressEditor/src/main/res/values/strings.xml
hypest: Yeah, "Replace?" works for me too, thanks!

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
SiobhyB: That makes a lot of sense! I've introduced that constant in https://github.com/wordpress-mobile/WordPress-Android/pull/14503/commits/9630c2955da873dbbd033849a6d1494876bbb582. I've gone ahead to use the `MEDIA_ID_NO_FEATURED_IMAGE_SET` suggestion for naming and used it in three different places, where there were checks for zero. How does that seem?

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
SiobhyB: >  With a line comment, we are seeking to add more info that is not obvious just by looking at the code. So in this case, I'd expect a comment that would answer the question: "Why are we setting mFeaturedImageId to mediaId?"

@hypest, from my side, I think I'm missing something. It feels correct to me that `setFeaturedImageId` would be called here, as I don't see anywhere else in the file where `mFeaturedImageId` is being updated. I wouldn't expect it to be automatically updated without a function call. Would you be able to help me understand why this doesn't look like you'd expect? That may help me to understand the sort of context that needs to be added or what to explore further in the code.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
jd-alexander: Ah, I didn't notice that. Thanks for pointing it out.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
jd-alexander: > That makes a lot of sense! I've introduced that constant in 9630c29. I've gone ahead to use the MEDIA_ID_NO_FEATURED_IMAGE_SET suggestion for naming and used it in three different places, where there were checks for zero. How does that seem?

Looking good @SiobhyB

hypest: Nice! Any opportunities to introduce the const (or define in a common place and reuse) on the "sending" side as well? Like, in the places where functions _use_ the `0` to denote clearing the setting, no just when checking against it.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergDialogFragment.kt
SiobhyB: Ah, good point! I've gone ahead to do that in https://github.com/wordpress-mobile/WordPress-Android/pull/14503/commits/7924341c7b430ec48d6fbb56d2f4d6a8866491c9, as part of that commit I also changed the reference to `activity` to `parentFragment`, to match the way the other interfaces are implemented. Let me know if you spot any issues with that!

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
SiobhyB: I've gone ahead to make some changes to the button on the JS side here https://github.com/WordPress/gutenberg/pull/31705#pullrequestreview-667361512

If there's a better way for defining the constant there, I'm all ears. :D

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
hypest: Ah, if there's no special meaning around the appearance of `setFeaturedImageId()` here (now that we've updated the original comment that was trying to elevate the importance) the we better just go ahead and remove the line comment as it doesn't add any context.

In other words, I'm not trying to say that the call should be somewhere else or not happening at all but, wondering if there's something special about it. If yes it makes sense to capture it in the line comment, otherwise it's actually better to remove the comment. Does that help?

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergDialogFragment.kt
hypest: Thanks! I [left a comment on the commit](https://github.com/wordpress-mobile/WordPress-Android/commit/7924341c7b430ec48d6fbb56d2f4d6a8866491c9#r51332384).

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
SiobhyB: Got it, thank you for the extra explanation! This thread's definitely helped me to see how comments can cause confusion and this one is pretty unecessary given the function itself is fairly self-explanatory. I've gone ahead to remove it in https://github.com/wordpress-mobile/WordPress-Android/pull/14503/commits/36b6e2ee848e98baab166230f2e1de04366b92dd.

hypest (approved):
Awesome work here @SiobhyB , thanks!

This looks good to me and I will leave it to other reviewers to finish their rounds before we start merging 👍

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
hypest: Looks like all comments/ideasof this thread have been addressed so, I'll mark it as resolved :)

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergDialogFragment.kt
jd-alexander: Ah, thanks for expounding on that. We can go ahead and keep this class then. The only thing I would change is probably the `mediaId` references since the name of the class is `GutenbergDialogFragment ` and there could be other use cases for it in the future. In the interest of time and not overcomplicating the issue we could even remove the `media` and make it just `id`. `STATE_KEY_MEDIA_ID = "state_key_media_id"` ==> `STATE_KEY_ID = "state_key_id"` Ideally, I would utilize generics, so we could make it a `param` value that isn't bound to a type and in this case, we could pass the `Int` but yes, we don't have to do that optimization at this time so we could probably evaluate changing the name for now. Let me know what you think.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -2484,12 +2485,32 @@ protected void setPostContentFromShareAction() {
...
-                ((GutenbergEditorFragment) mEditorFragment).sendToJSFeaturedImageId((int) mediaId);
+    private void setFeaturedImageId(final long mediaId, final boolean imagePicked, final boolean isGutenbergEditor) {
+        if (isGutenbergEditor) {
+            EditPostRepository postRepository = getEditPostRepository();
+            if (postRepository == null) {
+                return;
             }
+
+            int postId = getEditPostRepository().getId();
+            if (mediaId == 0) {
```
jd-alexander: @SiobhyB wdyt about importing `MEDIA_ID_NO_FEATURED_IMAGE_SET ` here?

jd-alexander (approved):
Thanks for these changes @SiobhyB I verified that the functionality here works as expected by utilizing the test cases on the Gutenberg PR. I left a few comments but none of them were blocking and the changes are minor, so this can be approved for merge. 🚢 

The lint error is probably fixed on `develop` so you could merge it in here to see if that resolves it.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
SiobhyB: @jd-alexander, that makes sense to me! I've gone ahead to import that constant in https://github.com/wordpress-mobile/WordPress-Android/pull/14503/commits/fb3ba9548da20bea53f50ca56e8bc6fc584e877e. Let me know if anything looks off to you. :)

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergDialogFragment.kt
SiobhyB: > In the interest of time and not overcomplicating the issue we could even remove the `media` and make it just `id`.

That makes sense! I've gone ahead to implement that change in https://github.com/wordpress-mobile/WordPress-Android/pull/14503/commits/40310c385d7c435912942c2a9b0b2ea6610d1445.

> Ideally, I would utilize generics, so we could make it a param value that isn't bound to a type and in this case, we could pass the Int but yes, we don't have to do that optimization at this time so we could probably evaluate changing the name for now. Let me know what you think.

That sounds great, I've created [an issue to track other places where this class](https://github.com/wordpress-mobile/WordPress-Android/issues/14676) can be used and have made a note of your comments for optimizing it in the comments here: https://github.com/wordpress-mobile/WordPress-Android/issues/14676#issuecomment-852024183. Does iterating on the class as part of that separate issue seem like a reasonable approach to you?

SiobhyB:
@jd-alexander, thank you for this. 🙇‍♀️ I've gone ahead to implement your suggestions, let me know what you think and I'll go ahead with the merge waterfall if all looks okay to you. :)

hypest:
👋 @SiobhyB , Joel will be on AFK for the next few weeks so, I'll go ahead and take over the last steps to finalize the review here.

I gave the PR a try locally using local builds and looks like I run into an issue with multiple gradle plugin versions detected:
```
Build file '/Users/stefanos/proj/a8c/android/tmp/t2/WordPress-Android/gutenberg-mobile/gutenberg/node_modules/@react-native-community/masked-view/android/build.gradle' line: 17

* What went wrong:
A problem occurred evaluating project ':android:@react-native-community_masked-view'.
> Failed to apply plugin [id 'com.android.internal.library']
   > Using multiple versions of the Android Gradle plugin in the same build is not allowed
```

The binary (composite) build works OK, only the local source mode encounters the issue. Perhaps some change in WPAndroid or gutenberg-mobile has introduce the mismatch, can you have a look? I don't think this PR has anything to do with the mismatch, but we might as well avoid merging as-is so we don't introduce a development flow breakage. wdyt?

SiobhyB:
@hypest, thank you for the heads up!  

> Perhaps some change in WPAndroid or gutenberg-mobile has introduce the mismatch, can you have a look? I don't think this PR has anything to do with the mismatch, but we might as well avoid merging as-is so we don't introduce a development flow breakage. wdyt?

There was a little bit of a window between [Android](https://github.com/wordpress-mobile/WordPress-Android/pull/14695) and [Gutenberg](https://github.com/WordPress/gutenberg/pull/32144) updating to 4.0.2, which explains what happened here. I've pulled the latest changes from Gutenberg now, so all should work again. :)

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergDialogFragment.kt
hypest: Makes sense to iterate in a separate effort, outside this PR 👍

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
hypest: Looks good to me!

hypest:
@SiobhyB , do you perhaps get this AndroidStudio issue `Cannot access 'androidx.activity.result.ActivityResultCaller' which is a supertype of 'org.wordpress.android.editor.gutenberg.GutenbergDialogFragment'. Check your module classpath for missing or conflicting dependencies`? See screenshot too. WPAndroid builds fine both via cli gradle and AndroidStudio (and dialog works fine too) so, not a blocker for the PR, just wondering if it's on my side alone or not.

Screenshot 2021-06-03 at 11 27 50

hypest (approved):
Let's get this merged @SiobhyB !

I think the [libraries misalignment error AndroidStudio is giving out](https://github.com/wordpress-mobile/WordPress-Android/pull/14503#issuecomment-853694243) is not a blocker but, can you please open a ticket and check it out in a separate PR? WDYT?

SiobhyB:
Thank you for this, @hypest! I'm going to begin the domino. :) 

Just noting our Slack conversation, also, for posterity: p1622711015069900/1622706783.069600-slack-C01NFTM4BMY

Although I'm not currently able to replicate the error you're seeing (likely due to an older version of AS), I'll look further into this and create a GH issue. ✅

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -2302,9 +2304,7 @@ private GutenbergPropsBuilder getGutenbergPropsBuilder() {
                 enableXPosts,
                 isUnsupportedBlockEditorEnabled,
                 unsupportedBlockEditorSwitch,
-                !isFreeWPCom, // Disable audio block until it's usable on free sites via "Insert from URL" capability
-                // Only enable reusable block in WP.com sites until the issue
-                // (https://github.com/wordpress-mobile/gutenberg-mobile/issues/3457) in self-hosted sites is fixed
+                !isFreeWPCom,
```
fluiddot, line 2307: These changes come from this [PR](https://github.com/wordpress-mobile/WordPress-Android/pull/14680), that it was required to merge it here due to conflicts in the Gutenberg reference.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergPropsBuilder.kt
```diff
@@ -14,7 +14,7 @@ data class GutenbergPropsBuilder(
     private val enableXPosts: Boolean,
     private val enableUnsupportedBlockEditor: Boolean,
     private val unsupportedBlockEditorSwitch: Boolean,
-    private val enableAudioBlock: Boolean,
+    private val isAudioBlockMediaUploadEnabled: Boolean,
```
fluiddot, line 17: These changes come from this [PR](https://github.com/wordpress-mobile/WordPress-Android/pull/14680), that it was required to merge it here due to conflicts in the Gutenberg reference.

libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergPropsBuilder.kt
```diff
@@ -29,7 +29,7 @@ data class GutenbergPropsBuilder(
             enableXPosts = enableXPosts,
             enableUnsupportedBlockEditor = enableUnsupportedBlockEditor,
             canEnableUnsupportedBlockEditor = unsupportedBlockEditorSwitch,
-            enableAudioBlock = enableAudioBlock,
+            isAudioBlockMediaUploadEnabled = isAudioBlockMediaUploadEnabled,
```
fluiddot, line 32: These changes come from this [PR](https://github.com/wordpress-mobile/WordPress-Android/pull/14680), that it was required to merge it here due to conflicts in the Gutenberg reference.

SiobhyB:
To close the loop: I've created an issue for the Android Studio error here: https://github.com/wordpress-mobile/WordPress-Android/issues/14773. As I'll be working on that class as part of my HACK week project, I can spend some time trying to see if it's something I can figure out next week.

PR #14519, opened 2021-04-23
Deeplink improvements

planarvoid:
This PR ties a few loose ends:
- move app link handling for `wordpress://stats` to the `StatsLinkHandler`
- move app link handling for `wordpress://notifications to the `NotificationsLinkHandler`
- move app link handling for `wordpress://read to the `ReaderLinkHandler`
- Move all the deeplink code into a separate package for better readibility

The only small difference in behaviour is that previously for `wordpress://stats` link we were selecting the primary site and showing stats for that site. For consistency I've changed that to show the stats for the currently selected site (just switching to the `Stats` screen). I think this change makes sense.

To test:
- Make sure you're logged in in the app
- Click on `wordpress://stats` link (in case you don't know how to create one, a link like `public-api.wordpress.com/mbar/?redirect_to=wordpress://stats` works as well
- Select the app to handle the link
- Notice the app opens the stats screen

To test:
- Make sure you're logged in in the app
- Click on `wordpress://notifications` link (in case you don't know how to create one, a link like `public-api.wordpress.com/mbar/?redirect_to=wordpress://notifications` works as well
- Select the app to handle the link
- Notice the app opens the notifications screen

To test:
- Make sure you're logged in in the app
- Click on `wordpress://read` link (in case you don't know how to create one, a link like `public-api.wordpress.com/mbar/?redirect_to=wordpress://read` works as well
- Select the app to handle the link
- Notice the app opens the reader screen

## Regression Notes
- other deeplinks

- I've added unit tests and tested the other links 

- unit tests

PR submission checklist:


renanferrari (approved):
Thanks for all the changes and specially for moving all classes to a single package, that was very much needed.

Code is looking pretty good and, in the future, I think we should probably move all cases in `DeepLinkingIntentReceiverActivity` to the view model in a similar way.

I've tested all scenarios you provided using the following command:

```
adb shell am start \
 -W -a android.intent.action.VIEW \
 -c android.intent.category.BROWSABLE \
 -d "wordpress://read"
```

 And everything seems to be working as expected.

 LGTM 👍

PR #14521, opened 2021-04-23
Add a "Set as Featured Image" Button to Image Block - Refactor for Featured Image PostModel update

jd-alexander:
## Description 

The main purpose of this PR is to refactor the logic that was being utilized to update the featured image id on the `PostModel` previously, we were utilizing the `PostRepository` that was within the `EditPostSettingsFragment.` This approach was deemed an area that needed improvement since, the Gutenberg side of things would be calling functionality within a fragment when the main point of entry is the `EditPostActivity`. I am also assuming there could have been app states where the fragment isn't created so this refactor ensures that the updates to the `PostModel` are done in an object that will be retained during the editing experience - `EditPostActivity`. Let us know what you think about the refactor done. Any improvements or ideas are more than welcomed. 

To be merged into the following PR: https://github.com/wordpress-mobile/WordPress-Android/pull/14503

## Testing

The test cases below are from the [main Gutenberg PR](https://github.com/WordPress/gutenberg/pull/31705) related to these changes. In addition, we should verify that the Tracks events edited/introduced with this PR work as expected.

### Image Block - Set and Remove Featured Image

1. In the app, navigate to _Home_ → _Blog Posts_ to create a new blog post, one which doesn't already have a featured image set.
2. Add a new image block, and upload an image to it.
3. Tap the image block's cog/gear icon to access its settings.
4. Observe the new **Set as Featured** button at the bottom of the settings and tap it.
5. Verify that you see a confirmation message about the newly set featured image and that the block now has a **Featured** banner overlaying it.
6. Tap the image block's cog/gear icon again and this time select the **Remove as Featured** option.
7. Verify that you see a confirmation message that the image has been removed and that the **Featured** banner is also removed.

### Image Block - Tracks Verification

1. After following the steps to set and remove an image as featured directly from the image block, navigate to the internal Tracks tool and choose the **Live View** option from the menu.
2. Enter `wpandroid_featured_image_picked_gutenberg_editor` into the tool's **Events** field and then the username associated with the account you tested with in the **Username** field. Click **Search**.
3. Verify that the event was logged correctly. Note, there is usually a delay of ~5 minutes before an event is logged in this tool.
4. Use the same steps to search for the `wpandroid_featured_image_removed_gutenberg_editor` event, in order to verify that this event fires after the steps to remove a featured image via the editor were followed.

### Post Settings - Set and Remove Featured Image

1. Tap the three icons to the upper right of the post editor to access **Post Settings**.
2. Tap existing featured image and then **Remove**.
3. Tap the back button to confirm that the **Featured** banner has been removed from the featured image, as expected. Tap the cog/gear icon to then confirm the option to **Set as Featured** appears.
4. Return to the **Post Settings** screen and set a new featured image.
5. Tap the back button. If the new featured image hasn't already been added to the editor via an image block, add it.  
6. Confirm that the **Featured** banner displays overlays the correct image and that  **Remove as Featured** appears in its settings.

### Post Settings - Tracks Verification

1. After following the steps to set a featured image via **Post Settings**, navigate to our internal Tracks tool and choose the **Live View** option from the menu.
2. Enter `wpandroid_featured_image_set_clicked_post_settings` into the tool's **Events** field and then the username associated with the account you tested with in the **Username** field. Click **Search**.
3. Verify that the event was logged correctly. Note, there is usually a delay of ~5 minutes before an event is logged in this tool.
4. Use the same steps to search for the `wpandroid_featured_image_remove_clicked_post_settings` event, in order to verify that this event fires after the steps to remove a featured image via **Post Settings** were followed.

## Regression Notes
The main unintended issues with these changes could be: 

* Modified track events aren't fired as expected. 
* The Post Settings featured image behavior gets broken. 

To ensure that these areas aren't impacted we will test these paths thoroughly. 

To test the areas of impact, the flows above were utilized.

N/A

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/posts/UpdateFeaturedImageUseCase.kt
```diff
@@ -0,0 +1,23 @@
+package org.wordpress.android.ui.posts
+
+import org.wordpress.android.fluxc.model.PostImmutableModel
+import org.wordpress.android.fluxc.model.PostModel
+import org.wordpress.android.ui.posts.EditPostRepository.UpdatePostResult
+import org.wordpress.android.ui.posts.EditPostRepository.UpdatePostResult.Updated
+
+class UpdateFeaturedImageUseCase {
```
SiobhyB:
```suggestion
import javax.inject.Inject

class UpdateFeaturedImageUseCase @Inject constructor() {
```

I'm only able to get the PR to build if I use the `@Inject` annotation here, let me know if this looks right to you!

SiobhyB (changes requested):
@jd-alexander, this works perfectly for me once I add the `@Inject` annotation to the `UpdateFeaturedImageUseCase.kt` file. I added what I used to get this working as a suggestion as part of the review, let me know if this is correct or if there's a different syntax that'd be preferable. 

I also submitted a couple of commits to fix small issues:

* Changed a couple of remaining references from `IMAGE_PICKED` to `IMAGE_PICKED_POST_SETTINGS `: https://github.com/wordpress-mobile/WordPress-Android/pull/14521/commits/ce3785e42339f2a8a16644a41cfd29c54d55bd26
* Removed unused imports from `EditPostSettingsFragment.java`: https://github.com/wordpress-mobile/WordPress-Android/pull/14521/commits/c1278d6a6262c401d72501be991b9d2d13ea7408

WordPress/src/main/java/org/wordpress/android/ui/posts/UpdateFeaturedImageUseCase.kt
jd-alexander: Good catch. Thank you 🙇🏾

jd-alexander: @SiobhyB I resolved this here 610348e and did another run-through of the PR. I will be adding PR notes later with the testing instructions so this can be reviewed and merged in.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -2481,12 +2482,32 @@ protected void setPostContentFromShareAction() {
...
-                ((GutenbergEditorFragment) mEditorFragment).sendToJSFeaturedImageId((int) mediaId);
+    private void setFeaturedImageId(final long mediaId, final boolean imagePicked, final boolean isGutenbergEditor) {
+        if (isGutenbergEditor) {
+            EditPostRepository postRepository = getEditPostRepository();
+            if (postRepository == null) {
+                return;
+            }
+
+            int postId = getEditPostRepository().getId();
+            if (mediaId == 0) {
```
jd-alexander, line 2493: This value will is under consideration to be hardcoded.

ashiagr (approved):
Hey @jd-alexander 👋!

All works as described, good job 👍. Also from your regression notes, I've tested that the older track events for (featured image on post settings) are not broken, even though the event label got changed. 

The only odd thing I noticed: `UpdateFeaturedImageUseCase` is formatted in commit 70dff2c but I still see it unformatted, can you re-check it? Feel free to merge after it.

jd-alexander:
> All works as described, good job 👍. Also from your regression notes, I've tested that the older track events for (featured image on post settings) are not broken, even though the event label got changed.

Thanks much for the review @ashiagr it's appreciated 🙇🏾 

> The only odd thing I noticed: UpdateFeaturedImageUseCase is formatted in commit 70dff2c but I still see it unformatted, can you re-check it? Feel free to merge after it.

Thanks for spotting. I will do that!

PR #14534, opened 2021-04-27
Update release toolkit version to 0.18.0

JavonDavis:
Updates release toolkit version to 0.17.2

To test: Create a temporary branch and run `bundle exec fastlane download_translations`... Verify that the strings get translated as expected

PR submission checklist:


fastlane/Fastfile
```diff
@@ -8,28 +8,28 @@ end
...
+  { glotpress: "pl", android: "pl",    google_play: "pl-PL",  promo_config: {} },
+  { glotpress: "pt-br", android: "pt-rBR", google_play: "pt-BR",  promo_config: {} },
+  { glotpress: "ru", android: "ru",    google_play: "ru-RU",  promo_config: {} },
+  { glotpress: "sr", android: "sr",    google_play: "sr",     promo_config: {} },
+  { glotpress: "sv", android: "sv",    google_play: "sv-SE",  promo_config: {} },
+  { glotpress: "th", android: "th",    google_play: "th",     promo_config: {} },
+  { glotpress: "tr", android: "tr",    google_play: "tr-TR",  promo_config: {} },
+  { glotpress: "vi", android: "vi",    google_play: "vi",     promo_config: {} },
+  { glotpress: "zh-cn", android: "zh-rCN", google_play: "zh-CN",  promo_config: {} },
+  { glotpress: "zh-tw", android: "zh-rTW", google_play: "zh-TW",  promo_config: {} },
```
AliSoftware, line 33: We have _way_ more than those Mag16 locales in GlotPress for the `strings.xml` files.
We need to download more than just those, so the list that you will pass to `android_download_translations` needs to be expanded to match the full list that is currently defined in `tools/exported-language-codes.xml`.

What I suggest to do is:

 - Rename that constant above to `ALL_LOCALES`
 - Add all the missing entries for locales that are listed in `tools/exported-language-codes.csv` to that `ALL_LOCALES` array, but with no `google_play` key, and with `promo_config: false`.
 - Create a `RELEASE_NOTES_LOCALES = ALL_LOCALES.reject { |h| h[:google_play].nil? }.map { |h| [h[:glotpress], h[:google_play]] }`

```ruby
ALL_LOCALES = [
  # Mag16, which we have release notes for
  { glotpress: "ar", android: "ar",    google_play: "ar",     promo_config: {} },
  { glotpress: "de", android: "de",    google_play: "de-DE",  promo_config: {} },
  { glotpress: "en-gb", android: "en-rGB", google_play: "en-US",  promo_config: {} },
  { glotpress: "es", android: "es",    google_play: "es-ES",  promo_config: {} },
  { glotpress: "fr", android: "fr-rCA",    google_play: "fr-CA",  promo_config: false },
  { glotpress: "fr", android: "fr",    google_play: "fr-FR",  promo_config: {} },
…
  { glotpress: "zh-tw", android: "zh-rTW", google_play: "zh-TW",  promo_config: {} },

  # Additional locales for which we have app strings in GlotPress, but support is unofficial, with no release notes
  { glotpress: "hr", android: "hr", promo_config: false },
  { glotpress: "cy", android: "cy", promo_config: false },
…
]

RELEASE_NOTES_LOCALES = ALL_LOCALES.reject { |h| h[:google_play].nil? }.map { |h| [h[:glotpress], h[:google_play]] }
```

 - Update the call sites for `android_download_translations` to use `ALL_LOCALES` as a parameter for `locales:`
 - Update the call sites for `gp_downloadmetadata` & `android_create_xml_release_notes` (from the `download_metadata_strings` lane) to use `RELEASE_NOTES_LOCALES` instead of `SUPPORTED_LOCALES.map {| hsh | [ hsh[:glotpress], hsh[:google_play] ]}`
 - Update the 3 uses of `SUPPORTED_LOCALES` that are in `ScreenshotFastfile` to use `ALL_LOCALES` instead
 - Delete the old `tools/exported-language-codes.csv` and `tools/update-translations.sh` files

fastlane/Fastfile
```diff
@@ -628,6 +628,24 @@ REPOSITORY_NAME="WordPress-Android"
...
+    # -----------------------------------------------------------------------------------
+    # Usage:
+    # bundle exec fastlane download_translations
+    #####################################################################################
+  lane :download_translations do |options|
+    # Adapt the GP project URL and lint_task as needed if testing a project other than WordPress
+    android_download_translations(
+      res_dir: File.join(ENV['PROJECT_NAME'], 'src', 'main', 'res'),
+      glotpress_url: 'https://translate.wordpress.org/projects/apps/android/dev/',
+      locales: SUPPORTED_LOCALES,
```
```suggestion
      locales: ALL_LOCALES,
```
With `ALL_LOCALES` defined as per previous comment above

fastlane/Fastfile
```diff
@@ -628,6 +628,24 @@ REPOSITORY_NAME="WordPress-Android"
...
+      lint_task: 'lintWordpressVanillaRelease'
```
AliSoftware, line 680: 🤔  Before that change, the legacy action was using `lintVanillaRelease` (it read that lint task name from the `ENV["validate_translations"]="lintVanillaRelease"` env var defined on line 58… and which can actually get removed while we're at it, because it wasn't used by anything other than the legacy action that you just replaced 😉

fastlane/Fastfile
```diff
@@ -638,16 +656,19 @@ REPOSITORY_NAME="WordPress-Android"
   #####################################################################################
   desc "Builds an app bundle"
   lane :build_for_translation_review do | options |
+    android_download_translations(
+      glotpress_url: 'https://translate.wordpress.com/projects/wporg/apps/android',
+      status_filter: ['current', 'waiting', 'fuzzy'],
+      locales: SUPPORTED_LOCALES,
```
AliSoftware: For this one you need to only pass the subset of locales that is supported by the GlotPress project you use as URL, which seems different from the main one for normal build.

This means that you need to only pass the subset of `ALL_LOCALES` that matches what is listed in `tools/review-language-codes.csv`.

What I suggest, similar to previous comment, is to have the `ALL_LOCALES` array have an additional key, e.g. `translator: true/false`, for each entry, and then use `ALL_LOCALES.select { |h| h[:translator] }` to derive the list to pass as parameter to this call site.

Another approach could be to list the glotpress locales and select entries matching those:
```ruby
translator_locales = %w[de es fr id it ja ko nl ru sv zh-cn zh-tw tr he pt ar]
translator_locales_map = ALL_LOCALES.select { translator_locales.include?(h[:glotpress]) }
```
(and then pass translator_locales_map variable to the locales: parameter)


Once this is done, you should take the occasion to also delete the `tools/review-language-codes.csv` file since it won't be used anymore.

fastlane/Pluginfile
```diff
@@ -6,5 +6,5 @@ group :screenshots, optional: true do
   gem 'rmagick', '~> 4.1'
 end
 
-gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', tag: '0.17.1'
+gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', branch: 'release/0.17.2'
```
AliSoftware: 🚨 Just as a reminder that, as you stated in the PR description, we should not forget to update this back to tag before merging this PR.

AliSoftware (changes requested):

fastlane/Fastfile
JavonDavis: Agreed! 👍  I'll opt to just pull out the subset using your second suggestion here, thanks @AliSoftware

JavonDavis:
Hey @AliSoftware! 👋  Addressed all the feedback and I think this is ready for another look now! Thank you! 🙇‍♂️

fastlane/Pluginfile
JavonDavis: Done! Updated to `0.18.0`

fastlane/Fastfile
JavonDavis: The list I used was `%w[de es fr id it ja ko nl ru sv zh-cn zh-tw tr he pt ar pl th sr vi]` which I think has all the locales that would be translated, let me know if you had any other questions/feedback on this, thanks!

fastlane/Fastfile
JavonDavis: Removed the env var 👍

fastlane/Fastfile
JavonDavis: Thanks for all the details here! Agree with the feedback and went through with these updates 🙂 Please Let me know if I missed anything!

fastlane/Fastfile
```diff
@@ -7,31 +7,65 @@ end
...
+  { glotpress: "pt-br", android: "pt-rBR", google_play: "pt-BR",  promo_config: {} },
+  { glotpress: "ru", android: "ru",    google_play: "ru-RU",  promo_config: {} },
+  { glotpress: "sr", android: "sr",    google_play: "sr",     promo_config: {} },
+  { glotpress: "sv", android: "sv",    google_play: "sv-SE",  promo_config: {} },
+  { glotpress: "th", android: "th",    google_play: "th",     promo_config: {} },
+  { glotpress: "tr", android: "tr",    google_play: "tr-TR",  promo_config: {} },
+  { glotpress: "vi", android: "vi",    google_play: "vi",     promo_config: {} },
+  { glotpress: "zh-cn", android: "zh-rCN", google_play: "zh-CN",  promo_config: {} },
+  { glotpress: "zh-tw", android: "zh-rTW", google_play: "zh-TW",  promo_config: {} },
+  { glotpress: "az", android: "az", promo_config: false},
```
AliSoftware, line 35: nit: just a suggestion for adding a comment, in order to make it clear why it suddenly changes format (no more google_play key) in the middle of the list

```suggestion
  # From this point are locales that are sto;; used for downloading `strings.xml`… but not for release notes – and thus don't need a `google_play` key. See `RELEASE_NOTES_LOCALES` below.
  { glotpress: "az", android: "az", promo_config: false},
```

fastlane/Fastfile
```diff
@@ -7,31 +7,65 @@ end
...
-  { glotpress: "ru",    google_play: "ru-RU",  promo_config: {} },
-  { glotpress: "sr",    google_play: "sr",     promo_config: {} },
-  { glotpress: "sv",    google_play: "sv-SE",  promo_config: {} },
-  { glotpress: "th",    google_play: "th",     promo_config: {} },
-  { glotpress: "tr",    google_play: "tr-TR",  promo_config: {} },
-  { glotpress: "vi",    google_play: "vi",     promo_config: {} },
-  { glotpress: "zh-cn", google_play: "zh-CN",  promo_config: {} },
-  { glotpress: "zh-tw", google_play: "zh-TW",  promo_config: {} },
+ALL_LOCALES = [
+  { glotpress: "ar", android: "ar",    google_play: "ar",     promo_config: {} },
```
AliSoftware, line 12:
```suggestion
  # First are the locales which are used for *both* downloading the `strings.xml` files from GlotPress *and* for generating the release notes XML files.
  { glotpress: "ar", android: "ar",    google_play: "ar",     promo_config: {} },
```

fastlane/Fastfile
```diff
@@ -477,11 +510,11 @@ REPOSITORY_NAME="WordPress-Android"
     download_path=Dir.pwd + "/metadata/android"
     gp_downloadmetadata(project_url: "https://translate.wordpress.org/projects/apps/android/release-notes/",
       target_files: files,
-      locales: SUPPORTED_LOCALES.map {| hsh | [ hsh[:glotpress], hsh[:google_play] ]},
+      locales: RELEASE_NOTES_LOCALES.map {| hsh | [ hsh[:glotpress], hsh[:google_play] ]},
```
AliSoftware: The `.map` part was already applied when you created the `RELEASE_NOTES_LOCALES` constant.

You should thus either remove the `.map` from the definition of `RELEASE_NOTES_LOCALES` on line 67, or keep it on line 67 but remove it from this line 513. Given it's used in the exact same way on line 517 below too, I suggest the former solution, so that we only do the `.map` once.

```suggestion
      locales: RELEASE_NOTES_LOCALES,
```

fastlane/Fastfile
```diff
@@ -477,11 +510,11 @@ REPOSITORY_NAME="WordPress-Android"
...
       source_locale: "en-US",
       download_path: download_path)
 
-    android_create_xml_release_notes(download_path: download_path, build_number: "#{options[:build_number]}", locales: SUPPORTED_LOCALES.map {| hsh | [ hsh[:glotpress], hsh[:google_play] ]})
+    android_create_xml_release_notes(download_path: download_path, build_number: "#{options[:build_number]}", locales: RELEASE_NOTES_LOCALES.map {| hsh | [ hsh[:glotpress], hsh[:google_play] ]})
```
AliSoftware: (See previous comment)
```suggestion
    android_create_xml_release_notes(download_path: download_path, build_number: "#{options[:build_number]}", locales: RELEASE_NOTES_LOCALES)
```

fastlane/Fastfile
```diff
@@ -638,16 +689,23 @@ REPOSITORY_NAME="WordPress-Android"
   #####################################################################################
   desc "Builds an app bundle"
   lane :build_for_translation_review do | options |
+
+  translator_locales = %w[de es fr id it ja ko nl ru sv zh-cn zh-tw tr he pt ar pl th sr vi]
+  translator_locales_map = ALL_LOCALES.select { translator_locales.include?(h[:glotpress]) }
```
AliSoftware: Indentation fix
```suggestion
    translator_locales = %w[de es fr id it ja ko nl ru sv zh-cn zh-tw tr he pt ar pl th sr vi]
    translator_locales_map = ALL_LOCALES.select { translator_locales.include?(h[:glotpress]) }
```

fastlane/Fastfile
```diff
@@ -638,16 +689,23 @@ REPOSITORY_NAME="WordPress-Android"
...
     unless options[:custom_version].nil?
       UI.message("Building custom version: #{options[:custom_version]}")
       cmd_params=" -PversionName=\"#{options[:custom_version]}\""
     end
 
-    sh("../tools/update-translations.sh review")
-    sh("../tools/update-translations.sh review waiting")
-    sh("../tools/update-translations.sh review fuzzy")
-
     android_merge_translators_strings(strings_folder:"./WordPress/src/main/res/")
```
AliSoftware: This step should be removed too as the merging is already handled by `android_download_translations` when passing multiple values (here, 3) for `status_filter`.

```suggestion
```

fastlane/Pluginfile
```diff
@@ -6,5 +6,5 @@ group :screenshots, optional: true do
   gem 'rmagick', '~> 4.1'
 end
 
-gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', tag: '0.17.1'
+gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', tag: '0.18.0'
```
AliSoftware, line 9: 🎉

fastlane/Fastfile
```diff
@@ -7,31 +7,65 @@ end
...
+  { glotpress: "hr", android: "hr", promo_config: false},
+  { glotpress: "cs", android: "cs", promo_config: false},
+  { glotpress: "pt-br", android: "pt-rBR", promo_config: false},
+  { glotpress: "en-ca", android: "en-rCA", promo_config: false },
+  { glotpress: "ms", android: "ms", promo_config: false},
+  { glotpress: "es-ve", android: "es-rVE", promo_config: false},
+  { glotpress: "gl", android: "gl", promo_config: false},
+  { glotpress: "is", android: "is"},
+  { glotpress: "es-co", android: "es-rCO", promo_config: false },
+  { glotpress: "kmr", android: "kmr", promo_config: false},
```
AliSoftware, line 66: 👍 

_(Hopefully we will soon be able to get rid of those ugly maps in a future PR when we do a deeper refactoring of our translation tooling and likely use https://github.com/wordpress-mobile/fastlane-locale-map/, which will make all that ugly long list way shorter and readable… but we're not there yet 😅 )_

AliSoftware (commented):
I noticed you removed the `exported-language-codes.csv` but forgot to also delete the `review-language-codes.csv` file, probably worth cleaning up too.

fastlane/Fastfile
AliSoftware: Oh and also we forgot to provide the param name to the `select` block (meaning that current code would crash with syntax error)
```suggestion
    translator_locales = %w[de es fr id it ja ko nl ru sv zh-cn zh-tw tr he pt ar pl th sr vi]
    translator_locales_map = ALL_LOCALES.select { |h| translator_locales.include?(h[:glotpress]) }
```

fastlane/Fastfile
```diff
@@ -30,6 +30,7 @@ ALL_LOCALES = [
   { glotpress: "vi", android: "vi",    google_play: "vi",     promo_config: {} },
   { glotpress: "zh-cn", android: "zh-rCN", google_play: "zh-CN",  promo_config: {} },
   { glotpress: "zh-tw", android: "zh-rTW", google_play: "zh-TW",  promo_config: {} },
+  # From this point are locales that are sto;; used for downloading `strings.xml`… but not for release notes – and thus don't need a `google_play` key. See `RELEASE_NOTES_LOCALES` below.
```
AliSoftware: Whoops, seems my fingers shifted 1 key to the right when I typed that comment (`still` -> `sto;;`)… 😅

fastlane/Fastfile
```diff
@@ -628,6 +663,24 @@ REPOSITORY_NAME="WordPress-Android"
...
   end
 
+  #####################################################################################
+    # download_translations
+    # -----------------------------------------------------------------------------------
+    # This lane download the string translations from GlotPress
+    # -----------------------------------------------------------------------------------
+    # Usage:
+    # bundle exec fastlane download_translations
+    #####################################################################################
```
```suggestion
  #####################################################################################
  # download_translations
  # -----------------------------------------------------------------------------------
  # This lane download the string translations from GlotPress
  # -----------------------------------------------------------------------------------
  # Usage:
  # bundle exec fastlane download_translations
    #####################################################################################
```

fastlane/Fastfile
```diff
@@ -628,6 +663,24 @@ REPOSITORY_NAME="WordPress-Android"
...
   end
 
+  #####################################################################################
+  # download_translations
+  # -----------------------------------------------------------------------------------
+  # This lane download the string translations from GlotPress
+  # -----------------------------------------------------------------------------------
+  # Usage:
+  # bundle exec fastlane download_translations
+    #####################################################################################
```
AliSoftware: Dammit.
```suggestion
  #####################################################################################
```

AliSoftware (approved):
LGTM!

PR #14532, opened 2021-04-26
Adding like enhancements unit testing

develric:
Part of the #14137 and #14138. It requires #14525 to be merged first.

This PR adds unit testing to cover the Handler, Use Case and Utils layers. I'm covering the View Models layer in a separate PR to keep it more manageable for the review.

# To test
To test it should be enough to check the CI is green.

## Regression Notes
N/A; behind feature flag.

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #14541, opened 2021-04-27
Like enhancements unit testing 2

develric:
This is a second part PR to the #14532. Adding unit testing to view model layers for likes enhancements.

## To test
To test it should be enough to check the CI is green.

## Regression Notes
N/A; behind feature flag.

PR submission checklist:

### DIFF
--- a/WordPress/src/test/java/org/wordpress/android/ui/engagement/EngagedPeopleListViewModelTest.kt
+++ b/WordPress/src/test/java/org/wordpress/android/ui/engagement/EngagedPeopleListViewModelTest.kt
@@ -1,6 +1,5 @@
 package org.wordpress.android.ui.engagement
 
-import androidx.arch.core.executor.testing.InstantTaskExecutorRule
 import androidx.lifecycle.MutableLiveData
 import com.nhaarman.mockitokotlin2.mock
 import com.nhaarman.mockitokotlin2.times
@@ -9,12 +8,10 @@ import com.nhaarman.mockitokotlin2.whenever
 import kotlinx.coroutines.InternalCoroutinesApi
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Before
-import org.junit.Rule
 import org.junit.Test
-import org.junit.runner.RunWith
 import org.mockito.ArgumentMatchers.anyLong
 import org.mockito.Mock
-import org.mockito.junit.MockitoJUnitRunner
+import org.wordpress.android.BaseUnitTest
 import org.wordpress.android.TEST_DISPATCHER
 import org.wordpress.android.test
 import org.wordpress.android.ui.engagement.AuthorName.AuthorNameString
@@ -51,11 +48,7 @@ import org.wordpress.android.util.analytics.AnalyticsUtilsWrapper
 import org.wordpress.android.viewmodel.Event
 
 @InternalCoroutinesApi
-@RunWith(MockitoJUnitRunner::class)
-class EngagedPeopleListViewModelTest {
-    @Rule
-    @JvmField val rule = InstantTaskExecutorRule()
-
+class EngagedPeopleListViewModelTest : BaseUnitTest() {
     @Mock lateinit var getLikesHandler: GetLikesHandler
     @Mock lateinit var readerUtilsWrapper: ReaderUtilsWrapper
     @Mock lateinit var analyticsUtilsWrapper: AnalyticsUtilsWrapper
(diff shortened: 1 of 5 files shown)


WordPress/src/test/java/org/wordpress/android/ui/engagement/EngagedPeopleListViewModelTest.kt
```diff
@@ -0,0 +1,518 @@
...
+import org.wordpress.android.ui.pages.SnackbarMessageHolder
+import org.wordpress.android.ui.reader.utils.ReaderUtilsWrapper
+import org.wordpress.android.util.analytics.AnalyticsUtilsWrapper
+import org.wordpress.android.viewmodel.Event
+
+@InternalCoroutinesApi
+@RunWith(MockitoJUnitRunner::class)
+class EngagedPeopleListViewModelTest {
+    @Rule
+    @JvmField val rule = InstantTaskExecutorRule()
```
khaykov: FYI, we have a `BaseUnitTest` class that you can extend, and it has both `@RunWith(MockitoJUnitRunner::class)` and `@Rule @JvmField val rule = InstantTaskExecutorRule()` :)

WordPress/src/test/java/org/wordpress/android/ui/engagement/EngagedPeopleListViewModelTest.kt
```diff
@@ -0,0 +1,518 @@
...
+
+        viewModel.start(listScenario)
+
+        getLikesState.value = likesState
+
+        requireNotNull(uiState).let {
+            assertThat(it.showLikeFacesTrain).isFalse
+            assertThat(it.showLoading).isFalse
+            with(it.engageItemsList) {
+                val likedItem = this.filterIsInstance<LikedItem>()
```
khaykov, line 200: TIL about `filterIsInstance` !

WordPress/src/test/java/org/wordpress/android/ui/engagement/utils/GetLikesTestUtils.kt
```diff
@@ -30,6 +49,61 @@ fun getDefaultLikers(numLikers: Int, likedItemType: LikeType, siteId: Long, item
...
+    TEST_CONFIG_1,
+    TEST_CONFIG_2,
+    TEST_CONFIG_3,
+    TEST_CONFIG_4,
+    TEST_CONFIG_5
+}
+
+fun getGetLikesState(testConfig: GetLikesTestConfig): GetLikesState {
+    return when (testConfig) {
+        TEST_CONFIG_1 -> {
```
khaykov, line 63: What do you think about adding a comment to each state explaining briefly what it represents?

WordPress/src/test/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostDetailViewModelTest.kt
```diff
@@ -772,6 +791,80 @@ class ReaderPostDetailViewModelTest {
...
+        val post = mock<ReaderPost>()
+        whenever(post.isWP).thenReturn(true)
+        viewModel.post = post
+        val likeObserver = init().likesUiState
+
+        getLikesState.value = likesState
+
+        assertThat(likeObserver).isNotEmpty
+        with(likeObserver.first()) {
+            assertThat(showLikeFacesTrain).isTrue
```
khaykov, line 839: Not sure, but shouldn't `showLikeFacesTrain` be false on failure and no cache?

khaykov (commented):
Thanks for all the tests, @develric ! Looks good to me, I just left a couple of non-critical questions/comments :)

WordPress/src/test/java/org/wordpress/android/ui/engagement/EngagedPeopleListViewModelTest.kt
develric: ahah 😄 ! Now that you say it, I think I even knew about it before but totally forgot! Changed to use it 🙇 .

WordPress/src/test/java/org/wordpress/android/ui/reader/viewmodels/ReaderPostDetailViewModelTest.kt
develric: So, I think we should rename the `showLikeFacesTrain` to something like `showLikeFacesTrainContainer`; it is actually the root container so it should be visible to show the empty state, but agree the naming is not good 👍 . I'm going to open a small separate PR for that if you agree 🙇

khaykov: Sounds good 👍

khaykov (approved):
Thanks for the changes! Looks good to me 🚀

PR #14545, opened 2021-04-28
Handle viewpost deeplink in view model

planarvoid:
This PR introduces 2 changes:
- moves handling of the `wordpress://viewpost` links to the viewmodel. They're now handled by the `ReaderLinkHandler` since they are opened in the Reader post detail. In addition to this I've changed the behaviour a little bit. Previously if the blog ID and post ID were not present/parsed, the app just finished. I've changed the behaviour to open default reader instead. I think this is better.
- Adds check whether the user is logged in before opening a link. If the user is not logged in, the URI is cached and the login is triggered for a result. Once we get the result, we retrigger the cached URL. This is a very generic approach and works for all the use cases. 

To test:
- Make sure you're logged in with a blog with remote ID X and post with remote ID Y
- open a `wordpress://viewpost?blogId=X&postId=Y` deeplink
- Notice the post detail is opened in the reader post detail
- Go back
- Notice you go to the main activity instead of closing the app

To test:
- Make sure you're logged out
- open a `wordpress://viewpost?blogId=X&postId=Y` deeplink
- Notice you're redirected to Login process
- Log in with a blog with remote ID X and post with remote ID Y
- Finish login
- Notice the post detail is opened in the reader post detail
- Go back
- Notice you go to the main activity instead of closing the app

To test:
- Make sure you're logged in
- open a `wordpress://viewpost` deeplink
- Notice the reader is opened

To test
- Test other links (stats, pages, read) without the user being logged in
- Notice the user is redirected to the desired link once login ends
- Notice the links work as expected when the user is logged in

## Regression Notes
- all the deeplinks

- I've tested multiple flows and everything is unit tested

- unit tests

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/ActivityLauncher.java
@@ -46,6 +46,7 @@ import org.wordpress.android.ui.domains.DomainRegistrationActivity;
 import org.wordpress.android.ui.domains.DomainRegistrationActivity.DomainRegistrationPurpose;
 import org.wordpress.android.ui.engagement.EngagedPeopleListActivity;
 import org.wordpress.android.ui.engagement.HeaderData;
+import org.wordpress.android.ui.engagement.ListScenario;
 import org.wordpress.android.ui.engagement.ListScenarioType;
 import org.wordpress.android.ui.history.HistoryDetailActivity;
 import org.wordpress.android.ui.history.HistoryDetailContainerFragment;
@@ -61,7 +62,6 @@ import org.wordpress.android.ui.main.SitePickerAdapter.SitePickerMode;
 import org.wordpress.android.ui.main.WPMainActivity;
 import org.wordpress.android.ui.media.MediaBrowserActivity;
 import org.wordpress.android.ui.media.MediaBrowserType;
-import org.wordpress.android.ui.engagement.ListScenario;
 import org.wordpress.android.ui.pages.PageParentActivity;
 import org.wordpress.android.ui.pages.PagesActivity;
 import org.wordpress.android.ui.people.PeopleManagementActivity;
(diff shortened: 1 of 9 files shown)


renanferrari (approved):
Thanks for the changes! Code is looking good and even though I couldn't make everything work as expected, it seems that the odd behaviors I found already existed before, so feel free to merge this once you read the comments below.

>- Make sure you're logged in with a blog with remote ID X and post with remote ID Y
>- open a wordpress://viewpost?blogId=X&postId=Y deeplink

I don't know if I understood this correctly, but is this kind of link supposed to work only for posts from the user's sites? Or is it supposed to work for any posts? 

I assume it's the latter, but I couldn't make it work. I tested with posts from the user's sites and also with posts from other sites, but in both cases I kept getting redirected to the Reader's main screen.

Apparently this was not working before as well, the difference is that I'd only see the Toast instead of getting redirected, so feel free to open an issue to track this separately from this PR.

>- Make sure you're logged out
>- open a wordpress://viewpost?blogId=X&postId=Y deeplink

I couldn't test this for the `viewpost` links specifically (because of the issue I described above), but it did work well for other links I tested. (And I like that the solution is quite simple!) 

One thing I wonder if it wouldn't be possible to open the Reader while logged out. It looks like we already do this for other post links (like the ones in #14471). Since this would change the existing behavior, I'd also open an issue to track it separately from this PR.

planarvoid:
thanks for the review @renanferrari !

> I don't know if I understood this correctly, but is this kind of link supposed to work only for posts from the user's sites? Or is it supposed to work for any posts?

> I assume it's the latter, but I couldn't make it work. I tested with posts from the user's sites and also with posts from other sites, but in both cases I kept getting redirected to the Reader's main screen.

> Apparently this was not working before as well, the difference is that I'd only see the Toast instead of getting redirected, so feel free to open an issue to track this separately from this PR.

I've just checked and you don't have to be logged in by the site X. The thing is that the post has to be published and the site has to be public. You can try the following deep link to a public post: `wordpress://viewpost?blogId=165726002&postId=1328` which should work for any account.

> One thing I wonder if it wouldn't be possible to open the Reader while logged out. It looks like we already do this for other post links (like the ones in #14471). Since this would change the existing behavior, I'd also open an issue to track it separately from this PR.

I think it should be possible, I've created an issue to handle this here https://github.com/wordpress-mobile/WordPress-Android/issues/14562

renanferrari:
>I think it should be possible, I've created an issue to handle this here #14562

Thank you for creating the issue!

>The thing is that the post has to be published and the site has to be public.

I had tried that but it wasn't working, in the end I figured out that I had to escape the `&` character in the URL for the adb command to work, like so:

```
adb shell am start \
 -W -a android.intent.action.VIEW \
 -c android.intent.category.BROWSABLE \
 -d "wordpress://viewpost?blogId=165726002\&postId=1328"
```
 
Such a silly thing 🤦‍♂️

Everything is working as expected now! LGTM 👍

planarvoid:
> I had tried that but it wasn't working, in the end I figured out that I had to escape the & character in the URL for the adb command to work, like so:

ah yeah, sorry I haven't mentioned that. It blocked me for a while as well :D

PR #14549, opened 2021-04-28
Update/ci tools release branch

JavonDavis:
See https://github.com/wordpress-mobile/WordPress-Android/pull/14494

Replaying changes onto the release branch

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -159,14 +159,6 @@ jobs:
       - update-gradle-memory
       - android/restore-gradle-cache
       - run:
-          name: Install other tools
-          command: |
-            /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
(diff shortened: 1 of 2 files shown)


jkmassel:
I presume the test for this is that you can cut a new beta? :)

AliSoftware (approved):
Looking good 👍 :shipit: 

Let's merge this and test it right away with the beta you planned to do.

JavonDavis:
> I presume the test for this is that you can cut a new beta? :) 

Yep @jkmassel 🙂

AliSoftware:
> I presume the test for this is that you can cut a new beta? :)

@jkmassel Yep. I tested that the same changes that were done in #14494 worked – by triggering a Release Build in that PR's branch – and (apart from a glitch due to my temporary change in order to limit side effects during the testing of the PR) it seemed to work as expected.

So the next step was to report the changes from #14494 on the release branch – this PR – then actually do a new beta – coincidentally, we needed a new beta anyway – and if all works as expected in the real test here, finally merge #14494.

PR #14557, opened 2021-04-29
Move Story option to the top of the main Create menu

renanferrari:
This PR is pretty straightforward and just updates the position of the "Story post" action inside the main Create menu in the My Site screen, from the bottom of the list to the top. 

For more context on this change, check this internal ref: pbArwn-26B-p2#comment-2975

Here's how the change looks like:

| Before | After |
| ------- | ----------- |
|story-last|story-first|

### To test

Notes: 

1. Open the app.
1. Make sure you have selected a site that supports stories.
1. On the My Site screen, tap the FAB.
1. Notice that the menu actions have the following order: "Story post", "Blog post" and "Site page".

To test:

## Regression Notes
Unlikely, but maybe post and page creation.

Manual testing.

None.

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModel.kt
```diff
@@ -118,6 +118,16 @@ class WPMainActivityViewModel @Inject constructor(
     private fun loadMainActions(site: SiteModel?) {
         val actionsList = ArrayList<MainActionListItem>()
 
+        if (SiteUtils.supportsStoriesFeature(site)) {
```
mzorz, line 129: I think this block should be moved below the `Add new` action [here](https://github.com/wordpress-mobile/WordPress-Android/pull/14557/files#diff-e536c944784cffa9b67cd62bea37b9d8e9f34a59927b3a59edfd3495956c46a6R131-R138), otherwise the bottom sheet looks like this right now on this PR:

Screenshot_20210430-104840

mzorz (changes requested):
Hi @renanferrari  - I made a comment regarding the order in which items are added to the list that should be changed first (I also wonder how the before/after screenshots in the description would reflect the change in the PR?).

Also, I think the order of the list could probably be reflected in the Posts list for consistency, wdyt? [Here's the place](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/viewmodel/posts/PostListCreateMenuViewModel.kt#L65-L82) where these items could switch their positioning. Would you run this question with a designer to double check? cc @kyleaparker  (ccing you Kylea but forgive me if you feel someone else should answer this question)

WordPress/src/main/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModel.kt
renanferrari: Oops, you're right! I was experimenting with some different solutions and somehow pushed the wrong version to the PR 🤦‍♂️ This was fixed in bf4a53b.

renanferrari:
Hey @mzorz 👋 Thanks a lot for the review!

>I think the order of the list could probably be reflected in the Posts list for consistency, wdyt?

I talked about this possibility with @leandroalonso and we decided to change only the main menu for now, since that was the one we experimented with during our A/B test. 

That said, I'm sure we can change the menu in the Posts list too, if we decide that's better. I see that the iOS PR is still open, so we can make sure we're consistent there as well: https://github.com/wordpress-mobile/WordPress-iOS/pull/16405

WordPress/src/main/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModel.kt
planarvoid: Thanks for the fix. Code-wise I don't think it's good that both the title and the action use the same object called `CreateAction`. I think that probably caused this error. Could we maybe add a test or two for these buttons?

mzorz: Thanks for the change @renanferrari! 🙇

mzorz (approved):
> I talked about this possibility with @leandroalonso and we decided to change only the main menu for now, since that was the one we experimented with during our A/B test.

Thanks for the context, that makes sense 👍 🙇 

> That said, I'm sure we can change the menu in the Posts list too, if we decide that's better. I see that the iOS PR is still open, so we can make sure we're consistent there as well: wordpress-mobile/WordPress-iOS#16405

Nice idea, I'd defer to a designer for proper validation, we could wait for @kyleaparker to chime in or we can do it later as well. Approving from my side as is 👍

kyleaparker:
>Also, I think the order of the list could probably be reflected in the Posts list for consistency, wdyt? Here's the place where these items could switch their positioning. Would you run this question with a designer to double check? cc @kyleaparker (ccing you Kylea but forgive me if you feel someone else should answer this question)

Hello! Sorry I don't have a lot of context on the reason for this change. What was the specific question you were looking for design feedback on?

mzorz:
> Hello! Sorry I don't have a lot of context on the reason for this change. What was the specific question you were looking for design feedback on?

Thank you @kyleaparker and sorry for the ping 🙏  - the order in which the options appear is changing on the My Site screen bottom sheet triggered from the FAB, from:
- Post 
- Page (if applicable)
- Story

To:
- Story
- Post
- Page (if applicable)

I was wondering if we should do the same in the Posts list FAB for consistency, where the order there currently goes like this:
- Post
- Story

To:
- Story
- Post


If we only merge this PR, then users will see a different order on each screen.

Hope that clarifies it, but let us know otherwise 👍

kyleaparker:
@mzorz thanks for clarifying! Yes I think they should definitely be in a consistent order 👍

mzorz:
Thank you @kyleaparker ! 🙇 🙏 

@renanferrari let me know when the change is ready for another look 👍 - I'll also link to the other PR as well so the changes remain consistent across platforms.

WordPress/src/main/java/org/wordpress/android/viewmodel/main/WPMainActivityViewModel.kt
renanferrari: @planarvoid Good idea! Added some tests in 8b70298.

renanferrari:
Thanks @mzorz and @kyleaparker! 🙇‍♂️ I've updated the order of the menu items in the Posts lists in 29df360.

mzorz (approved):
Tested, verified the lists match the intended order, and verified the unit tests run OK.
Thank you for your work @renanferrari 🙇
LGTM
:shipit:

PR #14528, opened 2021-04-26
Fixes Page Layout Picker not filtering layouts based on supported blocks

antonis:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/3417

`gutenberg-mobile`: https://github.com/wordpress-mobile/gutenberg-mobile/pull/3418
`gutenberg`: https://github.com/WordPress/gutenberg/pull/31184

To test:
* Follow testing steps on https://github.com/WordPress/gutenberg/pull/31184

## Regression Notes
1. Potential unintended areas of impact
None

2. What I did to test those areas of impact (or what existing automated tests I relied on)
N/A

3. What automated tests I added (or what prevented me from doing so)
N/A

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #14558, opened 2021-04-29
Add editor onboarding capability

dcalhoun:
* **Gutenberg:** https://github.com/WordPress/gutenberg/pull/31349
* **Gutenberg Mobile:** https://github.com/wordpress-mobile/gutenberg-mobile/pull/3437
* **WPiOS:** https://github.com/wordpress-mobile/WordPress-iOS/pull/16406

Add editor onboarding capability to control visibility of the feature. 

To test: Follow steps outlined in wordpress-mobile/gutenberg-mobile#3437.

## Regression Notes
  Modifications to build configuration could cause build failures. Modifications within Gutenberg editor lifecycle could cause editor to crash. 
  Ensured builds succeeded locally. Verified editor launched without error. 
  Includes no user-facing or critical routines to test. 

PR submission checklist:


WordPress/build.gradle
```diff
@@ -137,6 +139,8 @@ android {
         jalapeno { // Pre-Alpha version, used for PR builds, can be installed along release, alpha, beta, dev versions
             applicationIdSuffix ".prealpha"
             dimension "buildType"
+            // Used for development of editor onboarding, may be remove once finished
+            buildConfigField "boolean", "ENABLE_EDITOR_ONBOARDING", "true"
```
mchowning: Instead of adding a new `buildConfigField`, did you consider instead just checking if it's a debug build (`BuilcConfig.DEBUG` iirc)? Might be simpler. If you don't want to go that way, you might want to enable the editor onboarding for wasabi builds as well since that's what a lot of people use for their development builds (this is part of the reason I think just checking if it's a debug build might be simpler). 

Just a thought. This is **not** a blocker.

dcalhoun: > Instead of adding a new `buildConfigField`, did you consider instead just checking if it's a debug build (`BuilcConfig.DEBUG` iirc)? Might be simpler. 

I had not considered that, solely out of ignorance. It does seem leveraging `BuildConfig.DEBUG` would achieve what I want (enabling the feature for development purposes) in a simpler manner. Thanks for the idea!

build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
     ext.kotlin_ktx_version = '1.2.0'
     ext.wordPressUtilsVersion = '1.30.3-beta.3'
     ext.detektVersion = '1.15.0'
-    ext.gutenbergMobileVersion = 'v1.52.0'
+    ext.gutenbergMobileVersion = '3437-214e7b5314c4a30fc662493597a166aa00a957e2'
```
jhnstn: Should this revert back to pinned version before merging ?

dcalhoun: > Should this revert back to pinned version before merging ?

@jhnstn ultimately it will need to be updated to be the merge commit of wordpress-mobile/gutenberg-mobile#3437 into `develop` of `gutenberg-mobile`. An example of this can be found [here](https://github.com/wordpress-mobile/WordPress-Android/pull/14528/files). 

`ext.gutenbergMobileVersion` is currently set to `<PR-number>-<commit-sha>` to allow integration/testing of the unmerged work within wordpress-mobile/gutenberg-mobile#3437. More information on this practice can be found in [WPAndroid Integration](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/WPAndroid%20Integration.md).

During the release process, `ext.gutenbergMobileVersion` is set to the appropriate tag within `gutenberg-mobile`.

WordPress/build.gradle
hypest: And fwiw, verified that [Japapeno flavored builds for the installable builds are also `debug` variants](https://github.com/wordpress-mobile/WordPress-Android/blob/c7349f6734c230311bf0a1983f7b57ee40286b71/.circleci/config.yml#L130) so yeah, just checking for `BuildConfig.DEBUG` should be enough both for dev build as well as CI-built installable builds.

WordPress/src/main/java/org/wordpress/android/ui/prefs/AppPrefs.java
```diff
@@ -1186,6 +1189,14 @@ public static boolean shouldShowStoriesIntro() {
         return getBoolean(UndeletablePrefKey.SHOULD_SHOW_STORIES_INTRO, true);
     }
 
+    public static void setHasLaunchedGutenbergEditor(boolean hasLaunched) {
+        setBoolean(UndeletablePrefKey.HAS_LAUNCHED_GUTENBERG_EDITOR, hasLaunched);
```
mchowning, line 1193: nit: "has launched" feels slightly misleading to me because,based on that name, I would expect it to update to `true` as soon as Gutenberg is launched, not when it is closed. With that said, that distinction doesn't matter based on how we're using the variable, and I don't have a clearly better suggestion. The best I've come up with is something like "has used", which isn't great either. I don't view this as a blocker, just raising it in case you might have a better idea.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -2272,6 +2276,11 @@ private GutenbergPropsBuilder getGutenbergPropsBuilder() {
 
         boolean isFreeWPCom = mSite.isWPCom() && SiteUtils.onFreePlan(mSite);
 
+        int editorOnboardingPhasePercentage = 0;
```
mchowning: Since this variable is a constant, traditional java practice would say that we should make this a static variable [at the top of the file](https://github.com/wordpress-mobile/WordPress-Android/blob/04e523e0505945507852b57db0f57d752bcc6b1a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L297-L300) along these lines:

```
private static final int EDITOR_ONBOARDING_PHASE_PERCENTAGE = 0;
```

The reason for that is that it is slightly more memory efficient because it avoids the possibility of allocating the `int` multiple times. With that said, if you have reasons for not wanting to do that I certainly wouldn't consider this a blocker since this is an `Activity` class that isn't going to be instantiated a bunch of times.

mchowning (approved):
Looks good and is working well @dcalhoun . Nice work!

I left a couple of minor comments, but no blockers.

WordPress/src/main/java/org/wordpress/android/ui/prefs/AppPrefs.java
dcalhoun: The "has launched" name aligns with my original intent to set the value when the editor launched. However, I was unable to identify an appropriate [`Activity` lifecycle](https://developer.android.com/guide/components/activities/activity-lifecycle) that would run _after_ [`GutenbergPropsBuidler` was initialized](https://github.com/wordpress-mobile/WordPress-Android/blob/04e523e0505945507852b57db0f57d752bcc6b1a/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L2260-L2298) for the first time. Both `onCreate` and `onStart` appeared to run prior to the `GutenbergPropsBuilder`. 

The iOS implementation follows my original intent where [`gutenbergDidMount ` runs](https://github.com/wordpress-mobile/WordPress-iOS/blob/3e47fc112bcf8eb31192adff0847ac875d73310f/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift#L831-L841) after the initial call of [`gutenbergCapabilities`](https://github.com/wordpress-mobile/WordPress-iOS/blob/3e47fc112bcf8eb31192adff0847ac875d73310f/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift#L1048-L1060). So, the flag is truly set when the editor "launches" rather than when it closes. 

Are you aware of any lifecycles we might be able to hook into to make the "has launched" name continue to make sense? If not, do you believe it is better to retain the "has launched" name to match iOS or rename the Android name to more accurately represent the point at which it's set? 

I am also not dedicated to the "has launched" name on iOS. I think `hasUsedGutenbergEditor` could work for both Android and iOS, if you feel that is clearer.

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
dcalhoun: Ah, I was not aware of that ability or practice. I'll update the implementation to match your suggestion. Thank you!

WordPress/src/main/java/org/wordpress/android/ui/prefs/AppPrefs.java
mchowning: My initial thought was that we could just update the pref as soon as we send it off to the `GutenbergPropsBuilder`. That would ensure it was done as soon as the editor was started, but not until the initial value was sent to the editor.

Thinking about this some more, my concern with this approach is that the `GutenbergPropsBuilder` gets recreated and (with "new" props) when some things change ([see here](https://github.com/wordpress-mobile/WordPress-Android/blob/14d372ba5112f47300288e2fb0fff15f00039b5c/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java#L979)). If that happened, then I _think_ that if we made ^that^ suggested change, then the editor could start with showing the onboarding and then when some other prop changed so a new `GutenbergPropsBuilder` was used and sent across the bridge, the "new" value would hide onboarding (since it would think the editor had already been opened). Your current android implementation actually avoids that bug nicely, but I feel like it might occur on iOS (cc: @illusaen ).

illusaen: So when I tested it I didn't see the boolean flip even after I logged out and back in to reopen the editor.

dcalhoun: I spoke with @illusaen further and was able to confirm that neither one of us encountered an issue of the flag changing in the middle of the session for WPiOS. So, I believe we are safe to move forward with the current implementation in both WPAndroid and WPiOS.

I have discovered that WPAndroid includes a [`onEditorDidMount` callback](https://github.com/wordpress-mobile/WordPress-Android/blob/c43e8bc1f03b4bc8bba37f1b92b85aa2830f0726/libs/editor/WordPressEditor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java#L336-L353), which is similar to WPiOS' own `gutenbergDidMount`. So, I may open a follow up PR to more closely align the implementations, but in the meantime I plan to merge this work as the refactor shouldn't impact the functionality.

dcalhoun: Following up to share that I no longer plan to refactor this code. We ran into [race condition issues on iOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/16574/commits/5041436591d3869d4d43d4fab2cd7d01ca1877b0) where the `hasLaunchedGutenbergEditor` flag did, in fact, get set to the incorrect value during the middle of the first session. So, leveraging an end-of-life lifecycle hook seems most appropriate for this feature.

PR #14498, opened 2021-04-21
ExPlat: Sync implementation with latest updates from FluxC

renanferrari:
The latest ExPlat updates from FluxC introduces some breaking changes, so this PR simply makes sure we handle those so the app doesn't break. Another PR should follow this one introducing actual improvements to the current implementation.

I'll update this PR with a definitive FluxC version once https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1969 and https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1970 are merged.

To test:

Make sure the app builds.

## Regression Notes
1. Potential unintended areas of impact
None.

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #14089, opened 2021-02-16
ExPlat: Main Issue


renanferrari:
This issue centralizes all tasks and potential improvements to the ExPlat implementation on WPAndroid.

### Tasks
- [x] Initial A/A experiment implementation. (PR: #13960)
- [x] Initial A/B experiment implementation. (PR: #14286)
- [ ] Add support for logged-out experiments.
- [x] Add functionality to manage active experiments. (PR: #14559)
- [ ] Add UI to list active experiments (for internal testing only).
- [ ] Add functionality to manually edit assignments of active experiments (for internal testing only).
- [ ] Add support for other loading strategies (experiments that require loading screens, for example).
- [ ] Consider not fetching assignments if data collection is turned off.
- [ ] Extract implementation to a shared library (which one is yet TBD). (https://github.com/Automattic/Automattic-Tracks-Android/pull/114)

### Notes
- I'm going to update the tasks with links for the specific issues and PRs once they're up.

RenanLukas:
Issue moved to https://github.com/Automattic/Automattic-Tracks-Android/issues/116

PR #14559, opened 2021-04-29
ExPlat: Handle local Experiment registration

renanferrari:
This PR is a follow up to #14498 and is related to https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1970.

These changes can be divided into two parts:

- Part 1: Handling `Experiment` registration in the `ExPlat` class.
- Part 2: Providing a set of `Experiment`s to the `ExPlat` class for registration.

I thought about splitting this PR in two, but its changes are not so big overall and I think they make sense together, but let me know what you prefer.

#### Part 1

In this first part, the `ExPlat` class was updated in two ways:
- It now has a `start()` method that replaces the direct call we previously made to `forceRefresh()` once the app started. This method accepts a set of `Experiment`s, maps it to a list of experiment names which is then held internally and then calls `forceRefresh()`.
- The list of experiment names is then used so that we can check when we should retrieve assignments:
    - For `forceRefresh()` and `refreshIfNeeded()`, we check if the list is empty and if so, we prevent any calls to the API from being made.
    - For `getVariation()` we check if the list contains the name of the provided `Experiment` and if it doesn't, we just return `Control` directly (if `BuildConfig.DEBUG` is `true`, we throw a `IllegalArgumentException` instead).
    
#### Part 2

In this part, Dagger is being used to actually provide the set of `Experiment`s to the `ExPlat` class. Since our `Experiment` implementations were already being included in the dependency graph, I just created a `ExperimentModule` where we can add the experiments we want to register and bind them all into a `Set<Experiment>` by using a combination of `@Binds` and `@IntoSet` annotations. It also uses a `@Multibinds` annotation to provide an empty `Set` in case no `Experiment` is currently registered. 

This approach is a bit similar to the one we use to provide a `Map` of `ViewModel`s to the `ViewModelFactory` class and it introduces a similar extra step. To demonstrate, here's how it would work in practice to add a new `Experiment`:

1. We create a new class that extends from the `Experiment` class, and inject the `ExPlat` dependency using Dagger, just like we did before:

```kotlin
class ExampleExperiment @Inject constructor(exPlat: ExPlat) : Experiment("example_experiment", exPlat)
```

2. (new step) We then "register" this `Experiment` in the `ExperimentModule` like so:

```kotlin
@Binds @IntoSet fun exampleExperiment(experiment: ExampleExperiment): Experiment
```

And that's it. I also created an `ExampleExperiment` and left a commented out line in `ExperimentModule` with the code above as an example.

👉 Note: while this approach is pretty safe to use, I don't expect it to be final. Once we start working on adding functionality to list and edit registered experiments locally (as described in #14089), I expect we should be able to replace it with something similar to what we currently have with our remote configs, where we would just need to add an annotation to our `Experiment` class and the generated code would handle the rest for us. For now, other than that extra step, I don't see any major drawbacks that would prevent us from moving forward with this, so please let me know if you do!

### To test

**Without registered experiments**

1. Open the app.
1. On `logcat`, make sure you **don't** see any messages like: `ExPlat: fetching assignments successful with result: ...`.

Try logging in/out and restarting the app and make sure you still don't see the message.

**With registered experiments**

1. Uncomment the commented line in the `ExperimentModule` class and restart the app.
1. Open the app.
1. On `logcat`, notice a message like: `ExPlat: fetching assignments successful with result: ...`.

## Regression Notes
n/a

n/a

Added a few extra unit tests to account for new behaviors.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/WordPress.java
+++ b/WordPress/src/main/java/org/wordpress/android/WordPress.java
@@ -123,6 +123,7 @@ import org.wordpress.android.widgets.AppRatingDialog;
 import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.Field;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
@@ -366,7 +367,7 @@ public class WordPress extends MultiDexApplication implements HasAndroidInjector
         mStoryMediaSaveUploadBridge.init(this);
         ProcessLifecycleOwner.get().getLifecycle().addObserver(mStoryMediaSaveUploadBridge);
 
-        mExPlat.forceRefresh();
+        mExPlat.start(Collections.emptySet());
     }
 
     protected void initWorkManager() {
--- a/WordPress/src/main/java/org/wordpress/android/WordPress.java
+++ b/WordPress/src/main/java/org/wordpress/android/WordPress.java
@@ -185,7 +185,6 @@ public class WordPress extends MultiDexApplication implements HasAndroidInjector
     @Inject AppConfig mAppConfig;
     @Inject ImageEditorFileUtils mImageEditorFileUtils;
     @Inject ExPlat mExPlat;
-    @Inject Set<Experiment> mExperiments;
     @Inject @Named(APPLICATION_SCOPE) CoroutineScope mAppScope;
 
     // For development and production `AnalyticsTrackerNosara`, for testing a mocked `Tracker` will be injected.
@@ -369,7 +368,7 @@ public class WordPress extends MultiDexApplication implements HasAndroidInjector
         mStoryMediaSaveUploadBridge.init(this);
         ProcessLifecycleOwner.get().getLifecycle().addObserver(mStoryMediaSaveUploadBridge);
 
-        mExPlat.start(mExperiments);
+        mExPlat.forceRefresh();
     }
 
     protected void initWorkManager() {
--- a/WordPress/src/main/java/org/wordpress/android/WordPress.java
+++ b/WordPress/src/main/java/org/wordpress/android/WordPress.java
@@ -117,16 +117,17 @@ import org.wordpress.android.util.WPActivityUtils;
 import org.wordpress.android.util.analytics.AnalyticsUtils;
 import org.wordpress.android.util.experiments.ExPlat;
 import org.wordpress.android.util.config.AppConfig;
+import org.wordpress.android.util.experiments.Experiment;
 import org.wordpress.android.util.image.ImageManager;
 import org.wordpress.android.widgets.AppRatingDialog;
 
 import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.Field;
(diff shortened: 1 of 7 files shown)


WordPress/src/main/java/org/wordpress/android/WordPress.java
```diff
@@ -183,6 +185,7 @@
     @Inject AppConfig mAppConfig;
     @Inject ImageEditorFileUtils mImageEditorFileUtils;
     @Inject ExPlat mExPlat;
+    @Inject Set<Experiment> mExperiments;
```
planarvoid: I'm wondering whether it would make sense to inject this directly in the `ExPlat` class. The reason is that it would hide this internal piece of logic in there instead of using it in the shared application class. I see it's a singleton but I wouldn't expect it to outlive the application class. WDYT?

WordPress/src/main/java/org/wordpress/android/modules/ExperimentModule.kt
```diff
@@ -0,0 +1,13 @@
...
+import dagger.Module
+import dagger.multibindings.Multibinds
+import org.wordpress.android.util.experiments.Experiment
+
+@Module
+interface ExperimentModule {
+    @Multibinds fun experiments(): Set<Experiment>
+
+    // Copy and paste the line below to add a new experiment.
+    // @Binds @IntoSet fun exampleExperiment(experiment: ExampleExperiment): Experiment
```
planarvoid, line 12: Is it viable to introduce some kind of a lint rule that would fail if there is an experiment that is not registered in the module? Maybe that's unnecessary.

planarvoid (commented):
It's looking good and works well, thanks for the PR! I have one comment 👍

WordPress/src/main/java/org/wordpress/android/WordPress.java
renanferrari: That would make sense, for sure. The reason I avoided doing that in the first place was because I didn't want to generate a circular dependency (`ExPlat` depends on `Experiment` that depends on `ExPlat`).

That said, I tried using `dagger.Lazy` to inject the `Set<Experiment>` in the `ExPlat` constructor and it actually ended up working pretty well and the code got cleaner, specially because we no longer need the `ExPlat::start` method. 

You can check the changes here: 3eb27b7 – thanks for the suggestion!

WordPress/src/main/java/org/wordpress/android/modules/ExperimentModule.kt
renanferrari: This is probably viable, although I have no experience with custom lint rules so I can't say for sure.

That said, if that's something we want to do, we should probably do it in a different PR (specially since I guess it would need to be submitted to [WordPress-Lint-Android](https://github.com/wordpress-mobile/WordPress-Lint-Android)?).

planarvoid (approved):
Thanks for the changes, it's looking good 👍

PR #14561, opened 2021-04-29
[WP Stories] Fix stories error snackbar handling

mzorz:
Fixes a bug introduced in https://github.com/wordpress-mobile/WordPress-Android/pull/13533 that would make the StoryComposerActivity be launched automatically when a Story save process fails, right after presenting the error snackbar in the MySite Fragment as the callback was replicated in its `onDismiss` lambda function.

This PR:
~- removes the remaining code from MySiteFragment that should have been removed as part of https://github.com/wordpress-mobile/WordPress-Android/pull/13533~
- makes the  `onDismissAction` and `buttonAction` parameters to the `SnackbarMessageHolder` class explicit, and defines the behavior correctly on the `buttonAction` instead of letting it be assigned to the `onDismissAction` implicitly. 

**To test:**
1. apply this [patch](https://gist.github.com/mzorz/41d6f80f2395105fff6da6328b190ff5) on the stories library code, which will provoke an error in saving video slides each time.
2. run WPAndroid on this branch, and on the My Site screen tap on the FAB -> Add new Story post
3. record a video or select an existing video with the media picker
4. optionally add a text to it
5. publish it
6. wait on the My site screen until the `Unable to upload <post title> - 1 slide requires action - MANAGE` Snackbar.
7. wait for it to be dismissed, and observe you stay on the main screen, instead of being presented with the Story composer screen, as if you'd have pressed the `MANAGE` button on the snackbar.

## Regression Notes
N/A only the Stories actions are affected

N/A

N/A

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/mysite/SiteStoriesHandler.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/mysite/SiteStoriesHandler.kt
@@ -68,21 +68,18 @@ class SiteStoriesHandler
                     Event(
                             SnackbarMessageHolder(
                                     UiStringText(snackbarMessage),
-                                    UiStringRes(R.string.story_saving_failed_quick_action_manage)
-                            ) {
-                                val selectedSite = selectedSiteRepository.getSelectedSite()
-                                        ?: return@SnackbarMessageHolder
-                                _onNavigation.postValue(Event(OpenStories(selectedSite, event)))
-                                // TODO WPSTORIES add TRACKS: the putExtra described here below for NOTIFICATION_TYPE
-                                // is meant to be used for tracking purposes. Use it!
-                                // TODO add NotificationType.MEDIA_SAVE_ERROR param later when integrating with WPAndroid
-                                //        val notificationType = NotificationType.MEDIA_SAVE_ERROR
-                                //        notificationIntent.putExtra(ARG_NOTIFICATION_TYPE, notificationType)
-                                storiesTrackerHelper.trackStorySaveResultEvent(
-                                        event,
-                                        STORY_SAVE_ERROR_SNACKBAR_MANAGE_TAPPED
-                                )
-                            }
+                                    UiStringRes(R.string.story_saving_failed_quick_action_manage),
+                                    buttonAction = {
+                                        val selectedSite = selectedSiteRepository.getSelectedSite()
+                                                ?: return@SnackbarMessageHolder
+                                        _onNavigation.postValue(Event(OpenStories(selectedSite, event)))
+                                        storiesTrackerHelper.trackStorySaveResultEvent(
+                                                event,
+                                                STORY_SAVE_ERROR_SNACKBAR_MANAGE_TAPPED
+                                        )
+                                    },
+                                    onDismissAction = { }
+                            )
                     )
             )
         }
--- a/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/main/MySiteFragment.kt
@@ -1188,7 +1188,6 @@ class MySiteFragment : Fragment(R.layout.my_site_fragment),
                     allErrorsInResult(event.frameSaveResult).size,
                     errorText
             )
-
             uploadUtilsWrapper.showSnackbarError(
                     requireActivity().findViewById<View>(R.id.coordinator),
                     snackbarMessage,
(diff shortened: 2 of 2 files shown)


aforcier:
@mzorz I gave this a try. I initially had a hard time reproducing the issue, here's what I figured out:

* The issue only happens for the new My Site fragment.
* As far as I can tell, the new My Site fragment is still off by default.
* Merely enabling `MY_SITE_IMPROVEMENTS` in `build.gradle` is not enough, as [this randomizes based on your token](https://github.com/wordpress-mobile/WordPress-Android/blob/cb67bcdf11f90d4ccfdd3cde9138a8c8358ff13e/WordPress/src/main/java/org/wordpress/android/util/config/MySiteImprovementsFeatureConfig.kt#L42). There's probably a config I overlooked somewhere, but I ended up just forcing `isEnabled()` to true to test.

Once I made those changes I was able to reproduce the original issue with your patch applied, and check that it's resolved when checking out this branch (for the new My Site fragment).

However if I understand correctly, we shouldn't remove the subscriber from `MySiteFragment`. `SiteStoriesHandler` is only used by `MySiteViewModel`, which in turn is only used in the `MY_SITE_IMPROVEMENTS == true` case (via `ImprovedMySiteFragment`). So the two subscribers are never both registered for the same build of the app - and so we need the block in both places until we stop using `MySiteFragment` entirely. Otherwise there's no toast shown on error when using the old `MySiteFragment` with this branch. Let me know if I'm overlooking something 🙏

WordPress/src/main/java/org/wordpress/android/ui/mysite/SiteStoriesHandler.kt
```diff
@@ -68,21 +68,18 @@ class SiteStoriesHandler
...
-                                // TODO add NotificationType.MEDIA_SAVE_ERROR param later when integrating with WPAndroid
-                                //        val notificationType = NotificationType.MEDIA_SAVE_ERROR
-                                //        notificationIntent.putExtra(ARG_NOTIFICATION_TYPE, notificationType)
-                                storiesTrackerHelper.trackStorySaveResultEvent(
-                                        event,
-                                        STORY_SAVE_ERROR_SNACKBAR_MANAGE_TAPPED
-                                )
-                            }
+                                    UiStringRes(R.string.story_saving_failed_quick_action_manage),
+                                    buttonAction = {
```
aforcier, line 72: This was interesting, hadn't thought about this danger with trailing lambdas before. 😅

mzorz:
Thanks for the thorough testing @aforcier 🙇 

> However if I understand correctly, we shouldn't remove the subscriber from MySiteFragment.
>  So the two subscribers are never both registered for the same build of the app - and so we need the block in both places until we stop using MySiteFragment entirely. Otherwise there's no toast shown on error when using the old MySiteFragment with this branch. Let me know if I'm overlooking something 

Thanks for this @aforcier  - I could swear I saw both being run, but it may have been a glitch in between builds / branch switches.
Once I verified the MySite improvements work is behind a FeatureConfig flag (to the extent current [`develop`](https://github.com/wordpress-mobile/WordPress-Android/blob/cb67bcdf11f90d4ccfdd3cde9138a8c8358ff13e/WordPress/src/main/java/org/wordpress/android/util/config/MySiteImprovementsFeatureConfig.kt) shows), I returned the code back in `MySiteFragment` in c39ec2de55 👍 .

>  I initially had a hard time reproducing the issue, here's what I figured out:

> The issue only happens for the new My Site fragment.

True 👍 

> As far as I can tell, the new My Site fragment is still off by default.

I did some digging - it looks like `BuildConfig` false flag from [build.gradle](https://github.com/wordpress-mobile/WordPress-Android/blob/cb67bcdf11f90d4ccfdd3cde9138a8c8358ff13e/WordPress/build.gradle#L77) is totally [overlooked](https://github.com/wordpress-mobile/WordPress-Android/blob/cb67bcdf11f90d4ccfdd3cde9138a8c8358ff13e/WordPress/src/main/java/org/wordpress/android/util/config/MySiteImprovementsFeatureConfig.kt#L25) in current code, so from what I understand there’s no way the “old” code was run except for a brief amount of time in between installing and fetching things from the remote config, if it was the case that one's token ended up marking it false at first and then was overridden by the fetched remote firebase config. Hence there's what I can draw as a potential explanation as to why I saw the old code there being run at some earlier test.

> Merely enabling MY_SITE_IMPROVEMENTS in build.gradle is not enough, as this randomizes based on your token. There's probably a config I overlooked somewhere, but I ended up just forcing isEnabled() to true to test.

Also true. In my case I had to force `isEnabled()` to `false` as well 😬 .

I think it boils down to `BuildConfig` being overlooked, seems this was introduced [here](https://github.com/wordpress-mobile/WordPress-Android/pull/14452).

In any case, this should be ready for another look 👍

aforcier (approved):
Makes sense @mzorz - :shipit: !

PR #14568, opened 2021-04-30
Add retry to get expected like state

develric:
This PR adds a retry mechanism in the getLikes function. This is currently used in the reader post details when a user is liking/unliking a post and should appear/disappear from the likers train of faces. I'll add some more unit testing in a separate PR (also trying to see if we can even more simplify the getLikes function 🙃 ).

### To test
- Use the debugger to check there can be cases where one or more retry is attempted (especially when liking to get the user included in the likers response)
- Bring the NUM_RETRY to 1 and check there are cases when the user liked but it's not included in the list of likers. This tests that in case the NUM_RETRY is reached the logic still present a result in the end (showing what is possible to show)
- Smoke test both notifications and reader

## Regression Notes
N/A still behind feature flag

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/engagement/GetLikesUseCase.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/engagement/GetLikesUseCase.kt
@@ -21,9 +21,9 @@ import org.wordpress.android.fluxc.store.PostStore
 import org.wordpress.android.fluxc.store.PostStore.FetchPostLikesPayload
 import org.wordpress.android.fluxc.store.PostStore.OnPostLikesChanged
 import org.wordpress.android.fluxc.store.Store.OnChanged
-import org.wordpress.android.ui.engagement.GetLikesUseCase.BeInListRequirement.DON_T_CARE
-import org.wordpress.android.ui.engagement.GetLikesUseCase.BeInListRequirement.REQUIRE_TO_BE_THERE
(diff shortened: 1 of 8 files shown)


WordPress/src/main/java/org/wordpress/android/ui/engagement/GetLikesUseCase.kt
```diff
@@ -255,8 +308,17 @@ class GetLikesUseCase @Inject constructor(
     data class LikeGroupFingerPrint(val siteId: Long, val postOrCommentId: Long, val expectedNumLikes: Int)
     data class PaginationParams(val requestNextPage: Boolean, val pageLength: Int, val limit: Int)
 
+    enum class BeInListRequirement {
```
khaykov: Wonder if we can clarify this with something like `LikerInListRequirement` ?

WordPress/src/main/java/org/wordpress/android/ui/engagement/GetLikesUseCase.kt
```diff
@@ -255,8 +308,17 @@ class GetLikesUseCase @Inject constructor(
...
+        DON_T_CARE,
```
khaykov: I think we can do with `DONT_CARE` :)

khaykov (approved):
Works pretty well. Max I got is 4 retries for posts with a large number of likes :)

Also tested with liking on the web and then in the app - no issues 👍 

Just left a couple of semantic nitpicks :)

khaykov (approved):
Just for the context - all the PR's merged into this one were already reviewed 🚀

PR #14569, opened 2021-04-30
Fixing issue with multiple tap on same liker.

develric:
This is a quick PR to fix a possible crash when a user taps very quickly and multiple times on a liker in the list of likers. This is due since the show on the bottom sheet should not be actually called when the bottom sheet isn't null.

### To test
- Try to tap multiple times on the same liker and check the logcat for the following crash `java.lang.IllegalStateException: Fragment already added: UserProfileBottomSheetFragment`
- Smoke test to check the standard expected behaviour of the user profile bottom sheet is kept.

## Regression Notes
N/A behind feature flag.

PR submission checklist:


khaykov (approved):
I confirmed that the crash is gone 👍

PR #14573, opened 2021-04-30
Allow opening header blog by id and site url.

develric:
This PR allows to open the author site both by url (that is in a webview) or site id (that is in the reader) when tapping on the author avatar in the likers list header.

### To test
- In the reader, open a list of likers and tap on the header avatar. Check the author site is opened in the reader (since it was already fetched in the reader db)
- In the notifications, open a list of likers and tap on the header avatar. Majority of cases should open the author site in a webview (since it was not fetched already in the reader db) where before this patch nothing would have opened (given the site id would have been 0 even if the site url was available)

## Regression Notes
N/A behind feature flag.

PR submission checklist:


khaykov (approved):
Works well! I confirmed the original issue (header avatar not reacting to taps) and that this PR fixed it 👍

PR #14574, opened 2021-04-30
Avoid issues with negative site id

develric:
During testing we found that some reader post author site id could be negative. This PR enforce the logic checks to avoid issues with eventual negative site ids.

### To test
- Logic inspection of the code
- Force condition by debugger (or I can provide a couple of sample posts where this happen as well)

## Regression Notes
N/A behind feature flag.

PR submission checklist:


khaykov (approved):
Looks good! 👍

PR #14576, opened 2021-05-01
Merge release 17.2 into develop

JavonDavis:


oguzkocer (changes requested):
@JavonDavis It looks like `libs/stories-android` hash was updated in 2e5aeb992f699fbee140d43f4b0b35d4a2e22a57. Since there are no relevant PRs in the 4 commits this PR is merging, that's a curious change because even if there is a merge conflict we'd typically keep the hash from `develop`. Since there is no merge commit either, I am not entirely sure what's going on. Could you maybe look into it?

_P.S: I am marking this review as `Request changes` just to make sure it doesn't go unnoticed._

JavonDavis:
Thanks @oguzkocer ! Yep you're right looks like this got messed up resolving the merge! That's fixed now

oguzkocer (approved):

PR #14579, opened 2021-05-01
[WP Stories] finish the StoryComposerActivity when discarding a Story with errored frames

mzorz:
The changes in this PR aim at making sure to properly finish the StoryComposerActivity when the user decides to discard a Story with errored frames.
I observed after retrying several times and then choosing to discard, it would be stuck on the Activity given the StoryRepository would at some point load a new empty Story (effectively discarding the previous loaded one) but not being able to move forward. While there may be other causes for this, I thought it would be enough and more broad-covering to take note of whether the composer has been called with a StorySaveResult (which indicates it's in error handling mode) originally, and then perform the right action in the end, if such original StorySaveResult exists.

To test:
1. use this PR at the stories library https://github.com/Automattic/stories-android/pull/673
2. apply this [patch](https://gist.github.com/mzorz/41d6f80f2395105fff6da6328b190ff5) to the stories library
3. create a new story
4. add a video and publish
5. observe the error notification come up, tap on it to enter error handling mode in the StoryComposer
6. now tap retry several times (wait each time until it's tappable again)
7. tap the X on the top-left corner of the screen and discard the Story
8. observe you land in the My Sites screen.

## Regression Notes
N/A

StoryComposerViewModelTest (made sure to update and let them all pass)

Added one test that checks that onStoryDiscarded returns true if a originalSaveResult is passed and is a retry

PR submission checklist:


aforcier (approved):
Works well! :shipit: (Though it looks like there's a conflict with the stories lib hash which will need to be fixed first - latest `develop` probably makes sense to include https://github.com/Automattic/stories-android/pull/679.)

PR #14581, opened 2021-05-03
Fix margins on site info block

### LINKED ISSUE #14580: Fix cropped URL in My Site info
If the site title is long and wraps to a second line, the URL gets cropped in the my site info .

Screenshot_1620034376

planarvoid:
Fixes #14580 

This PR fixes the look of the site info block with the system font set to the largest possible value. In that case the site URL was overflowing the text block. Previously we were using top margin on both the icon and the site title, the bottom margin was set on the parent. This solution moves the 24dp margin to the bottom of the icon and removes all the margin from the title/subtitle. The title/subtitle view is centered so it looks good and has space to grow

To test:
- Make sure `MySiteImprovements` flag is turned on
- Set system font to largest value
- Go to a site with long site title (which wraps 2 lines)
- Make sure the site title and the URL are both visible
- Make sure you test other configurations (font sizes, shorter/longer site titles)

Before:

Screenshot_1620034376

After: 

Screenshot 2021-05-03 at 13 38 16

## Regression Notes
- Unexpected UI of the site title

- I've tested with multiple system font sizes/site titles

- There is no way to test this automatically

PR submission checklist:

### DIFF
--- a/WordPress/src/main/res/layout/my_site_info_block.xml
+++ b/WordPress/src/main/res/layout/my_site_info_block.xml
@@ -4,7 +4,7 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
-    android:layout_marginBottom="@dimen/margin_extra_extra_medium_large"
+    android:layout_marginBottom="@dimen/margin_medium"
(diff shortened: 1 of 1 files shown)


renanferrari (approved):
Thanks for the changes!

LGTM 👍

PR #14594, opened 2021-05-05
Use ConstraintLayout in quick actions block

### LINKED ISSUE #14591: Quick actions overlap to second line with largest font
Quick actions on the main site weirdly overlap to the second line even though they have enough space to grow to. As part of this issue check whether they are increasing the font size correctly

Screenshot 2021-05-04 at 14 31 57

planarvoid:
Fixes #14591 

This fix is a little bit tricky. The main thing that fixes the overlap with largest font is that the the label now has maxWidth 20dp wider than the icon. This causes the font sometimes overlap the icon and the quick actions block padding. However, I think this is ok since it only happens with the largest system font and the result looks really good. In order to implement this I've converted the quick actions block to constraint layout.

To test:
- Make sure the my site improvements flag is turned on
- Change system font to "Largest"
- Switch to a site with all of 4 actions 
- Check the font is not overlapping to the second line
- Switch to a site with all of 3 actions (where you're not an admin)
- Check the actions look as expected
- Change the system font back to standard
- Check the quick actions labels look like they did before both with 4 and 3 actions

Before: 

Screenshot 2021-05-05 at 11 31 29
Screenshot 2021-05-05 at 11 31 17
Screenshot 2021-05-05 at 11 30 41
Screenshot 2021-05-05 at 11 30 17

After: 

Screenshot 2021-05-05 at 11 32 20
Screenshot 2021-05-05 at 11 32 11
Screenshot 2021-05-05 at 11 31 54
Screenshot 2021-05-05 at 11 31 42

## Regression Notes
- I think there are no unintended areas of impact 

- I've tested everything mentioned in "To test" block and compared to the current develop

- N/A
 
PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/mysite/QuickActionsViewHolder.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/mysite/QuickActionsViewHolder.kt
@@ -16,8 +16,8 @@ class QuickActionsViewHolder(
         quickActionPagesButton.setOnClickListener { item.onPagesClick.click() }
 
         val pagesVisibility = if (item.showPages) View.VISIBLE else View.GONE
-        quickActionPagesContainer.visibility = pagesVisibility
-        middleQuickActionSpacing.visibility = pagesVisibility
+        quickActionPagesButton.visibility = pagesVisibility
+        quickActionPagesLabel.visibility = pagesVisibility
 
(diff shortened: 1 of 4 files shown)


renanferrari (approved):
Thanks for the changes. Code is looking good and everything works as expected.

PS: this used to be a problem even with the regular font in Portuguese, so I'm glad it finally got fixed! 🙂

LGTM 👍

PR #14596, opened 2021-05-05
Jetpack App (Basics): No Jetpack Sites Empty View

zwarm:
Parent #14302 

This PR implements the empty state UI for logging into the Jetpack app when the account has no jetpack sites.

Implementation Details
- Two large vector assets were added, don't let the number of lines cause panic. :)
- The colors in the `img_jetpack_empty_state` were changed to semantic where possible. I considered adding custom colors to the color file, but since they are not common I decide to leave them in the drawable itself. This follows the same pattern as the other illustrations. I'm opening to changing this if need be.
- The word `Error` was dropped from the fragment, viewmodel and navigationevent.  I was on the fence about including `Jetpack` and would love thoughts on whether it should be included. Also all files were left in WordPress source sets. This can be moved and would love opinions. 
- Refactored the way binding was being handled in `LoginNoSitesFragment`
- Removed the errorMessage from the `ShowNoJetpackSites` navigation event since it never changes and is set in the layout xml

To test:
- Download this branch and build a Jetpack App variant
- Tap the 'Login with WordPress' button
- Enter WordPress.com credentials for an account that is not linked to any Jetpack sites
- After logging in you should be brought to the No Jetpack Sites Found empty view

## Regression Notes

Manual and automated tests

Added new unit tests in `LoginNoSitesViewModelTest`

PR submission checklist:


osullivanchris (approved):
Works as expected and looks great.

Tried with light mode and dark mode. Buttons took me where I expected. Layout looks good.

Only issue I spotted was one time I logged in and didn't see a profile image above my name. Maybe it took a moment to load or something. I changed colour mode on the system settings, and it rejigged it and I saw it. If we are using code from the epilogue maybe this is pre-existing.

zwarm:
> Works as expected and looks great.
> 
> Tried with light mode and dark mode. Buttons took me where I expected. Layout looks good.
> 
> Only issue I spotted was one time I logged in and didn't see a profile image above my name. Maybe it took a moment to load or something. I changed colour mode on the system settings, and it rejigged it and I saw it. If we are using code from the epilogue maybe this is pre-existing.

Thanks for the feedback, I'll take a look. This code is brand new, so no pre-existing issues. It may just be a network lag when grabbing the avatar. I wonder if we should put a progress spinner on that. @osullivanchris wdyt?

osullivanchris:
> Thanks for the feedback, I'll take a look. This code is brand new, so no pre-existing issues. It may just be a network lag when grabbing the avatar. I wonder if we should put a progress spinner on that. @osullivanchris wdyt?

Ah ok. Let me know if you are able to reproduce/see the issue. I think it only happened to me once. But maybe it was cached after that, and this is relevant on first user experience.

As it happens, because it was a test account, I actually had no image on the account. So the image it was loading in was a placeholder style one. But maybe, that is tied to my account as an actual image and has to be loaded like any other.

What I was thinking is rather than a spinner we could show an empty profile image to everyone, or even just a grey circle, while the real image loads in on top of it. If thats getting complicated, a spinner in that location is also an option.

zwarm:
> What I was thinking is rather than a spinner we could show an empty profile image to everyone, or even just a grey circle, while the real image loads in on top of it. If thats getting complicated, a spinner in that location is also an option.

@osullivanchris - I pushed a change to show the empty profile just like the email password view does. Let me know how this looks works for you.

osullivanchris:
> @osullivanchris - I pushed a change to show the empty profile just like the email password view does. Let me know how this looks works for you.

Oh yeah that makes sense. Nice one. Always good to re-use an existing solution.

WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
```diff
@@ -962,9 +964,9 @@ private void showSiteAddressError(ShowSiteAddressError event) {
         slideInFragment(fragment, true, LoginSiteCheckErrorFragment.TAG);
     }
 
-    private void showNoJetpackSitesError(ShowNoJetpackSitesError event) {
-        LoginNoSitesErrorFragment fragment =
-                LoginNoSitesErrorFragment.Companion.newInstance(event.getErrorMessage());
-        slideInFragment(fragment, false, LoginNoSitesErrorFragment.TAG);
+    private void showNoJetpackSites() {
+        LoginNoSitesFragment fragment =
```
zwarm, line 966: FYI: This was moved to a single line in the next PR #14600

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginNoSitesFragment.kt
```diff
@@ -0,0 +1,152 @@
...
+    @Inject lateinit var meGravatarLoader: MeGravatarLoader
+    @Inject lateinit var uiHelpers: UiHelpers
+    private var loginListener: LoginListener? = null
+    private lateinit var viewModel: LoginNoSitesViewModel
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        initDagger()
+        initBackPressHandler()
+        with(FragmentLoginNoSitesEmptyViewBinding.bind(view)) {
```
ParaskP7: Praise (❤️): I really like what you did here, using extension functions with view bindings like that.

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginNoSitesFragment.kt
```diff
@@ -0,0 +1,152 @@
...
+                    loginEmptyViewUserSection.visibility = View.VISIBLE
+                }
+                is NoUser -> loginEmptyViewUserSection.visibility = View.GONE
+            }
+        })
+    }
+
+    private fun FragmentLoginNoSitesEmptyViewBinding.loadGravatar(avatarUrl: String) {
+        val url = meGravatarLoader.constructGravatarUrl(avatarUrl)
+        AvatarHelper.loadAvatarFromUrl(this@LoginNoSitesFragment, url, imageAvatar, object : AvatarRequestListener {
```
ParaskP7: Suggestion (💡): Since now we are at point were the 'loadAvatarFromUrl(...)` helper function doesn't need to `AvatarRequestListener`, I suggest making that optional and enhance this helper function with making the `AvatarRequestListener` nullable and default to `null`. This way this whole block can then be transformed to something more readable:
```
        AvatarHelper.loadAvatarFromUrl(
                fragment = this@LoginNoSitesFragment,
                avatarUrl = meGravatarLoader.constructGravatarUrl(avatarUrl),
                avatarView = imageAvatar
        )
```

Wdyt?

WordPress/src/main/java/org/wordpress/android/ui/accounts/LoginActivity.java
```diff
@@ -219,8 +219,10 @@ private void initViewModel() {
             LoginNavigationEvents loginEvent = event.getContentIfNotHandled();
             if (loginEvent instanceof ShowSiteAddressError) {
                 showSiteAddressError((ShowSiteAddressError) loginEvent);
+            } else if (loginEvent instanceof ShowNoJetpackSites) {
+                showNoJetpackSites();
             } else {
-                showNoJetpackSitesError((ShowNoJetpackSitesError) loginEvent);
+                // NO OP
```
ParaskP7: Minor (🔍): This `else` block is effectively not needed any more. Consider removing it altogether.

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
```diff
@@ -0,0 +1,126 @@
...
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toTopOf="parent"
+            app:constraint_referenced_ids="image_avatar,text_username, text_displayname"
+            tools:visibility="visible"/>
+
+        <ImageView
+            android:id="@+id/image_avatar"
+            android:layout_width="@dimen/login_empty_view_avatar_width"
+            android:layout_height="@dimen/login_empty_view_avatar_height"
+            android:layout_marginBottom="@dimen/login_empty_view_avatar_margin_bottom"
```
ParaskP7, line 31: Question (❓): The design specifies that this is `16dp` instead. Am I missing something?

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
```diff
@@ -0,0 +1,126 @@
...
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:paddingBottom="@dimen/margin_extra_medium_large"
+            app:layout_constraintBottom_toTopOf="@+id/login_empty_view_illustration"
+            app:layout_constraintStart_toStartOf="@+id/text_displayname"
+            app:layout_constraintTop_toBottomOf="@+id/text_displayname"
+            tools:text="\@droidtester2021" />
+
+        <com.google.android.material.textview.MaterialTextView
+            android:id="@+id/text_displayname"
```
ParaskP7: 1) Minor (🔍): I suggest reordering this view and moving that on top of the `text_username`, code wise, just to keep the design order so that it makes it easier to read through. I was confused initially while reading this layout file.
2) Minor (🔍): I suggest renaming the `android:id` to `text_display_name` as the current naming results in a typo.
3) Question (❓): The design specifies that there should exist a `4dp` bottom margin. Am I missing something?

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
```diff
@@ -0,0 +1,126 @@
...
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toBottomOf="@+id/login_empty_view_user_section"/>
+
+        <com.google.android.material.textview.MaterialTextView
+            android:id="@+id/login_error_message_title"
+            style="@style/ActionableEmptyStateTitle"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:layout_marginBottom="@dimen/margin_medium"
```
ParaskP7, line 77: Question (❓): The design specifies that this is `6dp` instead. Am I missing something?

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
```diff
@@ -0,0 +1,126 @@
...
+
+        <com.google.android.material.textview.MaterialTextView
+            android:id="@+id/login_error_message_text"
+            style="@style/ActionableEmptyStateSubtitle"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:gravity="center_horizontal"
+            android:lineSpacingExtra="@dimen/login_empty_view_line_spacing"
+            android:text="@string/login_no_jetpack_sites_error_message"
+            android:paddingBottom="@dimen/margin_extra_large"
```
ParaskP7: 1) Question (❓): Why did you use `padding` instead of `margin`?
2) Question (❓): From the design I am seeing that the constraints of this and the above `login_error_message_title` view should fill the parent (left and right). I am seeing that this is not done because of the `ActionableEmptyStateTitle` and `ActionableEmptyStateSubtitle`.  If you remove those you will observe that too. I am actually skeptical about reusing those styles, as those are `Help Activity Styles` it seems. Also, those styles contain `android:layout_xyz` specifics, which make those harder to reuse. Maybe we should start separating all that now and design these styles in a way that will not collide with the current ones or have any `android:layout_xyz` specifications. Wdyt?

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
```diff
@@ -0,0 +1,126 @@
...
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toBottomOf="@+id/login_error_message_text" />
+
+        <com.google.android.material.button.MaterialButton
+            android:id="@+id/button_secondary"
+            style="@style/Widget.LoginFlow.Button.Secondary"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:layout_marginBottom="@dimen/margin_extra_large"
```
ParaskP7, line 117: Question (❓): Is this bottom margin needed here?

WordPress/src/test/java/org/wordpress/android/ui/accounts/login/LoginNoSitesViewModelTest.kt
```diff
@@ -0,0 +1,192 @@
...
+    fun `given started, then application sign out is invoked`() {
+        setupAccountItem()
+
+        startViewModel()
+
+        verify(wordPress).wordPressComSignOut()
+    }
+
+    @Test
+    fun `given savedInstanceState is NoUser, when started, then state is recreated as NoUser`() {
```
ParaskP7: 1) Suggestion (💡): I would avoid using code specific language on test sentences, like classes or camel-case function references.
2) Question (❓): Since the `setupInstanceStateForNoUser()` and `setupInstanceStateForShowUser()` are only used once and are actually one-liners (ish), maybe extracting those to a method might convey that they are not. Did you do that just so that the setup becomes more obvious to the reader?
3) Minor (🔍): Consider moving the `Assertions.` to imports.

WordPress/src/test/java/org/wordpress/android/ui/accounts/login/LoginNoSitesViewModelTest.kt
```diff
@@ -0,0 +1,192 @@
...
+        startViewModel()
+
+        viewModel.writeToBundle(savedInstanceState)
+
+        verify(savedInstanceState).putSerializable(any(), argThat { this is ShowUser })
+    }
+
+    private val userName = "username"
+    private val displayName = "displayname"
+    private val avatarUrl = "avatarurl"
```
ParaskP7: Minor (🔍): Consider making those constants instead.

ParaskP7 (changes requested):
👋 @zwarm !

However, when testing, and although everything worked as expected, I noticed that when I click the `Try with another account` button, and then use the same account to try again, one that doesn't have any Jetpack sites connected to it, then instead of getting the `No Sites` empty screen again, I get redirected straight to the initial login screen, as if I automatically clicked the `Try with another account` button. I actually can see the `No Sites` empty screen shown for a split of a second and then I get redirected.

> The word Error was dropped from the fragment, viewmodel and navigationevent. I was on the fence about including Jetpack and would love thoughts on whether it should be included. Also all files were left in WordPress source sets. This can be moved and would love opinions.

I would indeed suggest to included the `Jetpack` everywhere that it makes sense, and actually, I would go a step further and restructure the packaging structure of `org.wordpress.android.ui.accounts`. As such, I would:
- At least, create a `jetpack` package within the `login` package and move the `Jetpack` related classes in it. This will make it obvious that those are used for Jetpack only (at least atm). Also, it might make it easier to potentially move them later on, to a different flavor or repo. So, whatever is being used through the `LoginActivity.showNoJetpackSites()`, I would move all that functionality into the newly created `jetpack` package and include the `Jetpack` naming on all those classes/files.
- Also, move the `Login` related classes to the `login` package. Some of them are still on the `accounts` package level and I am not sure why.

I have also left some minor (🔍) comments, few questions (❓) and some suggestions (💡) for you to consider. I am **NOT** going to approve this PR, will request changes for now, but to be honest it is not that something is blocking. It just that there are things to discuss.

Also, on the design level, I wasn't sure how to review the design vs. what I was seeing code wise, thus I left a handful of questions. Maybe you already discussed all that with Chris.

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginNoSitesFragment.kt
zwarm: I had thought about this, but didn't want to introduce null when others rely on it being non-nullable. Plus it's a Login library change and the risk was lower to use an empty handler. If you feel strongly, then I don't mind making the adjustment, just wanted you to be aware of why I didn't do it originally. @ParaskP7

WordPress/src/test/java/org/wordpress/android/ui/accounts/login/LoginNoSitesViewModelTest.kt
zwarm: > Did you do that just so that the setup becomes more obvious to the reader?

TBH, I wrote those methods first and then they only got used once, but I left it because it's clearer to read the test with them then without.

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
zwarm: @ParaskP7  - there were no specs for this view ... so this is based this off of Woo and then adjusted to match what Chris and I discussed. Where are you seeing a spec sheet for this?

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
zwarm: > 1. Minor (🔍): I suggest reordering this view and moving that on top of the `text_username`, code wise, just to keep the design order so that it makes it easier to read through. I was confused initially while reading this layout file.

Based on this comment, I let AS reorder the entire view. Thought you were confused before. I've committed it because I want you to see how AS believes it should be ordered.

> 3. Question (❓): The design specifies that there should exist a `4dp` bottom margin. Am I missing something?

@ParaskP7  - there were no specs for this view ... so this is based this off of Woo and then adjusted to match what Chris and I discussed. Where are you seeing a spec sheet for this?

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
zwarm: @ParaskP7 Same comment as below on specs

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
zwarm: @ParaskP7 ❓ 
`ActionableEmptyStateTitle` & `ActionableEmptyStateSubtitle` are used in `actionable_empty_view` which are used in empty views across the App. I wanted to keep this view aligned with the other empty views How did you map it to `Help Activity Styles` ?

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
zwarm: TBH, I couldn't tell you what I added margin/padding two weeks ago. Going to wait for your answer re: specs before adjusting anything else in the view. Thanks.

zwarm:
> * At least, create a `jetpack` package within the `login` package and move the `Jetpack` related classes in it. This will make it obvious that those are used for Jetpack only (at least atm). Also, it might make it easier to potentially move them later on, to a different flavor or repo. So, whatever is being used through the `LoginActivity.showNoJetpackSites()`, I would move all that functionality into the newly created `jetpack` package and include the `Jetpack` naming on all those classes/files.

What I did was create a new Jetpack package under `login` (including test) and moved the classes as is. The package name _should_ be enough to identify that the classes are used for Jetpack only. I felt adding Jetpack to the class name would be redundant at this point. After I address all the smaller issues, I will go back and reassess whether I want to rename individual classes and open that as a separate PR.


> * Also, move the `Login` related classes to the `login` package. Some of them are still on the `accounts` package level and I am not sure why.

I chose not to move other `Login` related classes to the package. The risk is unknown (but maybe small 🤷 ) and I don't want to introduce scope into the project at this time.

zwarm:
> * At least, create a `jetpack` package within the `login` package and move the `Jetpack` related classes in it. This will make it obvious that those are used for Jetpack only (at least atm). Also, it might make it easier to potentially move them later on, to a different flavor or repo. So, whatever is being used through the `LoginActivity.showNoJetpackSites()`, I would move all that functionality into the newly created `jetpack` package and include the `Jetpack` naming on all those classes/files.

What I did was create a new Jetpack package under `login` (including test) and moved the classes as is. The package name _should_ be enough to identify that the classes are used for Jetpack only. I felt adding Jetpack to the class name would be redundant at this point. After I address all the smaller issues, I will go back and reassess whether I want to rename individual classes and open that as a separate PR.


> * Also, move the `Login` related classes to the `login` package. Some of them are still on the `accounts` package level and I am not sure why.

I chose not to move other `Login` related classes to the package. The risk is unknown (but maybe small 🤷 ) and I don't want to introduce scope into the project at this time.

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginNoSitesFragment.kt
ParaskP7: Yeap, yeap, I totally understand, actually I had the same feeling before commenting on that. Then, I realised that it is only logically that if the client of that api doesn't want to provide a listener to do something when the request is finished then, he should have to be forced by the api, exactly what is happening in this case. Thus, I though I should comment on enhancing the api. However, feel free to ignore, I am not feeling strongly about it. 😄 

PS: Just to get a bit more knowledge on it I would ask Ferrari on it and the reasoning behind the `listener` not being nullable. Probably, it was just due to it was being used everywhere it was exposed to.

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
ParaskP7: > Based on this comment, I let AS reorder the entire view. Thought you were confused before. I've committed it because I want you to see how AS believes it should be ordered.

Yeap, AS can reorder those in a totally weird way, I have experienced that in the past and still I am not sure why it does that. I understand reformatting it, but reordering should happen like that. 🤔 

> @ParaskP7 - there were no specs for this view ... so this is based this off of Woo and then adjusted to match what Chris and I discussed. Where are you seeing a spec sheet for this?

I see, thanks for explaining that for me. PS: I was checking the specs on Zeplin.

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
ParaskP7: > I wanted to keep this view aligned with the other empty views.

This makes sense. 🙏 

> How did you map it to Help Activity Styles ?

I am seeing this comment on top of these styles, thus my asking about it: `<!--Help Activity Styles-->`

If this group comment is outdated, maybe we should consider replacing it with something better. Wdyt?

zwarm:
> However, when testing, and although everything worked as expected, I noticed that when I click the Try with another account button, and then use the same account to try again, one that doesn't have any Jetpack sites connected to it, then instead of getting the No Sites empty screen again, I get redirected straight to the initial login screen, as if I automatically clicked the Try with another account button. I actually can see the No Sites empty screen shown for a split of a second and then I get redirected.

@ParaskP7 - I can not recreate this scenario. I've tried at least a dozen times. Can you provide emulator specs please. Thanks

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
ParaskP7: @zwarm I think reformatting using AS made it actually worse, thus I am not doing that with the layout files, it reorder everything completely and without keeping a logical/readable order: 9a905ad28316033fb892c8bcb839186e2cc3827e 😄

ParaskP7:
> What I did was create a new Jetpack package under login (including test) and moved the classes as is. The package name should be enough to identify that the classes are used for Jetpack only. I felt adding Jetpack to the class name would be redundant at this point. After I address all the smaller issues, I will go back and reassess whether I want to rename individual classes and open that as a separate PR.

I agree. 👍 

> I chose not to move other Login related classes to the package. The risk is unknown (but maybe small 🤷 ) and I don't want to introduce scope into the project at this time.

Makes sense. 👍

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginNoSitesFragment.kt
zwarm: > Probably, it was just due to it was being used everywhere it was exposed to.

That would be my guess. Easy enough to ignore the listener if it's null in the `loadAvatarFromUrl` method, but how do we communicate back to the caller that it's required in those certain cases. Feels like an enhancement to the Login library OR put the responsibility back on the user to decide HOW they want to use it.

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
zwarm: LOL ... I didn't even see the comment. I used the AS find usages, well because, comments are only good on the day they are written. 😜  But yes, changing the comment in the style file is a good idea.

ParaskP7:
> @ParaskP7 - I can not recreate this scenario. I've tried at least a dozen times. Can you provide emulator specs please. Thanks

@zwarm , I am using a Pixel4 physical device. Actually, I could now reproduce it by doing the following and it seems to be consistent every time:
- Login with an account that doesn't have any Jetpack sites.
- When the `No Sites` empty screen is show, click the `Try with another account button`.
- Login with an account that does have at least one Jetpack site.
- Logout from that account.
- Login again with the first account that doesn't have any Jetpack sites.
- Now, instead of getting the `No Sites` empty screen again, you will get redirected straight to the initial login screen, as if I automatically clicked the Try with another account button.

Can you try the above instructions please?

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
zwarm: @ParaskP7 yes, letting AS reformat made it worse - which is why I committed it - just to show you. :)

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
zwarm: Actually, that comment is only over the `HelpActivitySingleText` and doesn't refer to the ones below it - at least that is how I am reading it. I'm going to leave it as is.

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
zwarm: I pushed the reformatted file in 9e49217. Yay

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginNoSitesFragment.kt
ParaskP7: Yeap, it is indeed a bit of a 🤔 puzzle, if it were me I would be comfortable to put the responsibility back to the user, but let's ignore that for now to progress with it. It just might be worthy it to talk to Ferrari and add an enhancement issue for later on.

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
ParaskP7: > @ParaskP7 yes, letting AS reformat made it worse - which is why I committed it - just to show you. :)

😅, yeap ordering it manually is the only way it seems.

WordPress/src/main/res/layout/fragment_login_no_sites_empty_view.xml
ParaskP7: > comments are only good on the day they are written.

True! 😝 

> Actually, that comment is only over the HelpActivitySingleText and doesn't refer to the ones below it - at least that is how I am reading it.

You are right, maybe what is missing is another `<!-- Empty State Styles -->` comment to make it more obvious that those styles are not part of the `<!--Help Activity Styles-->`. 🤷

ParaskP7:
@zwarm , actually, and in addition to the testing problem I stumbled and described above, I found that something even trickier happens when you trigger the back action afterwards to exit the app. Instead of exiting, the app enters a weird state when
the users seems to be logged-in.

If you manage to replicate the problem, try to swipe or press back and see what happens.

ParaskP7 (approved):
As discussed, I am going to approve this PR and we will create separate issue to deal with the rest.

PR #14600, opened 2021-05-05
Jetpack App (Basics): Site Address is not Jetpack

zwarm:
Parent #14302 

This PR implements the empty state UI when using a site address to log into the Jetpack app and it's not a Jetpack site or it's disconnected.

Implementation Details
- Introduced `LoginSiteCheckErrorViewModel` to handle button clicks and back press
- Refactored the way binding was being handled in `LoginSiteCheckErrorFragment`
- Refactored `ShowInstructions` navigation event to use a single const val for the URL, but still allow flexibility if this changes in the future by passing the value as a param

**Merge Instructions**
- Ensure PR #14596 has been merged
- Remove Not ready for merge label
- Merge as normal

**To test:**
- Download this branch and build a Jetpack App variant
- Tap the 'Enter your existing address' button
- Enter `http://do.wpmt.co/jp-disconnected/` or another site that has Jetpack installed but not connected
- You should be brought to an empty view with the message "The website {site} is not a Jetpack site"

## Regression Notes

Manual and automated tests

Added new unit tests in `LoginSiteCheckErrorViewModelTest`

PR submission checklist:


osullivanchris (approved):
Tested on phone in emulator, light and dark mode. Looks good. 

Two suggestions for improvement (not hitting 'Request changes' as they are not blocking):

- When I logged in initially, the keyboard was still up. Might just be an emulator issue. Had to tap the hardware down button to dismiss it.
- If we could bold the name of the site it might look nice. But not critical obviously.

zwarm:
Thanks, @osullivanchris  ... I will take a look at both of these. Good feedback.

zwarm:
@osullivanchris 
> * When I logged in initially, the keyboard was still up. Might just be an emulator issue. Had to tap the hardware down button to dismiss it.

Did not happen on a device. Let me know if you see this again.

> * If we could bold the name of the site it might look nice. But not critical obviously.

I pushed a change that bolds the name of the site. Pull the latest and give it a once over when you have the chance. Thanks.

osullivanchris:
> Did not happen on a device. Let me know if you see this again.

Ok cool. Sure thing!

> I pushed a change that bolds the name of the site. Pull the latest and give it a once over when you have the chance. Thanks.

Looks good! Adds a nice bit of clarity. Thanks for the fix.

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginSiteCheckErrorFragment.kt
```diff
@@ -2,95 +2,105 @@ package org.wordpress.android.ui.accounts.login
...
+        viewModel = ViewModelProvider(this@LoginSiteCheckErrorFragment, viewModelFactory)
+                .get(LoginSiteCheckErrorViewModel::class.java)
+    }
 
-        (activity as AppCompatActivity).supportActionBar?.let {
-            it.setDisplayHomeAsUpEnabled(true)
-            it.setDisplayShowTitleEnabled(true)
-        }
+    private fun FragmentLoginSiteCheckErrorBinding.initErrorMessageView() {
+        loginEmptyViewMessage.text =
```
ParaskP7: Suggestion (💡): Consider reformatting this line to:
```
        loginEmptyViewMessage.text = htmlMessageUtils.getHtmlMessageFromStringFormatResId(
                R.string.login_not_a_jetpack_site,
                "<b>$siteAddress</b>"
        )
```

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginSiteCheckErrorFragment.kt
```diff
@@ -110,4 +120,16 @@ class LoginSiteCheckErrorFragment : Fragment(R.layout.fragment_login_site_check_
 
         unifiedLoginTracker.track(step = Step.NOT_A_JETPACK_SITE)
     }
+
+    private fun initBackPressHandler() {
```
ParaskP7: Warning (⚠️): This function is unused. Probably you wanted to use that after `initDagger()` just like it is being done for the `LoginNoSitesFragment` class.

WordPress/src/main/res/layout/fragment_login_site_check_error.xml
```diff
@@ -1,25 +1,69 @@
...
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:layout_gravity="center"
+    android:fillViewport="true">
 
-    <include
-        layout="@layout/toolbar_login"/>
-
-    <ScrollView
+    <androidx.constraintlayout.widget.ConstraintLayout
```
ParaskP7, line 9: Suggestion (💡): There exist a lot of duplication here, that is, in relation to the `fragment_login_no_sites_empty_view.xml` file. Maybe, we should extract these duplicate views into a separate layout file and reuse that new layout here and there.

Actually, thinking a bit more about it, it seems to me that `LoginNoSitesFragment` and `LoginSiteCheckErrorFragment` share a lot of functionality as well, thus duplicating a lot of code. The main diff as far as I can see is the `ARG_SITE_ADDRESS`. As such, maybe those two fragments and functionality needs to be merged into one and then their corresponding layout files to be merged as well. I feel that this will reduce a lot of duplication and make any future maintainability more manageable and less error prone to, for example, updating one view while forgetting to update the other one as well.

Wdyt?

WordPress/src/test/java/org/wordpress/android/ui/accounts/login/LoginSiteCheckErrorViewModelTest.kt
```diff
@@ -1,58 +1,62 @@
...
+import org.wordpress.android.ui.accounts.LoginNavigationEvents.ShowInstructions
 import org.wordpress.android.ui.accounts.LoginNavigationEvents.ShowSignInForResultJetpackOnly
-import org.wordpress.android.ui.accounts.UnifiedLoginTracker
 
 @InternalCoroutinesApi
-class LoginNoSitesErrorViewModelTest : BaseUnitTest() {
-    @Mock lateinit var unifiedLoginTracker: UnifiedLoginTracker
-    @Mock lateinit var wordPress: WordPress
-    private lateinit var viewModel: LoginNoSitesErrorViewModel
+class LoginSiteCheckErrorViewModelTest : BaseUnitTest() {
```
ParaskP7: Suggestion (💡): I already left some testing related comment on the [accompanying PR](https://github.com/wordpress-mobile/WordPress-Android/pull/14596). Same apply to this PR as well.

ParaskP7 (changes requested):
👋 @zwarm !

Since this PR goes hand-to-hand with its [accompanying PR](https://github.com/wordpress-mobile/WordPress-Android/pull/14596), I have just added a few more comments here for you to consider. I am **NOT** going to approve this PR as well, will request changes for now, but to be honest it is not that something is blocking. It just that there are things to discuss, like merging functionality and removing duplication.

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginSiteCheckErrorFragment.kt
zwarm: Great catch :) Thanks

WordPress/src/main/res/layout/fragment_login_site_check_error.xml
zwarm: @ParaskP7 - I see your POV and I thought about it earlier, but made the choice to keep them separate. What it comes down to is this: Is using `if/case` statements to separate responsibility better than separate classes? One can argue this point from both sides :). Same goes for hiding/showing elements in a layout. You can argue that not having to hide/show makes it less error prone because you don't have to remember which case should hide or show elements. I decided to keep them individual fragment/layouts so that each had a single responsibility. However, I am open to the changing to meet the expectations of the majority. :)

ParaskP7: @zwarm , totally with you on that, my only fear is the amount of duplication and maintainability overhead, or even problems with that. Other than that, I agree with your point, that complexity just (kind of) shifts with one solution vs. the other.

ParaskP7 (approved):
As discussed, I am going to approve this PR and we will create separate issue to deal with the rest.

PR #14613, opened 2021-05-10
Jetpack App: Update infrastructure to support Jetpack app

JavonDavis:
This PR contains the following changes

1. Introduce support for an app option to our lanes to allow us to differentiate between jetpack and wordpress
2. Adopts the approach of managing the versions of these apps in a separate properties file and provides a gradle task to access and update the versions from this file 
3. Introduces installable builds for the Jetack app

To test:

- Testing the lanes used for building the bundle, uploading to the app store and triggering the CI job are the most important to check off here, that they work as expected when the app is wordpress/jetpack. 

PR submission checklist:


WordPress/build.gradle
```diff
@@ -60,9 +63,10 @@ android {
         if (project.hasProperty("versionName")) {
             versionName project.property("versionName")
```
oguzkocer: I don't think we want this anymore. We probably didn't want it for a some time. For context, it looks like [it was added to be able to pass the version name from command line](https://github.com/wordpress-mobile/WordPress-Android/commit/51bd1b015fb3c53393e6098b439e64cae7d32add).

WordPress/build.gradle
```diff
@@ -60,9 +63,10 @@ android {
...
         } else {
-            versionName "alpha-293"
+            versionName versionProperties.getProperty("wordpress.alpha.versionName")
```
oguzkocer: It'd be good to set this to a variable and reuse it in line 69 or anywhere else it'd be used: `versionProperties.getProperty("wordpress.alpha.versionName")`

WordPress/build.gradle
```diff
@@ -93,9 +97,15 @@ android {
 
     flavorDimensions "app", "buildType"
 
+
```
oguzkocer:
```suggestion
```

WordPress/build.gradle
```diff
@@ -93,9 +97,15 @@ android {
...
     productFlavors {
         wordpress {
             dimension "app"
+            // Only set the release version if one isn't provided
+            if (!project.hasProperty("versionName")) {
```
oguzkocer: Similar to my previous comment, I think we should remove this `if` statement and directly set the version name.

WordPress/build.gradle
```diff
@@ -104,6 +114,11 @@ android {
 
         jetpack {
             dimension "app"
+            // Only set the release version if one isn't provided
+            if (!project.hasProperty("versionName")) {
```
oguzkocer: Similar to my previous comment, I think we should remove this `if` statement and directly set the version name.

WordPress/build.gradle
```diff
@@ -455,12 +465,41 @@ task violationCommentsToGitHub(type: se.bjurr.violations.comments.github.plugin.
     ]
 }
 
-task printVersionName {
+task printWordPressVersionName {
+    doLast {
+        println android.productFlavors.wordpress.versionName-rc-1
```
oguzkocer:
```suggestion
        println android.productFlavors.wordpress.versionName
```

I am guessing this was a copy/paste error.

WordPress/build.gradle
```diff
@@ -455,12 +465,41 @@ task violationCommentsToGitHub(type: se.bjurr.violations.comments.github.plugin.
     ]
 }
 
-task printVersionName {
+task printWordPressVersionName {
```
oguzkocer: I don't think we want to add this many tasks for version name stuff. And for the tasks that we add, we should use [task configuration avoidance](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html) so it doesn't impact the build performance.

WordPress/build.gradle
```diff
@@ -455,12 +465,41 @@ task violationCommentsToGitHub(type: se.bjurr.violations.comments.github.plugin.
...
+}
+
+task printJetpackVersionCode {
+    doLast {
+        println android.productFlavors.jetpack.versionCode
+    }
+}
+
+
+task "updateVersionProperties" {
```
oguzkocer: It looks like this task is still a work in progress since it doesn't really do anything at the moment. It can be quite tricky to get the version number increased automatically. We could look into using [nebula plugin](https://nebula-plugins.github.io/) or at least use it as an inspiration.

Also, I don't mind this being a part of the project right now, but it'd be good to have this as a Gradle plugin so that we can use it in other apps as well.

oguzkocer (commented):
@JavonDavis I've left a few comments to the Gradle file changes. I hope you'll find them useful. I'd be happy to work with you on those if you'd like or open a PR against this branch to make my suggestions more concrete.

I have not reviewed the fastlane changes just yet, I can do that once the PR is ready for review. However, it'd be good to get another set of eyes on those files before it's merged as I am not all that comfortable with fastlane just yet.

WordPress/build.gradle
JavonDavis: The idea here for this task is just to update the properties file but the Nebula plugin looks a bit neater, I'm checking it out and their approach now.

AliSoftware:
@JavonDavis as discussed in Slack, and in the spirit of getting this merged during the 17.6 cycle before the next betas and before the final release, I suggest you change your PR to target `release/17.6` and not `develop` anymore. Then it should be ready to undraft once https://github.com/wordpress-mobile/WordPress-Android/pull/14870 with my fixes gets merged into this, right?

.circleci/config.yml
```diff
@@ -18,6 +18,12 @@ parameters:
   release_build:
     type: boolean
     default: false
+  app_name:
+    type: string
+    default: "WordPress"
+  product_key:
+    type: string
+    default: "wordpress"
```
AliSoftware: 💄 Nit: Could be nice to add comments to document those two

Especially I wondered at first if `app_name` had to match something special from the project (must this be the name of the folder in which the app is, aka always `WordPress` for both WP and JP because that's the folder at the repo root? Should it match the name of the product as defined somewhere in some `.properties` or `.gradle` file?… 🤔 ).

Given that the lanes use `app:` as the parameter to distinguish which app to build, I thought `app_name` was supposed to be the value for that parameter at first… until I realized `product_key` was the one for it and `app_name` is only used for display in Slack messages and doesn't have any impact in the execution of lanes. But this is why a comment would help 😅 

My suggestion is:
 - 💄  For this PR, or if not in the next one, just add comments to this parameters
 - 🔜  Optionally, and on a future PR (to not block this one, because this is just cleanup and non-blocker), consider renaming those params to avoid confusion with the similar but different names used in the lanes (e.g. `app_display_name` instead of `app_name`, and `app_key` instead of `product_key`?). This renaming is not critical and definitively for separate PR… if ever.

Gemfile.lock
```diff
@@ -1,3 +1,22 @@
+GIT
+  remote: https://github.com/wordpress-mobile/release-toolkit
+  revision: cb939d6f00ed0483eb4baf107361b7281bf9feac
```
AliSoftware: Since I approved [the release-toolkit PR](https://github.com/wordpress-mobile/release-toolkit/pull/258), feel free to merge it, then generate a new release-tookit version – see paaHJt-2d8-p2 for instructions and point your `fastlane/Pluginfile` to it now.

If you are still waiting for Oguz to review the PR before you feel confident in making a new release-toolkit release, I'm ok to merge this PR while still pointing on a toolkit branch for now, and we can do a subsequent PR to update this once the new toolkit release lands, but I feel we're ready to make that release now tho so why wait 😉

fastlane/Fastfile
```diff
@@ -217,12 +226,13 @@ REPOSITORY_NAME="WordPress-Android"
...
   desc "Updates a release branch for a new beta release"
   lane :new_beta_release do | options |
+    app = options[:app]
     android_betabuild_prechecks(options)
     send_strings_for_translation()
-    download_translations(options)
-    android_bump_version_beta()
-    new_version = android_get_app_version()
-    trigger_release_build(branch_to_build: "release/#{new_version}")
+    android_update_metadata(options)
```
AliSoftware: ⚠️   The revert of `download_translations(options)` line 222 back into `android_update_metadata` is incorrect…

…But I have fixed it in [this PR](https://github.com/wordpress-mobile/WordPress-Android/pull/14870/files#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688) targeting your branch here, so once #14870 is merged this diff should be outdated and this instance fixed.

fastlane/Fastfile
```diff
@@ -276,9 +286,9 @@ REPOSITORY_NAME="WordPress-Android"
     android_finalize_prechecks(options)
     configure_apply(force: is_ci)
     hotfix = android_current_branch_is_hotfix
-    android_bump_version_final_release() unless hotfix
-    download_translations(options) unless hotfix
-    version = android_get_release_version()
+    android_update_metadata(options) unless hotfix
+    android_bump_version_final_release(app: options[:app]) unless hotfix
```
AliSoftware: 🚨 🔴 The revert of `download_translations(options)` back into `android_update_metadata` is incorrect and need to be fixed. Unlike for the one above, I did **NOT** fix that one in the `finalize_release` branch in my other PR. so I'll let you fix this one yourself.

fastlane/Fastfile
```diff
@@ -394,19 +407,52 @@ REPOSITORY_NAME="WordPress-Android"
...
     if (options[:create_release])
-      create_gh_release(version: version, prerelease: true)
+      create_gh_release(version: version, prerelease: true, app: app)
+    end
+  end
+
+  #####################################################################################
+  # build_internal
+  # -----------------------------------------------------------------------------------
+  # This lane builds the app it for restricted internal testing and optionally uploads it
```
```suggestion
  # This lane builds the app for restricted internal testing, and optionally uploads it to PlayStore's Internal track
```

fastlane/Fastfile
```diff
@@ -425,11 +471,22 @@ REPOSITORY_NAME="WordPress-Android"
   #####################################################################################
   desc "Upload Build to Play Store"
   lane :upload_build_to_play_store do | options |
-    aab_file_path = bundle_file_path(options[:version])
+
+    app = options[:app]
+    package_name = app.downcase == "jetpack" ? JP_PACKAGE_NAME : WP_PACKAGE_NAME
```
AliSoftware, line 476: This is totally ok to merge as is to get this PR land ASAP, but as an improvement suggestion for a future/separate PR: 

A more Ruby-y way of doing this is to use a Hash for those constants. For example, change the constants you added [at the top of the Fastfile here](https://github.com/wordpress-mobile/WordPress-Android/pull/14613/files#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688R71-R80) to something like:

```
PACKAGE_NAME = { wordpress: 'org.wordpress.android', jetpack: 'com.jetpack.android' }.freeze
PREFIX = { wordpress: 'wpandroid', jetpack: 'jpandroid' }.freeze
APP_NAME = { wordpress: 'WordPress', jetpack: 'Jetpack' }.freeze
```
And then replace every place in the code where you use this kind of `downcase == 'jetpack'` test with something like this instead:
```suggestion
    package_name = PACKAGE_NAME[app.downcase.to_sym]
```

fastlane/Fastfile
```diff
@@ -650,15 +708,24 @@ REPOSITORY_NAME="WordPress-Android"
...
+      UI.message("Version specified for #{app} bundle is nil. Skipping ahead")
+      next
+    end
+
+    prefix = app.downcase == "jetpack" ? JP_PREFIX : WP_PREFIX
+    name="#{prefix}-#{version["name"]}.aab"
     apk_name=universal_apk_name(version)
-    aab_file="org.wordpress.android-#{options[:app]}-#{options[:flavor]}-#{options[:buildType]}.aab".downcase
+
+    aab_file="org.wordpress.android-#{app}-#{options[:flavor]}-#{options[:buildType]}.aab".downcase
```
AliSoftware, line 722: 💄  nitpicking (can wait a future PR):

Is there a reason why the aab file is not using the package name as prefix, i.e. start with `com.jetpack.android-` instead of `org.wordpress.android` for the case of Jetpack?

It doesn't really matter in practice, but just for consistency…

fastlane/Fastfile
```diff
@@ -669,12 +736,17 @@ REPOSITORY_NAME="WordPress-Android"
       sh("./gradlew clean >> #{logfile_path} 2>&1")
 
       sh("mkdir -p #{build_dir}")
-      UI.message("Running lint...")
-      sh("echo \"Running lint...\" >> #{logfile_path}")
-      sh("./gradlew lint#{options[:app]}#{options[:flavor]}Release >> #{logfile_path} 2>&1") unless is_ci
+      if options[:skip_lint].nil?
+        UI.message("Running lint...")
+        sh("echo \"Running lint...\" >> #{logfile_path}")
+        sh("./gradlew lint#{app}#{options[:flavor]}#{options[:buildType]} >> #{logfile_path} 2>&1") unless is_ci
```
AliSoftware, line 742: 💄  nitpick that is really non-blocker and can **totally** only be addressed in separate future PR:

It would be nice to have a helper method in that Fastfile to assemble different strings as CamelCase to build the gradle task with the right case. This is not a big deal at all since apparently gradle is case-insensitive so this works too, it would just improve readability both of the code in Fastfile and in the logs 🙂 

There's a `capitalize` method on `String` in ruby, so you can use this to turn `jetpack` into `Jetpack` and `vanilla` into `Vanilla`. This could look like this for example:

```ruby
def gradle_task(*args)
  ([args.first.downcase] + args[1..].map(&:capitalize)).join
end
```
and called like this:
```ruby
sh("./gradlew #{gradle_task('lint', app, options[:flavor], options[:buildType])} >> #{logfile_path} 2>&1") unless is_ci
```

Not tested, and again, you can keep that as a separate improvement in different future PR.

AliSoftware (commented):
I've reviewed the Draft in its current state and, looking at the code in GitHub, this overall looks good to me… with **one exception that requires a change**: the accidental revert of the obsolete `android_download_metadata` – which will crash the lane now that we don't have `tools/update-translations.sh` any more because the new action is implemented in full ruby – needs to be addressed. 🚨 

Once this is fixed, and since it doesn't seem to break existing lanes – as we tested this for new_beta_release with JP rc2 – I say let's merge this into `release/17.6` (change the PR's target branch first!) – even if it means starting another branch afterwards to tweak the potential remaining bugs we might find while using those lanes during 17.6 cycle and upcoming betas – to avoid keeping that PR and branch too-long-lived.

I've submitted my review as a comment only, because I don't want my "Request changes" (for the `download_translations` revert) to block you from merging the PR if you get it fixed before I come back, and also because once you change the PR's target branch, the diff might change, so could warrant checking the diff again, but wanted to post my review even while you're still in Draft to give you the push to move this forward and undraft it 🙂

fastlane/Fastfile
JavonDavis: Opened https://github.com/wordpress-mobile/WordPress-Android/issues/14874 to track this!

.circleci/config.yml
JavonDavis: Included comments on this! If there's strong feelings about the naming we can address in a separate PR for sure

fastlane/Fastfile
JavonDavis: This is just the generated name used by gradle when running `./gradlew bundle..` we could look into ways to align this more with flavors but not sure of any way right out the box, maybe @oguzkocer  could have an idea on this one!

fastlane/Fastfile
JavonDavis: Also tracking this in https://github.com/wordpress-mobile/WordPress-Android/issues/14874

WordPress/build.gradle
```diff
@@ -82,13 +85,8 @@ android {
...
 
-        // Allow versionName to be overridden with property. e.g. -PversionName=1234
-        if (project.hasProperty("versionName")) {
-            versionName project.property("versionName")
-        } else {
-            versionName "alpha-298"
-        }
-        versionCode 1060
+        versionName versionProperties.getProperty("wordpress.alpha.versionName")
+        versionCode versionProperties.getProperty("wordpress.alpha.versionCode").toInteger()
```
oguzkocer, line 89: Should we remove these values instead? The alpha version name & code is only valid for alpha builds, so having them in the `defaultConfig` feels confusing to me.

WordPress/build.gradle
oguzkocer:
```suggestion
tasks.register("updateVersionProperties") {
```

https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

WordPress/build.gradle
```diff
@@ -501,9 +500,23 @@ task violationCommentsToGitHub(type: se.bjurr.violations.comments.github.plugin.
...
     doLast {
-        println android.productFlavors.vanilla.versionName
+        ant.propertyfile(
+                file: "../version.properties") {
+            entry( key: "$key", value: "$value")
+        }
+    }
+}
+
+task "printVersionName" {
```
oguzkocer:
```suggestion
tasks.register("printVersionName") {
```

https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

WordPress/build.gradle
```diff
@@ -501,9 +500,23 @@ task violationCommentsToGitHub(type: se.bjurr.violations.comments.github.plugin.
...
+    doLast {
+        def key = "$productKey" + "."
```
oguzkocer, line 514: I don't think this makes it very clear that `productKey` is supposed to be passed as a project property.

.circleci/config.yml
```diff
@@ -18,6 +18,14 @@ parameters:
...
     type: boolean
     default: false
+  # Display name of the app and should match the usual public facing name (e.g. WordPress, Jetpack)
+  app_name:
+    type: string
+    default: "WordPress"
+  # The key used to reference the app in the Fastfile
+  product_key:
+    type: string
+    default: "wordpress"
```
oguzkocer, line 28: I don't think these are modified at all, are they? I see there is a Jetpack build, but I don't think it's actually using these information at all.

Also, I am concerned about using a default for these. It saves a few keystrokes when we have all the context, but when we don't, it'd help if CircleCI complains about the value missing so we deliberately pick the right one.

.circleci/config.yml
```diff
@@ -128,26 +136,49 @@ jobs:
...
+          command: |
+            if [ -n "$CIRCLE_PULL_REQUEST" ]; then
+              PR_NUMBER=$(basename $CIRCLE_PULL_REQUEST)
+              PREFIX="pr-${PR_NUMBER}"
+            else
+              PREFIX="$CIRCLE_BRANCH"
+            fi
+
+            VERSION_NAME="${PREFIX}-build-${CIRCLE_BUILD_NUM}"
+            echo "export VERSION_NAME=$VERSION_NAME" >> $BASH_ENV
```
oguzkocer: We are exporting this information twice, wouldn't it be better to do a step to calculate and export the version name and then use it in `Build WordPress APK` & `Build Jetpack APK` steps?

WordPress/build.gradle
oguzkocer: I don't have a solid advice on this, but the `$VERSION_NAME` in CircleCI and the `printVersionName` is pretty confusing to me because they use the same naming but refer to different things.

fastlane/Fastfile
```diff
@@ -217,12 +226,13 @@ REPOSITORY_NAME="WordPress-Android"
...
   desc "Updates a release branch for a new beta release"
   lane :new_beta_release do | options |
+    app = options[:app]
     android_betabuild_prechecks(options)
     send_strings_for_translation()
-    download_translations(options)
-    android_bump_version_beta()
-    new_version = android_get_app_version()
-    trigger_release_build(branch_to_build: "release/#{new_version}")
+    download_translations(app: app)
```
oguzkocer, line 232: I see that we use both `download_translations(options)` & `download_translations(app: app)`, I am guessing one of them is either incorrect or inconsistent. @AliSoftware Could you please take a look at this one as well?

oguzkocer (commented):
@JavonDavis I've done a review pass of this PR, but I have to say that I am very worried that we are adding even more complexity to an already confusing base. Looking at our CircleCI config is very scary for me, which I guess we can improve during BuildKite port, but it's a concern nonetheless. Similarly, the build flavor config is very bloated and I honestly couldn't keep up with it by just reading the code. I probably need to refactor it to be able to understand it.

So, my question is, do you have any plans to refactor the implementation, especially on the Gradle side?

---

Here are also a few general feedback which I hope will be useful in the future PRs:

1. This PR changes a lot of files (mostly strings) that don't really matter for this review. I think it's best to keep them out to make it easier to review.
2. I've no idea why the `UploadUtils` change was made as part of this PR. It'd be good to separate that if you can.
3. I see that this PR had 62 commits and that doesn't include the release toolkit changes. I think that's way too many iterations without a proper PR review. Another approach you can take in the future when you need to make this many changes is to open smaller PRs against a WIP branch, so they get officially reviewed and merged. It'll also make it easier to keep up with the iterations.
4. When the PR gets this much work in, it's harder for a reviewer to propose a different approach. For example, I can't really ask if we can try something different because it'd possibly discard a lot of work and put us behind on our "deadlines", so all I can do now is ask if there is further refactoring planned. (which is what I did)
5. One possible approach you might want to try is "prepare the workspace" for your intended changes by first doing a refactor PR without changing the end result. That'll make it easier to compare the result and to give feedback on just the refactor. Once the refactor is done, it'll be easier for you to add your changes in.

fastlane/Fastfile
AliSoftware: Both are valid and work, since in lanes the parameters you pass are in practice passed as à `Hash`. Which means that passing `(app: app)` in practice passes a dictionary with the key `app` and the appropriate value (extracted from thé `options` dictionary a couple of lines above) while passing `(options)` pass the whole dictionary of parameters the original lane was called with, which include `app` key… and maybe others.

In terms of conventions I don't mind either — passing `app: app` is explicit to only pass the needed Paramus, but passing the whole `options` is more flexible if the download_translations lane needs to extract more context one day (though… YAGNI) — but I agree we should be consistent everywhere for all calls at least.

WordPress/build.gradle
JavonDavis: Confirming you mean move these values to the zalpha flavor instead and not remove them completely? since that's the variant we use for alpha builds

.circleci/config.yml
JavonDavis: Yep this would be a cleaner approach here!

.circleci/config.yml
JavonDavis: They're being used when setting the parameters in the Release Build but we could just use the pipeline parameters directly in the job if I'm not mistaken so I'll clean that up.

JavonDavis:
@oguzkocer it seems the base branch hadn't been updated when you first gave this a scan, the diff should be a bit more manageable now. 

> So, my question is, do you have any plans to refactor the implementation, especially on the Gradle side?

To help get to a better state with the Jetpack app releases the idea here is to merge this and then handle refactors in subsequent PRs. 

> I've no idea why the UploadUtils change was made as part of this PR. It'd be good to separate that if you can.

Not seeing those changes nor was any intended change here for this, I'll assume this disappeared when the base branch was updated. 

> I see that this PR had 62 commits and that doesn't include the release toolkit changes. I think that's way too many iterations without a proper PR review. Another approach you can take in the future when you need to make this many changes is to open smaller PRs against a WIP branch, so they get officially reviewed and merged. It'll also make it easier to keep up with the iterations.

Yep! Good point the number of commits is also partly a result of keeping it up to date with develop and then the latest set of branch managing we need to do to handle the Jetpack release. 

> When the PR gets this much work in, it's harder for a reviewer to propose a different approach. For example, I can't really ask if we can try something different because it'd possibly discard a lot of work and put us behind on our "deadlines", so all I can do now is ask if there is further refactoring planned. (which is what I did)

Yeah I understand the issue here, but I think it'd be best to handle improving or changing the approach here in a smaller subsequent PR so that we can get to a better state with the JP/WP android release process. 

> One possible approach you might want to try is "prepare the workspace" for your intended changes by first doing a refactor PR without changing the end result. That'll make it easier to compare the result and to give feedback on just the refactor. Once the refactor is done, it'll be easier for you to add your changes in.

Agreed this is definitely should be the approach I take next time in a similar situation

JavonDavis:
@AliSoftware I've opened https://github.com/wordpress-mobile/WordPress-Android/pull/14885 that resolves the merge conflicts, I'm thinking to get this merged to help us get rid of the branch dance that needs to be done currently and work on refactoring any improvements in the gradle file or the Fastfile in smaller PRs.

JavonDavis:
> @AliSoftware I've opened #14885 that resolves the merge conflicts, I'm thinking to get this merged to help us get rid of the branch dance that needs to be done currently and work on refactoring any improvements in the gradle file or the Fastfile in smaller PRs.

To add more context as to why I opened a separate PR, the strings conflict scared me a bit and I opted to do the resolution on a different branch just to be safe on this branch. I'm happy to resolve the conflicts here since it's a simple change and maybe it'll make more sense to not have another PR to shift to for this one. 

EDIT:  Opted to keep everything in one PR to keep the discussion in one place

oguzkocer:
> To help get to a better state with the Jetpack app releases the idea here is to merge this and then handle refactors in subsequent PRs.

That makes me very happy, thank you! Let me know if you need any help with that. I am certainly confused right now and if you are as well, maybe we can figure it out together?

> Not seeing those changes nor was any intended change here for this, I'll assume this disappeared when the base branch was updated.

It's gone now 🥳 

---

@JavonDavis Can you re-request a review when this is ready for another look? I think you're planning to merge #14885 into this branch and might still be working on some of the feedback, so I am going to hold off on any more comments until your next ping.

P.S: It's also pretty late here 🙈

JavonDavis:
> I am certainly confused right now and if you are as well, maybe we can figure it out together?

@oguzkocer 😓  yep! this will be much simpler to look at in smaller PRs

> Can you re-request a review when this is ready for another look? I think you're planning to merge #14885 into this branch and might still be working on some of the feedback, so I am going to hold off on any more comments until your next ping.

I've re-requested the review because I think I addressed the major issues and opened issues to track some of the other feedback points, I'll continue opening issues for discussions that are still open, but unless there's a major issue standing out that shouldn't get into the release I'm ok with merging this and open issues around the pain points here, WDYT?

JavonDavis:
> P.S: It's also pretty late here 🙈

I know! 😅 🙈 I'll be signing off in a little bit

WordPress/build.gradle
JavonDavis: Tracking in https://github.com/wordpress-mobile/WordPress-Android/issues/14886

WordPress/build.gradle
JavonDavis: I'd like to propose tracking improvements here in https://github.com/wordpress-mobile/WordPress-Android/issues/14886

oguzkocer: Yes, I meant removing from defaultConfig and adding the version name & code separately for each flavor.

oguzkocer: I think we should remove the version from the `app` dimension and instead add it to each `buildType` by checking the app dimension. That means, we should introduce a beta & product version name for each app. That also means, `jalapeno` is the only flavor that takes the version name as a project property - which also limits the potential errors with passing the version name from command line.

We could even create a custom task just to build & release `jalapeno`, so it's all nicely contained.

What do you think @JavonDavis?

oguzkocer (approved):
@JavonDavis I am approving this because:
1. @AliSoftware needs this in release branch for his release duties
2. We have agreed that further improvements will be done as smaller follow up PRs. It'd be great if you can reference this PR in those, so at least it shows up in this PR's "mentioned" issues.

Thanks for working on this!

PR #14627, opened 2021-05-12
Stories - update init order

mzorz:
Fixes a crash identified in https://github.com/Automattic/stories-android/pull/681#issuecomment-839388382

Was able to reproduce on a fresh install (delete application data), and fixed in https://github.com/wordpress-mobile/WordPress-Android/commit/5ca542c07d4c47dc9ce27b7ae55c798b43067624 by moving `super.onCreate()` all the way down (actually just below `initViewModel()`) to make sure any Story library listeners instantiated in the ViewModel would be available before calling ComposeLoopFrameActivity's `onCreate()`. I also had to move the initialization call order of `setUseTempCaptureFile()` given it internally accesses the `storyViewModel` (and it can only exist after initializing the ViewModel itself).

Note this PR also updates the Stories library hash to point to the PR in Stories repo where the issue was noticed.

To test:
1. delete app's data
2. login to WP (a .com or jetpack site)
3. tap FAB to create a story
4. select one or two items from the media picker
5. observe the app doesn't crash with the `uninitialized prop` exception (it will fall back to the My Site screen the first run which is fixed in https://github.com/Automattic/stories-android/pull/684, but will work following tries)

## Regression Notes
N/A

N/A

N/A

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
+++ b/WordPress/src/main/java/org/wordpress/android/ui/stories/StoryComposerActivity.kt
@@ -142,7 +142,6 @@ class StoryComposerActivity : ComposeLoopFrameActivity(),
         setMediaPickerProvider(this)
         (application as WordPress).component().inject(this)
         initSite(savedInstanceState)
-        super.onCreate(savedInstanceState)
         setSnackbarProvider(this)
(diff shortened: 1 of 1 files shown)


aforcier (approved):
Works well! Stories library PR reviewed and merged also. :shipit:

PR #14642, opened 2021-05-14
Storage warning dialog on open editor.

### LINKED ISSUE #11223: AztecLoggingException: java.io.IOException: write failed: ENOSPC (No space left on device)
Sentry Issue: [WORDPRESS-ANDROID-8PN](https://sentry.io/organizations/a8c/issues/1490461672/?referrer=github_integration)

```
ErrnoException: write failed: ENOSPC (No space left on device)
    at libcore.io.Posix.writeBytes(Posix.java)
    at libcore.io.Posix.write(Posix.java:258)
    at libcore.io.BlockGuardOs.write(BlockGuardOs.java:313)
    at libcore.io.IoBridge.write(IoBridge.java:497)
    at java.io.FileOutputStream.write(FileOutputStream.java:186)
...
(32 additional frame(s) were not displayed)

IOException: write failed: ENOSPC (No space left on device)
    at libcore.io.IoBridge.write(IoBridge.java:502)
    at java.io.FileOutputStream.write(FileOutputStream.java:186)
    at java.io.DataOutputStream.writeShort(DataOutputStream.java:192)
    at java.io.ObjectOutputStream.writeStreamHeader(ObjectOutputStream.java:1795)
    at java.io.ObjectOutputStream.<init>(ObjectOutputStream.java:271)
...
(29 additional frame(s) were not displayed)

AztecLoggingException: java.io.IOException: write failed: ENOSPC (No space left on device)
    at org.wordpress.android.ui.posts.EditPostActivity$10.logException(EditPostActivity.java:1792)
    at org.wordpress.android.ui.posts.EditPostActivity.onSaveInstanceState(EditPostActivity.java:813)
...
(26 additional frame(s) were not displayed)

java.io.IOException: write failed: ENOSPC (No space left on device)
```

develric:
Fixes #11223 

The solution mimic the dialog already used in Stories (as highlighted [here](https://github.com/wordpress-mobile/WordPress-Android/issues/11223#issuecomment-805059838)) to show a warning to the user when the editor is opened and the available storage is below 10% threshold. We also added an option button to opt out and don't show the dialog again.

## To test

To fill in the storage space I used this SO link answer reported in the [original issue](https://stackoverflow.com/questions/33241465/android-test-out-of-storage-space).

### Storage is not low
- Create (or edit) a post or page and smoke test everything works as usual (no changes should be noticed)

### Storage is low
- Fill in the storage of the emulator to above 90%
- Create (or edit) a post or page
- Notice the dialog is presented as below

<p align=center><image src="https://user-images.githubusercontent.com/47797566/118242324-7e7e3e80-b49d-11eb-905f-206a2ec0f521.png" width=300/></p>

- Test the `VIEW STORAGE` button and check you are landed to the storage settings and you can come back to the app with the system back button
- Test the `CANCEL` (or tap outside the dialog) to make the dialog disappear
- Test the `DON'T SHOW AGAIN` button and check the dialog is not presented anymore when coming back to create (or edit) a post or page
- To check the case when settings intent cannot be resolved, you can use the debugger to force the scenario like in the left image below and getting the dialog reported on the right image below

| image | image |
|---|---|

### Analytics
Activate analytics and check you get the following

**On opening the editor when the dialog appears**
```
🔵 Tracked: storage_warning_shown, Properties: {"is_storage_settings_resolved":"true","source":"editor"}
```

**On Cancel (or tapping outside the dialog)**
```
🔵 Tracked: storage_warning_canceled, Properties: {"is_storage_settings_resolved":"true","source":"editor"}
```

**On View Storage**
```
🔵 Tracked: storage_warning_acknowledged, Properties: {"is_storage_settings_resolved":"true","source":"editor"}
```

**On Don't show again**
```
🔵 Tracked: storage_warning_dont_show_again, Properties: {"is_storage_settings_resolved":"true","source":"editor"}
```

## Regression Notes
Opening the editor

Manual testing

Added unit tests for the new VM introduced

PR submission checklist:


WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
```diff
@@ -482,6 +485,7 @@ protected void onCreate(Bundle savedInstanceState) {
         ((WordPress) getApplication()).component().inject(this);
         mDispatcher.register(this);
         mViewModel = new ViewModelProvider(this, mViewModelFactory).get(StorePostViewModel.class);
+        mStorageUtilsViewModel = new ViewModelProvider(this, mViewModelFactory).get(StorageUtilsViewModel.class);
```
mzorz, line 488: This is probably a minor optimization (not needed at all), but wondering if `new ViewModelProvider(this, mViewModelFactory)` could be called once and assigned to a local variable we could reuse? Feel free to overlook this comment if you don't think it's any better 👍

mzorz: For the record, I searched elsewhere out of curiosity so sharing this here https://www.bettercodebytes.com/the-cost-of-object-creation-in-java-including-garbage-collection/ (tl;dr we can probably just leave it as is 👍 😅 )

WordPress/src/main/java/org/wordpress/android/ui/prefs/AppPrefs.java
```diff
@@ -250,6 +250,9 @@
 
         // Used to determine if editor onboarding features should be displayed
         HAS_LAUNCHED_GUTENBERG_EDITOR,
+
+        // Used to indicate whether or not the device running out of storage warning shoudl be shown
```
mzorz:
```suggestion
        // Used to indicate whether or not the device running out of storage warning should be shown
```

minor typo 👍

WordPress/src/main/java/org/wordpress/android/viewmodel/storage/StorageUtilsViewModel.kt
```diff
@@ -0,0 +1,37 @@
...
+class StorageUtilsViewModel @Inject constructor(
+    @Named(BG_THREAD) private val bgDispatcher: CoroutineDispatcher,
+    private val storageUtilsProvider: StorageUtilsProvider
+) : ScopedViewModel(bgDispatcher) {
+    private val _checkStorageWarning = MutableLiveData<Event<Unit>>()
+    val checkStorageWarning: LiveData<Event<Unit>> = _checkStorageWarning
+
+    fun onStorageWarningInit() {
+        launch(bgDispatcher) {
+            delay(DIALOG_CHECK_DELAY)
```
mzorz, line 30: I don't really understand why we need to add a delay here - is it because the observers may not be attached before? This relationship seems a bit flaky, maybe there's a way in which we can just trigger the action wrapped in `onStorageWarningCheck()` directly from EditPostActivity, without calling `onStorageWarningInit()` only to wait for the observer to be called back later on?

I'd go with something simpler like:
1. removing `onStorageWarningInit()` and its call
2. replace the call to `onStorageWarningInit()` with `onStorageWarningCheck()` in `EditPostActivity` 
3. remove the observer in EditPostActivity

I may be missing something, is there a reason not to go this way? Let me know 🙏

mzorz (commented):
Thanks for implementing these changes @develric  🙇 

Tested, the functionality works as advertised. I left some comments about some things I wanted to double check with you. Let me know when this one is ready for another round 👍 

I also wonder whether the text in the Dialog seems relevant or not (as a user, it's difficult for me to connect how what I can read in the Dialog "Insufficient device storage"  is related to the activity of writing / editing a Post which is what I'm about to do). Maybe I'd add an explanation about why we need space for editing, but maybe this is too much and after all this is a personal preference only. Feel free to disregard this if this does not seem to be a concern 👍

WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java
develric: >tl;dr we can probably just leave it as is 👍 😅

Agree, and thanks for sharing that interesting post 🙇

WordPress/src/main/java/org/wordpress/android/viewmodel/storage/StorageUtilsViewModel.kt
develric: I should have mentioned it in the description, my bad sir 🙃

The reason why I placed a delay there is to try to avoid visual glitches like this below with the keyboard on editor opening that can happen by timing races (maybe not the best approach but should do its job hopefully 😄  )

<p align=center><image src="https://user-images.githubusercontent.com/47797566/118638533-3b8ed480-b7d7-11eb-9ae9-66b73a232acb.gif" width=300/></p>

I preferred to post to a live data due to its nature of being lifecycle aware (so no issues in case the user could open the editor and exit soon after it). Considered something like Handler.postDelayed, but as said the live data approach seemed cleaner to me.

This said, I think I can do a better job with naming and used the `start()` approach we already use with VMs all over the places (also placed a guard on `savedInstanceState` to avoid showing the dialog including when the VM could be re-created when the activity is destroyed in background (see the `don't keep activities` forced scenario). 

Hope this works better now 🙇

develric:
>as a user, it's difficult for me to connect how what I can read in the Dialog "Insufficient device storage" is related to the activity of writing / editing a Post which is what I'm about to do

Nice point 👍 ! I checked back the copy on the stories dialog and I think you were already doing similar reasoning there. Tried to mimic that wording like below. Let me know wdyt 🙇 

<p align=center><image src="https://user-images.githubusercontent.com/47797566/118639694-76ddd300-b7d8-11eb-8685-c328261cd26d.png" width=300/></p>

This should be ready for another pass @mzorz 🙇

WordPress/src/main/java/org/wordpress/android/viewmodel/storage/StorageUtilsViewModel.kt
mzorz: Oh I see! Understood, it's much clearer now! What do you think about adding this clarification in a comment in code? I'm still finding it difficult to understand if you read the code but lack this piece of context. Maybe also add a reference to this issue.

Nice call calling it `start()` and adding the safeguard for `savedInstanceState` 👍

mzorz (approved):
Thanks for the copy change, the method rename and safeguard, and the clarification about the added delay @develric 🙇 - I'd only request that we add a comment in code explaining the delay a bit more, even with a link to this issue so others who read the code may find it easier to understand by giving them more context. Then we can merge, giving this one approval so feel free to merge after the comment has been added 👍 
:shipit:

develric:
Merging this one after adding the required comment. Thanks for the review and all the feedbacks @mzorz 😄  , always appreciate reasoning with you about things 🙇  !

PR #14656, opened 2021-05-15
Reader: Respect src tag when tapping a Story preview

aforcier:
Fixes a recent issue where tapping a Story in the Reader detail view would open the photo viewer instead of opening the story in a browser for fullscreen playback.

The cause was a change to the structure of the rendered block, which moved the anchor tag further outside the nested HTML structure. This changed how taps were being registered by the WebView on Android. It looks like we've already had to solve a similar problem for Instagram embeds, so I just extended that solution to recognize story block URLs as well.

_Note: There's a [relevant Calypso PR](https://github.com/Automattic/wp-calypso/pull/52892) unmerged as of this writing that makes some CSS changes required by the same change that broke what this PR is fixing. However, the CSS changes are needed for fixes to the Calypso and WPiOS readers - the WPAndroid reader is unaffected, probably due to some customizations we have about `display: block` and such._

To test:

1. Follow a WP.com site with stories (wpstories DOT wordpress DOT com works)
2. Open a story post from that site in the Reader
3. Observe that the story preview loads correctly in the Reader detail view (should look like this):
device-2020-09-22-122031
4. Tap the story preview, and verify that it opens the device browser and plays the story in fullscreen mode

## Regression Notes

There should be none - this change specifically targets Story blocks in the Reader, identifying those by their URL structure.

N/A

N/A

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/reader/views/ReaderWebView.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/reader/views/ReaderWebView.java
@@ -193,13 +193,18 @@ public class ReaderWebView extends WPWebView {
     }
 
     /***
-     * isValidInstagramImageClick checks if this is an embedded instragram situation.
-     * If so, we want to ignore image click so that the iframe src gets handled
-     * The additional URL grab is an additional check for the instagram.com host
+     * Checks if the tapped image is a child of an anchor tag we want to respect.
+     * If so, we want to ignore the image click so that the wrapping src gets handled.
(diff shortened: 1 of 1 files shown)


aforcier:
Note: I'm targeting the 17.4 milestone with this PR but that will likely have already been cut by the time this can be reviewed. That's okay, let's just make sure to switch the base branch to the frozen `release/17.4` branch once that's cut so the fix can go out in a later beta cut.

mzorz (approved):
Thank you for this change @aforcier ! 
LGTM
:shipit:

PR #14660, opened 2021-05-17
Fix for choppy scrolling experience in Reader

### LINKED ISSUE #14636: Smooth the scrolling experience when browsing in Reader
### Expected behavior
Smooth scrolling up and down the Reader feed.

### Actual behavior
The header at the top prevents smooth scrolling the first couple of strokes in one direction. The header at the top catches the momentum of the scroll. The first few strokes are stunted, and then the next few, you can see the momentum of the page scroll smoothly. 

Recorded screen for reference:

https://user-images.githubusercontent.com/4505888/118044488-8ddd7a80-b32b-11eb-8558-847c8879a0ee.mp4

- Open Reader
- Begin scrolling up and down

ravishanker:
Fixes #14636 

Reader screen has a CollapsingToolbarLayout and a TabLayout that can feel scrolling up and down at time choppy as reported in the issue.

It is a subtle experience issue.  Seems to be due to using three flags combination on the layout_scrollFlags.  For the curious, this [post](https://medium.com/martinomburajr/android-design-collapsing-toolbar-scrollflags-e1d8a05dcb02) explains in detail.

To test:
Login to WordPress Android app.  Switch to Reader tab from the bottom navigation. Scroll up and down.  Scrolling should be is smoother than before without feeling like stuck when scrolling up or down

## Regression Notes
None

Regular regression tests

NA

PR submission checklist:

### DIFF
--- a/WordPress/build.gradle
+++ b/WordPress/build.gradle
@@ -87,6 +87,9 @@ android {
     // Gutenberg's dependency - react-native-video is using
     // Java API 1.8
     compileOptions {
+        //https://developer.android.com/studio/write/java8-support
+        coreLibraryDesugaringEnabled true
+
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
@@ -201,6 +204,8 @@ androidExtensions {
 }
 
 dependencies {
+    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.1.5'
+    
     implementation project(path:':libs:stories-android:stories')
     testImplementation project(path:':photoeditor')
     implementation project(path:':libs:image-editor::ImageEditor')
--- a/WordPress/build.gradle
+++ b/WordPress/build.gradle
@@ -87,9 +87,6 @@ android {
     // Gutenberg's dependency - react-native-video is using
     // Java API 1.8
     compileOptions {
-        //https://developer.android.com/studio/write/java8-support
-        coreLibraryDesugaringEnabled true
-
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
@@ -204,8 +201,6 @@ androidExtensions {
 }
 
 dependencies {
-    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.1.5'
-    
     implementation project(path:':libs:stories-android:stories')
     testImplementation project(path:':photoeditor')
     implementation project(path:':libs:image-editor::ImageEditor')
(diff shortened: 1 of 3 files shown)


renanferrari (commented):
Thanks for taking a look at this!

After doing a little bit of testing, I'm not sure if these changes actually solve the problem.

First, regarding the use of both `enterAlways` and `enterAlwaysCollapsed` flags at the same time, I believe this is fine according to the official documentation, which [describes `enterAlwaysCollapsed` as "an additional flag" to `enterAlways`](https://developer.android.com/reference/com/google/android/material/appbar/AppBarLayout.LayoutParams#SCROLL_FLAG_ENTER_ALWAYS_COLLAPSED). 

To clarify, the expected behavior here, as far as I understand, is for the `CollapsingToolbarLayout` to show up as soon as the scrolling starts (alongside the `TabLayout`) but to be expanded only when we reach the top of the scrolling content inside the `ViewPager`. This can be demonstrated by the Notifications screen, which uses a similar layout configuration and `enterAlways|enterAlwaysCollapsed`:

notifications-1


Second, looking at the video and description on the original issue, my understanding is that the undesirable behavior happens when trying to "fling" the content. Again, we can clearly see the difference in behavior when comparing with the Notifications screen:

|Notifications|Reader|
|---|---|
|notifications-2|following|

Interestingly, it seems that other tabs in the Reader actually behave as expected, which makes me think this is a problem with the "Following" tab specifically. Here's an example with the "Discover" tab:

discover

I encourage you to take a second look at this, but if it looks like it's going to be too much trouble, then we should probably leave this for a HACK Week, Groundskeeping, or something like that, as this doesn't have a high priority at the moment. Let me know what you think.

ravishanker:
> 
> I encourage you to take a second look at this, but if it looks like it's going to be too much trouble, then we should probably leave this for a HACK Week, Groundskeeping, or something like that, as this doesn't have a high priority at the moment. Let me know what you think.

@renanferrari,  Thanks for taking deeper look into the issue.  You're right , "Discover" tab doesn't have smooth scrolling issue.  However, "Following" tab has that due to its use of a legacy custom FilteredRecylerView and Filter tab inside it that is sticky.  I've made a minor tweak by disabling the flag on FilteredRecyclerView.  However, due to the filter that is designed to be sticky at the top, it will probably will never feel like Discover tab I'm afraid.  This is the best we can achieve for now without attempting to rewrite the legacy ReaderPostListFragment.java that is ~ 3K lines!

ravishanker:
@renanferrari I've tried the `exitUntilCollapsed` It seems to feel better and works like iOS one.  However now the top header doesn't scroll away as before (and same as iOS where it doesn't scroll away too).
I've also reverted all other changes limiting it to only this change now.  Please review again and if you're happy, merge it.  Thanks

renanferrari (approved):
Thanks for taking care of this, @ravishanker! 

LGTM 👍

PR #14651, opened 2021-05-14
Update intent flags for jetpack app

zwarm:
Fixes #14649 

This PR adds new intent flags when launching LoginActivity for Jetpack. 

To test:
Prep
- Download this branch
- Build a Jetpack flavor

Scenario 1
- Login with an account that doesn't have any Jetpack sites.
- When the No Sites empty screen is show, click the Try with another account button.
- Login with an account that does have at least one Jetpack site.
- Logout from that account.
- Login again with the first account that doesn't have any Jetpack sites.
- Note that the "No Sites empty view" is shown 

Scenario 2
- Login with an account that doesn't have any Jetpack sites.
- When the No Sites empty screen is show, tap the back button or swipe back
- Note that you are taken out of the app and not shown the Me view

## Regression Notes
1. Potential unintended areas of impact 🟢 
2. What I did to test those areas of impact (or what existing automated tests I relied on)
Manual testing and reran `LoginViewModelTest` and `LoginNoSitesViewModelTest`
3. What automated tests I added (or what prevented me from doing so) 🟢 

PR submission checklist:

- [X] I have completed the Regression Notes.
- [X] I have considered adding accessibility improvements for my changes.
- [X] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #14644, opened 2021-05-14
Jetpack App (Basics): Update magic link scheme

ashiagr:
SubTask #14620

This PR updates the magic link scheme for the Jetpack app. It is configured using the `build.gradle` based on the selected app.

Corresponding FluxC PR: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2003

To test:

- Install and launch the Jetpack app.
- Select "Login with WordPress.com".
- Enter email and click Continue.
- Click "Get a login link by email".
- Check email (it should be Jetpack-ified via diff `D61257`) and click "Log in to the app".
- Notice that the link opens the Jetpack app and login is successful.

Merge Instructions:

1. Wait for merge till corresponding FluxC PR is merged to develop.
2. Create FluxC tag and update it in the build.gradle (I can take care of it once this PR is approved).
3. Remove Not Ready for Merge.
4. Merge the PR.

## Regression Notes
1. Potential unintended areas of impact
WordPress app

2. What I did to test those areas of impact (or what existing automated tests I relied on)
Tested manually that "Get a login link by email" uses the `wordpress` scheme and opens the WordPress app from the email.

3. What automated tests I added (or what prevented me from doing so)
Added tests to check the scheme for both WordPress and Jetpack app.

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #14620, opened 2021-05-11
Jetpack App (Basics): Handle Magic Links


zwarm:
Parent #14293 

- [x] Update magic-link scheme for Jetpack app. @ashiagr #14644
- [x] Handle no site scenario for login via magic link. @ashiagr  #14662

leandroalonso:
JFYI: when requesting the magic link email the `scheme` property should be `jetpack`. Ie.:

```
curl https://public-api.wordpress.com/rest/v1.1/auth/send-login-email -H 'Content-Type: application/json' -d '{ "email": "$YOUR_WPCOM_EMAIL_HERE$", "client_id": $YOUR_CLIENT_ID$, "client_secret": "$YOUR_CLIENT_SECRET$", "scheme": "jetpack" }'
```

ashiagr:
Closed via #14662

PR #14662, opened 2021-05-17
Jetpack App (Basics): Magic links handle no sites

ashiagr:
SubTask #14620

### Description

When no sites are found after login via a magic link, the WordPress app displays a post-signup interstitial screen (see Epilogue Screen FieldGuide: `PCYsg-wY8`). 

> If the user doesn’t have any sites, they might see the Post-Signup Interstitial screen at the end of the flow, instead of the Epilogue screen (or after it, in some cases). This screen is only shown once and is only shown at the end of the primary flow. Despite its name, it is shown at the end of both the Signup and Login flows.

Once this screen is dismissed (using NOT RIGHT NOW option), the "My Site" tab for empty sites gets displayed. 

Similar behavior was seen for the Jetpack app (see: https://github.com/wordpress-mobile/WordPress-Android/pull/14644#pullrequestreview-659663043). 

This PR fixes this behavior and displays no jetpack sites screen for this scenario.

**Before** 

Post Login Interstitial | Add Site
----------------------|--------
post insterstitial | add new site

**After**
nosites

To test:

Jetpack App

1. Install and launch the Jetpack app.
2. Select "Login with WordPress.com".
3. Enter email (for an account with no jetpack sites) and click Continue.
4. Click "Get a login link by email".
5. Check email and click "Log in to the app".
6. Notice that the link opens the Jetpack app and no jetpack sites screen is shown.

WordPress App

1. Install and launch the WordPress app.
2. Select "Log in or sign up with WordPress.com".
3. Enter email (for an account with no sites) and click Continue.
4. Click "Get a login link by email".
5. Check email and click "Log in to the app".
6. Notice that the link opens the WordPress app and the post-login interstitial screen is displayed.

## Regression Notes
WordPress app

Tested WordPress app manually for magic link login (for accounts with sites and without sites)

Added unit tests for both WordPress app and Jetpack app for no site scenario on magic link login.

### Other Notes

1. ~~A flashing issue on the `LoginNoSitesFragment` screen is handled in PR #14651. This PR is not merged yet and you might see the issue in this PR.~~ Fixed with 9bc8471
2. Noticed a WordPress logo notification on the login Epilogue screen (after login with password for an account with no jetpack sites). On clicking this notification while the app is minimized, the "My Site" tab for empty sites gets displayed. Adding it as a known issue in the main task.

PR submission checklist:

### DIFF
--- a/WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginEpilogueFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginEpilogueFragment.java
@@ -29,6 +29,7 @@ import org.wordpress.android.ui.main.SitePickerAdapter;
 import org.wordpress.android.ui.main.SitePickerAdapter.SiteList;
 import org.wordpress.android.ui.main.SitePickerAdapter.SitePickerMode;
 import org.wordpress.android.ui.main.SitePickerAdapter.ViewHolderHandler;
+import org.wordpress.android.util.BuildConfigWrapper;
 import org.wordpress.android.util.StringUtils;
 import org.wordpress.android.util.analytics.AnalyticsUtils;
 import org.wordpress.android.util.image.ImageManager;
@@ -56,6 +57,7 @@ public class LoginEpilogueFragment extends LoginBaseFormFragment<LoginEpilogueLi
 
     @Inject ImageManager mImageManager;
     @Inject UnifiedLoginTracker mUnifiedLoginTracker;
+    @Inject BuildConfigWrapper mBuildConfigWrapper;
 
     public static LoginEpilogueFragment newInstance(boolean doLoginUpdate, boolean showAndReturn,
                                                     ArrayList<Integer> oldSitesIds) {
@@ -213,7 +215,7 @@ public class LoginEpilogueFragment extends LoginBaseFormFragment<LoginEpilogueLi
 
     @Override
     protected boolean isJetpackAppLogin() {
-        return mDoLoginUpdate && BuildConfig.IS_JETPACK_APP;
+        return mDoLoginUpdate && mBuildConfigWrapper.isJetpackApp();
     }
 
     private void bindHeaderViewHolder(LoginHeaderViewHolder holder, SiteList sites) {
--- a/WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginEpilogueFragment.java
+++ b/WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginEpilogueFragment.java
@@ -211,6 +211,11 @@ public class LoginEpilogueFragment extends LoginBaseFormFragment<LoginEpilogueLi
         }
     }
 
+    @Override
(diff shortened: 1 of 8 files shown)


zwarm (changes requested):
@ashiagr - I followed the test steps, but it is not working as I expect. I am able to login to Jetpack app with no jetpack sites, but I do not see the No Sites error view. I can repeat this process. Maybe I am testing wrong 🤷 

Steps
- Install branch
- Tapped Log in with WordPress.com
- Entered email address for WP site with no jetpack sites
- Clicked the Get a login link by email
- Open email and clicked link in password
- Shown the Login Epilogue view 
- Tapped done 
- I am then able to navigate around the app with an account that has no Jetpack Sites 

I poked around _very, very_ quickly in debug mode ... I see `LoginBaseFormFragment.onSiteChanged` getting called, which triggers the `onLoginFinished`, which kicks off the request to get sites. However when `LoginBaseFormFragment.isJetpackAppLogin` is called, `LoginEpilogueActivity` is not an instance of `LoginListener`, so the `isJetpackAppLogin()` returns false - which then triggers the call for sites without the Jetpack flag. 

I made sure I wasn't using local FluxC too.

ashiagr:
Nice catch @zwarm! I was testing with an account having no sites (not even wp sites) which didn't catch this one. 

I'm now returning true for `isJetpackAppLogin()` in the `LoginEpilogueFragment` in 79c9e69 for this scenario. Let me know what you think about it or if you still see this issue.

PS: There's a method visibility change in the login lib in af0cf4c. This PR can be merged to develop, I'll create a separate PR for merging it to `wordpress-mobile/WordPress-Login-Flow-Android`.

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginEpilogueFragment.java
```diff
@@ -211,6 +211,11 @@ public void onResume() {
         }
     }
 
+    @Override
+    protected boolean isJetpackAppLogin() {
+        return mDoLoginUpdate && BuildConfig.IS_JETPACK_APP;
```
zwarm: @ashiagr - This is only a suggestion, so feel free to ignore. Should we inject ConfigWrapper instead of direct access to BuildConfig? I go back & forth when we talk about Fragments - since we generally don't unit test them.

zwarm (approved):
@ashiagr - YAY, works like a charm. 🥇

WordPress/src/main/java/org/wordpress/android/ui/accounts/login/LoginEpilogueFragment.java
ashiagr: Right, we don't test fragments as such, so I didn't use the ConfigWrapper. I'm actually indifferent to this and can use either of the options. 

Suggestion applied in df37ee4 🙂.

zwarm: @ashiagr - thanks for the change.(Even though you didn't have to do it) :)

PR #14664, opened 2021-05-17
Merge release 17.3 into develop

JavonDavis:


WordPress/build.gradle
```diff
@@ -120,9 +120,9 @@ android {
             dimension "buildType"
             // Only set the release version if one isn't provided
             if (!project.hasProperty("versionName")) {
-                versionName "17.3-rc-1"
+                versionName "17.3"
             }
-            versionCode 1042
+            versionCode 1044
```
AliSoftware, line 125: what happened to 1043?

fastlane/Fastfile
```diff
@@ -273,7 +273,6 @@ REPOSITORY_NAME="WordPress-Android"
     android_finalize_prechecks(options)
     configure_apply(force: is_ci)
     hotfix = android_current_branch_is_hotfix
-    android_update_metadata(options) unless hotfix
```
AliSoftware, line 276: Any reason why you removed that line, i.e why we don't download the strings in `finalize_release` anymore? Seems we could easily missed some of the latest translations since last beta if we remove that line…

AliSoftware: In fact, checking out your branch and running `bundle exec fastlane download_translations` shows that there are quite a lot of strings that have been updated since but are not in this PR…

AliSoftware (commented):
I think we're missing some of the latest translations there

fastlane/Fastfile
JavonDavis: Ahhhh! thanks for catching this this was not suppose to be removed but replaced with the new download_translations lane

JavonDavis:
> I think we're missing some of the latest translations there

I think these translations changes you're seeing came in after the fact, due to issues being resolved with the translators

WordPress/build.gradle
JavonDavis: That is alpha-293, I'll have to take a look at the history here to see if there was an issue in conflict resolution when merging into develop

JavonDavis: oof! @AliSoftware  it's here https://github.com/wordpress-mobile/WordPress-Android/blob/84f3af53c47d67ca432c3de954ed77fa1e918e1b/WordPress/build.gradle#L65

AliSoftware: Ah good catch :+1: all good then

fastlane/Fastfile
jkmassel: @JavonDavis – does this mean we missed new translations in the most recent build?

jkmassel (approved):
I left one question that we'll need to sort out, but approving this so we can start the freeze

fastlane/Fastfile
AliSoftware: I'll let Javon reply in details and confirm, but yes we missed some indeed in the build submitted on Friday. Which I believe have since been fixed via 8be28f9 then https://github.com/wordpress-mobile/WordPress-Android/pull/14664/commits/6bbc145cd61c50b94e56aab1b00540bd9738c79e to re-download the latest translations 2 hours ago, which means that this PR should now be up-to-date with those latest translations, but that we'll still require to submit a new build for 17.3 and go thru Google validation process again before being able to start rollout iinm.

JavonDavis: Yep  @jkmassel after looking into this more with Olivier we did miss some of the new strings and addressed that in https://github.com/wordpress-mobile/WordPress-Android/pull/14664/commits/6bbc145cd61c50b94e56aab1b00540bd9738c79e and a new build submitted to Google

PR #14674, opened 2021-05-18
Merge release 17.4 into develop

JavonDavis:


mokagio (approved):
I was confused by GitHub not expanding the `values/strings.xml` and showing only red in the header, but looking at them I can see that there are both deletions (a lot) and new lines too 👍 

LGTM :shipit:

PR #14679, opened 2021-05-19
Issue/add unified comments feature flag

khaykov:
This PR adds feature flag and an entry point to the comment list for Unified Comment experience project.

In addition, legacy classes are deprecated with an explanatory message.

[Image from Gyazo](https://gyazo.com/9cb38e2a4c113f80be10670179823718)

Note: I added a detekt suppress method [here](https://github.com/wordpress-mobile/WordPress-Android/pull/14679/files#diff-9dfb54ffb216bc52a271e3148a25e73b0dbb41ca52e5cf0fba909f984e4cadaeR16) since the feature flag is temporary.

To test:
- Install `vanilla` variant of the app, and confirm that you do not see Unified Comments item on MySite screen.
- Install `wasabi` variant of the app, and confirm that you DO see Unified Comments, and clicking on it opens an empty screen.

Test with ImprovedMySiteFragment:
- Add `buildConfigField "boolean", "MY_SITE_IMPROVEMENTS", "true"` to `wasabi` build variant and install the app.
- Confirm that you DO see Unified Comments, and clicking on it opens an empty screen.

## Regression Notes
- Unified Comment row is visible in Vanilla variant.

- Manual and automatic testing.

- Added couple of automated tests to code that powers `ImprovedMySiteFragment`
PR submission checklist:


WordPress/build.gradle
```diff
@@ -82,6 +82,7 @@ android {
         buildConfigField "boolean", "IS_JETPACK_APP", "false"
         buildConfigField "String", "TRACKS_EVENT_PREFIX", '"wpandroid_"'
         buildConfigField "String", "PUSH_NOTIFICATIONS_APP_KEY", '"org.wordpress.android"'
+        buildConfigField "boolean", "UNIFIED_COMMENTS", "false"
```
develric: Not sure but was wondering if it could make sense to have 3 separated flags (1 per phase); in that case maybe we should reflect this in naming, wdyt?

WordPress/src/main/java/org/wordpress/android/ui/comments/CommentActions.java
```diff
@@ -9,8 +9,12 @@
  * actions related to comments - replies, moderating, etc.
  * methods below do network calls in the background & update local DB upon success
  * all methods below MUST be called from UI thread
+ *
+ * @deprecated
+ * Comments are being refactored as part of Comments Unification project. If you want to add any features or modify
```
develric: >If you want to add any features or modify

I think this can be just my bad English and in case I trust yours 😄 ...maybe we could say something like `If you are adding any features or modifying...`...sounds a bit less `you need to ask our permission` (just my 2 cents though 🙇 )

WordPress/src/test/java/org/wordpress/android/ui/mysite/SiteListItemBuilderTest.kt
```diff
@@ -430,6 +430,30 @@ class SiteListItemBuilderTest {
         assertThat(item).isNull()
     }
 
+    @Test
+    fun `unified comments item built if is the feature flag is enabled`() {
```
develric: >if is the feature flag is enabled

I guess the first `is` is by copy paste

WordPress/src/main/res/values/strings.xml
```diff
@@ -2108,6 +2108,7 @@
     <string name="my_site_btn_sharing">Sharing</string>
     <string name="my_site_btn_site_settings">Settings</string>
     <string name="my_site_btn_comments" translatable="false">@string/comments</string>
+    <string name="my_site_btn_unified_comments" translatable="false">Unified Comments</string>
```
develric, line 2110: I guess this name is temporary (otherwise we should remove the `translatable="false"` maybe), to differentiate between the two list items (legacy and new) but I wonder if we should just keep the menu and separate the call paths (opening the old and new activity) only inside the common Comments button callback. In this way we do not need to add an additional list item and do not need to come back renaming this button with the final `Comments` name. Let me know wdyt, I'm fine also with keeping this as it is 🙇 .

develric (commented):
Hey @khaykov 👋 , checked this one and it's looking good 👍  (code and test wise). Just left a few minor comments in case you think they can be worth. Let me know wdyt 🙇 .

WordPress/build.gradle
khaykov: Sounds good 👍

WordPress/src/main/java/org/wordpress/android/ui/comments/CommentActions.java
khaykov: Agree, your variant sounds better :)

WordPress/src/main/res/values/strings.xml
khaykov: I wanted to give users of wasabi version an option to check regular comments without toggling a feature flag - some folks might not know about it, and changing the flag adds an extra friction. This also gives us an opportunity to check both comment list without toggling feature flag or having a second device/emulator open.

khaykov:
Thanks for the review, @develric ! Hope I addressed all your comments :)

develric (approved):
Thanks for the changes and clarifications @khaykov 🙇 . I think this is ready to go 👍 !

